PR #231, opened 2017-09-02
[Destruction] Initial implementation

sMteX:
Added some basic things - ABC module, Cast Efficiency, Cooldown Tracker (albeit incomplete but that's more of a technical limitation than my fault), Destro spells. 

Soul Shard tracking is left commented, I'll have to tailor it specifically for Destruction, but I intend on having something similar to Affliction so I didn't delete it straight away. Will do tomorrow.


src/Parser/DestructionWarlock/Modules/Features/AlwaysBeCasting.js
```diff
@@ -0,0 +1,79 @@
...
+    SPELLS.SUMMON_INFERNAL_TALENTED.id,
+    SPELLS.DARK_PACT.id,
+    SPELLS.GRIMOIRE_OF_SACRIFICE_BUFF.id,
+    SPELLS.SHADOWFURY_TALENT.id,
+    SPELLS.GRIMOIRE_VOIDWALKER.id,
+    SPELLS.GRIMOIRE_FELHUNTER.id,
+    SPELLS.GRIMOIRE_SUCCUBUS.id,
+  ];
+
+  on_initialized() {
```
MartijnHols: This does nothing and could be omitted.

src/Parser/DestructionWarlock/Modules/Features/CastEfficiency.js
```diff
@@ -0,0 +1,350 @@
...
+
+class CastEfficiency extends CoreCastEfficiency {
+  static CPM_ABILITIES = [
+    ...CoreCastEfficiency.CPM_ABILITIES,
+    //Rotational spells
+    {
+      spell: SPELLS.CHAOS_BOLT,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+      recommendedCastEfficiency: 0.95,
```
MartijnHols: Cast efficiency can't really be set for abilities with no cooldown.

src/Parser/DestructionWarlock/Modules/Features/CooldownTracker.js
```diff
@@ -0,0 +1,56 @@
...
+class CooldownTracker extends CoreCooldownTracker {
+  static cooldownSpells = [
+    ...CooldownTracker.cooldownSpells,
+    {
+      spell: SPELLS.SOUL_HARVEST,
+      summary: [
+        BUILT_IN_SUMMARY_TYPES.DAMAGE,
+      ],
+    },
+    //TODO: see if we can add Havoc (it's not a buff on player, it's a debuff on enemies)
```
MartijnHols: You can have your own event handlers for activating cooldowns, see the core CooldownTracker's applybuff/removebuff events for info about how to do this. But seeing as it's a debuff it's probably only interesting to see the damage / debuffs done to this target instead of all available enemies, filtering on that might be more challenging (especially to do it with clean code) but can definitely be done.

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardDetails.js
```diff
@@ -0,0 +1,66 @@
...
+const soulShardIcon = 'inv_misc_gem_amethyst_02';
+
+class SoulShardDetails extends Module {
+  static dependencies = {
+    soulShardTracker: SoulShardTracker,
+  };
+
+  suggestions(when) {
+    const shardsWasted = this.soulShardTracker.shardsWasted;
+    when(shardsWasted).isGreaterThan(0)
```
MartijnHols: If possible please try to use percentage or per minute numbers for triggers so that something like wasting 10 on a 3 minute fight can get a higher prio than wasting 10 on a 15 minute fight. I recently wrote guidelines for suggestions, here: https://github.com/MartijnHols/WoWAnalyzer/blob/master/CONTRIBUTING.SPEC.md#suggestions

MartijnHols:
Thanks, see comments for optional feedback.

src/Parser/DestructionWarlock/Modules/Features/AlwaysBeCasting.js
sMteX: Didn't see it in the first place, same problem will probably be in Affliction ABC too (I copied), will delete.

src/Parser/DestructionWarlock/Modules/Features/CastEfficiency.js
sMteX: Ah, fair enough, that makes sense.

src/Parser/DestructionWarlock/Modules/Features/CooldownTracker.js
sMteX: I'll leave it as a future TODO, would be very nice to see how much damage I dealt to the target in that 10 second window, but seeing it would be difficult I'll leave it till later.

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardDetails.js
sMteX: My situation with this is pretty much identical to Iskalla's problem, and I'm torn apart as to how it should be. I agree with you that it should be somewhat proportionate to the fight length, but also that really no shard should be wasted regardless of the fight length. I'll try to change it but the thresholds will be tricky for both.

PR #233, opened 2017-09-02
[BrM] Added energy wasted

shighman:
Added energy wasted with tests

The energy is in its own js file because the next pr will include another statistic which uses it, committing this first one to keep it small.

Also removed Brewmaster from front of changelog, they know... lol


shighman:
This is the issue with taking so long to get a PR after writing the code, I was planning on converting the statistic to a percentage. Let me finish that before you review.

shighman:
This energy suggestion is not showing what i am trying to show, because energy is constantly regenerating doing a threshold i don't think delivers what i am trying to show.

src/Parser/BrewmasterMonk/Constants.js
```diff
@@ -26,3 +26,22 @@ export const SPELLS_WHICH_REMOVE_BOC = [
   SPELLS.IRONSKIN_BREW.id,
   SPELLS.PURIFYING_BREW.id,
 ];
+
+export const SPELLS_WITH_ENERGY_COST = {
```
MartijnHols: Specifying the `energyCost` in the spells object is preferred (and already available that way for talents), see `manaCost` for examples.

src/Parser/BrewmasterMonk/Modules/Features/Energy.js
```diff
@@ -0,0 +1,125 @@
...
+      }
+      this.addEnergyEvent(event, currentEnergy, SPELLS_WITH_ENERGY_COST[spellId]);
+    }
+  }
+
+  on_byPlayer_applybuff(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS_WITH_HASTE[spellId] !== undefined) {
+      debug && console.log('hero');
+      this.currentRegen = this.calcEnergyRegen([this.baseHaste, SPELLS_WITH_HASTE[spellId] / 100], event.timestamp);
```
MartijnHols: This only works with *one* ability at a time and this code is kinda hard to follow. Could this be refactored a bit? I think if you made `calcEnergyRegen` either a completely pure method or without any return values it would already be a fair bit easier to follow.

If you're interested in the currentHaste it might be best to put a dependency on AlwaysBeCasting and use its `currentHaste` property.

src/Parser/BrewmasterMonk/Modules/Features/Energy.js
```diff
@@ -0,0 +1,125 @@
...
+const debug = false;
+const BASE_REGEN = 10;
+export const MAX_ENERGY = 100;
+
+class Energy extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  energy = [];
```
MartijnHols: Those are a lot of public properties, this is generally a good indicator that some refactoring is needed.

src/Parser/BrewmasterMonk/Modules/Features/EnergyWaste.js
```diff
@@ -0,0 +1,53 @@
...
+  static dependencies = {
+    energy: Energy,
+    combatants: Combatants,
+  };
+
+  suggestions(when) {
+    const energyWastedFromBobPerc = (this.energy.energyWastedFromBOB / this.energy.bobCasts) / MAX_ENERGY;
+    
+    this.energy.bobCasts > 0 && 
+    when(energyWastedFromBobPerc).isGreaterThan(0.4)
```
MartijnHols: Shouldn't this be 0 by itself when bobCasts is 0? (I'd like to avoid optional suggestions outside of due to not having a talent/item/other static setup)

src/tests/Parser/Brewmaster/Energy.test.js
```diff
@@ -0,0 +1,112 @@
+import Energy from 'Parser/BrewmasterMonk/Modules/Features/Energy';
+import { processEvents } from './Fixtures/processEvents';
+import { EnergyFight, buffs } from './Fixtures/EnergyFight';
+
+describe('Brewmaster.Energy', () => {
```
MartijnHols: I'd prefer if the tests were based around requirements rather than per function; I think this would catch way more issues. For instance if you test `it('stacks multiple Haste gains properly', ...)` with the public API you could catch that `applybuff` doesn't do this right now.

src/tests/Parser/Brewmaster/Energy.test.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+  it('energy regen with 10% haste', () => {
+    const val = energy.calcEnergyRegen([0.1]);
+    expect(val).toBe(11);
+  });
+  it('energy regen with 1.52% haste', () => {
+    const val = energy.calcEnergyRegen([0.0152]).toFixed(2);
+    expect(val).toBe('10.15');
+  });
+  it('energy regen with 5.28% haste', () => {
```
MartijnHols: Running several tests with only values different; I think it would be cleaner to put this in an array and run a loop.

src/tests/Parser/Brewmaster/Energy.test.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+  it('energy regen with -10% haste', () => {
+    const val = energy.calcEnergyRegen([-0.1]);
+    expect(val).toBe(9);
+  });
+  it('energy regen with bad parameter (int) value haste sets as base value', () => {
+    const val = energy.calcEnergyRegen(1);
+    expect(val).toBe(10);
+  });
+  it('energy regen with bad parameter (string) value haste sets as base value', () => {
```
MartijnHols: Putting multiple assertions in a single test case would be preferred.

PR #239, opened 2017-09-03
(Holy Priest) Echo of Light (aka Mastery) breakdown

enragednuke:
Finally got to writing the module for the Echo of Light breakdown. As this isn't something that is available through WCL, it's been highly requested in respect to other things.

Also fixes a bug with the Expandable Statistic Box that was causing it to span across multiple columns unexpectedly.

image

edit: the columns above look the same but the 3rd column is now % of total and the 4th is % overheal instead.

### DIFF
--- a/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
+++ b/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
@@ -145,7 +145,7 @@ class MasteryBreakdown extends Module {
         icon={<SpellIcon id={SPELLS.ECHO_OF_LIGHT.id} />}
         value={`${formatNumber(this.healing)}`}
         label={(
-          <dfn data-tip={``}>
+          <dfn data-tip={`Echo of Light healing breakdown. As our mastery is often very finicky, this could end up wrong in various situations. Please report any logs that seem strange to @enragednuke on the WoWAnalyzer discord.`}>
             Echo of Light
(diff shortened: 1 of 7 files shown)


MartijnHols:
Thanks for the gif, very helpful. The list seems very long which breaks the layout a bit, what do you think about moving the data to a new tab instead?

src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
```diff
@@ -0,0 +1,200 @@
...
+    for (spell in this.effectiveHealDist) {
+      eHDPerc[spell] = this.effectiveHealDist[spell] / total;
+    }
+
+    const eOHD = {};
+    for (spell in this.effectiveHealDist) {
+      eOHD[spell] = this.effectiveOverhealDist[spell] / (this.effectiveHealDist[spell] + this.effectiveOverhealDist[spell]);
+    }
+
+    // Since JS objects lack order, we need to convert our dictionary to an array
```
MartijnHols: If it's not necessary for the data handling it's customary to wait with ordering data until the rendering so that the renderer can decide how the data should be sorted (and perhaps allow users to alter how a table is sorted).

enragednuke:
So using a tab at the bottom instead? That could work too, not opposed to that. As for the ordering issue, is there an example of this elsewhere in the code that I could base that off of?

PR #242, opened 2017-09-04
[Destruction] Fixed Summons and Grimoires 

sMteX:
Problem with these abilities is the shared cooldown, so I deleted the cooldown from CastEfficiency (so it still shows number of casts but doesn't interfere with suggestions) and added 2 separate modules that take care of combined casts. Also added module for unused Lord of Flames buffs.

In addition, I rewrote both parsers to use Combatants as dependency instead of accessing it via CombatLogParser, cleaned up few things and reworked wasted shards to be per minute (thresholds may be off right now but that can be now tweaked around). _Note:_ Destruction shard tracking is still not functional but I've changed it too so I don't forget about it.


rp4rk:
Looks good, this PR is a bit large. A few of the more maintenance-ish things such as the combatants dependency declarations could have been a separate PR.

PR #243, opened 2017-09-04
[Resto Druid] Added T19 2pc module

kfinch:
Added an item display for Resto Druid's T19 2pc bonus. Added module is basically just the display, all the backend work was already doing in the Mastery module.


MartijnHols:
Thanks!

PR #246, opened 2017-09-04
[FERAL] Added Cooldown tracker and Bleed uptimes

Thieseract:
Configured the cooldown tracker and added modules for bleed/dot uptimes.


MartijnHols:
Thanks!

PR #248, opened 2017-09-05
[Destruction] Added Destruction Soul Shard usage breakdown

sMteX:
I'm sorry this PR is bigger than I promised but I wanted to roll the changes out all at once so it's finished. 

I wish I could do the FRAGMENT_GENERATING_ABILITIES in the same manner as FRAGMENT_SPENDING_ABILITIES in SoulShardEvents (would get rid of the switch in _damage()), but Incinerate really complicates things as it relies on Enemies and Combatants module.


src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+        name: SPELLS[spellId].name,
+      },
+      targetID: event.targetID,
+      targetInstance: event.targetInstance,
+    };
+    let rawGain;
+    let gain = 0;
+    let waste = 0;
+
+    switch (spellId) {
```
rp4rk: Purely opinion here, but I think object literals are easier to use and read for this,

```javascript
const SHARD_GENERATION = {
  [SPELLS.CONFLAGRATE.id]: 5,
  ...
};

const rawGain = SHARD_GENERATION[spellid];
```

Not only are you not overriding the value of `rawGain`, but you also have an easy to maintain dictionary of shard generation values. This isn't a blocker.

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+      case SPELLS.IMMOLATE_DEBUFF.id:
+        rawGain = 1;
+        break;
+      case SPELLS.CONFLAGRATE.id:
+        rawGain = 5;
+        break;
+      case SPELLS.INCINERATE.id:
```
rp4rk: This changes things up a little from the previous suggestion:

```javascript
const SHARD_GENERATION = {
  [SPELLS.CONFLAGRATE.id]: () => 5,
  [SPELLS.INCINERATE.id]: () => {
    // Return shard generation here
  },
  ...
};

const rawGain = SHARD_GENERATION[spellid]();
```

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+      timestamp: event.timestamp,
+      type: 'soulshardfragment_gained',
+      ability: {
+        guid: spellId,
+        name: SPELLS[spellId].name,
+      },
+      targetID: event.targetID,
+      targetInstance: event.targetInstance,
+    };
+    let rawGain;
```
rp4rk: Maybe change `rawGain` to `gainBeforeCap` or `grossGain`?

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+    const spellId = event.ability.guid;
+    if (spellId === SPELLS.DIMENSIONAL_RIFT_CAST.id) {
+      this.processDimensionalRift(event);
+    }
+    else if (FRAGMENT_SPENDING_ABILITIES[spellId]) {
+      this.processSpenders(event);
+    }
+  }
+
+  //in order to de-clutter on_byPlayer_cast()
```
rp4rk: This comment isn't needed, but separating here is a good thing to do.

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+      type: 'soulshardfragment_gained',
+      ability: {
+        guid: spellId,
+        name: SPELLS[spellId].name,
+      },
+      targetID: event.targetID,
+      targetInstance: event.targetInstance,
+    };
+    const rawGain = 3;
+    let gain = 0;
```
rp4rk: Seems like a lot of repeating logic here from the previous method, maybe refactor this to something more generic?

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
```diff
@@ -0,0 +1,254 @@
...
+const FRAGMENTS_PER_SHARD = 10;
+
+const FRAGMENT_GENERATING_ABILITIES = [
+  SPELLS.IMMOLATE_DEBUFF.id,
+  SPELLS.CONFLAGRATE.id,
+  SPELLS.INCINERATE.id,
+  //Dimensional Rift would also be here, but that is evaluated on cast, not damage and would proc a lot of times
+];
+
+const FRAGMENT_SPENDING_ABILITIES  = {
```
rp4rk: This is also defined in SoulShardTracker, is there a reason it cannot be shared?

sMteX: That's true, I'll try to look at if something couldn't be factored out (but not sure about this, since I'm using those as object literals here, with fragments associated, whereas in Tracker, I only care about the ID, and I get the fragments from the event)

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
sMteX: Most likely just a cosmetic change but I can sure change that to be a little less misleading

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
sMteX: Yes, exactly this I had in mind, I just don't know if I could access the 2 dependent modules from inside the function (I need to look for T20 2p buff on the player, Havoc debuff on target and also if the event (if it critted) in order to determine how many fragments to add).

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
sMteX: I'll try to pull it out into processGenerators() or something like that, I agree that a lot is repeating there.

src/Parser/DestructionWarlock/Modules/SoulShards/SoulShardEvents.js
rp4rk: These short hand functions `() => x` will be bound to the parent scope here instead of `SHARD_GENERATION`, if the syntax was:

```javascript
const SHARD_GENERATION = {
  [SPELLS.INCINERATE.id]: function() {
    this.module...
  }
};
```
your concern would be entirely correct, as `this` will refer to the object you've defined it in.

---
I'd recommend [YDKJS - Scope and Closures](https://github.com/getify/You-Dont-Know-JS/tree/master/scope%20%26%20closures) as some reading if you want to become certain on these types of thoughts.

sMteX: Ah, I see.. but I still don't have an information about critting.. but I suppose I could do something like this:
`
const SHARD_GENERATION = {
   [SPELLS.CONFLAGRATE.id]: (isCrit) => 5,
   [SPELLS.INCINERATE.id]: (isCrit) => {
   },
   ...
};

const rawGain = SHARD_GENERATION[spellId](event.hitType);
`

most of the time it wouldn't be used but that doesn't matter very much, no?

rp4rk: `event` will be accessible inside this scope, so you shouldn't have a problem with that.

sMteX:
I refactored the Events a bit, leading to a cleaner source and less redundancy in the code. Even though seemingly the Tracker has some shared code with Events (the abilities), I don't think I can reduce that (at least not yet, if I will have to redefine the ID array again somewhere else, I would extract that as a constant array, but so far no need to), as in Events it is an object with associated fragments, but that's not needed in Tracker as the amount of fragments gained/spent is already in the events I'm listening to.

sMteX:
Refactored the Trackers again, leading to more polished code. Looks almost the same on the outside, shard/fragment generating effects are hidden when they generated nothing, which may or may not be desirable. I would most likely want to see that for example my affli legendary belt didn't generate any shards so I could perhaps replace it with a different legendary, but I don't want to see that when I don't have that belt. Don't know how to easily solve this without conditionals *somewhere* that checks whether I do or don't have said ability/effect.

PR #252, opened 2017-09-05
[Guardian] Clean up module dependencies

FaideWW:
Replaces most instances of `this.owner.modules` and `this.owner.selectedCombatant` in the guardian modules with the dependency injection pattern.  The remaining references are to `this.owner.fightDuration`, which does not currently have a better solution.

This PR also contains a change to the CastEfficiency of Thrash.  Previously, if Incarnation was talented, the maximum Thrash Casts were calculated based on the actual number of casts of Incarnation in the encounter.  Now, the maximum number of potential Incarnation casts is used to compute potential Thrash count, so poor Incarn cast efficiency will also lead to poor Thrash cast efficiency.  

This change should help better identify low Thrash counts, but was also necessary in order to remove the dependency on `owner.modules.abilityTracker` from within `CastEfficiency`.

### DIFF
--- a/src/Parser/GuardianDruid/Modules/Features/CastEfficiency.js
+++ b/src/Parser/GuardianDruid/Modules/Features/CastEfficiency.js
@@ -35,10 +35,9 @@ class CastEfficiency extends CoreCastEfficiency {
         if (!hasMightBuff) {
           return 6;
         }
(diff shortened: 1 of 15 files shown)


anom0ly:
Looks good! Thanks!

PR #254, opened 2017-09-05
[MM Hunter] fixed PR request

Jlassie82:
Fixed spell usage suggestion and other problem area wasn't being utilized at this point so I removed it.


anom0ly:
Looks good. Thanks!

PR #269, opened 2017-09-07
[Vengeance] Statistic mouseover events, ABC rework and code refactoring

ronaldpereira:


anom0ly:
Only a minor question on if you missed one of the refactor areas. Overall looks good though!

ronaldpereira:
I don't think I've missed any of the spell modules refactor areas. The only one I let it as icon is on the SpiritBomb.js:31, because it doesn't accept the SpellIcon module.

ronaldpereira:
Oh, now I see it. I'll compare it now. I messed up a bit with my commits :(

PR #278, opened 2017-09-08
[resto druid] Added support for T21.

kfinch:
T21 support for Resto Druids, using PTR logs to check my work. I also fixed some messed up dates in the changelog -_-

### DIFF
--- a/src/Parser/RestoDruid/CombatLogParser.js
+++ b/src/Parser/RestoDruid/CombatLogParser.js
@@ -22,9 +22,12 @@ import Sephuz from './Modules/Legendaries/Sephuz';
 import DarkTitanAdvice from './Modules/Legendaries/DarkTitanAdvice';
 import EssenceOfInfusion from './Modules/Legendaries/EssenceOfInfusion';
 import Tearstone from './Modules/Legendaries/Tearstone';
-import T20 from './Modules/Legendaries/T20';
-import T19_2Set from './Modules/Legendaries/T19_2Set';
 import DarkmoonDeckPromises from './Modules/Legendaries/DarkmoonDeckPromises';
+
+import T19_2Set from './Modules/Legendaries/T19_2Set';
+import T20 from './Modules/Legendaries/T20';
+import T21_2Set from './Modules/Legendaries/T21_2Set';
+import T21_4Set from './Modules/Legendaries/T21_4Set';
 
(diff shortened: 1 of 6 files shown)


src/Parser/RestoDruid/CombatLogParser.js
```diff
@@ -22,10 +22,13 @@ import Sephuz from './Modules/Legendaries/Sephuz';
 import DarkTitanAdvice from './Modules/Legendaries/DarkTitanAdvice';
 import EssenceOfInfusion from './Modules/Legendaries/EssenceOfInfusion';
 import Tearstone from './Modules/Legendaries/Tearstone';
-import T20 from './Modules/Legendaries/T20';
-import T19_2Set from './Modules/Legendaries/T19_2Set';
 import DarkmoonDeckPromises from './Modules/Legendaries/DarkmoonDeckPromises';
 
+import T19_2Set from './Modules/Legendaries/T19_2Set';
```
rp4rk: 👍

src/Parser/RestoDruid/CombatLogParser.js
```diff
@@ -104,8 +107,10 @@ class CombatLogParser extends CoreCombatLogParser {
     darkTitanAdvice: DarkTitanAdvice,
     essenceOfInfusion: EssenceOfInfusion,
     tearstone: Tearstone,
-    t20: T20,
     t19_2set: T19_2Set,
+    t20: T20,
+    t21_2set: T21_2Set,
```
rp4rk: I think this should be renamed `t21TwoSet` etc, snake case is creeping in here.

kfinch: All the other set bonus modules are named in this fashion, so I don't think it makes sense to change unless you want a broader refactor. Personally I prefer snake case to '2Two'.

kfinch:
For testing: Jx41dmtFw92vGT3n

src/Parser/RestoDruid/CombatLogParser.js
MartijnHols: Naming that I've used to keep it somewhat consistent and readable was `Tier21_4set`.

kfinch: So change it or keep it as is?

MartijnHols:
Thanks!

PR #279, opened 2017-09-08
[Destruction] First batch of talents

sMteX:
Added Backdraft and Roaring Blaze talents, and also support for Shadowburn. As I mentioned in latest commit, I'd love to show a statistic how much fragments did Shadowburn give over regular Conflagrate, but seems like there's no clean way of doing that (see comment in Events).

Tomorrow I'll do the refactor Zero wanted (remove `this.owner.error` from the code).


sMteX:
There might be conflicts between this PR and the refactor one I've just opened, mostly because in this PR, the `this.owner.error` is still in the code and in the other I've removed it, so hopefully it won't be as bad to merge.

sMteX:
Not overly sure about this (I don't like having ability names as keys, prefer IDs) but it seems to do exactly what I wanted. Tell me what do you think about it, if it's not a good idea, I can revert these commits and leave it combined together, I can live with it 😄

MartijnHols:
Thanks!

PR #282, opened 2017-09-08
[Destruction] Second batch of talents

sMteX:
Added tier 30 row for Destro talents - Reverse Entropy, Eradication and Empowered Life Tap. Pretty straightforward stuff. Also added the ELT module to Affli since it exists there too (albeit never used I'd say).

### DIFF
--- a/src/Parser/AfflictionWarlock/Modules/Talents/EmpoweredLifeTap.js
+++ b/src/Parser/AfflictionWarlock/Modules/Talents/EmpoweredLifeTap.js
@@ -1,0 +1,63 @@
+import React from 'react';
+
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import { formatNumber, formatPercentage } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+import getDamageBonus from '../WarlockCore/getDamageBonus';
+
+const ELT_DAMAGE_BONUS = 0.1;
+
+class EmpoweredLifeTap extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  bonusDmg = 0;
+  ELTuptime = 0;
+
+  on_initialized() {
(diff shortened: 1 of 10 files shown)


src/Parser/AfflictionWarlock/Modules/Talents/EmpoweredLifeTap.js
```diff
@@ -0,0 +1,63 @@
...
+
+const ELT_DAMAGE_BONUS = 0.1;
+
+class EmpoweredLifeTap extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  bonusDmg = 0;
+  ELTuptime = 0;
```
MartijnHols: Properties should start with lowercase characters.

src/Parser/DestructionWarlock/Modules/Talents/Eradication.js
```diff
@@ -0,0 +1,34 @@
...
+import Enemies from 'Parser/Core/Modules/Enemies';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+
+import getDamageBonus from '../WarlockCore/getDamageBonus';
+
+const ERADICATION_DAMAGE_BONUS = 0.15;
+
+//only calculates the bonus damage, output depends if we have the talent directly or via legendary finger (then it appears as either a Statistic or Item)
```
MartijnHols: I recommend showing it as a statistic always since it's basically the talent that gives you the damage, and the ring is just an indirect way of getting it. This makes the display of the talent more consistent and predictable. In the item display the only real gain is the talent. Example of how this was solved elsewhere:

https://github.com/MartijnHols/WoWAnalyzer/blob/cfe1de5908bbaa288fe0fddc7f6a5a85001631fc/src/Parser/HolyPaladin/Modules/Items/SoulOfTheHighlord.js#L22

MartijnHols:
Thanks! See the comments I left for feedback.

src/Parser/AfflictionWarlock/Modules/Talents/EmpoweredLifeTap.js
sMteX: Will fix in next PR :)

src/Parser/DestructionWarlock/Modules/Talents/Eradication.js
sMteX: Fair enough, I'll change it in next PR, along with the same issue in Affliction (Death's Embrace talent) :)

PR #283, opened 2017-09-09
[Vengeance+Havoc] Cast Efficiency rework

ronaldpereira:
Cast Efficiency is now less nitpicky: Some almost unused abilities or no CD abilities 'can be improved' warning is now disabled.

Also, corrected some getCooldown timers that was wrong in Vengeance CastEfficiency.


MartijnHols:
Thanks!

PR #284, opened 2017-09-09
Added suggestion module for high Nature's Essence overhealing.

kfinch:
I also fixed a bug where Druid modules getting Nature's Essence spellinfo were getting the Shaman spell by the same name instead. My fix was to name the druid version 'NATURES_ESSENCE_DRUID', not sure what you think about that solution.


MartijnHols:
Thanks!

PR #288, opened 2017-09-10
[Destruction] Added Fire and Brimstone + fixes

sMteX:
I fixed the suggestion from previous PR and added a module for Fire and Brimstone. 

Also added `targetID` and `targetInstance` back to the fragment events, and also a `damage` property. I should mention that I don't particularly like this solution (because it exists there even for events like shard refund, with `damage: 0` which doesn't make sense and goes kinda beyond the intended scope of the module). 

This could be worked around by tracking Incinerate damage events, and tracking the timestamp and targetID and targetInstance, but then in `on_finished()` where I count the bonus damage, I would have to make a lot of lookups by 3 variables (`timestamp`, `targetID` and `targetInstance` all need to match the current event I'm processing there) and I don't know how to efficiently do that without excessive use of loops and I thought that would be bad for performance. If anyone knows of a better solution, I'm open to suggestions.


anom0ly:
Thanks.

PR #289, opened 2017-09-10
[Destruction] Added Channel Demonfire and Havoc cleave

sMteX:
There are two things I'd like to change but don't know how to do any of them. First is to perhaps show wasted ticks of Channel Demonfire, but the problem is this:

Channel Demonfire fires 15 bolts at _random_ targets afflicted by Immolate (so far so good), but the bolts can _cleave_ targets around it. This cleave damage has the same ID as the normal damage, so we can't track by `event.ability.guid`, and we can't rely on `targetID` and `targetInstance` either, because it can cleave the targets with Immolate too (picture 2 Immolated targets next to each other and one bolt gets to one of them, the other one will take damage too but since both have Immolate, we can't tell which one was the original target). The only difference is that the cleave damage is about _half_ the normal damage, so perhaps I can push additional commit that tracks it by the damage. But then also the damage can crit, so it's like 4 different averages which complicates it a lot - cleave normal, cleave crit, direct normal, direct crit.

Problem with Havoc (it places a debuff on target that makes all single target spells to also strike the Havoc target) is simply because it doesn't generate any special events or something. There's no way to distinguish if `damage` event on target with Havoc was from a replicated spell, or if was from a direct cast at the Havoc target (but that should be discouraged, if you intend to cleave on the target, you just don't place Havoc on the same target and start hitting it directly). So instead I just counted all the damage done to a target if the target has Havoc, even if that isn't 100% accurate.


anom0ly:
Thanks.

PR #290, opened 2017-09-10
[FERAL] Changed exceptions to conditional assignment on combo point tracker

Thieseract:
Not sure if this is the correct way of doing things but it is the only way I could think to use conditional assignment and still keep the desired behavior.


MartijnHols:
Thanks!

src/Parser/FeralDruid/Modules/ComboPoints/ComboPointTracker.js
```diff
@@ -36,36 +20,39 @@ class ComboPointTracker extends Module {
   wasted = {};
 
   on_initialized() {
-    //initialize base abilities, the rest depends on talents and equip
-    pointGeneratingAbilities.forEach(x => {
+    const player = this.combatants.selected;
+
+    this.pointGeneratingAbilities = [
```
MartijnHols: I think it would be cleaner to put this property in the class with all static spells, and just have if-statements in this event listener that add spells to the array.

Thieseract: So this part would be better the way it was before? Just want to make sure I understand since I changed away from using a property with static spells and adding via if statements, to the conditional assignment setup based on feedback from rp4rk on my last PR.

MartijnHols: I meant something like this:

https://github.com/MartijnHols/WoWAnalyzer/blob/d1c857e5a5769934424d7114f0368d1ee80b30e2/src/Parser/HolyPaladin/Modules/Features/AlwaysBeCasting.js#L29-L61

So very similar to what you had, but without manually setting the gained/wasted properties. So really all it's different from the old code is that they conditionals would be above the for loops and without the this.spent/gained/wasted setting.

Thieseract: Got it, thanks

PR #291, opened 2017-09-10
[resto druid] Refactored Mastery module for ESLint

kfinch:
Changed Maps to Objects, used more functional techniques.


MartijnHols:
Thanks!

PR #292, opened 2017-09-10
[Destruction] Added Dimensional Rift breakdown

sMteX:
Really happy how that turned out :3 Now I'll move on some traits or set bonuses/legendaries. Suggestions are welcomed as always.


MartijnHols:
Thanks!

PR #294, opened 2017-09-11
(Holy Priest) Fix for mastery breakdown bug

enragednuke:
Basically, if something triggered mastery (and was not known in `CONSTANTS.js`) and nothing else known was applied within the 3 seconds prior to a mastery healing tick, it would cause the existing logic to crash. 

Also added a filter to avoid the table from being too congested when it expands to other spell ids.

### DIFF
--- a/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
+++ b/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
@@ -182,7 +182,7 @@ class MasteryBreakdown extends Module {
         {
           (
             <div>
-              Values under 0.5% of total are omitted.
+              Values under 0.5% of total (and Renew) are omitted.
             </div>
           )
         }
--- a/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
+++ b/src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
@@ -30,6 +30,7 @@ class MasteryBreakdown extends Module {
   on_finished() {
     // There's likely a far better way to do this, but my 2AM brain couldn't find it
     let total = 0;
+    delete this.effectiveHealDist[SPELLS.RENEW.id]; // see the comment at line ~150 for why this happens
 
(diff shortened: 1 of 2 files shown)


src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
```diff
@@ -136,16 +142,31 @@ class MasteryBreakdown extends Module {
         this._maxHealVal[tId] = {};
       }
 
+      let eh = event.amount + (event.absorbed || 0) + (event.overheal || 0);
```
MartijnHols: Please use descriptive var names.

src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
```diff
@@ -158,6 +179,13 @@ class MasteryBreakdown extends Module {
           </dfn>
         )}
       >
+        {
```
MartijnHols: The curly bracket and parens can be omitted.

src/Parser/HolyPriest/Modules/PriestCore/MasteryBreakdown.js
```diff
@@ -136,16 +142,31 @@ class MasteryBreakdown extends Module {
...
       } else {
-        this._healValByTargetId[tId][spellId] += event.amount + (event.absorbed || 0) + (event.overheal || 0);
+        this._healValByTargetId[tId][spellId] += eh;
       }
       this._maxHealVal[tId][spellId] = this._healValByTargetId[tId][spellId];
     }
   }
 
   statistic() {
+    const percOTHD = this.owner.getPercentageOfTotalHealingDone(this.healing);
```
MartijnHols: Please use descriptive var naming.

enragednuke: I felt it was fairly descriptive considering it's an acronym for the method call right after, but I'll tweak it.

MartijnHols:
Thanks!

PR #300, opened 2017-09-12
[FERAL] Added Savage Roar damage contribution

Thieseract:


src/Parser/FeralDruid/Modules/Talents/SavageRoarDmg.js
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  bonusDmg = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.SAVAGE_ROAR_TALENT.id);
+  }
+
+  on_damage(event) {
+    if (this.combatants.selected.hasBuff(SPELLS.SAVAGE_ROAR_TALENT.id, event.timestamp) && event.targetIsFriendly === false) {
```
MartijnHols: 1. It's generally cleaner to move `!event.targetIsFriendly` above this if-statement and do an early `return`; grouping unrelated if-statements makes it harder to see.
2. Is this supposed to not be filtered byPlayer (`on_byPlayer_damage`)?

Thieseract: Changed the targetIsFriendly check to a separate statement and changed to on_byPlayer_damage

PR #301, opened 2017-09-12
[Balance Druid] Unempowered LS module

Iskalla:
Don't think i should show a Statistic Box about this, but if necessary let me know.


rp4rk:
Could you fix the conflicts on this pretty please

PR #305, opened 2017-09-12
[Havoc] Update Cast Efficiency

ronaldpereira:
Vengeful Retreat does not show 'can be improved' anymore, as it is a really unused movement spell.
Eye Beam does not show 'can be improved' anymore, as it is a very situational spell.


anom0ly:
Thanks.

PR #307, opened 2017-09-13
[Vengeance + Havoc] Code refactoring and updated Fel Rush and Blade Dance cast recommendation

ronaldpereira:
Blade Dance recommended cast efficiency was a bit higher than the normal.
Also, I've made some code refactoring too: moved the SpiritBomb.js module to a specific folder, removed the if(!this.owner.error) from the on_initialized functions and refactored the immolation aura module (I forgot to do in the previous PRs).


rp4rk:
Cheers 👍

PR #308, opened 2017-09-13
Reworked Runic Power Tracking

yajinni:
Reworked Runic Power Tracking. Created a separate module to track generation and wasting. Deleted my old simple stat box module since the new tracker has it. Created a tab that shows a chart for this.


rp4rk:
Can we get a screenshot of the interface elements you've added?

yajinni:
Here is a screenshot: https://gyazo.com/d90067ca1da379c079042d9a6fb79627

PR #311, opened 2017-09-14
Fixed a bug where event.classResources wasn't checked for undefined

kfinch:
Observed issue in RestoDruid Innervate module, added to the other places just in case.


anom0ly:
Thanks.

PR #313, opened 2017-09-14
Disc Priest: Fix issue where friendly fire damage was considered a valid Atonement damage source

MartijnHols:
Depends on #312


MartijnHols:
Should add pet damage source tracking in another PR.

PR #317, opened 2017-09-16
Shadowpriest fixes

hasseboulen:
Fixed dead GCD time and 'this.owner.modules' deprectated warnings.
Removed the local Shadowpriest Haste module as this did the exact same thing as the Core one.


MartijnHols:
I don't think it *really* works the same, as the applybuff for lingering insanity has stacks with the old implemenation but not in new.

src/Parser/ShadowPriest/Modules/Core/Pet.js
```diff
@@ -1,6 +1,11 @@
 import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
 
 class Pet extends Module {
+  static dependencies = {
```
MartijnHols: This doesn't seem to do anything, if it does then please leave a comment why it's necessary, if not then it can be omitted 😅

src/Parser/ShadowPriest/Modules/Spells/Mindbender.js
```diff
@@ -7,11 +7,12 @@ const MINDBENDER_UPTIME_MS = 15000;
 const MINDBENDER_ADDED_UPTIME_MS_PER_TRAIT = 1500;
 
 class Mindbender extends Pet {
```
MartijnHols: This is missing the combatants dependency.

src/Parser/ShadowPriest/Modules/Spells/Shadowfiend.js
```diff
@@ -7,7 +7,7 @@ class Shadowfiend extends Pet {
   _pet = PETS.SHADOWFIEND;
```
MartijnHols: This is missing the combatants dependency.

MartijnHols:
Thanks!

PR #318, opened 2017-09-16
[Affli + Destro] Cooldown tracker rework

sMteX:
I added a support for cooldowns that aren't player buffs (we summon temporary pets instead), by tracking casts and adding a fixed duration to them (I checked, no legendary or set bonuses change the duration). So far only locally for Warlock specs, if it would be needed in other specs, it's of course possible to migrate that to Core Cooldown Tracker.


src/Parser/AfflictionWarlock/Modules/Features/CooldownTracker.js
```diff
@@ -11,44 +13,91 @@ class CooldownTracker extends CoreCooldownTracker {
...
+    const spellId = event.ability.guid;
+    const cooldownSpell = this.constructor.castCooldowns.find(cooldownSpell => cooldownSpell.spell.id === spellId);
+    if (cooldownSpell) {
+      // adding the fixed cooldown, now we need to remove it from activeCooldowns too
+      const cooldown = this.addFixedCooldown(cooldownSpell, event.timestamp);
+      this.activeCooldowns.push(cooldown);
+      debug && console.log(`%cCooldown started: ${cooldownSpell.spell.name}`, 'color: green', cooldown);
+    }
+    // super.on_byPlayer_cast(event) would call trackEvent anyway
+    this.trackEvent(event);
```
MartijnHols: Please do `super.on_byPlayer_cast && super.on_byPlayer_cast(event);` instead, so that you're calling it even when it changes.

src/Parser/AfflictionWarlock/Modules/Features/CooldownTracker.js
```diff
@@ -11,44 +13,91 @@ class CooldownTracker extends CoreCooldownTracker {
...
+  }
+
+  addFixedCooldown(cooldownSpell, timestamp) {
```
MartijnHols: This should be private; prefixed with an underscore (addCooldown too, but that's old code).

src/Parser/AfflictionWarlock/Modules/Features/CooldownTracker.js
```diff
@@ -11,44 +13,91 @@ class CooldownTracker extends CoreCooldownTracker {
...
+      end: timestamp + cooldownSpell.duration * 1000,
+      events: [],
+    };
+    this.pastCooldowns.push(cooldown);
+    return cooldown;
+  }
+
+  // on_event() might be more accurate but it would be most likely called much more
+  trackEvent(event) {
+    const finishedCooldowns = this.activeCooldowns.filter(cooldown => cooldown.end && cooldown.end < event.timestamp).map((_, index) => index);
```
MartijnHols: This would be cleaner:

```jsx
this.activeCooldowns
  .filter(cooldown => cooldown.end && cooldown.end < event.timestamp)
  .forEach((cooldown, index) => {
    debug && console.log(`%cCooldown ended: ${cooldown.spell.name}`, 'color: red', cooldown);
    this.activeCooldowns.splice(index, 1);
  });
```

Also might want to do `event.timestamp > cooldown.end` instead; it might be a little bit clearer that you're looking for cooldowns that have ended.

sMteX: Initially I thought about this too, I was worried it wouldn't work correctly (I remember that in C# if you iterate through a list with foreach and remove an element you'd get an exception that the collection changed while a iterating and I thought this would be the same case)

MartijnHols: Even better:

`this.activeCooldowns = this.activeCooldowns.filter(cooldown => cooldown.end && cooldown.end < event.timestamp);`

rp4rk:
This branch has conflicts, could they be resolved?

MartijnHols:
Thanks

PR #319, opened 2017-09-16
[core] Added a 'byAbility' util to the HealingDone module

kfinch:


kfinch:
Wait a sec, this might not be right. I think it should be 'event.ability.guid', not 'event.guid'

kfinch:
Which means the utility in DamageTaken is wrong too

MartijnHols:
Thanks!

PR #321, opened 2017-09-16
[Core] Added Archimonde's Hatred Reborn DPS meters on tooltip

ronaldpereira:
I inserted the DPS contribution to the Archimonde's Hatred Reborn legendary. It was a suggestion from a theorycrafter player in order to keep track of it maximum damage possible and DPS meters for it.

image


MartijnHols:
Thanks!

PR #324, opened 2017-09-16
[core] byAbility and byMagicSchool no longer return undefined when no entry yet

kfinch:
I tried to use byAbility in HealingDone and was immediately hit by the inconvenience of constant undefined checks. It seems to me that if I call byAbility on something that hasn't healed yet, I should get zeroes, not undefined.

A search of the code revealed that no one has used these util methods in DamageTaken either, so I gave them the same treatment.

By the way, how do you feel about using "if(!object)" to check for undefined? Searching the web for this question, I found a lot of answers for how you *could* check for undefined, but very few about how you *should* check for undefined.

### DIFF
--- a/src/Parser/Core/Modules/DamageTaken.js
+++ b/src/Parser/Core/Modules/DamageTaken.js
@@ -24,17 +24,19 @@ class DamageTaken extends Module {
   _byAbility = {};
   byAbility(spellId) {
     if(!this._byAbility[spellId]) {
-      this._byAbility[spellId] = new DamageValue(0, 0, 0);
+      return new DamageValue(0, 0, 0);
+    } else {
+      return this._byAbility[spellId];
     }
-    return this._byAbility[spellId];
   }
-  
+
   _byMagicSchool = {};
   byMagicSchool(magicSchool) {
     if(!this._byMagicSchool[magicSchool]) {
-      this._byMagicSchool[magicSchool] = new DamageValue(0, 0, 0);
+      return new DamageValue(0, 0, 0);
+    } else {
+      return this._byMagicSchool[magicSchool];
     }
-    return this._byMagicSchool[magicSchool];
   }
 
   on_toPlayer_damage(event) {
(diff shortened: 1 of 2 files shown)


MartijnHols:
Check for undefined with !object is fine if you considered that all the other falsy values aren't going to cause problems.

src/Parser/Core/Modules/DamageTaken.js
```diff
@@ -20,12 +20,20 @@ class DamageTaken extends Module {
   get total() {
     return this._total;
   }
+
   _byAbility = {};
   byAbility(spellId) {
+    if(!this._byAbility[spellId]) {
+      this._byAbility[spellId] = new DamageValue(0, 0, 0);
```
MartijnHols: I would prefer returning `new DamageValue(0, 0, 0);` making this simpler, storing it permanently in the array seems to have no benefits.

kfinch: I went back and forth between that way and the way you see it implemented. I ended up going with this way because I rationalized this would be better in the case of multiple gets for the same spell that did 0 healing. In practice the benefit of this is probably indistinguishable from zero. 

I'll change it to your way, you're right it's simpler.

kfinch:
Done

src/Parser/Core/Modules/DamageTaken.js
```diff
@@ -20,13 +20,23 @@ class DamageTaken extends Module {
   get total() {
     return this._total;
   }
+
   _byAbility = {};
   byAbility(spellId) {
-    return this._byAbility[spellId];
+    if(!this._byAbility[spellId]) {
+      return new DamageValue(0, 0, 0);
+    } else {
```
MartijnHols: This else can be omitted since the first if-statement returns and functionally it works like an exception to the normal flow so for max readability I feel it's a little bit better. 😅

kfinch: I know the else isn't needed, but stylistically I prefer it. When a function is strictly "one thing or the other thing" I like the 'else' because it makes it obvious at a glance that both lines can't happen. Just a preference thing I think.

PR #326, opened 2017-09-17
Enhance flametongue refresh pr

Nighteyez07:
Add Pandemic refresh check for Enhancement shaman


src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,58 @@
...
+import Module from 'Parser/Core/Module';
+
+const SPELLID = SPELLS.FLAMETONGUE_BUFF.id;
+const PANDEMIC_THRESHOLD = 11500;
+const debug = false;
+
+class Pandemic extends Module {
+  ft_timestamp = [];
+  early_refresh = [];
+  refresh_avearge;
```
MartijnHols: Please set a default value for properties.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,58 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  on_byPlayer_applybuff(event) {
+    if(event.ability.guid === SPELLID) {
```
MartijnHols: Use the spell reference `SPELLS.FLAMETONGUE_BUFF.id` straight up here is preferred to make it easier to see what spell is being checked.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  on_finished() {
+    let diff;
+
+    for (let i = 1; i < this.ft_timestamp.length; i++) {
+      diff = this.ft_timestamp[i] - this.ft_timestamp[i - 1];
+      if(diff < PANDEMIC_THRESHOLD) {
+        this.early_refresh.push(diff);
```
MartijnHols: It's generally preferred to do the actual data fetching in the related events, in this case `on_byPlayer_refreshbuff` instead of storing a buffer and processing it in on_finished. Instead of tracking the array `ft_timestamp` you would instead track something like the `lastFlameTongueAppliedTimestamp` which would be a simpler value.

Also this will probably break if FT is cast pre-combat.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+    debug && console.log(this.early_refresh);
+    if(this.early_refresh.length > 0) {
+      this.refresh_avearge = this.early_refresh.reduce((a, b) => a + b) / this.early_refresh.length;
```
MartijnHols: I'd prefer not doing complete calculations in the on_finished method. Since refresh_average is based purely on class properties you could also make refresh_average a getter that calculates it when called.

Nighteyez07:
Requested changes completed

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,54 @@
+import SPELLS from 'common/SPELLS';
+
+import Module from 'Parser/Core/Module';
+
+const PANDEMIC_THRESHOLD = 11500;
```
enragednuke: If I recall correctly, the duration of Flametongue is 16s, so the pandemic would be after 11.2s (11200) right?

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,54 @@
...
+
+class Pandemic extends Module {
+  flametongueTimestamp = 0;
+  earlyRefresh = 0;
+  refreshTimestamps = [];
+
+  on_byPlayer_applybuff(event) {
+    if(event.ability.guid === SPELLS.FLAMETONGUE_BUFF.id) {
+      if(this.flametongueTimestamp !== 0) {
+        if(event.timestamp - this.flametongueTimestamp < PANDEMIC_THRESHOLD) {
```
enragednuke: An applybuff event only fires when the buff is applied fresh. So you don't need to worry about pandemics in an applybuff event.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,54 @@
+import SPELLS from 'common/SPELLS';
```
enragednuke: I would also name this file something else since this module seems to serve solely as a flametongue early refresh checker. Perhaps just name it 'FlametongueRefresh' or something of the sort. It's not considered good practice to have multiple suggestions per module anyway, so you wouldn't want to add Frostbrand here to make it a "Pandemic" module.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
```diff
@@ -0,0 +1,54 @@
...
+
+import Module from 'Parser/Core/Module';
+
+const PANDEMIC_THRESHOLD = 11500;
+const debug = false;
+
+class Pandemic extends Module {
+  flametongueTimestamp = 0;
+  earlyRefresh = 0;
+  refreshTimestamps = [];
```
enragednuke: If you're not planning on making something with this var, maybe put debug conditionals with the 2 `push` calls below to preserve memory (or remove it altogether if you're confident in this module being finished and correct).

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
Nighteyez07: The rule is don't refresh with more than 4.5 seconds left on the debuff. 
Timestamp is record in ms, so entire buff is 16,000 ms, subtract 4.5 s (4500ms), and you get 11,500ms.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
Nighteyez07: Debating whether to add the times in the log where it was refreshed early. I'll wrap debug in it for now until I decide for sure which way to go.

src/Parser/EnhancementShaman/Modules/ShamanCore/Pandemic.js
Nighteyez07: True, if the buff fell off completely no point checking if refreshed early.

MartijnHols:
Thanks!

PR #330, opened 2017-09-18
Resto druid relic traits

buimichael:
image


buimichael:
@kfinch Feel free to comment if you find anything interesting.

kfinch:
Persistence at 2.8% can't possibly be right, I'm gonna take a closer look at that.

src/Parser/RestoDruid/Modules/Traits/EssenceOfNordrassil.js
```diff
@@ -0,0 +1,54 @@
...
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import calculateEffectiveHealing from 'Parser/Core/calculateEffectiveHealing';
+import SPELLS from 'common/SPELLS';
+
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+
+const ESSENCE_OF_NORDRASSIL_HEALING_INCREASE = 0.1;
```
kfinch: Shouldn't this be 0.05?

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
```diff
@@ -0,0 +1,55 @@
...
+    this.active = this.rank > 0;
+  }
+
+  on_byPlayer_heal(event) {
+    if(event.ability.guid !== SPELLS.HEALING_TOUCH.id || (event.ability.guid !== SPELLS.REGROWTH.id && event.tick) ) {
+      return;
+    }
+
+    this.healing += calculateEffectiveHealing(event, BLESSING_OF_THE_WORLD_TREE_HEALING_INCREASE * this.rank)/ this.rank;
+  }
```
kfinch: Stylistically I think it's better to use an affirmative 'if' instead of using an early return, like:

```
if(event.ability.guid === SPELLS.HEALING_TOUCH.id || (event.ability.guid === SPELLS.REGROWTH.id && !event.tick) ) {
  +    this.healing += calculateEffectiveHealing(event, BLESSING_OF_THE_WORLD_TREE_HEALING_INCREASE * this.rank)/ this.rank;
}
```

I think this mostly doesn't matter, just thought I'd throw it out there.

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+
+    // We are interested in the healing done on the last second of EoG
+    if(this.lastEoGApplied != null
+      && (this.lastEoGApplied + this.eogDuration - 1000) <= event.timestamp
+      && this.combatants.selected.hasBuff(SPELLS.ESSENCE_OF_GHANIR.id)
+      && HOTS_AFFECTED_BY_ESSENCE_OF_GHANIR.indexOf(spellId) !== -1) {
```
kfinch: Better to use includes() here, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
```diff
@@ -0,0 +1,72 @@
...
+      if(SPELLS.REGROWTH.id === spellId && event.tick !== true) {
```
kfinch: Technically you don't need to check for Regrowth here, could just check that it's not a tick. Also, `event.tick !== true` should be `!event.tick`.

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
```diff
@@ -0,0 +1,72 @@
...
+        return;
+      }
+        this.healing += event.amount/2;
```
kfinch: `event.amount` should be `(event.amount + (event.absorbed || 0))`

src/Parser/RestoDruid/Modules/Traits/Grovewalker.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    if(!HEALS_MASTERY_STACK.indexOf(spellId)) {
+      return;
+    }
+    if(SPELLS.REGROWTH.id === spellId && !event.tick) {
+      return;
+    }
```
kfinch: We're doing a check that boils down to "is this a resto HoT tick?" a lot. Maybe should be a util function somewhere.

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
```diff
@@ -0,0 +1,67 @@
...
+    const spellId = event.ability.guid;
+    if(SPELLS.WILD_GROWTH.id !== spellId) {
+      return;
+    }
+    this.manaGained += this.wgManaCost - (this.wgManaCost * (1 - INFUSION_OF_NATURE_REDUCTION));
+    this.freeRejuvs = this.manaGained / this.rejuvManaCost;
+  }
+
+  subStatistic() {
+    const oneRejuvenationThroughput = this.owner.getPercentageOfTotalHealingDone(this.owner.modules.treeOfLife.totalHealingFromRejuvenationEncounter) / this.owner.modules.treeOfLife.totalRejuvenationsEncounter;
```
kfinch: When my refactor is merged, this should be changed to use the one rejuv throughput helper in the new Rejuvenation module. In fact, the ToL utility doesn't exist after the refactor, so will have to be changed during the merge.

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
```diff
@@ -0,0 +1,67 @@
...
+ * Reduces the mana cost of Wild Growth by 2%
+ */
+class InfusionOfNature extends Module {
+  static dependencies = {
+    combatants: Combatants,
+    healingDone: HealingDone,
+  };
+
+  rank = 0;
+  healing = 0;
```
kfinch: This var not used?

src/Parser/RestoDruid/Modules/Traits/KnowledgeOfTheAncients.js
```diff
@@ -0,0 +1,60 @@
...
+
+  on_finished() {
+    // MP5
+    this.manaGained = ((this.owner.fightDuration/1000)/5)*KNOWLEDGE_OF_THE_ANCIENT_MANA_GAIN;
+  }
+
+  subStatistic() {
+    const rejuvManaCost = BASE_MANA * REJUVENATION_BASE_MANA;
+    const freeRejuvs = this.manaGained / rejuvManaCost;
+    const oneRejuvenationThroughput = this.owner.getPercentageOfTotalHealingDone(this.owner.modules.treeOfLife.totalHealingFromRejuvenationEncounter) / this.owner.modules.treeOfLife.totalRejuvenationsEncounter;
```
kfinch: Also needs to be switched during merge with my refactor

src/Parser/RestoDruid/Modules/Traits/SeedsOfTheWorldTree.js
```diff
@@ -0,0 +1,54 @@
...
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import calculateEffectiveHealing from 'Parser/Core/calculateEffectiveHealing';
+import SPELLS from 'common/SPELLS';
+
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+
+const SEEDS_OF_THE_WORLD_TREE_HEALING_INCREASE = 0.08;
```
kfinch: Seeds of the World Tree doesn't increase Living Seed by 8%, it increases _the portion of the initial heal that goes into the Living Seed by 8%_. Living Seed baseline is 25%, this constant should actually be (0.08 / 0.25). Inspection of Living Seed healing amounts confirms this.

src/Parser/RestoDruid/Modules/Traits/Grovewalker.js
```diff
@@ -0,0 +1,60 @@
...
+  healing = 0;
+
+  on_initialized() {
+    this.rank = this.combatants.selected.traitsBySpellId[SPELLS.GROVEWALKER.id];
+    this.active = this.rank > 0;
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    if(!HEALS_MASTERY_STACK.indexOf(spellId)) {
```
kfinch: This isn't the correct check, `HEALS_MASTERY_STACK.indexOf(spellId)` evaluates false for the item in index 0. Use HEALS_MASTERY_STACK.includes(spellId) instead.

src/Parser/RestoDruid/Modules/Traits/EssenceOfNordrassil.js
buimichael: Right, I even had it written down on the to-fix list but forgot about it. T_T

kfinch:
Ok I found the overcount in Persistence: oneRejuvenationThroughput is calculated using rejuv casts, but the amount of free rejuvs gained is from rejuv applications.

Also, we're assuming you never clip a rejuv (which is a bad assumption), but we already know about that shortcoming. Unfortunately, these are both overcounts.

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
buimichael: I believe @MartijnHols mentioned somewhere (?) that he enforces early returns.

src/Parser/RestoDruid/Modules/Traits/EssenceOfNordrassil.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/EternalRestoration.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/Grovewalker.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/Grovewalker.js
buimichael: This check is only relevant for regrowth as they use the same Id for the HoT part and the heal part.

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
buimichael: Removed.

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
buimichael: Added TODO

src/Parser/RestoDruid/Modules/Traits/KnowledgeOfTheAncients.js
buimichael: Added TODO

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
kfinch: I'll ask him what the reason is for that. I always preferred early returning only for exceptional behavior. When implementing pseudocode of "if X, then do A", I prefer:
```
if(X) {
   do A;
}
```
rather than:
```
if(!X) {
   return;
}
do A;
```

src/Parser/RestoDruid/Modules/Traits/Grovewalker.js
kfinch: I know why it's needed, I'm just saying we're doing this special case a lot, and it's all in service of detecting if a heal event was a resto HoT tick.

src/Parser/RestoDruid/Modules/Traits/SeedsOfTheWorldTree.js
buimichael: Done

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
buimichael: I believe it's to avoid potential nested if-cases. This is probably only applicable once you start nesting over a depth of 3.
https://3layer.com.br/sonar/rules/show/javascript:NestedIfDepth?layout=false

```
if(X) {
   some code;
   if(Y) {
      some code;
      if(Z) {
           do A;
      }
   }
}
```

```
if(!X) {
   return;
}

some code;
if(!Y) {
   return;
}
some code;
if(!Z) {
   return;
}

do A
```

buimichael:
@kfinch 
I'll keep the persistence module as if for now, and leave the TODO tag. We need to add the increased bonuses from dreamwalker, inc avg mastery stack and cultivation as they are a side effect of rejuvenation. (This holds true for all modules making use of the oneRejuvenationThroughput variable)

Also we need to adjust/subtract throughput if a a rejuv is applied outside of pandemic time frame

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
MartijnHols: Early returns are preferred Early returns are preferred when they improve readability and maintainability. If they don't then don't use them. It's more a guideline as it's near impossible to make rules that apply just as well everywhere.

I find nested if-statements harder to follow, change, add onto and remove as they require you to understand every single layer before you can modify the layers. On the other hand early returning keeps your method 1 layer deep and acts more like exceptions to the regular flow which is a lot easier to track, understand and modify. I also reckon often when you need multiple nested if-statements in one function you might need to start refactoring.

In this instance either could work, because of the specific Regrowth tick filtering I reckon it's more readable to nest the expression.

kfinch:
Ok I actually don't feel like making a new issue / thread, so I'll put more of my thoughts on Persistence calculation here,

We're leveraging the average healing per rejuv cast calculation when that's not really what we want. Consider the following instead: What we really want is to get the average amount of healing one second of rejuvenation does, and then multiply that by the number of extended seconds.

We can actually get healing per second from rejuv pretty easily: we just sum the buff uptime on all targets (for rejuv and germ) to get total seconds up, and then we divide by total healing done by those spells (and associated things if we want to add in cultivation / dreamwalker / mastery etc...). This gives us healing from one second of rejuv.

For total seconds of extended we don't have to change much, currently we're approximating by just counting the number of rejuv applications / refreshes. We can later extend by subtracting times it was refreshed early.

From this we have a much more reasonable calculation.

One other issue I raised was that the _last_ second of any rejuv's healing seems much more likely to be overheal than the _average_ second of a rejuv's healing. This is a more awkward issue to correct, but I think we could get pretty close by doing the following: We keep track of the average overhealing of the _last_ tick of every rejuv, and we correct our results by using the average overheal of the last tick rather than the average total overheal.

buimichael:
@kfinch I like the idea of that!

src/Parser/RestoDruid/Modules/Traits/ArmorOfTheAncients.js
```diff
@@ -0,0 +1,12 @@
+
+import Module from 'Parser/Core/Module';
+
+/**
+ * Armor of the Ancients
+ * Ironbark reduces all damage taken by the target by an additional 2%.
+ */
+class ArmorOfTheAncients extends Module {
+//TODO
```
kfinch: We can probably just do the DRPS calculation here. For the particular case of Ironbark, DRPS is going to be pretty comparable to HPS. Maybe include in tooltip the explanation that it's DRPS and also says average extra damage reduced per Ironbark cast.

buimichael:
@kfinch 
Please review the changes to persistence

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
```diff
@@ -0,0 +1,55 @@
...
+  rank = 0;
+  healing = 0;
+
+  on_initialized() {
+    this.rank = this.combatants.selected.traitsBySpellId[SPELLS.BLESSING_OF_THE_WORLD_TREE.id];
+    this.active = this.rank > 0;
+  }
+
+  on_byPlayer_heal(event) {
+    if(event.ability.guid !== SPELLS.HEALING_TOUCH.id || (event.ability.guid !== SPELLS.REGROWTH.id && event.tick) ) {
```
kfinch: Should be &&, not ||. As currently written this can never evaluate true, which is why in practice it's always showing value of 0.

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
```diff
@@ -0,0 +1,68 @@
...
+    const infusionOfNatureThroughput = oneRejuvenationThroughput * this.freeRejuvs;
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.INFUSION_OF_NATURE.id}>
+            <SpellIcon id={SPELLS.INFUSION_OF_NATURE.id} noLink /> Infusion of Nature
+          </SpellLink>
+        </div>
+        <div className="flex-sub text-right">
+          {formatPercentage(infusionOfNatureThroughput)} %
```
kfinch: Should add a wavy equals sign (like for Second Sunrise in the HolyPaladin traits output) for any value that we're fudging. Also a tooltip for how this number was derived.

src/Parser/RestoDruid/Modules/Traits/KnowledgeOfTheAncients.js
```diff
@@ -0,0 +1,63 @@
...
+    const knowledgeOfTheAncientsThroughput = oneRejuvenationThroughput * freeRejuvs;
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.KNOWLEDGE_OF_THE_ANCIENTS.id}>
+            <SpellIcon id={SPELLS.KNOWLEDGE_OF_THE_ANCIENTS.id} noLink /> Knowledge of the Ancients
+          </SpellLink>
+        </div>
+        <div className="flex-sub text-right">
+          {formatPercentage(knowledgeOfTheAncientsThroughput)} %
```
kfinch: Should add a wavy equals sign (like for Second Sunrise in the HolyPaladin traits output) for any value that we're fudging. Also a tooltip for how this number was derived.

src/Parser/RestoDruid/Modules/Traits/Persistence.js
```diff
@@ -0,0 +1,85 @@
...
+      .reduce((uptime, player) =>
+        uptime + player.getBuffUptime(spellId), 0);
+  }
+
+  subStatistic() {
+    const uptimeRejuvenation = Math.floor((this.getSpellUptime(SPELLS.REJUVENATION.id) + this.getSpellUptime(SPELLS.REJUVENATION_GERMINATION.id)) / 1000);
+    const totalRejuvenationHealing = this.rejuvenation.totalRejuvHealing;
+    const oneSecondRejuvenationHealing = totalRejuvenationHealing/uptimeRejuvenation;
+
+    const persistenceThroughput = this.owner.getPercentageOfTotalHealingDone(this.totalPersistanceValue * oneSecondRejuvenationHealing);
```
kfinch: Looks good. I'm working on enhancements to the Rejuvenation module to implement some of the optional things, but you should probably merge first.

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
kfinch: Oh, and the && should be an ||. Should be:
```
if(event.ability.guid !== SPELLS.HEALING_TOUCH.id && (event.ability.guid !== SPELLS.REGROWTH.id || event.tick) ) {
```

src/Parser/RestoDruid/Modules/Traits/InfusionOfNature.js
kfinch: I'm envisioning something like "Your last point in this trait saved you X mana. This healing value is derived from the assumption that you used that mana to cast more Rejuvenations."

buimichael: Great idea, done. Added a TODO to reduce the effect of this trait if the player didn't make use of the mana gained.

src/Parser/RestoDruid/Modules/Traits/BlessingOfTheWorldTree.js
buimichael: Ooops, done.

src/Parser/RestoDruid/Modules/Traits/KnowledgeOfTheAncients.js
buimichael: Great idea, done. Added a TODO to reduce the effect of this trait if the player didn't make use of the mana gained.

PR #332, opened 2017-09-19
[Demonology] Soul Shard Tracker fix

sMteX:
Didn't need to change anything apart from input/output spells.


MartijnHols:
Thanks!

PR #335, opened 2017-09-20
Add a `completeness` property to spec config that indicates how complete a spec's implementation is

MartijnHols:
I set all specs to *needs more work* by default. We will do a review of a spec analysis once this is increased to see if an increase is correct. The goal of this is to make it more obvious to users how complete a spec implementation is to make it clear that some specs still need more work and aren't reflective of the usefulness of the tool.

This indicator might become more obvious in the future.

I'd like to hear what you guys think about this before merging it into master.


MartijnHols:
Screenshots:
image
image

MartijnHols:
I'm not familiar enough with all specs to make a judgment right from the start, so I only changed the completeness of the specs I'm familiar enough with. Other specs will have to request a change via a PR or talk about it in #contributors.

PR #346, opened 2017-09-21
Still a lot of work left to do

CollCrom:
Need to add more legendaries and Judgment at least. Cast efficiency needs an overhaul to get around negative percent downtime and casting more spells than should be possible due to crusade. Could add something to make sure you are using trinkets at the right times as well


anom0ly:
Sounds good. Thanks!

PR #347, opened 2017-09-21
Chain of Thrayn

CollCrom:
Added a chain of Thrayn Module

### DIFF
--- a/src/Parser/RetributionPaladin/Modules/Items/ChainOfThrayn.js
+++ b/src/Parser/RetributionPaladin/Modules/Items/ChainOfThrayn.js
@@ -21,8 +21,9 @@ class ChainOfThrayn extends Module {
 	}
 
 	on_byPlayer_damage(event) {
-		if(this.combatants.selected.hasBuff(SPELLS.CRUSADE_TALENT.id) || this.combatants.selected.hasBuff(SPELLS.CRUSADE_TALENT.id)){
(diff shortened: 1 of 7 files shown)


src/Parser/RetributionPaladin/Modules/Items/ChainOfThrayn.js
```diff
@@ -0,0 +1,42 @@
...
+
+	damageDone = 0;
+
+	on_initialized() {
+		this.active = this.combatants.selected.hasWaist(ITEMS.CHAIN_OF_THRAYN.id);
+	}
+
+	on_byPlayer_damage(event) {
+		if(this.combatants.selected.hasBuff(SPELLS.CRUSADE_TALENT.id) || this.combatants.selected.hasBuff(SPELLS.CRUSADE_TALENT.id)){
+			this.damageDone += ((event.amount || 0) + (event.absorbed || 0)) * CHAIN_OF_THRAYN_INCREASE / (1 + CHAIN_OF_THRAYN_INCREASE);
```
anom0ly: You are checking for the `CRUSADE_TALENT.id` twice here, did you mean to select a different spell?

Also, isn't the 10% bonus from Thrayn already in the damage event? If so the formula would be `bonusDamage = raw - (raw / (1 + Increase))` I believe.

anom0ly:
Just double check the logic on the damage increase contribution. Other than that looks good.

CollCrom:
Yeah oops meant to do Avenging Wrath as well. I changed the calculation as well.

anom0ly:
Looks good. Thanks.

PR #349, opened 2017-09-21
Minor math bug fix

CollCrom:
Changed the bonus damage calculations. also made getting bonus damage is own method to get rid of some copy pasting


anom0ly:
Thanks.

anom0ly:
Thanks.

PR #351, opened 2017-09-21
[Marksmanship] Added checks to see if the hunter used pets at all, if not do not sho…

Pewtro:
…w pet contribution dmg, as it is pointless. Renamed 'wolves' to 'pets' to be more precise.


MartijnHols:
Thanks for the PR! As `dmgTooltip` as a property is only used in the `statistic()` method, could you please make it a local variable of this method instead? This would remove the need for an additional method (which should be camcelCase btw, just event handlers start with `on_` to indicate they're special), property, and listening to on_finished leading to shorter and easier to understand code.

MartijnHols:
Thanks!

PR #358, opened 2017-09-22
Updated descriptions and tracking.

yajinni:
Changed some efficiency thresh holds and wordings.
Big change is i split vampiric blood, icebound fortitude, and blood mirror into thier own modules.
Reason for this is to make custom suggestions for them.
The reason being is they are mostly defensive CDs. So you dont use them on cool down, and should be using them smartly. Sometimes you dont even use them on trivial content.
But i still wanted them to show on the minor side of things to highlight them better to a player then simply something int he efficiency cast list.  This are things you can improve and do better with if you find yourself either struggling or wanting to get better. So i wanted them to pop out more for people to see.


src/Parser/BloodDeathKnight/Modules/Core/BloodMirror.js
```diff
@@ -0,0 +1,29 @@
...
+import SpellLink from 'common/SpellLink';
+
+const BLOOD_MIRROR = 120;
+
+class BloodMirror extends Module {
+  static dependencies = {
+    abilityTracker: ABILITYTRACKER,
+  };
+
+    suggestions(when) {
```
MartijnHols: I think this could also be solved by the CastEfficiency module, why is this separate?

src/Parser/BloodDeathKnight/Modules/Core/IceboundFortitude.js
```diff
@@ -0,0 +1,29 @@
...
+import SpellLink from 'common/SpellLink';
+
+const ICEBOUND_FORTITUDE = 180;
+
+class IceboundFortitude extends Module {
+  static dependencies = {
+    abilityTracker: ABILITYTRACKER,
+  };
+
+    suggestions(when) {
```
MartijnHols: I think this could also be solved by the CastEfficiency module, why is this separate?

src/Parser/BloodDeathKnight/Modules/Core/VampiricBlood.js
```diff
@@ -0,0 +1,29 @@
...
+import SpellLink from 'common/SpellLink';
+
+const VAMPIRIC_BLOOD_COOLDOWN = 60;
+
+class VampiricBlood extends Module {
+  static dependencies = {
+    abilityTracker: ABILITYTRACKER,
+  };
+
+    suggestions(when) {
```
MartijnHols: I think this could also be solved by the CastEfficiency module, why is this separate?

src/Parser/BloodDeathKnight/Modules/Features/CastEfficiency.js
```diff
@@ -56,26 +55,19 @@ class CastEfficiency extends CoreCastEfficiency {
...
       category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
       getCooldown: haste => 30,
-      recommendedCastEfficiency: 0.85,
+      recommendedCastEfficiency: 0.90,
       isActive: combatant => combatant.hasTalent(SPELLS.BLOODDRINKER_TALENT.id),
       extraSuggestion: <span>Mostly used as a dps CD. Should be almost casted on CD. Good to use when your running to the boss or cant melee them.</span>,
     },
 
-
-    {
```
MartijnHols: Why was this removed? Using active racials like Arcane Torrent generate resources that can be used to improve play, using it on cooldown should be trivial.

src/Parser/BloodDeathKnight/Modules/Items/T20_2pc.js
```diff
@@ -11,22 +11,37 @@ import Combatants from 'Parser/Core/Modules/Combatants';
 class T20_2pc extends Module {
   static dependencies = {
     combatants: Combatants,
-  }
+  };
 
+  uptime=0;
```
MartijnHols: This could be a getter;

```jsx
get uptime() {
  return this.combatants.getBuffUptime(SPELLS.GRAVEWARDEN.id) / this.owner.fightDuration;
}
```

src/Parser/BloodDeathKnight/Modules/Items/T20_2pc.js
```diff
@@ -11,22 +11,37 @@ import Combatants from 'Parser/Core/Modules/Combatants';
...
 
   on_initialized() {
     this.active = this.combatants.selected.hasBuff(SPELLS.BLOOD_DEATH_KNIGHT_T20_2SET_BONUS_BUFF.id);
   }
 
+  on_finished() {
+    this.uptime = this.owner.modules.combatants.getBuffUptime(SPELLS.GRAVEWARDEN.id) / (this.owner.fightDuration);
```
MartijnHols: Only updating the value in `on_finished` is undesirable as it won't show progress while loading.

src/Parser/BloodDeathKnight/Modules/Items/T20_4pc.js
```diff
@@ -25,17 +26,22 @@ class T20_4pc extends Module {
     const spellId = event.ability.guid;
     if (spellId === SPELLS.DEATH_STRIKE.id && !this.combatants.selected.hasBuff(SPELLS.GRAVEWARDEN.id, event.timestamp)) {
       this.deathStrinkeBuffless += 1;
+    }else{
+      if (spellId === SPELLS.DEATH_STRIKE.id)
```
MartijnHols: Please use braces.

src/Parser/BloodDeathKnight/Modules/Core/BloodMirror.js
yajinni: Sorry for my bad explanation is the PR lol
Basically this is my goal and reason for separating it. Defensive CDs like this should not be used on CD and shouldnt be encouraged to used on CD like offensive abilities. So i wanted to simply show number of casts without showing the efficiency percent . To do that i had to split it off to make a custom suggestion that shows this. I wanted this to still be on the suggestion page under minor to bring a lil more attention to it instead of it being lost in the list of spells casted. So when people look they know to keep in mind they may need to use them more if needed. I plan on implementing them in the Cooldown tab also to show damage reduced or healing done because of them at a later time.

MartijnHols:
I sent a response to your DM, but to make it more obvious this PR is awaiting your action, here it is again;

```
Hey, in the future please use the PR to discuss changes, it keeps everything centralized.
I'm not sure about pulling spells out of the Cast Efficiency tab. You can already disable the can be improved indicator and suggestion, and I don't think showing the actual cast efficiency is detrimental and suggesting players to cast things on cooldown always. While defensive CDs should be used when they are needed, using them as often as possible does reduce damage taken and smooth it out, so casting is 4 out of 5 times is almost always going to be better than 3 ouf of 5 times. For example when a spell has a 1:30 CD and the first usage is at 2:00, using it before 0:30 at the latest would be optimal usage. Showing just the cast efficiency bar does not suggest casting it on cooldown, but does indicate the amount of missed casts relative to the max possible casts.
```

src/Parser/BloodDeathKnight/Modules/Core/BloodMirror.js
yajinni: Hello Maybe this pic will show what i want better:

image

MartijnHols:
I think the problem you're trying to solve would be better solved by using the `importance` property to set these spells to a static suggestion level of *minor* (as you do in your new modules), and setting the `recommendedCastEfficiency` to a lower value like 60%. It will still recommend a cast efficiency above 60%, but I think this is for the best as there really is no reason to not cast your spell ever.

I personally believe even making the importance for these suggestions static would not be an improvement. My argument is that I believe it is your job as a player to maximize your efficiency and that includes minimizing your damage intake at all times. This is even more important for tanks to make yourself a lot easier to heal and increasing your survivability (if done properly). I don't think most players would be inclined to start casting a defensive CD on cooldown when they see a suggestion of *>60% cast efficiency*, especially if you add the extra suggestion information you added.

In any case I'd like to keep using the tools available and if a new use cases arises modify or expand them to fit that use case. For the use case you're trying to solve here I believe a property could have been added to `CastEfficiency` that disabled the recommended cast efficiency. In fact, I just added that in my local environment a minute ago but I decided to revert it for the reasons listed above. I'm not unwilling to build it again if you are convincing enough.

yajinni:
Making the changes and putting in a new pull request

PR #362, opened 2017-09-22
[Resto Druid] Changed display of T20 2pc

kfinch:
Put the 'just the facts' statistic front facing, relegated the idealized throughput number to tooltip.


buimichael:
Don't forget to update the changelog, also i'd appreciate a screenshot of the feature (if applicable) instead of having to pull it locally to test it.

kfinch:
2t20

This isn't a new feature, and it's the same information just displayed differently. Seems to me that shouldn't be a changelog entry?

buimichael:
"Don't forget to update the changelog, but only include changes that users might notice."
Some users might be confused to why this information is no longer showing throughput in the main text prior to this PR.

kfinch:
@buimichael Done.

buimichael:
Thx

PR #364, opened 2017-09-22
Enhance maelstrom other tweaks pr

Nighteyez07:
Adjust a few values for issues and update display text for better context


Nighteyez07:
Isn't the idea of the progressive load though to be able to see the numbers adjust as the data is processed? From other PRs Zero has actually preferred that I do it within the events like I am now.

anom0ly:
Gotcha. Sounds good then - Rest of it looked fine.

PR #365, opened 2017-09-22
Added 5 combo point finisher tracking

Thieseract:
In the combo point tracker I added tracking for how many of each finisher was used at 5 combo points as well as a suggestion for when the percentage is too low.


MartijnHols:
Thanks!

PR #368, opened 2017-09-23
[Core] Kil'Jaedens Burning Wish cast efficiency removed

ronaldpereira:
It shows cast efficiency numbers, but this trinket hasn't a cast event on it, so it always show 0 casts per fight, also triggering suggestions and 'can be improved', even if the player has used it off CD.


anom0ly:
Looks good - Just need to make sure we come back and clean up the code so we don't leave commented code everywhere. Thanks!

PR #370, opened 2017-09-23
[Vengeance] Changed the spec completeness from good to great

ronaldpereira:
As I previously stated, I've now implemented the last open suggestion: the Tier 20 2 and 4 piece buff/debuff tracker.

So, the next steps is to wait for new suggestions and tier 21 become available to add it's tracker too.


MartijnHols:
Great work, thanks! 👍 By studying the T21 descriptions and perhaps some PTR logs you might get a long way towards implementing them already.

PR #374, opened 2017-09-23
[Core] Included a condition to only show pet damage when is greater than 0

ronaldpereira:
In DamageDone.js the tooltip was always showing pets damage, even if it was equal to 0. I've included a condition to only show pets DPS contribution when it is higher than 0, in order to only show pets DPS just when you used a pet in the fight.
Demo:
Destruction Warlock damage done tooltip:
image
Vengeance Demon Hunter damage done tooltip:
image

### DIFF
--- a/src/Parser/Core/Modules/DamageDone.js
+++ b/src/Parser/Core/Modules/DamageDone.js
@@ -52,7 +52,7 @@ class DamageDone extends Module {
         )}
         value={`${formatNumber(this.total.effective / (this.owner.fightDuration / 1000))} DPS`}
         label="Damage done"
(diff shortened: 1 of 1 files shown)


MartijnHols:
I like consistently showing the same information so there can be no confusion about why something isn't showing up (that could be caused by a bug or some other condition the user is unaware of). A user can not know why something doesn't appear when he doesn't get any feedback. Showing `Pets contributed 0 DPS.` does not cause this confusion.

At the same time it may be useless for a few specs that *never* have pets, but even for them they could get trinkets that contribute pet damage. But I also think the cost of always showing this is negligible. In conclusion I think it should stay as it is. Thanks for the PR but I won't be merging this at this time.

ronaldpereira:
That's okay. I didn't thought of another pet's damage other than literally having pets. I apologize for the mistake :disappointed:

PR #375, opened 2017-09-24
Nlc traits

anom0ly:
Initial Pass at NLC Traits - Should account for all of the non-static traits. 

Please review and let me know of any updates as needed.


src/Parser/Core/Modules/NetherlightCrucibleTraits/ChaoticDarkness.js
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.CHAOTIC_DARKNESS_TRAIT.id}>
+            <SpellIcon id={SPELLS.CHAOTIC_DARKNESS_TRAIT.id} noLink /> Chaotic Darkness
+          </SpellLink>
+        </div>
+        <div className="flex-sub text-right">
+        Healing: {formatPercentage(this.owner.getPercentageOfTotalHealingDone(healing))} % <br />
+        Damage: {formatPercentage(this.damage / this.damageDone.total.effective)} %
```
sMteX: You can use `this.owner.getPercentageOfTotalDamageDone(this.damage)` for the same purpose (same with other modules).

anom0ly: Ahh, I didn't realize. Updated to use that function. Thanks!

MartijnHols:
Do you have a screenshot of the output? Mostly wondering since I noticed <br/>s in the output and these don't always make the best looking output.

src/Parser/Core/Modules/NetherlightCrucibleTraits/ChaoticDarkness.js
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.CHAOTIC_DARKNESS_TRAIT.id}>
+            <SpellIcon id={SPELLS.CHAOTIC_DARKNESS_TRAIT.id} noLink /> Chaotic Darkness
+          </SpellLink>
+        </div>
+        <div className="flex-sub text-right">
+        Healing: {formatPercentage(this.owner.getPercentageOfTotalHealingDone(healing))} % <br />
```
MartijnHols: Could also write this at `X % healing & Y % damage` to be more concise

src/Parser/Core/Modules/NetherlightCrucibleTraits/NLCTraits.js
```diff
@@ -0,0 +1,57 @@
...
+    tormentTheWeak: TormentTheWeak,
+    darkSorrows: DarkSorrows,
+  };
+
+  statistic() {
+    return (
+      <StatisticsListBox
+        title="Netherlight Crucible"
+        tooltip="This provides an overview of the increased provide by the Netherlight Crucible traits."
+        style={{ minHeight: 186 }}
```
MartijnHols: It might be best to not set a min height, especially not this high, since it often won't reach near that height. Besides with Masonry aligning the tiles properly now we don't have to worry about height anymore.

src/Parser/Core/Modules/NetherlightCrucibleTraits/MurderousIntent.js
```diff
@@ -0,0 +1,41 @@
...
+    const murderousIntentUptime = this.combatants.selected.getBuffUptime(SPELLS.MURDEROUS_INTENT_BUFF.id) / this.owner.fightDuration;
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.MURDEROUS_INTENT_BUFF.id}>
+            <SpellIcon id={SPELLS.MURDEROUS_INTENT_BUFF.id} noLink /> Murderous Intent
+          </SpellLink>
+        </div>
+        <div className="flex-sub text-right">
+          Uptime: {formatPercentage(murderousIntentUptime)} %
```
MartijnHols: `x % uptime` would be more consistent

anom0ly:
Example Image:
image

I will make the other changes to be more consistent with the rest of the statistics.

anom0ly:
Made the changes you requested. One this with the `<br />` tags is that without them, the right had side pushed into the left had as it tried to keep all of that text on a single line. The line break removes this issue.

Image of Box without `<br />` tag:
image

kfinch:
Have you considered adding a switch somewhere (maybe in NLCTraits?) that allows individual specs to specify they want to show only healing or only damage? This could be one way of solving the awkward display issues you're discussing above. For the record, I like the first image better.

PR #376, opened 2017-09-24
Havoc init

CollCrom:
Changed the havoc spec to be a little more organized. Some spells were not included, some were in the ABC module but not on the GCD, etc. Also made the Ret and Havoc CD tracker less cluttered.


MartijnHols:
Thanks!

PR #377, opened 2017-09-24
[core] Fixed the green on the ABCHealing bar to be the same as the Healing Done green

kfinch:


MartijnHols:
They're different things, so maybe they should have different colors 🤔 I left it as it is as I couldn't decide a good approach. Maybe regular active time should be yellow or something?

kfinch:
I've got an idea, will implement when home.

kfinch:
Ok I tried yellow and it looked terrible. Best I could come up with is blue/purple/red. Please don't leave it as that green slightly off from the other green, it's triggering me.

kfinch:
healingabc-2

kfinch:
@MartijnHols how do you feel about the above colors?

MartijnHols:
I think it needs further experimentation 😅

MartijnHols:
I guess we can merge this with both bars the same green 🤷‍♂️

kfinch:
Man, the blue/purple/red kinda kinda grew on me. What about green/purple/red?

kfinch:
Fine fine, I've gone back to just changing the green. You should consider the blue/purple/red tho, it's rad.

MartijnHols:
I guess I'll merge, but the different greens never bothered me 😅

PR #380, opened 2017-09-24
[Balance Druid] Completeness

Iskalla:
The balance module currently tracks:

Standard modules: Damage done, Downtime
Cast Efficiency for all rotational spells (Moons and Incarnation with IFE need work)
Moons usage
Uptimes (Moonfire, Sunfire)
Astral Power overcapping
Lunar and Solar Empowerment Overcapping
Lunar Strike casts without Lunar Empowerment that hit less than 3 targets

Modules that are planned:

Dots casted outside of pandemic
Short casts (unemp SW, NM) over 200% haste
OI (Legendary bracers) Procs wasted
ED (Legendary helm) buff dropped
Couple  minor things to fix

It's also needed to add whispers and other haste trinkets to the haste module

I believe the current status should be Good.
Every core module is in, the only thing left is dot spamming (this is a fairly complex module since monkin dots spread, so even when casted outside pandemic in main target, they may be well used).
There are still legendarys to add and some in-depth stuff too, but is pretty detailed.
I'd also like to make a much more complex and detailed Astral Power module, but the current one is good enough to check how much was wasted - not the reason

Everything planned is in:
https://github.com/Iskalla/WoWAnalyzer/projects/1

I was doubting between letting it as WIP or good... but i think every main thing is in.

### DIFF
--- a/src/Parser/BalanceDruid/CONFIG.js
+++ b/src/Parser/BalanceDruid/CONFIG.js
@@ -7,7 +7,7 @@ import CHANGELOG from './CHANGELOG';
 export default {
   spec: SPECS.BALANCE_DRUID,
   maintainer: '@Iskalla',
-  completeness: SPEC_ANALYSIS_COMPLETENESS.NEEDS_MORE_WORK, // When changing this please make a PR with ONLY this value changed, we will do a review of your analysis to find out of it is complete enough.
+  completeness: SPEC_ANALYSIS_COMPLETENESS.GOOD, // When changing this please make a PR with ONLY this value changed, we will do a review of your analysis to find out of it is complete enough.
   changelog: CHANGELOG,
   parser: CombatLogParser,
 };


MartijnHols:
Thanks for the PR and your work so far, keep up the good work! 👍 The spec is actually really well worked out right now, with many statistics and items and a couple of suggestions. But I believe suggestions are the most important part of any analyzer and I think these might be lacking at this time. I believe a lack of suggestions is going to hold back adaptation of this spec's analyzer and warrants keeping it at *needs more work* for now.

Statistics are often repackaged things that are fairly easy to get straight from WCL, so them alone do not provide enough value. Providing meaningful suggestions to improve is probably the most important part of the analyzer. It helps players of any kind of knowledge get better without having to ask for help and without needing someone knowledgeable to manually review logs which can be a time consuming task, not to mention that it might be daunting to ask for a stranger's opinion. 

Here is an example log that I would expect a fair amount of suggestions on (with higher importance being the bigger and/or easier to fix things):
https://wowanalyzer.com/report/7mNHfzQA2ZnyM6tV/14-Heroic+Sisters+of+the+Moon+-+Kill+(4:13)/Luvzy
This player is green parsing, so surely he has more things to improve than just ABC? I'm unfamiliar with the spec, so unable to tell what the player is doing wrong in this log.

Iskalla:
As I told you in discord:
Im ok with leaving it as WIP, as I said up there, but...
This player is not doing anything wrong in the whole combat but his downtime (which is actually pretty poor and affects a lot). 
His problem is awful legendaries, so besides him having back luck and the downtime... he's pretty much playing perfectly

This is a bad log for real (ofc, made for testing purposes)
https://wowanalyzer.com/report/Y2fzGK4nrt7bv1Zg/1-+Goroth+-+Kill+(3:33)/Athawn

MartijnHols:
Aha, I wasn't aware gear could play such a big role. The log I linked is nearly 400k (about 45%) below epic performance, which I reckon is a lot. I wonder if there's any way to point out it's mostly a gear issue to players? If I were this player I'd be fairly worried about why my performance is so poor, being told it's a gear issue would be a huge relieve.

Iskalla:
Hmmm I suppose I could do something about that but I see 2 big problems:

- Legendaries are useful or not based on specific bosses, this legendary configuration could work potentially on some scenarios, so unless coding stuff for every raid its not really viable

- Tier sets change with each raid too, and there's not a way to simply check "how many bonuses do I have", would require code changes with each tier (even more since we can now do things like 4+2), and also different tier sets are used, depending on ST or AoE, rn

I simply can't think of a way to do this that doesnt require maintaining code on each tier (and i thought this was a thing we wanted to avoid)

MartijnHols:
It is, I was hoping you might have an idea. Let's keep thinking on it ;)

PR #383, opened 2017-09-25
(Holy Priest) Serendipity implemented

enragednuke:
Added spell IDs
 * Serendipity
 * Hallowed Ground trait
 * Holy Priest T20 2P
 * Holy Priest T20 4P

Added module 'Serendipity', which includes
 * "Wasted Serendipity" statistic
 * "Fully missed holy word casts (via Serendipity)" suggestion
 * "Holy Priest T20 2P Bonus" item

image

image

image


enragednuke:
I wonder if this would be better splitting the Serendipity effect for each holy word (so two holy word Serendipity modules). Then just perhaps make a central Serendipity module that draws dependencies to those two and makes the statistics/suggestions. The downside would be a lot of repetition but everything would be better spread. Thoughts?

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  // Holy Word reduction spells (aka things that apply Serendipity)
+  serendipityProccers = {
+    [SPELLS.GREATER_HEAL.id]: SPELLS.HOLY_WORD_SERENITY.id,
+    [SPELLS.FLASH_HEAL.id]: SPELLS.HOLY_WORD_SERENITY.id,
+    [SPELLS.PRAYER_OF_HEALING.id]: SPELLS.HOLY_WORD_SANCTIFY.id,
+  }
```
kfinch: Binding Heal isn't that special of a case, and this structure is barely used. Rethink the structure of what you're doing here to allow you to cut down on duplicated code below.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+
+  getAffectedSpell(spellId) {
+    if (Object.keys(this.serendipityProccers).indexOf(spellId.toString()) === -1) {
+      return -1;
+    }
+    return this.serendipityProccers[spellId];
```
kfinch: There's no need for this to be a special function. Where below you're calling `getAffectedSpell(spellId)` and checking for -1, instead just do `this.serendipityProccers[spellId]` and check for undefined.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+      this.serendipityReduction += 1.0;
+    }
+
+    // Check for Piety for serendipityProccers
+    if (this.combatants.selected.hasTalent(SPELLS.PIETY_TALENT.id)) {
+      this.serendipityProccers[SPELLS.PRAYER_OF_MENDING_CAST.id] = SPELLS.HOLY_WORD_SANCTIFY.id;
+    }
+
+    // Modify Sanctify CD based on trait
+    this.holyWordMaxCooldown[SPELLS.HOLY_WORD_SANCTIFY.id] = 60 - ((this.combatants.selected.traitsBySpellId[SPELLS.HALLOWED_GROUND_TRAIT.id] || 0) * SPELLS.HALLOWED_GROUND_TRAIT.coeff);
```
kfinch: The other talented/traited things are given a default value which is then modified on_initialized(), like with Light of the Naaru above. Why are you handling this differently?

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+    const spellId = event.ability.guid;
+    if (HOLY_WORD_SPELL_IDS.indexOf(spellId) !== -1) {
+      this.holyWordCooldowns[spellId] = event.timestamp + this.holyWordMaxCooldown[spellId] * 1000;
+      this.casts[spellId] += 1;
+
+      // We are giving a buffer of 75% of CD due to the fact that the large
+      // majority of players would intentionally use spells to push holy words
+      // off cooldown if needed. This is a feelycraft metric that is open for
+      // modification or outright removal depending on opinions.
+      this.fullOvercast[spellId] += Math.floor(this._tempOvercast[spellId] / 1000 / (this.holyWordMaxCooldown[spellId] * 0.75));
```
kfinch: All Suggestions are basically feelycraft metrics, but this one may be misguided. As a player reading the suggestion that I missed X casts, it wouldn't be obvious to me that you actually mean I missed 0.75 * X casts. Why aren't you just using 1.0 here? Isn't that what a 'full' overcast is? Maybe I'm missing an element of Holy Priest play, but it seems like this is just a 'cast efficiency' suggestion that you got fancy with.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+    const overcastAsPercCasts = (
+      (overcastAsCastsSerenity + overcastAsCastsSanctify) / (this.casts[SPELLS.HOLY_WORD_SERENITY.id] + this.casts[SPELLS.HOLY_WORD_SANCTIFY.id] + overcastAsCastsSanctify + overcastAsCastsSerenity)
+    );
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SERENDIPITY.id} />}
+        value={`${formatNumber(combinedOvercast)}s`}
+        label="Wasted Serendipity"
+        tooltip={`${formattedOvercastSerenity} wasted Serenity CD and ${formattedOvercastSanctify} wasted Sanctify CD. This is effectively ${formatPercentage(overcastAsPercCasts)}% of your potential Holy Word casts missed from solely wasted Serendipity.`}
```
kfinch: This still seems like an over-glorified cast efficiency metric to me. Should it really be expressed as 'wasted seconds' rather than just a cast efficiency percent?

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,227 @@
...
+      this.serendipityProccers[SPELLS.PRAYER_OF_MENDING_CAST.id] = SPELLS.HOLY_WORD_SANCTIFY.id;
+    }
+
+    // Modify Sanctify CD based on trait
+    this.holyWordMaxCooldown[SPELLS.HOLY_WORD_SANCTIFY.id] = 60 - ((this.combatants.selected.traitsBySpellId[SPELLS.HALLOWED_GROUND_TRAIT.id] || 0) * SPELLS.HALLOWED_GROUND_TRAIT.coeff);
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (HOLY_WORD_SPELL_IDS.indexOf(spellId) !== -1) {
```
kfinch: use `(HOLY_WORD_SPELL_IDS.includes(spellId))` here instead

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: It's similar to how I'm filtering out >80% OH Sanctify hits as being "ineffective". Because of how Serendipity works, it is possible to push spells out faster than its original CD. 

For example, You can sanctify, but realize you need it 38 seconds later. A good player would cast it and try to get it off CD for 38 seconds later while a bad one might hold the spell for 38 seconds despite being able to get it off CD for both events.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: This would be solved by my proposal to split this module into two sub-modules. See my first comment on this PR for what I mean by that. I'm starting to think more and more that it would be better but still would like an extra opinion.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: Ah right, sorry. My main language isn't JavaScript so I forget calling for an undefined index won't throw errors.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: :+1:

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: Would it be better as a percentage of wasted versus effective serendipity? I think that would be good.

enragednuke:
Thanks for all the comments, Serendipity is a really weird mechanic so my coding got really sloppy comparative to normal when writing this PR.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
kfinch: Yeah, that may be the best solution, if you're feeling like a refactor.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
kfinch: Hmm, so you're trying to capture part of what's missing in cast efficiency for healing CDs, namely that holding for a several seconds each time when it comes off CD is fine, but holding for a few full CDs is not fine.

Maybe the spin the 0.75 into a constant at the top of the module, and give some indication in the suggestion text of how you arrived at the conclusion they were holding Holy Words too long.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
kfinch: While it's complicated to calculate the cooldown, gameplay wise does this really play much differently from a a medium static CD? I assume you're casting Flash Heal / PoH because you need the healing, and the Serendipity proc is just a happy side effect. So the Holy Words are ~40 second CD abilities that are a high priority to use when available, typically such a thing would just be baked into the cast efficiency module.

It would make sense to call them 'wasted seconds' if Serendipity was a talent, because you could then suggest the player spec something else if they're not using it. Because it's not a talent, the only action the player can take is have better cast efficiency.

If you'd still prefer to keep this stat, my recommendation would be to change it to "per cast". I also wouldn't mind seeing the average CD reduction in the tooltip or something. Something like "Serendipity Reduction - 17s per Holy Word" ... "Serendipity cut an average of 17s off the CDs of your Holy Word: Sanctify and Holy Word: Serenity" ... "You held your Holy Words for an average of 13s after they became available."

kfinch:
It's cool, that's what reviews are for. Most of my comments are on design considerations that you are free to disagree with. Once you push the minor style stuff to the branch I'll approve.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: ```I assume you're casting Flash Heal / PoH because you need the healing, and the Serendipity proc is just a happy side effect.```

This actually isn't always the case. People cast Heal (slow ver.) very commonly to proc Serendipity. People often also cast Flash Heal for Serendipity if it can be brought off CD for heavy ST damage. I can't say it happens too often for Sanctify though, since PoH is quite expensive.

I want to keep this metric more towards some sort of wasted, since anything beneficial is probably not really relevant to know (it doesn't matter how much of a CD was reduced by Serendipity, but it does matter if someone is dumping spells that proc Serendipity while the respective spell is off CD). Since this is also Serendipity related, I imagine something like "average wasted Serendipity" would be a decent alternative (waiting isn't a bad thing since their CDs are decently sized, but as mentioned before, spamming when it's off CD is).

kfinch: I'll defer to your knowledge of the spec. Push the code style changes and I will approve.

enragednuke: I've had a lot of extra work at my job as of late, I'll let you know on Discord when I've gotten to doing all of the stylistic/refactor stuff mentioned.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,93 @@
...
+        'raw': formatNumber(this.sanctify.rawReduction / 1000),
+      },
+      'asCasts': {
+        'serenity': (this.serenity.overcast / 1000) / this.serenity.maxCooldown,
+        'sanctify': (this.sanctify.overcast / 1000) / this.sanctify.maxCooldown,
+      },
+    };
+
+    const percWastedVersusTotal = (this.serenity.overcast + this.sanctify.overcast) / (this.serenity.rawReduction + this.sanctify.rawReduction);
+    console.log(percWastedVersusTotal);
```
kfinch: put log lines behind a const 'debug' boolean, or remove entirely

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,93 @@
...
+// - "Holy Priest T20 2P Bonus" item
+class Serendipity extends Module {
+  static dependencies = {
+    combatants: Combatants,
+    serenity: SerenityReduction,
+    sanctify: SanctifyReduction,
+  }
+
+  statistic() {
+    const formatted = {
```
kfinch: Not sure that you need to make this 'formatted' structure. Most of these calculations are pretty short, why not do them inline? Or if you want the tooltip cleaner just make a bunch of const vals? Basically it looks like you made this a nested object for no good reason.

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
```diff
@@ -0,0 +1,93 @@
...
+        'wasted': formatNumber(this.serenity.overcast / 1000),
+        'raw': formatNumber(this.serenity.rawReduction / 1000),
+      },
+      'sanctify': {
+        'wasted': formatNumber(this.sanctify.overcast / 1000),
+        'raw': formatNumber(this.sanctify.rawReduction / 1000),
+      },
+      'asCasts': {
+        'serenity': (this.serenity.overcast / 1000) / this.serenity.maxCooldown,
+        'sanctify': (this.sanctify.overcast / 1000) / this.sanctify.maxCooldown,
```
kfinch: asCasts not used anywhere?

src/common/SPELLS_PRIEST.js
```diff
@@ -393,6 +399,13 @@ export default {
     coeff: 0.06,
   },
 
+  HALLOWED_GROUND_TRAIT: {
+    id: 196429,
+    name: 'Hallowed Ground',
+    icon: 'spell_holy_divineprovidence',
+    coeff: 3333,
```
kfinch: I think this value should be in the module, not in SPELLS. Also 'coeff' is a really vague name for this field.

src/common/SPELLS_PRIEST.js
```diff
@@ -431,6 +444,20 @@ export default {
     icon: 'inv_enchant_essencemagiclarge',
   },
 
+  // Holy Tier Sets
+  HOLY_PRIEST_T20_2SET_BONUS_BUFF: {
+    id: 242270,
+    name: 'Holy Priest T20 2P Bonus',
+    icon: 'spell_holy_guardianspirit',
+    coeff: 1000,
```
kfinch: I think this value should be in the module, not in SPELLS. Also 'coeff' is a really vague name for this field.

src/Parser/HolyPriest/Modules/PriestCore/SerendipityReduction/SerenityReduction.js
```diff
@@ -0,0 +1,90 @@
...
+      // We are giving a buffer of 75% of CD due to the fact that the large
+      // majority of players would intentionally use spells to push holy words
+      // off cooldown if needed. This is a feelycraft metric that is open for
+      // modification or outright removal depending on opinions.
+      this.effectiveFullOvercasts += Math.floor(this._tempOvercast / 1000 / (this.maxCooldown * FULL_OVERCAST_LENIENCE));
+      this._tempOvercast = 0.0;
+    }
+
+    // const affectedSpellId = this.getAffectedSpell(spellId);
+    if (Object.keys(this.serendipityProccers).indexOf(spellId.toString()) !== -1) {
```
kfinch: Should be `if(spellId.toString() in this.serendipityProccers)`, or alternately you could do 
```
const serendipityReductionFactor = this.serendipityProccers[spellId];
if(serendipityReductionFactor) {
...
```

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: forgot about it

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: oh, this was for the old tooltip logic I believe, yeah it can be removed

src/Parser/HolyPriest/Modules/PriestCore/Serendipity.js
enragednuke: Just for a cleaner look than a bunch of camel cased variables, no other reason.

src/Parser/HolyPriest/Modules/PriestCore/SerendipityReduction/SerenityReduction.js
enragednuke: ah yes, was too used to the structure I've been mimicking from other modules

src/common/SPELLS_PRIEST.js
enragednuke: I think it should remain here since it's a trait and it's value could be changed in a future patch (unlikely but still possible). Maybe change the variable name but keeping the value itself is good for future-proofing

src/common/SPELLS_PRIEST.js
enragednuke: I think it should remain here since it's a trait and it's value could be changed in a future patch (unlikely but still possible). Maybe change the variable name but keeping the value itself is good for future-proofing

enragednuke: Think of how the T20 2P for Disc Priest was recently nerfed in the patch. Had this value been used in multiple spots (as it is with my code), all you would have to do is change this one value in SPELLS files. Covers multiple bases and allows for new contributors to easily modify the value.

MartijnHols:
Thanks!

PR #392, opened 2017-09-26
Resto Druid: Added suggestions on sotf

buimichael:
Image


MartijnHols:
Would it be possible to prepend the suggestion with a simplified version of what was observed? e.g. `You didn't use all your [Wild Growth] procs from [Soul of the Forest].`

buimichael:
^ Done

PR #393, opened 2017-09-26
Resto druid flourish display changes

buimichael:
[Screenshot from Gyazo](https://gyazo.com/41713a2a4d8164bd51b6c7bce5e5fa74)

### DIFF
--- a/src/Main/ExpandableStatisticBox.js
+++ b/src/Main/ExpandableStatisticBox.js
@@ -8,6 +8,7 @@ class ExpandableStatisticBox extends React.PureComponent {
     label: PropTypes.node.isRequired,
     expanded: PropTypes.bool,
     children: PropTypes.node,
+    tooltip: PropTypes.string,
   };
 
   constructor() {
@@ -25,6 +26,7 @@ class ExpandableStatisticBox extends React.PureComponent {
       value: this.props.value,
       label: this.props.label,
       expanded: this.props.expanded,
+      tooltip: this.props.tooltip,
     });
   }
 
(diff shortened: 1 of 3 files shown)


kfinch:
Oh I am too late... I really hate 'average seconds extended' as a Flourish metric, because it's a really bad one. Breaking it down by cast doesn't make it any better. Also, it seems silly to show both Duration and # of HoTs, because the former will always be 6 times the latter. We should find a way to do throughput of extended HoTs sooner rather than later.

buimichael:
@kfinch 
I'd have to disagree that checking the amount of HoTs being extended as a bad metric. 
By breaking it down by cast we can see where we could have utilized a better flourish.
Also I don't see a problem with showing the same data in two different formats (same as we show healing done by number and by a bar atm)

I agree that HPS contribution should be used instead of average seconds in the main display. Flourish in a sense is very similar to Evangelism and tbh, what we want is probably exactly what the Evanglism box is showing:
 [Screenshot from Gyazo](https://gyazo.com/d3d0c8191ca6761c65a0be9817a5a97c)

Although tracking the healing contribution of each flourish may be complex to implement, especially accounting WG healing and rejuv healing. But it sure is possible.

kfinch:
There is one very important difference between Evangelism and Flourish, and that's that all Atonements are created equal but all HoTs are not. The healing done by 6 seconds of Cenarion Ward is much higher than that of 6 seconds of Wild Growth is much higher than 6 seconds of Cultivation / Germination / Rejuvenation is much higher than 6 seconds of Spring Blossoms / Regrowth.

buimichael:
Yeah I agree with that, we should probably show the duration/count of each HoT instead of treating them as equal.

PR #405, opened 2017-09-26
[Marksmanship] Minor Windburst cast efficiency update and CooldownTracker clean-up

Pewtro:
Made some minor changes to Windburst cast efficiency (bumping it from 0.7 to 0.85) as preliminary work before adding suggest etc to more spells

Cleaned up the cooldown tracker for MM to not include the movement-speed buff left on the ground. 

Before: 
before
After:
after


src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
```diff
@@ -12,6 +14,19 @@ class CooldownTracker extends CoreCooldownTracker {
...
     },
   ];
+
+  castEventSpells = [
+    SPELLS.WINDBURST_MOVEMENT_SPEED.id,
+  ];
+
+  on_byPlayer_cast(event) {
+    const spellID = event.ability.guid;
+    if (this.castEventSpells.indexOf(spellID) !== -1)  {
```
sMteX: This could be `this.castEventSpells.includes(spellID)` which is a little more readable and works the same.

src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
```diff
@@ -12,6 +14,19 @@ class CooldownTracker extends CoreCooldownTracker {
       ],
     },
   ];
+
+  castEventSpells = [
```
sMteX: Suggestion for future - if more spells are added here, perhaps it could be renamed as `ignoredSpells` or something similar as that's what's happening in `on_byPlayer_cast()`.

Issue #387, opened 2017-09-25
Cooldown Tracker should not include boss abilities


MartijnHols:
Astral Purge here should be excluded, the default view should only show player casts:

![image](https://user-images.githubusercontent.com/4565223/30831568-2a7a2ab2-a248-11e7-8d05-f39a554e0041.png)

sMteX:
Fixed with #407

PR #407, opened 2017-09-26
[Issue #387] Cooldown Tracker should not include boss abilities

sMteX:
**Original issue**: #387 

Not sure if there is some general solution to boss abilities, because even in the mentioned example, the casts are of Astral Vulnerability and it's from a player to a friendly player, so there's no way of knowing if that's a boss specific mechanic (just by looking at the event object). 

Instead I added a simple static array that contains spell IDs you wish to ignore. This can be furthermore extended by spec CooldownTrackers to add additional ignored spells (I suggested that in PR #405 and eventually rewrote that in this PR - works completely the same), for example see Marksmanship Hunter Cooldown Tracker.

Before: 
cooldowns-before

After:
cooldowns-after

### DIFF
--- a/src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
+++ b/src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
@@ -2,8 +2,6 @@ import SPELLS from 'common/SPELLS';
 
 import CoreCooldownTracker, { BUILT_IN_SUMMARY_TYPES } from 'Parser/Core/Modules/CooldownTracker';
 
-const debug = false;
-
 class CooldownTracker extends CoreCooldownTracker {
   static cooldownSpells = [
     ...CooldownTracker.cooldownSpells,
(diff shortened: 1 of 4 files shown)


kfinch:
I'm pretty sure you have write access, I'll let you do the update/merge when you're ready.

anom0ly:
Sorry for coming late to the game on this one.

Mistweavers make use of similar functionality. Can you make the same adjustments that were made to MM to the MW module?

Thanks.

src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
```diff
@@ -15,18 +13,10 @@ class CooldownTracker extends CoreCooldownTracker {
     },
   ];
 
-  castEventSpells = [
+  static ignoredSpells = [
+    ...CooldownTracker.ignoredSpells,
```
anom0ly: Can you extend these same changes to the Mistweaver Module also? Thanks.

MartijnHols:
http://www.wowhead.com/spell=207630/annihilation as well

src/Parser/MarksmanshipHunter/Modules/Features/CooldownTracker.js
MartijnHols: Wouldn't it make sense to do this in the core?

anom0ly: I believe my comment was not clear. MW make use of a similar function to remove the extra cast triggers for Chi Burst and Refreshing Jade Wind, for example. So to keep it consistent just also make those changes to the MW `CooldownTracker` module. I didn't mean to extend the core boss ability clean up to Mistweavers. I agree that should be done in the core parser.

sMteX:
I couldn't find any sensible log when there was a cast event of Annihilation during a cooldown, but looking at the Events tab, I looked for Annihilation cast and added that so I hope it's okay.

anom0ly:
Looks good from my end. Just need to aware this would need to be monitored tier by tier.

PR #409, opened 2017-09-27
added some more suggestions for the ret spec

CollCrom:


src/Parser/RetributionPaladin/Modules/Items/WhisperOfTheNathrezim.js
```diff
@@ -35,14 +37,25 @@ class WhisperOfTheNathrezim extends Module {
     return {
       item: ITEMS.WHISPER_OF_THE_NATHREZIM,
       result: (<dfn data-tip={`
-				The effective damage contributed by Whisper of the Nathrezim.<br/>
+        The effective damage contributed by Whisper of the Nathrezim.<br/>
 				Damage: ${this.owner.formatItemDamageDone(this.damageDone)}<br/>
```
sMteX: This is a little bit redundant since you're showing that as the item's main text.

CollCrom: So here are some examples where I want the cloak suggestion to show up

https://www.warcraftlogs.com/reports/RY9yFXkVpZqtzT1B#fight=2&type=damage-done


https://www.warcraftlogs.com/reports/aLKgjkyQNWbc2C91#fight=2&type=damage-done
untitled


https://www.warcraftlogs.com/reports/BMxLfbYy12pADGCN#fight=1&type=damage-done
untitled



Obviously these aren't the only things contributing to these lower than orange parse but they are part of it.

Issue #411, opened 2017-09-27
Netherlight Crucible Statistics Box Always Showing


anom0ly:
Thanks for reporting a new issue, it's much appreciated!

The following questions might help achieve a faster resolution, but you're in no way required to answer any them. Please remove this text and any questions you don't deem relevant.

 - What spec? All Specs
 - Link to an example report with the issue: https://wowanalyzer.com/report/1Ry2N98TjfbBck4t/6-Mythic+Mistress+Sassz'ine+-+Kill+(7:39)/%E5%A1%9E%E7%BA%B3%E9%87%8C%E5%A5%A5%E5%8F%9B%E5%BE%92
 - What's your Discord nickname? (if we can contact you on our Discord server) Anomoly
 - Expected behavior: 
Netherlight Crucible Statistics box is showing even when a player has no Netherlight Crucible traits. This should be hidden if no traits are present for the player.

anom0ly:
Fixed by PR #412

PR #412, opened 2017-09-27
Hot Fix: Issue #411 - NLC Traits Showing at All Times

anom0ly:
Hot Fix for Issue #411


src/Parser/Core/Modules/NetherlightCrucibleTraits/NLCTraits.js
```diff
@@ -29,6 +29,18 @@ class RelicTraits extends Module {
     darkSorrows: DarkSorrows,
   };
 
+  activeTraits = false;
+
+  on_initialized() {
+    // Deactive this module if none of the underlying modules are active.
+    Object.keys(NLCTraits.dependencies).forEach((key) => {
```
MartijnHols: Could make it fully functional;

```jsx
this.active = Object.keys(this.constructor.dependencies)
  .map(key => this.constructor.dependencies[key])
  .some(dependency => dependency.active);
```

anom0ly:
Adjusted based on your feedback. Had to change, but this implementation is working. Thanks again for the feedback.

PR #422, opened 2017-09-28
[Core] Added a spec information overlay

MartijnHols:
**Reviews welcome but please let me merge this myself.**

Example:

image

- Deprecated `config.footer` in favor of this overlay.
- I do want to make the button to open the overlay more pronounced, but will do that later.
- I'm not a big fan of hiding spec completeness behind the overlay, but it will have to do for now as I don't know of a quick decent alternative.
- Spec source links directly to spec directory on GitHub.
- Background gets blurred so it doesn't bleed through the background opacity which would make things a lot harder to read. 
- Minor Results refactoring to make it slightly more maintainable. 
- Maintainer should probably become an array some day, might need more refactoring I felt it was out of scope for now and the current solution is sufficient for a long time.
- I want to make the changelog an array with each entry an object with date, description, and changeType (or something like that) where changeType determine the icon type; edit, add, remove, bug(fix), and possibly others. I felt it was out of scope for now.

Nothing set in config has a decent default:

image

**Reviews welcome but please let me merge this myself.**


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -3,6 +3,7 @@
     <option name="myName" value="Project Default" />
     <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
     <inspection_tool class="ExceptionCaughtLocallyJS" enabled="false" level="WARNING" enabled_by_default="false" />
+    <inspection_tool class="HtmlUnknownAnchorTarget" enabled="false" level="WARNING" enabled_by_default="false" />
```
MartijnHols: First time running into the inspection; it's not going to be useful with React.

config/webpack.config.dev.js
```diff
@@ -256,6 +256,7 @@ module.exports = {
     fs: 'empty',
     net: 'empty',
     tls: 'empty',
+    __dirname: true,
```
MartijnHols: Without this `__dirname` is unavailable.

anom0ly:
I like the design / direction and the information. I'll echo @kfinch comment around the title being slightly larger than it potentially needs to be. Potentially dropping `analyzer` from the title would help reduce some of that clutter.

As for the `description` area, I think potentially what is listed in the example is more of the vision for the overall analyzer, just centered on some specifics for Paladins. I wonder if this area would be better served as a description of the specific maintainer / their contact information and links to the relevant class TC resources?

Other than that, I think this is a really solid, awesome work.

MartijnHols:
`all over the place`

Yeah it needs some more fiddling. Thanks for the feedback, will update when I can.

MartijnHols:
Thanks for the feedback.

I do think the title is the right size. I decreased the maintainer name size a bit, I reckon it looks decent enough. The description can't get larger as it wouldn't look right and take too much space. The maintainer positioning is to the right with some positioning to make it appear to float just underneath the title. I reckon it works ok enough. Slightly reduced the close button and aligned it to body instead of the container. Made the completeness/spec discussion/source code container the same size as the description, made everything a bit tighter vertically and made the changelog wider. I reckon it's good enough to merge now.

Disclaimer: I'm not a graphic designer, most of it is imitation and while I don't shy away from saying I'm a CSS wizard, I don't really know what I'm doing when it comes down to making things pretty until I accidentally turn the knobs just right.

With regards to the description area, you can enter whatever you want. I felt this most applicable for Holy Paladin.

MartijnHols:
image

PR #423, opened 2017-09-28
[Demonology] Added T20 set bonuses

sMteX:
Keeping it small. I wanted to add the ability to somewhat count the correct number of possible Call Dreadstalkers since the T20 2pc resets the cooldown, but that proved to be difficult to say the least. I tried doing `fightDuration / baseCD + numberOfResets` but that's kinda fundamentally wrong, since each time you reset the CD, it alters the timestamps at which the cooldowns would be "normally" ready.

In conclusion - no clue how to deal with ability resets, I left it uncorrected with base cooldown of 15 seconds. Also, if you haven't heard on Discord, Dead time for Demo is still severely broken and I've got precisely 0 clue why. Any help appreciated O:)

### DIFF
--- a/src/Parser/DemonologyWarlock/Modules/Items/T20_2set.js
+++ b/src/Parser/DemonologyWarlock/Modules/Items/T20_2set.js
@@ -36,7 +36,7 @@ class T20_2set extends Module {
       id: `spell-${SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id}`,
       icon: <SpellIcon id={SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id} />,
       title: <SpellLink id={SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id} />,
-      result: `${this.procs} resetted cooldowns on Call Dreadstalkers`,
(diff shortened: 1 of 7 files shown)


src/Parser/DemonologyWarlock/Modules/Items/T20_2set.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+    this._expectedCooldownEnd = event.timestamp + CALL_DREADSTALKERS_COOLDOWN;
+  }
+
+  item() {
+    return {
+      id: `spell-${SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id}`,
+      icon: <SpellIcon id={SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id} />,
+      title: <SpellLink id={SPELLS.WARLOCK_DEMO_T20_2P_BONUS.id} />,
+      result: `${this.procs} resetted cooldowns on Call Dreadstalkers`,
```
anom0ly: Could this be words `X resets of Call Dreadstalkers cooldown`? Resetted isn't a word so it just reads funny to me.

PR #424, opened 2017-09-29
[Enhancement] Alpha Wolf usage check

Nighteyez07:
Add detection for proper usage of Crash Lightning as priority spell directly after Feral Spirits used

### DIFF
--- a/src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
+++ b/src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
@@ -1,0 +1,41 @@
+import SPELLS from 'common/SPELLS';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+import Module from 'Parser/Core/Module';
+
+class AlphaWolf extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  feralSpiritCastCount = 0;
+  feralSpiritCastFlag = false;
+  missedAlphaWolfCount = 0;
+
+  on_byPlayer_cast(event) {
+    if(this.feralSpiritCastFlag) {
+      if(event.ability.guid !== SPELLS.CRASH_LIGHTNING.id) {
+        this.missedAlphaWolfCount += 1;
+      }
+      this.feralSpiritCastFlag = false;
+    }
+
+    if(event.ability.guid === SPELLS.FERAL_SPIRIT.id) {
+      this.feralSpiritCastFlag = true;
+      this.feralSpiritCastCount += 1;
+    }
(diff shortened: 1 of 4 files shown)


src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
```diff
@@ -0,0 +1,41 @@
...
+
+import Module from 'Parser/Core/Module';
+
+class AlphaWolf extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  feralSpiritCastCount = 0;
+  feralSpiritCastFlag = false;
```
sMteX: These two could be prefixed with `_` to indicate private variables (since the main variable for this module is `missedAlphaWolfCount`) but it's just a style issue.

src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
```diff
@@ -0,0 +1,41 @@
...
+  missedAlphaWolfCount = 0;
+
+  on_byPlayer_cast(event) {
+    if(this.feralSpiritCastFlag) {
+      if(event.ability.guid !== SPELLS.CRASH_LIGHTNING.id) {
```
MartijnHols: This will also trigger for boss triggered casts like on Sisters.

src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
```diff
@@ -0,0 +1,41 @@
...
+    if(event.ability.guid === SPELLS.FERAL_SPIRIT.id) {
+      this.feralSpiritCastFlag = true;
+      this.feralSpiritCastCount += 1;
+    }
+  }
+
+  suggestions(when) {
+    when(this.missedAlphaWolfCount).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest('Make sure Crash Lightning is always cast directly after Feral Spirits to activate Alpha Wolf, even on single target')
```
MartijnHols: Please include what was observed. E.g. You didn't always casts Chain Lightning immediately after Feral Spirits. 

Also please turn the spells into SpellLinks.

src/Parser/EnhancementShaman/Modules/ShamanCore/AlphaWolf.js
MartijnHols: "Flag" in the name can be omitted. If you want to make it clearer what it is how about `_didCastFeralSpirit`?

PR #189, opened 2017-08-22
[Guardian] Rage waste statistic

FaideWW:
Displays wasted rage as a statistic and suggestion, including % of total rage gained that was wasted and a breakdown of waste by spell.

PR #426, opened 2017-09-29
[Guardian] Minor fixes

FaideWW:
Fixes the following issues:
 - Total rage gained in the rage waste module was being calculated incorrectly
 - Defensive and utility cooldowns no longer show "Can be improved" in CastEfficiency
 - The Ironfur statistic box now shows effective uptime rather than total uptime


src/Parser/GuardianDruid/Modules/Features/RageWasted.js
```diff
@@ -73,7 +73,8 @@ class RageWasted extends Module {
         const realRageWasted = Math.floor((this._currentRawRage + RAW_RAGE_GAINED_FROM_MELEE - this._currentMaxRage) / 10);
         this.registerRageWaste(event.ability.guid, realRageWasted);
       }
-      this.totalRageGained += RAW_RAGE_GAINED_FROM_MELEE;
+      // Convert from raw rage to real rage
+      this.totalRageGained += RAW_RAGE_GAINED_FROM_MELEE / 10;
```
kfinch: So you're changing `this.totalRageGained` to be holding a 'real rage' value, but I see you're not changing how it's added to in other places. On line 67 I see `this.totalRageGained += event.resourceChange + event.waste;`, does this mean event.resourceChange and event.waste are expressed as real rage? If they're raw rage, this is now a bug. Also on line 89 you have `return this.totalWastedRage / this.totalRageGained;`, same question applies here.

FaideWW: This is intentional; see my discussion with Zero on #189.  `energize` events deal with "real rage", but `classResources` (and by extension my handling of melee attacks) are expressed as "raw rage".  

The bug being fixed here is that totalRageGained should be a "real rage" value, but here I was adding the "raw rage" value for melee attacks.

MartijnHols: *real rage* and *raw rage* are confusing, can we agree to call it *in-game-rage multiplied by 10* instead?

FaideWW: Sure. To clarify: 

The `energize` event properties `resourceChange` and `waste` represent in-game rage values.  

The property `classResources` represents in-game rage, multiplied by 10 (this is common among non-mana Power resources like rage, energy, Astral Power, etc. since presumably it allows the game engine to apply resource multipliers while still dealing with integer values).

Since melee attacks don't generate `energize` events, I store the current and maximum values from `classResources` (represented as in-game rage * 10) to compute potential rage waste when a melee attack occurs.  When I log the values for wasted rage and total rage gain, I convert into in-game rage to be consistent with the values from `energize` events.

kfinch: Ahh, that's weird of blizzard, but makes sense.

MartijnHols:
Thanks!!

PR #429, opened 2017-09-30
[resto druid] More flourish enhancements

kfinch:
I made the Flourish changes I was talking about, most obviously I switched the main stat to 'Average HoTs Extended'.

In the table I replaced seconds extended with 'HoT Power', which I explain in the tooltip.

I also refactored the backend because it was very out of date. I updated Wild Growths to tell you about casts extended rather than individual HoTs extended, which further obviates the need to check for ToL.

I updated the tooltip a bit.


kfinch:
new-flourish

kfinch:
@buimichael needs reapproval after I fixed missing semicolon

buimichael:
@kfinch One thing that I forgot, is the rejuvenation % SP. You need to adjust the %SP by the persistence level too.

kfinch:
Persistence is increased duration at same HPS. The %SP constants are indications of HPS, not total healing.

buimichael:
We may want to include this:
Incarnation: Tree of Life: Healing increased by 50%

kfinch:
Does the rejuv bonus apply to buffs applied during the duration, or is it a dynamic buff that applies to all rejuv ticks while ToL is up?

buimichael:
@kfinch It's 100% dynamic that applies to all rejuvs. Also in addition we could add a 15% modifier overall to all HoTs because of ToL.

PR #430, opened 2017-09-30
[Demonology] Legendaries, round 1

sMteX:
I'll split the legendaries in two, next batch will include Kazzak's Final Curse and maybe Wilfred's Sigil of Superior Summoning. I'll also look into all existing specs for some refactoring (I think messing with pet IDs isn't completely necessary since we have `byPlayerPet_damage` now).


src/Parser/DemonologyWarlock/Modules/Items/Legendaries/SindoreiSpite.js
```diff
@@ -0,0 +1,46 @@
...
+    // so friendly TKC damage doesn't get counted
+    if (!event.targetIsFriendly) {
+      this.bonusDmg += getDamageBonus(event, SINDOREI_SPITE_DAMAGE_BONUS);
+    }
+  }
+
+  item() {
+    return {
+      item: ITEMS.SINDOREI_SPITE,
+      result: `${formatNumber(this.bonusDmg)} damage contributed - ${this.owner.formatItemDamageDone(this.bonusDmg)}`,
```
kfinch: Use formatItemDamageDone() from the core CombatLogParser instead

src/Parser/DemonologyWarlock/Modules/Items/Legendaries/WakenersLoyalty.js
```diff
@@ -0,0 +1,64 @@
...
+    this._currentStacks = 0;
+  }
+
+  item() {
+    const tkcCasts = this.abilityTracker.getAbility(SPELLS.THALKIELS_CONSUMPTION_CAST.id).casts || 1; // it's for averages, so 1 in denominator as a default value
+    return {
+      item: ITEMS.WAKENERS_LOYALTY,
+      result: (
+        <dfn data-tip={`Average bonus damage per TKC cast: ${formatNumber(this.bonusDmg / tkcCasts)}<br />Average stacks of Wakener's Loyalty per TKC cast: ${(this._totalUsedStacks / tkcCasts).toFixed(2)}`}>
+          {formatNumber(this.bonusDmg)} damage contributed - {this.owner.formatItemDamageDone(this.bonusDmg)}.
```
kfinch: Use formatItemDamageDone() from the core CombatLogParser instead

src/common/SPELLS_WARLOCK.js
```diff
@@ -703,4 +703,11 @@ export default {
     name: 'Dreaded Haste',
     icon: 'spell_shadow_metamorphosis',
   },
+
+  // Demonology legendary effects
+  WAKENERS_LOYALTY_BUFF: {
```
kfinch: Use the spell's actual name when possible. This should just be WAKENERS_LOYALTY. If a spell's name makes it unclear what it's from (like a buff with a different name from the legendary that procced it), you can add clarification in a comment. In this case, I don't think a comment is needed.

src/Parser/DemonologyWarlock/Modules/Items/Legendaries/WakenersLoyalty.js
kfinch: Keeping the tooltip is fine, but the front facing number should be from formatItemDamageDone

src/Parser/DemonologyWarlock/Modules/Items/Legendaries/SindoreiSpite.js
sMteX: I'd like to show the damage contributed anyway, perhaps move it to tooltip? Same with Wakener's Loyalty.

src/common/SPELLS_WARLOCK.js
sMteX: I initially thought of it to be clear distinction between the item and the buff but considering one is in ITEMS and second in SPELLS, that kinda contradicts my argument.

src/Parser/DemonologyWarlock/Modules/Items/Legendaries/SindoreiSpite.js
kfinch: Oh wow for some reason I didn't even see that you had format item damage done as like the second format. I'm spacing out today. Yeah I think formatItemDamageDone should be solo here, and other notes should be in the tooltip.

src/common/SPELLS_WARLOCK.js
kfinch: Yeah there are a lot of cases with same name in ITEMS and SPELLS

src/Parser/DemonologyWarlock/Modules/Items/Legendaries/SindoreiSpite.js
sMteX: Ok, I'll rewrite it here, and in second PR I'm working on right now, I'll refactor it in other item modules as well (there will be other refactoring as well so I'll do that as well).

kfinch: Thanks! I'm determined to see a common format for showing Legendary/Set damage contribution. If you really think raw damage should show too, maybe make the argument for a change to formatItemDamageDone.

sMteX: I wouldn't mind being in the format as well, but I think it can stay in tooltip and show percentage / DPS by default.

PR #435, opened 2017-09-30
[core] custom cast efficiency thresholds

kfinch:
Zero: I was feeling constrained by the recommended, recommended-0.05, recommended-0.15 thresholds, so I added the ability to specify custom thresholds. Default behavior and behavior when only recommended is specified should remain exactly the same. I hope this implementation looks OK to you.

blazyb: Due to the nature of how Tranq works, I felt the analyzer was reporting not enough tranqs as a Major issue way too frequently. Considering adding some more explanation in extraSuggestion field. What do you think?

I also added handling for Stonebark CD reduction.


src/Parser/RestoDruid/Modules/Features/CastEfficiency.js
```diff
@@ -29,14 +32,16 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.IRONBARK,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: haste => 90,
+      getCooldown: (haste, combatant) => combatant.hasTalent(SPELLS.STONEBARK_TALENT.id) ? 60 : 90,
```
sMteX: `haste` can be omitted as it's not used. `(_, combatant)` should be enough. I guess it's a question of personal preference, if you want to express it's not affected by haste.

kfinch: gottem

kfinch:
Ok I further just updated talent getters in CastEfficiency, made Ironbark handle Xonis as well, and added Renewal.

src/Parser/Core/getCastEfficiency.js
```diff
@@ -64,6 +67,8 @@ export default function getCastEfficiency(CPM_ABILITIES, abilityTracker, combata
         overhealing,
         castEfficiency,
         recommendedCastEfficiency,
+        avgIssueCastEfficiency,
```
MartijnHols: I'd prefer spelling this out (average), consistent with the other props.

kfinch: Fixed

MartijnHols:
Thanks!

PR #448, opened 2017-10-03
Cast efficiency updates and added some cool down tracking.

yajinni:
New pull request based on changes asked to be done by zero in my previous PR after we talked.


src/Parser/BloodDeathKnight/Modules/Items/T20_4pc.js
```diff
@@ -25,17 +26,21 @@ class T20_4pc extends Module {
...
     if (spellId === SPELLS.DEATH_STRIKE.id && !this.combatants.selected.hasBuff(SPELLS.GRAVEWARDEN.id, event.timestamp)) {
       this.deathStrinkeBuffless += 1;
+    }else{
+      if (spellId === SPELLS.DEATH_STRIKE.id) {
+        this.deathStrikeTotal += 1;
+      }
     }
-  }
 
+  }
```
sMteX: The whole event handler should be rewritten as 
```jsx
if (spellId !== SPELLS.DEATH_STRIKE.id) {
   return;
}
if (!this.combatants.selected.hasBuff(SPELLS.GRAVEWARDEN.id, event.timestamp)) {
   this.deathStrinkeBuffless += 1;
}
this.deathStrikeTotal += 1;
```

I think in your implementation, if you intend to show "X / Y casts ....", then Y should include also those X casts. That is - your `deathStrikeTotal` was in fact `buffedDeathStrikes` because it was in the else branch. Total Death Strikes should count those casts buffed and unbuffed.

sMteX:
Thanks for the corrections of what I pointed out although I'm not sure about the `.idea/inspectionProfiles/Project_Default.xml` change if I can merge that. Feels to me that something like this should be local-only change but I've never worked with it. @MartijnHols would you please look at that if it's fine?

MartijnHols:
`Here is what you need to share:
- All the files under .idea directory in the project root except the workspace.xml and tasks.xml files which store user specific settings`

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

sMteX:
So if I understand correctly the changes to .gitignore and the Project_Default.xml should be reverted?

yajinni:
Ok fixed it so that the PR can go though. but i am sure all those extra files will come back once i start up my IDE will need to figure out why its happening and what to do about them later.

PR #451, opened 2017-10-03
Combo Strikes Implementation and CONFIG Updates

anom0ly:


src/Parser/WindwalkerMonk/Modules/Features/ComboStrikes.js
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Track Details on the last 3 spells used - Need to populate up to 4 first, then begin to modify the array.
+    if(this._lastFourSpellsUsed.length < 3) {
+        this._lastFourSpellsUsed.push({
+            ability: spellId,
+            timestamp: eventTimestamp,
+        });
+    } else {
+        this._lastFourSpellsUsed = this._lastFourSpellsUsed.slice(1);
```
sMteX: Not particularly some fault, just my curiosity - I presume this is to remove the first element (`slice` with only one parameter returns a *new* array with the elements equal to the elements of the original array from the index forward, so from index 1 it's all the original elements except the first one). Is there a particular reason for copying the array and overwriting `this._lastFourSpellsUsed` instead of using `this._lastFourSpellsUsed.splice(0, 1)` ?

anom0ly: Mainly because when using `.splice` the array stored for recall during the statistic would just be the final three spells used, and not those used when the mastery chain was broken. Arrays and their functions in JS is one of my shortcomings, so this is the way I got it to work. :)

I am pushing an update to just remove some `console.log` messages and also rename some variables, but if you see anything else please let me know.

MartijnHols:
Thanks! :)

PR #453, opened 2017-10-03
[Core] EnemyInstance and EnemyInstances

FaideWW:
Adds a new Entity type `EnemyInstance`, and its corresponding module `EnemyInstances`.

I discovered when working on another feature that there is no support in the existing `Enemies` module for tracking individual adds throughout an encounter, since adds share a `targetId` and are only separated by `targetInstance`.   This PR adds that support by accounting for `targetInstance` when handling Enemy data.

The existing `Enemies` module has not been touched, to preserve compatibility with existing code.


MartijnHols:
Thanks!

PR #456, opened 2017-10-04
[Marksmanship Hunter] Focus Generation Tracker

leapis:
Based on Parser/VengeanceDemonHunter/Modules/PainChart

(Hopefully there aren't too many style issues- while some parts of the code could be more efficient, I tried to prioritize clarity over anything else, as I'm very inexperienced with open source development)

This is a module that:

- determines focus generator casts (where applicable)
- determines focus generated from each source
- determines focus wasted from each source
- attempts to replicate similar numeric value to Check My WoW (only known passive focus generation resource)
- displays said data in accurate numeric form and "inaccurate" (prioritizing visual accuracy over numerical accuracy) graphical form

Improvements that can potentially be made to this module that I have not included in this pull request:
Major:
- set values for "acceptable" focus waste from each source of generation. Will require a MMS expert to generate these values (Putro has said he will take care of this at some point),  right now the only value that exists is for passive focus and it's an estimate of what I believe CMW's is

Minor:
- implement realistic haste tracking, as of right now it assumes flat focus gen rate based on initial haste (Note: CMW has been the hunter go-to for years and still does not have this, so I classify it as a minor improvement)
- account for volley (-3 focus per proc), this is minor because, again, CMW does not account for this
- Better visual display/ percentage statistics (just as an example, CMW uses a progress bar to display, for each generator, non-wasted/total-generated)
- Add clarity wherever possible

As a spin off, perhaps create a feature that simply displays % time focus capped or total focus wasted

Screenshot:
image


src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
+//Based on Main/Mana.js and Parser/VengeanceDemonHunter/Modules/PainChart
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import ChartistGraph from 'react-chartist';
```
MartijnHols: This isn't something that we made clear in any way and you don't have to fix it (at all), but we moved to use ChartJS instead of Chartist and intended for all new charts to be built with that instead. The status charts are an example of this. Since the only functional gain for migrating this for you would be tooltips you're in no way required to change this (but if you're motivated to do so it would be greatly appreciated! ;)).

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+import makeWclUrl from 'common/makeWclUrl';
+import SPELLS from 'common/SPELLS';
+
+import specialEventIndicators from 'Main/Chartist/specialEventIndicators';
+
+import 'Main/Mana.css';
+
+import FocusComponent from './FocusComponent';
+import './Focus.css';
+const formatDuration = (duration) => {
```
MartijnHols: This can be replaced with
```
import { formatDuration } from 'common/format';
```

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+      .then((json) => {
+        if (json.status === 400 || json.status === 401) {
+          throw json.error;
+        } else {
+          this.setState({
+            focusData: json,
+          });
+        }
+      });
+	const playerPromise = fetch(makeWclUrl(`report/events/${reportCode}`, { //We really just grab for player stats- I'm sure there's a better way
```
MartijnHols: This entire is unnecessary; the same request is already fetched by the main parser and the resulting events are what trigger the event handlers in modules. This should be changed to use that instead.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+  componentWillMount() {
+    this.load(this.props.reportCode, this.props.actorId, this.props.start, this.props.end);
+  }
+  componentWillReceiveProps(newProps) {
+    if (newProps.reportCode !== this.props.reportCode || newProps.actorId !== this.props.actorId || newProps.start !== this.props.start || newProps.end !== this.props.end) {
+      this.load(newProps.reportCode, newProps.actorId, newProps.start, newProps.end);
+    }
+  }
+  load(reportCode, actorId, start, end) {
+    const focusPromise = fetch(makeWclUrl(`report/tables/resources/${reportCode}`, {
```
MartijnHols: This data should be available in the existing events handler. Please use that instead. You can for the time being leave this as it is (since I estimate changing it will take some time), but please do at some point change this as additional requests are precious.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+          Loading...
+        </div>
+      );
+    }
+	const actorId = this.props.actorId;
+	
+	let maxFocus = 0;
+	let found = false;
+	let ii = 0;
+	while (!found){ //determine max focus (max focus is variable)
```
MartijnHols: This should be in a separate module, e.g. a Focus module. You could probably copy most of the existing ManaValues modules.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+		if (this.state.playerData.events[ii].sourceID === actorId &&(this.state.playerData.events[ii].type === 'cast' || this.state.playerData.events[ii].type === 'energize') ){
+			maxFocus = this.state.playerData.events[ii].classResources[0]['max'];
+			found = true;
+		}
+		else{
+			ii++;
+		}
+	}
+	
+	const focusGen = Math.round((10 + .1 * this.state.playerData.events[0].hasteRanged / 375)*100)/100; //TODO: replace constant passive FocusGen (right now we don't account for lust/hero or Trueshot)
```
MartijnHols: If you pass the parser to this component on creation (in your CombatLogParser), you can get this with `parser.combatants.selected.hasteRating`.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+    const { start, end } = this.props;
+	let passiveCap = 0; //counts time focus capped (in seconds)
+	let lastCatch = 0; //records the timestamp of the last event
+	
+    const cappedTimer = {
+      0:maxFocus,
+    };
+    this.state.playerData.events.forEach((events) => {
```
MartijnHols: This should be a separate module with the appropriate event listeners.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+	This is the graph data- since the graph is just a general point of reference, and I only record 1 data point for
+	every second, this ensures that the range of each section of the graph is accurate, at the cost of exact slope*/
+	lastCatch = 0;
+	const overCapBySecond = {};
+    const focusBySecond = {
+      0: maxFocus,
+    };
+	this.state.focusData.series[0].data.forEach((item) => {
+	  const secIntoFight = Math.floor((item[0] - start) / 1000);
+	  if (focusBySecond[secIntoFight] && item[1] <= 65 && focusBySecond[secIntoFight] > item[1]){ //aims to get highest peak
```
MartijnHols: Please extract the `65` magic number to a constant.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+    });
+
+
+    const abilitiesAll = {};
+    const categories = {
+      generated: 'Focus Generators',
+      //spent: 'Focus Spenders', //I see no reason to display focus spenders, but leaving this in if someone later wants to add them
+    };
+
+    this.state.playerData.events.forEach((event) => {
```
MartijnHols: This too should be a separate module.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,374 @@
...
+import 'Main/Mana.css';
+
+import FocusComponent from './FocusComponent';
+import './Focus.css';
+const formatDuration = (duration) => {
+  const seconds = Math.floor(duration % 60);
+  return `${Math.floor(duration / 60)}:${seconds < 10 ? `0${seconds}` : seconds}`;
+};
+
+const passiveWasteThreshold = .03; // (wasted passive focus generated) / (total passive focus generated), anything higher will trigger "CAN BE IMPROVED"
```
MartijnHols: It would be clearer if you included "percentage" in the name of this.

src/common/SPELLS_HUNTER.js
```diff
@@ -102,6 +102,28 @@ export default {
...
+	id: 75,
+	name: 'Auto Shot',
+	icon: 'ability_hunter_focusfire',
+  },
+  CRITICAL_FOCUS_FOCUSMODULE: {
+	id: 215107,
+	name: 'Critical Focus',
+	icon: 'ability_druid_replenish',
+  },
+  MULTISHOT_FOCUSMODULE: {
```
MartijnHols: What's the difference between this and http://www.wowhead.com/spell=2643/multi-shot ?

src/common/SPELLS_HUNTER.js
```diff
@@ -102,6 +102,28 @@ export default {
...
+    id: 213363,
+    name: 'Multi-Shot',
+    icon: 'ability_upgrademoonglaive',
+  },
+  ARCANE_SHOT_FOCUSMODULE: {
```
MartijnHols: What's the difference with http://www.wowhead.com/spell=185358/arcane-shot ?

src/common/SPELLS_HUNTER.js
leapis: Just the ID. This the id that corresponds with multi-shot with the events data. http://www.wowhead.com/spell=213363/multi-shot actually less correct (it's the lvl 1 version, whereas 2364 is the lvl 16-110 version), but that's the id I get. I originally translated it to 2643 within focus.js, but I decided it would probably be a better practice to simply leave the id as is and just provide the data for that id in SPELLS_HUNTER.js

src/common/SPELLS_HUNTER.js
leapis: Same as above, just the ID. same spell (almost), but the data that I'm parsing isn't the data that's in HUNTER_SPELLS

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
leapis: I'll make a note of it, and I'll make sure that I update it to chartJS sometime very soon.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.reportCode !== this.props.reportCode || this.props.focusMax !== this.props.focusMax || this.props.pWTracker !== this.props.pWTracker || this.props.aWTracker !== this.props.aWTracker || newProps.actorId !== this.props.actorId || newProps.start !== this.props.start || newProps.end !== this.props.end) {
+      this.load(newProps.reportCode, newProps.actorId, newProps.start, newProps.end);
+    }
+  }
+  load(reportCode, actorId, start, end) {
+
+ 
+    const focusPromise = fetch(makeWclUrl(`report/tables/resources/${reportCode}`, {
```
MartijnHols: This should already be available in the events under the classResources property. Caching this in an array should provide you with the same information without the need for an additional API call.

The events we fetch are a complete set with all data that's available from the combatlog (and thus also the only thing that WCL uses). The only thing we don't receive are events not directly related to the selected player. Any additional API calls for information about just the selected player should therefore be unnecessary. 

In this case I noticed you're also using the deaths property of the response, but that should be the same on the bossHealthPromise response and should be able to be used instead.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+  render() {
+        if (!this.state.focusData || !this.state.bossHealth || !this.props.passiveWaste) {
+      return (
+        <div>
+          Loading...
+        </div>
+      );
+    }
+
+	const actorId = this.props.actorId;
```
MartijnHols: You're mixing tabs and spaces, this makes the code look poorly indented and harder to follow in PRs. Please configure your IDE to use 2 spaces for indentation for this project.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+  	const maxFocus = this.props.focusMax;
```
MartijnHols: This block of code looks pretty grouped, could it be extracted to its own method? Would make it easier to understand (and later maintain).

I'm thinking without fully understanding what's happening here that this might be better suited for a specific module.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+		 focusBySecond[secIntoFight] = item[1];
+	  }
+	  else if (focusBySecond[secIntoFight] && item[1] > magicGraphNumber && focusBySecond[secIntoFight] < item[1]){ //aims to get lowest valley
+		 focusBySecond[secIntoFight] = item[1];
+	  }
+	  else if (!focusBySecond[secIntoFight]){
+		  focusBySecond[secIntoFight] = item[1];
+	  }
+	  lastCatch = Math.floor((item[0] - start) / 1000);
+	  if(item[1] > maxFocus - 1){
```
MartijnHols: empty statement

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+	/* no, you aren't seeing double- this does the same thing as above, but aims to maximize range where-ever possible.
+	This is the graph data- since the graph is just a general point of reference, and I only record 1 data point for
+	every second, this ensures that the range of each section of the graph is accurate, at the cost of exact slope*/
+	lastCatch = 0;
+	const overCapBySecond = {};
+  const focusBySecond = [];
+  const magicGraphNumber = Math.floor(maxFocus / 2);
+	this.state.focusData.series[0].data.forEach((item) => {
+	  const secIntoFight = Math.floor((item[0] - start) / 1000);
+	  if (focusBySecond[secIntoFight] && item[1] <= magicGraphNumber && focusBySecond[secIntoFight] > item[1]){ //aims to get highest peak
```
MartijnHols: Math.max might be able to simplify this

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+		 focusBySecond[secIntoFight] = item[1];
+	  }
+	  else if (focusBySecond[secIntoFight] && item[1] > magicGraphNumber && focusBySecond[secIntoFight] < item[1]){ //aims to get lowest valley
```
MartijnHols: Math.min might be able to simplify this

src/Parser/MarksmanshipHunter/Modules/FocusChart/MaxFocus.js
```diff
@@ -0,0 +1,11 @@
+import Module from 'Parser/Core/Module';
+
+class MaxFocus extends Module {
+  on_byPlayer_cast(event){
+    if (event.sourceID === this.owner.player.id && (event.type === 'cast' || event.type === 'energize')){
+      this._focus = event.classResources[0]['max'];
```
MartijnHols: I recommend looping over class resources to prevent this breaking for whatever reason in the future. A simple boss mechanic that adds a second class resource, some item, legendary set bonus or anything that messes with class resources could break this.

```jsx
event.classResources.forEach(classResource => {
  if (classResource.type === RESOURCE_TYPES.FOCUS) {
    if (classResource.max > this.max) {
      this.max = classResource.max;
    }
  }
});
```

src/Parser/MarksmanshipHunter/Modules/FocusChart/MaxFocus.js
```diff
@@ -0,0 +1,11 @@
+import Module from 'Parser/Core/Module';
+
+class MaxFocus extends Module {
+  on_byPlayer_cast(event){
+    if (event.sourceID === this.owner.player.id && (event.type === 'cast' || event.type === 'energize')){
```
MartijnHols: event.type can never be `enegerize` here since `on_byPlayer_cast` only gets `cast` events.

src/Parser/MarksmanshipHunter/Modules/FocusChart/MaxFocus.js
```diff
@@ -0,0 +1,11 @@
+import Module from 'Parser/Core/Module';
+
+class MaxFocus extends Module {
```
MartijnHols: I think you should be able to combine this and the `ActiveFocusWasted` into 1 module that tracks focus.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+      focusBySecond[i] = focusBySecond[i] !== undefined ? focusBySecond[i] : null;
+      overCapBySecond[i] = overCapBySecond[i] !== undefined ? overCapBySecond[i] : null;
+      bosses.forEach((series) => {
+        series.data[i] = series.data[i] !== undefined ? series.data[i] : null;
+      });
+      deathsBySecond[i] = deathsBySecond[i] !== undefined ? deathsBySecond[i] : undefined;
+    }
+	const wastedFocus = Math.round(passiveCap * focusGen);
+	const totalFocus = Math.floor(fightDurationSec * focusGen);
+	let passiveRating = "";
```
MartijnHols: Name doesn't seem to reflect what it is.

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
```diff
@@ -0,0 +1,353 @@
...
+	if ( passiveCap / totalFocus > passiveWasteThresholdPercentage){
+		passiveRating = "CAN BE IMPROVED";
```
MartijnHols: I think "Can be improved." would be more friendly

src/Parser/MarksmanshipHunter/Modules/FocusChart/PassiveFocusWasted.js
```diff
@@ -0,0 +1,25 @@
...
+	tracker = 0; //to tell if cappedTimer prop has updated, since we can't compare arrays
+
+  on_byPlayer_cast(event){
+		if (event.sourceID === this.owner.player.id &&(event.type === 'cast' || event.type === 'energize') ){
+			this.tracker++;
+			if(!this.cappedTimer){
+			}
+			const secIntoFight = (event.timestamp - this.owner.fight.start_time);
+			if(!this.cappedTimer[secIntoFight]){
+				this.cappedTimer[secIntoFight] = event.classResources[0]['amount'];
```
MartijnHols: see my previous classResource comment.

iirc the `amount` is the focus before the spell's cost from the cast event was applied, so you generally want to use `classResource.amount - classResource.cost`.

src/Parser/MarksmanshipHunter/Modules/FocusChart/PassiveFocusWasted.js
```diff
@@ -0,0 +1,25 @@
+import Module from 'Parser/Core/Module';
+
+class PassiveFocusWasted extends Module {
+
+	cappedTimer = [];
+	tracker = 0; //to tell if cappedTimer prop has updated, since we can't compare arrays
+
+  on_byPlayer_cast(event){
+		if (event.sourceID === this.owner.player.id &&(event.type === 'cast' || event.type === 'energize') ){
```
MartijnHols: Doesn't receive any `energize` events (see previous comment)

src/Parser/MarksmanshipHunter/Modules/FocusChart/Focus.js
leapis: sorry, I ended up changing IDEs about half way through, since my PC crashed and I was forced to move to my laptop T_T I'll get it fixed

leapis:
Not at all discouraged, it's a good learning experience. I originally thought this might be one of the easier things to do, but that was before I learned the nature of the the WCL API. That being said, the way I reorganized my modules should lend the focus tracker use for many future modules, so I do think it's much improved. 

Changes made:

- combined every focus tracking module (active, passive, max focus, focus by second, etc) into FocusTracker.js
- removed all WCL API calls (now I just use the event listeners [I took out boss health and player deaths]- I talked to Putro, I have a branch with boss health/player deaths, but that isn't needed with the current state of the MMS analyzer)
- migrated to Chart.js from Chartist
- changed indents (I think I got all of them... I'll keep looking though)
- implemented the various minor changes you requested

Screenshot:

image

src/Parser/MarksmanshipHunter/Modules/FocusChart/FocusTracker.js
```diff
@@ -0,0 +1,139 @@
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import RESOURCE_TYPES from 'common/RESOURCE_TYPES';
+
+
+class FocusTracker extends Module {
```
MartijnHols: This module still has very jumpy indents.

src/common/SPELLS_HUNTER.js
MartijnHols: The naming for keys for spells should be the exact (English) spell name, all caps snake case. If a name is already used, both names should be changed to indicate what they are. E.g. often there's a SPELL_CAST and a SPELL_DAMAGE, SPELL_BUFF, SPELL_HEAL and/or SPELL_TRAIT. Rarely it's more complex than that like ATONEMENT_HEAL_NON_CRIT, ATONEMENT_HEAL_CRIT, ATONEMENT_BUFF. Blizzard usually doesn't create duplicate spells for no reason, so finding out why it's there and what it's for - and giving it a name that makes it clear - will usually help you reason about it and apply them properly.

MartijnHols:
Looking good, thanks a lot for taking the time to make this and process the feedback!

PR #457, opened 2017-10-04
[Marksmanship Hunter] Aimed inside vulnerable tracker

Pewtro:
This tracks the amount of aimed shots that the player shoots inside and outside of vulnerable (our version of Colossus Smash basically), which is where we get the most of our damage from and this is one of the largest points of improvement for many players, so this visual representation of how many they're messing up should be helpful.


src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
```diff
@@ -0,0 +1,58 @@
...
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.AIMED_SHOT.id) {
+      return;
+    }
+    const enemy = this.enemies.getEntity(event);
+    if (enemy.hasBuff(SPELLS.VULNERABLE.id, event.timestamp)) {
+      this.inVulnAimed += 1;
+    }
+    if (!enemy.hasBuff(SPELLS.VULNERABLE.id, event.timestamp)) {
```
sMteX: Should be replaced with just `else {`.

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
```diff
@@ -0,0 +1,58 @@
...
+      this.outsideVulnAimed += 1;
+    }
+    this.totalAimed += 1;
+  }
+  suggestions(when) {
+    const aimedOutsideVuln = this.outsideVulnAimed;
+    when(aimedOutsideVuln).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span> You have {this.outsideVulnAimed} <SpellLink id={SPELLS.AIMED_SHOT.id} />s outside <SpellLink id={SPELLS.VULNERABLE.id} />. Try and minimize these, as they deal significantly less damage than their vulnerable counterparts. It should be noted that rarely, you will be casting non-vulnerable aimeds due to no procs and/or focus capping. </span>)
```
sMteX: 1) I'd rephrase the beginning as "You have casted X Aimed Shots ..."
2) Considering you mentioned it's possible that a player would have to cast some AS without Vulnerable, wouldn't you think it might be better to reconsider the thresholds for the suggestions? Maybe make them more tolerant? Or divide by fight duration, make them "AS outside Vulnerable *per minute*" and use that?

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
```diff
@@ -0,0 +1,58 @@
...
+          .regular(recommended + 1).major(recommended + 2);
+      });
+  }
+  statistic() {
+    const percentVulnInAimed = this.inVulnAimed / this.totalAimed;
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.VULNERABLE.id} />}
+        value={`${formatPercentage(percentVulnInAimed)}%`}
+        label="Vulnerable aimed shots"
```
sMteX: Should use capitalized spell name, `Aimed Shots`, looks better O:)

src/Parser/MarksmanshipHunter/CombatLogParser.js
```diff
@@ -24,12 +27,14 @@ class CombatLogParser extends CoreCombatLogParser {
     alwaysBeCasting: AlwaysBeCasting,
     castEfficiency: CastEfficiency,
     cooldownTracker: CooldownTracker,
+    vulnerableTracker: VulnerableTracker,
     vulnerabluptime: VulnerableUpTime,
```
sMteX: I must've missed that earlier, should be `vulnerableUptime`.

Pewtro: It's a module made by JLassie, but I am considering completely scratching it, it is currently a somewhat useless metric -- just feels rude, but I'll update that name for now.

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
Pewtro: I was going back and forth on this specific thing, glad someone else is of the same opinion on this 👍

src/Parser/MarksmanshipHunter/CombatLogParser.js
sMteX: Usefulness of the module is not up to me to judge, I just wanted to point out the name :) So `vulnerableUptime: VulnerableUpTime,` just to be clear which one I meant :)

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
Pewtro: It should really generally always be 100%, because it should be quite rare that you are forced to cast them outside, but I might've been a bit too strict when I made it, so it's a bit more lenient now, to allow for up to 2% of total aimeds to hit outside before it triggers the suggestion

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
```diff
@@ -15,41 +15,41 @@ class AimedInVulnerableTracker extends Module {
...
-      this.outsideVulnAimed += 1;
+    else {
+      this.outsideVulnerabilityAimed += 1;
     }
     this.totalAimed += 1;
   }
   suggestions(when) {
-    const aimedOutsideVuln = this.outsideVulnAimed;
-    when(aimedOutsideVuln).isGreaterThan(0)
+    const percentAimedOutsideVulnerable = 100/this.totalAimed * this.outsideVulnerabilityAimed;
```
sMteX: Some weird arithmetic here, if you rearrange the terms you get `(this.outsideVulnerabilityAimed / this.totalAimed) * 100` which would be a percentage between 0 - 100% which means you gave it tolerance of 0,02% instead of 2%. Furthermore you divide or multiply by 100 further down the code. I suggest you change this to just `const percentAimedOutsideVulnerable = this.outsideVulnerableAimed / this.totalAimed` and remove the `* 100` and `/ 100` further down. Nice jobs on the added SpellLink :)

src/Parser/MarksmanshipHunter/CombatLogParser.js
```diff
@@ -28,7 +28,7 @@ class CombatLogParser extends CoreCombatLogParser {
     castEfficiency: CastEfficiency,
     cooldownTracker: CooldownTracker,
     vulnerableTracker: VulnerableTracker,
-    vulnerabluptime: VulnerableUpTime,
+    vulnerablUptime: VulnerableUptime,
```
sMteX: Still missing an "e" here :D

Pewtro: 🙃 I'm officially blind - fixed

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
sMteX: I see you've updated the suggestion to be accurate but I still think using the decimal representation of percentage (e.g. 0.02 instead of 2) is more consistent with the rest of the code and allows for using `formatPercentage`.

src/Parser/MarksmanshipHunter/Modules/Features/AimedInVulnerableTracker.js
```diff
@@ -32,14 +32,14 @@ class AimedInVulnerableTracker extends Module {
     this.totalAimed += 1;
   }
   suggestions(when) {
-    const percentAimedOutsideVulnerable = 100/this.totalAimed * this.outsideVulnerabilityAimed;
-    when(percentAimedOutsideVulnerable).isGreaterThan(2)
+    const percentAimedOutsideVulnerable =(100/this.totalAimed * this.outsideVulnerabilityAimed)/100;
```
sMteX: Sorry to be nitpicky but the 100's cancel out so we're left with `this.outsideVulnerabilityAimed / this.totalAimed` anyway, please update it for readability's sake and I think I can merge that :)

sMteX:
Thanks!

PR #458, opened 2017-10-04
[Guardian] Filler spell spam detection

FaideWW:
This module detects when a player is spamming filler spells like Swipe and Moonfire instead of casting rage-generating, hard-hitting abilities like Mangle and Thrash.

There are several additional modules that were created to support filler spam detection:

`DynamicHaste` - tracks hasted cooldown progress through changing haste values (gaining and losing haste buffs).
`ActiveTargets` - estimates the number of targets that are active at a given time.

If there's enough interest I can generalize these and make them available in Core.

This PR also introduces Mangle cast efficiency, and fixes a bug with Thrash cast efficiency.


src/Parser/GuardianDruid/Modules/Features/AntiFillerSpam.js
```diff
@@ -0,0 +1,199 @@
...
+          .actual(`${formatPercentage(actual)}% unnecessary filler spells cast`)
+          .recommended(`${formatPercentage(recommended, 0)}% or less is recommended`)
+          .regular(recommended + 0.05).major(recommended + 0.1);
+      });
+  }
+}
+
+export default AntiFillerSpam;
+
+const GCD_SPELLS = {
```
MartijnHols: please define this prior to using it, better would be as a static property of the class so children can override it.

MartijnHols: I think another improvement would be to move this to the CastEfficiency abilities property.

src/Parser/GuardianDruid/Modules/Features/AntiFillerSpam.js
```diff
@@ -0,0 +1,199 @@
...
+      if (baseCD !== null) {
+        if (hastedCD) {
+          isOffCooldown = this.dynamicHaste.isHastedAbilityOffCooldown(lastCast, timestamp, baseCD);
+        } else {
+          isOffCooldown = timestamp - lastCast > baseCD;
+        }
+      }
+
+      const hasProc = proc && ((typeof proc === 'function') ? proc(event, combatant, targets, lastCast) : proc);
+      const meetsCondition = (condition !== undefined) ? ((typeof condition === 'function') ? condition(event, combatant, targets, lastCast) : condition) : true;
```
MartijnHols: I think it would be much easier to follow if this was written this with short if statements and return early when a required one fails.

src/Parser/GuardianDruid/Modules/Features/AntiFillerSpam.js
```diff
@@ -0,0 +1,199 @@
...
+    isFiller: false,
+    baseCD: null,
+    // Maul should never be considered a replacement for filler, but it should be tracked
+    condition: false,
+  },
+
+  // "Filler" spells
+  [SPELLS.MOONFIRE.id]: {
+    isFiller: ({ timestamp, targetID }, combatant, targets, lastCast) => {
+      if (combatant.lv75Talent === SPELLS.GALACTIC_GUARDIAN_TALENT.id) {
```
MartijnHols: `combatant.hasTalent(spellId)` is preferred

src/Parser/GuardianDruid/Modules/Features/AntiFillerSpam.js
```diff
@@ -0,0 +1,199 @@
...
+      const isIncarnation = combatant.hasBuff(SPELLS.INCARNATION_OF_URSOC.id, timestamp);
+      return isIncarnation;
+    },
+  },
+  [SPELLS.PULVERIZE_TALENT.id]: {
+    isFiller: false,
+    baseCD: null,
+    // A spell must meet these conditions to be castable
+    condition: ({ timestamp, targetID }, combatant, targets) => {
+      const pulverizeTalented = combatant.lv100Talent === SPELLS.PULVERIZE_TALENT.id;
```
MartijnHols: `combatant.hasTalent(spellId)` is preferred

src/Parser/GuardianDruid/Modules/Features/DynamicHaste.js
```diff
@@ -0,0 +1,85 @@
...
+      this._hasteLog[lastHasteIndex].haste = newHaste;
+    }
+
+    if (lastHaste.haste !== newHaste) {
+      // Conversely, only record a haste change if the values are actually different
+      this._hasteLog.push({ timestamp, haste: newHaste });
+    }
+  }
+
+  isHastedAbilityOffCooldown(lastCast, timestamp, baseCD) {
```
MartijnHols: I'll soon finish and release the new cooldown module, I reckon it will be a little bit simpler to use. I think the only main thing left of the module is writing tests (and refactoring and fixing issues I run into), so you can already see it in the branch I linked earlier. Making it work with "hasted cooldowns" should be as easy as adjusting the cooldown on every `changehaste` event (I might ask you about how this works by the time I get to that). I hope you won't mind later migrating this to avoid having two modules doing the same.

src/Parser/GuardianDruid/Modules/Features/ActiveTargets.js
```diff
@@ -0,0 +1,72 @@
...
+import { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';
+
+/*
+ * The amount of time elapsed without a combat event before a target is considered inactive.
+ * The ideal size of this window will vary; specs that produce a lot of combat events can
+ * have a lower threshold value.
+ */
+const ACTIVITY_THRESHOLD = 3000;
+
+class ActiveTargets extends Module {
```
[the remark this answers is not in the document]
FaideWW: Sure! I'll put that on my list of TODOs.

src/Parser/GuardianDruid/Modules/Features/DynamicHaste.js
FaideWW: Yeah that's not a problem, and I'd be happy to help.

Basically the gist of it is instead of reducing the cooldown directly, I accelerate the time passed based on the current haste value, and I do it iteratively until the the cooldown ends or we reach the current time.  

I found this was slightly easier to reason about since I didn't need to hold a list of all spells and whether or not they were on cooldown, just a timeline of the haste buffs gained and lost.

src/Parser/GuardianDruid/Modules/Features/ActiveTargets.js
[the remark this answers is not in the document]
FaideWW: I think it's out of scope for this feature, I'll open a separate PR for it.

src/Parser/GuardianDruid/Modules/Features/DynamicHaste.js
MartijnHols: Yeah it's definitely a much easier way to do what you had to do. My approach is a lot more complicated at the base, but I think once it's done it should be pretty easy to change its behavior via the methods provided.

src/Parser/GuardianDruid/Modules/Features/CastEfficiency.js
```diff
@@ -7,35 +7,94 @@ import CoreCastEfficiency from 'Parser/Core/Modules/CastEfficiency';
...
+
+        const gainedMangles = potentialGoreProcs * goreChance;
+        const totalMangles = manglesOnCD + gainedMangles;
+        const effectiveCD = fightLengthSec / totalMangles;
+
+        return effectiveCD;
+      },
       noSuggestion: true,
+
+      isOnGCD: true,
```
MartijnHols: This is nice! We should totally use this approach for AlwaysBeCasting, but that's out of scope for this PR.

MartijnHols:
Thanks!

PR #459, opened 2017-10-05
[Marksmanship Hunter] Remove Cyclonic Burst from Cooldown view

Pewtro:
Very short PR, that simply removes Cyclonic Burst (a ground effect from Windburst) from the cooldown view. 

The changes to CombatLogParser and AimedInVulnerableTracker are already in, but simply made it into this because of my merging to finally fix what I fucked some days ago.


MartijnHols:
Thanks!

PR #460, opened 2017-10-05
[Marksmanship] Trueshot CD reduction from artifact traits

Pewtro:
This adds a trueshot module for future usage for the advanced cooldown tracking, for now it simply has rank 0-7 of Quick Shot which reduces the CD of Trueshot with diminished values pr rank after initial 3. 

From my testing of various logs, it's accurate now when Convergence, Legendary boots and/or T192p aren't involved. They are on a future to-do list, but this solves the first hurdle when it comes to the CD of Trueshot. 

Example of before: 
image

Example of after: 
image

### DIFF
--- a/src/Parser/MarksmanshipHunter/Modules/Features/CastEfficiency.js
+++ b/src/Parser/MarksmanshipHunter/Modules/Features/CastEfficiency.js
@@ -2,9 +2,7 @@ import React from 'react';
 
 import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
-//import QuickShot from 'Parser/MarksmanshipHunter/Modules/Traits/QuickShot';
 import CoreCastEfficiency from 'Parser/Core/Modules/CastEfficiency';
-
 
 /* eslint-disable no-unused-vars */
 
@@ -58,7 +56,7 @@ class CastEfficiency extends CoreCastEfficiency {
       getCooldown: haste => 60,
       isActive: combatant => combatant.hasTalent(SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id),
       recommendedCastEfficiency: 1.0,
-      extraSuggestion: <span> The only time <SpellLink id={SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id}/> should be delayed is when the boss is under 25% hp, as you will then use it to generate <SpellLink id={SPELLS.BULLSEYE_BUFF.id}/> stacks as early on, and as often, as possible. </span>,
+      extraSuggestion: <span> The only time <SpellLink id={SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id} /> should be delayed is when the boss is under 25% hp, as you will then use it to generate <SpellLink id={SPELLS.BULLSEYE_BUFF.id} /> stacks as early on, and as often, as possible. </span>,
     },
     {
       spell: SPELLS.BARRAGE_TALENT,
@@ -83,9 +81,8 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.TRUESHOT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: (haste, combatant) => {
-        const availableCasts = combatant.owner.modules.trueshot.tsAvailableCasts;
-        return (combatant.owner.fightDuration / 1000) / availableCasts;
+      getCooldown: (_, combatant) => {
+        return combatant.owner.modules.trueshot.reducedCooldownWithTraits;
       },
       recommendedCastEfficiency: 1.0,
     },
--- a/src/Parser/MarksmanshipHunter/Modules/Features/CastEfficiency.js
+++ b/src/Parser/MarksmanshipHunter/Modules/Features/CastEfficiency.js
@@ -83,7 +83,11 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.TRUESHOT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: haste => 180, // TODO calculate cd reduction based on artifact
+     // getCooldown: haste => 180, // TODO calculate cd reduction based on artifact
+      getCooldown: (haste, combatant) => {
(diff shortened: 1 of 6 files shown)


src/Parser/MarksmanshipHunter/Modules/Spells/Trueshot.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    else if (quickShotRank === 6) {
+      this.trueShotCDReduction = 52;
+    }
+    else if (quickShotRank === 7) {
+      this.trueShotCDReduction = 58;
+    }
+    const cooldownWithTraits = 180 - this.trueShotCDReduction;
+    const tsAvailableCasts = (this.owner.fightDuration / 1000) / cooldownWithTraits;
+    return tsAvailableCasts;
```
sMteX: Now I'm not sure if it wouldn't mess up with the computation in CastEfficiency, but this method should probably get the amount of available casts as an integer, not return something like "1.43627Z6 available casts". Might want to round up or down (`Math.ceil() or Math.floor()`) before returning.

src/Parser/MarksmanshipHunter/Modules/Spells/Trueshot.js
```diff
@@ -0,0 +1,40 @@
...
+
+  trueShotCDReduction = 0;
+  totalFromCD = 0;
+
+  get tsAvailableCasts() {
+    const player = this.combatants.selected;
+    const quickShotRank = player.traitsBySpellId[SPELLS.QUICK_SHOT_TRAIT.id];
+
+    if (quickShotRank === 0) {
+      return;
```
sMteX: Returning here would return I think `undefined` which would mess up with the calculation in CastEfficiency. I understand that there's most likely a 0 probability of someone without this trait would be using the analyzer but it still should be accounted for (my suggestion below should solve this).

src/Parser/MarksmanshipHunter/Modules/Spells/Trueshot.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+    else if (quickShotRank === 5) {
+      this.trueShotCDReduction = 45;
+    }
+    else if (quickShotRank === 6) {
+      this.trueShotCDReduction = 52;
+    }
+    else if (quickShotRank === 7) {
+      this.trueShotCDReduction = 58;
+    }
```
sMteX: It's strange the CDR goes first by 10% and then decreases (which messes up the computations) but I suggest something like this instead:
```jsx
if (quickShotRank < 4) {
   this.trueShotCDReduction = quickShotRank * 10;
}
else {
   switch (quickShotRank) {
      case 4:
         this.trueShotCDReduction = 38;
         break;
      case 5:
         this.trueShotCDReduction = 45;
         break;
      case 6:
         this.trueShotCDReduction = 52;
         break;
      case 7:
         this.trueShotCDReduction = 58;
         break;
      default:
         break;
   }
}
```

src/Parser/MarksmanshipHunter/Modules/Spells/Trueshot.js
kfinch: Imo best to do all rounding when you display, and to leave values unfounded during calculations so you don't lose any precision you don't need to lose

sMteX: That's a fair point and I'd do that usually too although in this case ("get available casts") it does make sense to round the result I think. I don't know, maybe it doesn't really make a difference.

src/Parser/MarksmanshipHunter/Modules/Spells/Trueshot.js
Pewtro: oh yeah, I completely forgot about switch-cases.. derp, way better solution

kfinch: tbh I prefer the if/else 😂

Pewtro:
I'm sure there is a better way to do that code, but if I remove one of those calculations it returns that a 2m50s kill (with no additional CDR) could have used a 2m22s cooldown 142 times in the 2m51s fight.. I'll brood upon how to do it better, before submitting a new push to this. 

In regards to the future, it's a module where I intend to implement some more cooldown reduction in the future, once I get a good idea on how to handle the legendaries/tier/trinket properly, hence not creating anything in the CastEfficiency, so as to have everything in one place for when those CDR handlers are to be implemented. 

Might be better to call the class TrueshotCDR or something along those lines, I guess - if anything just for clarity.

sMteX:
Actually when I did the math right now (let's call `this.owner.fightDuration` just `duration`, `cooldownWithTraits` just `CD`).

When you rearrange `(duration / 1000) / CD` you get `duration / (CD * 1000)`. Then you get that value and divide the duration again with it in CastEfficiency - `(duration / 1000) / (duration / (CD * 1000))`. When you rearrange again (outer terms of the fraction divided by inner terms) you get `(duration * CD * 1000) / (duration * 1000)`. 

`duration` and `1000` cancel out and in the end you're left with just `duration`. It needs some testing but it _should_ be enough to return just `cooldownWithTraits` from the getter and in CastEfficiency just do 
```jsx
getCooldown: (haste, combatant) => combatant.owner.modules.trueshot.tsAvailableCasts,
```
(which could be by all means renamed to `getReducedCooldown` or something like that).

src/Parser/MarksmanshipHunter/Modules/Features/CastEfficiency.js
```diff
@@ -83,9 +81,8 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.TRUESHOT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: (haste, combatant) => {
-        const availableCasts = combatant.owner.modules.trueshot.tsAvailableCasts;
-        return (combatant.owner.fightDuration / 1000) / availableCasts;
+      getCooldown: (_, combatant) => {
+        return combatant.owner.modules.trueshot.reducedCooldownWithTraits;
```
sMteX: Thanks for the change - did you check it gives the same results? To the comment - if the lambda is just one return statement it can be shortened to 
```jsx
getCooldown: (_, combatant) => combatant.owner.modules.trueshot.reducedCooldownWithTraits,
```

MartijnHols:
Thanks!

PR #462, opened 2017-10-06
[resto druid] Added ironbark module

kfinch:
Added Ironbark module using a LazyLoaded stats box because an extra query is required. The stats generated here could be used to generate the NYI ArmorOfTheAncients trait, but I don't want to implement lazy loading in the Relic Trait box at the moment.


MartijnHols:
Thanks! 👍

PR #463, opened 2017-10-06
[Guardian] Improved Ironfur module

FaideWW:
Adds support for tracking individual stacks of Ironfur, and updates the statistic and suggestion accordingly.


MartijnHols:
Before I do a review, please fix the unit tests so the build succeeds.

MartijnHols:
Thanks!

PR #464, opened 2017-10-07
[Retribution] Added in a relic traits stat box

CollCrom:
The Wrath of the Ashbringer trait is a like hacky but I couldn't think of a better way to do it. It gets the average DPS of wings over the course of the fight then calculates how much damage 2 extra seconds of wings would be. I also couldn't figure out how to implement http://www.wowhead.com/spell=184759/sharpened-edge and am looking for suggestions on that. Otherwise should be good. Please let me know if there is anything you think I should change!

### DIFF
--- a/src/Main/App.js
+++ b/src/Main/App.js
@@ -97,6 +97,7 @@ class App extends Component {
       progress: 0,
       dataVersion: 0,
       bossId: null,
+      config: null,
     };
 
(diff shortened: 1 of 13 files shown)


MartijnHols:
There appears to be an issue with the build, check the Travis log to see it and please repair it in your PR.

CollCrom:
@MartijnHols Okay I passed the travis check. Please review when you have some free time! thanks

MartijnHols:
Thanks!

PR #466, opened 2017-10-08
Add spell usable tracking (whether a spell is on cooldown)

MartijnHols:
Based on spells as configured in CastEfficiency. Added a full test suite based on all requirements I could think of. 

Please review so this can be merged.


src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -0,0 +1,247 @@
...
+   * Whether the spell can be cast. This is not the opposite of `isOnCooldown`! A spell with 2 charges, 1 available and 1 on cooldown would be both available and on cooldown at the same time.
+   */
+  isAvailable(spellId) {
+    if (this.isOnCooldown(spellId)) {
+      const maxCharges = this.castEfficiency.getMaxCharges(spellId);
+      if (maxCharges > this._currentCooldowns[spellId].chargesOnCooldown) {
+        return true;
+      }
+      return false;
+    } else {
```
FaideWW: This else can be removed.

src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -0,0 +1,247 @@
...
+  }
+  /**
+   * Returns the amount of time remaining on the cooldown.
+   * @param {number} spellId
+   * @param {number} timestamp Override the timestamp if it may be different from the current timestamp.
+   * @returns {number|null} Returns null if the spell isn't on cooldown.
+   */
+  cooldownRemaining(spellId, timestamp = this.owner.currentTimestamp) {
+    if (!this.isOnCooldown(spellId)) {
+      return null;
```
FaideWW: What is the reason this is null rather than 0? Return type consistency would be nice here.

src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+  // Haste-based cooldowns gets longer/shorter when your Haste changes
+  // `newDuration = timePassed + (newCooldownDuration * (1 - progress))` (where `timePassed` is the time since the spell went on cooldown, `newCooldownDuration` is the full cooldown duration based on the new Haste, `progress` is the percentage of progress cooling down the spell)
+  on_changehaste(event) {
+    Object.keys(this._currentCooldowns).forEach(spellId => {
+      const cooldown = this._currentCooldowns[spellId];
+      const originalExpectedDuration = cooldown.expectedDuration;
+      const remainingDuration = this.cooldownRemaining(spellId, event.timestamp);
+      const timePassed = originalExpectedDuration - remainingDuration;
```
FaideWW: `timePassed` can be simplified to `event.timestamp - cooldown.start`, which would remove the need to call `this.cooldownRemaining`.

MartijnHols: I prefer the way since it closely matches what I want and therefore is less likely to break. I'm pretty sure calling `cooldownRemaining` doesn't really cost anything noticeable, so maintainability is the major point here.

src/Parser/Core/Modules/SpellUsable.js
MartijnHols: I prefer it this way, since it's a taste thing I'll leave it as it is 😜

src/Parser/Core/Modules/SpellUsable.js
MartijnHols: I want consumers of the function to take it being off cooldown into account when they're calling this method so they're less likely to create mistakes. This should probably throw instead.

PR #468, opened 2017-10-08
Frost Mage Support

Sharrq:
Added Frost Mage Support and a group of modules to start with. Still working on adding stuff, but this should be a good start.


src/Parser/FrostMage/Modules/Features/CooldownTracker.js
```diff
@@ -0,0 +1,133 @@
...
+        BUILT_IN_SUMMARY_TYPES.DAMAGE,
+      ],
+    },
+    {
+      spell: SPELLS.DRUMS_OF_THE_MOUNTAIN,
+      duration: 40,
+      summary: [
+        BUILT_IN_SUMMARY_TYPES.DAMAGE,
+      ],
+    },
```
sMteX: Bloodlust-type effects shouldn't be here. `castCooldowns` is only for cooldowns that doesn't place a buff on the player.

src/Parser/FrostMage/Modules/Features/FingersFrost.js
```diff
@@ -0,0 +1,62 @@
...
+class FingersFrostTracker extends Module {
+
+	overwrittenFingersFrostProcs = 0;
+	totalFingersFrostProcs = 0;
+
+	static dependencies = {
+		combatants: Combatants,
+	}
+
+	on_initialized(){}
```
sMteX: Empty function.

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
```diff
@@ -0,0 +1,50 @@
...
+class IcicleTracker extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  IciclesWasted = 0;
+
+  on_initialized() {
+	const hasGlacialSpike = this.combatants.selected.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id);
+	this.active = hasGlacialSpike;
```
sMteX: This can be shortened to `this.active = this.combatants.selected.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id);`

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+  
+  on_refreshbuff(event) {
```
sMteX: You should probably use `on_toPlayer_refreshbuff(event)`

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+  
+  suggestions(when) {
+    when(this.IciclesWasted).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span> You wasted {this.IciclesWasted} <SpellLink id={SPELLS.ICICLES.id}/>.  Try to cast <SpellLink id={SPELLS.GLACIAL_SPIKE.id}/> once you get to 5 Icicles to avoid wasting them.</span>)
+          .icon(SPELLS.ICICLES.icon)
+          .actual(`${formatNumber(actual)} Icicles Wasted`)
+          .recommended(`${formatNumber(recommended)} is recommended`)
+          .regular(3).major(10);
```
sMteX: It is recommended to use values relative to the `recommended`, so that if you decide to change the values they update as well.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
...
+import Module from 'Parser/Core/Module';
+
+class IcyVeinsDuration extends Module {
+  static dependencies = {
+		combatants: Combatants,
+	}
+	
+  on_initialized() {
+	const hasThermalVoid = this.combatants.selected.hasTalent(SPELLS.THERMAL_VOID_TALENT.id);
+	this.active = hasThermalVoid;
```
sMteX: Can be shortened to `this.active = this.combatants.selected.hasTalent(SPELLS.THERMAL_VOID_TALENT.id);`

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  IcyVeinsCasts = 0;
```
sMteX: Variables should be initiated in the beginning of the class, preferably under `dependencies`.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
...
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.ICY_VEINS.id) {
+      return;
+    }
+    this.IcyVeinsCasts += 1;
+  }
+
+  suggestions(when) {
+    const IcyVeinsUptime = (this.combatants.selected.getBuffUptime(SPELLS.ICY_VEINS.id) / 1000) / this.IcyVeinsCasts;
```
sMteX: I'd say "uptime" isn't accurate depiction of what is calculated here. "averageDuration" (since we're talking about Icy Veins it's unnecessary to repeat) would be better.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
...
+        return suggest(<span>Your <SpellLink id={SPELLS.ICY_VEINS.id}/> uptime can be improved. Make sure you use Frozen Orb to get Fingers of Frost Procs</span>)
+          .icon(SPELLS.ICY_VEINS.icon)
+          .actual(`${formatNumber(actual)} seconds Average Icy Veins Uptime`)
+          .recommended(`${formatNumber(recommended)} is recommended`)
+          .regular(40).major(30);
+      });
+  }
+
+  statistic() {
+    const IcyVeinsUptime = (this.combatants.getBuffUptime(SPELLS.ICY_VEINS.id) / 1000) / this.IcyVeinsCasts;
```
sMteX: See previous comment.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.ICY_VEINS.id} />}
+        value={`${formatNumber(IcyVeinsUptime)}s`}
```
sMteX: Take a look at `formatDuration` in "common/format" if that isn't more fitting.

src/Parser/FrostMage/Modules/Features/ThermalVoidDuration.js
```diff
@@ -0,0 +1,57 @@
...
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+
+class ThermalVoidDuration extends Module {
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.THERMAL_VOID_TALENT.id);
+  }
+
+  suggestions(when) {
```
sMteX: This whole suggestion is copied from Shadow Priest and wasn't changed.

src/Parser/FrostMage/Modules/Features/ThermalVoidDuration.js
```diff
@@ -0,0 +1,57 @@
...
+    <SpellLink id={SPELLS.DISPERSION.id} /> can be used to synchronize your cooldowns back in order or in case of an emergency if you are about to fall out of <SpellLink id={SPELLS.VOIDFORM.id} /> and you have a <SpellLink id={SPELLS.MINDBENDER_TALENT_SHADOW.id} /> active.
+				</span>)
+			.icon(SPELLS.VOIDFORM_BUFF.icon)
+			.actual(`${formatNumber(actual)} average Voidform stacks.`)
+			.recommended(`>${formatNumber(recommended)} stacks is recommended.`)
+			.regular(recommended).major(recommended - 5);
+});
+  }
+
+  statistic() {
```
sMteX: This was copied from Voidform too.

src/Parser/FrostMage/Modules/Features/WintersChill.js
```diff
@@ -0,0 +1,59 @@
...
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.ICE_LANCE_CAST.id) {
+      return;
+    }
+    const enemy = this.enemies.getEntity(event);
+    if (enemy.hasBuff(SPELLS.WINTERS_CHILL.id, event.timestamp)) {
+      this.inChillIceLance += 1;
+    }
+  }
+  on_applydebuff(event) {
```
sMteX: Should use `on_byPlayer_applydebuff(event)`

src/Parser/FrostMage/Modules/Features/WintersChill.js
```diff
@@ -0,0 +1,59 @@
...
+	  if(event.ability.guid !== SPELLS.WINTERS_CHILL.id) {
+		  return;
+	  }
+	  const enemy = this.enemies.getEntity(event);
+	  if(enemy.hasBuff(SPELLS.WINTERS_CHILL.id)) {
```
sMteX: This feels rather unnecessary - you just applied the Winter's Chill debuff so this condition will most likely always return true.

src/Parser/FrostMage/Modules/Items/ShardofExodar.js
```diff
@@ -0,0 +1,32 @@
...
+	};
+
+	LustDamage = 0;
+	
+  on_initialized() {
+    this.active = this.combatants.selected.hasFinger(ITEMS.SHARD_OF_EXODAR.id);
+  }
+
+  on_byPlayer_damage(event){
+	if (this.combatants.selected.hasBuff(SPELLS.HEROISM.id) || this.combatants.selected.hasBuff(SPELLS.BLOODLUST.id) || this.combatants.selected.hasBuff(SPELLS.TIME_WARP.id) || this.combatants.selected.hasBuff(SPELLS.NETHERWINDS.id) || this.combatants.selected.hasBuff(SPELLS.ANCIENT_HYSTERIA.id) || this.combatants.selected.hasBuff(SPELLS.DRUMS_OF_FURY.id) || this.combatants.selected.hasBuff(SPELLS.DRUMS_OF_RAGE.id) || this.combatants.selected.hasBuff(SPELLS.DRUMS_OF_THE_MOUNTAIN.id)) {
```
sMteX: A better solution would be to create an array with all the buffs, say
```jsx
const BLOODLUST_BUFFS = [
SPELLS.HEROISM.id,
SPELLS.BLOODLUST.id,
SPELLS.TIME_WARP.id,
....
];
```
and then in this event use `if (BLOODLUST_BUFFS.some(buff => this.combatants.selected.hasBuff(buff, event.timestamp)))`

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import SpellIcon from 'common/SpellIcon';
+import { formatPercentage, formatNumber } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Module from 'Parser/Core/Module';
+
+class FrozenOrb extends Module {
```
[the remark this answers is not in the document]
Sharrq: The FrozenOrb Module was something I was playing with to try and do cooldown reduction. I built some of it and never finished it because i figured i would wait until Zero has his cooldown tracking stuff implemented, and then i forgot to remove this file from the PR. Essentially though, the Statistic Box was Blizzard because i was testing the BlizzardHits portion to see if it was working. And Blizzard and Brain Freeze are in there because Blizzard Hits reduce the cooldown on FrozenOrb by 0.5 seconds and with T204pc, Brain Freeze Procs reduce the cooldown on Frozen Orb by 5 seconds. Either way ill remove this file

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+	
+	FrozenOrbCooldown = 60;
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+	FrozenOrbCasts = 0;
+	BlizzardHits = 0;
+	FightDuration = this.owner.fightDuration;
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+	
+	static dependencies = {
+		combatants: Combatants,
+	}
+	
+	on_initialized(){}
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+	}
+  }
+	
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.FROZEN_ORB_CAST.id) {
+      return;
+    }
+	PreviousTimestamp = event.timestamp;
+	OrbCooldown = 60;
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+	on_byPlayer_damage(event) {
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+		if(event.ability.guid === SPELLS.BLIZZARD_DEBUFF.id) {
+			this.OrbCooldown -= 0.5
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+    this.FrozenOrbCasts += 1;
+  }
+	
+	suggestions(when) {
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
```diff
@@ -0,0 +1,67 @@
...
+					.recommended(`Wasting none is recommended`)
+					.regular(recommended+ 0.1).major(recommended + 0.2);	
+			});
+	}
+	
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.BLIZZARD_DEBUFF.id} />}
+        value={`${formatNumber(this.BlizzardHits)}`}
```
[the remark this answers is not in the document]
Sharrq: I didnt mean for this file to get included in the PR since its unfinished, ill be removing it until its done.

src/Parser/FrostMage/Modules/Features/BrainFreeze.js
```diff
@@ -0,0 +1,73 @@
...
+class BrainFreezeTracker extends Module {
+
+	overwrittenBrainFreezeProcs = 0;
+	totalBrainFreezeProcs = 0;
+
+	static dependencies = {
+		combatants: Combatants,
+	}
+
+	on_initialized(){}
```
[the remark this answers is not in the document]
Sharrq: Yea, this was the first module i made, and I didnt know if on_initialized was required or not, so i just left it blank. Ill remove it.

src/Parser/FrostMage/Modules/Features/CastEfficiency.js
```diff
@@ -0,0 +1,105 @@
...
+	  noSuggestion: true,
+	  noCanBeImproved: true,
+	},
+	{
+	  spell: SPELLS.ICE_BLOCK,
+	  category: CastEfficiency.SPELL_CATEGORIES.DEFENSIVE,
+	  getCooldown: haste => 240,
+	  noSuggestion: true,
+	  noCanBeImproved: true,
+	},
```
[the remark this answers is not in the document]
Sharrq: I can do that. When i was copying one of the other CastEfficienty Modules i saw all the extra spells and they seemed like they were there just for the sake of being there ... which seemed like a waste to me ... so i didnt include the utility spells and such. But ill add them, that makes sense.

src/Parser/FrostMage/Modules/Features/CooldownTracker.js
Sharrq: Got it, ill get that fixed.

src/Parser/FrostMage/Modules/Features/FingersFrost.js
Sharrq: Same as the Brain Freeze thing, i didnt know if it was required or not, so i left it blank. ill remove it.

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
Sharrq: Ill get that fixed.

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
Sharrq: I guess i got a little confused about where that recommended number was coming from, so i was having a rough time figuring out how much i needed to adjust recommended since i didnt know the value of recommended. So i just put the numbers instead.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: will do.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: ok, will do.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: makes sense, ill change it.

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: see previous reply

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: ok, ill look at that.

src/Parser/FrostMage/Modules/Features/ThermalVoidDuration.js
Sharrq: Oops, ill fix that.

src/Parser/FrostMage/Modules/Features/WintersChill.js
Sharrq: Even though the debuff is on the target and not me, right?

src/Parser/FrostMage/Modules/Features/WintersChill.js
Sharrq: Ill take a second look at it, but Winter's Chill has a very very short duration, like just barely long enough to land one Instant Cast Ice Lance into it before it goes away, so i think that was to make sure that its still there when the spell hits, but ill double check.

src/Parser/FrostMage/Modules/Items/ShardofExodar.js
Sharrq: Yea i tried for about 2 hours to use an array instead and never could get it to work so i opted for listing them all out until i could figure out what i was doing wrong. But ill change it to what you suggested.

src/Parser/FrostMage/Modules/Features/WintersChill.js
sMteX: The point is though, if you get to this point, that means you have _just_ applied the Winter's Chill debuff, and while executing this function the time doesn't change. Makes sense in different events maybe, but when you're in `on_byPlayer_applydebuff(event)` and you check if the ability is Winter's Chill, then this will always return true.

Also, you should check the buff's presence based on specific timestamp, so `hasBuff(spellId, event.timestamp)`.

src/Parser/FrostMage/Modules/Features/WintersChill.js
sMteX: Yes, `onPlayer` is for debuffs applied on you (therefore you are the target of the spell), and `byPlayer` is for debuffs applied by you (you are the source of the spell).

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
sMteX: On the other hand, instead of "0:34 Average Icy Veins duration", "34s average Icy Veins duration" seems better. So maybe just round instead of `formatNumber`.

src/Parser/FrostMage/Modules/Features/IcicleTracker.js
sMteX: The `actual` value is equal to the value in `when()` and `recommended` is the same as the one used in `isLessThan()`, `isGreaterThan()` etc.

kfinch: The recommended is whatever the first number is. Doing them relative isn't required, we don't in the RestoDruid module.

src/Parser/FrostMage/Modules/Features/WintersChill.js
kfinch: Applied to you is actually toPlayer, not onPlayer

src/Parser/FrostMage/Modules/Features/WintersChill.js
kfinch: Sharrq, you should be checking damage events, check the damage was from Ice Lance and that the target has Winter's Chill

src/Parser/FrostMage/Modules/Items/ShardofExodar.js
```diff
@@ -0,0 +1,32 @@
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import { formatNumber } from 'common/format';
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class ShardOfExodar extends Module {
```
kfinch: Name it with items actual name, so ShardOfTheExodar

src/common/ITEMS.js
```diff
@@ -258,6 +258,18 @@ const ITEMS = {
     icon: 'inv_belt_92v4',
     quality: ITEM_QUALITIES.LEGENDARY,
   },
+  //Mage
+  SHATTERED_FRAGMENTS_SINDRAGOSA: {
+	  id: 151810,
+	  name: 'Shattered Fragments of Sindragosa',
+	  icon: 'inv_helmet_133',
+  },
+  SHARD_OF_EXODAR: {
```
kfinch: Use items actual name, so SHARD_OF_THE_EXODAR

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
Sharrq: I think formatNumber is already rounding it, cause when i did the math on it it was like 33.something. Ill take a look though.

src/common/SPELLS_MAGE.js
```diff
@@ -5,20 +5,262 @@
...
+	  id: 76613,
+	  name: 'Mastery: Icicles',
+	  icon: 'spell_frost_iceshard',
+  },
+  FROSTBOLT_CAST: {
+	  id: 116,
+	  name: 'Frostbolt',
+	  icon: 'spell_frost_frostbolt02',
+  },
+  FROSTBOLT_DEBUFF: {
```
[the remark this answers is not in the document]
kfinch: Interesting, I'll take a look when I'm in front of a computer again. For now it's fine to leave them in.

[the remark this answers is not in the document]
Sharrq: Essentially the ones with _CAST on them are the actual IDs that are in wowhead and _Debuff are the extra ones that were in the development tab. 

What would be really cool is if in the future you could add multiple IDs to a single spell so you dont have to keep creating new spells just to add an additional ID.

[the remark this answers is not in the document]
MartijnHols: `What would be really cool is if in the future you could add multiple IDs to a single spell so you dont have to keep creating new spells just to add an additional ID.`

This won't be done.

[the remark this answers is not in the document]
Sharrq: Well it was worth a shot :P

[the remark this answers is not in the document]
kfinch: There are a lot of weird extras sometimes. Might be best not to add them until you need to use them for something.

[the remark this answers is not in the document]
MartijnHols: You generally only want to add spells after investigating and finding what they are compared to other spells with the same name. Adding them without knowing, naming properly and taking it into consideration will only introduce bugs.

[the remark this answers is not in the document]
Sharrq: Ill fix them ... and everything else everyone suggested when i get home from work :)

src/Parser/FrostMage/CombatLogParser.js
```diff
@@ -0,0 +1,70 @@
...
+import ShardofExodar from './Modules/Items/ShardofExodar';
+
+
+class CombatLogParser extends CoreCombatLogParser {
+  static specModules = {
+    // Features
+    castEfficiency: CastEfficiency,
+    alwaysBeCasting: AlwaysBeCasting,
+    cooldownTracker: CooldownTracker,
+	WintersChillTracker: WintersChillTracker,
```
MartijnHols: Please be consistent with existing naming conventions: module names should start with a lower case.

src/Parser/FrostMage/Modules/Cooldowns/FrozenOrb.js
MartijnHols: Please use lower case class properties.

src/Parser/FrostMage/Modules/Features/CooldownTracker.js
Sharrq: So just to make sure i got this right, if it puts a buff on me it should be in cooldownSpells (like Icy Veins and Rune of Power), and if it doesnt put a buff on me it goes in castCooldowns (like Mirror Image and Frozen Orb)?

Would Bloodlust effects therefore go in cooldownSpells? or should it be removed alltogether?

src/Parser/FrostMage/Modules/Features/WintersChill.js
Sharrq: Actually i took another look at it, i think that second section was just to count the number of times Winter's Chill was applied and the top one was to count the number of Ice Lances that hit during winter's chill. I believe that is why there is a second one there ... that way i can figure out the number of times they didnt get an ice lance into winter's chill (for the suggestion). Is there a simpler way of getting the number of times a spell was cast or a buff was applied without manually counting it every time it is cast/applied?

sMteX: You can query the Ability Tracker to get casts of certain ability (by adding `abilityTracker: AbilityTracker,` into dependencies and then in suggestion probably do `const iceLanceCasts = this.abilityTracker.getAbility(SPELLS.ICE_LANCE.id).casts || 0;`), but that would count all Ice Lances for example, not just the ones in/outside of the buff.

Sharrq: I would need to get a count of the times Winter's Chill was applied to the target, not the number of times i cast Ice Lance. Can that be done with abilityTracker also?

Though I suppose i could just use the Brain Freeze Proc count from BrainFreeze.js since every Brain Freeze buffed Flurry will cause Winter's Chill.

sMteX: No, Ability Tracker only counts casts. You could list BrainFreeze as a dependency but I don't think it's that big of a deal to do 
```jsx
wintersChillCount = 0;

on_byPlayer_applydebuff(event) {
  if (event.ability.guid !== SPELLS.WINTERS_CHILL.id) {
    return;
  }
  this.wintersChillCount += 1;
}
```

Sharrq: Ok, i guess i was confused as to what was un-necessary and needed to be removed. I thought it was the on_applydebuff section, but your suggestion is almost identical to what is already in there ... so that must not be what was un-necessary. 

Or did you mean the Winter's Chill part of the on_byPlayer_cast section? If that is what you meant, then i think you may have assumed that Ice Lance is what caused the debuff therefore checking for the debuff is un-necessary ... but it is Flurry that debuff's the target with Winter's Chill..... So essentially, you get a Brain Freeze Proc, you expend that Proc on a free instant cast Flurry, Flurry debuff's the target with Winter's Chill, and you Cast Ice Lance during that Winter's Chill window. That is why im checking to see if Winter's Chill is there when Ice lance is cast/does damage (ill change it to the on damage one per kfinch's suggestion).

src/Parser/FrostMage/Modules/Items/ShardOfTheExodar.js
```diff
@@ -0,0 +1,43 @@
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import { formatNumber } from 'common/format';
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class ShardOfTheExodar extends Module {
+
+	damage = 0;
```
sMteX: See previous comment.

src/Parser/FrostMage/Modules/Items/ShardOfTheExodar.js
```diff
@@ -0,0 +1,43 @@
...
+  static dependencies = {
+		combatants: Combatants,
+	};
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasFinger(ITEMS.SHARD_OF_THE_EXODAR.id);
+  }
+
+  on_byPlayer_damage(event){
+    const BLOODLUST_BUFFS = [
```
sMteX: This array is constant and we usually extract constants out _before_ the class even begins. Just move it, no other changes needed.

src/Parser/FrostMage/Modules/Items/ShardOfTheExodar.js
```diff
@@ -0,0 +1,43 @@
...
+      ];
+    if (BLOODLUST_BUFFS.some(buff => this.combatants.selected.hasBuff(buff, event.timestamp))) {
+      this.damage += event.amount;
+    }
+  }
+
+  item() {
+    return {
+      item: ITEMS.SHARD_OF_THE_EXODAR,
+      result: `${formatNumber(this.damage)} Total Damage During Lust`,
```
sMteX: Standard format for items when it comes to showing damage is 
```jsx
result: `${this.owner.formatItemDamageDone(this.damage)}`,
```
This shows something like "7,38% / 126k DPS". You're free to add any additional info in the tooltip (look into other modules that use `<dfn>`tag in the item result)

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,59 @@
...
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Module from 'Parser/Core/Module';
+import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
+
+class IcyVeinsDuration extends Module {
```
kfinch: Should probably be called 'ThermalVoid'

src/Parser/FrostMage/Modules/Features/IcyVeinsDuration.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Module from 'Parser/Core/Module';
+
+class IcyVeinsDuration extends Module {
```
kfinch: Should probably be called 'Thermal Void'

src/Parser/FrostMage/Modules/Features/WintersChill.js
```diff
@@ -0,0 +1,60 @@
...
+    if (enemy.hasBuff(SPELLS.WINTERS_CHILL.id, event.timestamp)) {
+      this.iceLanceCasts += 1;
+    }
+  }
+  on_byPlayer_applydebuff(event) {
+	  if(event.ability.guid !== SPELLS.WINTERS_CHILL.id) {
+		  return;
+	  }
+		  this.wintersChillApplied += 1;
+	  }
```
kfinch: Indentation showing weird for me here

src/Parser/FrostMage/Modules/Items/ShardOfTheExodar.js
```diff
@@ -0,0 +1,44 @@
...
+  on_byPlayer_damage(event){
+    if (BLOODLUST_BUFFS.some(buff => this.combatants.selected.hasBuff(buff, event.timestamp))) {
+      this.damage += event.amount;
+    }
+  }
+
+  item() {
+    return {
+      item: ITEMS.SHARD_OF_THE_EXODAR,
+      result: `${this.owner.formatItemDamageDone(this.damage)} Total Damage During Lust`,
```
kfinch: 'total damage during lust' is an almost completely useless metric to show the player for SotE. The purpose of the item bonus displays is showing the player how much extra they got from equipping the item. Try going for showing "extra lusts" or something. Might be a pain to calculate, but I'd prefer you left out this module entirely than use this stat.

src/Parser/FrostMage/Modules/Items/Tier20_2set.js
```diff
@@ -0,0 +1,29 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  item() {
+    const T202setUptime = this.combatants.selected.getBuffUptime(SPELLS.FROST_MAGE_T20_2SET_BONUS_BUFF.id) / this.owner.fightDuration;
+    return {
+      id: `spell-${SPELLS.FROST_MAGE_T20_2SET_BONUS_BUFF.id}`,
+      icon: <SpellIcon id={SPELLS.FROST_MAGE_T20_2SET_BONUS_BUFF.id} />,
+      title: <SpellLink id={SPELLS.FROST_MAGE_T20_2SET_BONUS_BUFF.id} />,
```
kfinch: A few things here:

This should have an `on_initialized` that detects if player has the 2set and is only active then. Other SPELLS entries that have XXX_XXX_TXX_2SET_BONUS_BUFF are referring to a hidden aura that shows if the player has the set bonus. Those entries do NOT refer to the sets proc. The aura you're looking for for the frost mage 2t20 is this: http://www.wowhead.com/spell=242252/item-mage-t20-frost-2p-bonus. The 2t20 proc is called 'Frozen Mass', and should be named as such in its SPELLS entry.

Also, while you don't need to do this now, it is possible to calculate the actual percent damage increase you get from this bonus. That's a good candidate for future enhancement.

src/common/ITEMS.js
```diff
@@ -258,6 +258,18 @@ const ITEMS = {
     icon: 'inv_belt_92v4',
     quality: ITEM_QUALITIES.LEGENDARY,
   },
+  //Mage
+  SHATTERED_FRAGMENTS_SINDRAGOSA: {
```
kfinch: `SHATTERED_FRAGMENTS_OF_SINDRAGOSA` please

src/Parser/FrostMage/Modules/Features/BrainFreeze.js
```diff
@@ -0,0 +1,70 @@
...
+				});
+		}
+	}
+
+	statistic() {
+		return(
+			<StatisticBox
+				icon={<SpellIcon id={SPELLS.BRAIN_FREEZE.id} />}
+				value={`${formatNumber(this.totalProcs)}`}
+				label='Brain Freeze Procs'
```
kfinch: My rule of thumb when considering what should be shown as a statistic is it should be somewhat comparable between fights, and it should be somewhat in the player's control. Total procs is based on total fight time and random luck, and so fails both tests.

I'd like to see either 'procs per minute' here, or just another display of 'missed procs per minute'.

src/Parser/FrostMage/Modules/Features/FingersFrost.js
```diff
@@ -0,0 +1,60 @@
...
+					.regular(recommended+ 0.0).major(recommended + 0.05);
+			});
+	}
+
+	statistic() {
+		return(
+			<StatisticBox
+				icon={<SpellIcon id={SPELLS.FINGERS_OF_FROST.id} />}
+				value={`${formatNumber(this.totalProcs)}`}
+				label='Fingers of Frost Procs'
```
kfinch: Same suggestion as for Brain Freeze: PPM or missed PPM here.

src/Parser/FrostMage/Modules/Features/CastEfficiency.js
```diff
@@ -0,0 +1,187 @@
...
+    },
+    {
+      spell: SPELLS.ICE_LANCE,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    },
+    {
+      spell: SPELLS.FROZEN_ORB,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => 60,
```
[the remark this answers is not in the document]
Sharrq: Yea, i left it at the default cooldown until zero's cooldown reduction stuff was in place.

src/Parser/FrostMage/Modules/Items/ShardOfTheExodar.js
Sharrq: Yea, i tried to do the extra damage instead of total at one point, but all the examples of it that i found were examples where the legendary gave extra damage % not extra haste%, so i couldnt figure it out. And then i just left it at total. Probably should have just removed it. But i was considering changing it to (x possible lusts this fight) or something like that.

src/Parser/FrostMage/Modules/Items/Tier20_2set.js
Sharrq: Is there an existing spec that does this so i can potentially adopt that? I found examples of tier bonuses or talents that give a % damage increase using the getDamageBonus stuff .... but i didnt find any that give a boost to a secondary stat (like haste or in this case crit)

kfinch: Frozen Mass doesn't increase crit, it increases *crit damage*. It should be double checked, but I think this is just a +20% damage bonus that applies only to crits.

You can tell if a spell crit by checking the 'hitType' field in a damage event. A value of '2' indicates a crit (see https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/HIT_TYPES.js). So for each damage event you'd check if the player has Frozen Mass and if the damage was a crit, and if so use getDamageBonus with a value of 0.20.

kfinch:
Was just testing this out locally, and I got a "Winter's Chill Missed: -1". Looks like it happened because I cast Flurry -> Ice Lance -> Flurry -> Ice Lance during Bloddlust, which made the flurries land so close together that Winter's Chill was refreshed instead of applied. We obviously can't add a counter bump on refreshdebuff because Flurry applies the debuff multiple times. A possible solution would be counting Winter's Chill procs from consumed Brain Freeze, or honestly could just leave as is for now because this is an edge case.

kfinch:
Alright the quick fix is to use EnemyInstances here instead of Enemies, but I argue those two should be combined anyway...

Sharrq:
Made most of the requested changes. I tried doing the Missed Procs Per Minute, and i got it to work, i just really didnt like it, so I opted for Missed % instead. I figure its a good medium considering the longer the fight is and the more procs you get that you dont miss, the lower the percentage will be. Its not perfect, but i feel like it displays better. 

Also, i opted for moving Frozen Orb into the Cooldowns on Cast Efficiency but not Ebonbolt, thats just personal preference i suppose, but i consider Ebonbolt to be more of a rotational ability personally.

For Shard of Exodar i went for the x Lusts Used (x Possible) instead. 

For T20 2PC i did the damage calculation, but its not 100% accurate because it doesnt take into effect only crits .... im not sure how to do this. So if someone can tell me how to adjust it to only show the crit damage, then ill make the change. 

Also Sref, i know you said .includes would be better for the bloodlust section on Shard of the Exodar, but i tried it and it wasnt working, so i went back to .some. Maybe i did something wrong or there was an extra step, but it seems to be working properly with .some.

src/Parser/FrostMage/Modules/Items/Tier20_2set.js
```diff
@@ -2,26 +2,35 @@ import React from 'react';
...
+  damage = 0;
+
+  on_initialized() {
+	this.active = this.combatants.selected.hasBuff(SPELLS.FROST_MAGE_T20_2SET_BONUS_BUFF.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if (this.combatants.selected.hasBuff(SPELLS.FROZEN_MASS.id)) {
+      this.damage += getDamageBonus(event, 0.2);
+    }
```
sMteX: As kfinch said, AFAIK Frozen Mass increases _critical damage_, not all damage, so this should be encased in one more `if (event.hitType === HIT_TYPES.Crit)` provided you import HIT_TYPES from Parser/Core/HIT_TYPES.

MartijnHols: The preferred way is to return early instead of encasing with if-statements.

sMteX: Also it's recommended to extract the `0.2` before class as a constant, perhaps "FROZEN_MASS_DAMAGE_BONUS" (no magic numbers).

Sharrq:
Added those fixes for the Crit Damage stuff on Frozen Mass

kfinch:
The SotE calculation is correct for any fight less than 10 minutes, but will undercount beyond that. It could be improved, but I think this is a "we'll fix it in post" situation.

Sharrq:
Yea, i thought about adding in an additional calculation for after 10 minutes, but realistically a fight beyond 10 minutes isnt very common and then it starts getting complicated to calculate after that (When does the Bloodlust Debuff Fall off from when someone else gave you lust, when does your timewarp come off cooldown), plus you ideally want to line up lust with your cooldowns, so knowing when those come off. 

Im sure its doable though, with time.

MartijnHols:
Thanks a lot for the patience and effort from both @Sharrq and all reviewers!

PR #469, opened 2017-10-09
[Elemental Shaman] Storm Elemental bugfix

janvavra:
Just two very simple fixes of bugs with existing implementation for shamans using Storm Elemental (SE) talent

1. Maelstorm overview now works by registering maelstorm generating spell of the SE in SPELLS (doesn't work at all, when SE is selected on live)
2. Cast efficiency analysis switches between SE and Fire Elemental, based on talents selected (on live you are blamed for not casting FE if you have SE talented).


MartijnHols:
Thanks a lot!

PR #471, opened 2017-10-10
[Retribution] Spec completeness upgraded to good

CollCrom:
There are 5 main things you can do to improve your ret play:

1. Avoid capping your holy power
2. Using's Avenging Wrath/Crusade as much as possible
3. Making sure you're using holy power spenders inside of judgment
4. Using all of your blade of wrath procs
5. Using your artifact ability as much as possible.

I have implemented modules for all of these as well as all the top tier legendaries and some fluffy extra stats. I think using this implementation of the ret spec is a good starting spot but is no way complete yet. For this reason I think the spec completeness can be bumped up to good.

### DIFF
--- a/src/Parser/RetributionPaladin/CONFIG.js
+++ b/src/Parser/RetributionPaladin/CONFIG.js
@@ -20,7 +20,7 @@ export default {
   		Feel free to message me on discord or post in the spec discussion URL with any bugs or ideas for things I could work on!
   	</div>
   ),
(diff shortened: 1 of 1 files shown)


MartijnHols:
Thanks for the PR! I'll review this mostly by comparing suggestions given to what people point out during manual log reviews. I might also ask one of the ret mods to take a look at it.

MartijnHols:
I haven't heard back yet, from what I saw myself this looks good so I'll merge. I'll let you know if I get any more feedback. Thanks!!

PR #474, opened 2017-10-10
Expanded Windwalker CastEfficiency and change to a spell ID

Juko8:
Power strikes looks like it has some different spell ID's for the talent itself, the buff it grants and the consumption of the buff. I've changed it to the buff consumption that grants chi so it can be used for resource tracking, but unsure if that's the correct approach.


anom0ly:
Looks good. The way we have been treating it is to just append an identifier to the end of the spell name, so in this case it might be `POWER_STRIKES_BUFF` and `POWER_STRIKES_CAST` for the on use Spell ID.

PR #478, opened 2017-10-10
[Core] Fabricate `applybuff` events wherever they are missing

MartijnHols:
Fabricate `applybuff` events wherever they are missing so consumers can rely on just `applybuff` instead of having to do weird things with `combatantinfo` events and maybe even removebuff, buffstacks and refreshbuff events.

I think I fixed all modules that relied on the old behavior. Modules that forgot to take this into account should automatically be improved.


MartijnHols:
Fabricated events are already being shown in the events tab. This PR slightly changes this so they're colored differently;
image

I also modified the Light of Dawn reorderer to reveal the modified event as well:

image

src/Parser/Core/Modules/Helpers/ApplyBuffFixer.js
```diff
@@ -0,0 +1,130 @@
...
+        };
+        events.splice(firstEventIndex, 0, applybuff);
+        // It shouldn't happen twice, but better be safe than sorry.
+        this._buffsAppliedByPlayerId[targetId].push(spellId);
+      }
+    }
+    // endregion
+
+    // region `combatantinfo` based detection
+    // This catches buffs that never drop, such as Flasks and more importantly.
```
kfinch: and more importantly WHAT? I'm dying to know!

MartijnHols: I couldn't think of anything. The check is kinda useless.

MartijnHols: Actually I was thinking Atonements on other players. And beacons.

kfinch: yeah I figured you had something in mind, was mostly noting that that comment seemed to trail off mid sentence :)

PR #480, opened 2017-10-10
Holy Paladin: Add Divine Purpose cooldown reset of HS / LoD

MartijnHols:
Had to reorder HS Divine Purpose since it was bugged in logs and showing above the cast that triggered it.


MartijnHols:
Yeah just making sure. They're Blizzard combatlogs after all. 

Thanks for the review.

PR #483, opened 2017-10-11
[Core] Fix bug with scaling with Haste of CDs affected by CDRs & minor fixes

MartijnHols:
- Improved console output of main modules that need to be tuned properly
- Haste module now informs about not recognized spells to help reveal missing buffs
- Removed Lay on Hands from ABC's spells on GCD list since it isn't on the GCD
- Change Holy Shock's cooldown to consider Sanctified Wrath's CDR
- Added a unit test for the CDR bug fix

# Depends on #480  and #476


kfinch:
This PR had the changes from the previously merged one. I did the merge with Master, I'm pretty sure I did it right...

kfinch:
Ahh, after subtracting the stuff already merged in it looks like this is pretty straightforward, just a lot of debug stuff, some tests, and minor changes.

PR #484, opened 2017-10-11
Holy Paladin: Added a suggestion against casting Light of the Martyr while Holy Shock is available

MartijnHols:
image

### DIFF
--- a/src/Parser/HolyPaladin/Modules/PaladinCore/FillerLightOfTheMartyrs.js
+++ b/src/Parser/HolyPaladin/Modules/PaladinCore/FillerLightOfTheMartyrs.js
@@ -64,7 +64,7 @@ class FillerLightOfTheMartyrs extends Module {
     const inefficientLotmsPerMinute = this.inefficientCasts / (this.owner.fightDuration / 1000) * 60;
     when(inefficientLotmsPerMinute).isGreaterThan(0)
       .addSuggestion((suggest, actual, recommended) => {
-        return suggest(<span>You cast {this.inefficientCasts} <SpellLink id={SPELLS.LIGHT_OF_THE_MARTYR.id} />s while <SpellLink id={SPELLS.HOLY_SHOCK_CAST.id} /> was off cooldown. Try to <b>never</b> cast <SpellLink id={SPELLS.LIGHT_OF_THE_MARTYR.id} /> when something else is available.</span>)
+        return suggest(<span>You cast {this.inefficientCasts} <SpellLink id={SPELLS.LIGHT_OF_THE_MARTYR.id} />s while <SpellLink id={SPELLS.HOLY_SHOCK_CAST.id} /> was available. Try to <b>never</b> cast <SpellLink id={SPELLS.LIGHT_OF_THE_MARTYR.id} /> when something else is available.</span>)
           .icon(SPELLS.LIGHT_OF_THE_MARTYR.icon)
           .actual(`${this.inefficientCasts} casts while Holy Shock was available`)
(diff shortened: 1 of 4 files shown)


src/common/SPELLS_PALADIN.js
```diff
@@ -94,6 +94,11 @@ export default {
     icon: 'ability_paladin_lightofthemartyr',
     manaCost: 16500,
   },
+  MARAADS_DYING_BREATH_BUFF: {
```
kfinch: As far as I can see, no other spell with this name, so just call it `MARAADS_DYING_BREATH` - it's distinguished from the item by being `SPELLS.` rather than `ITEMS.`

src/Parser/HolyPaladin/Modules/PaladinCore/FillerLightOfTheMartyrs.js
```diff
@@ -6,25 +6,42 @@ import SpellLink from 'common/SpellLink';
...
+      // Not a filler
+      return;
+    }
 
-    const lightOfTheMartyrs = getAbility(SPELLS.LIGHT_OF_THE_MARTYR.id).casts || 0;
-    let fillerLotms = lightOfTheMartyrs;
-    if (this.maraadsDyingBreath.active) {
-      const lightOfTheDawns = getAbility(SPELLS.LIGHT_OF_DAWN_CAST.id).casts || 0;
-      fillerLotms -= lightOfTheDawns;
+    this.casts += 1;
```
kfinch: Slightly misleading: this isn't casts, this is "casts without Maraads". Maybe rename it `fillerCasts`, considering how it's used in suggestions?

MartijnHols: We're in the context "FillerLightOfTheMartyrs", I feel naming a properties "fillers" would be double which I feel would be detrimental.

src/common/SPELLS_PALADIN.js
MartijnHols: While true, suffixing it with `_BUFF` leaves no doubt about this and thus adds value imo.

src/Parser/HolyPaladin/Modules/PaladinCore/FillerLightOfTheMartyrs.js
kfinch: Clearly you should also rename the module LightOfTheMarty :smile:  ... but yeah that's a good enough explanation.

MartijnHols: If this becomes necessary then the class name will probably also become invalid. ;)

src/common/SPELLS_PALADIN.js
kfinch: While I've seen different ways of doing this, I thought the style thus far has been "always use the spell's exact name unless forced not too" e.g. when two spells have the same name.

kfinch: or I should say "the style is _supposed to be_"

MartijnHols: None of the rules are 100% hard (so they're more *guidelines*), if deviating from it makes things clearer then doing so is encouraged.

kfinch: ANARCHY CONFIRMED

sMteX: *\*sirens in the background**

kfinch:
guess I should merge this too, huh

PR #494, opened 2017-10-12
[Guardian] Fix display bugs in Ironfur and Frenzied Regen modules

FaideWW:
- Unbuffed Frenzied Regeneration statistic hides itself if there are 0 FR casts
 - Ironfur stack count returns 0 instead of undefined if there are no hits registered at 0 stacks
 - `1 stacks` -> `1 stack` in Ironfur statistic

### DIFF
--- a/src/Parser/GuardianDruid/Modules/Features/FrenziedRegenGoEProcs.js
+++ b/src/Parser/GuardianDruid/Modules/Features/FrenziedRegenGoEProcs.js
@@ -21,6 +21,9 @@ class FrenziedRegenGoEProcs extends Module {
   statistic() {
     const nonGoEFRegen = this.guardianOfElune.nonGoEFRegen;
     const GoEFRegen = this.guardianOfElune.GoEFRegen;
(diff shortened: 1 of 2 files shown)


src/Parser/GuardianDruid/Modules/Features/FrenziedRegenGoEProcs.js
```diff
@@ -21,6 +21,9 @@ class FrenziedRegenGoEProcs extends Module {
   statistic() {
     const nonGoEFRegen = this.guardianOfElune.nonGoEFRegen;
     const GoEFRegen = this.guardianOfElune.GoEFRegen;
+    if (nonGoEFRegen + GoEFRegen === 0) {
```
MartijnHols: Please use parenthesis so it's clear it's `(nonGoEFRegen + GoEFRegen) === 0`, not `nonGoEFRegen + (GoEFRegen === 0)`

MartijnHols:
thanks!

PR #496, opened 2017-10-12
[Retribution] Updated some helper text and gave spells charges

CollCrom:


CollCrom:
I have no idea, I have never touched it. How would I go about changing it back?

kfinch:
Looks like it got added in commit 86a645ab0dd20b67b422fc2ab7342c3b8d619c09 no idea why. The simple dumb way to fix it would be to copy pasta package-lock.json from master over your version on branch, then commit.

There's almost certainly the 'proper' way to do it with git commands, but I don't know what it is. Every damn time I need to revert something in Git I end up googling how to do it.

CollCrom:
Aight that should fix it!

PR #497, opened 2017-10-12
[Vengeance] Pain component bug fix

ronaldpereira:


src/common/SPELLS_DEMON_HUNTER.js
```diff
@@ -31,12 +31,12 @@ export default {
...
-  ARCANE_TORRENT: {
-    id: 202719,
-    name: 'Arcane Torrent',
-    icon: 'spell_shadow_teleport',
-  },
+	// Racial:
+	ARCANE_TORRENT_DH: {
+		id: 202719,
+		name: 'Arcane Torrent',
+		icon: 'spell_shadow_teleport',
```
kfinch: I take it this version is special because it gives DH resources?

MartijnHols: http://www.wowhead.com/spell=202719/arcane-torrent Fury, I think this spell should be called `ARCANE_TORRENT_FURY` instead.

src/common/SPELLS_DEMON_HUNTER.js
```diff
@@ -93,11 +93,13 @@ export default {
     name: 'Demon Spikes',
     icon: 'ability_demonhunter_demonspikes',
     painCost: 20,
+		charges: 2,
```
MartijnHols: This doesn't do anything, you can only define charges in cast efficiency.

src/common/SPELLS_DEMON_HUNTER.js
kfinch: The spell details in that wowhead link imply it generates both Fury and Pain... maybe `ARCANE_TORRENT_DH` actually is a reasonable name?

MartijnHols: `Silence all enemies within 8 yards for 2 sec and generate 15 Fury.  Non-player victim spellcasting is also interrupted for 3 sec.`

tooltip though :(

src/Parser/VengeanceDemonHunter/Modules/PainChart/PainComponent.js
```diff
@@ -45,7 +45,7 @@ const PainComponent = ({ categories, abilities }) => {
                     {wasted || '0'}
                   </td>
                   <td style={{ width: '25%', color: 'orange' }}>
-                    {canBeImproved && !ability.noCanBeImproved && 'Can be improved.'}
+                    {wasted > 0 ? 'Can be improved.' : ''}
```
MartijnHols: Within JSX it's common to do `{wasted > 0 %% 'Can be improved.'}`

ronaldpereira: I think the ternary conditional is more code-friendly. Almost every language has an ternary conditional in this exact same way. Putting the `{wasted > 0 %% 'Can be improved.}` is great because it's more objective, but `{wasted > 0 ? 'Can be improved.' : ''}` is easier to understand.

MartijnHols: I meant `{wasted > 0 && 'Can be improved.'}` which is common in dynamic languages

PR #498, opened 2017-10-12
[Vengeance] Added Cooldown Tracker for Metamorphosis

ronaldpereira:


src/Parser/VengeanceDemonHunter/CombatLogParser.js
```diff
@@ -35,6 +36,9 @@ class CombatLogParser extends CoreCombatLogParser {
     // Features
     alwaysBeCasting: AlwaysBeCasting,
     castEfficiency: CastEfficiency,
+		cooldownTracker: CooldownTracker,
+
+		// Spirit Bomb
     spiritBomb: SpiritBomb,
```
kfinch: This is one of the most tautological comments I've ever seen.

ronaldpereira:
I was looking into these lines of code and I thought "The role of each module isn't very clear, I should include this in a comment". I became a little bit lazy with my comments, but I can't stay in that way. Everyone have access to this code and everyone should be able to understand it though :smile: Thanks for the warning

PR #499, opened 2017-10-13
Reorganizing and adding to SPELLS_MONK

Juko8:
I've moved shared spells to the top and expanded the windwalker spells, also fixed some references in windwalker castefficiency. @anom0ly how do you feel about changing the spells to have 1 line each, i've been finding 600 lines hard to scroll through because every spell has 5. So far i have only changed the new spells and the shared spells to 1-line-each format.


Juko8:
I don't know what happened during the merge, but everything broke on my end afterwards and seems to be working for me after the revert.

kfinch:
Even after the revert, the Files Changed list showing an enormous number of changes I don't think you intended. I think I can cherrypick just the first 3 commits that have the stuff you want.

Juko8:
Yeah, that'd be best i believe, i haven't touched anything besides what those 3 commits have.

sMteX:
Personally I think this approach is making it less readable, and as for difficulty in navigating the file - Ctrl + F is your friend :) Just my two cents.

MartijnHols:
Last night I deduplicated all spells files, so you might have ran into merge conflicts. These will have to be resolved before we can merge this PR.

I don't think writing spells like this makes it more maintainable or readable. My main issue is that extra properties like `manaPerc` become less visible. I also prefer the old style because there's a lot more whitespace making it appear less daunting.

PR #501, opened 2017-10-13
[frost mage] Fixed ABC suggestion referencing Warlock abilities.

kfinch:


Sharrq:
Oops, missed that when i copied it, thanks for catching that!

Issue #400, opened 2017-09-26
Suggestion when an equipped legendary isn't the max item level (currently 970)


MartijnHols:
It would be a nice quality of life thing to check legendary item levels.

avelineestie:
I'll have a look at this.

fyruna:
This has been integrated, issue can be closed.

PR #506, opened 2017-10-13
Major suggestion if non-max ilvl legendary is used

fyruna:
Fix for issue #400 .

### DIFF
--- a/src/Parser/Core/Modules/Items/Legendaries.js
+++ b/src/Parser/Core/Modules/Items/Legendaries.js
@@ -21,7 +21,7 @@ class Legendaries extends Module {
     const slots = this.combatants.selected._gearItemsBySlotId;
     for(let index = 0; index < Object.keys(slots).length; index++) {
       const item = slots[index];
-      if (item.quality < ITEM_QUALITIES.LEGENDARY) {
+      if (item.quality === ITEM_QUALITIES.LEGENDARY) {
         this.legendaries.push(item);
       }
     }
@@ -33,7 +33,7 @@ class Legendaries extends Module {
         console.log(`Legendary ${item.id} is ${item.itemLevel}`);
       }
 
-      when(item.itemLevel === MAX_LEGENDARY_ILVL).isTrue().addSuggestion((suggest) => {
+      when(item.itemLevel < MAX_LEGENDARY_ILVL).isTrue().addSuggestion((suggest) => {
         return suggest(<span>You should consider upgrading <ItemLink id={item.id} /> to the maximum item level for legendaries (currently {MAX_LEGENDARY_ILVL}).</span>)
           .icon(item.icon)
           .staticImportance(SUGGESTION_IMPORTANCE.MAJOR);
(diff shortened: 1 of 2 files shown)


src/Parser/Core/Modules/Items/Legendaries.js
```diff
@@ -0,0 +1,45 @@
...
+
+  legendaries = [];
+  debug = false;
+
+  on_finished() {
+    const slots = this.combatants.selected._gearItemsBySlotId;
+    for(let index = 0; index < Object.keys(slots).length; index++) {
+      const item = slots[index];
+      if (item.quality === ITEM_QUALITIES.LEGENDARY) {
+        this.legendaries.push(item);
```
kfinch: We prefer using functional paradigms when possible, so here you should be looping with `slots.forEach( ...`

kfinch: Could also skip the if statement with a `filter`. Check out the functions available to a JS array, there's some pretty good stuff: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

src/Parser/Core/Modules/Items/Legendaries.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+  }
+
+  suggestions(when) {
+    for(const item of this.legendaries) {
```
sMteX: Same comment as kfinch had, this could most likely be replaced with 
```jsx
this.legendaries.forEach(item => {
.... 
});
```

src/Parser/Core/Modules/Items/Legendaries.js
```diff
@@ -0,0 +1,45 @@
...
+      if (this.debug) {
+        console.log(`Legendary ${item.id} is ${item.itemLevel}`);
+      }
+
+      when(item.itemLevel < MAX_LEGENDARY_ILVL).isTrue().addSuggestion((suggest) => {
```
sMteX: Could also be `when(item.itemLevel).isLessThan(MAX_LEGENDARY_ILVL)` but since we're not using the `actual` and `recommended` values it's not needed (just overall more used).

src/Parser/Core/Modules/Items/Legendaries.js
fyruna: I can't call forEach or filter on slots as the data is formatted like this:
image

kfinch:
Ah yes you're right it's an Object not an Array, but I see you figured out how to functionalize. One last nitpick while I'm being annoying: `Object.keys(slots).map(key => slots[key])` can be replaced with `Object.values(slots)`

kfinch:
Ok thanks! I'm done being difficult I swear.

fyruna:
@kfinch Rather have you be difficult than to break things or write ugly code. Right?

MartijnHols:
The functional approach takes some getting used to and will take some time until you prefer it (I personally disliked it for a long time), but once you get used to it you'll start to notice it's actually cleaner and easier.

I created a PR to patch up a few things I thought could be improved, if you have time @fyruna please consider reviewing it here: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/512

The two most important changes are the name of the class; I believe it should be more descriptive of what it does, and the suggestion text; the preferred style is to state what was observed te be wrong and only then suggest a course of action to solve it. The rest of the changes in the PR was just be me nitpicking.

PR #507, opened 2017-10-14
[Retribution] Blessing of the Ashbringer

CollCrom:
Since Blessing of the Ashbringer isn't always in the combatant info this check the uptime on Greater Blessing of Kings and Greater Blessing of Wisdom as well. If there is no BotA buff uptime it sets it to be equal to the lower of the those 2 values.

This was way harder than I thought it was going to be T_T


src/Parser/RetributionPaladin/Modules/PaladinCore/BlessingOfTheAshbringer.js
```diff
@@ -0,0 +1,61 @@
...
+const debug = false;
+
+class BlessingOfTheAshbringer extends Module {
+	static dependencies = {
+		combatants: Combatants,
+	};
+
+
+	//Blessing of the Ashbringer = BotA
+	botaUptime = 0;
```
MartijnHols: How about naming this just `uptime`? The context of the module is `BlessingOfTheAshbringer` so, specifying that it's about bota again is unnecessary (you wouldn't do `car.carSpeed`).

src/Parser/RetributionPaladin/Modules/PaladinCore/BlessingOfTheAshbringer.js
```diff
@@ -0,0 +1,61 @@
...
+
+	on_initialized() {
+		this.active = this.combatants.selected.traitsBySpellId[SPELLS.BLESSING_OF_THE_ASHBRINGER.id] > 0;
+	}
+
+
+	on_byPlayer_applybuff(event) {
+		const spellId = event.ability.guid;
+		const targetId = event.targetID;
+		this.botaUptime = this.combatants.selected.getBuffUptime(SPELLS.BLESSING_OF_THE_ASHBRINGER_BUFF.id) / this.owner.fightDuration;
```
MartijnHols: This would be better as a getter. I fear if this event isn't called in a long time the value will be off.

```jsx
get uptime() {
  return this.combatants.selected.getBuffUptime(SPELLS.BLESSING_OF_THE_ASHBRINGER_BUFF.id) / this.owner.fightDuration;
}
```

src/Parser/RetributionPaladin/Modules/PaladinCore/BlessingOfTheAshbringer.js
```diff
@@ -0,0 +1,61 @@
...
+		
+		if(spellId === SPELLS.GREATER_BLESSING_OF_WISDOM.id){
+			this.blessingOfWisdomUptime = this.combatants.players[targetId].getBuffUptime(spellId) / this.owner.fightDuration;
```
MartijnHols: this too could be a getter

src/Parser/RetributionPaladin/Modules/PaladinCore/BlessingOfTheAshbringer.js
```diff
@@ -0,0 +1,61 @@
...
+			debug && console.log(this.blessingOfWisdomUptime, 'wisdom uptime');
+		}
+		if(spellId === SPELLS.GREATER_BLESSING_OF_KINGS.id){
+			this.blessingOfKingsUptime = this.combatants.players[targetId].getBuffUptime(spellId) / this.owner.fightDuration;
```
MartijnHols: this too could be a getter

src/Parser/RetributionPaladin/Modules/PaladinCore/BlessingOfTheAshbringer.js
```diff
@@ -0,0 +1,61 @@
...
+			debug && console.log(this.blessingOfKingsUptime, 'kings uptime');
+		}
+
+		//If Blessing of the Ashbringer is undef/NaN set it to be 
+		//the lower of the uptimes between Kings and Wisdom
+		if(!this.botaUptime){
+			this.botaUptime = this.blessingOfWisdomUptime > this.blessingOfKingsUptime ? this.blessingOfKingsUptime : this.blessingOfKingsUptime;
```
MartijnHols: This could also be in the uptime-getter.

CollCrom:
I made the BotA uptime a getter but I am unsure how to go about making the Kings and Wisdom uptime into getters since they wouldn't know about targetId or sourceId

MartijnHols:
Couldn't you do `this.combatants.getBuffUptime(SPELLS.GREATER_BLESSING_OF_WISDOM.id)`? Then you don't need any targetId.

CollCrom:
Does this still work if there are multiple paladins giving out wisdom/kings?

CollCrom:
It's very hard to find a log with these conditions =/

MartijnHols:
Yes it does since you can only see buffs cast by the selected player or on the selected player. I also just made sure it doesn't bug out when BoW/BoK is cast on the selected player by someone else (realistically that would never happen): https://github.com/WoWAnalyzer/WoWAnalyzer/commit/705a0d9cddb427f311682eb7612961a8c570351b

CollCrom:
Thank you for clarifying! I made the suggested changes =]

PR #510, opened 2017-10-14
Added cooldown calculation of Serenity/SEF

Juko8:
Currently only dependant on amount of Split Personality traits - CoF calculation is probably not going to be added as it is already rarely used and doesn't reliably give extra casts and

### DIFF
--- a/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
+++ b/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
@@ -1,4 +1,7 @@
+import React from 'react';
+
 import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
 import CoreCastEfficiency from 'Parser/Core/Modules/CastEfficiency';
 
 class CastEfficiency extends CoreCastEfficiency {
@@ -39,7 +42,7 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.SERENITY_TALENT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: haste => 70,
+      getCooldown: (_, combatant) => combatant.owner.modules.sef.reducedCooldownWithTraits,
       isActive: combatant => combatant.hasTalent(SPELLS.SERENITY_TALENT.id),
     },
     // other spells
--- a/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
+++ b/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
@@ -1,7 +1,7 @@
-import React from 'react';
+
 
 import SPELLS from 'common/SPELLS';
-import SpellLink from 'common/SpellLink';
+
 import CoreCastEfficiency from 'Parser/Core/Modules/CastEfficiency';
 
 class CastEfficiency extends CoreCastEfficiency {
--- a/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
+++ b/src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
@@ -42,7 +42,7 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.SERENITY_TALENT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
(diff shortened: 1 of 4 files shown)


src/Parser/WindwalkerMonk/Modules/Spells/SEF.js
```diff
@@ -0,0 +1,40 @@
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SPELLS from 'common/SPELLS';
+
+class SEF extends Module{
```
MartijnHols: Please name this class (and its file) the full name (StormEarthAndFire) so it is much easier to spot in a long list of names.

src/Parser/WindwalkerMonk/Modules/Spells/SEF.js
```diff
@@ -0,0 +1,40 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  sefCDReduction = 0;
```
MartijnHols: This value is only set if `reducedCooldownWithTraits` is used before it. If it's not used it is 0. This would be error-prone. I think you should be able to turn this into a getter by moving lines 13-34 here.

Juko8: You want me to move sefCDReduction into the get reducedCooldownWithTraits() ?
I followed the way the Trueshot class was set up for marksmanship, so had imagined the layout wouldn't be problematic.

MartijnHols: No I meant something else. The value of `sefCDReduction` is set within `reducedCooldownWithTraits`, but I think making `sefCDReduction` a getter like `reducedCooldownWithTraits` that returns just its own value you can have both work without interfering with each other.

Juko8: I believe latest commit should be what you're asking for

src/Parser/WindwalkerMonk/Modules/Features/CastEfficiency.js
```diff
@@ -39,7 +42,7 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.SERENITY_TALENT,
       category: CastEfficiency.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: haste => 70,
+      getCooldown: (_, combatant) => combatant.owner.modules.sef.reducedCooldownWithTraits,
```
sMteX: This reference to StormEarthAndFire is not renamed yet.

src/Parser/WindwalkerMonk/Modules/Spells/StormEarthAndFire.js
```diff
@@ -0,0 +1,43 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  get sefCDReduction() {
+    const sefCDReduction = 0;
+    return sefCDReduction;
+  }
+
+  get reducedCooldownWithTraits() {
```
sMteX: I think all this is somewhat weird. Traits should be (haven't tested) available in on_initialized, I would do this calculation there. I would do this:
- remove sefCDReduction altogether
- make reducedCooldownWithTraits a class variable
- calculate the cool down reduction in on_initialized (there are no variables that would change it during the fight anyway)

In the end, in castEfficiency it would still be stormEarthAndFire.reducedCooldownWithTraits, just a variable and not a getter

Juko8: as mentioned in earlier comments, i copied the design from Marksmanship and its Trueshot class. I can't really provide much insight to design choices as i didn't make them.

src/Parser/WindwalkerMonk/Modules/Spells/SEF.js
MartijnHols: I think this part:
https://github.com/Juko8/WoWAnalyzer/blob/3d809b881f17913a04e330029a9deb8d3ba42320/src/Parser/WindwalkerMonk/Modules/Spells/StormEarthAndFire.js#L16-L37
should be the contents of  sefCDReduction`.

Also I think a better name for `sefCDReduction` would be `traitsCooldownReduction`. It's in the SEF class so prefixing it with that is unnecessary (you wouldn't do `car.carSpeed`).

Juko8: Alright, i believe i've gotten it looking like that now

Issue #388, opened 2017-09-25
List all legendaries in items


MartijnHols:
To make it easier to see what legendaries someone is using it would be nice to show all legendaries regardless of them being supported with its own module.

anom0ly:
With PR #511 I believe all legendaries have now been added to the `ITEMS` files. I don't believe there are any outstanding at this time.

Just want to confirm before this is closed.

sMteX:
Most likely all of them added (probably 7.3.5 Aman'thul's Vision is missing but that can be added once the tier lands).

PR #511, opened 2017-10-14
Add all remaining legendaries

avilene:
This should be it for #388: List all legendaries in item.

### DIFF
--- a/src/common/ITEMS.js
+++ b/src/common/ITEMS.js
@@ -862,12 +862,17 @@ const ITEMS = {
     icon: 'inv_pants_cloth_02',
     quality: ITEM_QUALITIES.LEGENDARY,
   },
-
-  // Subtlety Rogue Legendaries
-  SHADOW_SATYRS_WALK: {
-    id: 137032,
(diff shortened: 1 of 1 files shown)


MartijnHols:
Nice thanks! 👍 

What was actually meant with the issue was to show all items in the items box, such as for this log: https://wowanalyzer.com/report/kpLwvNVqCWZfMxHP/21-Mythic+Sisters+of+the+Moon+-+Kill+(7:38)/리로리로리:
image

This person has 2 legendaries equipped but since they're not being actively analysed they're not shown. The requested feature is to just show the legendaries regardless of them being equipped. So I'll merge this but leave the issue open for now.  Thanks again!

MartijnHols (approved):

fyruna:
@MartijnHols If i understand it correctly the idea is to show all the applicable legendaries in that box even if they aren't equipped? That would clutter it for quite a few players.

MartijnHols:
It will only show the 2 (soon 3) that are equipped, which are almost always relevant to analysis.

fyruna:
Okay I'll make it work!

Issue #31, opened 2017-05-22
Show concordance of legionfall uptime


buimichael:
This necessarily doesn't have to do with anything with your play style. Although it can be interesting to see how much this golden trait has contributed to.

MartijnHols:
I saw a Concordance of Legionfall display on the Shaman Discord that I'm pretty sure was done with this tool. Josh mentioned the guy who posted it is Exorsus Resto Shaman who maintains ERT. Hoping we can use that as basis.

![image](https://cloud.githubusercontent.com/assets/4565223/26501240/4bc90ff4-4239-11e7-9ba4-f7fb222200ae.png)

fyruna:
Working on this as we speak. Feedback is welcome!
![image](https://user-images.githubusercontent.com/5927637/31575615-36f0fa08-b0eb-11e7-9a13-96c3080f0c1e.png)

buimichael:
Fixed in #514

PR #514, opened 2017-10-14
Add Concordance of the Legionfall uptime tracker

avilene:
Adds a small tracker for the uptime of the Concordance of the Legionfall buff.
image

Should resolve issue #31 

To-do:
- [ ] Check wording for the box and tooltip
- [ ] Check if these values are the ones that should be displayed

### DIFF
--- a/src/Parser/Core/Modules/ConcordanceUptimeTracker.js
+++ b/src/Parser/Core/Modules/ConcordanceUptimeTracker.js
@@ -21,34 +21,41 @@ class ConcordanceUptimeTracker extends Module {
     combatants: Combatants,
   };
 
-  appliedBuff = null;
   statisticOrder = STATISTIC_ORDER.TRAITS(0);
 
-  statistic() {
-    const rank = this.combatants.selected.traitsBySpellId[SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_TRAIT.id];
+  get rank () {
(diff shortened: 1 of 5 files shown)


src/Parser/Core/Modules/ConcordanceUptimeTracker.js
```diff
@@ -0,0 +1,56 @@
...
+  AGILITY: SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_AGILITY,
+  STRENGTH: SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_STRENGTH,
+  VERSATILITY: SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_VERSATILITY,
+};
+class ConcordanceUptimeTracker extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  appliedBuff = null;
```
MartijnHols: This is only set if statistic is called, since this isn't guaranteed and it's not an event handler it is kinda unreliable. Would it be an idea to either remove this property or making it a getter? I'd prefer making it a getter; this separates the code leaving only display responsibilities to the statistic method.

fyruna: So moving this code up into a getter would be fine for you? No problemo.

MartijnHols (approved):
Thanks!

PR #515, opened 2017-10-14
[core / frost mage] Misc fixes

kfinch:
Updated PrePotion to use the new prepull buff handling. Should make the code less fragile with respect to potion durations.

Added check for prepull cast to ThermalVoid

Fixed a bug in WintersChill where it wasn't differentiating between multiple enemies with the same name. This should eliminate instances where missed winter's chill was a negative number. Fixed a bug where the suggestion wasn't correctly displaying number of missed winter's chill (was referencing `this.missed` when it needed to be using `missed`)


MartijnHols:
Thanks!

PR #517, opened 2017-10-14
[MMS HUNTER] Added Focus Capped Statisticbox

leapis:
Added a statisticbox for the focus tracking module. Right now it doesn't have a suggestion yet, but Putro has confirmed that he'll take care of that (and he has the best numbers for it, too). 

Screenshots:

image

image

### DIFF
--- a/src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
+++ b/src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
@@ -2,6 +2,7 @@ import React from 'react';
 import Module from 'Parser/Core/Module';
 import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
 import Icon from 'common/Icon';
+import { formatPercentage } from 'common/format';
 
 import FocusTracker from '../FocusChart/FocusTracker';
 
@@ -20,13 +21,13 @@ class TimeFocusCapped extends Module {
 
   statistic() {
     const totalFocusWaste = this.getTotalWaste;
-    const percentCapped = Math.round(this.focusTracker.secondsCapped/(this.owner.fightDuration/1000) * 10000)/100;
+    const percentCapped = formatPercentage(this.focusTracker.secondsCapped/(this.owner.fightDuration/1000));
     return (
       <StatisticBox
         icon = {<Icon icon='ability_hunter_focusfire' alt = 'Focus Wasted' />}
         label = 'Time Focus Capped'
         tooltip= {`You wasted <b> ${totalFocusWaste}  </b> focus. <br /> 
-        That's <b>  ${formatPercentage(totalFocusWaste/(this.owner.fightDuration/1000 * this.focusTracker.focusGen) * 10000)/100}% </b> of your total focus generated. 
+        That's <b>  ${formatPercentage(totalFocusWaste/(this.owner.fightDuration/1000 * this.focusTracker.focusGen))}% </b> of your total focus generated. 
         <br /> For more details, see the Focus Chart tab.`}
         value =  {`${percentCapped} %`}
       //Time not Focus-Capped: {Math.round((this.owner.fightDuration / 1000 - this.focusTracker.secondsCapped) * 100) / 100}s / {Math.floor(this.owner.fightDuration / 1000)}
--- a/src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
+++ b/src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
@@ -1,0 +1,103 @@
+import React from 'react';
+import {HorizontalBar} from 'react-chartjs-2';
+import Module from 'Parser/Core/Module';
+import StatisticsListBox, { STATISTIC_ORDER } from 'Main/StatisticsListBox';
+
+import FocusTracker from '../FocusChart/FocusTracker';
+
+class TimeFocusCapped extends Module {
+  static dependencies = {
+    focusTracker: FocusTracker,
+  };
+  centerStyle = {textAlign:'center'};
+
+  getData(){
+    this.cData = {
+      labels: [''],
+      datasets: [
+        {
+          label: 'Time Not Focus Capped',
+          backgroundColor: 'rgba(127,255,0,1)',
+          borderColor: 'rgba(0,0,0,1)',
+          borderWidth: 2,
+
+          data: [Math.round((this.owner.fightDuration / 1000)-this.focusTracker.secondsCapped)],
+        },
+        {
+          label: 'test2',
+          backgroundColor: 'rgba(255,255,255,1)',
+          borderColor: 'rgba(0,0,0,1)',
+          borderWidth: 2,
+
+          data: [Math.round(this.focusTracker.secondsCapped)],
+        },
+      ],
+    };
+    this.cOptions = {
+      scaleBeginAtZero: true,
+      tooltips:{
+        enabled: false,
+      },
+      legend: {
+        display: false,
+      },
+      scales: {
+        xAxes: [{
+          gridlines:{
+            display:false,
+          },
+          ticks:{
+            display:false,
+            max: Math.ceil(this.owner.fightDuration / 1000),
+          },  
+          barPercentage: 1,
+          categoryPercentage: 1,
+          stacked: true,
+        }],
+        yAxes: [{
+          gridlines:{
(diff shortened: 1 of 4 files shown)


MartijnHols:
For the bar might it be an idea to use the bars used in the past? e,g, from the holy paladin downtime:
image

leapis:
Okay here's the new look:
image
image

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+import Icon from 'common/Icon';
+
+import FocusTracker from '../FocusChart/FocusTracker';
+
+class TimeFocusCapped extends Module {
+  static dependencies = {
+    focusTracker: FocusTracker,
+  };
+
+  getTotalWaste(){
```
MartijnHols: This could be a getter; `get totalWaste()` so it can be accessed as a property.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+    let waste = this.focusTracker.secondsCapped * this.focusTracker.focusGen;
+    if (this.focusTracker.generatorCasts) {
+      Object.keys(this.focusTracker.activeFocusWasted).forEach((generator) => {
```
MartijnHols: How about a functional approach such as this?

```jsx
waste += Object.values(this.focusTracker.activeFocusWasted).reduce((sum, waste) => sum + waste, 0);
```

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+        waste += this.focusTracker.activeFocusWasted[generator];
+
+      });
+    }
+    return Math.round(waste);
+  }
+
+  statistic() {
+    const totalFocusWaste = this.getTotalWaste();
+    const tooltipData = "You wasted <b>" + totalFocusWaste + " </b> focus. <br /> That's <b>" + Math.round(totalFocusWaste/(this.owner.fightDuration/1000 * this.focusTracker.focusGen) * 10000)/100 + "% </b> of your total focus generated. <br /> For more details, see the Focus Chart tab.";
```
MartijnHols: I think a template string would make this cleaner, also please use formatPercentage instead of doing this manually;

```jsx
const tooltipData = `You wasted <b>${totalFocusWaste}</b> focus.<br />
That's <b>${formatPercentage(totalFocusWaste / (this.owner.fightDuration / 1000 * this.focusTracker.focusGen) * 100)}% </b> of your total focus generated.<br />
For more details, see the Focus Chart tab.`;
```

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+    const percentCapped = Math.round(this.focusTracker.secondsCapped/(this.owner.fightDuration/1000) * 10000)/100;
+    return (
+      <StatisticBox
+        icon = {<Icon icon='ability_hunter_focusfire' alt = 'Focus Wasted' />}
+        label = {'Time Focus Capped'}
```
MartijnHols: This doesn't need the curly braces; `label="Time Focus Capped"`

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+        tooltip= {tooltipData}
```
MartijnHols: Putting this in a temp variable is unnecessary.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+        value =  {percentCapped + " %"}
```
MartijnHols: This could be a template string as well.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+      //Time not Focus-Capped: {Math.round((this.owner.fightDuration / 1000 - this.focusTracker.secondsCapped) * 100) / 100}s / {Math.floor(this.owner.fightDuration / 1000)}
+      footer={(
+        <div className="statistic-bar">
+        <div
+          className="stat-health-bg"
+          style={{ width: `${(100-percentCapped)}%` }}
+          data-tip={`You spent <b>${100-percentCapped}%</b> of your time, or <b>${Math.round(Math.floor(this.owner.fightDuration/1000) - this.focusTracker.secondsCapped)}s</b> under the focucs cap.`}
```
MartijnHols: Typo: `focucs`

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+        >
+          <img src="/img/sword.png" alt="Time not focus capped" />
```
MartijnHols: I think this icon should be removed or replace by something more fitting. It does not appear relevant to the shown information.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,61 @@
...
+        </div>
+          <div
+            className="stat-overhealing-bg"
+            style={{ width: `${percentCapped}%` }}
+            data-tip={`You spent <b>${percentCapped}%</b>, or <b>${Math.round(this.focusTracker.secondsCapped)}s</b> of your time focus Capped.`}
+          >
+            <img src="/img/afk.png" alt="Time focus capped" />
```
MartijnHols: I think this icon should be removed or replace by something more fitting. It does not appear relevant to the shown information.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
leapis: I think it's relevance is a lot more clear to MMS hunters than it would be to other classes. While the icons might seem somewhat irrelevant with a lot of classes, the way in which we use focus makes them relevant when looking at the fight from the viewpoint of spenders = damage. Even our most common focus spenders generally do 5-10x more damage per cast when compared to our generators, and generators quite often make up less than 5% of our total damage in a fight. From this viewpoint, there are only 2 things happening when you're focus capped. Either you're actually "afk", or you're effectively "afk", ie only casting generators, which are incredibly weak in terms of damage and pretty much useless when you're already focus capped. Because this is the mindset that MMS hunters view their abilities, I think it's surprising appropriate to use a sword (symbolizing attacking) and an afk (symbolizing not attacking) icon, since our class finds very little benefit from non-focus spending abilities.

MartijnHols: The sword/afk icons are meant for the downtime statistic, and using them in two places to mean different things would be confusing, even if they might be appropriate enough.

leapis:
Okay, I've got a friend who's decent with graphic design thinking over the icons, and I'll make an issue for it too, but for now I've just removed the images and used just the bars.

src/Parser/MarksmanshipHunter/Modules/Features/TimeFocusCapped.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  statistic() {
+    const totalFocusWaste = this.getTotalWaste;
+    const percentCapped = formatPercentage(this.focusTracker.secondsCapped/(this.owner.fightDuration/1000));
+    return (
+      <StatisticBox
+        icon = {<Icon icon='ability_hunter_focusfire' alt = 'Focus Wasted' />}
+        label = 'Time Focus Capped'
+        tooltip= {`You wasted <b> ${totalFocusWaste}  </b> focus. <br />
```
MartijnHols: Nitpicking here, but I think indentation is key to making this readable. If the continued string is indented one more from the property, it stands out from the property list as being a continuation.

PR #518, opened 2017-10-14
[Frost Mage] Damage Count for Zann'esu Journey

Sharrq:
Added a Damage Count for Zann'esu Journey.

### DIFF
--- a/src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
+++ b/src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
@@ -14,7 +14,6 @@ class ZannesuJourney extends Module {
 
   damage = 0;
   stackCount = 0;
-  previousStackCount = 0;
 
   on_initialized() {
     this.active = this.combatants.selected.hasWaist(ITEMS.ZANNESU_JOURNEY.id);
@@ -24,38 +23,15 @@ class ZannesuJourney extends Module {
     if (event.ability.guid !== SPELLS.BLIZZARD.id) {
       return;
     }
-    this.previousStackCount = this.stackCount;
+    const buff = this.combatants.selected.getBuff(SPELLS.ZANNESU_JOURNEY_BUFF.id,event.timestamp);
+    this.stackCount = buff && buff.stacks || 0;
   }
 
   on_byPlayer_damage(event) {
     if (event.ability.guid !== SPELLS.BLIZZARD_DAMAGE.id) {
       return;
     }
-      this.damage += getDamageBonus(event, DAMAGE_BONUS * this.previousStackCount);
-      console.log('Blizzard Cast');
-      console.log(this.previousStackCount);
-      console.log(this.damage);
-  }
-
-  on_toPlayer_applybuff(event) {
-    if (event.ability.guid !== SPELLS.ZANNESU_JOURNEY_BUFF.id) {
-      return;
-    }
-    this.stackCount += 1;
-  }
-
-  on_toPlayer_applybuffstack(event) {
-    if (event.ability.guid !== SPELLS.ZANNESU_JOURNEY_BUFF.id || this.stackCount === 5) {
-      return;
-    }
-    this.stackCount += 1;
-  }
-
-  on_toPlayer_removebuff(event) {
-    if (event.ability.guid !== SPELLS.ZANNESU_JOURNEY_BUFF.id) {
-      return;
-    }
-    this.stackCount = 0;
+      this.damage += getDamageBonus(event, DAMAGE_BONUS * this.stackCount);
   }
 
   item() {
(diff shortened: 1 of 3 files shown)


src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
```diff
@@ -0,0 +1,69 @@
...
+
+class ZannesuJourney extends Module {
+
+  static dependencies = {
+		combatants: Combatants,
+	};
+
+  damage = 0;
+  stackCount = 0;
+  previousStackCount = 0;
```
kfinch: Why not just get stackCount by querying the entity at the moment of Blizzard cast? (rather than the manual tracking happening below)

I understand that you need previousStackCount so you can see how buffed damage events are.

If you're having the problem where the buff fades before the cast event happens, you could try an event reorderer.

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.BLIZZARD_DAMAGE.id) {
+      return;
+    }
+      this.damage += getDamageBonus(event, DAMAGE_BONUS * this.previousStackCount);
+      console.log('Blizzard Cast');
+      console.log(this.previousStackCount);
+      console.log(this.damage);
```
kfinch: Please put these in an `if(debug)` block

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
```diff
@@ -0,0 +1,45 @@
...
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasWaist(ITEMS.ZANNESU_JOURNEY.id);
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.BLIZZARD.id) {
+      return;
+    }
+    const buff = this.combatants.selected.getBuff(SPELLS.ZANNESU_JOURNEY_BUFF.id,event.timestamp);
```
kfinch: While this will work correctly, `getBuff` defaults to using the current timestamp, so you don't actually need to include the 2nd parameter.

Sharrq: im getting warnings because of the line beneath it that has && and || in the same line, but thats the only way i could get it to count correctly without it telling me that stacks is undefined.

Sharrq: nevermind, figured it out

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
```diff
@@ -0,0 +1,45 @@
...
+    this.stackCount = buff && buff.stacks || 0;
```
kfinch: You're right, this should be `this.stackCount = buff && (buff.stacks || 0);`
The lack of parens makes the order of operations unclear.

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
kfinch: I'm not talking about the line beneath it, I'm talking about this line right here. You can leave out the second parameter, `event.timestamp`

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
Sharrq: Out of curiosity, what would happen to the getDamageBonus() if the stack count is 0 or undefined? Would it register that Blizzard as doing 0 total damage? or would it register it as doing the normal damage with nothing extra?

Also it seems that each of these output differently when there is no stacks. Not sure which is right

(buff && buff.stacks) || 0
this Outputs a 0 when there is 0 stacks

buff && (buff.stacks || 0)
This Outputs "undefined" when there is 0 stacks ... but it doesnt error

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
Sharrq: Yea, i did remove that, i should have made a separate comment instead of adding that to this one

src/Parser/FrostMage/Modules/Items/ZannesuJourney.js
kfinch: Oh oops, yeah I made the wrong suggestion, should be `(buff && buff.stacks) || 0`

This is using javascript evaluating undefined as false when coerced to be a boolean. Basically that line says "if buff is defined and buff.stacks is defined, set to buff.stacks, otherwise set to 0".

MartijnHols:
Thanks!

PR #520, opened 2017-10-14
[core] Added StatTracker module

### LINKED ISSUE #406: Trinkets (and probably other buffs) that have a Haste rating increase effect add to base Haste instead of act as a Haste buff
This requires fixing the Haste module to apply these buffs differently from regular Haste buffs.

Research by @Iskalla: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/339#issuecomment-332317534

### LINKED ISSUE #417: Add Erratic Metronome to stacking haste buffs
Erratic Metronome is a trinket from Nighthold (Chronomatic Anomaly I believe), giving spells the chance to grant a stack of haste buff (Acceleration it's called I think), refreshing does NOT reset the duration (only ramps up the haste).

### LINKED ISSUE #482: [core] Prepull haste rating buffs are double counted
There is a bad interaction between the Haste module and the ApplyBuffFixer module. **Haste rating buffs that are applied prepull will be double counted because they will be present in both base haste rating from combatantinfo and in the fabricated applybuff events pulled from combatantinfo.** I don't think this is as simple as making Haste ignore events with `__prepull`, because then haste _percentage_ buffs applied prepull won't be handled at all. It seems likely that haste rating and haste percentage buffs should be handled differently.

I don't yet have hard proof of this, but it looks to be the case from the code.

kfinch:
I'm putting in this PR more because I want a review than because I think this should be merged exactly as is. Even if you approve, please _don't_ merge immediately.

This is a utility module intended to track the selected player's current stat rating. We pull initial stat rating from the combatantinfo, and track changes using a list of buffs. Obviously the list is currently quite sparse, and just there for testing. It can be freely added to, similar to the one in Haste.

The getters provided by this module are intended to replace what's currently in Combatant, but I've left Combatant untouched for now so this can be migrated over time.

I intend to use the functionality added here to assist in implementing a stat weight module for Resto Druids, which will be added in a later PR.

-----------------------------

Fixes #406, Fixes #482, Fixes #417


kfinch:
Note this only works on selected combatant because we only get all applied and removed buffs for the selected combatant. I don't know why you'd need this module functionality for anyone else anyway.

src/common/stats.js
```diff
@@ -8,6 +8,7 @@ Take the stat of an item at a given iLvL (for example, 300 Crit at 850 iLvL).
 To figure out how much crit that would have at 860 iLvL, you'd do: `300 * 1.15 ^ ((860 - 850) / 15) * 0.994435486 ^(860-850)` for 311 Crit.
 Once the stat at a given iLvL is found, the only variable that needs to be changed is `iLvL`
 */
+// can confirm this formula works if given ilevel/stat other than whatever the "base" is. - kfinch
```
MartijnHols: This would probably be better suited as a unit test than a comment.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,257 @@
...
+import SPECS from 'common/SPECS';
+import { calculateSecondaryStatDefault } from 'common/stats';
+import { formatMilliseconds } from 'common/format';
+
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+const debug = false;
+
+// TODO: add stam, speed, avoidance, leech
```
MartijnHols: I think adding them right away would be for the best.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,257 @@
...
+// TODO: stat constants somewhere else? they're largely copied from combatant
+class StatTracker extends Module {
```
MartijnHols: I believe this should be part of the Combatant class as soon as it's merged. If it's not we will soon find us with modules relying on this and it would become much harder to do. Migrating all existing usages of the properties of Combatant will be worth the trouble (and messy code) saved.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,257 @@
...
+  get currentVersRating() {
+    return this._stats.vers;
+  }
+
+  /*
+   * For percentage stats, the percentage you'd have with zero rating.
+   * These values don't change.
+   */
+  get baseCritPercentage() {
+    return 0.08;
```
MartijnHols: Isn't this different for each class/spec?

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,257 @@
...
+  get baseVersPercentage() {
+    return 0;
+  }
+
+  /*
+   * For percentage stats, this is the multiplier to go from rating to percent (expressed from 0 to 1)
+   * These values don't change.
+   */
+  get critRatingToPercent() {
+    return (1 / 40000);
```
MartijnHols: Parens are unnecessary and imo distracting.

src/Parser/Core/Modules/StatTracker.js
kfinch: yeah I'm going to add them before this gets merged, just pushing something out to be reviewed

src/Parser/Core/Modules/StatTracker.js
kfinch: My concern is that the functionality presented by this module only really works on the selected combatant, while the Combatant class is meant for all combatants.

Also, this functionality is implemented as a Module, and Combatant is not a module. Do you mean that this should be in Combatants? I'm not sure what you envision going where.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,257 @@
...
+// TODO: stat constants somewhere else? they're largely copied from combatant
+class StatTracker extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  static STAT_BUFFS = {
+    //// POTIONS ////
+    [SPELLS.POTION_OF_PROLONGED_POWER.id]: { str: 2500, agi: 2500, int: 2500 },
+    // probably no need to add flasks/food because they never fall
```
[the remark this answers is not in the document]
kfinch: Yeah there'd be nothing wrong with adding them, I guess there's no reason not to.

src/Parser/Core/Modules/StatTracker.js
kfinch: Uhh is it? It's currently implemented in Combatant as a flat 0.08.

src/Parser/Core/Modules/StatTracker.js
kfinch: I'll get rid of them.

src/common/stats.js
kfinch: Really, I should link to a math textbook or something that shows the function is ... commutative? I dunno..

MartijnHols: You don't have to do anything, just stating that proving this with a unit test would be more reliable and (excuse me) useful than a comment. You wouldn't do any math in such a test just check that given baseItemLevel X, baseStat Y and itemLevel Z, it returns the right (precalculated) value, even if those values aren't "base".

src/Parser/Core/Modules/StatTracker.js
MartijnHols: Hmmm yes this might be annoying. I quite dislike the Entities/Entity setup we have right now.

I think this would have been decent as a SelectedCombatant : Combatant class, but it doesn't solve the event handler issue.

src/Parser/Core/Modules/StatTracker.js
MartijnHols: I guess we can hope it's right 🙏

kfinch:
A quick review of what's currently using Combatant's stat getters:

Stamina: NONE
Strength: NONE
Agility: NONE
Intellect: NONE

Crit: NONE
Haste: Many places over many modules, largely cooldown related. It also looks like GuardianDruid has some sort of custom haste tracker already.
Mastery: Modules in RestoDruid, RestoShaman, and EnhancementShaman
Vers: A GuadianDruid module and a VengDH module

Armor: NONE
Block: NONE
Parry: NONE
Avoidance: NONE
Leech: NONE
Speed: NONE

kfinch:
@MartijnHols Continuing the class structure conversation: I think StatTracker is actually well suited as a classic Module, and the tracker itself is good as is. What's awkward is the stat constants, which don't seem to belong in Combatant or StatTracker. I propose adding a 3rd class (PlayerStatConstants?), just a utility basically, which is just stat conversion stuff. Combatant would keep its statRating getters, as its pulling them straight from the combatantinfo, but lose the percentage stuff. StatTracker would lose the stat percentage stuff (or maybe the percentage stuff calls PlayerStatConstants).

MartijnHols:
Yeah you're right it's better as a classic module. How about naming it SelectedCombatant and instead of moving StatTracker into Combatant, we move Combatant into StatTracker? This could be as simple as SelectedCombatant extending Combatant.

The alternative is to split Combatant completely up into different parts; base info (name, specid), stats, talents, traits, and gear. I don't think that would really be better, there's no gain in usage and I think it gets uglier having to specify all these different parts as dependencies even though they're basically just transformers for data from the combatant info.

kfinch:
I'm not sure I understand, SelectedCombatant needs to extend Module right? It couldn't also extend Combatant.

kfinch:
@MartijnHols Ok so my problem with moving Combatant into this module is that Combatant in its current form is used by almost every module. This would be an enormous refactor that I don't think should be tied to implementing stat tracking.

For now, how about something pretty close to what's in this PR. Combatant keeps all the info direct from the combatantinfo: spec, starting stat ratings, traits, talents, gear. It loses only the conversion to statPercentages. In the future we might want to stop combatant/entity from being in charge of tracking buffs on, but that shouldn't be part of this PR.

StatTracker becomes the one stop shop for player stat related things. It tracks the selected player's current stats and it also provides utilities for turning ratings into percentages.

Let me know what you think.

MartijnHols:
Merging Combatant into this class is not a prerequisite for merging this class, I'm just thinking about what the ideal end result would be.

On the other hand I wonder if it really is such an enormous refactor; I think we should be able to use almost an identical public API of the Combatant class so wouldn't the only change be from using `combatants.selected` to `selectedCombatant` or whatever the class ends up being called?

I recognize that buff tracking would be troublesome and requires more planning to work out, I haven't given it much thought yet. The Haste module as well might need changes so that the entire problem all these classes solve is solved well as a whole instead of with many separate ideas and solutions.

I'd like to avoid reaching a state where we have 2 solutions for the same problem, both applied only half way (we already have this issue in several other places and there's still no indicator that it will ever be resolved).

Basically; what would the ideal situation be with regards to stat, gear, talents and trait tracking, as well as buffs? And Haste. I guess this could all be combined to be called the *selected combatant* problem.

kfinch:
@MartijnHols Ok check out this terrible idea:

We have an Entities module which is in charge of creating and making public a list of objects representing entities in the encounter, and nothing else. Then, category of thing we want to know about entities, say stats, gear, talents, etc... is its own small module that isn't used directly but instead inserts that information in the relevant entity/combatant object. This way the programming logic of handling each category of information is separated into many class files, but from the user's perspective this is all just information that can be queried from an entity object.

As a concrete example, my StatTracker would be its own module as implemented, but when the selected combatant's stats change, the StatTracker module just modifies the appropriate value in the selectedCombatant object. Then the end user would only have to statically include Entities, and access it by say `this.entities.selectedCombatant.currentIntellect;` or something like that.

kfinch:
Note that this would allow existing accesses to work pretty much the same... even exactly the same if we add a simple compatibility layer.

MartijnHols:
Sounds good, except I'd make it not set the Entity's properties but just add getters to Entity for whatever API is decided to become public.

kfinch:
Not sure what inserting getters looks like in terms of syntax, but yeah that sounds good.

kfinch:
@MartijnHols to be clear, when you say adding a getter are you suggesting using `Object.defineProperty()`, or do you mean something else?

MartijnHols:
No, `get int() { return this.starTracker.int; }`

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,335 @@
...
+  on_toPlayer_changedebuffstack(event) {
+    this._changeBuffStack(event);
+  }
+
+  _changeBuffStack(event) {
+    const spellId = event.ability.guid;
+    const statBuff = this.constructor.STAT_BUFFS[spellId];
+    if(statBuff) {
+      // ignore prepull buff application, as they're already accounted for in combatantinfo
+      // we have to check the stacks count because Entities incorrectly copies the prepull property onto changes and removal following the application
```
MartijnHols: "incorrectly" maybe it's a feature ;)

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+  get baseSpeedPercentage() {
+    return 0;
+  }
+
+  /*
+   * For percentage stats, this is the multiplier to go from rating to percent (expressed from 0 to 1)
+   * These values don't change.
+   */
+  get critRatingToPercent() {
```
MartijnHols: I think this should be `critRatingPerPercent` so that it can be used for more things that just calculating the percentage.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,335 @@
...
+
+
+  }
+
+  _startingStats = {};
+  _stats = {};
+
+  on_initialized() {
+    this._startingStats = {
+      str: this.combatants.selected.strength,
```
MartijnHols: ok you'll think this is annoying, but please use the full name like in combatants; strength, agility, etc

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,335 @@
...
+  }
+  get speedRatingToPercent() {
+    throw new Error('Speed hasn\'t been implemented yet.');
+  }
+
+  /*
+   * For percentage stats, the current stat percentage as tracked by this module.
+   */
+  get currentCritPercentage() {
+    return this.baseCritPercentage + (this.currentCritRating * this.critRatingToPercent);
```
MartijnHols: this would become / then

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -0,0 +1,335 @@
...
+      if(event.oldStacks === 0 && event.prepull) {
+        debug && console.log(`StatTracker prepull application IGNORED for ${SPELLS[spellId] ? SPELLS[spellId].name : spellId}`);
+        return;
+      }
+      debug && console.log(`StatTracker: (${event.oldStacks} -> ${event.newStacks}) ${SPELLS[spellId] ? SPELLS[spellId].name : spellId}`);
+      this._changeStats(this._stats, statBuff, event.newStacks - event.oldStacks);
+    }
+  }
+
+  _changeStats(orig, change, factor) {
```
MartijnHols: seeing as this isn't reused *ever*, I think it would be cleaner to not have `orig` as a parameter but just directly set the value. Mutating parameters is generally undesirable for being hard to track and fragile, and if there's no reason at all to do it then I think we should avoid it.

src/Parser/Core/Modules/StatTracker.js
kfinch: 🤔

src/Parser/Core/Modules/StatTracker.js
kfinch: yeah yeah yeah

src/Parser/Core/Modules/StatTracker.js
kfinch: It's a conversion factor, and 'perPercent' sounds stupid! Seriously though, I originally had it like you suggest and found myself just dividing everywhere. If you really need the perPercent, you can just divide by it instead of multiply.

src/Parser/Core/Modules/StatTracker.js
kfinch: A holdover from an earlier design, I'll change it

kfinch:
Updated

kfinch:
@MartijnHols stop doing Pally stat weights and review these changes! :imp:

src/Parser/Core/Modules/StatTracker.js
MartijnHols: I think it makes more sense to do 12,000 / 400 to get the crit percentage than 12,000 * number I can't remember nor type. Isn't this most of the applications of this value?

Issue #399, opened 2017-09-26
Distance moved statistic


MartijnHols:
With the Holy Paladin mastery effectiveness module we learned it's possible to find a player's position and calculate distance. Using a similar technique it should be possible to get a high accuracy *distance moved* statistic. It could indicate ineffective play, probably reveals some fights are more movement intensive than other, but probably it would mostly be for the fun of it.

MartijnHols:
This might be a good first issue if you're good at math and worked with combatlogs in the past.

MartijnHols:
Fixed in #522

PR #522, opened 2017-10-14
Add statistic for Total Distance Moved (#399)

avilene:
Relates to issue #399.

Nothing much to add here. I might have missed an event or two to check for positions but I think I got most of them. I based most of this off the `MasteryEffectiveness` module for HolyPala.

image

### DIFF
--- a/src/Parser/Core/Modules/DistanceMoved.js
+++ b/src/Parser/Core/Modules/DistanceMoved.js
@@ -1,0 +1,74 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber } from 'common/format';
+
+import Module from 'Parser/Core/Module';
+
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+const debug = false;
+
+class DistanceMoved extends Module {
+    lastPositionUpdate = null;
+    totalDistanceMoved = 0;
+
+    // Events
+    on_cast(event) {
+        this.updatePlayerPosition(event);
+    }
+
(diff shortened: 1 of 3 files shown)


MartijnHols:
Please update the core changelog :D

fyruna:
Done @MartijnHols

src/Parser/Core/Modules/DistanceMoved.js
```diff
@@ -0,0 +1,74 @@
...
+        this.updateTotalDistance(event);
+        this.lastPositionUpdate = event;
+    }
+
+    statistic() {
+        debug && console.log(`Total distance moved: ${this.totalDistanceMoved} yds`);
+        return (
+            <StatisticBox
+            icon={<SpellIcon id={SPELLS.FRENETIC_SPEED_TALENT.id} />}
+            value={`${formatNumber(this.totalDistanceMoved)} yds moved`}
```
MartijnHols: How about making this show the total amount, e.g. `21,342 yards` (or w/e)? The label already shows what it is.

How much confidence do you have in the accuracy of this? I'm half thinking it might warrant adding a `≈` in front of the distance.

fyruna: Yeah it probably needs that in front of it, should I still format the number or not?

MartijnHols: Yeah I think the extra detail would be great.

fyruna: I'll add it in front.

MartijnHols (approved):

MartijnHols:
Thanks! This is cool :D

PR #526, opened 2017-10-15
[Frost Mage] Directory Structure

Sharrq:
Updated directory structure for frost mage, with the shared modules moved into a SharedModules folder inside "Mage"


MartijnHols:
Hey, I'd like to propose the following folder structure instead:

 * Mage/
    * Arcane
         * Modules
    * Fire
         * Modules
    * Frost
         * Modules
    * Shared
         * Modules

It's only a minor difference, but I think it makes more sense.

fyruna:
I agree with this exact comment that @MartijnHols wrote.

src/Parser/Mage/Frost/Modules/Items/Tier20_2set.js
```diff
@@ -5,7 +5,7 @@ import SpellLink from 'common/SpellLink';
 import Combatants from 'Parser/Core/Modules/Combatants';
 import Module from 'Parser/Core/Module';
 import HIT_TYPES from 'Parser/Core/HIT_TYPES';
-import getDamageBonus from '../MageCore/GetDamageBonus';
+import getDamageBonus from 'Parser/Mage/Shared/Modules/GetDamageBonus';
```
sMteX: Both this and Zann'esu Journey could be replaced with relative paths but that doesn't really change anything (and in this case it would make the import path even longer and perhaps less readable) but this should be equal to `../../../Shared/Modules/GetDamageBonus`.

src/Parser/Mage/Frost/CombatLogParser.js
```diff
@@ -7,6 +7,8 @@ import Talents from 'Main/Talents';
 import CoreCombatLogParser from 'Parser/Core/CombatLogParser';
 import DamageDone from 'Parser/Core/Modules/DamageDone';
 
+import ShardOfTheExodar from 'Parser/Mage/Shared/Modules/Items/ShardOfTheExodar';
```
sMteX: It's not a huge deal, mainly depends on if you want to use relative paths or absolute paths. The import path should be equal to (you start in Frost) `../Shared/Modules/Items/ShardOfTheExodar`

Sharrq: I changed this back to a relative path since its only one directory back and because i wanted to keep all the items together.

src/Parser/Mage/Frost/Modules/Items/Tier20_2set.js
Sharrq: Personally, i dont like using relative paths when i have to go backwards more than 1-2 directories, so i usually use absolute paths in that case just because its simpler to see it and know where its pointing as opposed to trying to decipher a relative path that goes back a bunch of directories. But thats just me. But when its in the same directory or one back, then i dont have a problem using relative paths.

sMteX: I agree, just wanted to point out the possibility :-)

Issue #398, opened 2017-09-26
Convert tab panel to horizontal tab list so it can work on mobile too and there's more space in general. Like the dev tools in chrome


MartijnHols:
Like the dev tools in chrome
![image](https://user-images.githubusercontent.com/4565223/30874446-9e866d46-a2f0-11e7-8d17-30c3bed23161.png)

anom0ly:
With PR #528 merged, can this issue now be closed?

MartijnHols:
Looks really sweet, on mobile too! 👍

PR #528, opened 2017-10-15
[UI] Make tab menu items horizontal

avilene:
As suggested in #398 I've made the tab menu items horizontal. 

image

### DIFF
--- a/src/Main/Results.css
+++ b/src/Main/Results.css
@@ -4,3 +4,30 @@
 .statistics > div {
   padding: 0 10px;
 }
+
+.results .navigation {
+  border-bottom: 1px solid rgba(0, 0, 0, .4);
+  box-shadow: 0 1px 0 0 rgba(255, 255, 255, .1);
+  margin-top: -10px;
+}
+.results .btn-link {
+  min-width: 120px;
+  min-height: 60px;
+  color: #aaaaaa;
+  outline: none;
+  padding: 5px;
+}
+
+.results .btn-link, .results .btn-link:hover, .results .btn-link:active {
+  text-decoration: none;
+}
+
+.results .btn-link:hover {
+  color: #eeeeee;
+  background-color: rgba(255, 255, 255, 0.1);
+}
+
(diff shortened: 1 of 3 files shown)


MartijnHols:
Hey thanks for tackling another issue! This is a good start, but I think it needs a few more changes before it's ready. The goal I had envisioned was to make it similar to the devtools tabs;

image

Another inspiration could be GitHub's tabs;
image
image

The gist of it is that the headers MENU and SUGGESTIONS can be removed with such a layout as the tab selection itself and the highlighting of the active menu item should suffice. This would save also save some vertical space from your current proposed solution. The difficulty in doing this would be in giving the minor importance toggle a good looking spot, and making the tabs fit neatly within the layout is probably also non trivial.

There are several solutions for fixing the mobile layout, a simple and temporary one would be to set the tab items to not wrap (white-space: nowrap), better might be similar to the devtools; horizontal scrolling, or perhaps a `...` button like I've seen in other layouts.

fyruna:
Okay! I can poke in the css files as well, right?

MartijnHols:
Yes

fyruna:
This is what I've made it into, suggestions are welcome.
image
image

Pewtro:
The blue outline in the top most picture when it's selected looks out of place

fyruna:
@Pewtro Good find, removed it. Also added a hover state on the non-selected tabs
image

MartijnHols:
I think we should use something other than orange as the active state. Our color scheme doesn't really include orange. I realize the current active state of the menu items is too hard to see, so maybe you can play around with the red border?

fyruna:
image
How about this, and perhaps making the text bold when its selected so its even more clear?

MartijnHols:
Looks good! Unsure about making the text bold, it might also make the tabs jump a bit, but feel free to give it a try.

fyruna:
Nah I'll leave it as is, bold might just be a tad overdone.

MartijnHols (approved):
Thanks!

Issue #505, opened 2017-10-13
Split the ITEMS.js


sMteX:
Would be nicer to split the ITEMS.js in the same manner as SPELLS are. It's getting quite bloated with all the items and having it per class and shared epics/leggos in separate file too would make it easier to maintain. 

In order not to break anything (or having to rename the occurences _everywhere_), it should be like SPELLS.js, therefore just a combination of all the ITEMS_*.js, but still the same file in the end so the imports are intact.

sMteX:
I took on the issue myself, expect PR soon :D

sMteX:
Fixed with #530 .

PR #530, opened 2017-10-15
[Refactor] Split the ITEMS into ITEMS_*

sMteX:
Fixes issue #505 . Everything else should stay intact (still imports as "common/ITEMS") so should be fine to merge. Also added some missed ones (at least Norgannon's Foresight).


MartijnHols:
Thanks very much!

PR #531, opened 2017-10-15
[Windwalker] Added tracking of Chi saved by Katsuo's Eclipse

Juko8:


src/Parser/WindwalkerMonk/Modules/Items/Katsuos.js
```diff
@@ -0,0 +1,34 @@
...
+  
+
+  on_initialized() {
+    this.active = this.combatatants.selected.hasFeet(ITEMS.KATSUOS_ECLIPSE.id);
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+
+    if (spellId === SPELLS.FISTS_OF_FURY_CAST.id &&! this.combatatants.selected.hasBuff(SPELLS.SERENITY_TALENT.id)) {
```
sMteX: Seeing as Serenity is actually a buff too, this should be 
```jsx
if (spellId === SPELLS.FISTS_OF_FURY_CAST.id && !this.combatants.selected.hasBuff(SPELLS.SERENITY_TALENT.id, event.timestamp))
```
As seen, please place the space between && and !
Also, are Fists of Fury the only Chi spender in WW?

src/Parser/WindwalkerMonk/Modules/Items/Katsuos.js
```diff
@@ -0,0 +1,34 @@
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import Module from 'Parser/Core/Module';
+
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class Katsuos extends Module {
```
sMteX: The module's name should be full length (unless it's something like the Priest cloak X'anshi, Shroud of Archbishop Benedictus) - so KatsuosEclipse (remember to rename it in the CombatLogParser too).

src/Parser/WindwalkerMonk/Modules/Items/Katsuos.js
Juko8: Katsuo's eclipse reduces the chi cost of Fists of Fury by 1, so it just counts the amount of Fists of Fury casts not made free by serenity.

sMteX: Oh wait, I misread the item's equip effect :D My bad, sorry ^^

src/Parser/WindwalkerMonk/Modules/Items/Katsuos.js
Juko8: Alright, done and all references should be updated too.

sMteX:
Nice, thank you, but you're still missing the `event.timestamp` parameter in `hasBuff`. You have to check if the player didn't have the buff at the time of the cast (without it it would work only if the player *didn't* have Serenity talent at all), but if they do have it selected, it would never count the Chi saved even if they didn't have the buff.

sMteX:
I personally would format the item output as "X Chi saved" as I think that's more natural way to put it but that's up to you, it's not blocking the PR. Thanks for the quick fixes :)

PR #532, opened 2017-10-15
[Windwalker] Updated AlwaysBeCasting

Juko8:
I tried using the `on_toPlayer_removebuff(event)` as seen in the Shadow Priest Module, but was not getting correct results from it.

Also, since Windwalker is an energy class with a fixed 1 second GCD, should all spells be listed as `STATIC_GCD_ABILITIES` instead of `ABILITIES_ON_GCD`?


MartijnHols:
Thanks!

PR #534, opened 2017-10-15
[Frost Mage] More Frost Legendary Damage Counts

Sharrq:
Damage Counts for Magtheridon's Banished Bracers and Shattered Fragments of Sindragosa.

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
+++ b/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
@@ -19,7 +19,7 @@ class ShatteredFragmentsOfSindragosa extends Module {
     if (event.ability.guid !== SPELLS.COMET_STORM_DAMAGE.id) {
       return;
     }
-      this.damage += event.amount;
+      this.damage += event.amount + (event.absorbed || 0);
   }
 
   item() {
--- a/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
+++ b/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
@@ -1,0 +1,33 @@
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class ShatteredFragmentsOfSindragosa extends Module {
+
+  static dependencies = {
+		combatants: Combatants,
+	};
+
+  damage = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasHead(ITEMS.SHATTERED_FRAGMENTS_OF_SINDRAGOSA.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.COMET_STORM_DAMAGE.id) {
+      return;
+    }
+      this.damage += event.amount;
+  }
+
+  item() {
+    return {
+      item: ITEMS.SHATTERED_FRAGMENTS_OF_SINDRAGOSA,
+      result: `${this.owner.formatItemDamageDone(this.damage)}`,
+    };
+  }
+}
+
+export default ShatteredFragmentsOfSindragosa;
--- a/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
+++ b/src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
@@ -30,7 +30,6 @@ class ShatteredFragmentsOfSindragosa extends Module {
       return;
     }
     this.legendaryProcs += 1;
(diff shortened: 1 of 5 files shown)


src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
```diff
@@ -0,0 +1,33 @@
...
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasHead(ITEMS.SHATTERED_FRAGMENTS_OF_SINDRAGOSA.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.COMET_STORM_DAMAGE.id) {
+      return;
+    }
+      this.damage += event.amount;
```
sMteX: You should add absorbed damage as well (Maiden, Avatar etc.) So `this.damage += event.amount + (event.absorbed || 0);`. Actually I'm never sure if it's absorb or absorbed, check to make sure.

Sharrq: Oops, forgot to add that. Should be added now

src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
```diff
@@ -0,0 +1,33 @@
...
+  damage = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasHead(ITEMS.SHATTERED_FRAGMENTS_OF_SINDRAGOSA.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.COMET_STORM_DAMAGE.id) {
+      return;
+    }
```
kfinch: This is trivially wrong if the player has taken Comet Storm (which is very uncommon, I know)

Sharrq: Yea, its even more rare for someone to be using Sindragosa and Comet Storm. Im not really sure how to exclude Comet Storm Damage that isnt related to the Legendary considering its the same ID (i checked)

kfinch: I can think of two approaches, I'll write them up when I'm back in front of computer

kfinch: Nevermind, 3 approaches:

1) The easy way out: check if player has Comet Storm, and if so either deactivate this module or replace the stat with "sorry, couldn't calculate because..."

2) Ignore Comet Storm damage that's within 1 second of a Comet Storm cast. This has the chance to clip procs that happen at same time as cast, but it's pretty good. You could even narrow that time window by closely inspecting logs to see the normal delay for a comet storm cast. 

3) Track hard cast count and use the legendaries buff to track proc count. Multiply total comet storm damage by the percentage of procs vs total.

sMteX:
Thanks for the changelog and also thanks @kfinch for the remark. I didn't realize Comet Storm might be a talent so that didn't occur to me that it could be a problem too.

src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
Sharrq: I opted for Number 3. Seemed like that was the simplest, best way to get a close approximation of the amount of damage that came from the legendary ... it wont be perfect, but it will be pretty close.

src/Parser/Mage/Frost/Modules/Items/ShatteredFragmentsOfSindragosa.js
```diff
@@ -0,0 +1,51 @@
...
+    if (event.ability.guid !== SPELLS.COMET_STORM_TALENT.id || !hasCometStormTalent) {
+      return;
+    }
+    this.cometStormCasts += 1;
+  }
+
+  on_toPlayer_applybuff(event) {
+    if (event.ability.guid !== SPELLS.RAGE_OF_THE_FROST_WYRM.id) {
+      return;
+    }
```
kfinch: Technically this should be `on_toPlayer_removebuff`, because it's when the player loses the buff that they consume it, but it's unlikely to ever make a difference.

Sharrq: The Helm actually uses two separate buffs, 

1. Shattered Fragments of Sindragosa, which stacks and counts the number of Frostbolts
2. Rage of the Frost Wyrm which triggers the Comet Storm.

When i was looking through the Events, I thought that Shattered Fragments stacked to 15, and then applied Rage of the Frost Wyrm which triggered the Comet Storm and then the buff was removed, but it appears that Shattered Fragments actually stacks to 14, removes itself and applies Rage of the Frost Wyrm, then the next Frostbolt removes Frost Wyrm and sends the Comet Storm. 

So long story short, ill change it.

PR #540, opened 2017-10-16
[Marksmanship] Tier 19 2p

Pewtro:
Adds a tracker for Tier 19 2p effectiveness if the hunter has that equipped.


src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,71 @@
...
+        .forEach(({ cost }) => {
+          const focusCost = cost || 0;
+          this.focusUpdates.unshift({
+            used: focusCost,
+          });
+        });
+    }
+    debug && console.log(`focusUpdates is now at: `, this.focusUpdates[0].used);
+    //1000ms/45focus = 22.2ms/focus
+    const COOLDOWN_REDUCTION_MS = 22.22222 * this.focusUpdates[0].used;
```
kfinch: Prefer this constant as a const above the class declaration, eg `const CDR_PER_FOCUS = 22.22222;`, and then here `const cdrMs = CDR_PER_FOCUS * this.focusUpdates[0].used;`

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,71 @@
...
+  a
+  item() {
+    return {
+      id: `spell-${SPELLS.HUNTER_MM_T19_2P_BONUS.id}`,
+      icon: <SpellIcon id={SPELLS.TRUESHOT.id} />,
+      title: <SpellLink id={SPELLS.HUNTER_MM_T19_2P_BONUS.id} />,
+      result: (
+        <dfn data-tip={`Your utilization of tier 19 2 piece: <br/> You effectively reduced Trueshots cooldown by: ${formatNumber(this.effectiveTrueshotReductionMs / 1000)} seconds.<br/> You wasted   ${formatNumber(this.wastedTrueshotReductionMs / 1000)} seconds of CDR.<br/> `}>
+          T19 2p CDR effectiveness: {formatPercentage(this.effectiveTrueshotReductionMs / (this.effectiveTrueshotReductionMs + this.wastedTrueshotReductionMs))}%
+        </dfn>
```
kfinch: This display has the potential to be very misleading. Item displays typically are intended to show what the legendary or set bonus gave you in additional throughput/effectiveness. If I read "T19 2p CDR effectiveness: 25%", I would naturally think "This set bonus effectively reduced the CD of my Trueshot by 25%". Of course that isn't **at all** what you're saying.

The stat you're actually displaying is like a worse version of cast efficiency. Let the cast efficiency module do what it's supposed to with regards to the additional CDR provided by this set. What you should be displaying here is the effective CDR applied to Trueshot.

Pewtro: Good point, fixing that in this and Ullrs then 👍

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+  item() {
+    return {
+      id: `spell-${SPELLS.HUNTER_MM_T19_2P_BONUS.id}`,
+      icon: <SpellIcon id={SPELLS.TRUESHOT.id} />,
+      title: <SpellLink id={SPELLS.HUNTER_MM_T19_2P_BONUS.id} />,
+      result: (
+        <dfn data-tip={`You wasted ${formatNumber(this.wastedTrueshotReductionMs / 1000)} seconds of CDR.<br/> `}>
+          You reduced <SpellLink id={SPELLS.TRUESHOT.id} /> by {formatNumber(this.effectiveTrueshotReductionMs / 1000)} seconds in total over the course of the fight.
```
kfinch: A minor wording thing, but this makes it sound like you reduced Trueshot rather than Trueshot's CD. It's also a bit wordy for a top level stat. Try: "Reduced Trueshot CD by X total seconds"

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,70 @@
...
+  wastedTrueshotReductionMs = 0;
+
+  focusUpdates = [];
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasBuff(SPELLS.HUNTER_MM_T19_2P_BONUS.id);
+  }
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.AIMED_SHOT.id && spellId !== SPELLS.PIERCING_SHOT_TALENT.id && spellId !== SPELLS.BURSTING_SHOT.id && spellId !== SPELLS.MARKED_SHOT.id && spellId !== SPELLS.WINDBURST.id) {
```
sMteX: Same comment I did in the Ullr's module applies here too.

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,70 @@
...
+      return;
+    }
+    if (event.classResources) {
+      event.classResources
+        .filter(resource => resource.type === RESOURCE_TYPES.FOCUS)
+        .forEach(({ cost }) => {
+          const focusCost = cost || 0;
+          this.focusUpdates.unshift({
```
sMteX: I'm trying to think about whether the array is really necessary. You alter it by `unshift`ing each affected spell's Focus cost, so on `[0]` you always have the most recent one, which you also use on lines 45 and 47, but you never really use the array any other way, like iterating through it, doing come math on it etc. Couldn't this be replaced by something like `this.lastFocusCost`?

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SpellUsable from 'Parser/Core/Modules/SpellUsable';
+import RESOURCE_TYPES from 'common/RESOURCE_TYPES';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import { formatNumber } from "../../../../common/format";
```
sMteX: This could be simple `'common/format'`.

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
Pewtro: It probably could, I can test a few things post raid - this was just a working solution 😊

sMteX: Sure thing, it's a good thing to have something that works :)

src/Parser/MarksmanshipHunter/Modules/Items/Tier19_2p.js
kfinch: Not 'could be', but 'should be'. I'm pretty sure you only do relative path for files inside spec, and use absolute path for common / utils

sMteX: Yeah, I think it happens when you allow IDE (at least WebStorm does it) to automatically (or more precisely, from the context menu made by Alt + Enter) import missing things.

PR #545, opened 2017-10-16
[Mage] Add Soul of the Archmage to items

avilene:
This is done so the player can see their legendaries; based off the Highlord one from paladin,

### DIFF
--- a/src/Parser/Mage/Fire/Modules/Items/SoulOfTheArchmage.js
+++ b/src/Parser/Mage/Fire/Modules/Items/SoulOfTheArchmage.js
@@ -1,0 +1,28 @@
+import React from 'react';
+
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class SoulOfTheArchmage extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  on_initialized() {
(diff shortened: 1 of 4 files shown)


src/Parser/Mage/Fire/Modules/Items/SoulOfTheArchmage.js
```diff
@@ -0,0 +1,28 @@
...
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class SoulOfTheArchmage extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_ARCHMAGE.id);
+    console.log(this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_ARCHMAGE.id));
```
sMteX: The log here is unnecessary (or prepend it with `debug &&`).

sMteX (changes requested):
Just the tiny log in Fire one, otherwise looks good. Thanks for the PR :)

kfinch:
These are fine as placeholders, but I may end up overwriting the Frost one when I implement Frozen Touch

src/Parser/Mage/Fire/Modules/Items/SoulOfTheArchmage.js
fyruna: Removed in 1c149be

fyruna:
@kfinch Obviously, but we should always make sure the player knows which legendaries they have; I would like to keep this around unless your Frozen Touch addition will actually include showing that the player has this equipped.

sMteX (approved):

PR #546, opened 2017-10-17
[mage] Added Rune of Power and Mirror Image modules.

kfinch:
Numbers for Fire won't be quite right due to interactions with Ignite, but it's fine for Frost.

Includes statistic informing player of the damage boost they got, suggestion if its below Incanter's Flow passive boost, and suggestion if player isn't standing in rune full duration.


kfinch:
Added MirrorImage as well, to Frost and Fire. Shows statistic with damage contribution, and a suggestion if its below Incanter's Flow passive boost.

src/Parser/Mage/Shared/Modules/Features/RuneOfPower.js
```diff
@@ -0,0 +1,79 @@
...
+    when(damageIncreasePercent).isLessThan(INCANTERS_FLOW_EXPECTED_BOOST)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>Your <SpellLink id={SPELLS.RUNE_OF_POWER_TALENT.id}/> damage boost is below the expected passive gain from <SpellLink id={SPELLS.INCANTERS_FLOW_TALENT.id}/>. Either find ways to make better use of the talent, or switch to <SpellLink id={SPELLS.INCANTERS_FLOW_TALENT.id}/>.</span>)
+          .icon(SPELLS.RUNE_OF_POWER_TALENT.icon)
+          .actual(`${formatPercentage(damageIncreasePercent)}% damage increase from Rune of Power`)
+          .recommended(`${formatPercentage(recommended)}% is the passive gain from Incanter's Flow`)
+          .regular(recommended).major(recommended-0.03);
+      });
+
+      const casts = this.abilityTracker.getAbility(SPELLS.RUNE_OF_POWER_TALENT.id).casts;
```
sMteX: If the Rune of Power is never cast, this would return `undefined` I believe (the object returned by `getAbility()` exists, but it doesn't have the `casts` property - https://github.com/WoWAnalyzer/WoWAnalyzer/blob/5b61270b21b0bd4c0dec0f05232c5238a023db72/src/Parser/Core/Modules/AbilityTracker.js#L23 , that one is set in `cast` event handler).

kfinch:
Fixed

sMteX:
My initial thought was to add `|| 0`, but returning and not showing the suggestion makes more sense (you would have to do `|| 1` in order not to divide by 0 and while that would still give correctly "average 0 seconds", it's ... weird to say the least :D).

kfinch:
Yeah I did the decision to do an early return because it made no sense to even go into that suggestion if you never cast RoP

PR #550, opened 2017-10-17
[core] Disabled DistanceMoved module

kfinch:
The numbers from the DistanceMoved module couldn't possibly be correct. I'm consistently seeing reports that the player moved many miles over the course of a few minute fight. This is happening across all logs I see.

This PR simply disables the module until we figure out what's wrong.


sMteX:
Although I don't know why we're dividing each distance by 100, if it wasn't there the distance would be even larger. My only guess is that the x and y properties of events are **not** measured in yards but something different and we would need to find out some conversion between the two.

kfinch:
Yeah I think this is a case of wrong conversion factor. I put in this PR so that we stop showing an embarrassingly wrong stat for every spec.

PR #556, opened 2017-10-17
[Windwalker / Mistweaver] Monk file structure

anom0ly:
Updated both Mistweaver and Windwalker spec modules to new file structure.

This PR should be merged before any additional changes to these 2 specs are merged.


src/Parser/Monk/Mistweaver/CHANGELOG.js
```diff
@@ -1,4 +1,5 @@
 export default `
+17-10-2017 - File structure updates for module. Should be no impact to end user (by anomoly)
```
kfinch: I thought the CHANGELOG files were supposed to be specifically for the benefit of the end user. This shouldn't have a changelog entry.

src/Parser/Monk/Windwalker/CHANGELOG.js
```diff
@@ -1,4 +1,5 @@
 export default `
+17-10-2017 - File structure updates for module. Should be no impact to end user (by anomoly)
```
kfinch: I thought the CHANGELOG files were supposed to be specifically for the benefit of the end user. This shouldn't have a changelog entry.

src/Parser/Monk/Mistweaver/CHANGELOG.js
anom0ly: Sounds good. I know I had added something to the MW last time we did a refactor, but I will remove.

src/Parser/Monk/Mistweaver/Modules/Traits/WhispersOfShaohao.js
```diff
@@ -4,7 +4,7 @@ import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 
 import Combatants from 'Parser/Core/Modules/Combatants';
-import SheilunsGift from 'Parser/MistweaverMonk/Modules/Spells/SheilunsGift';
+import SheilunsGift from 'Parser/Monk/Mistweaver/Modules/Spells/SheilunsGift';
```
kfinch: Why isn't this just relative path?

anom0ly: No reason - Updated to use relative path.

PR #559, opened 2017-10-17
[Windwalker] Added The Emperors Capacitor

Juko8:


src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -0,0 +1,80 @@
...
+  on_initialized() {
+    this.active = this.combatants.selected.hasChest(ITEMS.THE_EMPERORS_CAPACITOR.id);
+  }
+
+  on_byPlayer_applybuffstack(event) {
+    const spellId = event.ability.guid;
+    // You can only ever apply 1 stack at a time
+    if (SPELLS.THE_EMPERORS_CAPACITOR_STACK.id === spellId) {
+      this.totalStacks += 1;
+      if (this.currentStacks === 20) {
```
kfinch: Are you certain that an applybuffstack event is thrown when at max stacks?

src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -0,0 +1,80 @@
...
+      this.stacksUsed += this.currentStacks;
+      this.currentStacks = 0;
+      this.cjlCasts += 1;
+      this.averageStacksUsed = this.stacksUsed / this.cjlCasts;
+    }
+  }
+
+  item() {
+    const cracklingJadeLightning = this.abilityTracker.getAbility(SPELLS.CRACKLING_JADE_LIGHTNING.id);
+    const damage = cracklingJadeLightning.damageEffective;
```
kfinch: For whatever reason, `damageEffective` from the AbilityTracker is actually just amount, when you want amount + absorbed. You can do `const damage = cracklingJadeLightning.damageEffective + cracklingJadeLightning.damageAbsorbed;` Really, AbilityTracker should be fixed because this is misleading.

Also, `damageEffective` and `damageAbsorbed` will be undefined if you never cast CJL, so you might need to check for that.

Juko8: I can move the damagecalculation to a class variable and then calculate it in a `on_byPlayer_damage` check

src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
Juko8: It does not, and refreshbuff doesn't either. I have however changed it so that it now checks for chi spenders in casts and adds a wasted stack there if the cap has been reached

src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -0,0 +1,96 @@
...
+      this.cjlCasts += 1;
+      this.averageStacksUsed = this.stacksUsed / this.cjlCasts;
+    }
+  }
+
+  on_byPlayer_damage(event) {
+    const spellId = event.ability.guid;
+
+    if (spellId === SPELLS.CRACKLING_JADE_LIGHTNING.id) {
+      this.damage += event.amount;
```
kfinch: Remember that absorbed damage is effective damage, this should be `this.damage += event.amount + (event.absorbed || 0);`

src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -0,0 +1,96 @@
...
+      </dfn>),
+    };
+  }
+
+  suggestions(when) {
+    when(this.stacksWasted).isGreaterThan(0).addSuggestion((suggest, actual, recommended) => {
+      return suggest(<span> You wasted your <SpellLink id={SPELLS.THE_EMPERORS_CAPACITOR_STACK.id}/> stacks by using chi spenders while at 20 stacks </span>)
+        .icon(ITEMS.THE_EMPERORS_CAPACITOR.icon)
+        .actual(`${this.stacksWasted} Wasted stacks`)
+        .recommended(`Less than ${(recommended)} Wasted stacks is recommended`)
```
kfinch: This is going to say "Less than 0...."

kfinch:
Thanks!

PR #560, opened 2017-10-17
Fixed fights which had 0 voidform casts

hasseboulen:
Minor hotfix


hasseboulen:
I'm not quite sure if this does it.

PR #567, opened 2017-10-18
[Retribution] Added Tier 21

CollCrom:
Added in modules to generate results for both 2 and 4 piece of tier 21.

t21


MartijnHols:
I think the holy power gained should be shown as per minute so that its value can be more easily gauged.

CollCrom:
I think that this is a good suggestion and went ahead and changed all instances of items giving holy power to per minute. I left in the total holy power gained in a data tip.

src/Parser/RetributionPaladin/Modules/Items/LiadrinsFuryUnleashed.js
```diff
@@ -22,8 +21,10 @@ class LiadrinsFuryUnleashed extends Module {
 		const hpGained = this.holyPowerTracker.generatedAndWasted[SPELLS.LIADRINS_FURY_UNLEASHED_BUFF.id].generated;
 		return {
 			item: ITEMS.LIADRINS_FURY_UNLEASHED,
-			result: (
-				`${formatNumber(hpGained)} Holy Power gained from Liadrin's Fury Unleased.`
+			result: (<dfn data-tip={`
+				Total Holy Power Gained: ${formatNumber(hpGained)}`}>
+				{formatNumber(hpGained / this.owner.fightDuration * 60000)} Holy Power gained per minute from Liadrin's Fury Unleased.
```
anom0ly: More of a format thing. The portion `from Liadrin's Fury Unleased.` is potentially redundant because this is about the specific item.

src/Parser/RetributionPaladin/Modules/Items/Tier21_4set.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	item() {
+		return {
+			id: `spell-${SPELLS.RET_PALADIN_T21_4SET_BONUS.id}`,
+			icon: <SpellIcon id ={SPELLS.RET_PALADIN_T21_4SET_BONUS.id} />,
+			title: <SpellLink id={SPELLS.RET_PALADIN_T21_4SET_BONUS.id} />,
+			result : (<dfn data-tip={`
+				Total Holy Power Gained: ${formatNumber(this.holyPowerGained)}`}>
+				{formatNumber(this.holyPowerGained / this.owner.fightDuration * 60000)} Holy Power gained per minute from Tier 21 4 peice.
```
anom0ly: Similar comment to the above.  `from Tier 21 4 peice` is also slightly redundant.

src/Parser/RetributionPaladin/Modules/Items/Tier21_4set.js
```diff
@@ -0,0 +1,44 @@
...
+	on_initialized(){
+		this.active = this.combatants.selected.hasBuff(SPELLS.RET_PALADIN_T21_4SET_BONUS.id);
+	}
+
+	on_byPlayer_removebuff(event){
+		const spellId = event.ability.guid;
+		if(spellId !== SPELLS.RET_PALADIN_T21_4SET_BONUS_BUFF.id){
+			return;
+		}
+		this.holyPowerGained += 1;
```
anom0ly: Just to confirm the interaction here, the HP Gain is triggered on the debuff being removed.

CollCrom: Hmm not technically and I just thought about some cases where this method wouldn't work. I'll fix it

anom0ly: I was just trying to understand the interaction. I know nothing about Ret, so if there aren't edge cases where you never drop this buff this should be fine.

CollCrom: http://www.wowhead.com/spell=251868/item-paladin-t21-retribution-4p-bonus-judgment-reduces-spender-cost

Actually I was wrong and this method works. Since there is no duration on this buff every time the buff is removed you have cast a HP spender and net 1 holy power. At least according to the information on wowhead.

CollCrom: If you think it would make more sense I can check for the cast of a HP spender and check for the buff and then increment the hp gained

MartijnHols:
Could you update from master and resolve the conflicts please? It should be a minor fix; move what you added to SPELLS_PALADIN to SPELLS/PALADIN

PR #525, opened 2017-10-14
[Marksmanship] Patient Sniper tab and statistic

Pewtro:
Adds a statistic box with average damage gain from Patient Sniper: 
image

Aswell as a tab with significantly more detailed information:
image

~~Also moves folders around to adjust for new Class/Spec structure~~
Also deletes ProcTracker as it is currently not used at all, and was just a copy of some shaman code.

PR #575, opened 2017-10-20
[Marksmanship] Patient Sniper module

sMteX:
The original PR was #525 from @Pewtro, to which this PR is attributed in the end. I thought his original PR was quite unclear and repetitive and we've agreed that I could refactor it. Refactoring was quite a success but we've found out there's more intricate details to implement in order to be correct so the logic is more complicated than originally. I tried to explain my reasoning with comments in code but feel free to ask any additional questions. 

EDIT: I forgot - the suggestion thresholds are currently incorrect (they're left from previous implementation). As I expect this PR will probably stay here for a while, I asked Putro if he can come up with some feasible numbers so I can update this with new values, or he can update them in another PR.

EDIT2: Images can be found on original PR - only wording changed and a new column at the far right of the table.


src/Parser/Hunter/Marksmanship/Modules/Talents/PatientSniper/PatientSniperDetails.js
```diff
@@ -0,0 +1,97 @@
...
+    return this.patientSniperTracker.patientSniper[SPELLS.AIMED_SHOT.id].bonusDmg;
+  }
+  get bonusPiercingDamage() {
+    return this.patientSniperTracker.patientSniper[SPELLS.PIERCING_SHOT_TALENT.id].bonusDmg;
+  }
+  get bonusDamage() {
+    return this.bonusAimedDamage + this.bonusPiercingDamage;
+  }
+
+  suggestions(when) {
```
Pewtro: This should be updated to be 0.09, 0.08, 0.07 for minor, average and major as I mentioned, looking at some top logs they seem hover between 9 and 10% increased damage, and the moment it starts dropping lower it indicates a wrong rotation - as PS damage should be a damage increase from purely playing rotationally correctly (apply vulnerable --> generate focus --> aimed shotx2 late in the vuln window) 

These values are 'only' accurate on more ST focused encounters, but the moment we enter AoE , it becomes way less important to care about Patient Sniper, which then leads to automatically utilising it less - however I still think the recommended values should be for ST encounters, possibly include the fact that it may look bad on AoE, it could be because Patient Sniper isn't as important there.

src/Parser/Hunter/Marksmanship/Modules/Talents/PatientSniper/PatientSniperDetails.js
```diff
@@ -0,0 +1,97 @@
...
+
+  statistic() {
+    let tooltipText = `Your Aimed Shots ${this.hasPiercingShot ? 'and Piercing Shots ' : ''}did ${formatNumber(this.bonusDamage)} (${this.owner.formatItemDamageDone(this.bonusDamage)}) bonus damage thanks to Patient Sniper talent. Below you'll see them individually, and if you want to see more Patient Sniper information (such as without Trueshot windows), please check the "Patient Sniper Usage" tab in the menu. <br />`;
+    const aimed = this.abilityTracker.getAbility(SPELLS.AIMED_SHOT.id);
+    tooltipText += `Aimed Shot bonus damage: ${formatNumber(this.bonusAimedDamage)} (${formatPercentage(this.bonusAimedDamage / aimed.damageEffective)} %)`;
+    if (this.hasPiercingShot) {
+      const piercing = this.abilityTracker.getAbility(SPELLS.PIERCING_SHOT_TALENT.id);
+      tooltipText += `<br />Piercing Shot bonus damage: ${formatNumber(this.bonusPiercingDamage)} (${formatPercentage(this.bonusPiercingDamage / piercing.damageEffective)} %)`;
+    }
+    return (
```
Pewtro: Shouldn't this tooltipText be for both Piercing AND Aimed ? or are they still together if you have piercing shot selected?

sMteX: That's why there's tooltipText += ..., it adds the Piercing Shot part if you have it talented but the aimed part stays there regardless. I'll add screenshots later.

src/Parser/Hunter/Marksmanship/Modules/Talents/PatientSniper/PatientSniperDetails.js
sMteX: Ok I'll update the thresholds when I can :-)

src/Parser/Hunter/Marksmanship/Modules/Talents/PatientSniper/PatientSniperDetails.js
sMteX: With Piercing Shot:
image


Without Piercing Shot:
image

PR #582, opened 2017-10-20
[MMS Hunter] Displays Medallion casts paired with Trueshot

leapis:
Very short, just displays the number of medallion casts that were cast during Trueshot against the total number of Medallion casts as a tooltip. This is open for more updates once we implement a robust bullseye tracker, but for now it just displays paired casts.
A TS cast is considered "paired" if it is completely within the cast of medallion, with leeway of one second. Adding more detail/suggestions on incomplete pairings is possible, but not a very high priority at the moment. I'll probably flesh out a lot more medallion interactions once Bullseye is finished. Or maybe I won't get to it if Antorus renders Medallion obsolete.

Screenshots:
image


MartijnHols:
Has conflicts so can't merge yet

leapis:
Should be ready to go now- Putro wanted 10s of medallion to be during TS for it to count as "paired", which it now has (5 seconds of "leeway" means that TS can be cast up to 5 seconds before or be cast so that it ends 5 seconds after, since TS is 15 sec this means that there will be at least 10 seconds of TS and Medallion at the same time).

PR #583, opened 2017-10-21
[Marksmanship] Added support for Lock and Load and True Aim talents

Pewtro:


src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -0,0 +1,87 @@
...
+
+  statistic() {
+    const expectedProcs = Math.round(this.autoshots * 0.08);
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.LOCK_AND_LOAD_TALENT.id} />}
+        value={`${formatPercentage(this.totalProcs / expectedProcs, 1)}%`}
+        label={`of expected procs`}
+        tooltip={`You had a total of ${this.totalProcs} procs, and your expected amount of procs was ${expectedProcs}. <br /> You had ${this.noGainLNLProcs} with 2 lnl stacks remaining and ${this.halfLNLProcs} while you had 1 stack remaining. This resulted in you losing out on ${this.wastedInstants} instant casts of Aimed Shot due to overlapping.`}
+      />
```
sMteX: I suggest showing `wastedInstants` as the number in the statistic box. Seeing a percentage of "expected procs" doesn't seem very useful on the first look (as you said, it's a nice way to know if I was just unlucky or something, but I'd leave it into the tooltip instead). Also, I suggest adding "procs" after `${this.noGainLNLProcs}` and `${this.halfGainLNLProcs}` - "You had 2 with 2 LNL stacks remaining..." - I had 2 *what*?

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -0,0 +1,87 @@
...
+class LockAndLoad extends Module {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  fullLNLProcs = 0;
+  halfLNLProcs = 0;
+  noGainLNLProcs = 0;
+  totalProcs = 0;
+  autoshots = 0;
```
sMteX: Should be camelCased = `autoShots`.

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,73 @@
...
+    if (debuffId !== SPELLS.TRUE_AIM_DEBUFF.id) {
+      return;
+    }
+    //I can't use on_byPlayer_applydebuff as it won't work in conjunction with on_byplayer_applydebuffstack
+    //so we do it this way, and just make a check if _currentStacks is 0, in which case we set it to 1 to indicate it already has been applied.
+    if (this._currentStacks === 0) {
+      this._currentStacks = 1;
+    }
+    this._currentStacks += 1;
+    if (this._currentStacks === 10) {
```
sMteX: You could extract it as a constant before class like `MAX_STACKS`.

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,73 @@
...
+      this.startOfMaxStacks = event.timestamp;
+    }
+  }
+  on_byPlayer_removedebuff(event) {
+    const debuffId = event.ability.guid;
+    if (debuffId !== SPELLS.TRUE_AIM_DEBUFF.id) {
+      return;
+    }
+    if (this._currentStacks === 10) {
```
sMteX: Same here.

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,73 @@
...
+      this.timeAtMaxStacks += event.timestamp - this.startOfMaxStacks;
+    }
+    //ensures Trickshot cleaving doesn't count as multiple resets of stacks
+    if (this._currentStacks > 2) {
+      this.timesDropped += 1;
+    }
+    this._currentStacks = 0;
+  }
```
sMteX: Seems to me like it's missing adding to `timesDropped` if there were less than 3 stacks of the debuff (then it just resets the stacks).

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,73 @@
...
+  on_byPlayer_applydebuffstack(event) {
+    const debuffId = event.ability.guid;
+    if (debuffId !== SPELLS.TRUE_AIM_DEBUFF.id) {
+      return;
+    }
+    //I can't use on_byPlayer_applydebuff as it won't work in conjunction with on_byplayer_applydebuffstack
+    //so we do it this way, and just make a check if _currentStacks is 0, in which case we set it to 1 to indicate it already has been applied.
+    if (this._currentStacks === 0) {
+      this._currentStacks = 1;
+    }
```
sMteX: Interesting. I guess this would work but why it wouldn't work with applydebuffstack? Just out of curiosity.

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
Pewtro: Take this example: 
image

There are only 4 meaningful drops (one on switching to maiden first time, one on back to avatar), then once gain once onto maiden and once onto Avatar. If I just let it accept all "removedebuff" events, it'll rocket up to 18 resets - this is due to trickshot cleaving applying a debuff on main target, then it'd cleave to target 2 and apply the debuff and thus remove it from the initial target. Sometimes you'll be casting 1 arcane shot, thus apply an extra true aim debuff, before it knocking it off when trick cleaving. 

I'll add a bit of information in the tooltip to clarify that:
 ``You reset your stacks (when they were at 3 stacks or higher) ${this.timesDropped} times over the course of the encounter. <br />Your total amount of resets (including with trickshot cleaving) was: ${this.totalTimesDropped}``
 - does this sound better?

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
Pewtro: Because they happen in the "wrong" order for that to be possible: 
image

I see I wrote "in conjunction with on_byplayer_applydebuffstack" , but I actually meant "in conjunction with on_byplayer_removedebuff"

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,80 @@
...
+      this.timesDropped += 1;
+    }
+    this.totalTimesDropped += 1;
+    this._currentStacks = 0;
+  }
+
+  statistic() {
+    const percentTimeAtMaxTAStacks = formatPercentage(this.timeAtMaxStacks / this.owner.fightDuration);
+    //only shows if timesDropped is greater than 0, to only show it on encounters where it might matter - we expect full uptime on any ST boss.
+    if (this.timesDropped > 0) {
```
sMteX: Although I get what you mean by this, I think Zero said that you should always show statistic boxes even if the result was in a sense "perfect" (100% uptime on something, 0 <resource> wasted etc.).

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+
+import Module from 'Parser/Core/Module';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Enemies from 'Parser/Core/Modules/Enemies';
```
sMteX: Enemies import is unnecessary - it's not used.

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -0,0 +1,95 @@
...
+    combatants: Combatants,
+  };
+
+  fullLNLProcs = 0;
+  halfLNLProcs = 0;
+  noGainLNLProcs = 0;
+  totalProcs = 0;
+  autoShots = 0;
+  wastedInstants = 0;
+  _curruntStacks = 0;
```
sMteX: Typo here 😅

src/Parser/Hunter/Marksmanship/Modules/Talents/TrueAim.js
Pewtro: Fair enough - guess it'll get a Amazing! statistic when that issue is created 👍

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
Pewtro: 🤦‍♂️

PR #591, opened 2017-10-21
Arms warrior - fervor of battle update, cast efficiency update and spells added

BadBossy:
Was neccessary due talent issue's with uptime of colossus smash


BadBossy:
Yea, you are right, I can do the easier way of course, gonna rework it! thanks

PR #599, opened 2017-10-23
[Disc Priest] Atonement reordering issues

karlpralow:
This PR contains only the reordering when an atonement healing event occurs between 2 succesive damaging events because healing on self has no latency


src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
```diff
@@ -0,0 +1,30 @@
+import Module from 'Parser/Core/Module';
+import isAtonement from '../Core/isAtonement';
+
+class AtonementAttribution extends Module {
+
+  reorderEvents(events) {
+    // Push down atonement healing on self to the next atonement healing
```
MartijnHols: Why?

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
```diff
@@ -0,0 +1,30 @@
...
+    let movesToMake = [];
+
+    for(let i = 0; i < events.length; i++){
```
MartijnHols: please change this to a forEach;

```jsx
events.forEach((event, index) => {
  [rest of your code]
});
```

MartijnHols: Also I think if you name `event` and `index` differently you might get more understandable code. These are the self-healing Atonements right? So how about `atonementSelfHeal` or `atonementSelfHealEvent`?

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
```diff
@@ -0,0 +1,30 @@
...
+      if(events[i].type === "heal" && isAtonement(events[i]) && events[i].sourceID == events[i].targetID) {
+        // Find next atonement healing and move it right before it
+        for(let j = i + 1; j < events.length; j++) {
+          if(events[j].type == "heal" && isAtonement(events[j])) {
+            movesToMake.push({"From": i, "To": j - 1})
+            break;
```
MartijnHols: Ok so if I understand this correctly it moves (=prepares the instruction to move) Atonement self heal to the next Atonement? I think this is fragile and may break, such as if there's only 1 atonement active; on self: damage 1 -> damage 2 -> self heal 1 -> self heal 2 would reorder self heal 1 after 2, incorrectly. How about just breaking the loop if the first atonement heal found is a self heal?

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
```diff
@@ -0,0 +1,30 @@
...
+          }
+        }
+      }
+    }
+
+    for(let i = 0; i < movesToMake.length;i++){
+      events.splice(movesToMake[i].To, 0, events.splice(movesToMake[i].From, 1)[0]);
```
MartijnHols: I think this becomes unnecessary if you change the first for-loop to use `forEach`.

MartijnHols: I think the second splice should be put in a temp var so it can be clarified through proper var naming what it is.

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
```diff
@@ -0,0 +1,30 @@
...
+    atonementAtribution = new AtonementAttribution({
+      reorder: () => true,
+      toPlayer: () => true,
+      byPlayer: () => true,
+      toPlayerPet: () => false,
+      byPlayerPet: () => false,
+    });
+  });
+
+  it('Two Succesive Damaging Events with atonement on self between is reordered', () => {
```
MartijnHols: It's unclear to me what the end-goal is from this. Would this be right: `the Atonement self-healing triggered by two successive damaging events gets grouped with the rest of the Atonement healing`?

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
```diff
@@ -0,0 +1,30 @@
...
+    let result = atonementAtribution.reorderEvents(TwoSuccesiveDamagingEventsWithAtonementOnSelfBetween);
```
MartijnHols: I'd much rather see the actual array here so that I can assess what the test does without having to jump between files.

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(result[2].type).toBe("damage");
```
MartijnHols: This is pretty hard to understand. If you move the actual array to this method as proposed in the previous comment, would it be an idea to extract the interesting events from the array into temp variables and check those instead?

e.g.

```jsx
expect(result[2]).toBe(firstDamageEvent);
expect(result[3]).toBe(firstSelfHealing);
```

MartijnHols: Actually best case might be to move all events to temp vars and just check a new array. Example:

```jsx
const firstAtonementSelfHealEvent = { type: 'heal', sourceID: thisPlayer, targetID: otherPlayer1, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.ATONEMENT_HEAL_NON_CRIT.id } };
const firstDamagingEvent = { type: 'damage', sourceID: thisPlayer, targetID: enemy, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.PURGE_THE_WICKED_BUFF.id } };
const secondAtonementSelfHealEvent = { type: 'heal', sourceID: thisPlayer, targetID: thisPlayer, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.ATONEMENT_HEAL_NON_CRIT.id } };
const secondDamagingEvent = { type: 'damage', sourceID: thisPlayer, targetID: enemy, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.SHADOWBIND_TRAIT.id } };
const firstAtonementAllyHealEvent = { type: 'heal', sourceID: thisPlayer, targetID: otherPlayer1, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.ATONEMENT_HEAL_NON_CRIT.id } };
const secondAtonementAllyHealEvent = { type: 'heal', sourceID: thisPlayer, targetID: otherPlayer2, amount: amount, timestamp: timestamp, ability: { guid: SPELLS.ATONEMENT_HEAL_NON_CRIT.id } };

const originalEvents = [
  firstAtonementSelfHealEvent,
  firstDamagingEvent,
  secondAtonementSelfHealEvent,
  secondDamagingEvent,
  firstAtonementAllyHealEvent,
  secondAtonementAllyHealEvent,
];

expect(atonementAtribution.reorderEvents(TwoSuccesiveDamagingEventsWithAtonementOnSelfBetween)).toEqual([
  firstDamagingEvent,
  secondDamagingEvent,
  firstAtonementSelfHealEvent,
  firstAtonementAllyHealEvent,
  secondAtonementSelfHealEvent,
  secondAtonementAllyHealEvent,
]);

```

I reckon this makes the expected input -> transformation -> output the easiest to see.

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
MartijnHols: Tip: write a test for this situation before fixing this bug to ensure it's actually an issue.

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
karlpralow: The idea behind this is, right now the way atonement attribution is made is by simply associating a damage event to the next atonement events until it hits another damaging event. With the latency on self heal, if 2 damaging events occur simultaneously, the healing event of the first damaging event will heal the player before the 2nd damage event hits. Pushing it down groups it with the associated event. We can then handle the double damaging event issue

MartijnHols: Could you put that in a comment in the method please?

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
karlpralow: "damage 1 -> damage 2 -> self heal 1 -> self heal 2 would reorder self heal 1 after 2, incorrectly"
It woudn't, it would hit the next atonement, and place it right before it. (It moves from i (the self ato) to j - 1 (j behing the next atonement event, so j - 1 is right before it)

I know TDD, I actually did do the test beforehand :P

MartijnHols: In that case could you prove this with a test? 😇

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
karlpralow: Correct

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
karlpralow: Agreed

src/tests/Parser/DisciplinePriest/Features/AtonementAttribution.test.js
karlpralow: I agree 100%, will change it shortly

src/Parser/DisciplinePriest/Modules/Features/AtonementAttribution.js
karlpralow: Ok

karlpralow:
This module is trying to fix the atonement attribution problem. Right now how logs work, it associates all atonement healing events to the last damaging event. In many cases this is wrong. Because of latency and damaging events that occur very close together, this gets mixed because of the ordering of the events. My goal with this module is to make the general association to last damaging event reliable. This is very important for disc to fix this since we can't get any accurate values on multiple things like Crucible traits and such.

MartijnHols:
I see you're putting multiple reorder methods into the same class. I think you should create a different class (and perhaps PR) for each approach you wish to add, would make things cleaner. Especially with the new Normalizers setup (see existing event fixers).

PR #600, opened 2017-10-23
[Frost Mage] Thermal Void and Winters Chill Fixes

Sharrq:
Adjusted Thermal Void to exclude the uptime and cast for Icy Veins if it does not complete before the fight ends as it skews the average in a negative way.

Also adjusted Winters Chill to account for "Double Ice Lance" Hits within the Winter's Chill Window (which was causing this number to go negative or potentially hide a missed window).... This also counts the number of times the player managed to get 2 Ice Lance's in the window, but i havent decided how i want to use that data yet ... so for now it counts it and doesnt do anything further with it. Ill do something with that data at some point.

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
+++ b/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
@@ -22,18 +22,11 @@ class ThermalVoid extends Module {
   }
 
   on_toPlayer_applybuff(event) {
-    if(event.ability.guid === SPELLS.ICY_VEINS.id && event.prepull) { // catch prepull cast by buff being present on pull
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.ICY_VEINS.id) {
       this.casts += 1;
       this.buffApplied = event.timestamp;
     }
-  }
-
-  on_byPlayer_cast(event) {
-    if (event.ability.guid !== SPELLS.ICY_VEINS.id) {
-      return;
-    }
-    this.casts += 1;
-    this.buffApplied = event.timestamp;
   }
 
   on_finished() {
--- a/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
+++ b/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
@@ -14,29 +14,38 @@ class ThermalVoid extends Module {
 	}
 
   casts = 0;
+  buffApplied = 0;
+  extraUptime = 0;
 
   on_initialized() {
 	   this.active = this.combatants.selected.hasTalent(SPELLS.THERMAL_VOID_TALENT.id);
   }
 
   on_toPlayer_applybuff(event) {
-    const spellId = event.ability.guid;
-    if(spellId === SPELLS.ICY_VEINS.id && event.prepull) { // catch prepull cast by buff being present on pull
+    if(event.ability.guid === SPELLS.ICY_VEINS.id && event.prepull) { // catch prepull cast by buff being present on pull
       this.casts += 1;
+      this.buffApplied = event.timestamp;
     }
   }
 
   on_byPlayer_cast(event) {
-    const spellId = event.ability.guid;
-    if (spellId !== SPELLS.ICY_VEINS.id) {
+    if (event.ability.guid !== SPELLS.ICY_VEINS.id) {
       return;
     }
     this.casts += 1;
+    this.buffApplied = event.timestamp;
+  }
+
+  on_finished() {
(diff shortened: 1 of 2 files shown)


src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
```diff
@@ -14,29 +14,38 @@ class ThermalVoid extends Module {
...
   }
 
   on_byPlayer_cast(event) {
-    const spellId = event.ability.guid;
-    if (spellId !== SPELLS.ICY_VEINS.id) {
+    if (event.ability.guid !== SPELLS.ICY_VEINS.id) {
       return;
     }
     this.casts += 1;
+    this.buffApplied = event.timestamp;
```
kfinch: Rather than having one case for precasts with on_byPlayer_applybuff && prepull, and another for regular casts with on_byPlayer_cast, you could and probably should just catch both with on_byPlayer_applybuff.

src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
```diff
@@ -14,29 +14,38 @@ class ThermalVoid extends Module {
...
   }
 
-  suggestions(when) {
-    const averageDuration = (this.combatants.selected.getBuffUptime(SPELLS.ICY_VEINS.id) / 1000) / this.casts;
+  on_finished() {
+    if (this.combatants.selected.hasBuff(SPELLS.ICY_VEINS.id)) {
+      this.casts -= 1;
+      this.extraUptime = this.owner.currentTimestamp - this.buffApplied;
```
anom0ly: Is there a reason to remove the cast and not just show this as a shortened / clipped IV cast?

From an outsider perspective it seems like the cast in this case should count, however, tracking an additional metric of a clipped cast may make more sense.

kfinch: Yeah, it's because average Icy Veins uptime is an important frost mage performance metric (it works a bit similar to spriests void form, but is available much less frequently). If the fight ends with IV still up, it may come out as a too short uptime, even though this is not the player's fault, and it can throw off their overall metric. For this reason, we want to avoid including IV that is still up when fight ends.

anom0ly: Gotcha. Sounds good. Just wanted to ask the question. 😃

src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
Sharrq: I think you were the one that added the prepull stuff, so im not quite sure how that prepull event works

kfinch: I may have added the extra handling, yeah. I'll fix it myself later. Nothing will work wrong, just noticing a cleaner solution.

Sharrq: Yea i have no problem fixing it, just not sure how to combine them and make sure the pre pull works .... plus i dont know if i have a log with a prepull Icy Veins in it, so i dont know that i could even test it

src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
```diff
@@ -14,29 +14,38 @@ class ThermalVoid extends Module {
...
+  extraUptime = 0;
 
   on_initialized() {
 	   this.active = this.combatants.selected.hasTalent(SPELLS.THERMAL_VOID_TALENT.id);
   }
 
   on_toPlayer_applybuff(event) {
-    const spellId = event.ability.guid;
-    if(spellId === SPELLS.ICY_VEINS.id && event.prepull) { // catch prepull cast by buff being present on pull
+    if(event.ability.guid === SPELLS.ICY_VEINS.id && event.prepull) { // catch prepull cast by buff being present on pull
```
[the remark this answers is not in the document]
Sharrq: Ill fix those. I removed them because it seemed cleaner to not split it out like that and the single line without the constant is what i did in my other modules, so i was changing this one to match those. But ill change it back if thats the preferred way to do it ... and then ill change my other modules back the next time i modify them so they are standard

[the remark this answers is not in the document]
anom0ly: Thanks. I know for some of these smaller modules it seems potentially like an unneeded step, but just to keep the style consistent I would update. And that is fine for the other stuff, just slowly convert it over.

Sharrq:
All this should be fixed now

PR #603, opened 2017-10-23
[Windwalker] Added Fists of Fury tick tracking

Juko8:


src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
```diff
@@ -0,0 +1,67 @@
...
+            this.previousTickTimestamp = event.timestamp;
+            this.averageTicks = this.fistsTickNumber / this.fistsCastNumber;
+        }
+    }
+
+    suggestions(when) {
+        when(this.averageTicks).isLessThan(5).addSuggestion((suggest, actual, recommended) => {
+            return suggest(<span> You are cancelling your <SpellLink id={SPELLS.FISTS_OF_FURY_CAST.id} /> casts early and losing ticks </span>)
+                .icon(SPELLS.FISTS_OF_FURY_CAST.icon).actual(`${this.averageTicks.toFixed(2)} average ticks on each Fists of Fury cast`)
+                .recommended(`Aim to get 5 ticks with each Fists of Fury cast (not always true with T20)`)
```
anom0ly: My understanding is that T20 changes the playstyle to cancel FoF whenever you get a proc. If this is the case, would it be better to only have this suggestion if the player doesn't have T20 equiped?

src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
```diff
@@ -0,0 +1,67 @@
...
+    fistsCastNumber = 0;
+    averageTicks = 0;
+
+    isNewFistsTick(timestamp) {
+        return !this.previousTickTimestamp || (timestamp - this.previousTickTimestamp) > FISTS_OF_FURY_MINIMUM_TICK_TIME;
+    }
+
+    on_byPlayer_cast(event) {
+        const spellId = event.ability.guid;
+        if (spellId === SPELLS.FISTS_OF_FURY_CAST.id) {
```
anom0ly: More of a code style,it would potentially easier to read if you moved the logic out of the `if` statement and wrote the statement like `if(spellId !== SPELLS.FIST_OF_FURY_CAST.id) { return; }` This just cleans up the code below if additional logic is added.

src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
```diff
@@ -0,0 +1,67 @@
...
+            this.fistsCastNumber += 1;
+            // average ticks is calculated here in case you don't hit any ticks during a cast'
+            this.averageTicks = this.fistsTickNumber / this.fistsCastNumber;
+        }
+    }
+
+    on_byPlayer_damage(event) {
+        if (event.ability.guid === SPELLS.FISTS_OF_FURY_DAMAGE.id && this.isNewFistsTick(event.timestamp)) {
```
anom0ly: Same as the above question / comment.

anom0ly:
Thanks!

PR #604, opened 2017-10-23
Resto Druid Stat Weights

kfinch:
Added Resto Druid Stat Weight generator, and updated some core rendering things to include new panel rendering options.

Stat Weights: Generates stat weights based on actual talents / legendaries / healing done. There are many improvements I'd like to make to this module, but it's now in the stat where it can go live. It's currently at least as good as the other Resto Druid stat weight tools. While live stat values would be very helpful to increase accuracy of this module, I'm pushing this first. I've left many TODOs to swap out code once current stat values are implemented. I've hacked together temporary handling for Concordance and 2t19 because they're the biggest deal procs currently.

Core Updates: `item()` now accepts a raw React element, which will always be rendered after everything else. Module gets a new `extraPanel()` function, which takes a raw React element and renders it beneath the items panel.


src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,256 @@
...
+  totalOneInt = 0;
+  totalOneCrit = 0;
+  totalOneHasteHpm = 0;
+//  totalOneHasteHpct = 0;
+  totalOneMastery = 0;
+  totalOneVers = 0;
+  totalOneLeech = 0;
+
+  on_initialized() {
+    this.concordanceAmount = 4000 + (this.combatants.selected.traitsBySpellId[SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_TRAIT.id] || 0) * 300; // TODO remove when stat tracker added
```
MartijnHols: Doesn't this include 1 point too many because counting the first point twice?

kfinch:
544cfeaa-3b5d-4ba5-b926-8b8ffe92ff0c

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,256 @@
...
+      oneInt = amount * bonusFromOneInt;
+    }
+    this.totalOneInt += oneInt;
+
+    // CRIT //
+    let oneCrit = 0;
+    if(spellInfo.crit) {
+      const currCritPerc = this.combatants.selected.critPercentage; // TODO replace when dynamic stats
+      const bonusFromOneCrit = 1 / 40000; // TODO replace when stat constants exist
+      if(spellInfo.isLivingSeed) {
```
MartijnHols: Adding a special bool for this feels like adding unnecessary complexity to me. How about just checking the spell id if it's living seed? You have to do special handling anyway, so moving all this special handling to here seems fine to me.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,256 @@
...
+    if(healVal.overheal) {
+      return; // if a spell overheals, we know it couldn't have healed for more
+    }
+
+    const hotCount = target.activeBuffs()
+        .map(buffObj => buffObj.ability.guid)
+        .filter(buffId => getSpellInfo(buffId).masteryStack)
+        .length;
+
+    // INT //
```
MartijnHols: You could also use regions;

```
// region Intellect
[code]
// endregion
```

src/Parser/RestoDruid/SpellInfo.js
```diff
@@ -0,0 +1,247 @@
...
+    int: true,
+    crit: true,
+    hasteHpm: true,
+    hasteHpct: false,
+    mastery: true,
+    masteryStack: true,
+    vers: true,
+  },
+  [SPELLS.LEECH.id]: { // procs a percent of all your healing, so we ignore for weights and total healing
+    isLeech: true,
```
MartijnHols: see `isLivingSeed` comment

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Yeah you're right, I'll fix it.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Do those do anything special when inspecting code? Unfamiliar with them.

MartijnHols: Most decent editors make them collapsible. You could also add a block around the code by simply inserting { and } around it, also makes start and end easier to spot.

The best thing would be to extract to new methods.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: My thought was to avoid directly using the event and offload everything to the spellInfo, but you're right that in this case that may be stupid.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: They originally were their own methods, but there was a lot of changing method signatures and these were also methods that were going to be called from exactly one place. I think I'll go with the regions, to be honest.

MartijnHols: `these were also methods that were going to be called from exactly one place` that is not a bad thing. Readability and maintainability are very important.

kfinch: I started separating them out into their own functions and it really made it worse because of all the fields needed to be passed, repeated calculation, etc. I've cleaned up the code in `_handleHeal` and made them regions, I think it looks much better now.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Changed as recommended.

src/Parser/RestoDruid/SpellInfo.js
kfinch: Changed as recommended.

kfinch:
All issues addressed

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * theorize that it might lead to overhealing. We're more strict on heals that overheal: they're disregarded entirely even
+ * if they're only partial overhealing. This is because a spell that partially overheals will do the same amount of effective
+ * healing regardless of the raw strength of the heal. This approach can cause some possible weirdness. For example, consider
+ * a situation where there is a heal for 500 (0 overheal) followed immediately by a heal for 400 (+ 100 overheal). 500 of this
+ * healing would be counted towards stat weights. Now consider a situation where one heal does 900 (+ 100 overheal). Functionally,
+ * this situation is identical, but in this case none of the healing is counted towards stat weights. Counting all effective
+ * healing, even heals that partially overheal, fixes this issue and is overall a valid approach, but still not one I will be taking.
+ * I think an advantage of disregarding all partial overheals is that it gives "top off" heals an effective lower weight than
+ * "life saver" heals. Still, this is a decision I will revisit.
```
buimichael: Can't you get the targets currentHp and maxHp and calculate if the extra stat would result in overhealing or actual healing?

src/Parser/RestoDruid/SpellInfo.js
```diff
@@ -0,0 +1,245 @@
...
+    vers: true,
+  },
+  [SPELLS.LEECH.id]: { // procs a percent of all your healing, so we ignore for weights and total healing
+    multiplier: true,
+  },
+  [SPELLS.VELENS_FUTURE_SIGHT.id]: { // while active procs from any healing, so we ignore for weights and total healing
+    multiplier: true,
+  },
+};
+
```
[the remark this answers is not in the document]
kfinch: Note that the default returns vers:true and everything else false, so most of these handled already.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * The general formula is critChanceFromOneRating * healAmount, but we have to make sure to account for anything that makes a
+ * crit stronger than 2x (e.g. Drape of Shame), and we also have to account for if healAmount is already a crit. If the heal
+ * is already a crit then we double count the weight, so we first have to normalize it to the amount it would have done
+ * had it not crit. Special handling is needed for Living Seed obviously. We assume that higher crit will result in proportionally
+ * more Living Seeds, and so get the bonus to Living Seed as critChangeFromOneRating / critChance.
+ *
+ * Haste -
+ * Calculating the benefit to healing from being able to cast faster is a huge pain in the ass, particularly for Resto Druids
+ * who aren't very dependant on cast times. However, HoTs tick faster and effectively do increased healing per cast due to Haste.
+ * We make the decision to only calculate this benefit, which is fairly simple math:
```
MartijnHols: Doesn't this result in a very large undervaluation of Haste? The increased cast speed is an important part of the gain and for many other specs reason enough to use gear with Haste on it. Imagine if Resto Druid had 0 gain from HoTs ticking faster; would that mean Resto Druids would aim for 0 Haste? Your screenshot reveals Haste when just valuing HoT tick rate gain to already be close to Int, I reckon if you included cast speed (and thus reduced GCD) it would go much beyond.

I'm surprised by this approach and think it might be misleading. Is this the regular way to calculate the Haste value within the Resto Druid TC community?

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * This is the same calculation as for haste, healAmount / (1 + versPercentage) * versPercentageFromOneRating.
+ * We don't currently consider the damage reduction bonus from vers, and it's probably not a good thing to add directly to
+ * the weight as damage reduction isn't quite the same thing as healing. It's easy enough to calculate just by skimming the
+ * appropriate percentage of each toPlayer_damage. I will probably add it to the tooltip in the future.
+ *
+ * Leech -
+ * Calculating this is different depending on if the player does or does not already have any Leech. If the player does have
+ * Leech we calculate this pretty much the same as int: for every Leech heal the increase in power from one leech is
+ * simply 1 / totalLeechRating. If the player does not already have Leech, things are much more difficult. I have not yet
+ * settled on an approach, and so this calculation is not yet implemented.
```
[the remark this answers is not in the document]
kfinch: This is almost the exact thought I had on the matter. What bothers me is it's a very speculative value. I'll probably still implement and feelycraft if I think it's right or not 😛

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * A few special cases are considered. We're ignoring all Tranq healing, even ticks that don't overheal, under the assumption
+ * that the last ticks always overheal so what's the difference? I know this is a violation of the "would've" rule I detailed
+ * above. This decision may be revisited.
+ *
+ * Leech and Velens heals are disregarded for the purpose of the main stat's weights, as they scale with whatever procced them.
+ * It's easiest to just not count them.
+ *
+ * If a heal is not in the database of heals, it's assumed to scale only with Vers. This is generally a good assumption,
+ * and this is how most trinket heals behave.
+ */
```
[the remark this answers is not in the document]
kfinch: I'd love to add Avoidance, but I don't know how to detect if incoming damage is prevented by it or not.

src/Parser/RestoDruid/SpellInfo.js
```diff
@@ -0,0 +1,245 @@
...
+  [SPELLS.CENARION_WARD.id]: {
+    int: true,
+    crit: true,
+    hasteHpm: true,
+    hasteHpct: false,
+    mastery: true,
+    masteryStack: true,
+    vers: true,
+  },
+  [SPELLS.FRENZIED_REGENERATION.id]: { // TODO does this actually scale with nothing?
```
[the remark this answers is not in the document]
kfinch: Yeah Torty told me about the Mastery part at least. I'll fix this.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * The mastery calculation is almost the same as Haste's, in that you normalize the heal amount to be the amount it would
+ * have been without any of the stat, and then multiply by the percentage gain from one rating. The one wrinkle is that mastery's
+ * strength is multiplied by the number of hots on target. Fortunately, we can easily calculate the number of HoTs on the target
+ * of each heal.
+ *
+ * Versatility -
+ * This is the same calculation as for haste, healAmount / (1 + versPercentage) * versPercentageFromOneRating.
+ * We don't currently consider the damage reduction bonus from vers, and it's probably not a good thing to add directly to
+ * the weight as damage reduction isn't quite the same thing as healing. It's easy enough to calculate just by skimming the
+ * appropriate percentage of each toPlayer_damage. I will probably add it to the tooltip in the future.
```
MartijnHols: I think adding it not to the tooltip but as a new row in the panel might be better. I reckon DR is largely undervalued and giving it a prominent place will cause people to think about it (like they should) and choose for themselves which they value more; the versa stat weight for parsing well or the versa stat weight that helps more to kill bosses.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ *
+ * Leech -
+ * Calculating this is different depending on if the player does or does not already have any Leech. If the player does have
+ * Leech we calculate this pretty much the same as int: for every Leech heal the increase in power from one leech is
+ * simply 1 / totalLeechRating. If the player does not already have Leech, things are much more difficult. I have not yet
+ * settled on an approach, and so this calculation is not yet implemented.
+ *
+ * Special Cases -
+ *
+ * A few special cases are considered. We're ignoring all Tranq healing, even ticks that don't overheal, under the assumption
```
MartijnHols: How about summing all Tranq healing per player and looking at the sum; if sum didn't include overhealing include it, otherwise ignore?

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * that the last ticks always overheal so what's the difference? I know this is a violation of the "would've" rule I detailed
+ * above. This decision may be revisited.
+ *
+ * Leech and Velens heals are disregarded for the purpose of the main stat's weights, as they scale with whatever procced them.
+ * It's easiest to just not count them.
+ *
+ * If a heal is not in the database of heals, it's assumed to scale only with Vers. This is generally a good assumption,
```
MartijnHols: 👌

src/Parser/RestoDruid/SpellInfo.js
```diff
@@ -0,0 +1,245 @@
...
+    int: true,
+    crit: true,
+    hasteHpm: false,
+    hasteHpct: true,
+    mastery: true,
+    masteryStack: false,
+    vers: true,
+  },
+  [SPELLS.TRANQUILITY_HEAL.id]: {
+    ignored: true, // Dreamgrove theorycrafters say should be ignored because it always overheals, could unfairly skew weights against Haste if included. I'd like a better solution, but this will do for now.
```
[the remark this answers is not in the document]
kfinch: I also have mixed feelings about it. I think I included it originally in my analyzer because I was trying to compare values with RDSW and TwigIt and they both make this decision too.

The concrete effect of ignoring Tranq is a higher value for Haste HPM, because it's the only stat that doesn't benefit Tranq.

I'm considering melding the ticks for each player into one big heal, and considering things then. So tick sets on any player where even one of the ticks overheals will be disregarded, but if a player takes all 5 ticks and isn't overhealed then I count it. I could think of a few similar tricks.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+  // These are the total healing that would be gained if their respective stats ratings were increased by one.
+  totalOneInt = 0;
+  totalOneCrit = 0;
+  totalOneHasteHpm = 0;
+  totalOneMastery = 0;
+  totalOneVers = 0;
+  totalOneLeech = 0;
+
+  on_initialized() {
+    this.concordanceAmount = 3700 + (this.combatants.selected.traitsBySpellId[SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_TRAIT.id] || 0) * 300; // TODO remove when stat tracker added
```
MartijnHols: I recommend writing this like: `4000 + ((this.combatants.selected.traitsBySpellId[SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_TRAIT.id] - 1) || 0) * 300`

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+    // Leech is marked as a 'multplier' heal, so we have to check it before we do the early return below
+    const hasLeech = this.combatants.selected.leechRating > 0; // TODO replace when dynamic stats
+    if(hasLeech && spellId === SPELLS.LEECH.id && !healVal.overheal) {
+      this.totalOneLeech += amount / this.combatants.selected.leechRating; // TODO replace when dynamic stats
+    } else if(!hasLeech) {
+      // TODO this will be a pain to implement
+    }
+    // endregion
+
+    if(!spellInfo.multiplier) {
```
MartijnHols: This piece of code is unclear, half of it due to the property name not really explaining what it is but I can't think of anything better. Maybe add a comment?

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+      this.totalAdjustedHealing += healVal.effective;
+    }
+
+    if(spellInfo.multiplier || spellInfo.ignored || healVal.overheal) {
```
MartijnHols: I preferred the old way of doing this in 3 separate if-statements; made it clearer they're unrelated and don't depend on each other.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * it doesn't only matter how much is healed, but also how quickly. We'll count all healing that doesn't overheal even if we
+ * theorize that it might lead to overhealing. We're more strict on heals that overheal: they're disregarded entirely even
+ * if they're only partial overhealing. This is because a spell that partially overheals will do the same amount of effective
+ * healing regardless of the raw strength of the heal. This approach can cause some possible weirdness. For example, consider
+ * a situation where there is a heal for 500 (0 overheal) followed immediately by a heal for 400 (+ 100 overheal). 500 of this
+ * healing would be counted towards stat weights. Now consider a situation where one heal does 900 (+ 100 overheal). Functionally,
+ * this situation is identical, but in this case none of the healing is counted towards stat weights. Counting all effective
+ * healing, even heals that partially overheal, fixes this issue and is overall a valid approach, but still not one I will be taking.
+ * I think an advantage of disregarding all partial overheals is that it gives "top off" heals an effective lower weight than
```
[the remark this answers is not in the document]
kfinch: I plan to add more stuff to tooltips once I make my thoughts less ramble-y. I'm afraid if I cut down what I'm trying to say too much, it could just cause more confusion.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+
+const ARMOR_INT_MULTIPLIER = 1.05; // 5% int bonus from wearing all leather means each new point of int worth 1.05 vs character sheet int
+
+/*
+ * Resto Druid Stat Weights Methodology
+ *
+ * Approach -
+ * This module generates the players stat weights using the actual logged events. We keep a listing of all the player's healing spells
+ * along with which stats those spells scales with, and for each stat a heal scales with we do some simple math to find out
+ * how much more it would have healed if the player had one more of that stat. We compare the total healing increases
```
MartijnHols: `for each stat a heal scales with we do some simple math to find out how much more it would have healed if the player had one more of that stat`

If I understand the code correctly I think this isn't exactly right. It's actually `for each stat a heal scales with we do some simple math to find out how much healing one of that stat contributed`, no? Seeing as you seem to calculate the actual contribution of the last value of each stat, not the gain from adding 1 stat. It's no big deal though.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+ * if they're only partial overhealing. This is because a spell that partially overheals will do the same amount of effective
+ * healing regardless of the raw strength of the heal. This approach can cause some possible weirdness. For example, consider
+ * a situation where there is a heal for 500 (0 overheal) followed immediately by a heal for 400 (+ 100 overheal). 500 of this
+ * healing would be counted towards stat weights. Now consider a situation where one heal does 900 (+ 100 overheal). Functionally,
+ * this situation is identical, but in this case none of the healing is counted towards stat weights. Counting all effective
+ * healing, even heals that partially overheal, fixes this issue and is overall a valid approach, but still not one I will be taking.
+ * I think an advantage of disregarding all partial overheals is that it gives "top off" heals an effective lower weight than
+ * "life saver" heals. Still, this is a decision I will revisit.
+ *
+ * Stat Tracking -
```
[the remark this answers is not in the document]
kfinch: I'll make a few more updates to make the StatTracker fit better. Because of the manual handling of Concordance and 2t19, I'm satisfied enough with this module even without StatTracker.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: It does certainly undervalue Haste. This is a common calculation in the Resto Druid TC community, and we just say it's undervalued by ¯\_(ツ)_/¯

I could be mistaken, but I believe Resto Druid is less interested in faster cast speed than other healing classes. Our main heal is instant, and we do not often need to spam heals. Yes I know Haste reduces GCD for an instant spell, but unlike for a cast time spell Haste doesn't reduce the time in between when you decide to cast and when it lands.

If I can think of a good approach, I'll try to re-add Haste HPCT. I'm just not sure how best to do it.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Yeah that's a good idea, I'll give it more visibility one way or the other.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: That's exactly what I described when responding to blazyb above. It's what I plan to do.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: I considered this, but was all ¯\\\_(ツ)\_/¯

src/Parser/RestoDruid/Modules/Features/StatWeights.js
MartijnHols: We discussed this on Discord some time back and I concluded it's a major challenge when doing log-based stat weight calculation. 🤷‍♂️ I was thinking of a super simplified approach; 1% Haste = 1% more healing from all spells affected by Haste (GCD or channel), the beauty lies in that it's an extremely simple approach that is right in a way.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+    const hasteHpmForOnePercent = this._ratingPerOnePercent(this.totalOneHasteHpm);
+    const masteryForOnePercent = this._ratingPerOnePercent(this.totalOneMastery);
+    const versForOnePercent = this._ratingPerOnePercent(this.totalOneVers);
+    let leechForOnePercent = this._ratingPerOnePercent(this.totalOneLeech);
+    if(!hasLeech) {
+      leechForOnePercent = "NYI";
+    }
+
+    const hasteHpmTooltip = "HPM stands for 'Healing per Mana'. In valuing Haste, it considers only the faster HoT ticking and not the reduced cast times. Effectively it models haste's bonus to mana efficiency. This is typically the better calculation to use for raid encounters where mana is an issue.";
+    const leechTooltip = "Leech weight can currently only be calculated when player already has some Leech rating";
```
MartijnHols: `when player` -> `when you`

src/Parser/RestoDruid/Modules/Features/StatWeights.js
```diff
@@ -0,0 +1,316 @@
...
+      { stat: 'Leech', weight:leechWeight, ratingForOne:leechForOnePercent, tooltip:leechTooltip },
+    ];
+  }
+
+  extraPanel() {
+    const results = this._prepareResults();
+    return (
+      <div className="panel items">
+      <div className="panel-heading">
+        <h2><dfn data-tip="Weights are calculated using the actual circumstances of this encounter. Weights are likely to differ based on fight, raid size, items used, talents chosen, etc.">STAT WEIGHTS</dfn>
```
MartijnHols: Better not to capitalize; it's currently a design choice that's taken care of by CSS. It might change in the future.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: I think of it as the gain from one stat, but it's functionally the same thing.

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: yeah I did a shit job of explaining it, will add comment

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Fixing

src/Parser/RestoDruid/Modules/Features/StatWeights.js
kfinch: Fixing

kfinch: Jk it looks dumb as hell and is out of style with other resto tooltips, not fixing

kfinch:
Pls to re-approve when checks are done, I'm excited to get this in today!

src/Parser/RestoDruid/SpellInfo.js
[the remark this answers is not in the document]
buimichael: That's nice, however there's some trinkets/traits that scales with haste/crit. Not sure how much that'd impact the stat weights though. Some trinkets do 5-8% of your total healing, so my guess is that it'd be substantial.

buimichael:
GJ @kfinch !

src/Parser/RestoDruid/SpellInfo.js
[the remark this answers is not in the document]
kfinch: Any trinket that has scaling other than "vers only" can and should be added to the SpellInfo list

kfinch:
Ok, so post release improvements to implement:
- Finish up StatTracker, merge it, and integrate its functionality into StatWeights
- Sum Tranq healing on a per-player-per-cast basis, then unignore it
- Implement Leech weight when player doesn't have Leech
- Add Haste HPCT calculation?
- Add Avoidance calculation (if it's possible to separate out the damage effected by it)
- Add an 'info' button to the stat weights panel that would reveal the methodology write-up. I'm thinking a display similar to what's used for the "more information" button on each spec.

src/Parser/RestoDruid/SpellInfo.js
```diff
@@ -0,0 +1,245 @@
...
+    vers: false,
+  },
+  [SPELLS.XAVARICS_MAGNUM_OPUS.id]: { // TODO does it really scale with nothing (except stam)?
+    int: false,
+    crit: false,
+    hasteHpm: false,
+    hasteHpct: false,
+    mastery: false,
+    masteryStack: false,
+    vers: false,
```
buimichael: Prydaz scales with vers, confirmed that ingame

PR #606, opened 2017-10-24
[Frost Mage] Water Elemental Basic Support

Sharrq:
Adds the Water Elemental as a Pet in PETS and adds/reorganizes the Water Elemental Spells. This is essentially just basic support to be used later for making modules revolving around the water elemental.

### DIFF
--- a/src/common/SPELLS/MAGE.js
+++ b/src/common/SPELLS/MAGE.js
@@ -133,11 +133,6 @@ export default {
     name: 'Summon Water Elemental',
     icon: 'spell_frost_summonwaterelemental_2',
   },
-  WATER_JET: {
-    id: 135029,
-    name: 'Water Jet',
-    icon: 'ability_mage_waterjet',
-  },
   EBONBOLT: {
     id: 214634,
     name: 'Ebonbolt',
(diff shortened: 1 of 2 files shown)


src/common/SPELLS/MAGE.js
```diff
@@ -209,11 +221,6 @@ export default {
     name: 'Shatter',
     icon: 'spell_frost_frostshock',
   },
-  FREEZE: {
-    id: 231596,
```
kfinch: What's with this Freeze being different id / icon than the one added above?

Sharrq: I believe the old ID (231596) was incorrect and the new one (33395) is the correct one. 

This is the old one (which looks like it is the passive that gives the elemental the ability to use freeze)
http://www.wowhead.com/spell=231596/freeze

and this is the new one, which looks like the actual freeze ability
http://www.wowhead.com/spell=33395/freeze

kfinch: Have you confirmed this with logs? I know right now it doesn't matter because none of our code uses the SPELLS entry, but later on when someone does use it I don't want them surprised when it doesn't work.

Sharrq: I havent, but ill go digging through warcraft logs to see if i can find a log where someone didnt use lonely winter and used freeze .... that should be fun.

kfinch: If you can't find instances anywhere in logs, that's OK, just remove the FREEZE entry entirely from SPELLS. I think it's a good rule of thumb that you shouldn't add something to SPELLS until you've seen that it actually shows as you expect in logs.

Sharrq: The icon of the new ID matches the icon in game, so im almost positive that that is the correct ID, but ill still remove it for now.

Sharrq:
Im just going to close this and come back to it when i can actually add the analysis for the water elemental itself.

PR #609, opened 2017-10-24
[Unholy Death Knight]Added module for virulent plague uptime

jjs451:
Added basic module for Unholy's DoT uptime.  Made sure to double check copied code for Warlock spells this time 😁

### DIFF
--- a/src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueUptime.js
+++ b/src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueUptime.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import Module from 'Parser/Core/Module';
+import Enemies from 'Parser/Core/Modules/Enemies';
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import { formatPercentage } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+class VirulentPlagueUptime extends Module {
+  static dependencies = {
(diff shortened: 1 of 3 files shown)


src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueUptime.js
```diff
@@ -0,0 +1,40 @@
...
+import { formatPercentage } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+class VirulentPlagueUptime extends Module {
+  static dependencies = {
+    enemies: Enemies,
+  };
+
+  suggestions(when) {
+    const vpUptime = this.enemies.getBuffUptime(SPELLS.VIRULENT_PLAGUE.id) / this.owner.fightDuration;
```
kfinch: .... how does this line not error out? I thought 'enemies' was a mapping of entity objects, and that getBuffUptime only works on a single entity.

jjs451: I'm not sure, I stole it from other specs that track DoTs.  Blood dk, both dh specs, bear and cat druid, spriest, ele shamans, all three warlock specs, and arms warrior all use it in tracking uptimes.  Balance druid used `this.owner.modules.enemies.getBuffUptime()` though.

kfinch: Oh I missed that one, yeah that helper looks to do exactly what you need. Thanks.

PR #621, opened 2017-10-25
[Discipline Priest] Fix in atonement ordering for double damaging events

karlpralow:
This fixes the main problem with atonement tracking (when 2 damaging events happen simultaneously).
This does not handle the latency of atonement on self issue that will be fixed in another module


MartijnHols:
Hey again. There were some changes to the core structure of modules that changed the way you added "reorderEvents". Please update your code accordingly. It won't function as it is.

Examples: https://github.com/WoWAnalyzer/WoWAnalyzer/tree/master/src/Parser/Paladin/Holy/Normalizers

PR #625, opened 2017-10-25
Death Knight - Unholy - Update AlwaysBeCasting.js

Maxweii:
Reworded downtime suggestion to be more appropriate to melee/Unholy.


Pewtro:
Might want to add a .major() to this, to indicate when it becomes a big problem?

PR #628, opened 2017-10-26
[fire mage] Fixed bugs in 4pc t20 calculation

kfinch:
The implementation of Fire Mage Tier20 4set wasn't working at all in practice.

Unfortunately, the solution had to be somewhat complex in practice because of how the bonus was applied in logs, and due to fire mage cast time / spell travel time issues.

I also updated the display of the bonus.


kfinch:
fire_mage_t20_4set

PR #629, opened 2017-10-26
[Items] Added Cradle of Anguish to items

jjs451:
The goal is to report % uptime on the buff from the trinket.  Apparently WCL has issues with buffs that are applied pre-combat and never drop, so to get around that I went with this:
 `const uptimePercent = (this.combatants.selected.getBuffUptime(SPELLS.STRENGTH_OF_WILL.id) / this.owner.fightDuration) || 1;`  

My logic is that there are two scenarios where uptime could be 0, either it never dropped off or someone entered combat at under 50% hp and never got healed to 80%.  The latter seems extremely improbable and far, far out of the bounds of a normal use case.  This took little work to make so nothing lost if this is too janky for production.


src/Parser/Core/Modules/Items/CradleOfAnguish.js
```diff
@@ -0,0 +1,38 @@
...
+    // If the buff never fell off then getBuffUptime will report 0 due to WCL not tracking precombat buffs that don't drop/reapply
+    // setting uptimePercent to 1 gets around this
+    const uptimePercent = (this.combatants.selected.getBuffUptime(SPELLS.STRENGTH_OF_WILL.id) / this.owner.fightDuration) || 1;
+    return {
+      item: ITEMS.CRADLE_OF_ANGUISH,
+      result: (
+        <span>
+          <dfn
+            data-tip={`Buff was active for ${formatPercentage(uptimePercent)}% of the fight.`}
+          >
```
kfinch: This tooltip is not needed.

src/Parser/Core/Modules/Items/CradleOfAnguish.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import { formatPercentage } from 'common/format';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+class CradeOfAnguish extends Analyzer{
```
kfinch: Typo here and at the bottom, should be `CradleOfAnguish`

kfinch:
Remember also to pull from upstream master to resolve the conflict (which is probably just another item added to CombatLogParser in the "same" spot)

kfinch:
It looks like you didn't actually resolve the merge conflicts... look closely at the file changes in the 'fixed conflict' commit.

jjs451:
Shoot, didn't realize Git automatically added further commits to a branch to a PR.  I'm gonna close this for now while I figure out what I'm doing

kfinch:
Yeah anything on your GitHub repo branch gets auto added, which is nice most of the time. It won't get added until you push from your local, of course.

PR #630, opened 2017-10-26
[resto druid] Updated Prydaz entry in SpellInfo. Implemented Leech when player doesn't have Leech in StatWeights

kfinch:


src/Parser/Druid/Restoration/Modules/Features/StatWeights.js
```diff
@@ -147,8 +160,9 @@ class StatWeights extends Analyzer {
     const hasLeech = this.combatants.selected.leechRating > 0; // TODO replace when dynamic stats
     if(hasLeech && spellId === SPELLS.LEECH.id && !healVal.overheal) {
       this.totalOneLeech += amount / this.combatants.selected.leechRating; // TODO replace when dynamic stats
-    } else if(!hasLeech) {
-      // TODO this will be a pain to implement
+    } else if(!hasLeech && this.playerHealthMissing > 0) {
+      const bonusFromOneLeech = 1 / 23000;
+      this.totalOneLeech += healVal.raw * bonusFromOneLeech;
```
buimichael: Shouldn't there be a check whether the oneLeech would partially overheal or not? I guess it will proably never overheal since you taking the heal from one point of leech....

kfinch: Yeah, it's a very small amount. This is basically me implementing the simplest solution I could think of.

PR #633, opened 2017-10-26
[frost mage] Updated Winter's Chill display

kfinch:
We now also track frostbolt hits into Winter's Chill, and added a suggestion for it as well. Suggestion now triggers on missed PPM rather than total missed. Updated display of statistic to show % utilization, and added a stat for double icelance hits.

winters-chill


kfinch:
Updated in response to feedback from Mage discord

Sharrq:
You can also Glacial Spike into Flurry if you are talented into Glacial Spike.

kfinch:
This game is too hard

Sharrq:
Lol, I'm pretty sure that's the last one that is acceptable to be casted before flurry, so it should just be those 3. I think

PR #634, opened 2017-10-26
Fixes the z-index of the header overlay

fasib:
This gets fixed:
firefox_2017-10-26_18-16-37

The header's z-index of 10 got caped to 1 before this commit.


MartijnHols:
Thanks!

PR #635, opened 2017-10-26
(fix-ele) Remove Storm Totem which is a Summon Event and not a Buff.

fasib:
So that Totem Mastery can show its uptime (and not 0.00%...)


fasib:
Looks like travis dosen't handle to fast commits/updates and didn't update the build urls 😄

PR #637, opened 2017-10-26
[Discipline Priest] Atonement Normalizer

karlpralow:
This fixes all double successive damage events issues (almost all the atonement issues).
It basically goes down the events and keeps a list of the last atonement targets since last damage. 
When it encounters an ID it already saw, it pushes down the last damage down to split the atonement block


src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
```diff
@@ -0,0 +1,45 @@
...
+        _damageEventIndexes.push(eventIndex);
+        _encounteredTargetIDs = [];
+        return;
+      }
+
+      if(event.type === "heal" && isAtonement(event)) {
+
+        // We encountered a targetID we already encountered since the last damage
+        // event. We push down the last damage event here
+        if(_encounteredTargetIDs.indexOf(event.targetID) >= 0) {
```
MartijnHols: Since this doesn't check that the sourceId matches, I think this will break if you have 2 disc priests in the group.

src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
```diff
@@ -0,0 +1,45 @@
...
+          const lastDamageEvent = fixedEvents.splice(_damageEventIndexes[_damageEventIndexes.length -1], 1)[0];
```
MartijnHols: I recommend setting `__modified` to `true` so that this event is highlighted in the events tab:

`lastDamageEvent.__modified = true;`

src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
```diff
@@ -0,0 +1,45 @@
...
+          fixedEvents.splice(fixedEvents.length - 1, 0, lastDamageEvent);
+          _encounteredTargetIDs = [];
+          return;
+        }
+
+        //  We ignore atonement on self in the handling of repeating targetIDs
+        //  Because of latency issues
```
MartijnHols: I think this "because of" is confusing and perhaps even misleading. How about stating it's being handled separately in another normalizer?

src/tests/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.test.js
```diff
@@ -0,0 +1,106 @@
...
+      DamagingEvent1,
+      DamagingEvent2,
+      AtonementOnPlayer1,
+      AtonementOnPlayer2,
+      AtonementOnPlayer1,
+      AtonementOnPlayer2
+    ];
+
+    let result = atonementSuccessiveDamageNormalizer.normalize(AtonementOf2DamingEventsGroupedTogether);
+    expect(result[0]).toBe(AtonementOnPlayer2)
```
MartijnHols: I think this would be easier to read if you check that it matches the order of an array instead;

```
expect(result).toEqual([
  AtonementOnPlayer2,
  DamagingEvent1,
  ...etc
]);
```

rp4rk: Agree with Zero, you might find you have to use `toDeepEqual` to compare though and not `toEqual`.

MartijnHols: I reckon toEqual does the exact same check as what happens here

src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
karlpralow: the atonement from another priest shows here ? When I look at the initial events I only have events that are from the priest. What am I not understanding ?

src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
karlpralow: done

src/tests/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.test.js
karlpralow: toEqual worked

src/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.js
MartijnHols: Being healed by another Disc shows up.

PR #639, opened 2017-10-26
[Marksmanship] Bugfixes for Patient Sniper

sMteX:
Not proud of how I solved it but works as of now. 2 issues this solves:
1. If for some reason, Aimed Shot was cast but never landed and did damage (picture firing AS at Sisters of the Moon right before the boss despawns and another one spawns), it broke all Patient Sniper bonus damage from that point forward
=> solved by filtering out cast events that are too old - AS has a fast travel time, around <700ms so I added 1s max travel time
2. If Vulnerable debuff windows were slightly larger than 7000ms due to lag (removedebuff() was called let's say 70ms later than it *should've been*), then the difference between shot cast and start of the debuff window was > 7000ms which resulted in timeIntoVulnerable = 7 
=> solved by adding 100ms lag tolerance


Pewtro:
Looks good to me, gonna leave up for a bit if someone else has something to say about this :)

PR #643, opened 2017-10-27
Death Knight - Unholy - Create T20_2pc.js

Maxweii:
Creating tracker for tier 20 four piece buff uptime


Maxweii:
Hm I think I meant 2 piece then, I mix them up a lot, and that was next part was currently on. Though Keira or Khaz might get that done faster than I would

Maxweii:
Should be fixed now and pushed a thing for adding the tier set bonus for Unholy - ah as I type this fixing a wording on things so it matches

Pewtro:
To correctly monitor uptime you will need to track this: http://www.wowhead.com/spell=246995/master-of-ghouls

The buff you currently track is merely the hidden buff granted when tier202p is equipped and should be used in on_initialized() and anywhere else you need the icon, but is not the one you want to track the uptime of.

Pewtro:
Adding to my original request: Please ensure that both ``UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS_BUFF`` and ``UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS`` exist in SPELLS (right now only 1 does)

Maxweii:
Talked to Khazzak and he's going to finish this up/fix things for it.

PR #644, opened 2017-10-27
Death Knight - Adding Unholy 2 set Spell ID

Maxweii:
Added Unholy 2 piece tier set

### DIFF
--- a/src/common/SPELLS/DEATH_KNIGHT.js
+++ b/src/common/SPELLS/DEATH_KNIGHT.js
@@ -212,6 +212,13 @@ export default {
     name: 'Virulent Plague',
     icon: 'ability_creature_disease_02',
   },  
+  
+  // Unholy Tier Sets
+  // T20 2 Piece
+  UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS_BUFF: {
+    id: 242064,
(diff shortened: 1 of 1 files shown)


Pewtro:
Missing a ``},`` to close it off.

Maxweii:
Can do, sorry bout that :c

Pewtro:
You still need the other spellId, you will need both. One is for checking if tier20 is equipped (the first one you had), the other is checking for uptime (this new one). 

So you'll have a 
DK_UH_T19_2P_BONUS with the ID 242064
and a 
DK_UH_T19_2P_BONUS_BUFF with the ID 246995

It will then in your code read something like this: 

```jsx
on_initialized() {
    this.active = this.combatants.selected.hasBuff(SPELLS.DK_UH_T19_2P_BONUS.id);
}
```

and: 
```jsx
const uptime = this.combatants.selected.getBuffUptime(SPELLS.DK_UH_T19_2P_BONUS_BUFF.id) / this.owner.fightDuration;
```

With the const ofc being wrapped in the return like before, and then you can use either ``BONUS`` or ``BONUS_BUFF`` for the icons, tooltip, SpellLink whichever makes the most sense (sometimes one of them lacks an icon so you'd for the other one or something entirely different etc)

PR #664, opened 2017-10-27
[frost mage] Winter's Chill render fix and Brain Freeze appearance update

kfinch:
Winter's chill display could split strangely on wider windows, this fixes the issue.

Before:
winters-chill-before

After:
winters-chill-after

### DIFF
--- a/src/Parser/Mage/Frost/CHANGELOG.js
+++ b/src/Parser/Mage/Frost/CHANGELOG.js
@@ -1,4 +1,5 @@
 export default `
+27-10-2017 - Enhanced display of Winter's Chill and Brain Freeze statistics. (by Sref)
 17-10-2017 - Added Frozen Bomb, Unstable Magic, and Arctic Gale modules. (by Sref)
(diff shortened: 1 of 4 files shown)


kfinch:
Woops I intended to put second commit on different PR, but these are similar enough so whatever.... this second commit is updates to the BrainFreeze module to make it display in same style as Winter's Chill. Detection for expired procs was added, suggestion / stat reworded a bit.

brain-freeze

MartijnHols:
Please solve the merge conflicts so this can be merged.

kfinch:
I tried to do it in browser because it's a simple conflict, not sure that I did it right though...

kfinch:
I'm pretty sure that merge did resolve the conflict and just didn't show right... in any case I pushed another commit to clear things.

This commit added my name to Frost Mage maintainer, after Sharrq. While developing this spec with Sharrq we discussed if my name should be on maintainer, and decided "not until I've actually contributed several modules". At this point I think I have contributed several modules.

PR #665, opened 2017-10-27
Disc refactor

awlego:
Since the disc code was written, there have been a number of improvements to take the suggestions and rendering management out of the CombatLogParser module. I have refactored the disc code to use these new suggestions and rendering management.


rp4rk:
This is quite large, so bear with me!

src/Parser/Priest/Discipline/CHANGELOG.js
```diff
@@ -1,4 +1,5 @@
 export default `
+27-10-2017 - Refactored disc module. (Gao)
```
awlego: Not sure if internal changes that don't have external changes should be in this changelog.

awlego:
Sorry it's large. I tried to think of a reasonable way to break this up... but partial refactor PRs didn't seem great either.

rp4rk:
I've read through this entirely and I'm struggling to find anything worth blocking it, really really impressive work. Thank you!

PR #679, opened 2017-10-29
Scale stat weights with current HP (giving heals on low health targets a higher weight gain)

MartijnHols:
# Depends on #678. Clean diff: https://github.com/MartijnHols/WoWAnalyzer/compare/core-healer-stat-weights...stat-weights-hp-scaling

Regular weights:
image
Weighted weights:
image

No option to toggle this on yet as the PR is half questioning if we should bother.


src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
```diff
@@ -0,0 +1,373 @@
...
+    }
+    if (spellInfo.vers) {
+      this.totalOneVers += this._adjustGain(this._versatility(eventForWeights, healVal), targetHealthPercentage);
+    }
+  }
+  _adjustGain(gain, targetHealthPercentage) {
+    // We want 0-20% health to get value gain, and then linearly decay to 100% health
+    const maxValueHealthPercentage = 0.3;
+    const mult = 1 - Math.max(0, (targetHealthPercentage - maxValueHealthPercentage) / (1 - maxValueHealthPercentage));
+    return this.scaleWeightsWithHealth ? gain * mult : gain;
```
kfinch: Maybe just use `this.scaleWeightsWithHealth` to do an early return as first line? Doesn't matter that much though.

src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
```diff
@@ -0,0 +1,373 @@
...
+    if (spellInfo.mastery) {
+      this.totalOneMastery += this._adjustGain(this._mastery(eventForWeights, healVal), targetHealthPercentage);
+    }
+    if (spellInfo.vers) {
+      this.totalOneVers += this._adjustGain(this._versatility(eventForWeights, healVal), targetHealthPercentage);
+    }
+  }
+  _adjustGain(gain, targetHealthPercentage) {
+    // We want 0-20% health to get value gain, and then linearly decay to 100% health
+    const maxValueHealthPercentage = 0.3;
```
kfinch: Comment says 20%, code says 30%

src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
MartijnHols: I don't want to suggest I intended there to be multiple scalers mixed

rp4rk:
I do not think this makes sense, by optimising for healing lower health targets you're probably going to have a situation where more people fall low on health.

MartijnHols:
I think that is largely going to differ between healers. I believe this way of calculation weights would be perfectly fine for Holy Paladins as I reckon "topping off" is better left to other healers most of the time (e.g. let already rolling hots finish the job).

As seen in the screenshots the difference for Holy Paladins is minimal, this was a large reason for the PR in its current form (no toggle yet), to discuss the current formula (approach), analyze the difference from the default implementation (which already kinda lowers the value of topping off heals by not giving overhealing heals a weight value) and to question if the added complexity of this approach is worth the minimal differences in stat weights.

I'm also curious what it would show for other healers, in particular Restoration Shamans (I imagine it will more significantly increase their Mastery value). As for Restoration Druid I wonder if using this approach would maybe be detrimental for the reasons you stated.

kfinch:
@rp4rk It really depends on the difference in makeup of the heals typically used on low health targets vs 'top off' heals. My guess for Resto Druid would be that it would cause crit and maybe mastery to be weighted higher.

The important question is whether or not it's 'correct' to weight those stats a little higher than we otherwise would. I don't have an easy answer to that. I think it would be reasonable to include this feature, but only behind a toggle that defaults to off, and only with a tooltip that explains what's going on. If the difference between scaled and unscaled weights is small, we might just say "what's the point?"

MartijnHols:
Merged since it is off by default anyway. I don't think we'll use it but it does no harm for experimental reasons.

PR #686, opened 2017-10-29
[Ele Shami] Changed the overload StatisticBox layout to match the new left text-alignment.

fasib:
Old-Layout:
firefox_2017-10-29_23-24-57

New-Layout:
firefox_2017-10-29_23-25-04


fasib:
Other alignment ideas:
mspaint_2017-10-29_23-50-48

kfinch:
Looks nice, but I don't like that they're sorted by number of overloads. Order should be static IMO, for readability reasons I'd expect spells to always be in same order. 

I assume you did the sort because you thought the three digits look bad in the middle, why not make whatever spell that is always be first?

PR #689, opened 2017-10-30
Distance moved replaced by % of Time spent moving

janvavra:
Percentage of fight duration spent moving should be a better indicator of time wasted by movement.
I modified the basic algorithm for distance to measure time spent moving (as a time between events with different x/y coordinates). Distance is still reported in the tooltip. 
But there could be a better approach to this. Is this accurate enough? Should be with high event frequency.


src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -54,7 +56,13 @@ class DistanceMoved extends Analyzer {
     if (!this.lastPositionUpdate) {
       return;
     }
-    this.totalDistanceMoved += this.calculateDistance(this.lastPositionUpdate.x, this.lastPositionUpdate.y, event.x, event.y);
+    let distanceMoved = this.calculateDistance(this.lastPositionUpdate.x, this.lastPositionUpdate.y, event.x, event.y);
```
fasib: You should try to alway make the linter happy before you create a pull request or commit :)
With `npm run lint` or `yarn run lint`

```
./src/Parser/Core/Modules/Others/DistanceMoved.js

  Line 59:  'distanceMoved' is never reassigned. Use 'const' instead  prefer-const
```

janvavra: Thank you. Corrected and Travis is happy again.

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -66,13 +73,13 @@ class DistanceMoved extends Analyzer {
   }
 
   statistic() {
-    debug && console.log(`Total distance moved: ${this.totalDistanceMoved} yds`);
+    debug && console.log(`Time spend moving: ${this.timeSpendMoving / 1000} s, Total distance moved: ${this.totalDistanceMoved} yds`);
     return (
       <SmallStatisticBox
```
Pewtro: Should be ``Time spent moving: ``

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -66,13 +73,13 @@ class DistanceMoved extends Analyzer {
...
         icon={<Icon icon="spell_fire_burningspeed" />}
-        value={`≈${formatThousands(this.totalDistanceMoved)} yards`}
-        label="Distance moved"
-        tooltip={`≈${formatThousands(this.totalDistanceMoved / (this.owner.fightDuration / 1000) * 60)} yards per minute. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
+        value={`≈${formatPercentage(this.timeSpendMoving / (this.owner.fightDuration))}%`}
+        label="Fight spend moving"
+        tooltip={`In ≈${formatThousands(this.timeSpendMoving / 1000)} seconds of movement you moved by ≈${formatThousands(this.totalDistanceMoved)} yards. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
       />
```
Pewtro: ``you moved X yards. `` would sound better than ``you moved by X yards``

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -66,13 +73,13 @@ class DistanceMoved extends Analyzer {
...
+    debug && console.log(`Time spend moving: ${this.timeSpendMoving / 1000} s, Total distance moved: ${this.totalDistanceMoved} yds`);
     return (
       <SmallStatisticBox
         icon={<Icon icon="spell_fire_burningspeed" />}
-        value={`≈${formatThousands(this.totalDistanceMoved)} yards`}
-        label="Distance moved"
-        tooltip={`≈${formatThousands(this.totalDistanceMoved / (this.owner.fightDuration / 1000) * 60)} yards per minute. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
+        value={`≈${formatPercentage(this.timeSpendMoving / (this.owner.fightDuration))}%`}
+        label="Fight spend moving"
+        tooltip={`In ≈${formatThousands(this.timeSpendMoving / 1000)} seconds of movement you moved by ≈${formatThousands(this.totalDistanceMoved)} yards. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
```
Pewtro: Grammar here aswell ``Fight spent moving``

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -6,12 +6,14 @@ import { formatThousands } from 'common/format';
 import Analyzer from 'Parser/Core/Analyzer';
 
 import SmallStatisticBox, { STATISTIC_ORDER } from 'Main/SmallStatisticBox';
+import { formatPercentage } from "../../../../common/format";
```
Pewtro: Should be imported like so ``common/format``

janvavra: Thanks for noticing, fixed.

src/Parser/Core/Modules/Others/DistanceMoved.js
janvavra: Embarrassing .. addressed.

src/Parser/Core/Modules/Others/DistanceMoved.js
janvavra: Addressed.

src/Parser/Core/Modules/Others/DistanceMoved.js
janvavra: Indeed it would, addressed.

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -66,13 +74,13 @@ class DistanceMoved extends Analyzer {
...
-    debug && console.log(`Total distance moved: ${this.totalDistanceMoved} yds`);
+    debug && console.log(`Time spent moving: ${this.timeSpentMoving / 1000} s, Total distance moved: ${this.totalDistanceMoved} yds`);
     return (
       <SmallStatisticBox
         icon={<Icon icon="spell_fire_burningspeed" />}
-        value={`≈${formatThousands(this.totalDistanceMoved)} yards`}
-        label="Distance moved"
-        tooltip={`≈${formatThousands(this.totalDistanceMoved / (this.owner.fightDuration / 1000) * 60)} yards per minute. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
+        value={`≈${formatPercentage(this.timeSpentMoving / (this.owner.fightDuration))}%`}
+        label="Fight spent moving"
```
kfinch: Shouldn't this be "Time spent moving"?

src/Parser/Core/Modules/Others/DistanceMoved.js
```diff
@@ -66,13 +74,13 @@ class DistanceMoved extends Analyzer {
...
+        tooltip={`In ≈${formatThousands(this.timeSpentMoving / 1000)} seconds of movement you moved ≈${formatThousands(this.totalDistanceMoved)} yards. Consider this when analyzing the fight, as some fights require more movement than others. Unnecessary movement can result in a DPS/HPS loss.`}
```
kfinch: "Unnecessary movement can result in a DPS/HPS loss" is not really true for Tanks or Melee DPS, though I'm not sure if a reword is needed.

src/Parser/Core/Modules/Others/DistanceMoved.js
kfinch: I guess it makes a little more sense if read with the value, like "20% Fight spent moving"...

janvavra: I agree, sounds better. Addressed.

src/Parser/Core/Modules/Others/DistanceMoved.js
janvavra: This whole statistic calls for spec depended treatment in the future.  I would leave it as it is for now.

Sharrq: I assume this also would count movement as a result of using abilities like Roll, Blink, Chi Torpedo, Charge, Heroic Leap, Disengage, Harpoon, etc ... when those really shouldnt count against the player ... imo

janvavra: @Sharrq this is why I created this from a simple distance moved in yards. Using any movement increase lowers % of time spend moving in fight (it counting for you).

kfinch: @Sharrq Well in this case it wouldn't "count against them" much, because when you use a movement ability you don't spend much time moving because you're moving very fast.

You're basically pointing out one of the advantages of switching from Distance Moved to Time Moving.

Sharrq: Fair point, i didnt think about that.

Pewtro: Basically my idea for Marksmanship (and to a degree BM because movement is irrelevant for them) is to implement their own spec-specific version of this and ensure that movement while X buff is up doesn't count as it allows movement while casting and then to alter the tooltip to showcase that information correctly. In that case it would only show the movement when it is actually a possible detriment (and show total movement in tooltip or something).

In any case I think this version of a movement module is significantly better than what we had before, as it's more lenient on the mobility spells.

janvavra:
@kfinch 
>Do we have any idea how accurate these values really are? 

I dont think you can measure it in game in a better way then we are doing now. Maybe some custom addon logging your XY position every X ms -> basically what we have now in logs, just with regular intervals.

I tried to look at this vid https://youtu.be/IiDjhqmJukg and used corresponding log (https://www.warcraftlogs.com/reports/JGNZpztPYL7CFKv4#fight=1&type=damage-done) to predict movement time and distance traveled based on current implementation. Using eyeball test it corresponds pretty well. Example from the beggining of the vid https://i.imgur.com/PBe5q1i.jpg

>Are there more events we can look at to get more data points? 

Most common events we don't watch here  are all the buff events (remove/apply buff/debuff). From my testing adding them has negligible effect on final values. But this may change in future content.



>This calculation assumes that if the player moved in between events, then they spent that entire time moving; that's a pretty big assumption.


After adding buff/debuff events, I think we are at the limit of our possible accuracy. In addition to ~ we can explicitly mention, that it is an "approximation" and true value may be lower. Can we estimate the possible error? Maybe we can also do some checks (like if distance between watched events in the log is greater then X) and then outright refuse to estimate this.

MartijnHols:
I'd like to preserve the distance moved in yards as it's a nice meme statistic. Would it be possible to show both?

janvavra:
@MartijnHols meme is still preserved on mouse-over. You want to show it on main screen?

MartijnHols:
Would be nice if you could achieve that.

janvavra:
Meme request fulfilled. 
image
Could use some help in tooltip wording:
tooltip distance:
image
tooltip Time spent:
image

kfinch:
I like the look of the mini double stat. Still thinking on how to word the tooltips.

PR #692, opened 2017-10-30
[frost mage] Fixed IceLance suggestion tip

kfinch:
Ice Lance suggestion tip showed actual as 'missed' when it was actually the percentage hit. Uncapitalized 'recommended' to match style.

These changes are entirely to output text, and not to code.


kfinch:
Changed my mind, it **should** show 'missed' to match the other suggestion's style. Changed it to show missed instead of hit, and updated numbers so it triggers at same thresholds.

kfinch:
@Pewtro look again, I updated it

Pewtro:
Looks good to me!

PR #694, opened 2017-10-31
[Core/FrostMage] Cancelled Casts

Sharrq:
Module for showing the % of casts that were cancelled. Tooltip also shows the breakdown of spells cast, completed, and cancelled. Also filters out spells with a cast time, but the cast time was made instant by some proc or other event (Like Blizzard - Instant if Frozen Orb is up) .... but if the dev includes a pure instant cast spell in the list, then it will likely screw with the results because Instant Cast Spells dont have a "begincast", but spells that are sometimes instant do.

### DIFF
--- a/src/Parser/Core/Modules/CancelledCasts.js
+++ b/src/Parser/Core/Modules/CancelledCasts.js
@@ -13,58 +13,61 @@ class CancelledCasts extends Analyzer {
     combatants: Combatants,
   };
 
-  castsStarted = 0;
+  castsCancelled = 0;
   castsFinished = 0;
-  beginCastTimestamp = 0;
-  castTimestamp = 0;
+  beginCastSpellId = 0;
+  wasCastStarted;
 
-  static CANCELABLE_ABILITIES = [
+  static IGNORED_ABILITIES = [
   ];
 
   on_byPlayer_begincast(event) {
     const spellId = event.ability.guid;
-    if (!this.constructor.CANCELABLE_ABILITIES.includes(spellId)) {
+    if (this.constructor.IGNORED_ABILITIES.includes(spellId)) {
       return;
     }
-    this.beginCastTimestamp = event.timestamp;
-    this.castsStarted += 1;
+    if (this.wasCastStarted === true) {
+      this.castsCancelled += 1;
+    }
+    this.beginCastSpellId = event.ability.guid;
+    this.wasCastStarted = true;
   }
 
   on_byPlayer_cast(event) {
     const spellId = event.ability.guid;
-    const castTimestamp = event.timestamp;
-    const castTime = castTimestamp - this.beginCastTimestamp;
-    if (!this.constructor.CANCELABLE_ABILITIES.includes(spellId)) {
+    if (this.constructor.IGNORED_ABILITIES.includes(spellId)) {
       return;
     }
-    if (castTime < 100) {
-      this.castsStarted -= 1;
-    } else {
+    if (this.beginCastSpellId !== spellId && this.wasCastStarted === true) {
+      this.castsCancelled += 1;
+    }
+    if (this.beginCastSpellId === spellId && this.wasCastStarted === true) {
       this.castsFinished += 1;
     }
+    this.wasCastStarted = false;
   }
 
-  get cancelledCasts() {
-    return this.castsStarted - this.castsFinished;
+  get totalCasts() {
+    return this.castsCancelled + this.castsFinished;
   }
 
   on_finished() {
-    debug && console.log(formatMilliseconds(this.owner.fightDuration), 'Casts Started:', `${formatNumber(this.castsStarted)}`);
-    debug && console.log(formatMilliseconds(this.owner.fightDuration), 'Casts Completed:', `${formatNumber(this.castsFinished)}`);
+    debug && console.log(formatMilliseconds(this.owner.fightDuration), 'Casts Finished:', `${formatNumber(this.castsFinished)}`);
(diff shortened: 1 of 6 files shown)


Sharrq:
That could work.

So essentially we would just need to make a list of spells to ignore cast events for because they cast mid cast (Like Shimmer, Combustion, and Fire Blast).

Sharrq:
Works for me, ill see about switching that up tonight if i have time, otherwise ill do it tomorrow.

Sharrq:
Changed it to ignore specified spells instead of only check for specified spells. Also moved the spec file to the Mage Shared instead of Frost mage

src/Parser/Core/Modules/CancelledCasts.js
```diff
@@ -0,0 +1,79 @@
...
+
+  static IGNORED_ABILITIES = [
+  ];
+
+  on_byPlayer_begincast(event) {
+    const spellId = event.ability.guid;
+    if (this.constructor.IGNORED_ABILITIES.includes(spellId)) {
+      return;
+    }
+    if (this.wasCastStarted === true) {
```
kfinch: should be simply `if (this.wasCastStarted) {`

src/Parser/Core/Modules/CancelledCasts.js
```diff
@@ -0,0 +1,79 @@
...
+    this.beginCastSpellId = event.ability.guid;
+    this.wasCastStarted = true;
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (this.constructor.IGNORED_ABILITIES.includes(spellId)) {
+      return;
+    }
+    if (this.beginCastSpellId !== spellId && this.wasCastStarted === true) {
```
kfinch: `if (this.beginCastSpellId !== spellId && this.wasCastStarted) {`

src/Parser/Core/Modules/CancelledCasts.js
```diff
@@ -0,0 +1,79 @@
...
+      this.castsCancelled += 1;
+    }
+    if (this.beginCastSpellId === spellId && this.wasCastStarted === true) {
```
kfinch: `if (this.beginCastSpellId === spellId && this.wasCastStarted) {`

src/Parser/Core/Modules/CancelledCasts.js
```diff
@@ -0,0 +1,79 @@
...
+      this.castsCancelled += 1;
+    }
+    this.beginCastSpellId = event.ability.guid;
+    this.wasCastStarted = true;
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (this.constructor.IGNORED_ABILITIES.includes(spellId)) {
+      return;
```
kfinch: Technically, each ignored ability is still a finished cast, so you should increment the 'castsFinished' counter before returning.... or you could stop counting all instant casts because you have no choice but to finish them, maybe the more useful stat is the percentage of spells with cast times you finished.

Sharrq: Wouldn't incrementing the casts finished for ignored spells just skew the results considering we are ignoring those spells?

I tested this, if you go to events and only look for begin cast. The numbers match the modules total cats

kfinch: Depends on what you mean by "skewed"

Sharrq:
Yes I tested it. If you go to events and filter for event.type=="begincast" and count the results, the number matches the total casts.

Channels are .... Complicated. They all display their events differently. For example demon wrath for warlock is damage, cast, followed by x applybuff instances, and then damage. Drain soul is completely different. I don't think there is anything we can do to code around covering all of them

kfinch:
Sure, but we don't even look at damage and applybuff events. My main question was about making sure this CancelledCasts module doesn't do anything obviously wrong/stupid when confronted with a channel.

src/Parser/Core/Modules/CancelledCasts.js
Sharrq: Currently the module isn't counting instant casts, unless they are spells that have car times that are sometimes instant. Wouldn't incrementing competed/total casts just lower the cancelled percentage?  I'm already not counting instant casts just by excluding things without a begin cast, so I don't see the advantage of counting these instant casts

kfinch:
Oh, and so I don't have to make another ordering PR, could you make the statistics CORE(2) ?

Sharrq:
Ah. As far as I know it doesn't break anything when you use a channel, but it's hard to tell and test considering each channel is handled differently..

Sharrq:
Sent the fixes for the wasCastStarted stuff and the statistic priority thing.

Im waiting for the LFR queue to pop so i can test Ray of Frost to confirm that nothing gets screwed up because of that ... since its the only channel frost has.

src/Parser/Core/Modules/CancelledCasts.js
kfinch: Oh yeah I misread the logic

kfinch:
Ok, changes look good. I'll merge pending the Ray of Frost results

Sharrq:
Ray of Frost doesnt break anything. It doesnt have a begincast so it never gets counted. And the totalcasts was correct, 56 in events, 56 in the totalcasts.

Sharrq:
Was using this log for reference
https://www.warcraftlogs.com/reports/wJrmf9ka1Qx7b64v/

PR #697, opened 2017-10-31
[core] Fixed some issues in BaseHealerStatWeights

kfinch:
Made Leech stop counting for absorbs.

Fixed extraneous `1 *` in a calculation.

Fixed hasteHpct calculation.

### DIFF
--- a/src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
+++ b/src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
@@ -145,6 +145,9 @@ class BaseHealerStatWeights extends Analyzer {
   }
   _leech(event, healVal) {
     const spellId = event.ability.guid;
+    if (event.type !== 'heal') {
+      return; // leech doesn't proc from absorbs
+    }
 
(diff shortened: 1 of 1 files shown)


src/Parser/Core/Modules/Features/BaseHealerStatWeights.js
```diff
@@ -170,7 +173,7 @@ class BaseHealerStatWeights extends Analyzer {
     }
     const currInt = this.statTracker.currentIntellectRating;
     // noinspection PointlessArithmeticExpressionJS
-    const healIncreaseFromOneInt = (1 * ARMOR_INT_MULTIPLIER) / currInt;
+    const healIncreaseFromOneInt = ARMOR_INT_MULTIPLIER / currInt;
```
MartijnHols: That was on purpose since it's more technically right. :( I guess `ARMOR_INT_MULTIPLIER` would be 0.05 and we do `(1 + ARMOR_INT_MULTIPLIER)`

kfinch: I mean they're equally technically right.

PR #721, opened 2017-11-01
[Beast Mastery] BM Modules + other small things

Pewtro:
## Bigger things
Implemented 2 Bestial Wrath modules (one for contribution, one for uptime)
Implemented 2 Dire Beast modules (one for information, one for uptime)
Implemented Killer Cobra module
Implemented Qapla module
Implemented Titan's Thunder module
Implemented a simple Mark of the Claw uptime tracker

#### Minor things:
Fixes formatting in marksmanship modules
Fixes spelling error in ``MARK_OF_THE_ANCIENT_PRIESTESS``.


src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,119 @@
...
+
+let COOLDOWN_REDUCTION_MS = 12000;
+
+class DireBeast extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    spellUsable: SpellUsable,
+  };
+
+  DBCasts = 0;
```
MartijnHols: This should be just `casts`. Please don't include the name of the class in a property name (no `car.carSpeed`, just `speed` would be enough).

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,119 @@
...
+  effectiveBWReduction = 0;
+  wastedBWReduction = 0;
+  BestialWrathCast = 0;
```
MartijnHols: Please start properties with lower case characters.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,119 @@
...
+    if (this.combatants.selected.hasBuff(SPELLS.HUNTER_BM_T19_4P_BONUS.id)) {
+      COOLDOWN_REDUCTION_MS = 16000;
+    }
+    if (spellId === SPELLS.DIRE_BEAST.id) {
+      this.DBCasts += 1;
+      const bestialWrathIsOnCooldown = this.spellUsable.isOnCooldown(SPELLS.BESTIAL_WRATH.id);
+      if (bestialWrathIsOnCooldown) {
+        this.remainingBestialWrathCooldown = this.spellUsable.cooldownRemaining(SPELLS.BESTIAL_WRATH.id);
+        const reductionMs = this.spellUsable.reduceCooldown(SPELLS.BESTIAL_WRATH.id, COOLDOWN_REDUCTION_MS);
+        if (this.remainingBestialWrathCooldown > 3000) {
```
MartijnHols: Please extract this magic number.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,119 @@
...
+      this.DireBeastCastsSinceLastBW += 1;
+    }
+    if (spellId === SPELLS.BESTIAL_WRATH.id) {
+      this.BestialWrathCast = event.timestamp;
+      this.DireBeastCastsSinceLastBW = 0;
+    }
+  }
+
+  statistic() {
+    const gainedBestialWraths = this.effectiveBWReduction / 90000;
```
MartijnHols: Please extract this magic number.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrathUptime.js
```diff
@@ -0,0 +1,74 @@
...
+
+import SPELLS from 'common/SPELLS/index';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatisticBox from "Main/StatisticBox";
+import SpellIcon from "common/SpellIcon";
+import { formatPercentage, formatNumber } from "common/format";
+import Icon from "common/Icon";
+import SpellLink from "../../../../../../common/SpellLink";
```
MartijnHols: This import can be simpler: from 'common/SpellLink';

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
```diff
@@ -0,0 +1,95 @@
...
+      } else {
+        BESTIAL_WRATH_TOTAL_MODIFIER = BESTIAL_WRATH_BASE_MODIFIER;
+      }
+
+    }
+    if (spellId !== SPELLS.KILL_COMMAND.id || spellId !== SPELLS.COBRA_SHOT.id || spellId !== SPELLS.MULTISHOT.id) {
+      return;
+    }
+
+    if (this.combatants.selected.hasBuff(SPELLS.HUNTER_BM_T20_2P_BONUS.id) && this.combatants.selecteded.hasBuff(SPELLS.BESTIAL_WRATH.id)) {
```
MartijnHols: How about checking for the 2PT20 bonus in the on_initialized call? It doesn't change during the fight.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
```diff
@@ -0,0 +1,95 @@
...
+import getDamageBonus from "Parser/Hunter/Shared/Core/getDamageBonus";
+import StatisticBox from "Main/StatisticBox";
+import SpellIcon from "common/SpellIcon";
+import ITEMS from "common/ITEMS/HUNTER";
+
+const T204P_MODIFIER = 0.15;
+const T20_2P_MODIFIER_PR_STACK = 0.015;
+const BESTIAL_FURY_MODIFIER = 0.15;
+const BESTIAL_WRATH_BASE_MODIFIER = 0.25;
+let BESTIAL_WRATH_TOTAL_MODIFIER = 0;
```
MartijnHols: Shouldn't this be a class property since it's used by an instance of BestialWrath? Running multiple instances would run into problems.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
Pewtro: Why would I put it there? Would I simply say 
``if (this.combatants.selected.hasBuff(SPELLS.HUNTER_BM_T20_2P_BONUS.id) {
this.T20_2P_MODIFIER_PR_STACK = 0.015 
}`` and then otherwise have it set as 0?

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
Pewtro: I'm not sure as to what you mean, but I made a small change in how I utilise this in the latest push

src/Parser/Hunter/BeastMastery/Modules/Features/MarkOfTheClaw.js
```diff
@@ -0,0 +1,32 @@
...
+
+import SPELLS from 'common/SPELLS/index';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatisticBox from "Main/StatisticBox";
+import SpellIcon from "common/SpellIcon";
+import { formatPercentage } from "common/format";
+
+class MarkOfTheClaw extends Analyzer {
```
kfinch: Please reconsider this module. Mark of the Claw makes a very small contribution to damage, the uptime is completely outside of the player's control, and the enchant is not BeastMastery specific so I don't know why it's in the BeastMastery directory.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
kfinch: you should `let bestialWrathBonus` inside the class so it's an instance variable, and you should set its value exactly once in an `on_initialized()` function. Then probably also have a `currentBestialWrathBonus` value that's set to bestialWrathBonus on cast, and is then boosted throughout.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
```diff
@@ -0,0 +1,93 @@
...
+    const spellId = event.ability.guid;
+    //resets the strength of bestial wrath every time it's cast and adjust for Bestial Fury talent.
+    if (spellId === SPELLS.BESTIAL_WRATH.id) {
+      if (this.combatants.selected.hasTalent(SPELLS.BESTIAL_FURY_TALENT) || this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_HUNTMASTER.id)) {
+        BESTIAL_WRATH_TOTAL_MODIFIER = BESTIAL_FURY_MODIFIER + BESTIAL_WRATH_BASE_MODIFIER;
+      } else {
+        BESTIAL_WRATH_TOTAL_MODIFIER = BESTIAL_WRATH_BASE_MODIFIER;
+      }
+    }
+    if (spellId !== SPELLS.KILL_COMMAND.id || spellId !== SPELLS.COBRA_SHOT.id || spellId !== SPELLS.MULTISHOT.id) {
```
kfinch: I assume you defined `AFFECTED_ABILITIES` for this, remember to use it.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
```diff
@@ -0,0 +1,93 @@
...
+
+const AFFECTED_ABILITIES = [
+  SPELLS.KILL_COMMAND.id,
+  SPELLS.COBRA_SHOT.id,
+  SPELLS.MULTISHOT.id,
+];
+
+const debug = false;
+
+class BestialWrath extends Analyzer {
```
kfinch: This class is doing a whole lot, it's possible it should be split into several. What does the player gain from knowing the damage boost of using a baseline cooldown? I think you should split damage boost reports up by the different talents and tier bonuses that boost it, so the player knows what they're getting from each.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrathUptime.js
```diff
@@ -0,0 +1,74 @@
...
+              alt='Average Focus At Trueshot Cast'
+              style={{
+                height: '1.3em',
+                marginTop: '-.1em',
+              }}
+            />
+          </span>
+        )}
+        label={`Bestial Wrath information`}
+        tooltip={`You started your average Bestial Wrath window with ${averageFocusAtBW} focus, and you had an overall uptime of ${formatPercentage(this.percentUptime)}% on Bestial Wrath.`}
```
kfinch: Looks like you put two unrelated statistics in the same box, and it's causing display issues. As I suggested above, maybe drop the uptime display.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,122 @@
...
+        )}
+        label={`Direbeast information`}
+        tooltip={`You cast Dire Beast ${this.casts} times. <br/> <ul> <li>This reduced Bestial Wraths cooldown by ${(this.effectiveBWReduction / 1000).toFixed(1)} seconds in total, which resulted in you gaining ${formatNumber(gainedBestialWraths, 2)} extra Bestial Wrath casts. </li> <li>You lost out on ${this.wastedBWReduction / 1000} seconds of CD reduction by casting Dire Beast while Bestial Wrath wasn't on cooldown or while cooldown had less than ${COOLDOWN_REDUCTION_MS / 1000} seconds remaining. </li><li> You cast ${this.badDBCasts} Dire Beasts while there was less than 3 seconds remaining of Bestial Wrath cooldown.</li></ul>`}
+      />
+    );
+  }
+  suggestions(when) {
+    when(this.badDBCasts).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>Delay casting <SpellLink id={SPELLS.DIRE_BEAST.id} /> if there is less than 3 seconds cooldown remaining on <SpellLink id={SPELLS.BESTIAL_WRATH.id} />.</span>)
```
kfinch: It's not clear to me from reading relevant tooltips *why* it's so bad to cast Dire Beast with 3 seconds or less remaining on Bestial Wrath. If it's not clear to me it's probably also not clear to a newbie BM player looking for advice in suggestions. Perhaps give some more explanation.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
```diff
@@ -0,0 +1,122 @@
...
+              id={SPELLS.DIRE_BEAST.id}
+              style={{
+                height: '1.3em',
+                marginTop: '-.1em',
+                filter: 'grayscale(100%)',
+              }}
+            />
+          </span>
+        )}
+        label={`Direbeast information`}
```
kfinch: It looks again like you've clumped several semi related stats into the same box. This has forced you to hide stat meaning in the tooltip, which reduces readability.

You should probably decide which of these stats is the most important, and then keep only it while putting the other stats in the tooltip. Alternately, if they're all really important, you could put them in seperate stat boxes. 

The problem is that if they're so unrelated that the best label you can come up with is "Direbeast information", they probably shouldn't be together.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeastUptime.js
```diff
@@ -0,0 +1,31 @@
...
+    const uptime = this.combatants.selected.getBuffUptime(SPELLS.DIRE_BEAST_BUFF.id) / this.owner.fightDuration;
+    return uptime;
+  }
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.DIRE_BEAST.id} />}
+        value={`${formatPercentage(this.percentUptime)}%`}
+        label={`Dire Beast Uptime`}
+        tooltip={`If this tooltip shows over 100% uptime, it's because you on average had more than 1 Dire Beast buff active at all times. In actuality you had an average of ${(this.percentUptime).toFixed(2)} up throughout the fight.`}
```
kfinch: Why does the tooltip say `In actuality you had an average of ${(this.percentUptime).toFixed(2)}` when this.percentUptime is also the stat shown in the value? I don't understand what the difference is here.

src/Parser/Hunter/BeastMastery/Modules/Talents/KillerCobra.js
```diff
@@ -0,0 +1,57 @@
...
+    const spellId = event.ability.guid;
+    if (!this.combatants.selected.hasBuff(SPELLS.BESTIAL_WRATH.id)) {
+      return;
+    }
+    if (spellId !== SPELLS.COBRA_SHOT.id) {
+      return;
+    }
+    const killCommandIsOnCooldown = this.spellUsable.isOnCooldown(SPELLS.KILL_COMMAND.id);
+    if (killCommandIsOnCooldown) {
+      this.spellUsable.reduceCooldown(SPELLS.KILL_COMMAND.id, COOLDOWN_REDUCTION_MS);
```
kfinch: you can just use `this.spellUsable.endCooldown(SPELLS.KILL_COMMAND.id);` here, no need for the COOLDOWN_REDUCTION_MS stuff.

src/Parser/Hunter/BeastMastery/Modules/Traits/TitansThunder.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.TITANS_THUNDER.id) {
+      return;
+    }
+    debug && console.log(`stacks:`, this._currentStacks);
+    this.totalTTCasts += 1;
+    if (this.spellUsable.cooldownRemaining(SPELLS.BESTIAL_WRATH.id) < 30) {
```
kfinch: No magic numbers

src/Parser/Hunter/BeastMastery/Modules/Traits/TitansThunder.js
```diff
@@ -0,0 +1,132 @@
...
+                height: '1.3em',
+                marginTop: '-.1em',
+                filter: 'grayscale(100%)',
+              }}
+            />
+          </span>
+
+        )}
+        label={`Titan's Thunder information`}
+        tooltip={`You cast Titan's Thunder ${this.totalTTCasts} times, of which ${this.badTTCasts} were bad casts. <br/> Bad casts indicate that they were used without Dire Beasts up, or if using Dire Frenzy it was used without Bestial Wrath up. <br/> You cast Titan's Thunder ${this.shouldHaveSavedTT} times where you should have delayed casting it, this occurs when you cast Titan's Thunder when there is less than 30 seconds remaning on Bestial Wrath cooldown.`}
```
kfinch: Check your wording: ` Bad casts indicate that they were used without Dire Beasts up, or if using Dire Frenzy it was used without Bestial Wrath up.`

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
Pewtro: It breaks if I use it there like I use elsewhere, and I couldn't figure out why 🤔

kfinch: What's "it"? What line were you using, and what error did you get?

Pewtro: it being ``AFFECTED_ABILTIES``
line code used elsewhere where it works intentionally: 
``if (AFFECTED_ABILITIES.every(id => spellId !== id))``

kfinch: ok sure but what error do you get when you use it here?

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeastUptime.js
Pewtro: One shows an actual percentage number (say 170%), and one shows a regular number which in this case would be 1.7. 

It should probably say ``you had an average of X Dire Beasts up throughout the fight.``

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrath.js
Pewtro: I agree with the splitting it up, so I'll do that

src/Parser/Hunter/BeastMastery/Modules/Features/MarkOfTheClaw.js
Pewtro: Fair enough I guess

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrathUptime.js
Pewtro: I can split the two modules to showcase information better, but I think it's quite relevant to show uptime, as a large part of the spec revolves around reducing the cooldown of BestialWrath and optimising damage inside it, so having low uptime would suggest incorrect play.

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
Pewtro: I'll split it into two modules, to promote clarity

src/Parser/Hunter/BeastMastery/Modules/Spells/DireBeast/DireBeast.js
Pewtro: That makes sense, will add!

src/Parser/Hunter/BeastMastery/Modules/Talents/KillerCobra.js
Pewtro: Nice!

src/Parser/Hunter/BeastMastery/Modules/Features/MarkOfTheClaw.js
kfinch: It might be reasonable in the future to include throughput or "average stat gain" calculations for each of the neck enchants, but I don't think including a full sized statistic that doesn't expose any interesting information is valuable.

src/Parser/Hunter/BeastMastery/Modules/Spells/BestialWrath/BestialWrathUptime.js
kfinch: Well then isn't uptime just a poor proxy for cast efficiency in this case?

As for optimizing damage inside of Bestial Wrath, that does sound like a useful stats box.

Pewtro: I need to do some work on Cast Efficiency, but it's a bit of a pita because the spec has 2 untracked cooldown reducing/resetting events that happen often on spells that reduce Bestial Wrath cooldown (so cast efficiency on BW is currently saying stuff like 7/3). 

I asked the BM experts, and they'd like an uptime module atleast as long as Cast Efficiency is bonkered. 

I'm not sure if damage inside BW windows is possible without creating 3 or 4 versions because of different viable builds, will look into it.

Pewtro:
Should have handled most requests, sorry for the large PR 😓

src/Parser/Hunter/BeastMastery/Modules/Talents/BestialFury.js
```diff
@@ -0,0 +1,58 @@
...
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import { formatNumber } from 'common/format';
+import getDamageBonus from "Parser/Hunter/Shared/Core/getDamageBonus";
+import StatisticBox from "Main/StatisticBox";
+import SpellIcon from "common/SpellIcon";
+import ITEMS from "common/ITEMS/HUNTER";
+
+//Calculates the actual % increase in damage from Bestial Fury
+const BESTIAL_FURY_MODIFIER = 1 - (1 / 1.4 * 1.25);
```
kfinch: I think this calculation is wrong, this comes out to ~0.107, while I think the correct calculation is `(1.4 / 1.25) - 1`, which is 0.12.

I'll use the following terminology: a "multiplier" is the amount you multiply with damage to get new damage, e.g. 1.4, and a "bonus" is the amount of the multiplier over 100%, which is multiplier - 1, e.g. 0.4. The correct formula is `newMultiplier / oldMultiplier === relativeMultiplier === relativeBonus + 1`

If you want a more concrete example why my formula is correct: if X is the old multiplier and Y is the new multiplier, your formula is `1 - (1 / Y * X)`, mine is `(Y / X) - 1`. Lets consider X = 1.0 and Y = 1.5. I think we easily see without a formula that this is a relative increase of 50%, since "X = 1.0" means we started with no bonus at all. Your formula does `1 - (1 / 1.5 * 1) === 0.33...` mine does `(1.5 / 1) - 1 === 0.5`.

src/Parser/Hunter/BeastMastery/Modules/Items/Tier20_2p.js
```diff
@@ -0,0 +1,78 @@
...
+      this.currentStacks += 1;
+      debug && console.log('t20 buffer cast and stacks = ', this.currentStacks);
+    }
+  }
+
+  on_byPlayer_damage(event) {
+    if (!this.combatants.selected.hasBuff(SPELLS.BESTIAL_WRATH.id, event.timestamp)) {
+      return;
+    }
+    this.bonusDmg += getDamageBonus(event, (1 - ((1 / (1 + this.bestialWrathBaseModifier)) * (1 + (this.currentStacks * T20_2P_MODIFIER_PR_STACK)))));
```
kfinch: Pretty sure this calculation is incorrect (see my previous calculation comment). Pretty sure the correct calculation is `getDamageBonus(event, (this.currentStacks * T20_2P_MODIFIER_PR_STACK) / (1 + this.bestialWrathBaseModifier));`. Remember it's "newMultiplier / oldMultiplier === relativeMultiplier"

src/Parser/Hunter/BeastMastery/Modules/Items/Tier20_2p.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  on_byPlayerPet_damage(event) {
+    if (!this.combatants.selected.hasBuff(SPELLS.BESTIAL_WRATH.id, event.timestamp)) {
+      return;
+    }
+    this.bonusDmg += getDamageBonus(event, (1 - ((1 / (1 + this.bestialWrathBaseModifier)) * (1 + (this.currentStacks * T20_2P_MODIFIER_PR_STACK)))));
```
kfinch: See prev

src/Parser/Hunter/BeastMastery/Modules/Items/Tier20_2p.js
Pewtro: It needs to be ``getDamageBonus(event, (1+this.bestialWrathBaseModifier+(this.currentStacks * T20_2P_MODIFIER_PR_STACK)) / (1 + this.bestialWrathBaseModifier));`` surely then?

kfinch: Not quite, it would be that `- 1`. However, what I put it is the same thing in a reduced form. Sorry I should have mentioned I reduced it.

kfinch:
Thanks! Sorry I did a bad job of explaining the math, but I swear it's correct. I should clearly latex it out or something.

PR #728, opened 2017-11-03
Disc Legendary Estel Avg Haste

awlego:
This calculates the avg haste value the legendary Estel, Dhjahna's Inspiration gives a player as well as showing them their buff uptime.

I have to use a normalizer since sometimes the apply/remove buffs are the the wrong order.


src/Parser/Priest/Discipline/Modules/Items/Estel.js
```diff
@@ -0,0 +1,98 @@
...
+  get timePerHastePercentage() {
+    return Object.keys(this._timePerHastePercentage).map(key => this._timePerHastePercentage[key]);
+  }
+
+  on_byPlayer_applybuff(event) {
+    const spellId = event.ability.guid;
+
+    if (spellId === SPELLS.ESTEL_DEJAHNAS_INSPIRATION_BUFF.id) {
+      console.log(`num atonements found: ${this.atonementModule.numAtonementsActive}`);
+      this.lastHasteValue = this.atonementModule.numAtonementsActive;
```
karlpralow: The Number of Atonement module is to avoid since it's not precise.. (Maybe @rp4rk fixed it ?)

karlpralow:
What is the purpose of the Normalizer ?

src/Parser/Priest/Discipline/Normalizers/EstelNormalizer.js
```diff
@@ -0,0 +1,49 @@
...
+            haveOpenBuff = false;
+            fixedEvents.push(event);
+          }
+        } else {
+          if ((event.type) === "applybuff") {
+            haveOpenBuff = true;
+            fixedEvents.push(event);
+          } else if (event.type === "removebuff") {
+            // I have never seen this case.
+          }
```
karlpralow: This is useless code (else if { comment }).. I would remove this

src/Parser/Priest/Discipline/Normalizers/EstelNormalizer.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      } else {
+        fixedEvents.push(event);
```
karlpralow: Inversing your conditionals would reduce the nesting and make the code much clearer.

src/Parser/Priest/Discipline/Normalizers/EstelNormalizer.js
```diff
@@ -0,0 +1,49 @@
+import EventsNormalizer from 'Parser/Core/EventsNormalizer';
+
+import SPELLS from 'common/SPELLS';
+
+class EstelNormalizer extends EventsNormalizer {
```
karlpralow: What exactly is the purpose of this normalizer ?

src/Parser/Priest/Discipline/Modules/Items/Estel.js
karlpralow: Try to avoid using the atonement module, it does not always give the correct count unless @rp4rk fixed it

src/Parser/Priest/Discipline/Normalizers/EstelNormalizer.js
awlego: Which conditionals are you suggesting inverting?

src/Parser/Priest/Discipline/Modules/Items/Estel.js
awlego: I will look into this.

MartijnHols: @karlpralow I think that's a weird recommendation; having the best logic to track active Atonements in one module would seem better to me than having different Atonement tracking mechanisms across multiple modules. Even if the Atonement module is off, why would another implementation be better, and better yet why reinvent the wheel when you could fix the one you already have?

Issue #444, opened 2017-10-02
Auto select right fight & player if this is provided in the URL


MartijnHols:
Currently we only extract the report code from WCL URLs, picking up the fight id and player id could make importing even easier.

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=49980158)](https://www.bountysource.com/issues/49980158-auto-select-right-fight-player-if-this-is-provided-in-the-url?utm_source=49980158&utm_medium=shield&utm_campaign=ISSUE_BADGE)

awlego:
Opened a PR for this issue. See https://github.com/WoWAnalyzer/WoWAnalyzer/pull/729.

Pewtro:
Pretty sure this is already being done?

awlego:
Yes I did it a while ago. There in an upgrade to it that puts their name in
the URL instead of the player ID, but the actual auto selection is done.

-awlego / gao

PR #729, opened 2017-11-03
Auto Select Fight/Player if in WCL URL

awlego:
Implemented auto selecting of fight and player if provided in the passed URL.

### DIFF
--- a/src/Main/App.js
+++ b/src/Main/App.js
@@ -78,7 +78,11 @@ class App extends Component {
   }
 
   getPlayerFromReport(report, playerName) {
-    return report.friendlies.find(friendly => friendly.name === playerName);
+    const fetchByNameAttempt = report.friendlies.find(friendly => friendly.name === playerName);
+    if (!fetchByNameAttempt) {
+      return report.friendlies.find(friendly => friendly.id === parseInt(playerName, 10));
+    }
+    return fetchByNameAttempt;
   }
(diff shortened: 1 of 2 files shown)


src/Main/App.js
```diff
@@ -78,7 +78,11 @@ class App extends Component {
   }
 
   getPlayerFromReport(report, playerName) {
-    return report.friendlies.find(friendly => friendly.name === playerName);
+    const fetchByNameAttempt = report.friendlies.find(friendly => friendly.name === playerName);
+    if (!fetchByNameAttempt) {
+      return report.friendlies.find(friendly => friendly.id === parseInt(playerName, 10));
```
MartijnHols: `parseInt()` is outdated JS, please use `Number()` instead.

MartijnHols:
Fixes #444

src/Main/App.js
awlego: I originally was going to include fight name and player name, but it requires blocking on the report be loaded which is certainly possible, but felt less clean. I can certainly try that route again.

I'll update to use Number().

awlego:
Updated to use Number().

MartijnHols:
Thanks! I would prefer an approach that shows the fight and player names in the URL, but it's not required.

awlego:
Thanks for the approval. I have some free time this weekend so I'll see if I can get the actual names in another PR.

PR #735, opened 2017-11-03
[frost mage] Added support for Ice Time and Lady Vashj's Grasp legendaries

kfinch:

PR #738, opened 2017-11-04
[frost mage] Upgraded completeness to 'good'

kfinch:
@Sharrq and I believe Frost Mage completeness should be upgraded to 'Good'.

We've implemented the many suggestions from Mage discord, and we now offer statistics and suggestions for every facet of the core Frost Mage rotation. Analysis is provided for all reasonably analyzable talents. All legendaries are supported, along with t20 2set and 4set (pending the PRs listed below).

While t19 2set is still commonly used, there is unfortunately no reasonable way to make a statistic out of it.

We have not implemented a Stat Value module or a Relic Traits Value module, and this is an intentional decision. Frost Mage SimC is actively maintained and is the go-to gearing tool for the Mage community. Due to the heavily proc based nature of Frost Mage gameplay we believe sims will be able to provide better gearing advice than we could through logs analysis.

Depends on #737 #736 #735

### DIFF
--- a/src/Parser/Mage/Frost/CONFIG.js
+++ b/src/Parser/Mage/Frost/CONFIG.js
@@ -12,7 +12,7 @@ export default {
   spec: SPECS.FROST_MAGE,
   maintainer: '@Sharrq and @sref',
   maintainerAvatar: SharrqAvatar,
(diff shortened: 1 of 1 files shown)


MartijnHols:
Thanks for your hard work @kfinch and @Sharrq! This spec is well worked out 👍

PR #739, opened 2017-11-04
Shadowpriest ABC fix mostly

hasseboulen:


src/Parser/Core/Modules/Haste.js
```diff
@@ -37,6 +37,7 @@ class Haste extends Analyzer {
     [SPELLS.VOIDFORM_BUFF.id]: 0.01,
     [SPELLS.ICY_VEINS.id]: 0.3,
     [SPELLS.BONE_SHIELD.id]: 0.1, // Blood BK haste buff from maintaining boneshield
+    [SPELLS.MIND_QUICKENING.id]: (800 / 37500), // shadow priest trait
```
kfinch: Because this is a rating bonus and not a haste percent bonus, it's handled instead in StatTracker.

src/Parser/Priest/Shadow/Modules/Core/Haste.js
```diff
@@ -0,0 +1,80 @@
+import CoreHaste from 'Parser/Core/Modules/Haste';
+import SPELLS from 'common/SPELLS';
+
+const debug = false;
+
+const VOIDFORM_HASTE_PER_STACK = 0.01;
+const LINGERING_INSANITY_HASTE_PER_STACK = 0.01;
```
kfinch: Lingering Insanity tooltip just says the Voidform bonus fades over time, so you probably don't need two seperate constants here.

src/Parser/Priest/Shadow/Modules/Core/Haste.js
```diff
@@ -0,0 +1,80 @@
...
+
+
+class Haste extends CoreHaste {
```
kfinch: Maybe I'm missing something, but it's not clear to me why you need to extend CoreHaste. Why not just register Lingering Insanity and Voidform buffs in the core Haste module and let it handle things?

It looks like some manner of special mechanics happening here, I just don't understand what they are.

src/Parser/Priest/Shadow/Modules/Features/AlwaysBeCasting.js
```diff
@@ -4,11 +4,20 @@ import SpellLink from 'common/SpellLink';
 import { formatPercentage } from 'common/format';
 import { STATISTIC_ORDER } from 'Main/StatisticBox';
 
+import CoreAlwaysBeCasting from 'Parser/Core/Modules/AlwaysBeCasting';
 import SPELLS from 'common/SPELLS';
 
-import CoreAlwaysBeCasting from 'Parser/Core/Modules/AlwaysBeCasting';
+import Haste from '../Core/Haste';
+
+const debug = false;
```
kfinch: Value assigned but never used, Travis was yelling about this.

src/Parser/Priest/Shadow/Modules/Core/Haste.js
hasseboulen: I haven't successfully managed to get the Core Haste module correct with just adding the lingering and voidform buffs as this results in way to little haste overall. Most likely related to lingering insanity.

There is also a gap inbetween when you lose voidform and before lingering insanity is applied, due to not actually gaining a buff, but rather updating when you start to lose stacks.

src/Parser/Priest/Shadow/Modules/Core/Haste.js
hasseboulen: Yes, it was only added for clarity reasons. Can be removed if wanted.

src/Parser/Core/Modules/Haste.js
hasseboulen: Alright, I wondered a bit about stat bonuses. I will adjust.

MartijnHols:
Looks good, I think I broke this some time ago when implementing the buff stack Haste changes, sorry about that.

Could you revert your changes to `CONTRIBUTING.REVIEW.md` so that this can be merged?

hasseboulen:
I had some problems with removing changes to code-reviews.md. Closed the pull request until I have time to fix it.

PR #740, opened 2017-11-04
Add spell timeline to cast efficiency suggestions

MartijnHols:
image

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -8,6 +8,9 @@
     <inspection_tool class="JSConstructorReturnsPrimitive" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JSDuplicatedDeclaration" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JSMethodCanBeStatic" enabled="false" level="WARNING" enabled_by_default="false" />
+    <inspection_tool class="JSPotentiallyInvalidConstructorUsage" enabled="false" level="WARNING" enabled_by_default="false">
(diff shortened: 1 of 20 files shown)


MartijnHols:
`Is there any way to change the scaling of the timeline display? I'm not sure if the scaling should be dynamic or what, but the current static scaling is way too zoomed in. It's basically unusable for 60s+ CDs.`

Yes, but I don't really want to dynamically change it based on cooldown length and I think the current size is decent for shorter cooldowns. I reduced it by 20%, any more will require hiding every second second (teehee).

`The red marking doesn't seem to be working correctly, it's extending beyond when the CD is supposed to have ended. I know the CD is set correctly because the tooltip is showing the right number.`

Fixed it. Oops. Also new color since red was confusing.

`The 'Cooldown' tooltip is showing the CD to 3 decimal places, which seems excessive.`

image
I think it makes sense for cooldowns such as this, or that are reduced by 1.5s or something by other abilities. I know it's pretty useless for 30s/60s/static CDs but consistency again.

MartijnHols:
Changed the CD to 1 decimal. 👍

kfinch:
I was about to suggest one decimal but I see you've beaten me to it.

kfinch:
As for scaling, I think it might need to scale based on ability's cooldown? I first checked out this feature by opening up a Resto Druid log, and all their cast efficiency spells are long CD. The feature is unattractive and not user friendly when the spell's CD takes the fill width of the screen.

As for changing the red to a light gray, I think it looks too low contrast now? Or at least glancing at it it wasn't immediately clear to me which was 'on-CD' and which was 'off-CD'

MartijnHols:
Fixed both issues, thanks once again for the review!

PR #741, opened 2017-11-04
[resto druid] Reworked NaturesEssence

kfinch:
The NaturesEssence module was intended to make sure you were hitting enough injured targets with WildGrowth, but its method of detection had a high rate of false positives and false negatives.

I've used a completely different method that is more complicated but more directly measures what it's supposed to measure.

### DIFF
--- a/src/Parser/Druid/Restoration/Modules/Features/NaturesEssence.js
+++ b/src/Parser/Druid/Restoration/Modules/Features/NaturesEssence.js
@@ -6,46 +6,83 @@ import Analyzer from 'Parser/Core/Analyzer';
 import SPELLS from 'common/SPELLS';
 import Combatants from 'Parser/Core/Modules/Combatants';
 
-const MINOR = 0.07;
-const AVERAGE = 0.12;
-const MAJOR = 0.17;
+import SuggestionThresholds from '../../SuggestionThresholds';
+
+const HEAL_WINDOW_MS = 500;
+const RECOMMENDED_HIT_THRESHOLD = 3;
 
 class NaturesEssence extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
 
-  effectiveHealing = 0;
-  overHealing = 0;
+  casts = 0;
+  castsWithTargetsHit = []; // index is number of targets hit, value is number of casts that hit that many targets
+
+  castHits = 0;
+  totalHits = 0;
+  healTimestamp = undefined;
 
   on_initialized() {
     this.active = this.combatants.selected.traitsBySpellId[SPELLS.NATURES_ESSENCE_TRAIT.id] > 0;
   }
 
   on_byPlayer_heal(event) {
-    const spellId = event.ability.guid;
+    if (event.ability.guid !== SPELLS.NATURES_ESSENCE_DRUID.id) {
+      return;
+    }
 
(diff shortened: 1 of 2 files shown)


buimichael:
So basically you're changing the check on the amounts of OH on Nature's Essence to check the amounts of people a single cast of WG hit? I don't see why one would exclude the other, both are valid suggestions imo.

kfinch:
I think you're mixing up 'suggestion' and 'calculation'. Both the old calculation and my new calculation are done for the purpose of detecting if the player is casting WG at a good time and on good targets, and then offering a suggestion if they aren't. The relevant question is "which calculation is doing a better job at detecting if the player is casting WG at the right times?"

There are two deficiencies in the Nature's Essence overheal method that I'm seeking to fix.
1) When the player Wild Growths an isolated target, and there aren't enough nearby players to produce 6 HoTs. Lets look at the extreme example where only the primary (injured) target is hit. The Nature's Essence overheal method detects one Nature's Essence with 0% overheal, and thinks you've done a good job. My method detects 1 target hit, and correctly realizes that was a bad cast.
2) When the player is consistently hitting several wounded targets, but less than the full 6. For example consider one case where the player casts 6 WGs and hits 4 wounded targets each time. Those were all reasonable casts. Now consider the case where the player casts 6 WGs, hitting 6 targets 4 of the times and 0 targets 2 of the times. The two full overheal casts were bad. Both of these cases would show the same overheal on Nature's Essence, but my new calculation would be able to detect the two bad casts in the second case.

buimichael:
Yeah I like your calculation. But there's on scenario that i'd like detected as well and that is:

3. Consider the case where the player casts WGs and does hit your threshold of 3 targets each time, however these WGs were cast before damage occuring, thus making the initial burst of WG and Nature's essence overheal. In your implementation those WG would be marked as "good" WG's, which is not the case. I'd like to include this scenario in your calcs, if possible.

kfinch:
Yes, it would be great to detect when the most raiders are injured, but that requires another API call, which is frowned upon.

kfinch:
Put another way, "I'd like to include it too, but it's not possible"

buimichael:
That's not what I meant, you can keep the old implementation + the new one and merge them. All we need to check for is the amount overhealing nature's essence is doing. (The old implementation).
So :
if WG targets is bigger than 3, proceed to check if those 3+ hits were just overheals or not. If overheal add them to "bad WG usage" or add them to another variable to track the OH of those WGs

kfinch:
The new implementation does check Nature's Essence overheal, just in an 'all or nothing' manner. When the new method checks targets hit, it's counting a hit as "hit and didn't completely overheal".

Is your suggestion that I shouldn't go all or nothing, and should allow "partial credit" on Nature's Essence overheal? I can agree that adds a bit more accuracy in some cases, but I chose "all or nothing" because it was an easier metric to show to the player... basically I did it for user readability reasons.

buimichael:
I agree that your approach is 100% more reliable in the sense of checking what we're actually looking for, Although I think that we should try to implement the point I mentioned in my previous comments to also include accurate amounts of overhealing on natures essence/wg.

Either way this solution is better than the current implementation.

kfinch:
There is room for enhancement.. I was thinking about showing an 'effectiveness rating' for WG that combines overhealing with a consideration of how many targets were hit by WG... for now I'll merge this in though because it's better than what was there before.

PR #749, opened 2017-11-06
WW Energizing Elixir Resource gains (Chi calc'ed not shown)

terndrup:
Calculates the amount of energy and chi gained from the Energizing Elixir talent for Windwalker parses, using the energize event.

NOTE: The chi is calculated and available in the `chiGained` field but the engine is currently always returning a static amount of chi wasted on cast (incorrectly) and therefore it is not shown yet.

Coryn - Peak of Serenity

### DIFF
--- a/src/Parser/Monk/Windwalker/Modules/Talents/EnergizingElixir.js
+++ b/src/Parser/Monk/Windwalker/Modules/Talents/EnergizingElixir.js
@@ -20,8 +20,7 @@ class EnergizingElixir extends Analyzer {
   eeCasts = 0;
 
   on_initialized() {
-    this.active = this.combatants.selected.hasTalent(SPELLS.ENERGIZING_ELIXIR_TALENT
-      .id);
+    this.active = this.combatants.selected.hasTalent(SPELLS.ENERGIZING_ELIXIR_TALENT.id);
   }
 
   on_byPlayer_cast(event) {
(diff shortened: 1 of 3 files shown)


rp4rk:
Can you fix the lint error in this, please?

src/Parser/Monk/Windwalker/Modules/Talents/EnergizingElixir.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+  chiGained = 0;
+  energyGained = 0;
+  energyWasted = 0;
+  chiSaved = 0;
+  eeCasts = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.ENERGIZING_ELIXIR_TALENT
+      .id);
```
anom0ly: Style item - But can you move this all to a single line?

src/Parser/Monk/Windwalker/Modules/Talents/EnergizingElixir.js
```diff
@@ -0,0 +1,65 @@
...
+        this.energyGained += event.resourceChange - event.waste;
+      }
+      if (event.resourceChangeType === ResourceTypes.CHI) {
+        this.chiWasted += event.waste;
+        this.chiGained += event.resourceChange - event.waste;
+      }
+    }
+  }
+
+  statistic() {
```
anom0ly: For the statistics box, to keep this consistent with other specs, can you re-write it as the following:
```js
  statistic() {
    return (<StatisticBox icon={<SpellIcon id={SPELLS.ENERGIZING_ELIXIR_TALENT.id }/>}
      value={this.energyGained}
      label={(
        <dfn data-tip={`from ${this.eeCasts} Energizing Elixir Casts`}>
          Energy gained
        </dfn>
        )}
        />
      );
    }
  }
```

This just moves the casts portion to the tooltip.

terndrup: Will do, thanks!

terndrup:
Made the requested changes and updated Changelog 👍

PR #750, opened 2017-11-06
Shadow priest fixes, mostly ABC module

hasseboulen:
Remade the pull request.
Old request can be found here: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/739

Already reviewed and should be good to go.


hasseboulen:
Ah, good call! I will change that in a near PR.

PR #751, opened 2017-11-06
Added suggestions on mana levels on encounter finished. Fixes #82

### LINKED ISSUE #82: Add mana usage to suggestions (suggestion on mana graph)
Right now we have the mana graph, which is great.

Would be nice to add a suggestion with the various severity levels if you finished with too much mana, as some may not check that tab at all.

buimichael:
[Screenshot from Gyazo](https://gyazo.com/1848f79eff7d6154832cc1b2eabeca16)

(Ignore the branch name, I got the issue numbers mixed up). This is not 100% accurate since it gets the endingMana from the last cast event from the player, so the endingMana variable can be potentially higher (if the player didn't cast anything in the last X seconds of the encounter). 

This can be solved by calculating the players MP5 (or mana gained per second) and checking the time differences between the last cast event and the on_finished timestamp, and calculating the players endingMana. However this is not an easy task as there's many edge cases to consider that may alter the MP5 (unless there's a property for that). (Such as items, traits, spells etc.)

### DIFF
--- a/src/Parser/Core/Modules/ManaValues.js
+++ b/src/Parser/Core/Modules/ManaValues.js
@@ -1,10 +1,21 @@
+import React from 'react';
+
 import Analyzer from 'Parser/Core/Analyzer';
 import RESOURCE_TYPES from 'common/RESOURCE_TYPES';
+import { formatPercentage } from 'common/format';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import ROLES from 'common/ROLES';
 
 class ManaValues extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  }
+  static SUGGESTION_MANA_BREAKPOINT = 0.1;
+
   lowestMana = null; // start at `null` and fill it with the first value to account for users starting at a non-default amount for whatever reason
   endingMana = 0;
 
+  baseMana = 110000;
   manaUpdates = [];
 
   on_byPlayer_cast(event) {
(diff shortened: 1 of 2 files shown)


buimichael:
Mana thresholds are Major > 30%, Regular > 20%, Minor >10%

buimichael:
@kfinch Omitting that line is good enough?

src/Parser/Core/Modules/ManaValues.js
```diff
@@ -26,9 +37,29 @@ class ManaValues extends Analyzer {
...
             used: manaCost,
           });
+          // The variable 'max' is constant but can differentiate by racial/items.
+          this.baseMana = max;
         });
     }
   }
+
+  suggestions(when) {
+    if(this.combatants.selected.spec.role !== ROLES.HEALER) {
```
MartijnHols: I think this should be in the on_initialized method, with this suggestion being its own Analyzer (so adding a new analyzer with just this suggestion and this.active set with this conditional). More SRP.

src/Parser/Core/Modules/ManaValues.js
```diff
@@ -26,9 +37,29 @@ class ManaValues extends Analyzer {
             max: max,
             used: manaCost,
           });
+          // The variable 'max' is constant but can differentiate by racial/items.
+          this.baseMana = max;
```
MartijnHols: I'm unsure if calling the *max* amount of mana "baseMana" is going to always be right. How about actually naming this `max`?

src/Parser/Core/Modules/ManaValues.js
```diff
@@ -26,9 +37,29 @@ class ManaValues extends Analyzer {
...
+
+  suggestions(when) {
+    if(this.combatants.selected.spec.role !== ROLES.HEALER) {
+      return;
+    }
+    const manaPercentageLeft = this.endingMana/this.baseMana;
+
+    when(manaPercentageLeft).isGreaterThan(this.constructor.SUGGESTION_MANA_BREAKPOINT)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>You ended the fight with too much mana. If possible, try to deplete all your mana when the boss dies. A good rule of thumb is having the same mana % as the bosses health %.
```
MartijnHols: `You ended the fight with too much mana` How about `You had mana left at the end of the fight.`? Technically you can never have too much mana.

Also I'd prefer just writing out *percentage* instead of %, or how about "levels"?

src/Parser/Core/Modules/ManaValues.js
```diff
@@ -26,9 +37,29 @@ class ManaValues extends Analyzer {
...
+          Mana is indirectly tied with healing throughput and should be optimized. </span>)
+          .icon('inv_elemental_mote_mana')
+          .actual(`${formatPercentage(manaPercentageLeft)}% (${this.endingMana}) mana left`)
```
MartijnHols: `endingMana` should have a formatNumber or formatTHousands I think

kfinch:
@buimichael yeah just omitting it works for me

buimichael:
Made some changes according to the requested changes.

PR #753, opened 2017-11-07
[Windwalker] Added CJL damage from SEF clones in The Emperors Capacity item breakdown

Juko8:


src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -65,6 +68,21 @@ class TheEmperorsCapacitor extends Analyzer {
...
 
+  _isPermanentPet(guid) {
+      return Object.keys(PETS).map(key => PETS[key]).every(pet => pet.id !== guid);
+  }
+
+  on_byPlayerPet_damage(event) {
+      const pet = this.pets.getSourceEntity(event);
+      const spellId = event.ability.guid;
+      if (!this._isPermanentPet(pet.guid)) {
+         return;
```
kfinch: Can you explain why you go to the trouble of excluding some pets here? Does any pet other than SEF clones use CJL? If another pet used CJL wouldn't you want it included?

src/Parser/Monk/Windwalker/Modules/Items/TheEmperorsCapacitor.js
```diff
@@ -65,6 +68,21 @@ class TheEmperorsCapacitor extends Analyzer {
     }
   }
 
+  _isPermanentPet(guid) {
+      return Object.keys(PETS).map(key => PETS[key]).every(pet => pet.id !== guid);
```
kfinch: it looks like the only pets listed in the PETS object are Warlock and Priest things, so won't this always return true for a Monk's pet? Also why is this function called `_isPermanentPet`? I really don't get what this is for.

PR #754, opened 2017-11-07
[core] Added handling for Haste rating changes to Haste analyzer

kfinch:
Handling for rating changes had previously been moved to the StatTracker module, but this meant that the Haste module no longer was tracking haste rating changes at all.

I've modified StatTracker so it generates a fabricated event any time stats change, and I've modified Haste to listen for those events and apply the appropriate change (including firing its own changehaste event).

It's plausible that the Haste module as a whole is due for a refactor, but in this PR I make the 'lightest touch' change I can while still implementing the desired functionality in a non-hacky manner.

### DIFF
--- a/src/Parser/Core/Modules/StatTracker.js
+++ b/src/Parser/Core/Modules/StatTracker.js
@@ -294,24 +294,52 @@ class StatTracker extends Analyzer {
         debug && console.log(`StatTracker prepull application IGNORED for ${SPELLS[spellId] ? SPELLS[spellId].name : spellId}`);
         return;
       }
-      debug && console.log(`StatTracker: (${event.oldStacks} -> ${event.newStacks}) ${SPELLS[spellId] ? SPELLS[spellId].name : spellId}`);
-      this._changeStats(statBuff, event.newStacks - event.oldStacks);
+
+      const before = this._stats;
+      const delta = this._changeStats(statBuff, event.newStacks - event.oldStacks);
+      const after = this._stats;
+      this._triggerChangeStats(event, before, delta, after);
+      debug && console.log(`StatTracker: (${event.oldStacks} -> ${event.newStacks}) ${SPELLS[spellId] ? SPELLS[spellId].name : spellId} - Change: ${this._statPrint(delta)}`);
+      debug && this._debugPrintStats(this._stats);
     }
   }
 
   _changeStats(change, factor) {
-    this._stats.strength += this._getBuffValue(change, change.strength) * factor;
-    this._stats.agility += this._getBuffValue(change, change.agility) * factor;
-    this._stats.intellect += this._getBuffValue(change, change.intellect) * factor;
-    this._stats.stamina += this._getBuffValue(change, change.stamina) * factor;
-    this._stats.crit += this._getBuffValue(change, change.crit) * factor;
-    this._stats.haste += this._getBuffValue(change, change.haste) * factor;
-    this._stats.mastery += this._getBuffValue(change, change.mastery) * factor;
-    this._stats.versatility += this._getBuffValue(change, change.versatility) * factor;
-    this._stats.avoidance += this._getBuffValue(change, change.avoidance) * factor;
-    this._stats.leech += this._getBuffValue(change, change.leech) * factor;
-    this._stats.speed += this._getBuffValue(change, change.speed) * factor;
-    debug && this._debugPrintStats();
+    const delta = {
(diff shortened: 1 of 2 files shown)


src/Parser/Core/Modules/StatTracker.js
```diff
@@ -252,28 +252,55 @@ class StatTracker extends Analyzer {
   }
 
   /*
+   * Helpers to get stat percentage from an arbitrary rating.
+   */
+   getCritPercentage(rating) {
+     return this.baseCritPercentage + (rating / this.critRatingPerPercent);
```
MartijnHols: I'm not sure about including the base in this as when just reading the method name I'd expect it to do just rating / ratingPerOnePercent and the base would come as a surprise. How about moving this to the methods below?

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -294,24 +321,52 @@ class StatTracker extends Analyzer {
...
+   */
+  _triggerChangeStats(event, before, change, after) {
+    this.owner.triggerEvent('changestats', {
+      timestamp: event ? event.timestamp : this.owner.currentTimestamp,
+      type: 'changestats',
+      sourceID: event ? event.sourceID : this.owner.playerId,
+      targetID: this.owner.playerId,
+      reason: event,
+      before,
+      change,
```
MartijnHols: How about naming this `delta` also?

src/Parser/Core/Modules/StatTracker.js
kfinch: It made sense to me to be able to get a stat percentage given an arbitrary rating, and you can see I actually use the `getHastePercentage` function in the Haste module. Which methods below are you suggesting I move to? I'm not sure what you're recommending here.

src/Parser/Core/Modules/StatTracker.js
kfinch: Hah, I was thinking you were going to suggest making all of them 'change'. I chose `change` over `delta` for the outward facing field because I felt it was easier to understand.

MartijnHols: Delta is used by browser for things like this, while `change` is usually a method.

src/Parser/Core/Modules/StatTracker.js
MartijnHols: current* methods

src/Parser/Core/Modules/StatTracker.js
kfinch: Alright I'll change it

src/Parser/Core/Modules/StatTracker.js
kfinch: I'm still not sure what you mean... are you saying something like `get currentCritPercentage(rating = this.currentCritRating) {`? That doesn't work because `get` methods can't have any arguements.

MartijnHols: I guess it doesn't matter much as I've been meaning to suggest some other changes to the stat handling that should also make this easier to solve.

Let's focus on crit since it has all the interesting bits.

Right now we have:
- startingCritRating - rating: the crit at pull including any buffs a player had then (base + gear + rating buffs)
- currentCritRating - rating: base + gear + rating buffs
- baseCritPercentage - percentage: base
- critRatingPerPercent - rating: rating per one percent
- getCritPercentage(value) - percentage: base + value
- currentCritPercentage - percentage: base + gear + rating buffs

How about changing this to:
- baseCritRating or baseCritPercentage (rating would make things simpler so we wouldn't have to mix percentage and rating behaviors more than once)
- gearCritRating
- buffCritRating
- currentCritRating - total current rating: base + gear + buff
- critRatingPerPercent - rating: rating per one percent
- getCritPercentage(value) - percentage: value / critRatingPerPercent
- currentCritPercentage - currentCritRating + crit percentage buffs

In this example I consider Haste to behave the same, since Haste does `current * (1 + gain) + gain` I don't think we could get a meaningful `buffCritPercentage` property so I left it out for all stats.

kfinch: I made this PR small because I was trying to connect StatTracker to Haste *without* an involved refactor. I've reverted the StatTracker so there are no longer getStatPercentage(value) functions, and I now use the existing suite to get the same behavior in Haste.

If you think StatTracker's API could be improved, you're welcome to open a PR for it.

PR #761, opened 2017-11-08
Focustracker performance fix

leapis:
I figure it's probably easier to tell me what's stylistically wrong with this when you can see the code yourself. The issue was with extrapolateFocus(), as it was called after every event- now it only iterates appropriate indexes each call, instead of the entire thing. Vastly decreased load times on my machine.

### DIFF
--- a/src/Parser/Hunter/Shared/FocusChart/FocusTracker.js
+++ b/src/Parser/Hunter/Shared/FocusChart/FocusTracker.js
@@ -8,6 +8,7 @@ class FocusTracker extends Analyzer {
     combatants: Combatants,
   }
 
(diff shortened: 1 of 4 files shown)


src/Parser/Hunter/Shared/FocusChart/FocusTracker.js
```diff
@@ -8,6 +8,8 @@ class FocusTracker extends Analyzer {
     combatants: Combatants,
   }
 
+  lastEventTimestamp = 0;
+  isFinished = false;
```
MartijnHols: This is unused.

Issue #391, opened 2017-09-26
Adding Boss features info


Iskalla:
When reviewing the Completeness of Moonkin, Zero came across a log with a green parse, that only had one problem: downtime.
After analyzing the log, its obvious that the main problems are lack of proper legendaries for the fight and lack of tier sets.

I've been thinking about a possible solution to reflect this in the analysis that requires as little maintenance each tier as possible.
Best idea I've come up with is to have some sort of static info on encounters, with specific parameters, that would allow us to somehow check for best leggos, tier sets (ofc would have to make it clear that this is based on general features of the encounter and won't be 100% exact at all times). This would only have to be changed once per tier.

Once this is done, an analysis for optimal (or usable) talents / legendaries should be easy to implement - maybe add a tab about it? -, tier sets would require more each-tier-coding, but maybe there's a cleaner solution, if someone came up with something...

For now I've thought of something like this:

![image](https://user-images.githubusercontent.com/11649446/30848509-9e4d3a22-a29f-11e7-90fa-95e0c1eb37c2.png)

With that info I could make something work for moonkin, idk if other classes would need more info.
A bit of feedback and any suggestion on what else to add would be appreciated

This relates directly to issue #16

MartijnHols:
This looks exactly like what I had in mind! 👌 A few questions and suggestions mixed:

 - How about changing `singleTarget` to targetType which is an enum? Would make it more obvious that single target means it's not considered an AoE boss (true or false isn't as explicit).
 - Is addWaves the same as meaning a AoE boss?
 - Should this include a cleave property?
 - What determines add health; at what point is it low, medium or high?
 - How could secondaryPrioTarget be used?
 - What determines high, medium or low movement?

Also it would be nice if the boss ids object was used for boss ids and the difficulties were via an enum.

Iskalla:
> How about changing singleTarget to targetType which is an enum? Would make it more obvious that single target means it's not considered an AoE boss (true or false isn't as explicit).
> Also it would be nice if the boss ids object was used for boss ids and the difficulties were via an enum.

Yeah, all of this sounds better :)


> Is addWaves the same as meaning a AoE boss?

Not necessarily. An AoE boss is a boss where you spend most of the time hitting more than one target. 
A good example is Mistress: It's a ST boss with add waves, you need to dps adds at certain points of the fights, but you are most of the time doing ST damage to the boss
[...]

PR #765, opened 2017-11-09
[Frost Mage] Minor Blizzard Suggestion

Sharrq:
This adds a suggestion if you are on a fight that is primarily or partially AoE and you never cast Blizzard. This is meant to be a start to evaluating talents and spell cast choices based on the current fight ... For example, telling the player that they took a talent that gave them no benefit because its a Single Target fight, etc, etc.

I also plan on doing more with the Blizzard Module in the future, but getting this working opens up some doors to do further evaluation on the other modules, so I would like to get this in first, so i can do the others. At some point ill circle back to this and add more analysis as far as Blizzard Usage and Damage.

### DIFF
--- a/src/Parser/Mage/Frost/CHANGELOG.js
+++ b/src/Parser/Mage/Frost/CHANGELOG.js
@@ -1,4 +1,5 @@
 export default `
+08-11-2017 - Added Suggestion for casting Blizzard on AoE fights. (by Sharrq)
 04-11-2017 - Reworked and updated Glacial Spike module. (by Sref)
 04-11-2017 - Added support for T20 4set. (by Sref)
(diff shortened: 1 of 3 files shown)


kfinch:
Enabling such a module is part of a larger conversation about if we should report differently depending on the encounter. If we were to implement something like this, the "AOE fights" part would probably want to be core.

I don't think this should be merged as is, needs more discussion on if we actually go down the road of fight specific analysis.

MartijnHols:
Please see https://github.com/WoWAnalyzer/WoWAnalyzer/issues/391 and help us figure out how to solve this everywhere. I mostly agree with @kfinch. I'm not yet sure if we should let all spec maintainers implement a different solution for this or if we should wait for a core-solution.

PR #768, opened 2017-11-09
[Unholy Death Knight] updated changelog to new format

jjs451:


jjs451:
I made a horrible mistake, fixed it, and now the site doesn't crash when you try to open the changelog

PR #786, opened 2017-11-11
[Frost Mage] Misc Additions/Fixes

Sharrq:
Added a damage counter to show how much extra damage the player got from Double Ice Lance. Also modified the cast time on Ebonbolt to include the Cast Time of the spell. That way the Cast Efficiency can be a little more accurate when it is determining how many casts were possible.

These changes were per some feedback from the Mage Discord

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Features/CastEfficiency.js
+++ b/src/Parser/Mage/Frost/Modules/Features/CastEfficiency.js
@@ -16,7 +16,7 @@ class CastEfficiency extends CoreCastEfficiency {
     {
       spell: SPELLS.EBONBOLT,
       category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
-      getCooldown: haste => 45,
+      getCooldown: haste => 45 + (3 / (1+haste)),
       recommendedCastEfficiency: 0.90,
     },
     {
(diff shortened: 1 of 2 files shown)


Sharrq:
The only warning that is showing up in my console is in the Brewmaster Config

console

kfinch:
While your implementations look fine, I don't really think either should be added.

For the double Ice Lance thing, the damage boost from it isn't really the number listed, it's the number listed minus the opportunity cost of that global, because if you hadn't double ice lanced you'd would have done something else with that global instead. This "damage vs replacement" problem is the reason why I haven't shown damage numbers for other things, like Brain Freeze.

The Ebonbolt thing is a good call, but it doesn't sit right with me that we're listing the wrong cooldown. Is there any other precedent for cooldown'd spells with cast times, and how are they handled?

Sharrq:
I only really added the double il damage thing because it was suggested from the mage discord. I can remove it, i didnt feel that strongly for it anyway.

The Ebonbolt thing though ... I agree that a separate setting for it would be better, but i dont really know how to go about implementing that. I could take a look at it and see if i can figure it out ... but im not sure if it would just get removed if/when Cast Effiency gets combined with SpellUsable? So i opted for doing it this way with the expectation that when cast efficiency gets redone that ill have to probably change it ..... but that could be wrong.

kfinch:
Yeah, remove the double ice lance damage thing. If the guy in Mage discord complains about it, send him to me and I'll explain in detail why it shouldn't be included.

As for Ebonbolt, I think you're right it can be left as is. Just add a comment explaining what you've done.

kfinch:
Er, when I say "left as is" I mean with the cast time included like you've done

Sharrq:
Removed the Double IL Damage Counter and added a comment for the Ebonbolt Change

PR #796, opened 2017-11-12
[core] Updated appearance of DamageDone and HealingDone tooltips

kfinch:
Changed the 'totals' part of each tooltip to be more concisely worded. For the 'pets contribution' part of DamageDone, updated the display and also made it only show if the player had >0 pet damage.


kfinch:
total-damage

total-healing

PR #797, opened 2017-11-12
[Marksmanship] Improvements and new suggestions

### LINKED ISSUE #703: [Marksmanship] A Murder of Crows module
Update a A Murder Of Crows module to handle casts properly in terms of boss health (don't use sub 25%, rather save for 20% of boss) --> show bullseye stacks gained from AMOC? 

Make this be the Cast Efficiency module and remove AMoC Cast Efficiency suggestions and tie them to this module instead.

### LINKED ISSUE #704: [Marksmanship] Trueshot in Execute
Add check for trueshot in execute (should probably be combined with the bullseye tracker)

Pewtro:
Some general improvements (and config update to reflect who is actively maintaining aswell as some updated text in it) and closes two issues (trueshot in execute and aMoC saving)

Fixes #704 
Fixes #703


src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
```diff
@@ -109,6 +113,14 @@ class Trueshot extends Analyzer {
           .regular(recommended - 10)
           .major(recommended - 20);
       });
+    when(this.executeTrueshots).isLessThan(1)
+      .addSuggestion((suggest, actual, recommended) => {
+      return suggest(<span>You should make sure to have atleast 1 <SpellLink id={SPELLS.TRUESHOT.id}/> cast during execute (where you are buffed by <SpellLink id={SPELLS.BULLSEYE_TRAIT.id}/>) to get the most out of <SpellLink id={SPELLS.TRUESHOT.id}/> as possible.</span>)
```
kfinch: "to get the most out of Trueshot as possible" doesn't make sense. Should be either "to get the most out of Trueshot", or "to get as much out of Trueshot as possible".

src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js
```diff
@@ -7,27 +7,51 @@ import SPELLS from 'common/SPELLS';
...
     }
     this.damage += event.amount;
   }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.A_MURDER_OF_CROWS_SPELL.id) {
+      return;
+    }
+    if (event.targetInstance === undefined && event.maxHitPoints && this.bossIDs.indexOf(event.targetID) !== -1) {
```
kfinch: Why checking for no targetInstance and also if target is a boss? Are there any bosses with more than one instance?

Also, the last condition should be `this.bossIDs.includes(event.targetID)`

src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js
```diff
@@ -37,6 +61,17 @@ class AMurderOfCrows extends Analyzer {
       />
     );
   }
+  suggestions(when) {
+    when(this.shouldHaveSaved).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>You should generally save <SpellLink id={SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id} /> when the boss has under 25% hp so that it is ready to use when the boss hits 20% and you can start getting <SpellLink id={SPELLS.BULLSEYE_TRAIT.id}/> quicker.</span>)
```
kfinch: "Don't cast between 20 and 25%" is a strangely specific suggestion, I hope that it's relevant as you think. I can see from the descriptions of Bullseye and Murder of Crows that this probably makes sense, tho.

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
Pewtro: Agreed

src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js
Pewtro: Checks if target is boss to ensure it doesn't give false flags if the player is using on an add that has between that amount of health (which could be fine - especially if the mob then dies within 15 seconds so that crows is reset) 

Should probably work without the targetInstance check 🤔 I shamelessly stole it from ``Parser/Hunter/Marksmanship/Modules/Traits/Bullseye.js`` where part of this code was originally in until we figured out how we wanted to do it

src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js
Pewtro: Yeah it is very specific - but it's a general rule of thumb, that helps the player quickly rack up bullseye stacks which is a large part of our damage. (Each tick adds a stack for a total of 16 stacks over 15 seconds (or until 30 in total)). 

I'll degrade it to be average importance and emphasize generally.

kfinch: Ehh you don't have to change it, was just making sure it needed to be that specific

src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js
```diff
@@ -7,27 +7,55 @@ import SPELLS from 'common/SPELLS';
 import StatisticBox from "Main/StatisticBox";
 import SpellIcon from "common/SpellIcon";
 import { formatNumber } from "common/format";
+import SpellLink from "common/SpellLink";
+import { formatPercentage } from "../../../../../common/format";
```
kfinch: Imports from out of spec directory should always be absolute path, not relative.

Pewtro: auto import reeeeee hahaha

kfinch: H E L P I N G

PR #798, opened 2017-11-12
[Marksmanship] upgraded spec completeness

Pewtro:
I believe the Marksmanship is in a really good state, and is a potential candidate for "great" status, however as long as a few bugs exist and the implementation of the last legendaries (regardless of useless/unused they are), I won't push for that.

We've implemented many suggestions from the hunter discord, and the module now offers statistics, metrics and suggestions for everything relevant for Marksmanship hunters. 

Depends on #797


MartijnHols:
Thanks!

ps. please migrate the changelog to the new format.

PR #799, opened 2017-11-13
Cast Efficiency / Spell Usable merge

### LINKED ISSUE #543: Connecting CastEfficiency to SpellUsable
CastEfficiency also needs to use SpellUsable in one way or another for calculating the max possible casts of a spell. Making modules interact with SpellUsable to reduce/reset the cd of spells is very easy and if CastEfficiency relies on that we could get pretty accurate max cast amounts. Difficulty with this lies with SpellUsable only really tracking spells that are on cooldown while an ABC module would have to estimate every time it could have gone on cooldown. 

The solution I had in mind is to have a module simulate casts from the start of the fight. Each event look if an ability was off cooldown and if so simulate a cast, then the total possible casts would be the new max casts.
In order to account for cooldown resets, resetCooldown needs to no longer throw an error when an ability is already on cooldown so implementers of it can call it even when it is on cooldown so our max possible casts would also take that into account. But maybe we shouldn't do this.

### LINKED ISSUE #611: Move abilities out of cast efficiency
I want to move the CPM_ABILITIES (and rename it) out of Cast Efficiency since it would also be good to use for SpellUsable and with #612 AlwaysBeCasting. Having it in CastEfficiency makes usage hard due to circular dependencies, and makes no sense anymore.

In fact I think if we add a isOnGlobalCooldown property it could also be used for ABC. It also needs to be extendable by core modules such as the Deceiver's Grand Design item module.

kfinch:
I've put up this PR in an incomplete state so I can more easily get comment / review, DO NOT MERGE THIS.

This is an attempt to get Cast Efficiency reports to use data from SpellUsable. While I haven't done the big rename yet, the intent is that all current CastEfficiency classes be renamed to 'CastInfo', and the new CastEfficiencyDisplay class added be renamed to simply 'CastEfficiency '.

For notes on design decisions / challenges in implementation, see the comments in CastEfficiencyDisplay.

Will fix #543 and fix #611


kfinch:
Some notes / discussion of the new cast efficiency calculation:

My new method calculates cast efficiency as the percentage of time a spell is on cooldown. The method on live calculates it as casts / "rawMaxCasts". These method are very similar, but for reasons I'll soon explain the new method returns slightly to moderately lower efficiency, particularly when a spell's cooldown is a large percentage of the fight duration.

Lets compare a situation, and show the calculation using the new method and the live method: Cooldown is 1 minute, Fight lasted 2:30, player casted spell at 0:10 and 2:00.
- Live method: ```
casts === 2
rawMaxCasts === fightDuration / cooldown === 150 / 60 = 2.5
maxCasts === ceil(rawMaxCasts) === 3
castEffic === casts / rawMaxCasts === 2 / 2.5 = 80%
userFacingResult === "2 / 3 casts (80% efficiency)"
```
- New method: ```
casts === 2
maxCasts === ceil(fightDuration / averageCooldown) === 3
timeOnCooldown === 90  // from 0:10 to 1:10, and from 2:00 to 2:30, when the fight ended...
castEffic === timeOnCooldown / fightDuration === 90 / 150 === 60%
userFacingResult === "2 / 3 casts (60% efficiency)"
```

The critical difference here is that the new method doesn't give credit for spell cooling down after the fight has ended. I think there's a strong argument that the new way is a better way of capturing efficiency, but I wanted to put them both out there for discussion.

MartijnHols:
The reason for the old approach was to allow for some slack if there was very little time for another cast. A 2 minute cooldown in a 2:20 fight *could* be cast twice but you generally don't want to as that would be inefficient as there would be nothing to heal. Old: 85.7% cast efficiency, new: below default recommended threshold.

kfinch:
@MartijnHols  That logic is sensible when discussing healing cooldowns, but is not sensible when discussing DPS cooldowns. Wouldn't it make sense to adopt this new calculation, and simply lower the thresholds for big healing cooldowns?

kfinch:
Some notes on how the fields of CPM_ABILITIES will be handled in new system:
- spell / name / category / isActive : same as before
- getCooldown : called during analysis by SpellUsable, meaning conditional cooldowns (like PW:S with and without rapture) can be handled here if desired
- charges : same as before, needed by SpellUsable.
- recommendedCastEfficiency / noCanBeImproved / noSuggestion / extraSuggestion / isUndetectable / importance : same as before
- getCasts : currently used in two circumstances: to break up buffed / unbuffed versions of basic spells in the cast efficiency chart, and to correct the cast count for Penance, which shows a cast event per bolt. The cast count isn't actually used or needed by the new cast efficiency calculation, so my current solution is to accept the given number for the 'actual casts' display, but not use it in the cast efficiency calculation. Note that the penance cast efficiency calculation appears to work mostly right even with the spurious cast events, although I think it's screwing up the cooldown calculation from SpellUsable. My hope is that this will be fixed by the planned EventEnhancer feature.
- getMaxCasts : currently used for spells with cooldowns that vary based on state, which appear to be PW:S (rapture), Void Bolt (void form), and Mind Blast (void form). My hope is that this field can be eliminated entirely by instead handling these cases in the getCooldown function.

kfinch:
Ok I thought more about the relative merits of '% on CD' vs 'casts/rawMaxCasts'. First of all, the issue is largely irrelevant for shorter CDs, where the two calculations will come out almost the same. The calculations are also the same if the cooldown ends the fight available. The specific case where they diverge a lot is when the cooldown is large relative to the fight duration and its cast near the end of the fight.

Lets consider an edge case, and think about the relative merits: Fight duration: 3:20, Cooldown: 2:00, Actual casts @ 1:00 and 3:10.
- **casts/rawMaxCasts method** : casts = 2, rawMaxCasts = (200 / 120) = 1.67, cast efficiency = 1.20, clamped to **100%**
- **% on CD method** : Spell on CD from 1:00 to 3:00 and 3:10 to 3:20, which is 65% of the time. Cast efficiency = **65%**

Argument for the casts/rawMaxCasts method: long cooldowns are often lined up with things and/or saved for specific times, the player should be given full credit as long as they cast the maximum number of times, regardless of when they cast.

Argument for % on CD method: The player doesn't know for certain when the fight is going to end, so leaving the spell available for an extended period of time was a mistake with the *potential* to cause a lost cast, even if in this particular circumstance it ended up not mattering.

I also want to more directly address the comment ```A 2 minute cooldown in a 2:20 fight could be cast twice but you generally don't want to as that would be inefficient as there would be nothing to heal.```
If a spell should be saved for specific circumstances rather than cast on cooldown, then *it should not be entered in cast efficiency with a threshold of 80%* Regardless of if we change the cast efficiency calculation, I'm of the belief that the big healing cooldowns should have their threshold lowered a lot for exactly this reason.

MartijnHols:
In no situation can you cast such a spell 3 times, so how could 65% be better than actual 100%?

kfinch:
As I said above, because the player didn't know at the time that they wouldn't get a 3rd cast, so holding the CD a full minute was a mistake.

Anyone reviewing a DPS player's logs would have called that a mistake upon seeing it.

kfinch:
Put another way, "an unpunished mistake is still a mistake"

MartijnHols:
We're doing log analysis, I believe we should look at what *actually happened*, not at what *could have happened*, as much as possible.

kfinch:
But this is log analysis, we'd be making a suggestion based on when the player actually cast their spells, pointing out that it should have been cast earlier

MartijnHols:
Casting earlier for no reason is never better if it doesn't lead to more total casts

Pewtro:
Log analysis is also about teaching how the player should treat future (similiar) encounters, as it is looking at what actually happened. 
While in this specific log, it may not have led to an extra cast, it should still be treated as a misplay to have held the CD for that long imo. Holding CDs for no apparent reason = not good - even despite not getting an extra cast in the end. 
Kill times cannot be estimated reliably in real-time, and therefore not using cooldowns actively and as often as possible (outside of holding for specific lineups or for execute) is a misplay and should notify the user.

kfinch:
I'll repeat a suggestion I made in dev chat, which is that we use the '% on CD' calculation, but make it suppress a potential suggestion if the player gets maximum possible casts, operating under the assumption that they planned their CDs for this.

Lets consider another hypothetical scenario, and this time I'm going to to tailor it to be friendly to the new calculation: Fight Duration: 4:30, Cooldown: 2:00. There's time for three casts, but you have to be quick about it, missing the third would generally be understandable. Alice casts twice, at 0:10 and 2:20. Robert casts twice, at 2:10 and 4:15.
- **casts/rawMaxCasts method** : calculates the same for both players, efficiency = 2/2.25 = **88.9%**, likely no suggestion unless threshold set very high
- **% on CD method** : Alice was pretty prompt about using her cooldown, ends up with a similar looking calculation to the current method and gets the same efficiency at **88.9%**. Robert was most certainly not prompt, in fact it's pretty clear how he missed his cast. He only kept the spell on CD for 135 out of 270 seconds, gets an efficiency of **50%** and a suggestion. The best part is, if he's confused as to how this is 50%, there's a fancy new spell timeline feature in the suggestion that will show him what he missed.

kfinch:
@MartijnHols alright, refactored and cleaned up, take another look when you can.

kfinch:
I've looked through to make sure, but literally the only interesting changes are in one class, the new cast efficiency handler: https://github.com/kfinch/WoWAnalyzer/blob/206d7b423b34f2c031308afafdb471e698450322/src/Parser/Core/Modules/CastEfficiency.js

All the other file changes are just handling the rename of the old CastEfficiency classes to 'Abilities', and the old CPM_ABILITIES to 'ABILITIES', along with a couple tweaks to ABILITIES to update deprecated behavior.

src/Parser/Core/Modules/Abilities.js
```diff
@@ -0,0 +1,149 @@
...
+    ROTATIONAL: 'Rotational Spell',
+    ROTATIONAL_AOE: 'Spell (AOE)',
+    ITEMS: 'Item',
+    COOLDOWNS: 'Cooldown',
+    DEFENSIVE: 'Defensive Cooldown',
+    OTHERS: 'Spell',
+    UTILITY: 'Utility',
+    HEALER_DAMAGING_SPELL: 'Damaging Spell',
+  };
+  static ABILITIES = [
```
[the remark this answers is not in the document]
kfinch: One of my planned enhancements after this was to make it non static, and then change the ability function signatures so they don't need to be passed everything.

Are you suggesting do all that now, or just the 'non-static' part and save the rest for later?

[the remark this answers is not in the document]
MartijnHols: No that's great :D

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -4,182 +4,210 @@ import SpellLink from 'common/SpellLink';
...
-    });
+
+  /*
+   * Gets the number of ms the given spell has been on CD since the beginning of the fight.
+   * Only works on spells entered into CastEfficiency list.
+   */
+  _getTimeOnCooldown(spellId) {
+    const history = this.spellHistory.historyBySpellId[spellId];
+    if(!history) {
+      return;
```
MartijnHols: Please return `null` when alternating between a return value and nothing.

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -4,182 +4,210 @@ import SpellLink from 'common/SpellLink';
...
+   */
+  _getAverageCooldown(spellId) {
+    const history = this.spellHistory.historyBySpellId[spellId];
+    if(!history) {
+      return;
+    }
+
+    let lastStartChargeTimestamp = null;
+    let recharges = 0;
+    const totalRechargingTime = history
```
MartijnHols: What's the difference between this and `_getTimeOnCooldown`?

In any case it should probably be its own method.

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -4,182 +4,210 @@ import SpellLink from 'common/SpellLink';
...
-    return ability ? (ability.getCooldown(this.haste.current, this.combatants.selected) * 1000) : undefined;
+
+  /*
+   * Gets the average number of ms the given spell took to cooldown (or generate a charge)
+   * Only works on spells entered into CastEfficiency list.
+   */
+  _getAverageCooldown(spellId) {
+    const history = this.spellHistory.historyBySpellId[spellId];
+    if(!history) {
+      return;
```
MartijnHols: return null

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -4,182 +4,210 @@ import SpellLink from 'common/SpellLink';
...
         return;
       }
-      when(cpm.castEfficiency).isLessThan(cpm.recommendedCastEfficiency)
+      const ability = abilityInfo.ability;
+      when(abilityInfo.castEfficiency).isLessThan(abilityInfo.recommendedCastEfficiency)
         .addSuggestion((suggest, actual, recommended) => {
-          return suggest(<span>Try to cast <SpellLink id={cpm.ability.spell.id} /> more often. {cpm.ability.extraSuggestion || ''} <a href="#spell-timeline">View timeline</a>.</span>)
-            .icon(cpm.ability.spell.icon)
-            .actual(`${cpm.casts} out of ${cpm.maxCasts} possible casts; ${formatPercentage(actual)}% cast efficiency`)
+          return suggest(<span>Try to cast <SpellLink id={ability.spell.id} /> more often. {ability.extraSuggestion || ''} <a href="#spell-timeline">View timeline</a>.</span>)
```
MartijnHols: Using the Wrapper component in `common/Wrapper` is recommended instead of spans to reduce the amount of unnecessary elements being rendered.

src/Parser/Core/Modules/CastEfficiency.js
kfinch: Will fix

src/Parser/Core/Modules/CastEfficiency.js
kfinch: Will fix

src/Parser/Core/Modules/CastEfficiency.js
kfinch: TimeOnCooldown gets total time on cooldown, up to and including spell still cooling down at the current time.

AverageCooldown gets the average time it takes spell to recharge, disregards a cooldown that's in progress at the current time.

I considered trying to merge more of the functionality, but it seemed like I was twisting things too much so I made them seperate methods.

src/Parser/Core/Modules/CastEfficiency.js
kfinch: I assume that means <Wrapper> instead of <span>. I don't know enough about html to understand why it makes a difference, but I'll do it.

src/Parser/Core/Modules/CastEfficiency.js
MartijnHols: I meant the statement this comment was one; it gets the totalRechargingTime which should be equal to _getTimeOnCooldown so could it be removed?

src/Parser/Core/Modules/CastEfficiency.js
MartijnHols: Wrapper doesn't render anything, span renders span

kfinch: I don't know what that means 😆

MartijnHols: it's fake https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/common/Wrapper.js

src/Parser/Core/Modules/CastEfficiency.js
kfinch: Oh, this totalRechargingTime includes hooks to bump recharge count, and also disregards the time of the current cooldown. There is probably room to merge some code.

src/Parser/Core/Modules/CastEfficiency.js
kfinch: I understand the Wrapper is like a "fake object", I just don't understand what it means to 'render span', and why it's bad that we're rendering span.

MartijnHols: The more elements we render the slower things get, see Events tab. A bunch of spans won't matter but doing this properly everywhere will improve render performance.

MartijnHols: Rendering is hundreds of times slower than just running JS.

kfinch:
Ok I did the fixes, including merging of the get cooldown logic. I do end up calling basically the same stuff twice, I could probably fix that too.

NoSuggest / NoCanBeImproved is added when player achieves max casts, btw, although how I implemented that is kinda sloppy. There's a check for it when setting canBeImproved and a seperate check when building suggestions... probably could implement better.

kfinch:
annnd just pushed a commit that improved both those implementations.

MartijnHols:
Nice 👍 

I'm still not sold on the percentage of time on CD method. 

OLD:
Cast Efficiency = casts / raw max possible casts
Pros:
- Predictable: reflects what the user expects; he sees x casts / y max and cast efficiency is connected to that. It only slightly varies based on fight duration.
Cons:
- Allows for slack which might result in reduced performance.

NEW:
Cast Efficiency = timeOnCooldown / fightDuration
Pros:
- Might motivate players to cast an ability earlier and better on cooldown, reducing the risk they're missing extra casts if the fight turns out taking longer than expected.
- Better matches the spell timeline (although this is less important as matching the shown cast amounts).
Cons:
- Less predictable: it is connected to time on cooldown which might end up much more out of balance than cast ratio. E.g. Lay on Hands with a 8 minute cooldown cast 5 minutes into a 6 minute fight would show 16.667% even though the player actually cast 100%. Making an exception for this by setting CE to 100% when the player had all casts would still fail for a 8:30 minute fight where it is unreasonable to require the player to cast it inefficiently.
- Punishes players for holding cooldowns to align with tactics.

Another alternative would be to use the old method but have it based on the ceiled casts, this would completely match the expected cast efficiency from casts / max casts while being more strict about using the max amount of casts.

I also thought about making a switch between those approaches, but I think this might be too confusing.

kfinch:
It's pretty easy for me to switch back to using casts/rawMaxCasts if needed, but allow me to make one more argument in favor of 'time on cooldown'.

I initially didn't set out to change the way cast efficiency was calculated, this came about because as I was implementing cast efficiency for spells whose cooldowns can be dynamically reduced or reset, it occurred to me that 'max casts' was no longer a 100% accurate calculation. Some cooldown reducers are random procs, others depend on how frequently you cast a different spell. I decided I needed a different way to calculate cast efficiency, and '% time on cooldown' seemed to be the natural fit. Afterall, DPS guides talking about spells with cooldowns often advice "cast whenever available".

The other reason I wanted to switch to '% time on cooldown' is that I think it's actually an easier metric for players to understand than the 'casts/rawMaxCasts' number. You'll see I just pushed a tweak to the suggestion text, I think what I've changed it to makes things much more clear than "X% cast efficiency".

I was also intending to update the display of the cast efficiency chart a bit, I wanted to make the 'actual casts / max casts' a seperate column from the '% cast efficiency' bar, and then add a new header for that bar that would say "Percentage on Cooldown". Unfortunately, something about the table setup is messing with me and I can't get it to render right. I'll keep trying.

kfinch:
Ok I figured out the chart, you should check out what I've done with it. Maybe still some weirdness with text justification, but I think you'll get the idea.

kfinch:
cast-effic

MartijnHols:
Ahh that makes a lot of sense thanks. This is good to merge as soon as you rename "Percent on Cooldown" to "Time on cooldown" :D

kfinch:
Ahh yeah that makes sense. It's done.

Issue #762, opened 2017-11-08
[Restoration Shaman] Gift of the Queen Hit Counter Incorrect


rp4rk:
Gift of the Queen is not accounting for both the spell IDs from the intial hit and the echo from the artifact. This should be an easy fix.

See https://github.com/WoWAnalyzer/WoWAnalyzer/blob/d62914e25fad68d25ea2884f4fd49f139857271c/src/Parser/Shaman/Restoration/CombatLogParser.js#L156

PR #803, opened 2017-11-14
Implementing Gift of the Queen Replica for Restoration Shaman

HawkCorrigan:


HawkCorrigan:
Fix regarding issue #762.
Also includes Cloudburst Tracking for the Second Hit.

src/Parser/Shaman/Restoration/CombatLogParser.js
```diff
@@ -145,24 +146,29 @@ class CombatLogParser extends CoreCombatLogParser {
...
     const chainHealAvgHits = chainHealHits / chainHealCasts;
     const maxChainHealTargets = this.modules.combatants.selected.hasTalent(SPELLS.HIGH_TIDE_TALENT.id) ? 5 : 4;
     const chainHealTargetEfficiency = chainHealAvgHits / maxChainHealTargets;
 
-    const hasDeepWaters = this.modules.combatants.selected.traitsBySpellId[SPELLS.DEEP_WATERS.id] > 0;
     const giftOfTheQueenHits = giftOfTheQueen.healingHits || 0;
-    const giftOfTheQueenAvgHits = giftOfTheQueenHits / giftOfTheQueenCasts / (hasDeepWaters ? 2 : 1);
+    const giftOfTheQueenDuplicateHits = giftOfTheQueenDuplicate.healingHits || 0;
+    const giftOfTheQueenAvgHits = (giftOfTheQueenHits+giftOfTheQueenDuplicateHits) / (giftOfTheQueenCasts+giftOfTheQueenDuplicateCasts);
     const giftOfTheQueenTargetEfficiency = giftOfTheQueenAvgHits / 6;
```
anom0ly: Question: Does the target efficiency change with the additional of the Deep Waters trait?

I would believe this would be modified to be either 6 or 12 targets, depending of if you have the trait or not?

Also, if that is the case, does the lower suggestion change?

src/Parser/Shaman/Restoration/CombatLogParser.js
```diff
@@ -145,24 +146,29 @@ class CombatLogParser extends CoreCombatLogParser {
     const unbuffedHealingSurges = healingSurges - twHealingSurges;
     const totalTwUsed = twHealingWaves + twHealingSurges;
     const giftOfTheQueenCasts = giftOfTheQueen.casts || 0;
+    const giftOfTheQueenDuplicateCasts = giftOfTheQueen.casts || 0;
```
anom0ly: What is the point of setting the cast event to the `Duplicate` casts also? This seems redundant and could be handled by knowing if the player has the trait or not.

HawkCorrigan: yup that's a typo i didn't catch, should be giftDuplicate.casts

src/Parser/Shaman/Restoration/CombatLogParser.js
HawkCorrigan: It doesn't need to adjust to 12 target if you divide by the sum of Normal and Duplicate casts.
You could also just take the amount of normal casts and divide by 12, but you would have a weird edge case if the fight ends/you die in the 3 second window before the second cast goes through

src/Parser/Shaman/Restoration/CombatLogParser.js
anom0ly: So just so you are aware - when I did a quick check on the logs, the `Duplicate` event didn't seem to have an associated Cast event. I would just double check to make sure that a cast event is actually generated so this number isn't always zero and therefore influences the below calculation.

src/Parser/Shaman/Restoration/CombatLogParser.js
anom0ly: Makes sense. Understood.

PR #804, opened 2017-11-14
[Discipline priest] Fix for double disc priests null

karlpralow:


kfinch:
Looks like this broke a test in `src/tests/Parser/Priest/Discipline/Normalizers/AtonementSuccessiveDamage.test.js`. It's a peculiarity of the test and not a problem with your fix, but you still need to check.

karlpralow:
@kfinch Yes both concern the same issue, but #801 fixes it with null checking, maybe not a bad thing to have but this fixes the source of the issue

PR #807, opened 2017-11-14
[Discipline Priest] Added a column for absolute healing in atonement source tab

karlpralow:


MartijnHols:
screenshot please

karlpralow:
image

MartijnHols:
I don't think "relative" and "absolute" are clear, how about making it a toggle instead?

karlpralow:
When comparing differrent things, it's better to have it in absolute so you can compare it to non atonement related things.

Quick example, one of our relic traits is a 5% increase to shield. So if I wanted to compare this trait to Shadowbind NLC trait, I need to actually compare the absolute value of atonement provided by Shadowbind, not the relative.

karlpralow:
I made it into a toggle. Screenshot 
image

MartijnHols:
nit pick, but could you move it down 1 line so it's on the same level as name and healing done? no reason for it not to be :D

karlpralow:
Yep! Done 
image

PR #809, opened 2017-11-14
[Unholy Death Knight] Added t20 2 piece

jjs451:
Finished an old PR on behalf of Maxweii

### DIFF
--- a/src/common/SPELLS/DEATH_KNIGHT.js
+++ b/src/common/SPELLS/DEATH_KNIGHT.js
@@ -410,12 +410,18 @@ export default {
   },
   
   // Unholy Tier Sets
-  // T20 2 Piece
-  UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS_BUFF: {
-    id: 246995,
-    name: 'T20 2 Set Bonus',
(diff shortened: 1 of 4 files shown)


src/common/SPELLS/DEATH_KNIGHT.js
```diff
@@ -410,12 +410,18 @@ export default {
   },
   
   // Unholy Tier Sets
-  // T20 2 Piece
-  UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS_BUFF: {
-    id: 246995,
-    name: 'T20 2 Set Bonus',
+  // T20 2 Piece and buff
+  UNHOLY_DEATH_KNIGHT_T20_2SET_BONUS: {
```
Pewtro: ``and buff``` ? Surely it's only the t202p

jjs451: The intent is that its describing the next 2 entries in the spell file.  When I add T21 there'll be a new comment after master of ghouls indicating that the T21 spells are beginning.  The way the file is laid out, each comment defines everything under it until theres a new comment

PR #810, opened 2017-11-15
Shadow priest, 4 legendaries

hasseboulen:
Added support for 4 shadow spec legendaries.

About the twins painful touch:
The extra DOTs applied might not have the same timestamp (and the caster might not have the spread buff when applied due to this) so I added an latency buffer.

Anunds seared shackles:
Due to losing the buff to void bolts on cast rather than on damage event, the buff have to be manually tracked.


src/Parser/Priest/Shadow/Modules/Items/ZenkaramIridisAnadem.js
```diff
@@ -0,0 +1,58 @@
...
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasHead(ITEMS.ZENKARAM_IRIDIS_ANADEM.id);
+  }
+
+  on_byPlayer_damage(event){
+    if(
+      this.combatants.selected.hasBuff(SPELLS.VOIDFORM_BUFF.id) &&
+      event.ability.type === SCHOOLS.ids.SHADOW){
+      this.bonusDamage += event.amount -
```
MartijnHols: The recommended approach for calculating these kind of damage bonuses is to use `calculateEffectiveDamage`. For example this is missing absorbed damage which more often than not is important in fights.

src/Parser/Priest/Shadow/Modules/Items/TwinsPainfulTouch.js
```diff
@@ -0,0 +1,63 @@
...
+  on_byPlayer_applydebuff(event){
+    const spellID = event.ability.guid;
+    if(spellID === SPELLS.MIND_FLAY.id && this.combatants.selected.hasBuff(SPELLS.THE_TWINS_PAINFUL_TOUCH.id)){
+      this.mindflayTimestamps.push(event.timestamp);
+      return;
+    }
+
+    if(spellID === SPELLS.VAMPIRIC_TOUCH.id || spellID === SPELLS.SHADOW_WORD_PAIN.id){
+      // needs a latency buffer (and due to this, we cant check if combatant has the buff either)
+      if(this.mindflayTimestamps.find(mindflayTimestamp => event.timestamp >= mindflayTimestamp && event.timestamp < (mindflayTimestamp + LATENCY_BUFFER))){
```
MartijnHols: `.some` instead of `.find` would suffice

src/Parser/Priest/Shadow/Modules/Items/TwinsPainfulTouch.js
```diff
@@ -0,0 +1,63 @@
...
+  mindflayTimestamps = [];
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasFinger(ITEMS.THE_TWINS_PAINFUL_TOUCH.id);
+  }
+
+  on_byPlayer_applydebuff(event){
+    const spellID = event.ability.guid;
+    if(spellID === SPELLS.MIND_FLAY.id && this.combatants.selected.hasBuff(SPELLS.THE_TWINS_PAINFUL_TOUCH.id)){
+      this.mindflayTimestamps.push(event.timestamp);
```
MartijnHols: Do you really need to track an array of this? How about just tracking the last timestamp?

src/Parser/Priest/Shadow/Modules/Items/AnundsSearedShackles.js
```diff
@@ -0,0 +1,55 @@
...
+  buffStacksSinceLastCast = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasWrists(ITEMS.ANUNDS_SEARED_SHACKLES.id);
+  }
+
+  on_byPlayer_damage(event){
+    const spellID = event.ability.guid;
+    if(spellID === SPELLS.VOID_BOLT.id){
+      this.bonusDmg += event.amount - (event.amount / (1 + this.buffStacksSinceLastCast * DAMAGE_INCREASE_PER_STACK));
```
MartijnHols: The recommended approach for calculating these kind of damage bonuses is to use `calculateEffectiveDamage`. For example this is missing absorbed damage which more often than not is important in fights.

src/Parser/Priest/Shadow/Modules/Items/TwinsPainfulTouch.js
hasseboulen: Good call, it's a leftover from earlier. Will get it fixed.

src/Parser/Priest/Shadow/Modules/Items/ZenkaramIridisAnadem.js
hasseboulen: Ah, I will get it changed.

src/Parser/Priest/Shadow/Modules/Items/AnundsSearedShackles.js
hasseboulen: Ah, I will get it changed.

src/Parser/Priest/Shadow/Modules/Items/TwinsPainfulTouch.js
hasseboulen: Is this for performance, readability or something else? They should both work if I'm not mistaken.

PR #815, opened 2017-11-15
[Discipline Priest] Fix to support SWP in the PtW Module

karlpralow:


src/Parser/Priest/Discipline/Modules/Features/PurgeTheWicked.js
```diff
@@ -16,17 +16,27 @@ class PurgeTheWicked extends Analyzer {
     enemies: Enemies,
   };
 
+  dotSpell;
+
+  on_initialized() {
+    if(this.owner.modules.combatants.selected.hasTalent(SPELLS.PURGE_THE_WICKED_TALENT.id)) {
```
kfinch: Getting the Combatants object from `this.owner.modules.combatants` is technically deprecated, in the future you should include combatants as a dependency and then access with `this.combatants`.

PR #816, opened 2017-11-15
[Disc] Soul of the High Priest as Talent Box

awlego:
Moved Soul of the High Priest info to have the Twist of Fate talent box.
Also added suggestions on Twist of Fate healing percentages.


src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
     this.healing += calculateEffectiveHealing(event, TWIST_OF_FATE_HEALING_INCREASE);
   }
 
+  suggestions(when) {
+    if (this.hasSoulOfTheHighPriest === false) {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.minor)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different talent than <SpellLink id={SPELLS.TWIST_OF_FATE_TALENT.id}/>.</span>)
```
anom0ly: Can you be more descriptive as to why you would pick a different talent?

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
...
+            .icon(SPELLS.TWIST_OF_FATE_TALENT.icon)
+            .actual(`${formatPercentage(actual)}% of total healing`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended.`)
+            .regular(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.regular)
+            .major(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.major);
+        });
+    } else {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(SuggestionThresholds.TWIST_OF_FATE_LEGO_PERCENT.minor)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different legendary other than <ItemLink id={ITEMS.SOUL_OF_THE_HIGH_PRIEST.id}/>.</span>)
```
anom0ly: Can you be more descriptive as to why you would pick a different legendary?

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
     this.healing += calculateEffectiveHealing(event, TWIST_OF_FATE_HEALING_INCREASE);
   }
 
+  suggestions(when) {
+    if (this.hasSoulOfTheHighPriest === false) {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.minor)
```
MartijnHols: This is very long and I don't think moving the suggestion threshold out of this class makes this much easier to maintain, but it does make it harder to find the actual thresholds. In general you fiddle with suggestion thresholds one at a time and having the entire context of the suggestion allows you to be more precise. 

I recommend going for a hardcoded value for the initial threshold and then do regular `recommended - 0.005` and major similar to that.

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
     this.healing += calculateEffectiveHealing(event, TWIST_OF_FATE_HEALING_INCREASE);
   }
 
+  suggestions(when) {
+    if (this.hasSoulOfTheHighPriest === false) {
```
MartijnHols: This can be simplified: `!this.hasSoulOfTheHighPriest`

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
...
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.minor)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different talent than <SpellLink id={SPELLS.TWIST_OF_FATE_TALENT.id}/>. Castigation will give a consistent 3-5% increase and Schism provides a significant DPS increase if more healing is not needed.</span>)
```
MartijnHols: Could you try to make it clear in the suggestion *why* this is recommended?

MartijnHols: I just realized this and it's nowhere else so there's no way you could have known, but for future suggestion instead of wrapping it in `<span>`  tags I recommend using the `Wrapper` component in `common/Wrapper`. This reduces the amount of needless elements which makes the source cleaner and more speedy.

MartijnHols: This does seem like more of a l2p thing for guides than something that belong to a gameplay improving suggestion, how will you ensure it's only shown when it is accurate?

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
...
+            .recommended(`>${formatPercentage(recommended)}% is recommended.`)
+            .regular(SuggestionThresholds.TWIST_OF_FATE_LEGO_PERCENT.regular)
+            .major(SuggestionThresholds.TWIST_OF_FATE_LEGO_PERCENT.major);
+        });
+    }
+  }
+
   statistic() {
-    if(!this.hasTwistOfFate) { return; }
+    if(!this.active) { return; }
```
MartijnHols: Nit pick: you should generally return `null` in places where the default is a React node. `undefined` is normally not an accepted value (although I manually handle the `statistic` method to account for it, it's still best practice and might break in the future).

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +63,32 @@ class TwistOfFate extends Analyzer {
...
+            .icon(SPELLS.TWIST_OF_FATE_TALENT.icon)
+            .actual(`${formatPercentage(actual)}% of total healing`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended.`)
+            .regular(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.regular)
+            .major(SuggestionThresholds.TWIST_OF_FATE_TALENT_PERCENT.major);
+        });
+    } else {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(SuggestionThresholds.TWIST_OF_FATE_LEGO_PERCENT.minor)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different legendary other than <ItemLink id={ITEMS.SOUL_OF_THE_HIGH_PRIEST.id}/>. If the burst healing during dangerous situations from Twist of Fate is not needed, you will find greater average healing from several other legendaries.</span>)
```
MartijnHols: see above

src/Parser/Priest/Discipline/Modules/Items/SoulOfTheHighPriest.js
```diff
@@ -0,0 +1,40 @@
...
+    return {
+      item: ITEMS.SOUL_OF_THE_HIGH_PRIEST,
+      result: <span>This gave you <SpellLink id={SPELLS.TWIST_OF_FATE_TALENT.id} />.</span>,
+    };
+  }
+
+  suggestions(when) {
+    when(this.hasPickedOtherTalent).isFalse()
+    .addSuggestion((suggest) => {
+      return suggest(<span>When using <ItemLink id={ITEMS.SOUL_OF_THE_HIGH_PRIEST.id} /> please make sure to pick another talent in the same talent row. Your choices are <SpellLink id={SPELLS.CASTIGATION_TALENT.id} /> or <SpellLink id={SPELLS.SCHISM_TALENT.id} />.</span>)
```
MartijnHols: Please try to include what was found so the user understands why this is being suggested. e.g. "You did not pick a second talent in the first row."

awlego:
Sorry for the slow update -- I got caught up with life. Let me know if you want any more changes.

src/Parser/Priest/Discipline/Modules/Items/SoulOfTheHighPriest.js
```diff
@@ -0,0 +1,40 @@
...
+  on_initialized() {
+    this.active = this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_HIGH_PRIEST.id);
+    this.hasPickedOtherTalent = this.combatants.selected.hasTalent(SPELLS.CASTIGATION_TALENT.id) || 
+                                this.combatants.selected.hasTalent(SPELLS.SCHISM_TALENT.id);
+  }
+
+  item() {
+    return {
+      item: ITEMS.SOUL_OF_THE_HIGH_PRIEST,
+      result: <span>This gave you <SpellLink id={SPELLS.TWIST_OF_FATE_TALENT.id} />.</span>,
```
MartijnHols: Please use the Wrapper class from `common/Wrapper` instead of spans to wrap elements without changing how they're displayed.

src/Parser/Priest/Discipline/Modules/Items/SoulOfTheHighPriest.js
MartijnHols: Please use the Wrapper class from `common/Wrapper` instead of spans to wrap elements without changing how they're displayed.

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +62,32 @@ class TwistOfFate extends Analyzer {
     this.healing += calculateEffectiveHealing(event, TWIST_OF_FATE_HEALING_INCREASE);
   }
 
+  suggestions(when) {
+    if (!this.hasSoulOfTheHighPriest) {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(0.05)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different talent than <SpellLink id={SPELLS.TWIST_OF_FATE_TALENT.id}/>. Castigation will give a consistent 3-5% increase and Schism provides a significant DPS increase if more healing is not needed.</span>)
```
MartijnHols: Please use the Wrapper class from `common/Wrapper` instead of spans to wrap elements without changing how they're displayed.

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +62,32 @@ class TwistOfFate extends Analyzer {
...
+            .icon(SPELLS.TWIST_OF_FATE_TALENT.icon)
+            .actual(`${formatPercentage(actual)}% of total healing`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended.`)
+            .regular(0.045)
+            .major(0.025);
+        });
+    } else {
+      when(this.owner.getPercentageOfTotalHealingDone(this.healing)).isLessThan(0.03)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>Consider picking a different legendary other than <ItemLink id={ITEMS.SOUL_OF_THE_HIGH_PRIEST.id}/>. If the burst healing during dangerous situations from Twist of Fate is not needed, you will find greater average healing from several other legendaries.</span>)
```
MartijnHols: Please use the Wrapper class from `common/Wrapper` instead of spans to wrap elements without changing how they're displayed.

MartijnHols: Twist of Fate could be a spell link.

src/Parser/Priest/Discipline/Modules/Spells/TwistOfFate.js
```diff
@@ -60,8 +62,32 @@ class TwistOfFate extends Analyzer {
...
+            .recommended(`>${formatPercentage(recommended)}% is recommended.`)
+            .regular(0.025)
+            .major(0.02);
+        });
+    }
+  }
+
   statistic() {
-    if(!this.hasTwistOfFate) { return; }
+    if(!this.active) { return; }
```
MartijnHols: This can be omitted as this method is never called for inactive modules.

PR #820, opened 2017-11-17
[Resto Shaman] Refactored Spec Modules

anom0ly:
Refactored Restoration Shaman spec to bring it in line with the current module / spec code.


src/Parser/Shaman/Restoration/Modules/Talents/HighTide.js
```diff
@@ -39,6 +45,24 @@ class HighTide extends Analyzer {
...
   }
+
+  statistic() {
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.HIGH_TIDE_TALENT.id} />}
+        value={`${formatPercentage(this.healing)} %`}
+        label={(
+          <dfn data-tip={'The percentage of your healing that is caused by High Tide.'}>
```
MartijnHols: Tooltips like these would be best omitted to reduce the amount of tooltips so tooltips are primarily used for things that reveal interesting/important info so that users don't learn they are uninteresting.

PR #829, opened 2017-11-18
Implemented Redux

MartijnHols:
- Moved a lot of code out of App to make it more maintainable
- Removed `config.footer` which has been deprecated for a long time
- Change URL routing to be completely Redux based so that we only have one state store instead of Redux + URL (also we can't use it properly as the URL being to a selected player still requires us to fetch report and combatants so 1 page can't be completely isolated)
- Refactored a lot of code
- Moved the document title updater out of App to a new component called `DocumentTitleUpdater` so that it's more isolated and doesn't pollute App (also causes it to update more reliably)
- Greatly improved network error handling; it should now always show a decent error page instead of an alert, and properly account for all APIs potentially reporting the report is private (this can for example happen if fights overview and combatants were cached and afterwards the report was made private)

Things I'll solve later/in a future PR:

- Replace usages of `react-router-dom`'s Link with a Redux push
- Further move code form App to subcomponents (e.g. report, combatants and events fetching)
- Also handle parse errors more controlled


kfinch:
I appreciate seeing this in a PR, and the splitting out of a lot of stuff from App into their own files looks nice. I'm not sure if anyone else on this project is familiar enough with App to meaningfully review / catch errors... so I dunno what to do about a review.

One small thing I did notice is that you commented footers in the spec CONFIGs rather than deleting outright... is that what you intended, for them to be left as a reminder to put the info elsewhere?

MartijnHols:
Aye

MartijnHols:
well yolo then

PR #831, opened 2017-11-18
Sub rogue refactoring

TsaboTavok:
Refactoring of the current sub rogue metrics.

### DIFF
--- a/src/Parser/Rogue/Subtlety/CombatLogParser.js
+++ b/src/Parser/Rogue/Subtlety/CombatLogParser.js
@@ -2,7 +2,11 @@ import CoreCombatLogParser from 'Parser/Core/CombatLogParser';
 import DamageDone from 'Parser/Core/Modules/DamageDone';
 import CastEfficiency from './Modules/Features/CastEfficiency';
 
-import { ShadowBladesUptime, ShadowDance, DarkShadowContribution, DarkShadowEvis, SymbolsOfDeathUptime } from './Modules';
+import ShadowBladesUptime from "./Modules/Features/ShadowBladesUptime";
+import ShadowDance from "./Modules/Features/ShadowDance";
+import SymbolsOfDeathUptime from "./Modules/Features/SymbolsOfDeathUptime";
+import DarkShadowContribution from "./Modules/Talents/DarkShadowContribution";
+import DarkShadowEvis from "./Modules/Talents/DarkShadowEvis";
 
 class CombatLogParser extends CoreCombatLogParser {
   static specModules = {
--- a/src/Parser/Rogue/Subtlety/CombatLogParser.js
+++ b/src/Parser/Rogue/Subtlety/CombatLogParser.js
@@ -1,110 +1,22 @@
-import React from 'react';
-
-import Icon from 'common/Icon';
 import CoreCombatLogParser from 'Parser/Core/CombatLogParser';
-import SPELLS from 'common/SPELLS';
-import SpellIcon from 'common/SpellIcon';
-import StatisticBox from 'Main/StatisticBox';
-
-import ShadowDance from './Modules/Features/ShadowDance';
+import DamageDone from 'Parser/Core/Modules/DamageDone';
 import CastEfficiency from './Modules/Features/CastEfficiency';
 
-function formatThousands(number) {
-  return (`${Math.round(number || 0)}`).replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,');
-}
-
-function formatNumber(number) {
-  if (number > 1000000) {
-    return `${(number / 1000000).toFixed(2)}m`;
-  }
-  if (number > 10000) {
-    return `${Math.round(number / 1000)}k`;
-  }
-  return formatThousands(number);
-}
-
-function formatPercentage(percentage) {
-  return (Math.round((percentage || 0) * 10000) / 100).toFixed(2);
-}
+import { ShadowBladesUptime, ShadowDance, DarkShadowContribution, DarkShadowEvis, SymbolsOfDeathUptime } from './Modules';
 
 class CombatLogParser extends CoreCombatLogParser {
   static specModules = {
+    //Feature
+    damageDone: [DamageDone, { showStatistic: true }],
+    shadowBladesUptime : ShadowBladesUptime,
     shadowDance: ShadowDance,
     castEfficiency: CastEfficiency,
+    symbolsOfDeathUptime:  SymbolsOfDeathUptime,
+
+    //Talents
+    darkShadowContribution: DarkShadowContribution,
+    darkShadowEvis: DarkShadowEvis,
   };
-
-  generateResults() {
-    const results = super.generateResults();
-
-    const fightDuration = this.fightDuration;
-
-    results.statistics = [
-      <StatisticBox
-        icon={<Icon icon="class_rogue" alt="Damage Per Second" />}
-        value={`${formatNumber(this.modules.damageDone.total.effective / fightDuration * 1000)} DPS`}
-        label={(
-          <dfn data-tip={`The total damage done recorded was ${formatThousands(this.modules.damageDone.total.effective)}.`}>
-            Damage done
-          </dfn>
-        )}
-      />,
-      <StatisticBox
-        icon={<SpellIcon id={SPELLS.SHADOW_BLADES.id} />}
-        value={`${formatPercentage((this.modules.combatants.selected.getBuffUptime(SPELLS.SHADOW_BLADES.id) / this.fightDuration))} %`}
-        label={(
-          <dfn data-tip={'Shadow Blades up time'}>
-            Shadow Blades up time
-          </dfn>
-        )}
-      />,
-
-      <StatisticBox
-        icon={<SpellIcon id={SPELLS.SHADOW_DANCE.id} />}
-        value={`${formatPercentage((this.modules.combatants.selected.getBuffUptime(SPELLS.SHADOW_DANCE_BUFF.id) / this.fightDuration))} %`}
-        label={(
-          <dfn data-tip={'Shadow Dance up time'}>
-            Shadow Dance up time
-          </dfn>
-        )}
-      />,
-
-      <StatisticBox
-        icon={<SpellIcon id={SPELLS.SYMBOLS_OF_DEATH.id} />}
-        value={`${formatPercentage((this.modules.combatants.selected.getBuffUptime(SPELLS.SYMBOLS_OF_DEATH.id) / this.fightDuration))} %`}
-        label={(
-          <dfn data-tip={'Symbols of Death up time'}>
-            Symbols of Death up time
-          </dfn>
-        )}
-      />,
-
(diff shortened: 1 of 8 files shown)


src/Parser/Rogue/Subtlety/CombatLogParser.js
```diff
@@ -1,110 +1,22 @@
...
-  if (number > 10000) {
-    return `${Math.round(number / 1000)}k`;
-  }
-  return formatThousands(number);
-}
-
-function formatPercentage(percentage) {
-  return (Math.round((percentage || 0) * 10000) / 100).toFixed(2);
-}
+import { ShadowBladesUptime, ShadowDance, DarkShadowContribution, DarkShadowEvis, SymbolsOfDeathUptime } from './Modules';
```
kfinch: Rather than doing this import using an index.js in Modules, current style is to just import one per line, and skip the index.

I'm guessing the point of doing it this way is so you don't have to change a lot of imports if you move one of the modules? It's probably not a big deal you're doing it this way, but should talk to @MartijnHols about it and see what he thinks.

src/Parser/Rogue/Subtlety/Modules/Features/ShadowBladesUptime.js
```diff
@@ -0,0 +1,37 @@
...
+
+  statistic() {
+    const shadowDanceUptime = this.combatants.selected.getBuffUptime(SPELLS.SHADOW_DANCE_BUFF.id) / this.owner.fightDuration;
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SHADOW_DANCE_BUFF.id} />}
+        value={`${formatPercentage(shadowDanceUptime)} %`}
+        label={(
+          <dfn data-tip={'Shadow Dance up time'}>
+            Shadow Dance up time
```
kfinch: Only include tooltips if they add new information (which this one does not). Also, 'uptime' is one word.

src/Parser/Rogue/Subtlety/Modules/Features/ShadowBladesUptime.js
```diff
@@ -0,0 +1,37 @@
...
+
+
+class ShadowBladesUptime extends Analyzer {
+  static dependencies = {
+    enemies: Enemies,
+    combatants: Combatants,
+  };
+
+  statistic() {
+    const shadowDanceUptime = this.combatants.selected.getBuffUptime(SPELLS.SHADOW_DANCE_BUFF.id) / this.owner.fightDuration;
```
kfinch: It looks like you mixed up class name and statistic? Class is called ShadowBladesUptime, this is getting Shadow Dance uptime.

src/Parser/Rogue/Subtlety/Modules/Features/ShadowDance.js
```diff
@@ -41,6 +50,23 @@ class ShadowDance extends Analyzer {
       this.inShadowDance = false;
     }
   }
+
+  statistic() {
+    const shadowBladesUptime = this.combatants.selected.getBuffUptime(SPELLS.SHADOW_BLADES.id) / this.owner.fightDuration;
```
kfinch: It looks like you mixed up class name and statistic? Class is called ShadowDance, this is getting Shadow Blades uptime.

src/Parser/Rogue/Subtlety/Modules/Features/ShadowDance.js
```diff
@@ -41,6 +50,23 @@ class ShadowDance extends Analyzer {
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SHADOW_BLADES.id} />}
+        value={`${formatPercentage(shadowBladesUptime)} %`}
+        label={(
+          <dfn data-tip={'Shadow Blades up time'}>
+            Shadow Blades up time
```
kfinch: Only include tooltips if they add new information (which this one does not). Also, 'uptime' is one word.

src/Parser/Rogue/Subtlety/Modules/Features/SymbolsOfDeathUptime.js
```diff
@@ -0,0 +1,37 @@
...
+
+  statistic() {
+    const symbolsOfDeathUptime = this.combatants.selected.getBuffUptime(SPELLS.SYMBOLS_OF_DEATH.id) / this.owner.fightDuration;
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SYMBOLS_OF_DEATH.id} />}
+        value={`${formatPercentage(symbolsOfDeathUptime)} %`}
+        label={(
+          <dfn data-tip={'Symbols of Death uptime'}>
+          Symbols of Death uptime
```
kfinch: Only include tooltips if they add new information (which this one does not). Also, 'uptime' is one word.

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadow.js
```diff
@@ -0,0 +1,23 @@
...
+    shadowDance: ShadowDance,
+  };
+
+  bonusDmg = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.DARK_SHADOW_TALENT.id);
+  }
+
+}
```
kfinch: I'm confused, this module doesn't appear to do anything?

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadowContribution.js
```diff
@@ -0,0 +1,42 @@
...
+
+  statistic() {    
+    const danceDamage = this.shadowDance.totalDamageDoneInShadowDance * 0.3 / 1.3;
+    return (
+        <StatisticBox
+          icon={<SpellIcon id={SPELLS.DARK_SHADOW_TALENT.id} />}
+          value={`${formatNumber(danceDamage / this.owner.fightDuration * 1000)} DPS`}
+          label={(
+            <dfn data-tip={`Total damage increase is ${formatNumber(danceDamage)} in ${this.shadowDance.totalShadowDanceCast} Shadow Dance casts. Dark Shadow increase 30% damage during Shadow Dance.`}>
+              Increased from Dark Shadow talent
```
kfinch: Remove `Dark Shadow increase 30% damage during Shadow Dance.` , this is just a restatement of Dark Shadow's tooltip, which will be present in the statistic box display because you're using Dark Shadow as the icon.

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadowEvis.js
```diff
@@ -0,0 +1,45 @@
...
+
+  statistic() {    
+    const danceEvis = this.shadowDance.totalEviscerateDamageInShadowDance / (this.shadowDance.totalShadowDanceCast * 2);
+    return (
+        <StatisticBox
+          icon={<SpellIcon id={SPELLS.EVISCERATE.id} />}
+          value={`${formatPercentage(danceEvis)} %`}
+          label={(
+            <dfn data-tip={`Your Eviscerate casts in Shadow Dance / (Shadow Dance casts * 2). Your actual / max possible casts is ${danceEvis}. This number includes Eviscerates cast from Death from Above. Subtlety rogue should cast as many as possible (usually 2 times) Eviscerates in a Shadow Dance to get benefit from 30% damage increasing of Dark Shadow talent.`}>
+            Actual/Possible Eviscerates in Shadow Dance
```
kfinch: This is a confusing stat... maybe redo this to be "average Eviscerate casts during Shadow Dance"?

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadow.js
TsaboTavok: Its base for 2 other DarkShadow modules, right now only checks if they need to be enabled

src/Parser/Rogue/Subtlety/Modules/Features/ShadowBladesUptime.js
TsaboTavok: fixed

src/Parser/Rogue/Subtlety/Modules/Features/ShadowBladesUptime.js
TsaboTavok: fixed

src/Parser/Rogue/Subtlety/Modules/Features/ShadowDance.js
TsaboTavok: fixed

src/Parser/Rogue/Subtlety/Modules/Features/ShadowDance.js
TsaboTavok: fixed

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadowContribution.js
TsaboTavok: fixed

TsaboTavok:
Hey, thanks for the review. Original idea was just to refactor and then start fixing the features 1 by 1, but might as well make some changes now.

src/Parser/Rogue/Subtlety/CombatLogParser.js
TsaboTavok: I contacted Martijn, its just the way I've done it before and I don't really have a strong opinion on using it here or not.

MartijnHols: I'd prefer keeping it consistent across all specs unless we're going to deprecate one method in favor of another. Having alternative solutions to simple problems would make it harder for us to make cross-spec improvements.

I think it's a good idea, I've been thinking about if we could apply it everywhere but there are several problems, and I don't think the variances I considered solve those either.
I don't want to have to edit 2 files instead of 1 to enable modules for no real tangible gain. I was also thinking about returning a default from the Modules index.js file like the index file in the SPELLS and ITEMS directories and other locaitons, but that would mean we have 2 different approach for core and spec-specific modules and it gets harder with the alternative when there are options (`damageDone: [DamageDone, { showStatistic: true }],`). Another alternative would have been to just move the entire specModules into the index.js but I don't think it makes things easier to use.

As for the rest I trust @kfinch his judgment.

TsaboTavok: Alright, removed for now.

src/Parser/Rogue/Subtlety/Modules/Talents/DarkShadowEvis.js
```diff
@@ -0,0 +1,35 @@
...
+
+class DarkShadowEvis extends DarkShadow {
+  static dependencies = {
+    enemies: Enemies,
+    combatants: Combatants,
+    shadowDance: ShadowDance,
+  };
+
+  statistic() {    
+    const danceEvis = this.shadowDance.totalEviscerateDamageInShadowDance / this.shadowDance.totalShadowDanceCast;
```
kfinch: `totalEviscerateDamageInShadowDance` is a pretty confusing name for what's actually 'total evis casts in shadow dance'. I understand you increment it on damage events, but still.

PR #833, opened 2017-11-19
[Frost Mage] Update to new Cast Efficiency

Sharrq:
Updated to new Cast Efficiency & Removed Icy Veins/Frozen orb Suggestions


Sharrq:
Yes, that is what i was going for. I wanted to reduce it to default.

If it turns out to be too low, i can raise it again later, but 95% is a bit much.

kfinch:
Neato

PR #834, opened 2017-11-19
[core] Added support for multiple spellIds to abilities

kfinch:
Abilities can now define 'additionalSpells', to be used associating multiple cast Ids that share a cooldown, or to associate a buff Id with a differing cast Id.

Used this to add Grimoire of Service to cast efficiency, rather than needing custom module.


kfinch:
@sMteX I believe this mostly obsoletes the Grimoire of Service cast effic module, you can see in Files Changed what I've done. The custom cast effic thresholds were copied from your module, if you'd prefer to fallback on defaults let me know.

grimoire

src/Parser/Core/Modules/Abilities.js
```diff
@@ -29,6 +29,7 @@ class Abilities extends Analyzer {
      * Available properties:
      *
      * required spell {object} The spell definition with { id, name and icon }
+     * optional additionalSpells {array} An array of additional spell definitions with { id, name and icon }. Can be used if an ability has multiple cast IDs, or to associate a buff ID with a cast ID. The 'spell' field is still required even if you define this, as the spell field is used for the user facing display of the ability.
```
MartijnHols: I think instead of adding another property, spell should also allow for an array to be passed to make usage simpler.

kfinch: I know it looks awkward, but I did it this way for two reasons:
1) Makes it possible to specify the 'main' spell when showing user facing elements. See in the image above the suggestion is about Grimoire of Service because that's the one in the 'spell' field.
2) Easier on dev, don't have to check type of spell every time I use it.

Looking at usage of ABILITIES, it's not currently touched from that many places, so I could change it. I guess I'd specify the 'main' spell by whichever is first in the array?

MartijnHols: I'd rather have spell be of type `object|arrayOf(object)` with a second property indicating which is main, than have spell mean two things and having a second property for extending it if you get what I mean.

Selecting first would make sense I think. I can't think of a situation where the main spell can't be first.

kfinch: Technically the spell field doesnt mean two things, because it counts as one of the IDs to be tracked even if additionalSpells is defined. 

But ok ok, I'll change it.

kfinch:
Changed it as requested. I think it made the code uglier, but it does appear to work.

src/Parser/Core/Modules/SpellHistory.js
```diff
@@ -27,11 +27,20 @@ class SpellHistory extends Analyzer {
       // We're only interested in abilities in Abilities since that's the only place we'll show the spell history, besides we only really want to track *casts* and the best source of info for that is Abilities.
       return null;
     }
-    if (!this.historyBySpellId[spellId]) {
-      this.historyBySpellId[spellId] = [];
+
+    let mainSpellId;
+    if(Array.isArray(ability.spell)) {
```
MartijnHols: same

src/Parser/Core/Modules/Abilities.js
```diff
@@ -121,10 +121,16 @@ class Abilities extends Analyzer {
    */
   getAbility(spellId) {
     return this.constructor.ABILITIES.find(ability => {
-      if (ability.spell.id === spellId) {
-        return !ability.isActive || ability.isActive(this.combatants.selected);
+      if (Array.isArray(ability.spell)) {
```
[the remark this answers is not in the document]
kfinch: Oh, sure

src/Parser/Core/Modules/SpellHistory.js
kfinch: yaya

kfinch:
done

PR #835, opened 2017-11-19
Adding new Sub Abilities.

TsaboTavok:
Adding new ~~Cast Efficiency~~ Abilities

### DIFF
--- a/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
+++ b/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
@@ -60,6 +60,55 @@ class Abilities extends CoreAbilities {
       recommendedCastEfficiency: 0.80,
       extraSuggestion: "Should be used with every Symbols of Death", 
     },
+
+    //Non-CDs
+    
+    {
+      spell: SPELLS.NIGHTBLADE,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    },    
+    {
+      spell: SPELLS.EVISCERATE,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    },    
+    {
+      spell: SPELLS.BACKSTAB,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      isActive: combatant => !combatant.hasTalent(SPELLS.GLOOMBLADE_TALENT.id),
+      getCooldown: haste => null,
+    },       
+    {
+      spell: SPELLS.GLOOMBLADE_TALENT,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      isActive: combatant => combatant.hasTalent(SPELLS.GLOOMBLADE_TALENT.id),
+      getCooldown: haste => null,
+    }, 
+    {
+      spell: SPELLS.SHADOWSTRIKE,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    }, 
+    {
+      spell: SPELLS.SHURIKEN_STORM,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    },
+    {
+      spell: SPELLS.SHURIKEN_TOSS,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: haste => null,
+    },    
+    {
+      spell: SPELLS.MARKED_FOR_DEATH_TALENT,
+      category: CastEfficiency.SPELL_CATEGORIES.ROTATIONAL,
+      isActive: combatant => combatant.hasTalent(SPELLS.MARKED_FOR_DEATH_TALENT.id),
+      getCooldown: haste => null,
+    },
+
+
+
   ];
 }
 
--- a/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
+++ b/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
@@ -11,8 +11,9 @@ class Abilities extends CoreAbilities {
       spell: SPELLS.VANISH,
       category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
       getCooldown: haste => 90,
-      //Vanish is often delayed.
-      recommendedCastEfficiency: 0.90,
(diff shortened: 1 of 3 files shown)


src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
```diff
@@ -20,16 +22,78 @@ class Abilities extends CoreAbilities {
...
     {
       spell: SPELLS.SYMBOLS_OF_DEATH,
-      category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: haste => 30, // TODO: Add T20 4Sets, reduce cd by 5 sec.
+      category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+      getCooldown: (haste, combatant) => (30 - (combatant.hasBuff(SPELLS.SUB_ROGUE_T20_2SET_BONUS.id) ? 5 : 0)), 
       recommendedCastEfficiency: 0.95,
+      importance: ISSUE_IMPORTANCE.MAJOR,
+      extraSuggestion: "This is the most important rotational ability, try to allways use it on cooldown.",      
     },
```
Pewtro: Allways is spelt ``always``

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
```diff
@@ -20,16 +22,78 @@ class Abilities extends CoreAbilities {
...
     {
       spell: SPELLS.SHADOW_DANCE,
-      category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+      category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
       getCooldown: haste => 60, // TODO: Reduced by a passive.
       charges: 2,
-      recommendedCastEfficiency: 0.95,
+      recommendedCastEfficiency: 1,
+    },
```
Pewtro: you have something with a recommendedCastEfficiency of 1, but no extraSuggestion to tell people how they should be using it if they don't follow this metric?

TsaboTavok: This is old code, 1 is more correct then 0.95. Changed to noSuggestion, because the metric is not usable before cooldown reduction is implemented.

kfinch: Just so you know, the framework for cooldown reduction is handled for Cast Efficiency as of yesterday. The logic for the reduction still needs to be implemented on your end, if that's what you meant by it not being usable.

Pewtro: And fwiw it doesn't have any ``noSuggestion`` added either

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
```diff
@@ -10,26 +11,90 @@ class Abilities extends CoreAbilities {
       spell: SPELLS.VANISH,
       category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
       getCooldown: haste => 90,
-      recommendedCastEfficiency: 1.0,
+      //Vanish is often delayed.
+      recommendedCastEfficiency: 0.90,
     },
```
Pewtro: Adding a bit more annoying comments here: You sure you don't want anything here in the suggestion about it often being delayed?

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
```diff
@@ -10,26 +11,90 @@ class Abilities extends CoreAbilities {
...
       charges: 2,
-      recommendedCastEfficiency: 0.95,
+      noSuggestion: true,
+    },
+    {
+      spell: SPELLS.GOREMAWS_BITE,
+      category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+      getCooldown: haste => 60,
+      recommendedCastEfficiency: 0.9,
+    },
```
Pewtro: You don't want any extraSuggestion here to aid the player in how this should be used? afaik it's generally used in combination with other stuff aswell, ye?

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
TsaboTavok: Yeah I saw that you were working on it, but I will first need to implement Combo Point tracking so this has to be delayed =)

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
TsaboTavok: Vanish usage depends a on legionaries and talents, for now I'm trying to get all of the general features in. Putting a line like "try not to delay it too much" doesn't seem helpful to me.

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
TsaboTavok: The exact condition again depends on talents but I think I found suitable wording for it

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
```diff
@@ -10,26 +11,92 @@ class Abilities extends CoreAbilities {
...
+      noSuggestion: true,
+    },
+    {
+      spell: SPELLS.GOREMAWS_BITE,
+      category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+      getCooldown: haste => 60,
+      recommendedCastEfficiency: 0.9,
+      // TODO: Track usage sepratly, not during Dance and with atleast 50 energy missing. 
+      extraSuggestion: "Use as often as possible without waisting Combo Points and Energy.", 
+    },
```
Pewtro: ``waisting`` should be ``wasting``

src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
Pewtro: It's helpful until you've implemented those things, and if the legendaries do get implemented will they override this, or how will that be handled? If they don't, it should some sort of suggestion regardless.
A general rule of thumb as an extraSuggestion or the likes is generally implemented and sounds helpful to me, if that is possible. 
I understand you want to get the general features implemented, but telling people they're not casting X spell enough without giving them a reason to as to why they're casting it too little or when they should be casting it more, can just lead to confusion or incorrect play in their hunt to use it more.

TsaboTavok: Ok agreed, no suggestions for now.

PR #836, opened 2017-11-20
Fixed division by zero Edgecase

HawkCorrigan:
This "fix" might break some things, I'm not sure how often formatPercentage was used and needs to return a number. Definitely affects RestoShaman and RestoDruid.
Shows N/A instead of Nan in casting efficiencies if things like chainheal are not used.

I mostly wanna have some feedback if a change like this is wanted before i start checking other abilities/specs.

### DIFF
--- a/src/Parser/Shaman/Restoration/Modules/Spells/ChainHeal.js
+++ b/src/Parser/Shaman/Restoration/Modules/Spells/ChainHeal.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { formatPercentage} from 'common/format';
 import SpellIcon from 'common/SpellIcon';
 import SpellLink from 'common/SpellLink';
 import SPELLS from 'common/SPELLS';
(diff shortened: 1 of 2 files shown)


HawkCorrigan:
Yup just noticed that too.
Question still stands if a change like this is wanted in general.

kfinch:
As for the general concept of handling NaN / null / undefined inputs to the format functions, you may have something of a point, it's a little bit weird that a few of the format functions coerce those values to zero. I'm not convinced that returning "N/A" is the best solution, maybe it would be better to return NaN in these cases, and let the end user do with it as they will.

kfinch:
For example, returning "N/A" from formatPercentage wouldn't work even when formatPercentage is used correctly, because things using it offer a trailing "%", meaning you'd end up displaying "N/A %". 

You could make formatPercentage include the % symbol itself, and then remove the percents after it from everything that calls formatPercentage, but then there is a spacing issue. Stats boxes that include a percent in big lettering look better with a space in-between the number and the '%', while smaller percentages in suggestions and tooltips look better without the space. Currently that spacing is handled by the dev instead of the format function.

HawkCorrigan:
I am asking because at the moment it return NaN average targets if no chainheals were used.
IMO NaN is the worst thing that could be displayed, because most users don't know what it means.
I thought about displaying it as "0 average targets hit", but that's kind of wrong so i don't really like it either.

First i just checked it with a ternary operator in the html-tags, but that seems to be kind of hacky, so I thought about running it through a format-Function.
formatPercentage already used the 2 digits after decimal points, but I missed the part of multiplying it by 100 so ye, that won't work.

So maybe just have a formatNumber(int precision) which doesn't change the number itself? IDK tbh

kfinch:
For your specific case, just do `${isNaN(avgHits) ? "N/A" : avgHits.toFixed(2)}` for the statistic. For the suggestion, you should probably do an early return if avgHits is NaN.

The more I think of it, the more I believe the format functions should be about formatting numbers, and should leave it up the dev to check that their value is actually a number before passing it in.

PR #838, opened 2017-11-20
[Discipline Priest] Fixed value for Undead Racial

karlpralow:
In the Touch of the Grave Undead racial, the direct healing wasn't added in the count. The spell generates healing from atonement but also direct healing on self. This PR adds the healing to the value of the racial and gives a breakdown in the mouse over.

### DIFF
--- a/src/Parser/Priest/Discipline/Modules/Spells/TouchOfTheGrave.js
+++ b/src/Parser/Priest/Discipline/Modules/Spells/TouchOfTheGrave.js
@@ -18,7 +18,8 @@ class TouchOfTheGrave extends Analyzer {
     combatants: Combatants,
   };
 
-  healing = 0;
+  atonementHealing = 0;
+  directHealing = 0;
   damage = 0;
 
   on_byPlayer_damage(event) {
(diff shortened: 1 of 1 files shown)


src/Parser/Priest/Discipline/Modules/Spells/TouchOfTheGrave.js
```diff
@@ -28,32 +29,45 @@ class TouchOfTheGrave extends Analyzer {
...
+
+    const atonementPct = this.owner.getPercentageOfTotalHealingDone(atonementHealing);
+    const directPct = this.owner.getPercentageOfTotalHealingDone(directHealing);
+    const totalPct = atonementPct + directPct;
+
+    // since we can't directly get undead racial status, if we found 0 damage,
     // assume they aren't undead and do not load the module
     if(damage === 0) { return; }
-
+    console.log(this.directHealing);
```
MartijnHols: You left this in here. While you're at it could you change the above return statement to return `null` instead? It's preferred to either always return nothing (void) or always return a value, but not switch between the two.

PR #839, opened 2017-11-20
[Marksmanship] Fix infinity error on trueshot

Pewtro:
If one were to precast trueshot and not cast it anymore in a fight, you'd end up with this result: 
image

This updates fixes it so that it'll end up looking like this: 
image

### DIFF
--- a/src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
+++ b/src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
@@ -25,13 +25,28 @@ class Trueshot extends Analyzer {
   totalCastsPrTS = 0;
   executeTrueshots = 0;
   startFocusForCombatant = 0;
+  prepullTrueshots = 0;
+
+  on_byPlayer_applybuff(event) {
+    const buffId = event.ability.guid;
+    if (buffId !== SPELLS.TRUESHOT.id && !event.prepull) {
+      return;
+    }
+    //ensures we only get 1 of these events
+    if (this.trueshotCasts === 0) {
+      //adds 1 to trueshotCasts to properly show that it was cast prepull
+      this.trueshotCasts += 1;
+      this.prepullTrueshots += 1;
+    }
+  }
 
   on_byPlayer_cast(event) {
-    //checks for the combatants start focus and assigns it once - useful if the combatant has precast trueshot
-    if (this.startFocusForCombatant === 0) {
+//checks if we had a prepull trueshot, in which case the firstCast done symbolises our starting focus of that one trueshot
+    if (this.prepullTrueshots > 0 && this.startFocusForCombatant === 0) {
       event.classResources.forEach(classResource => {
         if (classResource.type === RESOURCE_TYPES.FOCUS && classResource['amount'] > this.startFocusForCombatant) {
           this.startFocusForCombatant += classResource['amount'];
+          this.accumulatedFocusAtTSCast += this.startFocusForCombatant;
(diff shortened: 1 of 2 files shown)


Pewtro:
Didn't know I could make prepull stuff like that, how nice. 
I've moved it into a prepull check aswell as added an uptime suggestion to the module.

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
```diff
@@ -23,23 +24,46 @@ class Trueshot extends Analyzer {
...
   totalCastsPrTS = 0;
   executeTrueshots = 0;
+  startFocusForCombatant = 0;
+  prepullTrueshots = 0;
 
-  on_byPlayer_cast(event) {
+  on_byPlayer_applybuff(event) {
     const buffId = event.ability.guid;
-    if (buffId !== SPELLS.TRUESHOT.id) {
+    if (buffId !== SPELLS.TRUESHOT.id && !event.prepull) {
```
kfinch: Shouldn't this be `||` not, `&&` ?

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
```diff
@@ -23,23 +24,46 @@ class Trueshot extends Analyzer {
...
       return;
     }
-    this.trueshotCasts += 1;
-    this.accumulatedFocusAtTSCast += event.classResources[0]['amount'] || 0;
-    if (this.combatants.selected.hasBuff(SPELLS.BULLSEYE_TRAIT.id, event.timestamp)) {
-      this.executeTrueshots += 1;
+    //ensures we only get 1 of these events
+    if (this.trueshotCasts === 0) {
```
kfinch: I'm unclear on how you could ever get more than one, is this a 'just in case' kind of thing?

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
Pewtro: Well I only care about prepull events and trueshot events in this section, since everything else is handled in ``on_byPlayer_cast``

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
Pewtro: It registered more than one if I didn't do it, if I recall correctly, but otherwise, ye just a "just in case"

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
kfinch: Yes, you only care about an event that is Trueshot AND Prepull, but this statement is for an early return. The logical negation of "A && B" is "!A || !B". For a concrete example, a buff that is Prepull but isn't Trueshot would not early return as currently written.

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
kfinch: It registered more than one because your early return statement is incorrect, as I noted above.

src/Parser/Hunter/Marksmanship/Modules/Spells/Trueshot.js
Pewtro: right... fuck me - fixed

PR #843, opened 2017-11-21
[core] Fixed an issue where spell timelines weren't rendered properly for spells with charges

### LINKED ISSUE #793: Cooldowns of abilities with stacks don't work properly in the timeline tab
image

The correct behavior would be:
1. Show it on cooldown as long as at least 1 charge is recharging
2. Show it on cooldown when no charges are available

2 would be weird since then the first cast would show no cooldown. I think the best solution is to show both stacks with a opacity of like 33% so that the indicator gets darker when multiple stacks are on cooldown.

kfinch:
DON'T MERGE just yet

Fixes #793 - I chose option 1 for fixing this, because option 1 was obviously the correct choice.

The reason I say "don't merge this yet" is because the cooldown tooltip on charged spells is now a bit janky -- it shows the total amount of contiguous time the spell was recharging, which typically has nothing to do with the spell's actual cooldown. I put up this PR to solicit ideas as to how this should be handled. My current best idea is that some sort of marker should delineate each gained charge, and the tooltip on each segment should display the amount of time it took that charge to be gained.

### DIFF
--- a/src/Main/Timeline/Events.js
+++ b/src/Main/Timeline/Events.js
@@ -52,6 +52,8 @@ class Events extends React.PureComponent {
     const { events, start, totalWidth, secondWidth, className } = this.props;
     const fixedEvents = this.fabricateEndCooldown(events);
 
+    let beginTimestamp;
+
     return (
       <div className={`events ${className || ''}`} style={{ width: totalWidth }}>
         {fixedEvents.map((event, index) => {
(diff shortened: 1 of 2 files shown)


kfinch:
Ok, what I currently have is definitely better than what's on live, but a few issues to work out. In some cases it looks like cooldown bars are stacking up for spells with stacks, which is not intended. Also there is no delimiter between stack recharges, I'd like some sort of tick mark or something in between recharges, and I'm not sure the best thing to put there.

kfinch:
Ok delimiter added, now two issues remaining:
1) sometimes stacked cooldown bars for spells with stacks
2) sometimes missing end cooldown bar for spells with stacks

kfinch:
Ahaha those two issues were actually the same issue! Got them both fixed, and also fixed the icon vertical alignment issue.

kfinch:
timeline-with-stacks

PR #847, opened 2017-11-22
[Windwalker] Tier 21 4PC item breakdown

Juko8:


src/Parser/Monk/Windwalker/Modules/Items/T21_4set.js
```diff
@@ -0,0 +1,60 @@
...
+     if (spellId === SPELLS.BLACKOUT_KICK.id && this.comboBreaker === true) {
+         this.damage += (event.amount + (event.absorbed || 0))*(175/275);
+     }
+ }
+
+ on_byPlayerPet_damage(event) {
+     const spellId = event.ability.guid;
+
+     if (spellId === SPELLS.BLACKOUT_KICK.id && this.comboBreaker === true) {
+         this.damage += event.amount + (event.amount + (event.absorbed || 0)) * (175 / 275);
```
kfinch: use the `calculateEffectiveDamage` helper from src/Parser/Core/calculateEffectiveDamage here, also move the `(175 / 275)` into a named constant pls (no magic numbers)

src/Parser/Monk/Windwalker/Modules/Items/T21_4set.js
```diff
@@ -0,0 +1,60 @@
...
+ on_initialized() {
+     this.active = this.combatants.selected.hasBuff(SPELLS.WW_TIER21_4PC.id);
+ }
+
+ // Combo Breaker buff fades after Blackout Kick is cast but before it deals damage
+ on_byPlayer_cast(event) {
+  const spellId = event.ability.guid;
+
+  if (spellId === SPELLS.BLACKOUT_KICK.id && this.combatants.selected.hasBuff(SPELLS.COMBO_BREAKER_BUFF.id)) {
+    this.comboBreaker = true;
```
kfinch: You're setting this true when conditions are met, but then you're never unsetting it. This will result in incorrect behavior as all Blackout Kicks after the first proc will be counted as buffed whether or not they actually are.

You could easily fix by setting the flag false when the buffed damage happens, but let me suggest instead using the little known 'bufferTime' feature of the hasBuff function. Basically, you could skip the 'comboBreaker' flag entirely, and instead on_byPlayer_damage your check would be `if (spellId === SPELLS.BLACKOUT_KICK.id && this.combatants.selected.hasBuff(SPELLS.COMBO_BREAKER_BUFF.id), null, 500)) {` Check out the documentation for this feature in src/Parser/Core/Entity.js

Juko8: This seems to break when trying to do it with the pet damage check since the pets don't have the buff

kfinch: I'm not sure what you mean.. you shouldn't check the pet for the buff, just check the player.

kfinch:
You could almost certainly get away with a buffer bigger than 32 ms, but yeah this looks good. I'll merge once Travis build has passed.

PR #859, opened 2017-11-22
Sub rogue - resource tracker - changes from CR

TsaboTavok:
Made changes suggested in CR.


src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
```diff
@@ -52,7 +52,7 @@ class ResourceTracker extends Analyzer {
     this.generated += gain;  
 
     const eventResource = this.getResource(event);
-    if(eventResource !== undefined) {
+    if(!!eventResource) {
       this.current = eventResource.amount;
     }
```
Pewtro: Shouldn't this just be ``if(!eventResource)``

src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
```diff
@@ -74,7 +74,7 @@ class ResourceTracker extends Analyzer {
     
     if (!(spellId in this.spendersObj)) {
       this.initSpenderAbility(spellId);
-  }
+    }
 
     const eventResource = this.getResource(event);
```
Pewtro: you should include this, as said in the original review, to ensure no crashes (i've been bugfixing tons in my modules because of this, despite thinking stuff isn't possible, it will inevitably happen, might aswell be safe) 
```jsx 
if (!cost) { 
      return; 
}

src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
MartijnHols: Actually it's checking if it *is* defined, so in this instance no exclamation mark at all is required to turn into just `if(eventResource) {`. `!!` converts a value to a boolean, but JS will do this automatically anyway. It's only really useful if you want to convert a random value (1, 0, "1", "", []) into a boolean without inverting its value.

PR #864, opened 2017-11-23
Changed Cast Efficency for Sheild of the Righteous

yajinni:


src/Parser/Paladin/Protection/Modules/Features/Abilities.js
```diff
@@ -39,7 +39,7 @@ class Abilities extends CoreAbilities {
     {
       spell: SPELLS.SHIELD_OF_THE_RIGHTEOUS,
       category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-      getCooldown: haste => null,
+      getCooldown: haste => 16,
```
kfinch: I'm pretty sure SotR scales with haste. Also if there is anything else that modifies cooldown, it should go here.

yajinni: your right. it does scale with haste. Will make changes.

yajinni:
Changes have been made. It was initially marked like that because we are still in the most begining stages.
As i mentioned without it you get hit buy trucks as paladins and 1 main source of damage. They stack haste so they can try to use it as much as possible. So tracking it is important. Its left at a more lax 80% so to give wiggle room to those that might be saving it up for a big event. But it also has 3 charges, so at high haste its around for most of the time.

PR #866, opened 2017-11-23
[Windwalker] Added Chi breakdown tab, suggestions and chi wasted statistic

Juko8:
The formatting in the Chi tab isn't the prettiest right now, but i'm not very experience with editing it. It does show everything, so it can always be improved later


kfinch:
Looks cool, but SOMETHING'S NOT QUITE RIGHT where I've circled.

chi

Juko8:
I honestly have no clue how to fix that

kfinch:
Well what's it supposed to look like? Just how it is but without the `|` ? In any case, check where it's rendered in ChiBreakdown.

Juko8:
It's a bar like the ones showing chi generated, but i don't know how to hide it when width is 0, i kinda expected it to just be hidden in that case. Googling it didn't help me either.

kfinch:
ohhhh, just a sec I think I can figure it out

src/Parser/Monk/Windwalker/Modules/Chi/ChiBreakdown.js
```diff
@@ -0,0 +1,130 @@
...
+                    />
+                  </td>
+                  <td style={{ width: 50, paddingRight: 5, textAlign: 'center' }}>
+                    <dfn data-tip={`${formatPercentage(ability.wasted / totalWasted)} %`}>{ability.wasted}</dfn>
+                  </td>
+                  <td style={{ width: '30%' }}>
+                    <div
+                      className={'performance-bar '}
+                      style={{ width: `${(ability.wasted / totalWasted) * 100}%` }}
+                    />
```
kfinch: Maybe wrap this in a conditional, only render when `ability.wasted === 0`

kfinch:
Although actually could just leave it, I didn't realize they were even supposed to be bars

kfinch:
Yeah ok I take it back, it's fine.

PR #870, opened 2017-11-25
Unholy - Runic Power/Festering Strike Module

Maxweii:
Updating values on Runic Power wasted percentages.


Maxweii:
Went through things with Khazzak about Festering Strike values, found it was a rather high value so lowered to reflect in game performance more. Runic Power values as well for same reason.

PR #874, opened 2017-11-26
Sub Rogue - Cooldown Reductions for Dance and T21_2P Symbols

TsaboTavok:
Just adding cooldown reductions.

### DIFF
--- a/src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
+++ b/src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
@@ -92,6 +92,12 @@ class ResourceTracker extends Analyzer {
 
     //Re-sync current amount, to update not-tracked gains.
     this.current = eventResource.amount - cost;
+
+    this.onSpent(cost);
+  }
+
+  onSpent(spent) {
+
   }
   
   shouldProcessCastEvent(event) {
--- a/src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
+++ b/src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
@@ -10,21 +10,21 @@ class ComboPointTracker extends ResourceTracker {
 		spellUsable: SpellUsable,
   };
 
+  danceReduction;
+
   on_initialized() {
     this.resourceType = 4;
     this.resourceName = "Combo Points";
+    this.danceReduction = this.combatants.selected.hasTalent(SPELLS.ENVELOPING_SHADOWS_TALENT.id) ? 2500 : 1500;
   }
 
-  onSpent(spent)  {    
-    const combatant = this.combatants.selected;
-
+  onSpent(spent)  {
     if (this.spellUsable.isOnCooldown(SPELLS.SHADOW_DANCE.id)) {
-      const reduction = combatant.hasTalent(SPELLS.ENVELOPING_SHADOWS_TALENT.id) ? 2500 : 1500;
-
-			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SHADOW_DANCE.id, reduction * spent );
+			this.spellUsable.reduceCooldown(SPELLS.SHADOW_DANCE.id, this.danceReduction * spent );
     }
-    if(combatant.hasBuff(SPELLS.SUB_ROGUE_T21_2SET_BONUS.id)){
-			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SYMBOLS_OF_DEATH.id, 200 * spent );
+    if(this.combatants.selected.hasBuff(SPELLS.SUB_ROGUE_T21_2SET_BONUS.id 
+      && this.spellUsable.isOnCooldown(SPELLS.SYMBOLS_OF_DEATH.id))){
+			this.spellUsable.reduceCooldown(SPELLS.SYMBOLS_OF_DEATH.id, 200 * spent );
     }
   }
 }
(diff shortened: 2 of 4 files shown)


src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
```diff
@@ -92,6 +92,12 @@ class ResourceTracker extends Analyzer {
 
     //Re-sync current amount, to update not-tracked gains.
     this.current = eventResource.amount - cost;
+
+    this.onSpent(cost);
+  }
+
+  onSpent(spent) {
+
   }
```
Pewtro: You've added a method that does nothing? Either remove it for now, or include what it's supposed to do

src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
```diff
@@ -1,16 +1,32 @@
...
     this.resourceType = 4;
     this.resourceName = "Combo Points";
   }
+
+  onSpent(spent)  {    
+    const combatant = this.combatants.selected;
+
+    if (this.spellUsable.isOnCooldown(SPELLS.SHADOW_DANCE.id)) {
+      const reduction = combatant.hasTalent(SPELLS.ENVELOPING_SHADOWS_TALENT.id) ? 2500 : 1500;
+
```
Pewtro: Move this ``const reduction`` into your ``on_initialized`` please. Alternatively you can create a constant called REDUCTION above the class indicator, and then in ``on_initialized`` add 1000 to it if the player has the talent, that would be the most consistent with other similiar modules.

src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
```diff
@@ -1,16 +1,32 @@
...
+			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SHADOW_DANCE.id, reduction * spent );
+    }
+    if(combatant.hasBuff(SPELLS.SUB_ROGUE_T21_2SET_BONUS.id)){
+			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SYMBOLS_OF_DEATH.id, 200 * spent );
+    }
```
Pewtro: You need to add a check for Symbols of death actually being off cooldown, or this can lead to errors/crashes.

src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
```diff
@@ -1,16 +1,32 @@
...
+  onSpent(spent)  {    
+    const combatant = this.combatants.selected;
+
+    if (this.spellUsable.isOnCooldown(SPELLS.SHADOW_DANCE.id)) {
+      const reduction = combatant.hasTalent(SPELLS.ENVELOPING_SHADOWS_TALENT.id) ? 2500 : 1500;
+
+			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SHADOW_DANCE.id, reduction * spent );
+    }
+    if(combatant.hasBuff(SPELLS.SUB_ROGUE_T21_2SET_BONUS.id)){
+			this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.SYMBOLS_OF_DEATH.id, 200 * spent );
```
Pewtro: You've never defined ``cooldownReduction``, yet you keep adding to it - and you never do anything with either - this needs to be defined if you intend to do something with it, otherwise it's unnecessary and should be removed.

src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
TsaboTavok: I added a hook that can be used by implementations. By default it does nothing, Who is this wrong?

Pewtro: Where and how do you currently use it? If no where, it's unnecessary code and shouldn't be in this PR.

TsaboTavok: In ComboPointTracker.

Pewtro: Ok, I see now. You made ResourceTracker handle both energy AND combo-points. That could be a point of future refactoring for clarity possibly 😄

Pewtro:
Leaving open for now for more chance of more feedback.

src/Parser/Rogue/Subtlety/Modules/RogueCore/ComboPointTracker.js
```diff
@@ -1,15 +1,31 @@
...
   static dependencies = {
     combatants: Combatants,
+		spellUsable: SpellUsable,
   };
 
+  danceReduction;
+
   on_initialized() {
     this.resourceType = 4;
     this.resourceName = "Combo Points";
```
kfinch: While it probably doesn't matter, resourceType and resourceName probably should be set baseline, rather than in on_initialized()

src/Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker.js
```diff
@@ -92,6 +92,12 @@ class ResourceTracker extends Analyzer {
 
     //Re-sync current amount, to update not-tracked gains.
     this.current = eventResource.amount - cost;
+
+    this.onSpent(cost);
+  }
+
+  onSpent(spent) {
```
kfinch: This is an interesting design, though I think it would be more sensibly implemented as a callback registry. Basically keep an array of `_onSpentCallbacks`, do something like 
```
onSpent(callback) {
  this._onSpentCallbacks.push(callback);
}
```
... and then when the module detects resources spent, it does `this._onSpentCallbacks.forEach(callback => callback(amount, event));` ... this pattern allowing other analyzers to register things that happen on resource spend, rather than having to pack it all into the ComboPointTracker / EnergyTracker.

TsaboTavok: So Observable pattern? Sounds good, will change it when I work on it next. I imagine I can register observers in on_initialized ?

PR #890, opened 2017-11-28
Add a stats on pull display (fixes #708)

### LINKED ISSUE #708: [Core] Show info box with stats
A nice little info-box with the stats the player had on this specific kill: 

Something like this: 
image

Maybe structured in this manner and placed under items/legendaries/tier boxes: 

Stat | Percentage | Rating | 
-----|-------------|---------
Mastery | X % | Y (number) | 
Crit | X % | Y (number) | 
Haste | X % | Y (number) | 
Verstatility | X % | Y (number) | 

Also acts as an alternative (in terms of this information possibly being more relevant to DPS) to the statweights developed for healers. 

A further development could be to include how much you **actually** critted during said encounter (maybe on mouseover tooltip or something of the like)

MartijnHols:
image

Positioned under items panel.


MartijnHols:
I realized I messed a few things up like not including base percentages in the shown stats. I'll push fixes tonight.

MartijnHols:
Fixed all issues

MartijnHols:
Ohh accidentally also added a talent display.

image

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -236,45 +247,61 @@ class StatTracker extends Analyzer {
       case SPECS.RESTORATION_DRUID:
         return 66667;
       default:
-        throw new Error('Mastery hasn\'t been implemented for this spec yet.');
+        console.error('Mastery hasn\'t been implemented for this spec yet.');
+        return 99999999;
```
kfinch: Ok, so instead of crashing the StatDisplay will now simply display a wrong answer?

MartijnHols: It will show 0, which is obviously wrong

kfinch: Won't it show base mastery, which is less obviously wrong? Also, do you want to push to display something that is obviously wrong?

I get that you're probably looking at including all of the Mastery conversion factors as not your problem, but I would argue that now that you're adding stat percentage display it *is* your problem. 

The good news is there is apparently a simpler formula for the different specs Mastery conversion, which I am looking into.

MartijnHols: I looked into it briefly, it did not seem like an easy to tackle problem and I figured if it's obviously wrong it won't be misleading. It should be easy to implement for spec maintainers once they get to it.

kfinch: https://i.imgur.com/G1F1ECN.jpg

PR #903, opened 2017-12-01
Added Carafe of searing light

buimichael:


Pewtro:
I've merged the disc version of this, so if you could update to use those spells (they are currently in the item folder, however upon further inspection I do prefer your version of this with the spells in OTHERS, but if you do a fresh pull and update both to use the same spells one way or the other I'll merge) 
So either update to use his version and I'll fix it (since it's my bad), or you do a pull -> remove his and update his module to use your spells. 

Other than that, it looks ready to go

PR #904, opened 2017-12-01
[core] Fixed a bug where changebuffstack events were often given the wrong timestamp

kfinch:
The `...buff` used when triggering the `changebuffstack` contains the timestamp field from the internal buff object, which is the timestamp when the buff was *initially applied*, rather than the timestamp the stack changed.

This was causing some truly esoteric bugs in the Haste module.


kfinch:
This bug was causing particular problems in the Ret Paladin timeline / cast effic, because it mostly comes into play when there is a stacking haste buff and also CDs that scale with haste; Ret has both those things.

kfinch:
The wrong timestamp was being passed through into `changehaste` fabbed events, and then SpellUsable listens `on_changehaste` and uses that event's timestamp to calculate `timePassed`. The resulting timePassed would end being negative, which then fucked up the rest of the calculations.

MartijnHols:
good job

PR #905, opened 2017-12-01
[frost mage] Tier21 2pc and Mastery

kfinch:
Added support for Tier21 2pc. Flurry bolt tracking is a little try-hard because of delay issues, but I believe it will work correctly in nearly all situations, and will work close to correctly in very odd (high haste) situations.

I wanted to add the 4pc too, but haven't yet been able to find any logs with it.

Also added Frost Mage mastery ratios so that Mastery percent shows correctly on the stats pane.


Sharrq:
Thanks! I kept telling myself i needed to get to that but first week of new raid has been a little crazy

Pewtro:
It's good but needs a merge conflict fix

kfinch:
I shouldn't have merged other people's Mastery implementations 😛

kfinch:
@Pewtro 
I should say the fix is for a more rare a case than just "high haste", it's basically "high haste, and the player gets two procs in a row, and the player makes the (incorrect) choice to use them back to back"

Also please don't start the merge unless you're going to finish it, things are a little messed up on the branch now.

kfinch:
Ok gottem

Pewtro:
Yeah I did finish the original merge, then travis was slow so I moved onwards 😛

PR #906, opened 2017-12-01
[Retribution] Added in Ret mastery

CollCrom:


src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -5,7 +5,7 @@ import Analyzer from 'Parser/Core/Analyzer';
 
 import Abilities from './Abilities';
 
-const debug = false;
+const debug = true;
```
kfinch: 🤔

CollCrom: Whoops <_<

PR #908, opened 2017-12-01
Sub Rogue mastery

TsaboTavok:
Just supporting mastery


src/Parser/Core/Combatant.js
```diff
@@ -88,6 +88,8 @@ class Combatant extends Entity {
 	      return 0.048 + this.masteryRating / 66666.6666666;
       case SPECS.RETRIBUTION_PALADIN:
         return 0.14 + this.masteryRating / 22850;
+	    case SPECS.SUBTLETY_ROGUE:
+	      return 0.2208 + this.masteryRating / 144.9261221;
       default:
```
Pewtro: You sure about this number here should probably be increased with a factor of 100

TsaboTavok: Sorry, was contemplating why I have to do this in 2 places in the first place.

PR #910, opened 2017-12-01
Added garothi feedback conduit

buimichael:
resto druids gets extra love as usual

### DIFF
--- a/src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
+++ b/src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
@@ -1,11 +1,12 @@
 import React from 'react';
 
 import ITEMS from 'common/ITEMS';
+import SPELLS from "common/SPELLS";
+import { formatThousands } from "common/format";
+import { calculateSecondaryStatDefault } from 'common/stats';
 
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
-import SPELLS from "../../../../common/SPELLS";
-import { formatThousands } from "../../../../common/format";
 
 const BASE_ITEM_LEVEL = 935;
 const BASE_HASTE_VALUE = 872;
@@ -14,7 +15,6 @@ class GarothiFeedbackConduit extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
-  currentILvl = 0;
   baseILvl = 935;
   staticHaste = 0;
   lastProcc = null;
@@ -22,12 +22,12 @@ class GarothiFeedbackConduit extends Analyzer {
   currentUptime = 0;
   currentProccValue = 0;
 
-  totalAverageHasteGained = 0;
   on_initialized() {
     this.active = this.combatants.selected.hasTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id);
     if(this.active) {
-      this.currentILvl = this.combatants.selected.getTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id).itemLevel;
-      this.staticHaste = Math.round(BASE_HASTE_VALUE* Math.pow(1.15,((this.currentILvl-BASE_ITEM_LEVEL)/15)) * Math.pow(0.994435486, (this.currentILvl-BASE_ITEM_LEVEL)));
+      const currentILvl = this.combatants.selected.getTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id).itemLevel;
+      this.staticHaste = calculateSecondaryStatDefault(BASE_ITEM_LEVEL, BASE_HASTE_VALUE, currentILvl);
+      console.log("static haste: " + this.staticHaste);
     }
   }
 
@@ -53,21 +53,15 @@ class GarothiFeedbackConduit extends Analyzer {
     }
   }
 
-  on_finished() {
-    this.totalProccValue.forEach(function(item) {
-      const uptimePercent = item[1]/this.owner.fightDuration;
-      this.totalAverageHasteGained += uptimePercent * item[0];
-    }.bind(this));
-  }
-
   item() {
+    const avgHaste =
+      this.totalProccValue.reduce((acc, proc) => acc + (proc[0] * proc[1]), 0) / this.owner.fightDuration;
+
     return {
       item: ITEMS.GAROTHI_FEEDBACK_CONDUIT,
       result: (
-        <dfn
-          data-tip={`This is the average haste rating gained</b>`}
-        >
-          {formatThousands(this.totalAverageHasteGained)} average haste rating gained.
+        <dfn>
+          {formatThousands(avgHaste)} average haste rating gained.
         </dfn>
       ),
     };
--- a/src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
+++ b/src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
@@ -1,0 +1,77 @@
+import React from 'react';
+
+import ITEMS from 'common/ITEMS';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SPELLS from "../../../../common/SPELLS";
+import { formatThousands } from "../../../../common/format";
+
+const BASE_ITEM_LEVEL = 935;
+const BASE_HASTE_VALUE = 872;
+
+class GarothiFeedbackConduit extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  currentILvl = 0;
+  baseILvl = 935;
+  staticHaste = 0;
+  lastProcc = null;
+  totalProccValue = [];
+  currentUptime = 0;
+  currentProccValue = 0;
+
+  totalAverageHasteGained = 0;
+  on_initialized() {
+    this.active = this.combatants.selected.hasTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id);
+    if(this.active) {
+      this.currentILvl = this.combatants.selected.getTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id).itemLevel;
+      this.staticHaste = Math.round(BASE_HASTE_VALUE* Math.pow(1.15,((this.currentILvl-BASE_ITEM_LEVEL)/15)) * Math.pow(0.994435486, (this.currentILvl-BASE_ITEM_LEVEL)));
+    }
+  }
+
+  on_byPlayer_changebuffstack(event) {
+    if (event.ability.guid !== SPELLS.FEEDBACK_LOOP.id) {
+      return;
+    }
+    if(this.lastProcc == null) {
+      this.lastProcc = event.timestamp;
+    } else {
+      this.currentUptime += event.timestamp - this.lastProcc;
+      this.currentProccValue = event.oldStacks * this.staticHaste;
+      this.lastProcc = event.timestamp;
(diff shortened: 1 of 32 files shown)


Pewtro:
This will need an update now that carafe got added

src/Parser/Druid/Restoration/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,60 @@
...
+      this.totalIntellect -= 325;
+    }
+    if(spellId === SPELLS.FLASK_OF_THE_WHISPERED_PACT.id) {
+      this.totalIntellect -= 1300;
+    }
+  }
+
+  item() {
+    const hpmWeight = this.statWeights._getGain(STAT.HASTE_HPM) / (this.statWeights.totalOneInt || 1);
+    const throughput = (hpmWeight * this.totalAverageHasteGained) / this.totalIntellect;
```
kfinch: The `_getGain()` function literally returns the amount of healing gained from one rating over the course of the encounter. Just multiply it by the average rating gain to get throughput, no reason to play games with Int.

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import ITEMS from 'common/ITEMS';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SPELLS from "../../../../common/SPELLS";
+import { formatThousands } from "../../../../common/format";
```
kfinch: Use absolute path for stuff in common / Core

buimichael:
Note: I may have been a little brain dead implementing this after a 12 hours work session. I don't know how to compare the values in WCL but the results looks what's expected from the trinkets.

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,77 @@
...
+  totalProccValue = [];
+  currentUptime = 0;
+  currentProccValue = 0;
+
+  totalAverageHasteGained = 0;
+  on_initialized() {
+    this.active = this.combatants.selected.hasTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id);
+    if(this.active) {
+      this.currentILvl = this.combatants.selected.getTrinket(ITEMS.GAROTHI_FEEDBACK_CONDUIT.id).itemLevel;
+      this.staticHaste = Math.round(BASE_HASTE_VALUE* Math.pow(1.15,((this.currentILvl-BASE_ITEM_LEVEL)/15)) * Math.pow(0.994435486, (this.currentILvl-BASE_ITEM_LEVEL)));
```
kfinch: You should check out `common/stats/calculateSecondaryStatDefault`

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,77 @@
...
+        this.lastProcc = null;
+      }
+    }
+  }
+
+  on_finished() {
+    this.totalProccValue.forEach(function(item) {
+      const uptimePercent = item[1]/this.owner.fightDuration;
+      this.totalAverageHasteGained += uptimePercent * item[0];
+    }.bind(this));
```
kfinch: Please use a lambda instead. `function(item) { ... }.bind(this)` is literally identical to `item => { ... }`

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  item() {
+    return {
+      item: ITEMS.GAROTHI_FEEDBACK_CONDUIT,
+      result: (
+        <dfn
+          data-tip={`This is the average haste rating gained</b>`}
```
kfinch: Don't include tooltip if it's just a restatement of the main display.

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
buimichael: Woops, my IDE seems to auto add them like that :/ fixed

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
buimichael: Oh, neat.

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
buimichael: Done

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
buimichael: :)

src/Parser/Druid/Restoration/Modules/Items/GarothiFeedbackConduit.js
buimichael: Doing this approach showed a lot less throughput total rather than taking the avgHaste * hpmValue / totalInt. Any idea why?

kfinch: Uhh that's a good question, I'll take a closer look.

src/Parser/Core/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,70 @@
...
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+const BASE_ITEM_LEVEL = 935;
+const BASE_HASTE_VALUE = 872;
+
+class GarothiFeedbackConduit extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  baseILvl = 935;
```
kfinch: This is a duplicate of BASE_ITEM_LEVEL

src/Parser/Druid/Restoration/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -0,0 +1,64 @@
...
+      this.totalIntellect -= 1300;
+    }
+  }
+
+  item() {
+    const avgHaste =
+      this.totalProccValue.reduce((acc, proc) => acc + (proc[0] * proc[1]), 0) / this.owner.fightDuration;
+    //TODO: Add the healing from each haste rating in the future:
+    const hpmWeight = this.statWeights._getGain(STAT.HASTE_HPM) / (this.statWeights.totalOneInt || 1);
+    const throughput = (hpmWeight * avgHaste) / this.totalIntellect;
```
kfinch: Ok, I've had the time to more closely look over this calculation.

Psuedocode -
Your throughput calculation: `(hasteHealingFromOne / intHealingFromOne) * avgHaste / "totalInt"`
Honestly the more I look at it the less sense it makes. The units aren't even right, see: `(healing / healing) * scalar / statAmount = 1 / statAmount` You want units of healing, and ended up with 1 / statAmount. All of this is even ignoring that 'totalInt' doesn't properly reflect procs. It's a miracle this produced a value that looked kinda correct.

My suggested throughput calculation: `hasteHealingFromOne * avgHaste`
This just takes the healing gained from one haste, and multiplies by haste gain. Units: `healing * scalar = healing`.

kfinch: Ohh, your calculation is supposed to be percentage of total throughput, not raw throughput. The calculation is still wrong, but I better understand how it got a number that looked right.

buimichael: The reasoning for having a throughput number is for users more easily compare with other trinkets. 
For example- Let's say we have another trinket with a healing procc, it has the same amount of INT as Garothis but the healing procc did 3% healing. How do you as a user decide if 3% healing is better than 1.9k haste rating? 
* Normally (as in Tortys spreadsheet too afaik) we translate 3% healing to INT by taking 0.03 * totalIntUnbuffed. 
* So for a character with 70k INT, 3% healing would be worth 2100 INT. 
* Assuming haste is weighted at 1.3, the 1.9k haste rating gained by Garothis would be worth 2.47k int. Thus making Garothis stronger.
* Using the variable ``hasteHealingFromOne`` that would yield roughly 2.25% in this example, showing Garothis way worse than the healing proc trinket.

Now I understand that your ``hasteHealingFromOne`` *should* yield the correct healing value per rating, but it seems odd that it differs by 2/3 of the expected values comparing to the values in the spreadsheets. This may cause some confusion for users trying to verify/calculate the worth of the trinket I think.

buimichael: On another hand, I guess the ``hasteHealingFromOne`` should yield the most accurate values, since it's a haste procc, it may procc during times of downtime and maybe that's why it's showing a lot less value than expected?

buimichael: To extend, i'd like to take the approach of using ``xxxHealingFromOne`` variables from constant stat buffs such as stat sticks and see if they yield a more close to the expected values.

kfinch: Ah, I understand how you came up with this calculation now. The calculation's error is the assumption that `gainedInt / totalUnbuffedInt = percentGainedThroughput`. I would expect the resulting number to overstate things significantly. Using the unbuffed int value is just wrong, especially considering how many int buffs there are. Once flask, food, rune, and concordance are factored in, the player's actual int will always be higher than the unbuffed int, and it's the actual int that matters when trying to calculate percent throughput gain. Also, the set of spells that scale from int is different from the set that scale with haste, that too could factor in. Also, the hasteHealingFromOne value ignores Tranq, while your the way you've done your calculation effectively unignores Tranq; that's another explanation for your value being higher.

The purpose of the StatValues calculator was to remove a lot of the guesswork inherent in Torty's spreadsheet. Your calculation just adds that guesswork back in.

buimichael: Fair enough ^^ I'll push the changes tomorrow

src/Parser/Druid/Restoration/Modules/Items/GarothiFeedbackConduit.js
```diff
@@ -3,58 +3,31 @@ import { formatThousands, formatPercentage, formatNumber } from 'common/format';
...
       item: ITEMS.GAROTHI_FEEDBACK_CONDUIT,
       result: (
         <dfn
-          data-tip={`The throughput shown is calculated by using the int on "Stats on pull" and the stat weights of HPM.</b>`}
+          data-tip={`The throughput shown is calculated by using the amount of healing one rating of haste yielded and multiplied with the average haste gained.
+            The amount of healing one rating of haste yielded was ${formatNumber(this.statWeights._getGain(STAT.HASTE_HPM))}</b>`}
         >
           {formatThousands(avgHaste)} average haste rating gained.<br />
-          {formatPercentage(throughput)} % / {formatNumber((this.healingDone.total.effective * throughput) / this.owner.fightDuration * 1000)} HPS <br />
+          {formatPercentage(healing / this.healingDone.total.effective)} % / {formatNumber(healing / this.owner.fightDuration * 1000)} HPS <br />
```
kfinch: One final nitpick, but you could just use `this.owner.formatItemHealingDone(healing)` here

PR #912, opened 2017-12-02
[Beast Mastery] Fixes Tier192p, adds Cobra Commander module

Pewtro:


Pewtro:
Okay, I removed the event.ability.name, and replaced it with another filter that may or may not break it on some logs if they resummon pets or something else that I haven't accounted for - hopefully not tho.

Pewtro:
Ok, I've filtered through quite a few logs, and haven't found one that breaks this solution.

src/Parser/Hunter/BeastMastery/Modules/Traits/CobraCommander.js
```diff
@@ -0,0 +1,53 @@
...
+    if (!selectedSneakySnake) {
+      return;
+    }
+    this.sneakySnakeDamage += event.amount;
+  }
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.COBRA_COMMANDER.id} />}
+        value={`${formatNumber(this.sneakySnakeDamage)}`}
```
MartijnHols: You don't need to use a string literal when not merging strings.

PR #923, opened 2017-12-03
[Sub Rogue] Specter usage

TsaboTavok:
Specter  remains the BIS trinket for Antorus on ST fights.
Checking that it is used together with Dance to benefit from 30% Dark Shadow buff.


TsaboTavok:
Yeah I think I need to relax a lot of the requirements on various abilities, but right now this is in line with all  other recommendations that I have.  I'll change this one now, but I still plan to make a full parse of all of my suggestions, and try to have one off mistakes fall in to Minor or Average categories.

PR #928, opened 2017-12-03
[core] Core items update (E through P)

kfinch:
ALL - Added item's tooltip to comment at top of file, and also added new `primaryStat` field to SPECS.

EngineOfEradication - Partial rework where average stat gain is now displayed instead of uptime, and where average orbs collected is also tracked. Added suggestion when player isn't good about collecting orbs.

Eonar's Compassion - Improved tooltips layout for readability. Info displayed should be the same.

Highfather's Machination - Fixed a bug where multiple stacks being consumed by a heal wasn't being counted properly, and also improved tooltip display.

Infernal Cinders - Removed superfluous tooltip.


src/common/PRIMARY_STATS.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
MartijnHols: Did you consider that according to feasts "stamina" is a primary stat for tanks?

kfinch: I did consider it, but I wasn't sure of a solution. I thought for other things tanks had their damage stat as primary?

kfinch: There's no actual bug yet because the new field is only hooked into EoE.

kfinch: Ok my main argument for doing it this way is that it's easy to convert to stam for a tank:
```
if (spec.role === ROLES.TANK) {
  return 'Stamina';
} else {
  return spec.primaryStat;
}
```
But if I made STAMINA primary stat for all tank specs, there'd be no way to get their 'damage primary'.

PR #938, opened 2017-12-05
[Frost Mage] Lower Thermal Void Avg Time

Sharrq:
Lowered threshold for Thermal Void warnings partly because it wont be as possible to hit 45 seconds as T20 4pc fades away. Also because 45 seconds was a tad bit high considering its partly based on RNG

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
+++ b/src/Parser/Mage/Frost/Modules/Features/ThermalVoid.js
@@ -45,7 +45,7 @@ class ThermalVoid extends Analyzer {
           .icon(SPELLS.ICY_VEINS.icon)
           .actual(`${formatNumber(actual)} seconds Average Icy Veins Duration`)
           .recommended(`${formatNumber(recommended)} is recommended`)
-          .regular(35).major(30);
+          .regular(37).major(33);
       });
   }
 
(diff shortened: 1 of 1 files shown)


Sharrq:
Yea i used 45 because the 80% parses were all around 47ish, but i think ive personally only managed to get to 45 like twice when with two frozen orbs. Been thinking about lowering it for a while and just never pulled the trigger. But without t20 4pc it is going to be a lot harder to get a second frozen orb within TV so the 5 proc reduction was mostly to compensate for that.

i forgot that Thermal Void was changed to add 10 seconds to it and therefore thought it was 20 seconds base thats why the major was 30, Ill adjust it real quick before raid time.

kfinch:
Oh, I didn't realize really high parses were getting 47.... and oh I just realized why they must be using Frozen Touch rather than Splitting Ice. The highest tier should be hard to achieve, maybe it's OK at 45? I almost think the thresholds should be different depending on if you have FT or not.

Sharrq:
Even with Frozen Touch though, its RNG, its not like they could have played better to get those extra 5 procs (most of the time). But i think getting that high mostly requires getting 2 frozen orbs within the TV Window, and i think thats just going to be borderline impossible without t20 4pc. 

I might lower it to 40 seconds for now and wait for a bigger pool of logs to review and adjust once people start getting t21.

PR #943, opened 2017-12-07
BaseHealerStatValues exposes the amount of healing one rating does

buimichael:
This is useful for comparing static stat values. I.e. if you're comparing a log where you had a trinket with healing procc with a stat stick. Or just interested in how much healing each stat did.

img

### DIFF
--- a/src/Parser/Core/Modules/Features/BaseHealerStatValues.js
+++ b/src/Parser/Core/Modules/Features/BaseHealerStatValues.js
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import SPELLS from 'common/SPELLS';
-import { formatNumber } from 'common/format';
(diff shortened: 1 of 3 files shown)


MartijnHols:
`16396 healing gained per 1 rating` what kind of healing? total or per second? Also please use the formatThousands formatter on the number.

buimichael:
Actual healing. So as a user if you're curious about how much healing the 450 leech stat you have on your wrists you'd take 16396 * 450 =  7378000 healing. Which is also the reason i'm not use formatThousand as these kind of calculation would be heavily skewed by the rounding.

MartijnHols:
formatThousands does no significant rounding

MartijnHols:
Also could you please change the tooltip to "total healing" then?

buimichael:
Done

PR #947, opened 2017-12-07
[Discipline Priest] Atonement per Penance Bolt in Atonement Source Tab

kpralow:
image

### DIFF
--- a/src/Parser/Priest/Discipline/CHANGELOG.js
+++ b/src/Parser/Priest/Discipline/CHANGELOG.js
@@ -1,4 +1,6 @@
 export default `
+07-12-2017 - Added healing per penance bolt in Atonement Source Tab (by Oratio)
+01-12-2017 - Added Support for Carafe of Searing Light (by Oratio)
 15-11-2017 - Moved Soul of the High Priest value to show the talent Twist of Fate (by Gao)
 15-11-2017 - Added a toggle absolute healing in atonement sources (By Oratio)
 10-13-2017 - Fixed a bug with some trinket damage events causing a crash. (By Gao)
--- a/src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
+++ b/src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
@@ -57,7 +57,6 @@ class AtonementHealingBreakdown extends React.Component {
               const performanceBarPercentage = healing.effective / highestHealing;
 
               return ([
-
                 <tr key={ability.guid}>
                   <td style={{ width: '30%' }}>
                     <SpellLink id={ability.guid}>
@@ -84,11 +83,11 @@ class AtonementHealingBreakdown extends React.Component {
                   </td>
                 </tr>
 
-              ,(bolts && bolts.map((name,index) => {
+              , (bolts && bolts.map((value,index) => {
 
-                const penanceBarPercentage = name / healing.effective;
+                const penanceBarPercentage = value / healing.effective;
 
-                if(!name) return null;
+                if(!value) return null;
 
                 return (<tr>
                   <td style={{ width: '30%', paddingLeft:50 }}>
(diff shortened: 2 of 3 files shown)


src/Parser/Priest/Discipline/CHANGELOG.js
```diff
@@ -1,4 +1,6 @@
 export default `
+07-12-2017 - Added healing per penance bolt in Atonement Source Tab (by Oratio)
```
MartijnHols: String-based changelog style is deprecated. Please convert to the new format (see other specs/core) when possible.

src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
```diff
@@ -52,11 +52,11 @@ class AtonementHealingBreakdown extends React.Component {
           {bySource && Object.keys(bySource)
             .sort((a, b) => bySource[b].healing.effective - bySource[a].healing.effective)
             .map((spellId) => {
-              const { ability, healing } = bySource[spellId];
+              const { ability, healing, bolts } = bySource[spellId];
 
               const performanceBarPercentage = healing.effective / highestHealing;
 
-              return (
+              return ([
```
MartijnHols: Please use the Wrapper component in `common/Wrapper` instead of returning an array.

src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
```diff
@@ -82,9 +82,42 @@ class AtonementHealingBreakdown extends React.Component {
...
-            })}
-        </tbody>
+
+              , (bolts && bolts.map((value,index) => {
+
+                const penanceBarPercentage = value / healing.effective;
+
+                if(!value) return null;
+
+                return (<tr>
```
MartijnHols: I think it's more readable to start the element on the next line so that indentation is saner.

src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
```diff
@@ -82,9 +82,42 @@ class AtonementHealingBreakdown extends React.Component {
...
+                  <td style={{ width: '30%', paddingLeft:50 }}>
+                    <SpellLink id={ability.guid}>
+                      <Icon icon={ability.abilityIcon} />{' '}
+                      {ability.name} Bolt {index + 1}
+                    </SpellLink>
+                  </td>
+                  {!this.state.absolute && (
+                  <td style={{ width: 60, paddingRight: 5, textAlign: 'center' }}>
+                    {(Math.round(value / totalAtonement.effective * 10000) / 100).toFixed(2)}%
```
MartijnHols: How about replacing `totalAtonement.effective` with `applicableTotal` which is defined above the return statement that takes care of the `this.state.absolute` toggle (on = its value is total, off = atonement relative)? I think this would reduce the amount of duplicate code.

src/Parser/Priest/Discipline/Modules/Features/AtonementHealingDone.js
```diff
@@ -47,6 +60,15 @@ class AtonementHealingDone extends Analyzer {
     this.bySource[spellId] = this.bySource[spellId] || {};
     this.bySource[spellId].ability = ability;
     this.bySource[spellId].healing = (this.bySource[spellId].healing || new HealingValue()).add(amount, absorbed, overheal);
+
+    if(spellId === SPELLS.PENANCE.id) {
+      if(!this.bySource[SPELLS.PENANCE.id].bolts) {
+        this.bySource[SPELLS.PENANCE.id].bolts = [0,0,0,0];
```
MartijnHols: While I don't expect Penance to gain any bolts any time soon, I think it would be better to write this without having to hard-code the amount of bolts. How about `this.bySource[SPELLS.PENANCE.id].bolts[this._lastPenanceBoltNumber] = (this.bySource[SPELLS.PENANCE.id].bolts[this._lastPenanceBoltNumber] || 0) + amount + (absorbed || 0);`?
ps. amount is always defined, absorbed may be undefined. The old method would not include any healing with absorbed not defined.

src/Parser/Priest/Discipline/CHANGELOG.js
kpralow: Oh nice new feature. I'll update the complete log in another PR right after this

src/Parser/Priest/Discipline/Modules/Features/AtonementHealingBreakdown.js
kpralow: I tried working with Wrappers but because these were <tr> it would break the display when I tried adding them in. Am I missing something that would let me do it easyly?

MartijnHols: The Wrapper component doesn't add any elements to the DOM so should be fine in a table.

kpralow: If I replace [ by <wrapper> and ] by </wrapper> and if I surround the 2nd block with {}, I get this 
image

MartijnHols: It needs to be capitalized, and you need to import the component from `common/Wrapper`. See other modules for examples.

kpralow:
@MartijnHols Everything should be fixed from your review, thank you

PR #956, opened 2017-12-09
Added 'copy Pawn String to clipboard' button to StatValues panel

kfinch:
Zero help me! I have no idea how to CSS a button at all! This looks like garbage!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -146,7 +146,7 @@
     "ansi-styles": {
       "version": "3.2.0",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.0.tgz",
-      "integrity": "sha512-NnSOmMEYtVR2JVMIGTzynRkkaxtiq1xnFBcdQD/DnNCYPoEPsVJhM98BDyaoNOQIi7p4okdi3E27eN7GQbsUug==",
+      "integrity": "sha1-wVm41b4PnlpvNG2rlPFs4CIWG4g=",
       "requires": {
         "color-convert": "1.9.0"
       }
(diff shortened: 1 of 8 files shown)


kfinch:
Currently doesn't look like a button at all:

shitty-button

kfinch:
OK I tried... Dunno how to make shift up a bit tho.

ok-button

Pewtro:
``vertical-align`` maybe?

kfinch:
Oh I think when I tried vertical align I put it as a class instead of a style because I was way too tired.

MartijnHols:
I think a copy to clipboard solution would be a better suited as a function than a React component. This could also be done with a library, but probably not the one you're using now. Then you could just programmatically give feedback instead of using a focus-hack.

Also Travis is failing.

kfinch:
@MartijnHols can you explain how a function would solve this? I'm having trouble envisioning what that would look like.

The Travis fail is just an imports thing, I knew this wasn't ready for merge so I just hadn't fixed it yet.

MartijnHols:
Button onClick -> trigger method that calls the function to copy a string to clipboard & shows a toaster that it happened. I believe we want full control over how such a button looks like, so I think using a React component to do it seems undesirable.

kfinch:
Good point, the React thing did seem awkward... Maybe I'm using the wrong library, but it wasn't obvious how to bind copying to a function. 

Also can you point to am example in code of this function binding? I dunno why, but I'm having real trouble thinking about what this would look like.

MartijnHols:
https://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript

kfinch:
Hopefully I'll have time to get back to this and get this fixed before I go on vacation...

kfinch:
I'll reimplement something that actually works some other time....

PR #963, opened 2017-12-10
Acrid Catalyst Injector

CollCrom:
screen shot 2017-12-10 at 5 09 19 pm

Could probably stand to refactor a bit. Am open to suggestions about this implementation.


CollCrom:
also the wowhead link is totally wrong on the effect.

http://www.wowdb.com/items/151955-acrid-catalyst-injector?bonusIDs=3610,1472

kfinch:
I have an idea of how I could generalize 'average stat gain' for use by all Trinket modules, using Entity and StatTracker. I'm sure it will be rad if I get around to implementing.

kfinch:
@CollCrom see the helper method in #995 , this should help with implementing this.

PR #965, opened 2017-12-11
destro mods

superderpicus:
a couple mods were pointless, so i removed them (fnb, demonfire). Couple of mods made no sense, like eradication/life tap bonus damage, instead of uptime, so I combined it to show both. havoc is not needed if you didnt cast it, added line to hide it when its pointless.


src/Parser/Warlock/Destruction/Modules/Talents/Backdraft.js
```diff
@@ -83,7 +83,7 @@ class Backdraft extends Analyzer {
     );
   }
 
-  statisticOrder = STATISTIC_ORDER.OPTIONAL(0);
+  statisticOrder = STATISTIC_ORDER.CORE(5);
```
kfinch: The 'OPTIONAL' order is typically used for talents, which Backdraft apparently is. What's your justification for moving it to core?

src/Parser/Warlock/Destruction/Modules/Talents/ChannelDemonfire.js
```diff
@@ -1,43 +0,0 @@
...
-
-import Analyzer from 'Parser/Core/Analyzer';
-import Combatants from 'Parser/Core/Modules/Combatants';
-
-import SPELLS from 'common/SPELLS';
-import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
-import SpellIcon from 'common/SpellIcon';
-import { formatNumber, formatPercentage } from 'common/format';
-
-class ChannelDemonfire extends Analyzer {
```
kfinch: Can you explain why this module should be deleted?

src/Parser/Warlock/Destruction/Modules/Talents/EmpoweredLifeTap.js
```diff
@@ -54,7 +54,7 @@ class EmpoweredLifeTap extends Analyzer {
     );
   }
 
-  statisticOrder = STATISTIC_ORDER.OPTIONAL(3);
```
kfinch: Same as with Backdraft, this is a talent right?

src/Parser/Warlock/Destruction/Modules/Talents/Eradication.js
```diff
@@ -27,26 +28,39 @@ class Eradication extends Analyzer {
...
-        label="Damage contributed"
+        value={`${formatPercentage(uptime)} %`}
+        label="Eradication Uptime"
         tooltip={`Your Eradication talent contributed ${formatNumber(this.bonusDmg)} total damage (${formatPercentage(this.owner.getPercentageOfTotalDamageDone(this.bonusDmg))} %)`}
       />
     );
   }
 
-  statisticOrder = STATISTIC_ORDER.OPTIONAL(2);
+  statisticOrder = STATISTIC_ORDER.CORE(4);
```
kfinch: Same as with Backdraft, this is a talent right?

src/Parser/Warlock/Destruction/Modules/Talents/Eradication.js
```diff
@@ -27,26 +28,39 @@ class Eradication extends Analyzer {
...
-  statistic() {
+  statistic() {    
+    const uptime = this.enemies.getBuffUptime(SPELLS.ERADICATION_DEBUFF.id) / this.owner.fightDuration;
     return (
       <StatisticBox
         icon={<SpellIcon id={SPELLS.ERADICATION_TALENT.id} />}
-        value={`${formatNumber(this.bonusDmg / this.owner.fightDuration * 1000)} DPS`}
-        label="Damage contributed"
+        value={`${formatPercentage(uptime)} %`}
+        label="Eradication Uptime"
```
kfinch: For talents, we typically like to express the effectiveness of the talent in terms of percentage damage gained, so that it can be used as a point of reference vs. other talents in the same tier. That's why this showed damage instead of uptime. It could be worthwhile to show both, or one in the stats box and the other in the tooltip.

src/Parser/Warlock/Destruction/Modules/Talents/FireAndBrimstone.js
```diff
@@ -1,69 +0,0 @@
...
-import Combatants from 'Parser/Core/Modules/Combatants';
-
-import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
-import SpellIcon from 'common/SpellIcon';
-import { formatNumber, formatPercentage } from 'common/format';
-import SPELLS from 'common/SPELLS';
-
-import SoulShardEvents from '../SoulShards/SoulShardEvents';
-
-class FireAndBrimstone extends Analyzer {
```
kfinch: Can you explain why this module should be deleted?

src/Parser/Warlock/Destruction/Modules/Talents/SoulHarvestTalent.js
```diff
@@ -21,18 +21,19 @@ class SoulHarvestTalent extends Analyzer {
...
+    const bonusDmg = this.soulHarvest.talentBonusDmg;    
+    const uptime = this.combatants.selected.getBuffUptime(SPELLS.SOUL_HARVEST_TALENT.id) / this.owner.fightDuration;
     return (
-      <StatisticBox
+      <SmallStatisticBox
         icon={<SpellIcon id={SPELLS.SOUL_HARVEST_TALENT.id} />}
-        value={`${formatNumber(bonusDmg / this.owner.fightDuration * 1000)} DPS`}
-        label="Damage contributed"
+        value={`${formatPercentage(uptime)} %`}
+        label="Soul Harvest Uptime"
```
kfinch: Same story as Eradication, we've preferred damage contribution over uptime, to make talent effects comparable.

src/Parser/Warlock/Destruction/Modules/Talents/Eradication.js
superderpicus: that’s what I did, the damage is still in the tooltip, but it’s more helpful to see uptime in the box.

src/Parser/Warlock/Destruction/Modules/Talents/ChannelDemonfire.js
superderpicus: takes up too much space, could opt for a small stat box but still ultimately pretty pointless to see dps from demonfire

src/Parser/Warlock/Destruction/Modules/Talents/Backdraft.js
superderpicus: I have no idea how this works and this made it go where I wanted

src/Parser/Warlock/Destruction/Modules/Talents/FireAndBrimstone.js
superderpicus: takes up too much space for what it is, if you really wanted to keep it for “comparison” reasons highly suggest small box, looks bad as it was

src/Parser/Warlock/Destruction/Modules/Talents/SoulHarvestTalent.js
superderpicus: dmg is still there

superderpicus:
what is this for? areas of improvement or talent comparison? I will make a module later for talents that puts them in one of those fancy lil boxes like crucible, they don’t deserve a big box

kfinch:
@superderpicus There is a lot of variance over the specs, but generally the statistic boxes are for talent comparison/effectiveness and core spell effectiveness. We use 'suggestions' for informing the player of areas where they need to improve.

There is currently a refactor in the works for making suggestions better organized and more prominently displayed, but for now if a module detects something the player is messing up, it should be output as a suggestion.

You do need to talk to @sMteX before I can merge this, as I believe he is the maintainer of the Destro analyzer, and I'm not interested in overseeing a spec power struggle via PRs.

src/Parser/Warlock/Destruction/Modules/Talents/Backdraft.js
kfinch: Ordering of stats boxes is left to right and then top to bottom, with some possible rearrangement based on box sizes. Core comes first, then Optional. It's generally been the style that Core is for core spells / mechanics, and Optional is for talents.

Changing these orders may have put these boxes where you wanted for the report you were testing with, but a player with different talents selected could mix things up again. (Due to some boxes being missing or others being added by the nature of which talents were selected)

kfinch:
Also, re: ` I will make a module later for talents that puts them in one of those fancy lil boxes like crucible, they don’t deserve a big box`

I think that is a good idea, but don't remove functioning modules now with the promise that their data will be added in at an unspecified later date. Either keep them in as small stats boxes, or leave the modules unmolested and only remove them in the same PR that you add a 'talents box'

superderpicus:
right

Pewtro:
Just adding my two-cents to this. 

Removing/altering anything without talking to the current maintainer is, frankly, rude and should be avoided.
Removing something without creating the replacement in the same PR, **unless** it's getting removed for a good reason, also not good. 
There is an overall precedent and coding style for various modules, which is generally recommended to adhere to, and this includes how talents in general are handled. 
I've reworked my modules talents into a small statistic box, and I'd like that idea to spread to more specs (because I think it's neater), but they are still using damage labels, unless there's a very good reason for them not to. 
Another note on ``I have no idea how this works and this made it go where I wanted``, don't just change something if you don't know how it works or why it does what it does, that's a bad idea in any code project.

superderpicus:
cleary we have a different idea of how this should look so I’m gonna go ahead and bail

PR #970, opened 2017-12-12
Results 2.0

MartijnHols:
# Don't merge yet please. TLDR: checkout, review, give code feedback, add your own spec

This is a rework of the results page. There are quite a few things left to do, but the main parts are in place and working. I'd like to focus this PR on reviews about the checklist, how it's defined, how it works, what it includes, my choice of texts, etc.

The goal of this review is to gather opinions and prepare the branch for checklists for other specs (I'd like to support more than just the Holy Paladin checklist when merging this, going to try to do a "big announcement" since I believe this to be the biggest change since the introduction of suggestions).

When reviewing I recommend checking out this branch and running a Holy Paladin log through it. See instructions at the bottom of this PR.

image
image


kfinch:
Some more styling requests:
- If a spec hasn't yet implemented Checklist, the initially loading analysis should default to the Suggestions tab, if possible.
- I think the 'acceptable' color would look better as a deeper blue.

MartijnHols:
TODO:
* image
* image
* docs

MartijnHols:
TODO:
* Generalize some of the rules and requirements that can be

Pewtro:
I found out why the bar was white - it was whenever I didn't use all 3 arguments (minor, average, major), despite it being for mostly a true/false scenario.

Pewtro:
As I mentioned on Discord, I think there is way too much waste of space in the way it looks right now. 
The below images are me having 6 expandable menus (non of them expanded at that point), yet there is a ton of deadspace until getting to the statistic boxes.
deadspace1
The statistic boxes do not fill out properly, leaving us with this very empty space on the right hand side. 
deadspace2

- Previewing this to other hunters in the community they expressed desire to have everything on the screen, more visible - and not needing to scroll so much, especially in the way it'd end up currently as soon above. Solution would be to:
   - make the hero vs enemy banner waaay smaller (it was taking ~40% of my screen on a 1920x1080  resolution) 
   - Ensure the statistic boxes automatically fill under the checklist to remove deadspace.
   - Make the statistic boxes fill everything below the checklist
- A radical suggestion would be to move the checklist under the statistic boxes, much like we do now and have the suggestions tab be called Checklist/Suggestions and then have the toggle between one or the other be in that tab. 
- On terms of colour, I think it looks way better turning Mage into Rogue, it gives a logical progression in colouring scheme. Red -> Orange -> Yellow -> Green. When messing with that yesterday, I got very positive feedback as opposed to having it as mage or shaman. Might even be worth completely ditching the class colours if it ends up restricting what we want to portray.

MartijnHols:
Hey thanks for the feedback. I think the white space underneath the checklist will slowly disappear as more checklist rules are added. Do you think what you have is a complete list? I realize the amount of empty space varies based on the amount of items in the items panel, but I don't think there's a good solution to making that more consistent; I think hiding items behind a click is worse than scrolling a little.

The scrolling is setup in such a way that most users should find the information they want at the top; the checklist and items. The statistics are a more advanced display hence why they're further down now. Making the statistics fill nicely underneath the checklist would be very hard so unfortunately I won't be working on that. I also believe we should have some breathing room in our UI (this is important in UI design). This comes back to the header which needs the breating room to not make the entire interface look like a mess.

I just changed the extra panels next to statistics to take no space when there is no panel. This should remove that empty space. Do note that unforunately due to the way statistics are configured there's no way to have more than 3 statistics per row at this time.

The reason I chose blue for the *minor* color was primarily because it was the color for minor suggestions 😊 I'm not entirely sure about changing it to yellow as I think that makes the performance appear worse than is intended (blue is more neutral). I'll give it a try and see how it looks.

MartijnHols:
I'm not at all thinking this is the best possible interface, but it's the best I can come up with at this time. I dislike quite a few things about this version, like important elements of the header (e.g. the background the gradient) and the default tab at the bottom. And the stretched out statistics are definitely going to become another one of them, and if there's a lot of empty space underneath the checklist that will bug me a lot too but it's hard to come up with something better within the current approach. Maybe it will bug me enough in a month that I'll start working on another rework.

MartijnHols:
Now previewable on a "test server": https://test.wowanalyzer.com/report/LRchdHVAmWtNQ8Fj/22-Mythic+Harjatan+-+Kill+(5:54)/Zerotorescue

I'll fix tests later, don't worry about those when branching.

MartijnHols:
* Holy Paladin: Add Wasted IoLs to `Use core spells as often as possible`
* Add a tooltip to bars to explain what a filled vs empty bar means and maybe also the thresholds

MartijnHols:
Planning to release this Saturday afternoon (instead of previously mentioned Friday). Will try to appropriately share these big changes with the world (e.g. make a Reddit post).

Issue #685, opened 2017-10-29
Suggestion for enchants


MartijnHols:
Report missing or cheap enchants

PR #974, opened 2017-12-13
EnchantChecker

jjs451:
Aims to address issue #685.  It works just fine but I'm not sure about the style used here so feedback is extra welcome 
image

### DIFF
--- a/src/Parser/Core/Modules/Items/EnchantChecker.js
+++ b/src/Parser/Core/Modules/Items/EnchantChecker.js
@@ -45,7 +45,7 @@ class EnchantChecker extends Analyzer {
     .forEach(item => {
         // inverting everything felt improper but its what made sure only one suggestion fires
         const missingEnchant = !gear[item].hasOwnProperty('permanentEnchant');
-        const missingMaxEnchant = !missingEnchant && !MAX_ENCHANT_IDS.includes(gear[item].permanentEnchant)
+        const missingMaxEnchant = !missingEnchant && !MAX_ENCHANT_IDS.includes(gear[item].permanentEnchant);
         const itemUrl = "https://wowhead.com/item=" + gear[item].id;
         when(missingEnchant).isTrue()
           .addSuggestion((suggest, actual, recommended) => {
--- a/src/Parser/Core/Modules/Items/EnchantChecker.js
+++ b/src/Parser/Core/Modules/Items/EnchantChecker.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import ItemLink from 'common/ItemLink';
-import ITEM_QUALITIES from 'common/ITEM_QUALITIES';
 
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
--- a/src/Parser/Core/Modules/Items/EnchantChecker.js
+++ b/src/Parser/Core/Modules/Items/EnchantChecker.js
@@ -54,11 +54,11 @@ class EnchantChecker extends Analyzer {
               .staticImportance(SUGGESTION_IMPORTANCE.MAJOR);
           });
         when(missingMaxEnchant).isTrue()
-        .addSuggestion((suggest, actual, recommended) => {
-          return suggest(<span>Your <a href={itemUrl}>{ENCHANTABLE_SLOTS[item]}</a> has a cheap enchant.  You should consider upgrading it from a "Word of" enchant to a "Binding of" enchant.</span>)
-            .icon(gear[item].icon)
-            .staticImportance(SUGGESTION_IMPORTANCE.MINOR);
-      });
+          .addSuggestion((suggest, actual, recommended) => {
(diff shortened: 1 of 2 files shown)


src/Parser/Core/Modules/Items/EnchantChecker.js
```diff
@@ -0,0 +1,66 @@
...
+    // iterating with keys instead of value because the values don't store what slot is being looked at
+    Object.keys(gear)
+    .forEach(item => {
+        // inverting everything felt improper but its what made sure only one suggestion fires
+        const missingEnchant = !gear[item].hasOwnProperty('permanentEnchant');
+        const missingMaxEnchant = !missingEnchant && !MAX_ENCHANT_IDS.includes(gear[item].permanentEnchant);
+        const itemUrl = "https://wowhead.com/item=" + gear[item].id;
+        when(missingEnchant).isTrue()
+          .addSuggestion((suggest, actual, recommended) => {
+            return suggest(<span>Your <a href={itemUrl}>{ENCHANTABLE_SLOTS[item]}</a> has a missing enchant.</span>)
```
kfinch: Use SpellLink (https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/common/ItemLink.js) instead of building your own itemUrl.

src/Parser/Core/Modules/Items/EnchantChecker.js
```diff
@@ -0,0 +1,66 @@
...
+              .icon(gear[item].icon)
+              .staticImportance(SUGGESTION_IMPORTANCE.MAJOR);
+          });
+        when(missingMaxEnchant).isTrue()
+          .addSuggestion((suggest, actual, recommended) => {
+            return suggest(<span>Your <a href={itemUrl}>{ENCHANTABLE_SLOTS[item]}</a> has a cheap enchant.  You should consider upgrading it from a "Word of" enchant to a "Binding of" enchant.</span>)
```
kfinch: Use SpellLink (https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/common/ItemLink.js) instead of building your own itemUrl.

src/Parser/Core/Modules/Items/EnchantChecker.js
```diff
@@ -0,0 +1,66 @@
...
+
+  suggestions(when) {
+    const gear = {};
+    Object.keys(ENCHANTABLE_SLOTS).forEach(slot => gear[slot] = this.combatants.selected._getGearItemBySlotId(slot));
+
+    // iterating with keys instead of value because the values don't store what slot is being looked at
+    Object.keys(gear)
+    .forEach(item => {
+        // inverting everything felt improper but its what made sure only one suggestion fires
+        const missingEnchant = !gear[item].hasOwnProperty('permanentEnchant');
```
kfinch: The item object is something we just pulled from WCL's JSON, so I don't believe that using `hasOwnProperty` is strictly needed. Nothing wrong with doing it tho, is fine to leave as is.

src/Parser/Core/Modules/Items/EnchantChecker.js
kfinch: Oh also the line `You should consider upgrading it from a "Word of" enchant to a "Binding of" enchant.` makes sense for Cloak and Ring, but not for Necklace.

You could make the suggestion more generic, or go try-hard and swap in the appropriate string based on slot. It's up to you.

src/Parser/Core/Modules/Items/EnchantChecker.js
jjs451: The goal is just checking if the item has an enchant, if there's another way to check if the property exists I'd be fine with that.

src/Parser/Core/Modules/Items/EnchantChecker.js
jjs451: I'm getting errors trying to use ItemLink, probably because the item needs to exist in ITEMS somewhere to be used.  If using my own link is an issue I could just not link the item.  The item JSON includes the icon, which gets shown and should be enough information for the player.

src/Parser/Core/Modules/Items/EnchantChecker.js
jjs451: Changed to "Your [ITEM] has a missing/cheap enchant.  Apply a strong enchant to increase your DPS/healing/survivability against a challenging boss."

jjs451:
image

src/Parser/Core/Modules/Items/EnchantChecker.js
kfinch: `const missingEnchant = gear[item].permanentEnchant === undefined` or `const missingEnchant = !gear[item].permanentEnchant` if you're not worried about permanentEnchant possibly being zero.

Your way is fine though.

src/Parser/Core/Modules/Items/EnchantChecker.js
kfinch: Oh right, you're linking by id rather than an Item from ITEMS, but still I'm a little surprised that's not working...

I feel like there's a better solution here, but I'm not thinking of it. At the least, you should add `target="_blank" rel="noopener noreferrer"` to the a href

kfinch:
Thanks!

PR #977, opened 2017-12-14
Added Lightning Rod Damage Display

HawkCorrigan:
Not too sure about looks/layout.
Just shows the damage done by the talent.

### DIFF
--- a/src/Parser/Shaman/Elemental/Modules/Talents/LightningRod.js
+++ b/src/Parser/Shaman/Elemental/Modules/Talents/LightningRod.js
@@ -28,12 +28,16 @@ class LightningRod extends Analyzer {
 
   on_byPlayer_damage(event) {
     if (event.ability.guid===SPELLS.LIGHTNING_ROD_DAMAGE.id){
-      this.lightningRodDamage+=event.amount;
+      this.lightningRodDamage+=event.amount+(event.absorbed || 0);
     }
   }
 
-  get rawUpdate() {
+  get getPercentageUptime() {
     return this.enemies.getBuffUptime(SPELLS.LIGHTNING_ROD_DEBUFF.id) / this.owner.fightDuration;
+  }
+
+  get getDamagePerSecond() {
+    return this.lightningRodDamage/this.owner.fightDuration*1000;
   }
 
   statistic() {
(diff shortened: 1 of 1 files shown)


src/Parser/Shaman/Elemental/Modules/Talents/LightningRod.js
```diff
@@ -21,21 +20,33 @@ class LightningRod extends Analyzer {
...
 
+  lightningRodDamage=0;
+
   on_initialized() {
     this.active = this.combatants.selected.hasTalent(SPELLS.LIGHTNING_ROD_TALENT.id);
   }
 
+  on_byPlayer_damage(event) {
+    if (event.ability.guid===SPELLS.LIGHTNING_ROD_DAMAGE.id){
+      this.lightningRodDamage+=event.amount;
```
kfinch: Remember to count absorbed damage, should be `this.lightningRodDamage += event.amount + (event.absorbed || 0);`

src/Parser/Shaman/Elemental/Modules/Talents/LightningRod.js
```diff
@@ -21,21 +20,33 @@ class LightningRod extends Analyzer {
...
+    }
+  }
+
   get rawUpdate() {
```
kfinch: I know this isn't part of your changes, but it's bothering me that this neither 'raw' nor an 'update'. Probably should read `percentUptime`

src/Parser/Shaman/Elemental/Modules/Talents/LightningRod.js
```diff
@@ -21,21 +20,33 @@ class LightningRod extends Analyzer {
...
-      <StatisticBox
-        icon={<SpellIcon id={SPELLS.LIGHTNING_ROD_TALENT.id} />}
-        value={`${formatPercentage(this.rawUpdate)} %`}
-        label={'Uptime'}
-      />
+      <StatisticBox icon={<SpellIcon id={SPELLS.LIGHTNING_ROD_TALENT.id} />}
+        value={(
+          <span>
+            {`${formatPercentage(this.rawUpdate)} %`}<br/>
+            {formatNumber(this.lightningRodDamage)}{' '}
```
kfinch: It's completely inconsistent with style to show raw damage rather than damage as a percentage of total. I assume you're doing this so the player doesn't mix up the uptime and the damage?

It's more typical for spells like this to show the damage alone in the statistic and then the uptime in the tooltip (or vice versa).

HawkCorrigan:
Still don't really satisified with how it looks but dont know how to solve it really.

Pewtro:
You could use the formatItemDamageDone to display the damage done / dps - I generally like how that looks. 

Also you have some stray ``</li></ul>`` in your tooltip

kfinch:
I'll mark approved so it's not waiting on me, but I'll hold off on merge until you comment that you're satisfied with the look.

kfinch:
Ehh I'll assume this is fine, can be tweaked later. Will get at least something in for the trinket.

PR #980, opened 2017-12-14
[resto druid] Rewrote efflo module, fixing the no-cast bug and also opening up easy implementation of 'early refresh' suggestion

### LINKED ISSUE #968: BUG - Efflorescence shows uptime even if efflorescence was never cast.
https://www.warcraftlogs.com/reports/jQrF6DNxHwMRGnth#fight=25&type=healing&translate=true&source=25
https://wowanalyzer.com/report/jQrF6DNxHwMRGnth/25-Mythic+Felhounds+of+Sargeras+-+Kill+(4:59)/%E6%BD%AE%E7%BA%A2

According to the log this druid didn't cast Efflo a single time, yet it shows up as a 10% uptime with 0 healing done

kfinch:
I came in just to fix #968 , and instead rewrote the whole module because it needed it. My rewrite also makes it easy for a later PR to add "don't early refresh Efflo" suggestion.

### DIFF
--- a/src/Parser/Druid/Restoration/Modules/Features/Efflorescence.js
+++ b/src/Parser/Druid/Restoration/Modules/Features/Efflorescence.js
@@ -9,84 +9,66 @@ import Analyzer from 'Parser/Core/Analyzer';
 
 import SuggestionThresholds from '../../SuggestionThresholds';
 
-const debug = false;
+const DURATION = 30000;
 
 class Efflorescence extends Analyzer {
 
-  lastCast = null;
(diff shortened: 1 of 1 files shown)


kfinch:
Because this is basically a rewrite, it may be easier to look at the new version rather than the changes:
https://github.com/kfinch/WoWAnalyzer/blob/e90f76a38dc70180d8b8bfc495c1ed5fd436eff1/src/Parser/Druid/Restoration/Modules/Features/Efflorescence.js

PR #990, opened 2017-12-15
Adding Scourge Strike Efficiency to Unholy Death Knight

Bicepspump:
Added a new module that keeps track of Scourge Strike casts against targets with 0 festering wounds. This is a common mistake for players that don't keep track of their target's debuffs properly.


src/Parser/DeathKnight/Unholy/CombatLogParser.js
```diff
@@ -41,11 +44,12 @@ class CombatLogParser extends CoreCombatLogParser {
     // items
     coldHeart: ColdHeart,
     tier20_2p: Tier20_2p,
+    coldHeart: ColdHeart,
```
jjs451: It would be a good idea to name this coldHeartEfficiency to match the name of the class it comes from, as well as make it clear that this is different than the regular item statistic.

Bicepspump: That's entirely my bad. I'm a bit new to this so I didn't know my new commits would end up in the Push Request I added before.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,86 @@
...
+
+  on_byPlayer_applybuffstack(event){
+    this.addStack(event);
+  }
+
+  addStack(event){
+    const spellID = event.ability.guid;
+    if(spellID === SPELLS.COLD_HEART_BUFF.id){
+      this.buffColdHeart+=1;
+	if(this.buffColdHeart > 20) {
```
jjs451: Is it possible for buffColdHeart to ever go above 20?  Once its at 20 stacks we shouldn't be seeing any more events for it until there's a cast

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,86 @@
...
+	this.buffColdHeart=20;
+	}
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId == SPELLS.CHAINS_OF_ICE.id) {
```
jjs451: Don't forget to use `===` to check for equality.  IIRC Travis sometimes gets unhappy if you use `==`

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,86 @@
...
+       		        	this.correctColdHeartCasts++;
+				}
+			}
+                 }
+	    if (this.buffColdHeart == 20) {
+                 this.correctColdHeartCasts++;
+        }
+  }
+
+  }
```
jjs451: Looks like the logic is good.  Cleaning up the brackets will make it easier on the eyes of the admin who review this :)

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
Bicepspump: I was afraid the program would count over 20. I tried removing that bit and it broke the program, showed 0%.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
Bicepspump: I'm sorry that it's so messy, I didn't know it would be included in the first pr I made.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
Pewtro: Why is addStack even it's own logic, you could just put it all in ``on_byPlayer_applybuffstack`` ? 
If it can't go above 20 however, it shouldn't trigger a ``on_byPlayer_applybuffstack`` event once it's at 20 stacks. You can somewhat see this logic in marksmanship legendary WarbeltOfTheSentinelArmy module for reference.

Pewtro:
Since you mentioned yourself, it's pretty messy, I'll wait a bit then toss a review - so you can have some time to clean up what you want to. I'll give it a look tonight with some stuff, if no one else gets around to it.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,86 @@
...
+        });
+  }
+  statistic() {
+    const castEfficiency = this.correctColdHeartCasts/this.totalColdHeartCasts;
+        return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.CHAINS_OF_ICE.id} />}
+        value={`${formatPercentage(castEfficiency)} %`}
+        label={'Cold Heart Efficiency'}
+        tooltip={`${formatPercentage(1-castEfficiency)}% of Cold Heart casts were made incorrectly. Either there wasn't 20 stacks of cold heart, or you didn't have the Unholy Strength buff up with more than 16 stacks of Cold Heart.`}
```
jjs451: I think the tooltip is unnecessary here.  The goal for all tooltips on this project is to have them contain new and useful information.  Here the tooltip is repeating what was covered in the suggestion.  You could instead use the tooltip to show where the castEfficiency came from.  Something like "{this.correctColdHeartCasts} out of {this.totalColdHeartCasts} were done correctly"

Bicepspump: Changed it to what you suggested

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
Bicepspump: It took the logic from another module. I could try to do it a different way but I don't really see a reason to do so if the current code works.

src/Parser/DeathKnight/Unholy/Modules/Features/ScourgeStrikeEfficiency.js
```diff
@@ -0,0 +1,89 @@
...
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FESTERING_WOUND.id){
+      this.targets[event.targetID] = event.stack;
+    }
+  }
+
+  on_byPlayer_removedebuffstack(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FESTERING_WOUND.id){
+      this.targets[event.targetID] = event.stack;
```
Pewtro: You should probably check for targetID **AND** targetInstance, since there can be multiple mobs with the same ID up at any given moment

src/Parser/DeathKnight/Unholy/Modules/Features/ScourgeStrikeEfficiency.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  on_byPlayer_removedebuff(event){
+    // keeps track of when the debuff drops off, useful for multitarget targets like DI
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FESTERING_WOUND.id){
+      this.targets[event.targetID] = 0;
+    }
```
Pewtro: Same thing with targetInstance here

src/Parser/DeathKnight/Unholy/Modules/Features/ScourgeStrikeEfficiency.js
```diff
@@ -0,0 +1,89 @@
...
+    const percentCastZeroWounds = this.scourgeStrikeCastsZeroWounds/this.totalScourgeStrikeCasts;
+    const strikeEfficiency = 1 - percentCastZeroWounds;
+    when(strikeEfficiency).isLessThan(0.60)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>You are casting <SpellLink id={SPELLS.SCOURGE_STRIKE.id}/> too often.  When spending runes remember to cast <SpellLink id={SPELLS.FESTERING_STRIKE.id}/> instead on targets with no stacks of <SpellLink id={SPELLS.FESTERING_WOUND.id}/></span>)
+            .icon(SPELLS.SCOURGE_STRIKE.icon)
+            .actual(`${formatPercentage(actual)}% of Scourge Strikes were used with Wounds on the target`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended`)
+            .regular(recommended - 0.30).major(recommended - 0.40);
+        });
```
Pewtro: This seems like a rather large jump from minor to average and then followed by a smaller jump from average to major.
This is your current scheme: >60% --> >30% --> >20%.
Wouldn't a 60 -> 40 -> 20 make more sense?

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,96 @@
...
+    const spellID = event.ability.guid;
+    if(spellID === SPELLS.COLD_HEART_BUFF.id){
+      this.buffColdHeart=event.stack;
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    this.unholyStrengthRemaining = 15000-(event.timestamp - this.unholyStrengthStart);
+    if (spellId === SPELLS.CHAINS_OF_ICE.id) {
```
Pewtro: no magic numbers, put duration as a constant at the top

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -0,0 +1,96 @@
...
+       this.totalColdHeartCasts++;
+    	    if  (this.unholyStrengthRemaining>0) {
+		         if (this.unholyStrengthRemaining <4000) {
+		            if (this.buffColdHeart < 20) {
+			            if (this.buffColdHeart > 16) {
+                    this.correctColdHeartCasts++;
```
Pewtro: again no magic numbers in terms of the stacks and duration remaining

src/Parser/DeathKnight/Unholy/Modules/Features/ScourgeStrikeEfficiency.js
Bicepspump: Can I simply just use targetInstance everywhere or do I have to use both?

Pewtro: You have to use both.
If you simply check for instance, it could trigger off the same instance of a unit, despite them having different IDs 
Boss1 spawns addType1 x2 (so one is instance 1 and the other is instance 2) 
Boss1 spawns addType2 x2 (so one is instance 1 and the other is instance 2) 

Now imagine you hit the instance 2 of the addType1 and apply festering wound there, the analyzer would allow you to hit addType2 instance 2 and apply calculations there, even tho it wasn't the same mob.

src/Parser/DeathKnight/Unholy/Modules/Features/ScourgeStrikeEfficiency.js
jjs451: Tacking onto that, I would raise the suggestion threshold to 80 or 85%.  Scourge Strike should almost never be cast without a Festering Wound on the target as bursting wounds provides significantly more damage than scourge strike itself.  There are occasional situations in burst AoE where you might spam SS on a target with no wounds but generally speaking we should see this number >90% when playing the rotation correctly.

Pewtro:
Thanks for the PR, looks good with the newest changes 😄

PR #993, opened 2017-12-15
[Marksmanship] rework t21 + descriptions of items and spells

Pewtro:
Reworks T214P  from: 

image

To: 
image

Also fixes an issue with t212p where focus gain was negative, and cleaned the code up in there.

Added descriptions to all tier, talents, traits and other items for Marksmanship.


Pewtro:
Feedback on the wording of the procs is welcome.. 

It can proc up to 3 times per cast, this was the best my tired brain could come up with.

PR #994, opened 2017-12-15
[Unholy Death Knight] Fix for crash when WCL is down

jjs451:
A line in the scourge of worlds module was causing a crash, decided to speed up plans to remove the module completely.  Feedback on the module is that it felt pretty useless to report something that was less than 1% of your damage


Pewtro:
Makes sense to me. 
By fire be purged!

PR #995, opened 2017-12-15
[core] Added better buff stack tracking to Entities.

kfinch:
Added better buff stack tracking to Entity / Entities, including a utility function that can be used to calculate 'average stacks' of a buff. Also added an ErraticMetronome module (a trinket from Nighthold), as a demonstration of how average stack calculation can be done.

This should help with the implementation of PR #963


MartijnHols:
Thanks!

PR #996, opened 2017-12-15
[Core] update Personnel Prototype Decimator to include procs and hits

Pewtro:
This isn't something easily trackable in Warcraft Logs, since it has no cast event, thought it'd be interesting information. 

image


kfinch:
Thanks!

PR #982, opened 2017-12-14
Fury warrior

Maldark:
Added initial support for Fury Warrior. Checks for Enrage and Frothing Berserker uptime as well as general ability usage. Also checks for correct potions and if the Juggernaut stacks dropped at some point.

PR #998, opened 2017-12-15
Fury warrior

Maldark:
Fixed all of the issues mentioned by @MartijnHols in #982 

- Changed class names to PascalCase
- Added getters for uptimes
- Added suggestion thresholds
- Replaced <span> with <Wrapper>
- Replaced some low recommendedEfficiencies withfixed minor importance.
- Fixed missing dots
- Fixed insufficient extraSuggestions
- Fixed Shockwave recommendedEffiency being unrealisticly high

### DIFF
--- a/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
+++ b/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
@@ -21,33 +21,39 @@ class EnrageUptime extends Analyzer {
 
   get suggestionThresholds() {
     return {
-      actual: this.enrageUptime,
       isLessThan: {
-        minor: 0.07,
-        average: 0.065,
-        major: 0.06,
+        minor: 0.7,
+        average: 0.65,
+        major: 0.6,
       },
       style: 'percentage',
     };
   }
 
   suggestions(when) {
-    when(this.suggestionThresholds.actual).isLessThan(this.suggestionThresholds.isLessThan.minor)
+    const {
+        isLessThan: {
+          minor,
+          average,
+          major,
+        },
+      } = this.suggestionThresholds
+
+    when(this.enrageUptime).isLessThan(minor)
       .addSuggestion((suggest, actual, recommended) => {
         return suggest(<Wrapper>Your <SpellLink id={SPELLS.ENRAGE.id} /> uptime can be improved.</Wrapper>)
           .icon(SPELLS.ENRAGE.icon)
           .actual(`${formatPercentage(actual)}% Enrage uptime`)
           .recommended(`>${formatPercentage(recommended)}% is recommended`)
-          .regular(this.suggestionThresholds.isLessThan.average).major(this.suggestionThresholds.isLessThan.major);
+          .regular(average).major(major);
       });
   }
 
   statistic() {
-    const enrageUptime = this.enrageUptime;
     return (
       <StatisticBox
         icon={<SpellIcon id={SPELLS.ENRAGE.id} />}
-        value={`${formatPercentage(enrageUptime)} %`}
+        value={`${formatPercentage(this.enrageUptime)} %`}
         label="Enrage uptime"
       />
     );
--- a/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
+++ b/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
@@ -9,7 +9,7 @@ import { formatPercentage } from 'common/format';
 import StatisticBox from 'Main/StatisticBox';
 import Combatants from 'Parser/Core/Modules/Combatants';
 
-class enrageUptime extends Analyzer {
+class EnrageUptime extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
@@ -39,4 +39,4 @@ class enrageUptime extends Analyzer {
   }
 }
 
-export default enrageUptime;
+export default EnrageUptime;
--- a/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
+++ b/src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
@@ -37,7 +37,7 @@ class EnrageUptime extends Analyzer {
           average,
           major,
         },
-      } = this.suggestionThresholds
+      } = this.suggestionThresholds;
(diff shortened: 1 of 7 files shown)


src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
```diff
@@ -8,27 +8,42 @@ import SpellLink from 'common/SpellLink';
...
+        average: 0.065,
+        major: 0.06,
+      },
+      style: 'percentage',
+    };
+  }
+
+  suggestions(when) {
+    when(this.suggestionThresholds.actual).isLessThan(this.suggestionThresholds.isLessThan.minor)
       .addSuggestion((suggest, actual, recommended) => {
```
Pewtro: just use ``this.enrageUptime`` and in the suggestions you can define the thresholds like so 
```javascript
    const {
      isGreaterThan: {
        minor,
        average,
        major,
      },
    } = this.suggestionThresholds;
```
and then simply refer to them as ``minor``, ``average`` and ``major``

src/Parser/Warrior/Fury/Modules/BuffDebuff/EnrageUptime.js
```diff
@@ -8,27 +8,42 @@ import SpellLink from 'common/SpellLink';
...
           .recommended(`>${formatPercentage(recommended)}% is recommended`)
-          .regular(recommended - 0.05).major(recommended - 0.10);
+          .regular(this.suggestionThresholds.isLessThan.average).major(this.suggestionThresholds.isLessThan.major);
       });
   }
 
   statistic() {
-    const enrageUptime = this.combatants.selected.getBuffUptime(SPELLS.ENRAGE.id) / this.owner.fightDuration;
+    const enrageUptime = this.enrageUptime;
     return (
```
Pewtro: no need to define it when you can just call it further down with ``this.enrageUptime``

Maldark:
The requested changes have been made.

src/Parser/Warrior/Fury/Modules/Features/Abilities.js
```diff
@@ -108,47 +116,53 @@ class Abilities extends CoreAbilities {
       noCanBeImproved: true,
       extraSuggestion: 'Only used in a combat with AoE or as a filler with the Wrecking Ball talent',
     },
+    {
+      spell: SPELLS.BERSERKER_RAGE,
+      category: Abilities.SPELL_CATEGORIES.UTILITY,
+      getCooldown: haste => 45,
+      recommendedEfficiency: 0.95,
```
kfinch: Is using Berserker Rage on cooldown this critical? Keep in mind that a recommendedEfficiency of 0.95 for a spell with 45 second CD means that you'd consider it an issue for the player to wait longer than 2 seconds to cast it after it becomes available. Basically you're asserting there's never a reason to hold this.

I don't know much about Fury Warriors, so this may be correct, just double checking.

src/Parser/Warrior/Fury/Modules/BuffDebuff/Juggernaut.js
```diff
@@ -27,12 +28,26 @@ class juggernautReset extends Analyzer {
...
     }
   }
+  
+  get suggestionThresholds() {
+    return {
+      isGreaterThan: {
+        minor: 0,
+      },
+      style: 'percentage',
+    };
```
Pewtro: Something I noticed when implementing my own checklist yesterday was that you'll need all 3 levels of severity of the issue for it to work properly and not simply be a white bar, so might want to look into that

src/Parser/Warrior/Fury/Modules/Features/Abilities.js
Maldark: You are probably right that it is unrealisticly harsh. I'll tune it down a bit, but I don't see it being of interest any time soon, as there are currently 0 logged kills on warcraftlogs where anyone has used this talent.

src/Parser/Warrior/Fury/Modules/BuffDebuff/Juggernaut.js
Maldark: I have yet to work on the checklist, so I wouldn't know. Can the severity levels have the same value? In general Juggernaut stacks should never drop - but seeing as there may be adds it could be unavoidable. Having Juggernaut drop on a Patchwerk boss, however, is a major issue.

Pewtro: Well all of these things ``suggestionThresholds`` you're making are strictly speaking only a thing because of the checklist as that's where they're used. 
I never checked if they were the same, right now it's PERMANENTLY a minor issue, it'll never go to average or major. 
What you can do is minor: 0.5
average: 0.7
major: 0.9
Then if it becomes 1, it's a major issue
Or maybe like so:
minor: 0.5 (or 0, whatever)
average: 0.9
major: 1.1
In which case if it drops twice, it's a major issue, but otherwise average if it drops (then in the suggestion include a note on AoE about it)

src/Parser/Warrior/Fury/Modules/Features/Abilities.js
Pewtro: Doesn't matter if it's used a lot or not, if you make it, you should make it properly. 
What I sometimes use for unconventional abilities like this one, is deduct a bit of usage from a simc baseline profile with it, and work from there.

Maldark: That's a good way of determining realistic use - I'll do that. Thanks

src/Parser/Warrior/Fury/Modules/BuffDebuff/Juggernaut.js
Maldark: I made this kind of suggestionThreshold due to @MartijnHols's suggestion on my previous PR #982. I do, however, agree with your interpretation that it would permanently be a minor issue, but my testing showed it as a major issue regardless of it only having a minor threshold. I'm guessing it's because there is only one value, and seeing as we're below that (and there is no additional average or major thresholds) it defaults to major. I'll get it fixed for my next commits.

Maldark:
Fixed the Berserker Rage concerns @kfinch pointed out and added all 3 levels of severity to the Juggernaut reset suggestionThreshold as suggested by @Pewtro

PR #999, opened 2017-12-16
[Hunter] Marksmanship checklist

Pewtro:
As I said on discord, I'm not really happy with how the stacking is looking currently with a lot of screen space being wasted. 

For reference: 
https://cdn.discordapp.com/attachments/316864385052049420/391381619900022784/unknown.png
https://cdn.discordapp.com/attachments/316864385052049420/391381738774986763/unknown.png

My proposed fix would be to somehow make the statistics boxes auto allocate under the checklist, and to also fill out all the way to the fight, and that way move everything quite far up, so we don't end up with having 2+ pages worth of scrolling to get information. 

But here is what I've been working on:
image

image

image


MartijnHols:
Looking through it now that it's merged, a few comments:

* How about changing `Average Aimed Shot Aimed Shot casts/Trueshot` to `Average Aimed Shot Aimed Shot casts per Trueshot`?
* The focus icon seems out of place since focus is not capitalized. How about capitalizing it and linking to https://www.wowhead.com/focus ?
* How about renaming `Did not save A Murder of Crows for execute`  to `Missed A Murder of Crows casts during execute`?
* What does `Execute Trueshot Trueshot casts` mean?
* How about renaming `Bullseye resets` to `Bullseye buff lost`? It's probably fine for most Hunters to understand, but it's a little hard to grasp without good spec knowledge.
* Downtime rule is missing spell icons (they're so pretty)
* comma in downtime rule is obsolete
* What is `casting downtime`?
* I'm not entirely 100% on `Cancelled casts`; is this channeled casts canceled out of total channeled casts?

Pewtro:
* Agreed - done.
* Capitalized and added a link.
* It's a check if they cast trueshot when boss was between 25 and 20% health (we enter execute at 20%), so generally you save AMoC for the 20% mark if you get it back once boss is ~25% hp to build our Bullseye stacks fast. I'll think of a better wording. UPDATE: I've reworded it to ``A Murder of Crows casts right before execute``
* Updated to be ``Trueshot casts with Bullseye up``
* Updated to be ``Bullseye lost, while boss was under 20%`` 
* Added spell icons
* Updated
* Changed to just be ``Downtime`` 
* Changed to be ``Channeled casts cancelled``

PR #1000, opened 2017-12-16
Fixing bug where GotQ/Cloudburst percent is show incorrectly

hartra344:
Found and fixed this in the Checklist PR but since this is a live bug now I decided to push it as a separate PR


Pewtro:
PR number 1000 😅

PR #1001, opened 2017-12-16
Add Clawing Shadows alias for Scourge Strike

layday:
Clawing Shadows is a third-tier talent that replaces
Scourge Strike.


Pewtro:
We'll get this in immediately once travis is done

PR #1002, opened 2017-12-16
[Core] Fix for SpellUsable Reduction of Charge Spells

Sharrq:
I noticed that the console log was frequently telling me that Phoenix's Flames was used while it was on cooldown, so i went to look at SpellUsable to figure out why.

The way Phoenix's Flames works is it has 3 charges, 45 second cooldown each. And ignite damage has a chance to erupt and reduce the cooldown of Phoenix's Flames by 10 seconds.

I believe the issue is that if multiple charges of Phoenix's Flames are on cooldown, and there is less than 10 seconds left on the cooldown of a charge, then SpellUsable ends the cooldown and grants a charge, and then stops.

But in game, if you have less than 10 seconds left on a Phoenix Flames charge, then it ends the cooldown on that charge, and the remainder of that 10 seconds reduces the next charge. So for example. If i have 3 charges on cooldown and 5 seconds until i gain a charge, and ignite procs. Then it is supposed to reduce the one charge by 5 seconds, gain a charge, and then reduce the cooldown on the next charge by the remaining 5 seconds.

I believe what i have done here fixes it (and it fixed my issue with Phoenix Flames .. or seems to have), but im not sure if this is how ALL charge spells work, nor am i sure if there is a better way to accomplish this.

### DIFF
--- a/src/Parser/Core/Modules/SpellUsable.js
+++ b/src/Parser/Core/Modules/SpellUsable.js
@@ -101,7 +101,7 @@ class SpellUsable extends Analyzer {
    * @param {boolean} resetAllCharges Whether all charges should be reset or just the last stack. Does nothing for spells with just 1 stack.
    * @param {number} timestamp Override the timestamp if it may be different from the current timestamp.
    */
-  endCooldown(spellId, resetAllCharges = false, timestamp = this.owner.currentTimestamp) {
+  endCooldown(spellId, resetAllCharges = false, timestamp = this.owner.currentTimestamp, remainingCDR = 0) {
     if (!this.isOnCooldown(spellId)) {
       throw new Error(`Tried to end the cooldown of ${spellId}, but it's not on cooldown.`);
     }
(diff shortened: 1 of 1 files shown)


src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -101,7 +101,7 @@ class SpellUsable extends Analyzer {
    * @param {boolean} resetAllCharges Whether all charges should be reset or just the last stack. Does nothing for spells with just 1 stack.
    * @param {number} timestamp Override the timestamp if it may be different from the current timestamp.
    */
-  endCooldown(spellId, resetAllCharges = false, timestamp = this.owner.currentTimestamp) {
+  endCooldown(spellId, resetAllCharges = false, timestamp = this.owner.currentTimestamp, remainingCDR = 0) {
```
kfinch: Please also add remainingCDR to the documentation comment above. For a core module like SpellUsable I think it's important we keep the documentation up to date.

src/Parser/Core/Modules/SpellUsable.js
```diff
@@ -118,6 +118,7 @@ class SpellUsable extends Analyzer {
       cooldown.chargesOnCooldown -= 1;
       this._triggerEvent('updatespellusable', this._makeEvent(spellId, timestamp, 'restorecharge', cooldown));
       this.refreshCooldown(spellId, timestamp);
+      this.reduceCooldown(spellId,remainingCDR,timestamp);
```
kfinch: Please add an if statement so `reduceCooldown` is only called here when `remainingCDR` is not 0. Calling `reduceCooldown` with reduction equal to 0 after almost every `endCooldown` is superfluous, and it might also cause confusing debug logging.

Sharrq:
added those changes

PR #1003, opened 2017-12-17
[Hunter] Checklist update for Marksmanship

Pewtro:
Adjusts things as suggest by ZeroToRescue
Updates positioning of most modules
Fixes a small bug I hadn't noticed when changing things originally for this module in CancelledCasts
Ensures that Bullseye shows up at the start of the pull, which makes it so that the modules don't jump around while loading in.
Adds a background to focus capped time to better indicate it being bad: 
image


MartijnHols:
Could you also add a changelog entry for your checklist please?

Pewtro:
It already has this?
image

MartijnHols:
lol I was looking at the wrong spec sorry my bad

PR #1013, opened 2017-12-19
Merge magic schools in Damage Taken

### LINKED ISSUE #384: Merge magic schools in Damage Taken statistic
It's rarely useful to know the specific magic school of an incoming damage source, and potentially confusing to a user.  

Conversely, it's extremely useful to know whether or not an incoming damage source is magical, or physical (some damage reductions like Armor only work on physical, others like Spell Reflect only work on magic).

I propose that the graphic for the type breakdown in the Damage Taken statistic box be consolidated into only Physical vs Magic, to highlight the distinction between the two.  This would enable users to identify at a quick glance whether they took mostly physical, mostly magic, or a combination of both, in a given fight.

The current school-by-school breakdown can be moved into the statistic tooltip, or toggleable by click.

[Bountysource](https://www.bountysource.com/issues/49713481-merge-magic-schools-in-damage-taken-statistic?utm_source=49713481&utm_medium=shield&utm_campaign=ISSUE_BADGE)

Zeboot:
Aims to resolve #384  by merging the different magic schools of incoming damage into a general "magic damage" bar segment on the damage taken graph. The detailed school-by-school split can still be viewed on the tooltip and the bar can be toggled between the two modes by clicking on it.


src/Parser/Core/Modules/DamageTaken.js
```diff
@@ -65,9 +65,44 @@ class DamageTaken extends Analyzer {
...
+    </ul>
+    Click the bar to switch between simple and detailed mode.`;
+  }
+
+  handleClick(){
+    const barSimple = document.getElementById("damage-taken-bar-simple");
+    const barDetailed = document.getElementById("damage-taken-bar-detailed");
+    if((!barSimple || !barDetailed)) return; //return if one of the two bars isn't found.
+    const detailed = barDetailed.style.display.toLowerCase() !== "none";
+    if(detailed){
```
MartijnHols: Toggling DOM elements with React is usually done differently. With React the regular way to do this is to store the state of the element and only render the elements you want. If you're interested you can find a full tutorial here https://reactjs.org/docs/hello-world.html, I personally couldn't recommend React as a front-end library more.

It's important to note that our application is structured in a way that does not fully use the React way and mixes React approaches with plain ES6 classes. This causes a lot of the best practices to work slightly different and often require much poorer solutions, so while I will recommend a solution close to *the React way*, it's not entirely how this is normally done with React.

You really are in no way required to learn these things for future PRs, I'm just saying all this to try and explain my reasoning behind my suggestion below and maybe poke your interest.

As I said earlier *the React way* would be to only render the elements applicable. Assuming you have a property named `detailView` in the class that should show the detail view if `true` and the simple view if `false`, I'd suggest solving this with the following:

```jsx
handleClick(){
  this.detailView = !this.detailView;
}
```

```jsx
        footer={(
          <Wrapper>
            {this.detailView ? (
              <div onClick={() => this.handleClick()} className="statistic-bar">
                {Object.keys(this._byMagicSchool)
                  .map(type => {
                    const effective = this._byMagicSchool[type].effective;
                    if(effective === 0) return ``;
                    return (
                      <div
                        key={type}
                        className={`spell-school-${type}-bg`}
                        style={{ width: `${effective / this.total.effective * 100}%` }}
                        data-tip={tooltip}
                      />
                    );
                  })}
              </div>
            ) : (
              <div onClick={() => this.handleClick()} className="statistic-bar">
                {(() => {
                  const physical = (this._byMagicSchool[MAGIC_SCHOOLS.ids.PHYSICAL])?this._byMagicSchool[MAGIC_SCHOOLS.ids.PHYSICAL].effective : 0;
                  const magical = this.total.effective - physical;
                  const simplifiedValues = {};
                  simplifiedValues[MAGIC_SCHOOLS.ids.PHYSICAL] = physical;
                  simplifiedValues[MAGIC_SCHOOLS.ids.SHADOW] = magical; // use shadow as placeholder for general magic
                  return Object.keys(simplifiedValues).map(type => {
                    const effective = simplifiedValues[type];
                    return (
                      <div
                        key={type}
                        className={`spell-school-${type}-bg`}
                        style={{ width: `${effective / this.total.effective * 100}%` }}
                        data-tip={tooltip}
                      />
                    );
                  });
                })()
                }
              </div>
            )}
          </Wrapper>
        )}
```

Do note I didn't apply the other suggestions about the rendering of these bars.

src/Parser/Core/Modules/DamageTaken.js
```diff
@@ -82,29 +117,44 @@ class DamageTaken extends Analyzer {
...
-                    }
-                  />
-                );
-              })}
+        footer={
+            [(<div id="damage-taken-bar-detailed" onClick={() => this.handleClick()} className="statistic-bar" style={{display: "none"}}>
+                {Object.keys(this._byMagicSchool)
+                .map(type => {
+                  const effective = this._byMagicSchool[type].effective;
+                  if(effective === 0) return ``;
```
MartijnHols: I think it would be cleaner if this were a `.filter(..)` before the map. Example:

```jsx
Object.keys(this._byMagicSchool)
  .filter(type => this._byMagicSchool[type].effective !== 0)
  .map(type => {
    // etc
```

Zeboot:
Thanks for the feedback! I'll attempt to implement these changes later tonight (I initially attempted to have conditional rendering such as this just couldn't figure out how exactly to do it, your feedback will definitely help there)!

Zeboot:
I believe the feedback should be addressed, though I'm not sure whether the way I implemented the toggle is the cleanest one (might be cleaner in a different class than the one I used)

PR #1014, opened 2017-12-19
Added a module that tracks the average amount of wounds Apocalypse po…

Bicepspump:
New module for Apocalypse usage.


src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
...
+	}
+  }
+
+  //Logic that both counts the amount of Apocalypse cast by the player, as well as the amount of wounds popped by those apocalypse.
+  on_byPlayer_cast(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.APOCALYPSE.id){
+		this.totalApocalypseCasts+=1;
+		if(this.targets.hasOwnProperty(encodeTargetString(event.targetID, event.targetInstance))) {
+			const currentTargetWounds = this.targets[encodeTargetString(event.targetID, event.targetInstance)];
```
kfinch: While I believe your method of buff tracking does work, it is already implemented between the Entity and EnemyInstances classes, making your logic in applydebuffstack, removedebuffstack, and removedebuff methods unneeded. No need to 'reinvent the wheel'. I'm sorry I didn't warn you about this sooner in Discord, I think I misunderstood what you were trying to do.

Instead of using the above lines to get buff stacks, you can do something like:
```
const target = enemies.getEntity(event); // can be null for a "not noteworthy" target
const currentTargetWounds = target && target.hasBuff(SPELLS.FESTERING_WOUND.id)
    ? target.getBuff(SPELLS.FESTERING_WOUND.id).stacks
    : 0;
```

src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
...
+			if(currentTargetWounds > 5){
+				this.apocalypseWoundsPopped=this.apocalypseWoundsPopped + 6;
+				console.log(this.apocalypseWoundsPopped);
```
kfinch: Log lines for normal behavior should not make it to production code. Easiest is to remove this line.

If you want to leave in these debug prints for later development, you can hide it behind a debug flag like `debug && console.log(this.apocalypseWoundsPopped);` along with `const debug = false` at the top of the class. Then when you're on dev environment and you need to debug, you simply change the flag to true in order to turn on logging for this class.

src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
...
+			} else {
+				this.apocalypseWoundsPopped=this.apocalypseWoundsPopped + currentTargetWounds;
+			}
+		}
+	}
+  }
+
+  suggestions(when) {
+	//Takes the average amount of casts and rounds it to one decimal digit
+    const averageWoundsPopped = parseFloat(Math.round(this.apocalypseWoundsPopped/this.totalApocalypseCasts*10)/10).toFixed(1);
```
kfinch: what's all the parsing and rounding and multiplying for? Isn't this identical to `(this.apocalypseWoundsPopped/this.totalApocalypseCasts).toFixed(1)` ?

src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Enemies from 'Parser/Core/Modules/Enemies';
```
kfinch: Currently you're not using this import, but there's a change I suggest below where you do. If you do, you want to be using `'Parser/Core/Modules/EnemyInstances'` instead of `'Parser/Core/Modules/Enemies'`. Those two modules are identical, except `EnemyInstances` is capable of discerning between different mobs with the same name. (Really, `Enemies` shouldn't exist at all, I plan at some point to remove it and leave only `EnemyInstances`)

src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
...
+				console.log(this.apocalypseWoundsPopped);
+			} else {
+				this.apocalypseWoundsPopped=this.apocalypseWoundsPopped + currentTargetWounds;
+			}
+		}
+	}
+  }
+
+  suggestions(when) {
+	//Takes the average amount of casts and rounds it to one decimal digit
```
kfinch: This comment is just a restatement of what the line of code below it is doing, and so is not needed.

src/Parser/Deathknight/Unholy/Modules/Features/Apocalypse.js
```diff
@@ -0,0 +1,92 @@
...
+          return suggest(<span>You are casting <SpellLink id={SPELLS.APOCALYPSE.id}/> with too few <SpellLink id={SPELLS.FESTERING_WOUND.id}/> on the target. When casting <SpellLink id={SPELLS.APOCALYPSE.id}/>, make sure to have at least 6 <SpellLink id={SPELLS.FESTERING_WOUND.id}/> on the target.</span>)
+            .icon(SPELLS.APOCALYPSE.icon)
+            .actual(`An average ${(actual)}% of Festering Wounds were popped by Apocalypse`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended`)
+            .regular(recommended - 1).major(recommended - 2);
+        });
+  }
+
+  statistic() {
+    const averageWoundsPopped = parseFloat(Math.round(this.apocalypseWoundsPopped/this.totalApocalypseCasts*10)/10).toFixed(1);
```
kfinch: what's all the parsing and rounding and multiplying for? Isn't this identical to `(this.apocalypseWoundsPopped/this.totalApocalypseCasts).toFixed(1)` ?

Bicepspump:
Didn't mean to dismiss the review. Not sure why it fails either 🤔

MartijnHols:
The tests are failing because something is crashing the rendering of the Root. Run tests locally with `npm test` to reproduce the issue. Make sure the home page is rendering without new errors.

Before doing any of that try to merge from master.

kfinch:
@Bicepspump Updating code in an approved PR automatically dismisses the approval, wasn't your fault that happened. Once you get the tests fixed this is good to merge.

kfinch:
@MartijnHols according to Travis, the live version of WoWAnalyzer has failing tests... this may not be his fault.

MartijnHols:
`master`  doesn't seem to have an issue in Root, I hope merging with master will resolve the failed Travis checks.

Bicepspump:
I think there was a weird issue here. The Apocalypse.js had been created in a folder which didnt have a capital K in DeathKnight. I think this is what caused the previous issue.

MartijnHols:
Thanks! 👍

Issue #954, opened 2017-12-08
Show artifact relics equipped


MartijnHols:

Pewtro:
Do you have a battleplan how this would go about being added? I have some sparetime sunday to sink into something, this could be one of them

MartijnHols:
Well still thinking about if we should do this at all. Artifact relics might be interesting because they reflect more than just stats, the only other things that do this are legendaries, trinkets and set bonuses. Still it seems weird to do this without simply showing the full gear set.

I was thinking it could go underneath the talentbox, but I realize in the results page redesign it wouldn't really fit. So a tab with the entire equipped gear visible would make more sense.

PR #1017, opened 2017-12-20
Added gear tab

Nekorsis:
Nothing fancy, just thought it would be nice thing to have. I think it's a part of #954 issue, all that it's left is to find out how to get relics data.


MartijnHols:
Thanks for the PR! Could you please include a screenshot of what it looks like?

src/Main/Gear.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    return(
+      <div>
+        <div style={{marginLeft: 'auto', marginRight: 'auto', display: 'block', width: '90%'}}>
+          {currentPlayer.gear.map(item => {
+            return (
+              <div style={{display: 'inline-block', textAlign: 'center'}}>
+                {item.itemLevel}
+                <a key={item.id} href={`${toolTipBaseUrl}${item.id}`} style={{ margin: '5px', display: 'block' }}>
+                  <img  alt='' target='_blank' rel='noopener noreferrer' src={`${iconBaseUrl}${item.icon}`}/>
```
MartijnHols: Please use the Icon component in `common/Icon` for rendering icons.

src/Main/Gear.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+    return(
+      <div>
+        <div style={{marginLeft: 'auto', marginRight: 'auto', display: 'block', width: '90%'}}>
+          {currentPlayer.gear.map(item => {
+            return (
+              <div style={{display: 'inline-block', textAlign: 'center'}}>
+                {item.itemLevel}
+                <a key={item.id} href={`${toolTipBaseUrl}${item.id}`} style={{ margin: '5px', display: 'block' }}>
```
MartijnHols: `key` should be on the first container element or it will have no effect

src/Main/Gear.js
```diff
@@ -0,0 +1,45 @@
...
+    playerName: PropTypes.string.isRequired,
+  };
+  render() {
+    const playersList = this.props.combatants.players;
+    const iconBaseUrl = 'http://render-us.worldofwarcraft.com/icons/56/';
+    const toolTipBaseUrl='http://www.wowhead.com/item=';
+    let currentPlayer;
+    for (const prop in playersList) {
+      if (playersList[prop]._combatantInfo.name === this.props.playerName) {
+        currentPlayer=  playersList[prop]._combatantInfo;
```
MartijnHols: You can get the current player with `this.props.combatants.selected`. Also I recommend changing the prop from taking the `combatants` prop to just `combatant` or `selectedCombatant` or something similar. That way you don't need the playerName.

Nekorsis: Done, Gear now takes selectedCombatant  as the only prop.

src/Main/Gear.js
Nekorsis: My bad, already fixed.

src/Main/Gear.js
Nekorsis: Done. Should i also consider createng ItemIcon and ItemLink components, just like with SpellLink, SpellIcon ?

Nekorsis:
And this how it looks like: 
https://i.imgur.com/QV4GQw9.png
https://i.imgur.com/0xC4Z4t.png
Probable should use table of some sort, maybe bootstrap ?
Also it's seems like logs themselves don't have any relics info or am i missing something ?

MartijnHols:
I reckon the "gems" of artifact weapons are the relics:

image

What do you think about including a border to indicate item rarity (epic/legendary/etc)? And how about hiding the cloak when not equipped?

MartijnHols:
If you push your changes to `Nekorsis:master` this PR will be updated to include those changes. It might be easier to create a new branch for this feature and work from that instead, if you choose to do that you could also edit your PR to track the new branch.

Nekorsis:
So yeah, i added relics.
About borders - relics don't have a quality property and because of it i turned off borders for now until i figure out how to get relics quality.
Also i don't really like the double map in render, should i somehow refactor this ?

src/Main/Gear.js
```diff
@@ -0,0 +1,79 @@
...
+          return '#0070dd';
+        case 2:
+          return '#1eff00';
+        case 1:
+          return '#ffffff';
+        default:
+          break;
+      }
+    };
+    const artifactData = data.filter(item => item.quality === 6 && item.gems); // get artifact from list of items to render relics
```
MartijnHols: This kinda qualifies as a hack. You can achieve the same while using the array methods normally;
```jsx
const artifact = data.find(item => item.quality === 6);
const artifactData = artifact ? artifact.gems : null;
```
I think it's safe to assume a player only has *one* artifact.

Also even more nitpicky: the var name `artifactData` is a bit vague. How about `relics`?

src/Main/Gear.js
```diff
@@ -0,0 +1,79 @@
...
+    return(
+      <div>
+        <div style={{marginLeft: 'auto', marginRight: 'auto', display: 'block', width: '90%'}}>
+          {data.map(item => {
+            if (item.id === 0) return null;
```
MartijnHols: When working with an array it would be slightly cleaner to use the `filter` method to filter results. Example:

```jsx
{data
  .filter(item => item.id !== 0)
  .map(item => (
    <div>...rest
```

src/Main/Gear.js
```diff
@@ -0,0 +1,79 @@
...
+                    :
+                    null
+                  }
+                >
+                  <Icon icon={item.icon} />
+                </a>
+              </div>
+            );
+          })}
+          {artifactData[0].gems.map(item => {
```
MartijnHols: This might crash without an artifact equipped. I realize we'll have to fix a lot more things to support next expansion, but the more things we can account for in the first design the better.

src/Main/Gear.js
```diff
@@ -0,0 +1,79 @@
...
+import PropTypes from 'prop-types';
+import Icon from 'common/Icon';
+
+class Gear extends React.PureComponent {
+  static propTypes = {
+    selectedCombatant: PropTypes.object.isRequired,
+  };
+  render() {
+    const toolTipBaseUrl='http://www.wowhead.com/item=';
+    const data = Object.values(this.props.selectedCombatant._gearItemsBySlotId);
```
MartijnHols: `data` is a bit of a vague name, how about `items` or `gear`?

Nekorsis:
Fixed variable names and render.
Sorry about the names, sometimes i forget to do a proper naming.

MartijnHols:
Thanks! Just a few more things; please update from master to fix the conflict, and ensure you're not running into any errors in the console. Travis seems to be having issues running here, hoping it's resolved when you update from master but it might be you have to recreate the PR to have Travis pass.

PR #1018, opened 2017-12-20
Runic Power Pooling for DA

Bicepspump:
Added a module that tracks how much runic power the player pools before using DA.


src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+		this.totalDarkArbiterCasts++;
+		this.totalRunicPowerPooled += this.currentRP;
+		}
+	}
+
+  suggestions(when) {
+    const averageRpPooled = (this.totalRunicPowerPooled/this.totalDarkArbiterCasts).toFixed(1);
+    when(averageRpPooled).isLessThan(this.maxRP-20)
+ 		  .addSuggestion((suggest, actual, recommended) => {
+			return suggest(<span>You are casting <SpellLink id={SPELLS.DARK_ARBITER_TALENT.id}/> without enough runic power. Make sure to pool some runic power before you cast <SpellLink id={SPELLS.DARK_ARBITER_TALENT.id}/>.</span>)
```
MartijnHols: When you need a container for multiple elements without the desire to style it, please use the Wrapper component from `common/Wrapper` instead of `span`. Usage of `span` for this is deprecated (so you might find this in older code).

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+
+class RpPoolingDA extends Analyzer {
```
MartijnHols: Am I right in concluding that this is an analyzer only for Dark Arbiter and there are no other modules for the same talent? If so how about naming it DarkArbiter instead, I think that would make it easier to identify later.

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+  totalDarkArbiterCasts = 0;
+  currentRP = 0;
+  maxRP = 0;
+  
+   on_toPlayer_energize(event){
+	   this.currentRP = event.classResources[0]['amount']/10 || 0;
+	   this.maxRP = event.classResources[0]['max']/10;
+   }
+   
+  on_byPlayer_cast(event){
```
MartijnHols: I think casts also give you an update of class resources, you might want to take those values into account for a higher accuracy.

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+    this.active = this.combatants.selected.hasTalent(SPELLS.DARK_ARBITER_TALENT.id);
+  }
+
+  totalRunicPowerPooled = 0;
+  totalDarkArbiterCasts = 0;
+  currentRP = 0;
+  maxRP = 0;
+  
+   on_toPlayer_energize(event){
+	   this.currentRP = event.classResources[0]['amount']/10 || 0;
```
MartijnHols: Since a player can have multiple resources active at the same time, RP might not always be at the same index. I recommend doing a `event.classResources.find` for the resource type instead. See the `RageWasted` module for an example.

MartijnHols: The order of this expression can be error sensitive; `undefined / 10` will result in NaN which is one of the most unpredictable types in JS there is so should be avoided at all times. It would be safer to write this as `(event.classResources[0]['amount'] || 0) / 10`.

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+  on_byPlayer_cast(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.DARK_ARBITER_TALENT.id){
+		this.totalDarkArbiterCasts++;
+		this.totalRunicPowerPooled += this.currentRP;
+		}
+	}
+
+  suggestions(when) {
+    const averageRpPooled = (this.totalRunicPowerPooled/this.totalDarkArbiterCasts).toFixed(1);
```
MartijnHols: I recommend making this a getter in the class. This separates the code more and makes it reusable.

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+    when(averageRpPooled).isLessThan(this.maxRP-20)
```
MartijnHols: A totally nit picky comment: in the new results rework where we're introducing the Checklist it's recommended to write suggestion thresholds differently. It might benefit to start doing this already so that you don't have to rewrite as much later when you want to implement the checklist. Here's an example:

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/46d812bdda19ad4eb06e7d372913835e926b6e1d/src/Parser/Paladin/Holy/Modules/Talents/AuraOfSacrifice.js#L35-L57

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
```diff
@@ -0,0 +1,65 @@
...
+			return suggest(<span>You are casting <SpellLink id={SPELLS.DARK_ARBITER_TALENT.id}/> without enough runic power. Make sure to pool some runic power before you cast <SpellLink id={SPELLS.DARK_ARBITER_TALENT.id}/>.</span>)
+				.icon(SPELLS.DARK_ARBITER_TALENT.icon)
+				.actual(`${averageRpPooled} of runic power were pooled on average`)
+				.recommended(`>${(recommended)} is recommended`)
+				.regular(recommended - 20).major(recommended - 40);
+        });
+  }
+
+  statistic() {
+    const averageRpPooled = (this.totalRunicPowerPooled/this.totalDarkArbiterCasts).toFixed(0);
```
MartijnHols: This could re-use the getter that you'd define for the suggestion: less duplicate code :D

src/Parser/DeathKnight/Unholy/Modules/Features/RpPoolingDA.js
jjs451: There's already another module named DarkArbiter.  It tracks how much RP is dumped while DA is ticking.  That one should probably get renamed at some point to make it easy to see at a glance what's different about it

MartijnHols:
Thanks!

PR #1025, opened 2017-12-21
[Fire Mage] Remove Direct Damage Crits Suggestion

Sharrq:
I believe my analysis is wrong for the direct damage crits which is causing people to get higher numbers than they should have. Removing the suggestion until I can take a closer look at this.


Sharrq:
Also, travis is probably going to fail, my console is getting a error in the Blood DK Unending Thirst module

image

PR #1028, opened 2017-12-21
[Discipline Priest] Bug Fix for the usage of Light Spawn Glyph

karlpralow:


src/Parser/Priest/Discipline/Modules/Core/ShadowfiendAbilityTracker.js
```diff
@@ -0,0 +1,18 @@
...
+
+class ShadowfiendAbilityTracker extends CoreAbilityTracker {
+
+  on_byPlayer_cast(event) {
+    if (super.on_byPlayer_cast) {
+      super.on_byPlayer_cast(event);
+    }
+    const spellId = event.ability.guid;
+    if (spellId === SPELLS.SHADOWFIEND_WITH_GLYPH_OF_THE_SHA.id || spellId === SPELLS.LIGHTSPAWN.id) {
+      event.ability.guid = SPELLS.SHADOWFIEND.id;
```
MartijnHols: How about making a Normalizer for this instead? I wouldn't expect this kind of behavior in an Analyzer and I think it will be hard to find for others, and it won't be shown in the events tab.

PR #1031, opened 2017-12-22
[Unholy Death Knight] Checklist support

jjs451:
A solid first pass at a checklist for unholy.  Feels a little sparse right now, waiting for feedback from some theroycrafters on what they want to see.  Core spells feels especially spare because core spells are tied to runes and I haven't figured out how to do rune tracking yet.


src/Parser/DeathKnight/Unholy/CHANGELOG.js
```diff
@@ -1,4 +1,9 @@
 export default  [
+  {
+    date: new Date('12-21-2017'),
```
MartijnHols: This date format is not supported, please provide dates as y-m-d.

src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueUptime.js
```diff
@@ -14,24 +14,38 @@ class VirulentPlagueUptime extends Analyzer {
     enemies: Enemies,
   };
 
+  get vpUptime() {
```
MartijnHols: Since your class is already called `VirulentPlague`, it can be omitted from all properties.

PR #1043, opened 2017-12-23
Implemented Legendary Boots for Elemental Shaman

HawkCorrigan:
Not too sure about the equip and the class Resource access.
The whole Maelstrom stuff only uses index 0 so i guess that it will always have that index.

Deciding if it's active could also be done via item check, but that seems a little harder than just checking the buff

### DIFF
--- a/src/Parser/Shaman/Elemental/Modules/Items/TheDeceiversBloodPact.js
+++ b/src/Parser/Shaman/Elemental/Modules/Items/TheDeceiversBloodPact.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import ItemIcon from 'common/ItemIcon';
+import ItemLink from 'common/ItemLink';
+import { formatNumber } from 'common/format';
+
+
+class TheDeceiversBloodPact extends Analyzer {
+  static dependencies = {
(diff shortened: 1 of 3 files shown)


src/Parser/Shaman/Elemental/Modules/Items/TheDeceiversBloodPact.js
```diff
@@ -0,0 +1,37 @@
...
+
+class TheDeceiversBloodPact extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  extraMaelstrom = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasBuff(SPELLS.THE_DECEIVERS_BLOOD_PACT_EQUIP.id);
```
anom0ly: You could use the `.hasFeet` function to also check the item ID, and enable based on that.

src/Parser/Shaman/Elemental/Modules/Items/TheDeceiversBloodPact.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  on_byPlayer_energize(event) {
+    if (event.ability.guid===SPELLS.THE_DECEIVERS_BLOOD_PACT_BUFF.id) {
+        this.extraMaelstrom+=event.classResources[0].amount;
```
anom0ly: For Shaman I believe this is fine to code a index `0` just for other classes, such as druids, their main resource potentially changes.

Pewtro:
Getting this in, possible edits to be made in the feature however with the above comments 😄

PR #1050, opened 2017-12-24
Fix #1044: Added a *casting time* bar to the spell timeline that shows you the amount of time spent channeling or waiting for the GCD

### LINKED ISSUE #1044: Add a "downtime" view to the timeline
e.g. show every GCD and point out when the user isn't doing anything.

https://www.reddit.com/r/wow/comments/7floz2/warcraftlogs_experts_any_way_to_see_your_dps/

MartijnHols:
image

Might have to tweak the colors a bit. I thought about what color I thought the GCD would be and figured brown seemed right, but in hindsight it doesn't look that well.

Fixes #1044

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2017-12-25'),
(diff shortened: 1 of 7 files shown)


MartijnHols:
image

Better colors. I wanted to save purple for channeling bars should we ever want to add those. Blue is close so makes sense still I think.

MartijnHols:
Ohh yeah I forgot to mention in #1056 that it depends on this.

PR #1053, opened 2017-12-25
[Core] Remove Downtime Suggestion for Eonar

Sharrq:
This uses the new boss fight settings to disable the Downtime StatisticBox/Suggestion for Eonar since the stat is pretty useless on that fight.

This is just the first of some other boss specific settings that I think will help. Part of Issue #391

### DIFF
--- a/src/Parser/Core/Modules/AlwaysBeCasting.js
+++ b/src/Parser/Core/Modules/AlwaysBeCasting.js
@@ -172,7 +172,8 @@ class AlwaysBeCasting extends Analyzer {
     downtime: '/img/afk.png',
   };
   statistic() {
-    if (!this.showStatistic || this.owner.boss.fight.noDowntimeSuggestion === true) {
+    const boss = this.owner.boss;
+    if (!this.showStatistic || (boss && boss.fight.disableDowntimeStatistic)) {
       return null;
     }
 
--- a/src/Parser/Core/Modules/AlwaysBeCasting.js
+++ b/src/Parser/Core/Modules/AlwaysBeCasting.js
@@ -172,40 +172,41 @@ class AlwaysBeCasting extends Analyzer {
     downtime: '/img/afk.png',
   };
   statistic() {
-    if (!this.showStatistic) {
+    if (!this.showStatistic || this.owner.boss.fight.noDowntimeSuggestion === true) {
       return null;
     }
 
     return (
-      <StatisticBox
-        icon={<Icon icon="spell_mage_altertime" alt="Downtime" />}
-        value={`${formatPercentage(this.downtimePercentage)} %`}
-        label="Downtime"
-        tooltip={`Downtime is available time not used to cast anything (including not having your GCD rolling). This can be caused by delays between casting spells, latency, cast interrupting or just simply not casting anything (e.g. due to movement/stunned).<br/>
-        <li>You spent <b>${formatPercentage(this.activeTimePercentage)}%</b> of your time casting something.</li>
-        <li>You spent <b>${formatPercentage(this.downtimePercentage)}%</b> of your time casting nothing at all.</li>
-        `}
-        footer={(
-          <div className="statistic-bar">
-            <div
-              className="stat-health-bg"
-              style={{ width: `${this.activeTimePercentage * 100}%` }}
-              data-tip={`You spent <b>${formatPercentage(this.activeTimePercentage)}%</b> of your time casting something.`}
-            >
-              <img src={this.constructor.icons.activeTime} alt="Active time" />
-            </div>
-            <div
-              className="remainder DeathKnight-bg"
-              data-tip={`You spent <b>${formatPercentage(this.downtimePercentage)}%</b> of your time casting nothing at all.`}
-            >
-              <img src={this.constructor.icons.downtime} alt="Downtime" />
-            </div>
+    <StatisticBox
+      icon={<Icon icon="spell_mage_altertime" alt="Downtime" />}
+      value={`${formatPercentage(this.downtimePercentage)} %`}
+      label="Downtime"
+      tooltip={`Downtime is available time not used to cast anything (including not having your GCD rolling). This can be caused by delays between casting spells, latency, cast interrupting or just simply not casting anything (e.g. due to movement/stunned).<br/>
+      <li>You spent <b>${formatPercentage(this.activeTimePercentage)}%</b> of your time casting something.</li>
+      <li>You spent <b>${formatPercentage(this.downtimePercentage)}%</b> of your time casting nothing at all.</li>
+      `}
+      footer={(
+        <div className="statistic-bar">
+          <div
+            className="stat-health-bg"
+            style={{ width: `${this.activeTimePercentage * 100}%` }}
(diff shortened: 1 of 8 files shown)


src/Parser/Core/Modules/AlwaysBeCasting.js
```diff
@@ -172,40 +172,41 @@ class AlwaysBeCasting extends Analyzer {
     downtime: '/img/afk.png',
   };
   statistic() {
-    if (!this.showStatistic) {
+    if (!this.showStatistic || this.owner.boss.fight.noDowntimeSuggestion === true) {
```
MartijnHols: Boss may be null/undefined in which case this will crash. Also checking for `=== true` can be omitted.

src/Raids/AntorusTheBurningThrone/EonarLifebinder.js
```diff
@@ -8,6 +8,7 @@ export default {
   headshot: Headshot,
   fight: {
     vantusRuneBuffId: 250150,
+    noDowntimeSuggestion: true,
```
MartijnHols: How about `disableDowntimeStatistic` and `disableDowntimeSuggestion`?

I think by default it might be more desirable to just do `disableDowntimeSuggestion` and leave the statistic on so you can at least compare your own downtime with that of other players to find reasons why your performance is lower than that of others. What do you think?

Sharrq: The problem is your down time is completely dictated by how high or low your raids dps is. If the group is low on damage, then there is less downtime between packs,  if your groups damage is high then the is a lot of downtime waiting for spawns.

But for healing maybe the downtime stat is still useful?

Sharrq:
Made suggested changes

src/Raids/AntorusTheBurningThrone/EonarLifebinder.js
MartijnHols: Yeah and the personal downtime stat may be useless by itself. But if you're comparing yourself to other players with their (better) reports, the downtime stat might reveal that you and your raid - even if it's outside your own control - have more downtime. The reason for that might be due to personally being slow to cast things, move platforms or your raid being slow. I think having the statistic makes it easier to see.

If you're opposed to the statistic how about showing an alert box in it / apart of it that explains it's less accurate?

Sharrq:
I came up with a different plan and im working on it now. 

Im going to leave the downtime statistic box as is (with the option to still disable it via fight settings), but im also adding a warning box to show up under the checklist/suggestions to warn people that some statistics might be inaccurate. I should have it ready to be looked at shortly. Just cleaning up.

kfinch:
Oh neat, sounds good to me

Sharrq:
Alright, just sent it in.

Essentially i added the WarningDisplay, which can be used for other things in the futue. 
image


I also removed the disable Statistic setting from Eonar, but left the logic in AlwaysBeCasting so that it could still be used for some other purpose later (if needed)

Sharrq:
I figured out how to separate the warning display from the checklist.

image

MartijnHols:
I'd like to suggest incorporating it with the checklist panel and using an icon instead of the header.

How about something like this?

image

Sharrq:
if its incorporated with the checklist though, then it wouldnt show up if the checklist is not implemented, right?

If thats the case then maybe we go with mine until everyone has a checklist and then switch to yours?

I like yours better, but im just wondering what would happen if they have no checklist.

MartijnHols:
Right, good point. How about adding it to the suggestions panel also? I don't think the separate warning box looks appealing, especially on specs with a lot of checklist rules.

Sharrq:
Hmm, it could also get buried in a ton of suggestions. Can it be forced to the top of the suggestion list?

Sharrq:
Alright. I reattached it to the Checklist/Suggestions/About Section so it shows on all 3 of those menus. Also added the Warning Icon to it and removed the panel heading/lines/etc.

On the Checklist View
image

Suggestions View
image

About View
image

MartijnHols:
Top of suggestions would be fine, but then maybe also consider top of checklist.

The suggestions warning is missing the border, makes it look a bit weird.
I don't think the about page should have the warning. I don't think it matters that it disappears going to that tab, one should already be aware by then.

Could the warning text be any shorter? I think it's a lot of text that we could perhaps reword to be shorter. I'd prefer it not going over 3 lines in the warning.

There seems to be a lot of margin at the left side of the icon. How about making that about the same as the icon for the checklist items so it lines up? In my example I tried that and it looked fine.
How about a bit more of a brighter orange? e.g. #FFA100
I think the icon should be at least as big as the text, in my example it was slightly bigger, I think that looks slightly better.
It has a border at the bottom while it's the last item, I think that looks misplaced and should be removed.

src/Parser/Core/Modules/Features/WarningDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+      return (
+        <div>
+          <div style={{ padding: 0 }}>
+            <div className="flex-sub content-middle" style={{ margin: '10px 30px 10px 30px', color: '#FC9F06', opacity: 0.8 }}>
+                <div style={{ fontSize: '4em', lineHeight: 1, marginRight: 20 }}>
+                  <WarningIcon />
+                </div>
+                <div>
+                  <h5>
+                    Because of the way this encounter was designed, some statistics and suggestions may be inaccurate. Therefore it is recommended that you do not use this encounter as a guide for improving your overall play, and that you instead use these statistics and suggestions to compare yourself against other members of your raid and for improving your play for this encounter specifically.
```
MartijnHols: How about making this text configurable in the boss settings? That way we can specifically state what's wrong.

src/Parser/Core/Modules/Features/WarningDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+    if (boss && boss.fight.displayAccuracyWarning) {
+      return (
+        <div>
+          <div style={{ padding: 0 }}>
+            <div className="flex-sub content-middle" style={{ margin: '10px 30px 10px 30px', color: '#FC9F06', opacity: 0.8 }}>
+                <div style={{ fontSize: '4em', lineHeight: 1, marginRight: 20 }}>
+                  <WarningIcon />
+                </div>
+                <div>
+                  <h5>
```
MartijnHols: You shouldn't use `h5` for styling; it's considered unused and its styling might change in the future. If you want to give the any specific styling you can use the `style` attribute on the `div`.

Sharrq:
So because of the way the Checklist/Suggestions/About is clumped together, It just automatically showed on all three, i didnt intentionally do that. 

I do think it looks better on top though even if it is showing on about also.

image

image

Sharrq:
I made the requested changes ... Not sure what is up with the gitignore thing that got sent in. I didnt have the Warning Icon locally so i created it from your commit, so i told Github to ignore it assume it would just not upload it, but it uploaded gitignore instead.

Not sure how to remove that from the commit.

MartijnHols:
Just remove the line from the file and add the icon back to the repo.

PR #1054, opened 2017-12-25
Added ISB-tracking checklist items for Brewmasters

emallson:
This PR introduces two checklist items that track the use of Ironskin Brew for Brewmaster Monks. These checklist items give an at-a-glance view of:

1. The percentage of hits mitigated via Ironskin Brew (100% is the goal, in practice 99%+ is obtainable). Discounting fights with a large number of negligible hits while not actively tanking (which don't exist in the current relevant tiers as far as I know), this takes into account time not actively tanking better than a simple uptime check. The high requirements on uptime are set because having ISB inactive (ever) while actively tanking can easily result in a dead tank.

2. The amount of duration wasted by over-capping the ISB buff. Most (all?) tank AM buffs have a cap of 3x the duration of a single application of a buff (taking into account traits that increase the duration,  ie Potent Kick in this case). I added code to track the amount of duration lost by exceeding this cap. Ideally, none will be wasted.

The first item is the first thing checked in the Brewmaster discord when someone asks a damage-intake-related-question. The second item supports diagnosing the cause of low uptime.

(I am `emallson` on the brm discord if anyone cares to double-check or discuss numbers there; haven't joined the WoWAnalyzer discord yet but it'll be the same name when I do.)

### DIFF
--- a/src/Parser/Monk/Brewmaster/CHANGELOG.js
+++ b/src/Parser/Monk/Brewmaster/CHANGELOG.js
@@ -1,11 +1,16 @@
 import React from 'react';
-import { WOPR, Zerotorescue } from 'MAINTAINERS';
+import { WOPR, Zerotorescue, emallson } from 'MAINTAINERS';
 
 export default [
   {
     date: new Date('3000-01-01'),
     changes: <span style={{ color: 'red' }}>Changed completion status to <i>Not actively maintained</i> as @wopr resigned. Any help is welcome to continue support for this spec, see GitHub for more information.</span>,
     contributors: [Zerotorescue],
+  },
+  {
+    date: new Date('2017-12-24'),
+    changes: 'Added ISB uptime and clipping checklist items.',
+    contributors: [emallson],
   },
   {
     date: new Date('2017-08-24'),
--- a/src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
@@ -56,7 +56,6 @@ class IronSkinBrew extends Analyzer {
         this.durationLost += this.currentDuration - this.durationCap;
         this.currentDuration = this.durationCap;
       }
-      console.log(`ISB buff applied: ${this.currentDuration}`);
       // add this duration to the total duration
       this.totalDuration += addedDuration;
       this.lastDurationCheck = event.timestamp;
--- a/src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
@@ -21,6 +21,48 @@ class IronSkinBrew extends Analyzer {
   hitsWithoutIronSkinBrew = 0;
   damageWithoutIronSkinBrew = 0;
 
+  // duration tracking to record clipping of buff
+  lastDurationCheck = 0;
+  totalDuration = 0;
+  durationLost = 0;
+  currentDuration = 0;
+  durationPerCast = 6000; // base
+  durationPerPurify = 0;
+  durationCap = -1;
+
+  on_initialized() {
+    this.durationPerCast += 500 * this.combatants.selected.traitsBySpellId[SPELLS.POTENT_KICK.id];
+    this.durationCap = 3 * this.durationPerCast;
+    this.durationPerPurify = 1000 * this.combatants.selected.traitsBySpellId[SPELLS.QUICK_SIP.id];
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS.IRONSKIN_BREW.id === spellId || SPELLS.PURIFYING_BREW.id === spellId) {
+      // determine the current duration on ISB
+      if (this.currentDuration > 0) {
+        this.currentDuration -= event.timestamp - this.lastDurationCheck;
+        this.currentDuration = Math.max(this.currentDuration, 0);
+      }
+      // add the duration from this buff application (?)
+      let addedDuration = 0;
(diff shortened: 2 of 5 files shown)


emallson:
Commit 9fde4a6 works around the bug that can be seen in this log: https://wowanalyzer.com/report/g6kqfdVyBYbLKrTh/35-Heroic+The+Coven+of+Shivarra+-+Kill+(8:12)/Eisenpelz shows 0% ISB uptime, but as can be seen [on my casts](https://www.warcraftlogs.com/reports/g6kqfdVyBYbLKrTh/#fight=35&source=35&type=casts) ISB was cast 68 times. If you look at the data [on WCL](https://www.warcraftlogs.com/reports/g6kqfdVyBYbLKrTh/#fight=35&source=35&type=auras) the same issue manifests, so it isn't a WoWA bug.

src/Parser/Monk/Brewmaster/CHANGELOG.js
```diff
@@ -1,12 +1,17 @@
 import React from 'react';
-import { WOPR, Zerotorescue } from 'MAINTAINERS';
+import { WOPR, Zerotorescue, emallson } from 'MAINTAINERS';
 
 export default [
   {
```
MartijnHols: If you wish to give maintaining the spec a go feel free to remove this message (perhaps announce you'll be working on it) and update the maintainer and status in the config.

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
```diff
@@ -21,17 +21,71 @@ class IronSkinBrew extends Analyzer {
   hitsWithoutIronSkinBrew = 0;
   damageWithoutIronSkinBrew = 0;
 
+  // duration tracking to record clipping of buff
```
MartijnHols: We would prefer it if properties that are for internal usage are prefixed with an underscore.

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
```diff
@@ -21,17 +21,71 @@ class IronSkinBrew extends Analyzer {
...
+  durationLost = 0;
+  currentDuration = 0;
+  durationPerCast = 6000; // base
+  durationPerPurify = 0;
+  durationCap = -1;
+
+  buffShouldBeApplied = false;
+  buffNeverApplied = true;
+
+  on_initialized() {
```
MartijnHols: Your additions here seem to be about a different analysis from what the original analyzer did. How about making a new module for it instead? I think keeping things separate and small modules will make maintenance and reuse later easier.

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
```diff
@@ -57,7 +111,20 @@ class IronSkinBrew extends Analyzer {
     }
   }
 
+  // returns true if we hit the WCL bug with 100%/0% uptime on ISB
+  get wclUptimeBug() {
```
MartijnHols: You could fix this perhaps more consistently with a "Normalizer". Normalizers modify the events by changing, removing or adding new fabricated events to keep it more consistent and solve bugs in the combatlog. One such instance that we always considered was buffs not appearing despite the spell casts that should trigger them was being cast. In such instances you could fabricate an `applybuff`.

In fact if the ISB cast still triggers an `refreshbuff` event, then the `ApplyBuff` Normalizer should already take care of fabricating an `applybuff` even for you. Did you verify this in a broken log?

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
```diff
@@ -80,6 +147,30 @@ class IronSkinBrew extends Analyzer {
...
+      isLessThan: {
+        minor: 0.99,
+        average: 0.98,
+        major: 0.95,
+      },
+      style: 'percentage',
+    };
+  }
+
+  get clipSuggestionThreshold() {
```
MartijnHols: Every suggestion threshold should have a suggestion tied to it.

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
```diff
@@ -80,6 +147,30 @@ class IronSkinBrew extends Analyzer {
       });
   }
 
+  get suggestionThreshold() {
```
MartijnHols: Please update the original suggestion to use these values so that you don't have to keep the two values in sync.

MartijnHols: When you have 2 `suggestionThresholds` in one class please consider prefixing both with a name so they can be identified in isolation.

MartijnHols:
Thanks for the PR! This was a very good first PR.

Please consider talking to WOPR about maintaining this. He previously resigned as maintainer then mentioned making a few more changes but haven't seen or heard anything since, so I think it's safe to assume he's still resigned but asking won't hurt.

src/Parser/Monk/Brewmaster/Modules/Spells/IronSkinBrew.js
emallson: I hadn't thought of using a normalizer for that. I'll look into doing so.

I did confirm this with a broken log (and that my hack works around it, showing 100% ISB uptime after the log is finished parsing).

emallson:
I would like to go ahead and get this merged. To support other suggestions, I'm planning to do some refactoring anyway (assuming WOPR has resigned) and I'll fix things like the _ for private/internal properties and splitting the suggestions into separate modules as I do so.

Also re: removing the no-maintainer message, I'll talk to WOPR and then remove it afterwards.

PR #1056, opened 2017-12-25
Refactor ABC: mostly implementing the `isOnGCD` prop from Abilities

### LINKED ISSUE #612: Add "isOnGlobalCooldown" property to (cast efficiency) abilities definition
Currently ability info is spread over several files, but it would be much easier to just define spells in one place and mark those that are on GCD as such and those that aren't not.

MartijnHols:
Also implements a default for the suggestions, but most specs already override those so that won't change much.

This makes it so we no longer have to specify a second separate list of spec abilities to list everything on the GCD, just specifying the Abilities and using the `isOnGCD` prop will be enough. This simplifies the setup.

Fixes #612


MartijnHols:
That will be the preferred method yes, but I left `ABILITIES_ON_GCD` in case anyone has reasons to want to add other spell ids there... Shouldn't be the case I think.

Issue #443, opened 2017-10-02
Add a "you died and were dead for x seconds" suggestion


MartijnHols:
It could also check if the user had a CD available, and if he had used a Healthstone the fight. It might also be nice to suggest using abilities that might help (to point out the available CDs).

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=49979110)](https://www.bountysource.com/issues/49979110-a-you-died-and-were-dead-for-x-seconds-suggestion?utm_source=49979110&utm_medium=shield&utm_campaign=ISSUE_BADGE)

janvavra:
this module should be able to be to deal with or be disabled on Antorus final boss fight (multiple deaths are part of the encounter design)

PR #1066, opened 2017-12-26
[Core] Track Deaths and Time Spent Dead

Sharrq:
Tracks player deaths and resurrections, calculates time spent dead, has a suggestion for avoiding damage and using defensives, and adds a boss specific setting to disable on Argus.

For Issue #443


src/Parser/Mage/Fire/Modules/Features/DeathTracker.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import CoreDeathTracker from 'Parser/Core/Modules/DeathTracker';
+
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatNumber } from 'common/format';
+
+class DeathTracker extends CoreDeathTracker {
```
MartijnHols: I think this suggestion should be a part of the core version of this module so that it's implemented for everyone for free.

How about making the `You can also use defensive spells like <SpellLink id={SPELLS.BLAZING_BARRIER.id}/> or <SpellLink id={SPELLS.ICE_BLOCK.id}/> to avoid taking unavoidable damage. Additionally, ensure you are using Health Potions and Healthstones to heal yourself when you are very low.` part a property in the core so that we can use the config setting in CombatLogParser to change it as desired:

```jsx
deathTracker: [DeathTracker, { extraSuggestion: <Wrapper>You can also use defensive spells like <SpellLink id={SPELLS.BLAZING_BARRIER.id}/> or <SpellLink id={SPELLS.ICE_BLOCK.id}/> to avoid taking unavoidable damage. Additionally, ensure you are using Health Potions and Healthstones to heal yourself when you are very low.</Wrapper> }],
```

Keeping it that simple wouldn't require everyone to extend and override the core module.

src/Parser/Core/Modules/DeathTracker.js
```diff
@@ -0,0 +1,53 @@
...
+
+  deathTimestamp = 0;
+  resurrectTimestamp = 0;
+  timeDead = 0;
+  castCount = 0;
+  isAlive = true;
+
+  on_toPlayer_death(event) {
+    this.deathTimestamp = this.owner.currentTimestamp;
+    if (debug) {console.log("Player Died @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));}
```
MartijnHols: This could be shorter: `debug && console.log(...);` or without the curly brackets.

The strict code style requires if-statements with curly brackets to be multiline.

src/Parser/Core/Modules/DeathTracker.js
```diff
@@ -0,0 +1,53 @@
...
+
+const debug = true;
+
+class DeathTracker extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  deathTimestamp = 0;
+  resurrectTimestamp = 0;
```
MartijnHols: This could occur twice, how about naming this `lastResurrectionTimestamp` to make it more transparent that it might not include all resses?

src/Parser/Core/Modules/DeathTracker.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  on_finished() {
+    if (this.isAlive === false) {
+      this.timeDead += this.owner.currentTimestamp - this.deathTimestamp;
+    }
+  }
+
+  suggestions(when) {
+      when(this.castCount).isLessThan(1)
```
MartijnHols: As it is any implementers of the suggestion will override this suggestion.

MartijnHols: Please consider adding suggestion thresholds for both suggestions.

src/Parser/Mage/Fire/CombatLogParser.js
```diff
@@ -33,6 +34,7 @@ class CombatLogParser extends CoreCombatLogParser {
     cooldownThroughputTracker: CooldownThroughputTracker,
     damageDone: [DamageDone, { showStatistic: true }],
     cancelledCasts: CancelledCasts,
+    deathTracker: DeathTracker,
```
MartijnHols: How about enabling this in the core CombatLogParser so it's enabled for every spec?

src/Parser/Mage/Frost/Modules/Features/DeathTracker.js
```diff
@@ -0,0 +1,27 @@
...
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatNumber } from 'common/format';
+
+class DeathTracker extends CoreDeathTracker {
+
+  suggestions(when) {
+    const timeDeadSeconds = this.timeDead / 1000;
+    const boss = this.owner.boss;
+    if (!boss || !boss.fight.disableDeathSuggestion) {
```
kfinch: Does the death event contain any info about how the player died? Is it possible to ignore specifically the "RP kill" but still detect the rest?

src/Parser/Mage/Fire/Modules/Features/DeathTracker.js
kfinch: I agree, all the DeathTracking stuff should be core, only different between specs that I can think of would be the text of the suggestion.

src/Parser/Core/Modules/DeathTracker.js
```diff
@@ -0,0 +1,53 @@
...
+  on_byPlayer_cast(event) {
+    this.castCount += 1;
+  }
+
+  on_finished() {
+    if (this.isAlive === false) {
+      this.timeDead += this.owner.currentTimestamp - this.deathTimestamp;
+    }
+  }
+
```
kfinch: for less repeated code, could add in 
```
get timeDeadSeconds() {
   return (this.timeDead / 1000).toFixed(0);
}

get timeDeadPercent() {
   return this.timeDead / this.owner.fightDuration;
}
```

I've found `get` statements to be really nice for cleaning up code, as you can now use those in your suggestions without having to do the calculation each time, and it's still clear what you're doing.

src/Parser/Core/Modules/DeathTracker.js
```diff
@@ -0,0 +1,53 @@
...
+    this.isAlive = true;
+  }
+
+  on_byPlayer_cast(event) {
+    this.castCount += 1;
+  }
+
+  on_finished() {
+    if (this.isAlive === false) {
+      this.timeDead += this.owner.currentTimestamp - this.deathTimestamp;
```
kfinch: If you want to be really extra, can skip this `on_finished` statement and rename timeDead to `completedTimeDead`, then add 
```
get timeDead() {
   return this.completedTimeDead + (this.isAlive ? 0 : this.owner.currentTimestamp - this.deathTimestamp);
}
```
The advantage of this (other than swag), is that the timeDead value will update correctly as the fight loads, which the user sometimes sees. To be clear, you don't need to update this, I'm mostly just preaching the gospel of the `get` statement.

src/Parser/Mage/Frost/Modules/Features/DeathTracker.js
Sharrq: The problem with Argus, is walking into the tree doesnt count as a "ressurection". So regardless, it will say you were dead from the RP to the end of the fight.

kfinch: Ohhh yeah that makes it difficult.

Sharrq:
Alright, I think i got everything that was suggested ... but i might have missed something.

MartijnHols:
Could you add a changelog entry please? Then this is all good to go!

Sharrq:
added changelog and fixed the conflict in the changelog

MartijnHols:
Log where someone died:
https://wowanalyzer.com/report/RjH6AnYdP8GWzX4h/2-Heroic+Aggramar+-+Kill+(6:23)/Kantasai

Screenshot:
deathtracker

PR #1071, opened 2017-12-26
[Hunter] Focus Usage Module (shared)

### LINKED ISSUE #1058: [Hunter] show a focus usage diagram
Make something alike the vulnerable applications module, But in a shared module showing focus usage spread on the varying abilities in the specs

Pewtro:
This adds a module in which it breakdowns what the player used their focus on in the given encounter. 
Then it reorders many modules to make it look prettier

For Marksmanship:
image

For Beast Mastery:
image

fixes #1058

### DIFF
--- a/src/Parser/Hunter/Marksmanship/Modules/Features/Abilities.js
+++ b/src/Parser/Hunter/Marksmanship/Modules/Features/Abilities.js
@@ -48,7 +48,7 @@ class Abilities extends CoreAbilities {
       getCooldown: haste => 30,
       isActive: combatant => combatant.hasTalent(SPELLS.EXPLOSIVE_SHOT_TALENT.id),
       recommendedEfficiency: 0.95,
-      extraSuggestion: <span><SpellLink id={SPELLS.EXPLOSIVE_SHOT_TALENT.id}/> should be used on cooldown, and you should aim to hit it in the center of the mobs, as that will be where it does the most dmg.</span>,
+      extraSuggestion: <span><SpellLink id={SPELLS.EXPLOSIVE_SHOT_TALENT.id} /> should be used on cooldown, and you should aim to hit it in the center of the mobs, as that will be where it does the most dmg.</span>,
     },
     {
       spell: SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED,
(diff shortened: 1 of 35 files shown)


src/Parser/Hunter/Marksmanship/Modules/Features/Abilities.js
```diff
@@ -90,95 +90,159 @@ class Abilities extends CoreAbilities {
...
     {
       spell: SPELLS.TRUESHOT,
-      category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
-      getCooldown: (_, combatant) => 180 - combatant.owner.modules.quickShot.traitCooldownReduction,
-      recommendedEfficiency: 1.0,
-    },
+      category:
+      Abilities.SPELL_CATEGORIES.COOLDOWNS,
+      getCooldown:
+        (_, combatant) => 180 - combatant.owner.modules.quickShot.traitCooldownReduction,
```
MartijnHols: What happened here?

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+        this.focusSpenderCastsBM[spellId].focusUsed += VOLLEY_FOCUS_COST;
+      }
+      this.lastVolleyHit = event.timestamp;
+    }
+  }
+
+  focusUsageChart() {
+    let totalFocusUsed = 0;
+    const items = [];
+    const bmFocusSpenders = [[SPELLS.COBRA_SHOT.id], [SPELLS.MULTISHOT.id], [SPELLS.KILL_COMMAND.id], [SPELLS.REVIVE_PET_AND_MEND_PET.id], [SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id], [SPELLS.BARRAGE_TALENT.id], [SPELLS.VOLLEY_ACTIVATED.id]];
```
MartijnHols: Does every entry need to be an array? I'm having a hard time figuring out why you need this, and I'm thinking your code below might not be working as you're expecting it to.

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+        if (this.focusSpenderCastsMM[focusSpender].casts > 0) {
+          items.push({
+            color: this.focusSpenderCastsMM[focusSpender].color,
+            label: this.focusSpenderCastsMM[focusSpender].name,
+            spellId: focusSpender,
+            value: this.focusSpenderCastsMM[focusSpender].focusUsed,
+          });
+        }
+      });
+      mmFocusSpenders.map(ability => totalFocusUsed += this.focusSpenderCastsMM[ability].focusUsed);
```
MartijnHols: This should be a reducer

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+    }
+  }
+
+  focusUsageChart() {
+    let totalFocusUsed = 0;
+    const items = [];
+    const bmFocusSpenders = [[SPELLS.COBRA_SHOT.id], [SPELLS.MULTISHOT.id], [SPELLS.KILL_COMMAND.id], [SPELLS.REVIVE_PET_AND_MEND_PET.id], [SPELLS.A_MURDER_OF_CROWS_TALENT_SHARED.id], [SPELLS.BARRAGE_TALENT.id], [SPELLS.VOLLEY_ACTIVATED.id]];
+    const mmFocusSpenders = [[SPELLS.AIMED_SHOT.id], [SPELLS.MARKED_SHOT.id], [SPELLS.REVIVE_PET_AND_MEND_PET.id], [SPELLS.BARRAGE_TALENT.id], [SPELLS.VOLLEY_ACTIVATED.id], [SPELLS.WINDBURST.id], [SPELLS.PIERCING_SHOT_TALENT.id], [SPELLS.BLACK_ARROW_TALENT.id], [SPELLS.EXPLOSIVE_SHOT_TALENT.id]];
+    if (this.combatants.selected.spec === SPECS.MARKSMANSHIP_HUNTER) {
+      mmFocusSpenders.forEach(focusSpender => {
```
MartijnHols: This could be a map with the if statement as a filter before it

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+  SPELLS.EXPLOSIVE_SHOT_TALENT.id,
+];
+
+//code grabbed from Parser/Paladin/Holy/Modules/PaladinCore/CastBehavior.js
+
+class FocusUsage extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  focusSpenderCastsBM = {
```
MartijnHols: How about extending this module in the spec specific folders by overriding a single property instead? I'd prefer all spec specific code to be in the spec folders exclusively, and the lists of spenders are mostly config which is a good reason to extend and override a class.

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+    [SPELLS.COBRA_SHOT.id]: {
+      casts: 0,
+      focusUsed: 0,
+      name: "Cobra Shot",
```
MartijnHols: Single quotes is preferred. How about using the id and pulling the name from the spells object instead? That way you could optionally make it a SpellLink.

src/Parser/Hunter/Shared/Modules/Features/FocusUsage.js
```diff
@@ -0,0 +1,323 @@
...
+        </div>
+      </div>
+    );
+  }
+  statistic() {
+    return (
+      <div className="col-lg-3 col-md-4 col-sm-6 col-xs-12">
+        <div className="row">
+          <StatisticsListBox
+            title={<span> Focus usage</span>}
```
MartijnHols: This can be a (double-quoted since React) string.

src/Parser/Hunter/Marksmanship/Modules/Features/Abilities.js
Pewtro: Interesting, probably when I played around with the sentinel spell. Nothing is changed

Pewtro:
While I didn't make it able to extend it directly, I made it waaay less spec specific now to a degree where you simply add spells into 1 list

Pewtro:
I'll look into making this extendable at a later date (probably as part of the survival module in the new year), but if this could be merged as it is - that'd be great 😄

PR #1077, opened 2017-12-26
Fix Windwalker ABC by setting the default GCD

MartijnHols:


MartijnHols:
Need to remake this when Abilities/ABC refactor is in. This version isn't correct.

PR #1081, opened 2017-12-26
[BrM Monk] Maintenance & Internal Changes

emallson:
This PR has three main changes:

1. I marked myself as the maintainer and added some info about where to go for discussion. This follows discussion with @shighman (WOPR) via Discord PMs and I have his blessing to take over maintenance (he also said he didn't have anything in the pipeline to be added).
2. I added the mastery data for Brewmasters.
3. I removed a bunch of code for dealing with stagger and hacks around ISB buff bugs in WCL. Specifically, I added an `Analyzer` that generates reliable stagger pool update events for other modules to use (so each module doesn't have to maintain its own stagger tracking data) and replaced my hack-around for ISB buff events not showing (commit 9fde4a6e845c3bcebf2a3d6c06d68ea36cf31b15) with an `EventNormalizer` that generates them if it detects that they are missing. As part of this I also addressed some of the comments from my initial PR (private variables should now be _prefixed, though many are removed since stagger tracking is now centralized).

(in hindsight, this PR isn't very focused; I can split it into a couple of smaller PRs if that'd make review easier)

The main publicly-visible changes from this will be (a) me being maintainer new, (b) if the ISB bug happens, ISB uptime will now visibly update during parsing instead of jumping to 100% uptime at the end. The rest of the changes are for me in future work.


emallson:
I hadn't realized there were tests (they're in the old location and I hadn't looked). My stagger implementation changes broke them. I've fixed them, but run across an odd issue that I'm not sure how to handle properly.

Basically, the log contains all the events we need to determine the exact amount of damage in the stagger pool at any one time. The old implementation estimated the amount in the pool based on the amount each tick did. The new one instead constructs it exactly by measuring the amount put in (via absorb events) and the amount removed (via stagger ticking and the various means of purification). 

The resulting stagger numbers on real logs are identical (once I remembered to put in a trait I'd forgotten), but the new way is much easier to do cool things with since I can attribute the cause of every point of damage leaving the pool (e.g. removed by stagger tick, removed by Purifying Brew, or removed by the Quick Sip trait). However, on the test "logs" the numbers differ because the tick amounts in the test log are not 100% realistic (based on my napkin math). For the moment I've updated the numbers in the tests, but they're only really giving a consistency check at the moment. I probably need to talk with WOPR about it and then update the test data accordingly.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Fabricate events corresponding to stagger pool updates. Each stagger
+ * tick, purify, and stagger absorb generates one event.
+ */
+class StaggerFabricator extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  _has_t20_4pc = false;
```
MartijnHols: Please use camelCase for properties, e.g. `_hasTier20_4set`, the underscore is only here to avoid it becoming 204. Alternatively my preference is `has4PT20`.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+  _PURIFY_AMOUNT = PURIFY_BASE;
+  _stagger_pool = 0;
```
MartijnHols: Using a private property for this while making it public is unnecessary, please just call this `staggerPool`.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+  _has_quick_sip = false;
```
MartijnHols: `_hasQuickSip`.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+    if(player.hasTalent(SPELLS.ELUSIVE_DANCE_TALENT.id)) {
+      this._PURIFY_AMOUNT += ELUSIVE_DANCE_PURIFY;
+    }
+    this._PURIFY_AMOUNT += player.traitsBySpellId[SPELLS.STAGGERING_AROUND.id] * STAGGERING_AROUND_PURIFY;
+    this._has_quick_sip = player.traitsBySpellId[SPELLS.QUICK_SIP.id] > 0;
+    this._has_t20_4pc = player.hasBuff(SPELLS.XUENS_BATTLEGEAR_4_PIECE_BUFF_BRM.id);
+  }
+
+  on_toPlayer_absorbed(event) {
+    if (event.ability.guid === SPELLS.STAGGER.id) {
```
MartijnHols: It's preferred to avoid nesting if-statements. This could also be written like so at the top of the method:

```jsx
if (event.ability.guid !== SPELLS.STAGGER.id) {
  return;
}
```

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+      const amount = event.amount + (event.absorbed || 0);
+      this._stagger_pool += amount;
+      debug && console.log("triggering stagger pool update due to absorb");
+      this.owner.triggerEvent(EVENT_STAGGER_POOL_UPDATE, this._fab(event, amount));
+    }
+  }
+
+  on_toPlayer_damage(event) {
+    if (event.ability.guid === SPELLS.STAGGER_TAKEN.id) {
```
MartijnHols: It's preferred to avoid nesting if-statements. This could also be written like so at the top of the method:

```jsx
if (event.ability.guid !== SPELLS.STAGGER_TAKEN.id) {
  return;
}
```

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      const amount = event.amount + (event.absorbed || 0);
+      this._stagger_pool -= amount;
+      debug && console.log("triggering stagger pool update due to stagger tick");
+      this.owner.triggerEvent(EVENT_STAGGER_POOL_UPDATE, this._fab(event, -amount));
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.PURIFYING_BREW.id) {
+      const amount = this._stagger_pool * this._PURIFY_AMOUNT;
```
MartijnHols: If you added the `currentPurifyAmount` prop this could be just `this.currentPurifyAmount`.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      this._stagger_pool -= amount;
+      debug && console.log("triggering stagger pool update due to purify");
+      this.owner.triggerEvent(EVENT_STAGGER_POOL_UPDATE, this._fab(event, -amount));
+    } else if (this._has_quick_sip && event.ability.guid === SPELLS.IRONSKIN_BREW.id) {
```
MartijnHols: It might be worth placing a comment here to explain what Quick Sip is and does. I personally can't remember a single trait name, just their effects. It might also help slightly to name it `hasQuickSipTrait`.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      const amount = this._stagger_pool * ISB_QUICK_SIP_PURIFY;
+      this._stagger_pool -= amount;
+      debug && console.log("triggering stagger pool update due to ISB + Quick Sip");
+      this.owner.triggerEvent(EVENT_STAGGER_POOL_UPDATE, this._fab(event, -amount));
+    }
+  }
+
+  on_toPlayer_heal(event) {
+    if (this._has_t20_4pc && GIFT_OF_THE_OX_SPELLS.indexOf(event.ability.guid) !== -1) {
```
MartijnHols: The preferred method of writing this would be like so:

```jsx
if (!this._has4PT20 || !GIFT_OF_THE_OX_SPELLS.includes(event.ability.guid)) {
  return;
}
...rest
```
Alternatively you could consider making it two if-statements if that helps readability/scanability.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+import { GIFT_OF_THE_OX_SPELLS } from '../../Constants';
+
+const PURIFY_BASE = 0.4;
+const ELUSIVE_DANCE_PURIFY = 0.2;
+const STAGGERING_AROUND_PURIFY = 0.01;
+const ISB_QUICK_SIP_PURIFY = 0.05;
+const T20_4PC_PURIFY = 0.05;
+const debug = false;
+
+export const EVENT_STAGGER_POOL_UPDATE = "stagger_pool_update";
```
MartijnHols: Game's naming style is usually similar to `updatestaggerpool`

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  _fab(event, amount) {
+    return {
+      timestamp: event.timestamp,
+      type: EVENT_STAGGER_POOL_UPDATE,
+      ability: event.ability,
+      extraAbility: event.extraAbility,
+      kind: amount > 0 ? KIND_STAGGER_ADDED : KIND_STAGGER_REMOVED,
```
MartijnHols: How about making two events instead? e.g. `addstagger` and `removestagger`. We prefer small methods, and implementers will rarely need both for the same reason (and using kind is kinda awkward with the constants).

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+    if (this._has_t20_4pc && GIFT_OF_THE_OX_SPELLS.indexOf(event.ability.guid) !== -1) {
+      const amount = this._stagger_pool * T20_4PC_PURIFY;
+      this._stagger_pool -= amount;
+      debug && console.log("triggering stagger pool update due to T20 4pc");
+      this.owner.triggerEvent(EVENT_STAGGER_POOL_UPDATE, this._fab(event, -amount));
+    }
+  }
+
+  _fab(event, amount) {
+    return {
```
MartijnHols: We usually add the original event under the `reason` prop, in case it has info that is relevant to listeners (e.g. what triggered the change).

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      timestamp: event.timestamp,
+      type: EVENT_STAGGER_POOL_UPDATE,
+      ability: event.ability,
```
MartijnHols: I suggest removing this and relying on reason instead. Technically the ability of this event should be STAGGER if it were consistent with other events, and it would be better to let listeners rely on the common approach of using `reason` for the original event.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      extraAbility: event.extraAbility,
```
MartijnHols: same

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+      kind: amount > 0 ? KIND_STAGGER_ADDED : KIND_STAGGER_REMOVED,
+      amount: Math.abs(amount),
+      pooled_damage: this._stagger_pool,
```
MartijnHols: camelCase please. This could be `current` to make it consistent with `classResources`.

MartijnHols: It's unclear if this is the old or new value which is likely to cause confusing and may lead to errors. I recommend using `oldStaggerPool` and `newStaggerPool` instead so that there can be no confusion.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,104 @@
...
+ * Fabricate events corresponding to stagger pool updates. Each stagger
+ * tick, purify, and stagger absorb generates one event.
+ */
+class StaggerFabricator extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  _has_t20_4pc = false;
+  _PURIFY_AMOUNT = PURIFY_BASE;
```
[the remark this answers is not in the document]
MartijnHols: That getter could then lead to suggestions about when someone could have had a really good Purify

emallson:
Thanks for the review, I'll address those comments.

My day job is Rust, which has exactly the opposite naming convention (snake_case rather than camelCase) so switching back and forth is going to be FUN.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
emallson: The idea with that was to (strongly) indicate that although you can *read* the value, it should not be modified except within this module. A comment may suffice for that, though.

src/Parser/Monk/Brewmaster/Modules/Normalizers/IronskinBrew.js
```diff
@@ -0,0 +1,161 @@
...
+      // is done because the bug occurs when we have 100% uptime, and if
+      // we don't set the timestamp to be late enough we can actually
+      // get much less than 100% uptime
+      const firstDmg = events.find(event => event.type === "damage" && event.targetID === targetID);
+      const firstDmgIdx = events.indexOf(firstDmg);
+      events.splice(firstDmgIdx, 0, fab("applybuff", firstDmg.timestamp - 1));
+      let currentBuffTime = DURATION;
+      // now we go through the event list and fabricate correct buff
+      // events
+      for(let j = firstDmgIdx; j < events.length; j++) {
```
MartijnHols: I still don't really understand what this does. If I understand it correctly the only issue you're trying to fix is if ISB has a 100% uptime it never shows, right? So then wouldn't it suffice to insert it at the same spot as ApplyBuff does; after the combatantinfos? Why do you need to removebuff and refresh?

emallson: The idea had been to reconstruct *each* event that would be present. In theory, this is an initial ApplyBuff, then (I thought) a bunch of RefreshBuff events. The RemoveBuff stuff is present to check when I've gotten it wrong. However, upon reviewing more logs I've realized that the refreshbuff doesn't actually fire for ISB even on not-bugged logs (I expected it to behave like RJW, which has a refreshbuff event at each cast where the buff is already applied).

So, to match actual logs you are correct: I can just add an ApplyBuff immediately after the combatantinfos. This will simplify the module dramatically. I'll update the PR with that this evening.

src/tests/Parser/Brewmaster/Stagger.test.js
```diff
@@ -1,55 +1,68 @@
...
-  });
   it('Tracks the amount of stagger missing from the fight', () => {
     const earlyFightEnd = 6000;
     const myOwner = { fight: { end_time: earlyFightEnd } };
-    processEvents(EarlyFinish, stagger);
+    processEvents(EarlyFinish, fab, stagger);
     stagger.owner = myOwner;
     stagger.triggerEvent('finished');
-    expect(stagger.staggerMissingFromFight).toBe(285);
+    expect(stagger.staggerMissingFromFight).toBe(484);
```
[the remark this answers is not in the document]
emallson: The `staggerMissingFromFight` value *had been* calculated based on the most recent tick of stagger damage. The new method tracks the actual amount in the pool over the entirety of the fight. The stagger ticks in the fixture are not what they would be in an actual fight. This led to the old method underestimating the amount of damage missing *on the fixture*.

I spoke to WOPR about it via Discord PMs and his immediate response was that the test data was probably wrong, which lines up with my math. I noted this in my [previous comment](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/1081#issuecomment-354019767).

The test data needs to be replaced, but I'm not totally sure what to do for it yet (dump a real log to a file?) because stagger is a complicated piece of shit.

src/tests/Parser/Brewmaster/T20_4pc.test.js
```diff
@@ -1,24 +1,42 @@
...
   it('how many gift of the ox orbs were absorbed as a heal', () => {
-    processEvents(SimpleFight, item);
+    processEvents(SimpleFight, fab, item);
     expect(item.orbsEaten).toBe(1);
   });
   it('how much stagger was reduced by absorbing the gift of the ox orbs', () => {
-    processEvents(SimpleFight, item);
-    expect(item.staggerSaved).toBe(15);
+    processEvents(SimpleFight, fab, item);
+    expect(item.staggerSaved).toBe(23.450000000000003);
```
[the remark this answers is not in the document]
emallson: see above comment.

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -0,0 +1,111 @@
...
+export const EVENT_STAGGER_POOL_REMOVED = "removestagger";
+
+/**
+ * Fabricate events corresponding to stagger pool updates. Each stagger
+ * tick, purify, and stagger absorb generates one event.
+ */
+class StaggerFabricator extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  }
```
MartijnHols: Missing semicolon

src/Parser/Monk/Brewmaster/Modules/Normalizers/IronskinBrew.js
```diff
@@ -0,0 +1,97 @@
...
+    const isbCasts = {};
+    const isbBuffs = {};
+    let isbBuffAbility;
+    for(let i = 0; i < events.length; i++) {
+      const event = events[i];
+      const sourceID = event.sourceID;
+      const targetID = event.targetID;
+      const spellId = event.ability ? event.ability.guid : null;
+
+      if(event.type === "cast" && SPELLS.IRONSKIN_BREW.id === spellId) {
```
MartijnHols: Please don't use [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions), they're often harder to read and more importantly inconsistent with the rest of the codebase.

PR #1086, opened 2017-12-28
Abilities refactor

MartijnHols:
# Don't merge, just review

I refactored the Abilities to be easier to use.

* Normalized all property names,
* moved cast efficiency specific settings to a sub-object,
* import everything in a class for neat getters and setters,
* ability objects validation,
* allow for multiple types of values on almost all properties including a static value,
* set `this` to abilities so we don't have to keep adding parameters,

I migrated Holy Paladin, no other spec will work atm. I plan to migrate everything or if that proofs too much work add backwards compatibility.
Should note that all parameters except Haste in `cooldown` and `channel` will be considered deprecated.
I am considering adding a prop for marking spells as healing spells for the healing ABC.


MartijnHols:
This branch is now running at https://test.wowanalyzer.com to make it easier to verify if the new version will work with your spec.

MartijnHols:
# I know it's a lot to look at. Please at least check your own spec.

I reckon I fixed all issues. Please review so this can be merged. Things to look at:

* Specs importing properly (https://test.wowanalyzer.com)
* Previously cast efficiency suggestions were opt-out, now they're opt-in. Since I **manually** converted all Abilities classes I tried my best to add `suggestion: true` anywhere that didn't have a `noSuggestion` nor `noCanBeImproved` but did have a `cooldown` (spells without cooldown doesn't have a cast efficiency). I might have forgotten this once or twice, I also might have added it once or twice too many. Please ensure all spells you want suggestions for are marked with `suggestion: true`.

I'd really like to remove some more of the deprecated things still remaining in this code, but since this is now in a mergable state I'd like to merge it first and do that in a future PR instead.

Pewtro:
Marksmanship logs are stuck on initializing. Example: https://test.wowanalyzer.com/report/awKLZA4RgjTWhNbQ/2-Mythic+Felhounds+of+Sargeras+-+Kill+(4:43)/Putro
Beast-Mastery works. Example: https://test.wowanalyzer.com/report/fRa4VL3QJjcDnCNy/92-Mythic+Imonar+the+Soulhunter+-+Kill+(7:34)/Moofz

MartijnHols:
Looks to be an issue with the item spell. Thanks, investigating...

MartijnHols:
Aha, it extends the item but doesn't destructure the base's dependencies.

Pewtro:
👍 Everything else looks good for the hunter specs

emallson:
Looks good for Brewmaster. Plugged in a bunch of logs from our `#brew_logs` channel plus my own stuff and everything loads fine. The changes to the abilities file also check out (mumblegrumble will have to update my PR)

FaideWW:
Guardian loads fine, and the code changes look okay to me.

Juko8:
Windwalker is working and looking good

Sharrq:
Out of curiosity, would it be possible to incorporate the cast time for certain spells (like Ebonbolt) into Cast Efficiency as part of this?

For example, Ebonbolt is an ability that has a 45 second cooldown (not reduced by haste) and you want to use it as close to on cooldown as possible. It also has a 3 second cast time (reduced by haste). So once it comes off cooldown you need to spend 3 secondsish casting the next one, but those 3 seconds count as the spell not being used. So in some cases, at the end of the fight those 3 second increments are the difference between being able to get another cast or not, so the cast efficiency claims you missed a cast when you didnt. 

I have a bandaid on it by making the cooldown 45 + (3 / (1 + haste)) ... but that would account for the haste you have when the ability is cooling down, not the haste you have when you are actually casting it. So if the cast time could be incorporated, then that would solve that and go towards making cast efficiency that much more accurate :)

MartijnHols:
I don't think you should adjust the cooldown or even the efficiency of that spell; the *time on cooldown* indicator should be just that. If you want to account for channel time you could reduce the recommended cast efficiency.

kfinch:
Could make an effectively undocumented change to "time on cooldown or casting" to fix this issue, but I'm not sure how annoying that would be to implement

kfinch:
Also, Frost Mage and Resto Druid look fine to me

Sharrq:
i suppose ... it just seems awkward that you could literally hit the spell the millisecond that it comes off cooldown every time it comes off cooldown and still have cast efficiency tell you that you missed a cast.

PR #1090, opened 2017-12-28
ABC refactor: implement good support for channeled abilities

MartijnHols:
I rewrote most of ABC to be much simpler. ABC now goes for active time instead of downtime which is easier to calculate.
Channels are now handled through a single class that does event fabrication to make using its info trivial and move the challenge of marking channels properly to a single spot.
GlobalCooldown is in its own class to be isolated. It also triggers a new event to make usage much simpler and move away complexity out of ABC.


MartijnHols:
Ok if it were up to me this would be good to go

Pewtro:
If it doesn't break anything and it works then 👌 to me, but I haven't tinkered enough with channel to be able to discern much

PR #1082, opened 2017-12-26
[Warlock] Updates for new code style

sMteX:
I'm afraid this will be a bit painful to review but bear with me please. I've updated the Abilities, removed them from AlwaysBeCasting, reworked nearly all (I think I missed one intentionally) suggestions to use the object style, so I should be a little bit more ready for possible checklist implementation.

PR #1092, opened 2017-12-29
[Affliction] Affliction checklist

sMteX:
Implemented the checklist for Affliction. Also threw in the values for Affliction Mastery.

**!! Extends and requires #1082 , please merge that one before this one.**


src/Parser/Warlock/Affliction/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,198 @@
...
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.DARK_PACT_TALENT,
+            when: combatant.hasTalent(SPELLS.DARK_PACT_TALENT.id),
+          }),
+        ];
+      },
+    }),
+    new Rule({
+      name: 'Try to avoid being inactive for a large portion of the fight',
```
MartijnHols: This name makes sense for healers, but I think for a DPS you can actually tell someone to be active. You could name it 'Always be casting' instead

PR #1097, opened 2017-12-29
[BrM Monk] Replaced CD-based RJW suggestion with uptime-based suggestion

emallson:
Rushing Jade Wind has a fixed duration, but variable uptime. It is possible (and recommended) to leave RJW off cooldown quite a bit and still maintain high uptime. (See for example, [this fight](https://wowanalyzer.com/report/g6kqfdVyBYbLKrTh/7-Heroic+Felhounds+of+Sargeras+-+Kill+(3:59)/Eisenpelz/cast-efficiency) where I have <60% cast efficiency but > 85% uptime) This PR replaces the Abilities-based suggestion with an uptime-based suggestion.

I didn't see a way to do uptime-based stuff mostly automatically, so this is done as a new module.

(Also fixed final reviews from previous PR)

### DIFF
--- a/src/Parser/Monk/Brewmaster/Modules/Spells/RushingJadeWind.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Spells/RushingJadeWind.js
@@ -4,8 +4,7 @@ import SpellLink from 'common/SpellLink';
 import { formatPercentage } from 'common/format';
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
-
-const debug = false;
+import Wrapper from 'common/Wrapper';
 
 // the buff events all use this spell
 export const RUSHING_JADE_WIND_BUFF = SPELLS.RUSHING_JADE_WIND_TALENT; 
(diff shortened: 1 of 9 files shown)


src/Parser/Monk/Brewmaster/Modules/Spells/RushingJadeWind.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  on_finished() {
+    if(!this._active) {
+      this.inactiveDuration += this.owner.fight.end_time - this._lastDropped;
+    }
+  }
+
+  get uptime() {
```
MartijnHols: You can get uptime through `this.combatants.selected.getBuffUptime(SPELLS.RUSHING_JADE_WIND_TALENT.id)`, no need to do all this inactive time tracking.

src/Parser/Monk/Brewmaster/Modules/Spells/RushingJadeWind.js
```diff
@@ -0,0 +1,71 @@
...
+    const downtime = this.inactiveDuration / this.owner.fightDuration;
+    return 1.0 - downtime;
+  }
+
+  // using a suggestion rather than a checklist item for this as RJW is
+  // purely offensive
+  suggestions(when) {
+    when(this.uptime).isLessThan(0.95)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>You had low uptime on <SpellLink id={SPELLS.RUSHING_JADE_WIND.id} />. Try to maintain 100% uptime by refreshing the buff before it drops.</span>)
```
MartijnHols: Please use the `Wrapper` component from `common/Wrapper` when wrapping without any intent to have the container affect styling.

emallson:
Thanks for the feedback, I hadn't thought to check on the combatant object itself (I was looking in the `Abilities` module and related stuff). Cleaned it up

src/Parser/Monk/Brewmaster/Modules/Spells/RushingJadeWind.js
```diff
@@ -15,51 +14,17 @@ class RushingJadeWind extends Analyzer {
...
-    }
-    const downtime = this.inactiveDuration / this.owner.fightDuration;
-    return 1.0 - downtime;
-  }
-
   // using a suggestion rather than a checklist item for this as RJW is
   // purely offensive
   suggestions(when) {
-    when(this.uptime).isLessThan(0.95)
+    const uptime = this.combatants.selected.getBuffUptime(RUSHING_JADE_WIND_BUFF.id) / this.owner.fightDuration;
```
MartijnHols: For future reference it's generally best practice to make a getter for these checks so it can be used elsewhere too, it's easier to find and the suggestions method has less logic.

PR #1101, opened 2017-12-30
[Affliction] HOTFIX - The Master Harvester

sMteX:
That kind of a "way-too-late-at-night" bug. Master Harvester doesn't give the talent, just the effect sometimes procs. `ITEMS` was unnecessary then.


kfinch:
All of our fixes are hotfixes!

sMteX:
I'd say majority, sounds... less.. bad.. :smile:

PR #1102, opened 2017-12-30
[BrM Monk] Legendary Shoulders + Chest (and associated spells)

emallson:
Added support for the legendaries:

 - Sal'salabim's Lost Tunic (Chest)
 - Stormstout's Last Gasp (Shoulders)

and updated Breath of Fire suggestion + associated checklist item.

This also eliminates the error-spam in the dev log when Sal'salabim's (and, to a lesser extent, Stormstout's) is used since the cooldown resets (or extra charge) are correctly accounted for.


src/Parser/Monk/Brewmaster/Modules/Items/StormstoutsLastGasp.js
```diff
@@ -0,0 +1,48 @@
...
+  damage = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasShoulder(ITEMS.STORMSTOUTS_LAST_GASP.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if(event.ability.guid === SPELLS.KEG_SMASH.id) {
+      this.damage += KEG_SMASH_DMG_FACTOR * event.amount;
+    }
```
Pewtro: You should consider using the getBonusDamage module for this type instead, it allows you to use the correct modifiers as constants for clarity

src/Parser/Monk/Brewmaster/Modules/Spells/BreathOfFire.js
```diff
@@ -0,0 +1,44 @@
...
+
+class BreathOfFire extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    enemies: Enemies,
+  };
+
+  get uptime() {
+    return this.enemies.getBuffUptime(SPELLS.BREATH_OF_FIRE_DEBUFF.id) / this.owner.fightDuration;
+  }
```
Pewtro: Wouldn't this "break" in any AoE scenario - since you have multiple debuffs up at once?

emallson: Potentially. I looked at how AffLock debuffs were tracked and then did the same thing. It should probably be changed to the same metric as ISB (fraction of hits mitigated with it) since it may never be applied to adds that don't hit you. It works well enough on most fights in Antorus that it is worth replacing the old one (which completely breaks with the chest since we only want to cast often enough to maintain the buff).

src/Parser/Monk/Brewmaster/Modules/Items/StormstoutsLastGasp.js
emallson: Oh interesting, I'll check that out

emallson:
I don't think I'm going to have time to mess with trying to move BoF to the ISB-like '% of hits mitigated' metric for a few days at this point. My most recent logs on Coven show 92% uptime on the debuff (despite the fact that we used the "melee tank" strat where the tanks don't touch the shadow add so a mdps can tank it). I don't totally understand the code in the Enemies.getBuffUptime function, but it seems to be giving reasonable numbers.

PR #1106, opened 2017-12-30
[Balance] Added Astral Power usage tab

Gebuz:
and fixed a bug with the emerald dreamcatcher.


src/Parser/Druid/Balance/Modules/ResourceTracker/AstralPowerDetails.js
```diff
@@ -0,0 +1,62 @@
...
+    astralPowerTracker: AstralPowerTracker,
+  };
+  
+  suggestions(when) {     
+    resourceSuggest(when, this.astralPowerTracker, {
+      spell: SPELLS.RELENTLESS_STRIKES,
+      minor: 0.0,
+      avg: 0.02, 
+      major: 0.5,
+      extraSuggestion: <Wrapper> You are wasting more Astral Power. Always prioritize spending it over avoiding the overcap of any other ability. </Wrapper>,
```
kfinch: "You are wasting more Astral Power" sounds awkward. More than what? Maybe just say "You are wasting Astral Power"?

src/Parser/Druid/Balance/Modules/ResourceTracker/AstralPowerDetails.js
```diff
@@ -0,0 +1,62 @@
...
+import AstralPowerBreakdown from './AstralPowerBreakdown';
+
+class AstralPowerDetails extends Analyzer {
+  static dependencies = {
+    astralPowerTracker: AstralPowerTracker,
+  };
+  
+  suggestions(when) {     
+    resourceSuggest(when, this.astralPowerTracker, {
+      spell: SPELLS.RELENTLESS_STRIKES,
```
kfinch: What's with making `SPELLS.RELENTLESS_STRIKES` the spell? Is this just because it has an icon you like?

src/Parser/Druid/Balance/Modules/ResourceTracker/AstralPowerTracker.js
```diff
@@ -0,0 +1,17 @@
+import Combatants from 'Parser/Core/Modules/Combatants';
+import RESOURCE_TYPES from 'common/RESOURCE_TYPES';
+
+import ResourceTracker from 'Parser/Rogue/Subtlety/Modules/ResourceTracker/ResourceTracker';
```
kfinch: I take it you're importing Rogue's ResourceTracker because it's an obvious base, but I think it's generally not a good idea to import another spec's module like that. If the Subtlety dev decides to change something he may mess up your module and he'll probably not notice.

Instead, you should make a generalized ResourceTracker in Core somewhere (probably in Modules), and then make both the Subtlety and Balance specs extend it. I imagine in the future there will be other specs that have interest in that as well.

src/Parser/Druid/Balance/Modules/ResourceTracker/AstralPowerDetails.js
Gebuz: yeah I forgot to replace it.

Gebuz:
I moved the ResourceTracker to core, let me know if I did it incorrectly. I also removed the duplicate astral power statistic box and changed the original one to show per minute.

Gebuz:
Astral Power tab should be good now.
I also disabled Empowerment overcapping when using the Emerald Dreamcatcher as it will naturally overcap a lot of both empowerments.

PR #1111, opened 2017-12-30
Change Mastery per rating to a coefficient based approach and implemented Mastery coefficients for all specs

MartijnHols:


MartijnHols:
I didn't implement all the coefficients available to each spec, this might end up biting me in the ass some day but for now the only goal of this stat was to calculate the Mastery percentage shown on the character sheet, which the current coefficients are about. I'm too lazy to change it now.

PR #1112, opened 2017-12-31
Add Prydaz/Healthstone/Mark of the Ancient Priestess to SpellInfo for…

MartijnHols:
… healing stat values

### DIFF
--- a/src/Parser/Core/Modules/Features/SpellInfo.js
+++ b/src/Parser/Core/Modules/Features/SpellInfo.js
@@ -72,6 +72,27 @@ export default {
     mastery: false,
     vers: true,
   },
+  [SPELLS.XAVARICS_MAGNUM_OPUS.id]: { // Prydaz
+    int: false,
+    crit: false,
+    hasteHpct: false,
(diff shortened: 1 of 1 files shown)


src/Parser/Core/Modules/Features/SpellInfo.js
```diff
@@ -72,6 +72,27 @@ export default {
...
+    int: false,
+    crit: false,
+    hasteHpct: false,
+    mastery: false,
+    vers: false, // not 100% sure
+  },
+  [SPELLS.MARK_OF_THE_ANCIENT_PRIESTESS.id]: {
+    int: false,
+    crit: true,
+    hasteHpct: false,
```
kfinch: Does Mark have any haste PPM scaling? Because if so it should probably have `hasteHpct: true`

MartijnHols: No

MartijnHols: Unless both my source and Wowhead lied

PR #1113, opened 2017-12-31
 Update StatValues: Show HPS per 1 rating in a column

MartijnHols:
image

Looking this over I actually don't know why someone would want to know this. What applications does it have? I dislike the tooltip on every piece, but moving it to a column when the info isn't actually useful for anything seems worse. Maybe just remove it if it doesn't have a use?


MartijnHols:
image

HPS per 1 rating makes more sense, using decimals it's not half bad

anom0ly:
@MartijnHols Looks like when I went to resolve the merge conflict, I inadvertently pulled in the wrong portion. Unfortunately I can't push an update to your fork / branch directly it looks like. Can you take a look and just push a quick fix for the ESLint error? Thanks and sorry for screwing this up

MartijnHols:
image
it should work 😱

anom0ly:
I am probably just terrible at using Git / command line and screwed something up. I will look into it though.

PR #1119, opened 2018-01-01
[frost mage] Updated suggestions to new format, added new Checklist entry

kfinch:
New checklist entry "Fish for Procs" checks that you're spamming Frostbolt when you have no procs (rather than casting Ice Lance or Flurry without procs, which is almost always wrong)

Updated some suggestion texts / numbers too.


Koskun:
If you have a chance could you take a look at Water Jet.  WoWAnalyzer isn't catching any casts of it at all (verified on multiple comparisons of logs to what the Analyzer says).

Sharrq:
open an issue for it instead of commenting on a closed merged PR.

PR #1120, opened 2018-01-01
Fix Brewmaster's GCDs and channeling

MartijnHols:
It's still missing a lot of spells from the abilities config which will cause the timeline to not be complete, but I won't be fixing that.


emallson:
This cleans up the GCDs in the timeline quite a bit. The only issue I see is that Roll still overlaps, but I've no idea what to do about that. It isn't a large issue though.

MartijnHols:
I'll take that as an apporval :D

PR #1122, opened 2018-01-02
[Fire Mage] Suggestion for Phoenix Flames Charges before Combustion

Sharrq:
Added suggestion for casting Combustion with less than 2 charges of Phoenix Flames saved. Also added a function to SpellUsable that gives the option to check how many charges are available for a spell.


src/Parser/Mage/Fire/Modules/Features/Combustion.js
```diff
@@ -0,0 +1,52 @@
...
+        minor: 0,
+        average: 1,
+        major: 2,
+      },
+      style: 'number',
+    };
+  }
+
+  suggestions(when) {
+    when(this.lowPhoenixFlamesCharges).isGreaterThan(this.phoenixFlamesChargesSuggestionThresholds.isGreaterThan.minor)
```
kfinch: Please use the new suggestion format, where you pass in the suggestion thresholds object to `when` and you don't have to include `isGreaterThan`, `regular`, or `major` at all (because they're automatically read from the suggestion thresholds object)

src/Parser/Mage/Fire/Modules/Features/Combustion.js
```diff
@@ -0,0 +1,52 @@
...
+    if (spellId !== SPELLS.COMBUSTION.id) {
+      return;
+    }
+    const phoenixFlamesCharges = this.spellUsable.chargesAvailable(SPELLS.PHOENIXS_FLAMES.id);
+    if (phoenixFlamesCharges < 2) {
+      this.lowPhoenixFlamesCharges += 1;
+    }
+  }
+
+  get phoenixFlamesChargesSuggestionThresholds() {
```
anom0ly: Since you are already in the Combustion module, you could shorten then to `suggestionThresholds()` to just clean up the below usage of it.

src/Parser/Mage/Fire/Modules/Features/Combustion.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      actual: this.lowPhoenixFlamesCharges,
```
kfinch: This would work better as a percentage of Combustion casts where charges were low, because currently it's effectively skewed to be more strict on longer fights (with more Combustion casts)

src/Parser/Mage/Fire/Modules/Features/Combustion.js
Sharrq: im confused ... i am using the new suggestion format .... its right there in the section you highlighted
isGreaterThan(this.phoenixFlamesChargesSuggestionThresholds.isGreaterThan.minor)

not sure what i am missing here.

src/Parser/Mage/Fire/Modules/Features/Combustion.js
Sharrq: I decided not to do that because you cast like 3-4 Combustion's per fight, so missing one would immediately drop you to 60-70% and that felt awefully punishing. I can change it, it just seemed like it was more punishing than it should be.

src/Parser/Mage/Fire/Modules/Features/Combustion.js
sMteX: I think the point is - when you're constructing the suggestions object, you no longer need to use this verbose syntax and can simplify it to
```jsx
when(this.phoenixFlamesChargesSuggestionThresholds)
return suggest(<Wrapper>You cast <SpellLink id={SPELLS.COMBUSTION.id}/> {this.lowPhoenixFlamesCharges} times with less than 2 charges of <SpellLink id={SPELLS.PHOENIXS_FLAMES.id}/>. Make sure you are saving at least 2 charges while Combustion is on cooldown so you can get as many <SpellLink id={SPELLS.HOT_STREAK.id}/> procs as possible before Combustion ends.</Wrapper>)
    .icon(SPELLS.COMBUSTION.icon)
    .actual(`${formatNumber(this.lowPhoenixFlamesCharges)} Combustion Casts`)
    .recommended(`<${formatNumber(recommended)} is recommended`);
});
```

src/Parser/Mage/Fire/Modules/Features/Combustion.js
kfinch: Just code the suggestions with these numbers in mind. Another problem with the current implementation is that on short fights (eg with only time for 2 combustions), it's literally impossible to get a major issue, even if you fuck them both up (and who starts their 1st combustion with 1 PF charge anway?)

I think you have the right idea that missing even one should be a minor issue. maybe try thresholds of 100%, 65%, 45%? Or if you want to be a little harsher, 100%, 80%, 60%. Basically play in to your knowledge of where the likely breakpoints will be.

Sharrq: thats true, i guess it isnt as big of a deal since there is no statistic so the % isnt really visible to the user anyway. Ill change it.

Sharrq:
made suggested changes and resolved conflicts

Sharrq:
I had forgotten to add the comment for chargesAvailable ... so NOW i think i made all the requested changes

kfinch:
Those merge conflicts tho -_-

Sharrq:
I fixed those conflicts in one of the last two commits (i think) ... not sure why Travis hasnt completed though.

Sharrq:
Finally got travis to do his job.

PR #1125, opened 2018-01-02
[Balance] Added Starlord GCD reduction

Gebuz:
- Timeline and ABC should be more accurate now.
 - Increased ABC threshold for balance (most top logs have 4+% downtime, so expecting <2% doesn't seem reasonable). Might be worth considering making it a core change though.
 - Added myself as balance maintainer so I can enjoy DMs

### DIFF
--- a/src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
+++ b/src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
@@ -1,0 +1,56 @@
+import SPELLS from 'common/SPELLS';
+import CoreGlobalCooldown from 'Parser/Core/Modules/GlobalCooldown';
+import Channeling from 'Parser/Core/Modules/Channeling';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+import Abilities from '../Features/Abilities';
+import Haste from './Haste';
+
+/**
+ * The talent Starlord reduces GCD and cast time of empowered Lunar Strikes and Solar Wraths by 20%. 
+ * Since Solar Wrath cast time == GCD the GCD needs to be reduced.
+ */
+class GlobalCooldown extends CoreGlobalCooldown {
+  static dependencies = {
+    abilities: Abilities,
+    haste: Haste,
+    combatants: Combatants,
+    channeling: Channeling,
+  };
+
+  isStarlordBuffedSolarWrath(event) {
+  	return event.ability.guid === SPELLS.SOLAR_WRATH_MOONKIN.id && this.combatants.selected.hasTalent(SPELLS.STARLORD_TALENT.id) && this.combatants.selected.hasBuff(SPELLS.SOLAR_EMP_BUFF.id);
+  }
+
+  on_byPlayer_begincast(event) {
+    if (this.isStarlordBuffedSolarWrath(event)) {
+      this.triggerStarlordGlobalCooldown(event, 'begincast');
+    } 
+    return;
+  }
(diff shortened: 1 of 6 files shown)


src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
```diff
@@ -0,0 +1,56 @@
...
+    haste: Haste,
+    combatants: Combatants,
+    channeling: Channeling,
+  };
+
+  isStarlordBuffedSolarWrath(event) {
+  	return event.ability.guid === SPELLS.SOLAR_WRATH_MOONKIN.id && this.combatants.selected.hasTalent(SPELLS.STARLORD_TALENT.id) && this.combatants.selected.hasBuff(SPELLS.SOLAR_EMP_BUFF.id);
+  }
+
+  on_byPlayer_begincast(event) {
```
MartijnHols: My suggestion was to override the `getCurrentGlobalCooldown` method and making it return a 20% reduced GCD when you have the buff and it's an applicable event, would that not suffice?

It will take some more time to review this as it is since it's harder to grasp what behavior is being changed. I think this might break the channel/gcd time behavior but I haven't reviewed the base class to make sure.

src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
```diff
@@ -0,0 +1,56 @@
...
+    } 
+    super.on_byPlayer_cast(event);
+  }
+
+  triggerStarlordGlobalCooldown(event, trigger) {
+    this.owner.triggerEvent('globalcooldown', {
+      type: 'globalcooldown',
+      ability: event.ability,
+      timestamp: event.timestamp,
+      duration: Math.max(750, this.getCurrentGlobalCooldown(event.ability.guid) * 0.8), //Starlord multiplier
```
MartijnHols: Are you sure the minimum is 750 with the buff? Most of the time the minimum GCD is exactly 50% of the default, and this buff seems to reduce the *default* by 20%.

src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
Gebuz: yeah that would probably have been simpler, I'll take a look at that.

src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
Gebuz: yeah afaik the only way to reduce GCD below the minimum 750 is the Elisande buff from NH, atleast for balance. I think it's the same for all casters.

src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+class GlobalCooldown extends CoreGlobalCooldown {
+  static dependencies = {
+    abilities: Abilities,
+    haste: Haste,
+    combatants: Combatants,
+    channeling: Channeling,
+  };
+
+  getCurrentGlobalCooldown(spellId = null) {
```
MartijnHols: Try to avoid copying any code, especially from core classes as they may change at any time. You can access the base method by using `super.getCurrentGlobalCooldown(spellId)`, if you put this in a var you can don't need to do the GCD calculating twice.

Gebuz: It uses super now

PR #1127, opened 2018-01-02
[BrM Monk] Brew Cooldown Reduction

emallson:
**tl;dr** one of brewmaster's two main mechanics is reducing brew cooldowns via your rotation (the other is Stagger). this adds a checklist item + stat tracking how much the cooldown was reduced.

This adds modules for the relevant abilities (Black Ox Brew, Keg Smash, Tiger Palm) and items (Anvil-Hardened Wristwraps), and a module for aggregating this data (BrewCDR).

Also correctly sets the cooldowns for ISB and BoB (but not PB -- waiting on abilities refactor (maybe?) to do linked CDs. everything can just reference ISB CD to see PB CD).

The brew CDR formula is set up to display it in a manner similar to games like League of Legends (e.g. if it shows 60% effective CDR, then your brews CDs were effectively 40% of what they'd normally be).

If you have any suggestions for improving how I estimate when we get a Face Palm proc, I'd love to hear them. Right now I'm estimating the damage of FP vs non-FP Tiger Palm hits and then picking the one that is closer. This gives reasonable results, but certainly isn't perfect (though I can't say I've seen it be wrong first-hand).

(Another 500-line PR \o/ At least its all one feature this time)


emallson:
After some discussion on discord about how to do it, I ended up implementing the shared CD for brews via a fake ability (using an unused spell ID).

I tried putting both ISB and PB in the same spell using the ability format, but that did not work in basically any sense of the word. Thus, instead ISB and PB both start the cooldown of the fake shared brew themselves and other spells reduce this cooldown instead of reducing ISB/PB cooldowns.

src/Parser/Monk/Brewmaster/Modules/Features/Checklist.js
```diff
@@ -78,6 +80,24 @@ class Checklist extends CoreChecklist {
         ];
       },
     }),
+    new Rule({
+      name: (
```
MartijnHols: You should only use the Wrapper component when combining strings and JSX. In all other cases please use strings or template literals.

src/Parser/Monk/Brewmaster/Modules/Items/AnvilHardenedWristwraps.js
```diff
@@ -0,0 +1,46 @@
...
+      const actualReduction = this.isb.reduceCooldown(WRISTS_REDUCTION);
+      this.cdr += actualReduction;
+      this.wastedCDR += WRISTS_REDUCTION - actualReduction;
+    } 
+  }
+
+  item() {
+    return {
+      item: ITEMS.ANVIL_HARDENED_WRISTWRAPS,
+      result: (
```
MartijnHols: Please use a template literal instead, using JSX here is unnecessary.

src/Parser/Monk/Brewmaster/Modules/Spells/TigerPalm.js
```diff
@@ -0,0 +1,138 @@
...
+          this.normalHits += 1;
+        }
+      }
+    }
+  }
+
+  on_toPlayer_damage(event) {
+    // record the last known AP to estimate the amount of damage a
+    // non-FP TP should do
+    if (event.hasOwnProperty("attackPower") && event.attackPower > 0) {
```
MartijnHols: `event.hasOwnProperty("attackPower")` should be either just `event.attackPower` or `event.attackPower !== undefined`.

src/Parser/Monk/Brewmaster/Modules/Spells/TigerPalm.js
```diff
@@ -0,0 +1,138 @@
...
+  fpHits = 0;
+  bocFpHits = 0;
+  lastAttackPower = 0;
+
+  cdr = 0;
+  wastedCDR = 0;
+  fpCDR = 0;
+  wastedFpCDR = 0;
+
+  boc_buff_active = false;
```
MartijnHols: Please use camelCase naming for class properties.

src/Parser/Monk/Brewmaster/Modules/Spells/TigerPalm.js
```diff
@@ -0,0 +1,138 @@
...
+const FACE_PALM_AP_RATIO = 1.281;
+const TIGER_PALM_REDUCTION = 1000;
+const FACE_PALM_REDUCTION = 1000;
+
+class TigerPalm extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    boc: BlackoutCombo,
+    brews: SharedBrews,
+  }
```
MartijnHols: Missing semicolon

src/Parser/Monk/Brewmaster/Modules/Spells/TigerPalm.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  // predicate function returning `true` if we think this is an FP proc,
+  // `false` otherwise
+  // 
+  // we're going to do this by using halfway between non-fp/fp as a
+  // cutoff. anything >= the cut-off is FP, and anything < the cut-off
+  // is non-FP
+  _face_palm_proc(event) {
```
MartijnHols: Please use camelCasing naming for methods

src/common/SPELLS/MONK.js
```diff
@@ -529,6 +529,11 @@ export default {
     name: 'Staggering Around',
     icon: 'ability_monk_fortifyingale_new',
   },
+  FAKE_SHARED_BREWS: {
+    id: 9001,
```
MartijnHols: I recommend using a sub-zero spell id instead. This may conflict in the future.

src/Parser/Monk/Brewmaster/Modules/Spells/KegSmash.js
```diff
@@ -0,0 +1,88 @@
...
+  bocHits = 0;
+
+  cdr = 0;
+  bocCDR = 0;
+  wastedCDR = 0;
+  wastedBocCDR = 0;
+
+  _next_target = 0;
+
+  _boc_buff_active = false;
```
MartijnHols: Please use camelCase naming for class properties

src/Parser/Monk/Brewmaster/Modules/Spells/KegSmash.js
```diff
@@ -0,0 +1,88 @@
...
+
+const KEG_SMASH_REDUCTION = 4000;
+const BOC_KEG_SMASH_REDUCTION = 2000;
+
+class KegSmash extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    boc: BlackoutCombo,
+    brews: SharedBrews,
+  }
```
MartijnHols: Missing semicolon

src/Parser/Monk/Brewmaster/Modules/Spells/KegSmash.js
```diff
@@ -0,0 +1,88 @@
...
+  on_byPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if(SPELLS.BLACKOUT_COMBO_BUFF.id === spellId) {
+      this._boc_buff_active = false;
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if(SPELLS.KEG_SMASH.id === spellId) {
```
MartijnHols: It's preferred to return early instead of nesting large pieces of code in if-statements. As code grows this quickly becomes hard to understand and maintain.

emallson:
Thanks for the review.

I've got two questions for you:

1. This is a big monolithic PR because nothing in it really does anything (correctly, at least) without everything. The only thing I could maybe split out and have something useful would be BoB usage, but even that depends on tracking brew cooldowns. Would you rather it be split into smaller units that don't do anything (yet) instead of one huge blob?

2. I don't quite understand what you mean by this:
> I would have greatly preferred if you added support for shared CDs to the core SpellUsable/cast efficiency spells. This would have been much cleaner, easier to use and understand (since it would actually match reality), less likely to break and better reusable.

You mean modify the `Core` module for `SpellUsable` so that it works for shared CDs / charges instead of doing the hack with the fake ability?

MartijnHols:
1. It looks like `KegSmash`, `AnvilHardenedWristwraps`, `PurifyingBrew`, `TigerPalm` and `BlackOxBrew` could have been in separate PRs as they're new modules on top of what you have here. This would have made for a couple of small PRs that would be pretty easy to review. It's fine to make PRs depend on other PRs, you can achieve this by making branches at any point you feel a part is ready.

But I suppose it is likely the basis was still in flux at the time you made those other modules so it might have been hard. Again you're not required to split them, it just helps make reviewing easier and usually quicker. I do understand you kept it together in this instance and that makes sense too, it just makes it harder to review now.

2. Yes, this problem isn't unique to your spec and it would make the project better for everyone if this were handled properly rather than requiring a "hack". We don't require this really because we do want you to be able to develop things as you see fit and I understand if modifying core modules might appear more challenging, but now we're still left with the issue in the core modules and if anyone were to want to try to tackle it, they're going to have to try to fit it in your implementation without breaking anything without sufficient knowledge about the spec.

PR #1130, opened 2018-01-03
[Fire Mage] Marquee Bindings of the Sun King

fuzzys:
Hello and first of all I want to say this is my first time contributing to this (or anything like this) so forgive any mistakes and let me know what I can do to improve. This is a really cool project and I've wanted to learn how to code for awhile so I thought it'd be a good way practice.

This adds the legendary Marquee Bindings of the Sun King to Fire Mages.

- detects how many times it procs
- shows how many of those procs are overwritten, expired, or used up
- calculates DPS from the combined buffed Pyroblasts

I checked it with a few logs and the numbers look to be adding up correctly except for the DPS, which is somehow off by a percentage point despite the damage done being added up correctly.

I commented the whole thing so people here who know better can tell what I'm trying to do. Please let me know if there's anything to fix or improve, and thank you!

### DIFF
--- a/src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
+++ b/src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
@@ -3,127 +3,137 @@ import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
-import ItemDamageDone from 'Main/ItemDamageDone';
 import ItemLink from 'common/ItemLink';
-import { formatPercentage } from 'common/format';
+import { formatPercentage, formatNumber } from 'common/format';
+import Wrapper from 'common/Wrapper';
+import SpellLink from 'common/SpellLink';
 
 class MarqueeBindingsOfTheSunKing extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
   
-  damage = 0;
-  legendaryProcs = 0;
-  overwrittenProcs = 0;
+  baseProcs = 0;
+  refreshedProcs = 0;
   buffedPyroblasts = 0;
-  expiredProcs = 0
-  
-  lastBuffTimestamp;
-  lastPyroCastTimestamp;
+  buffChecked = 0;
+  combustionRefresh = 0;
   
   on_initialized() {
   this.active = this.combatants.selected.hasWrists(ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING.id);
   }
 
-	// get total count of procs
-  
-    on_toPlayer_applybuff(event) {
-	  if (event.ability.guid !== SPELLS.KAELTHAS_ULTIMATE_ABILITY.id) {
-	  return;
+  // count procs + refreshes  
+  on_toPlayer_applybuff(event) {
+    if (event.ability.guid !== SPELLS.KAELTHAS_ULTIMATE_ABILITY.id) {
+    return;
     }
-  this.legendaryProcs += 1;
-  }
+    this.baseProcs += 1;
+  } 
   
-   // get count of refreshes
-  	on_byPlayer_refreshbuff(event) {
+  on_byPlayer_refreshbuff(event) {
     if (event.ability.guid !== SPELLS.KAELTHAS_ULTIMATE_ABILITY.id) {
-			return;
-		}
-		this.overwrittenProcs += 1;
-		this.legendaryProcs += 1;
+      return;
+    }
+    if (this.combatants.selected.hasBuff(SPELLS.COMBUSTION.id))
+    {this.combustionRefresh +=1;
+    return;
+    }
+    this.refreshedProcs += 1;
 	}
-	
-	
-	// this section needs explaining:
-	// from looking through logs, the order of events is buff expires -> pyroblast cast begins within ~100ms -> pyroblast hits within 4.5 seconds
-	// i'm not sure this is the best way to accomplish this but it seems to work
-	
-	//get timestamp when buff expires
-	
-	on_byPlayer_removebuff(event) {
+  
+  // when buff expires, start check for matching pyroblast
+ 
+  on_byPlayer_removebuff(event) {
     if (event.ability.guid !== SPELLS.KAELTHAS_ULTIMATE_ABILITY.id) {
-			return;
-		}
-		this.lastBuffTimestamp = this.owner.currentTimestamp;
-			}
-	
-	// for each pyroblast, check if buff expired within the past 100ms. if so, count it as buffed by the legendary
-	// additionally, set timestamp of cast to do a damage check
-	// i suppose it's possible for a player to instant cast pyroblast with hot streak within 100ms of the buff ending so it checks that this isn't the case.
-	
-	on_byPlayer_cast(event) {
+      return;
+    }
+    this.buffChecked=1;
+    }
+  
+  on_byPlayer_cast(event) {
     if (event.ability.guid !== SPELLS.PYROBLAST.id) {
-			return;
-		}
-    if (this.lastBuffTimestamp > this.owner.currentTimestamp - 100 && !this.combatants.selected.hasBuff(SPELLS.HOT_STREAK) ) {
-			this.buffedPyroblasts += 1;
-			this.lastPyroCastTimestamp = this.owner.currentTimestamp;
-		}
-		}
-
-	// after being cast, pyroblast takes a few seconds to travel and hit the target.
-	// allows 3 seconds for the pyroblast to hit, then resets timestamp so additional instant pyroblasts within the time limit don't count.
-	// this is messy and i don't know if there's an easier way
+      return;
+     }
+	 	 
+   if (this.buffChecked === 0) {
+       return;
+     }  
+   
+  // make sure that this isn't a hot streak instant pyroblast that just happened to cast as buff fell off
+     
+     if (this.combatants.selected.hasBuff(SPELLS.KAELTHAS_ULTIMATE_ABILITY.id, event.timestamp, 100) && !this.combatants.selected.hasBuff(SPELLS.HOT_STREAK)) {
+      this.buffedPyroblasts += 1;
+      this.buffChecked = 0;
+    }
+    }	
 	
-	on_byPlayer_damage(event) {
-    if (event.ability.guid !== SPELLS.PYROBLAST.id) {
-			return;
-		}
-		
-	if (this.owner.currentTimestamp - 3000 < this.lastPyroCastTimestamp)	
-			this.damage += event.amount + (event.absorbed || 0);
-			this.lastPyroCastTimestamp = 0;
-			console.log(this.damage);
-		}
+  // math
+	
+  get totalProcs() {
+    return (this.baseProcs + this.refreshedProcs + this.combustionRefresh);
+  }
+ 
+  get expiredProcs() {
+    return (this.totalProcs - this.refreshedProcs - this.buffedPyroblasts - this.combustionRefresh);
+  }
 
-  item() {
-	  
-	  // i don't think this is the best section to put this code
-	  	this.expiredProcs = (this.legendaryProcs - this.overwrittenProcs - this.buffedPyroblasts);
+  get refreshPercentage() {
+    return (this.refreshedProcs) / (this.totalProcs);
+  }
 
+// item display
+   
+item() {
+  return {
+  item: ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING,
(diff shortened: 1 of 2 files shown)


Gebuz:
Hey, awesome that you want to contribute! 
Just a thing to consider about what you actual want to display, it might make more sense to only display number of procs, overwritten and expired. Showing the raw DPS from the buffed pyroblasts might be slightly misleading, as the actual DPS increase from the effect would need to include the loss from the spells you would have been casting had you not been casting the buffed pyro (and that is pretty complicated to estimate).

vituscze:
Even with perfect play you cannot avoid all overwrites. Looking at the T21 profile, it overwrites on average 1.1 procs out of 15.4 (or about 7.2%).

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+	
+	on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.PYROBLAST.id) {
+			return;
+		}
+		
+	if (this.owner.currentTimestamp - 3000 < this.lastPyroCastTimestamp)	
+			this.damage += event.amount + (event.absorbed || 0);
+			this.lastPyroCastTimestamp = 0;
+			console.log(this.damage);
```
kfinch: console logs for non-error behavior shouldn't make it to production. Please either remove this or hide it behind a debug flag.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+	
+	// for each pyroblast, check if buff expired within the past 100ms. if so, count it as buffed by the legendary
+	// additionally, set timestamp of cast to do a damage check
+	// i suppose it's possible for a player to instant cast pyroblast with hot streak within 100ms of the buff ending so it checks that this isn't the case.
+	
+	on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.PYROBLAST.id) {
+			return;
+		}
+    if (this.lastBuffTimestamp > this.owner.currentTimestamp - 100 && !this.combatants.selected.hasBuff(SPELLS.HOT_STREAK) ) {
```
kfinch: the `hasBuff` function has an optional 'buffer' field that could be used instead of saving timestamps like this. Basically, this check could instead be `if (this.combatants.selected.hasBuff(SPELLS.KAELTHAS_ULTIMATE_ABILITY.id, event.timestamp, 100)) {`. See hasBuff in the Entity class for more details on the function.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+	if (this.owner.currentTimestamp - 3000 < this.lastPyroCastTimestamp)	
+			this.damage += event.amount + (event.absorbed || 0);
+			this.lastPyroCastTimestamp = 0;
+			console.log(this.damage);
+		}
+
+  item() {
+	  
+	  // i don't think this is the best section to put this code
+	  	this.expiredProcs = (this.legendaryProcs - this.overwrittenProcs - this.buffedPyroblasts);
```
kfinch: calculations for derived values can be put in getter functions. In this case, you'd have a seperate function
```
get expiredProcs() {
   return this.legendaryProcs - this.overwrittenProcs - this.buffedPyroblasts;
}
```
and then you'd access the value simply with `this.expiredProcs` ... this is in same namespace as instance variables, so of course you'd also have to delete the `this.expiredProcs = 0` at the top of the class.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+
+    return {
+		
+	 //	i debugged this and the console log is returning the expected damage done but the math for dps isn't working as i expect?
+	 // instead it returns something within about one % of the correct number
```
kfinch: How are deciding what 'the correct number' is? What calculation are you comparing to?

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+		
+      item: ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING,
+	  result: <ItemDamageDone amount={this.damage} />,
```
kfinch: Damage numbers for an item are intended to show "increase in damage vs not having the the item". That's difficult for this legendary because there is a large opportunity cost to hardcasting a Pyroblast, and hardcasting a Pyroblast is not something you would do if you didn't have the legendary.

I don't think there is a reasonable way to calculate the opportunity cost (if you think of something let me know). There are two options I can think of here:
1) Show a `~` before the damage amount, explain in tooltip the opportunity cost issue.
2) Don't show a damage value at all, instead show procs per minute or something.

Also, the damage tally you're using completely misses the damage added to your Ignite by the proc'd pyro. I think there may be an Ignite module to help with this calculation?

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+    };
+  }
+  
+	// this is wordy, improve? also: change recommended numbers
+	// i couldn't find the icon for the bindings so i used pyroblast
+  
+    suggestions(when) {
+    when(this.overwrittenProcs).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>Your buff from <ItemLink id={ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING.id} /> was overwritten {formatPercentage(this.overwrittenProcs/this.legendaryProcs)}% of the time. This happens when the buff procs from an instant cast Pyroblast during Hot Streak before it is hard cast. If possible avoid using Fire Blast or Phoenix's Flames during Heating Up while you have the Marquee Bindings of the Sun King buff active.</span>)
```
kfinch: Please use `<Wrapper>` instead of `<span>`. Wrapper is basically the same thing but with much better render performance. I know you probably saw a lot of spans in the existing code, and we'll eventually clean it up, but for now the policy is "no new spans".

Get the import with
```
import Wrapper from 'common/Wrapper ';
```

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+          .icon(SPELLS.PYROBLAST.icon)
+          .actual(`${this.overwrittenProcs} overwritten buffs`)
+          .recommended(`0 is recommended`)
+		  .regular(recommended - 0.0).major(recommended - 0.0);
+      });
+	  
+	  when(this.expiredProcs).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>Your buff from <ItemLink id={ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING.id} /> expired {formatPercentage(this.expiredProcs/this.legendaryProcs)}% of the time. This happens when Pyroblast isn't hard cast during the fifteen seconds the buff is up. Note that this can also happen if the buff procs shortly before the fight ends when you wouldn't have had time to use it.</span>)
```
kfinch: Same thing as above, Wrapper instead of span please.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,129 @@
...
+      item: ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING,
+	  result: <ItemDamageDone amount={this.damage} />,
+    };
+  }
+  
+	// this is wordy, improve? also: change recommended numbers
+	// i couldn't find the icon for the bindings so i used pyroblast
+  
+    suggestions(when) {
+    when(this.overwrittenProcs).isGreaterThan(0)
```
kfinch: We've recently updated to a new suggestions format, I'd like to see it implemented here. Basically, we create a suggestions object that's passed into `when`, and this obviates the need for the `isGreaterThan`, `regular`, and `recommended` calls.

For an example of this in action, see https://github.com/WoWAnalyzer/WoWAnalyzer/blob/4b27611bf802cffa4b7aa380ff6242cf26df5e33/src/Parser/Mage/Frost/Modules/Features/GlacialSpike.js#L47

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
sMteX: Huh, didn't know that - maybe I could refactor some of my Warlocks' code using this buffer, thanks!

Sharrq:
I could be reading it wrong, and im at work so i cant try it, but wouldnt subtracting a flat 3 seconds from every pyroblast damage event be a bit flawed? Would that not also count an instant pyroblast immediately after the hardcasted pyroblast?

Marquee Bindings has proven to be rather difficult to get right .... I managed to get it working and then in antorus the instant cast pyroblasts didnt have a begincast event anymore and I had to start over. If you can manage to get it to work then great, but i dont think allowing a flat "travel time" is going to work without issue.

fuzzys:
Thanks for the feedback everyone! Made an update with the requested changes.

I removed damage calculation for now since it's more complicated than I thought (though I'd love to find a way to estimate it.)

I edited the two suggestions slightly, which I could use feedback on: first of all I excluded refreshes during Combustion from being counted against the player because as far as I can tell it's good to do Combustion normally rather than hard cast a Pyroblast in the middle of one, and I set the minimum suggestion to 10% refreshes before popping up based on vituscze's 7.2% number. For expires I set the suggestion limit to start when you have 1 expired buff since that ideally wouldn't be happening.

Sharrq:
Even using the damage number wouldnt be 100% accurate. 

There are plenty of ways in which a pyroblast can do more or less than whatever baseline you set. Gear and stats being the biggest.

Additionally, bosses and mobs that take extra damage, mobs that have less health remaining than the amount of damage the pyroblast did (i.e. the pyroblast killed them), etc. 

Hard Coding values and saying "anything above this is a buffed hard cast and anything below it is not" is not a very good place to be.

When i did finally get this to work I did so by looking at begincast events and cast events and comparing their timestamps.

So for example, spells that have a cast time have a begincast event (when you start casting it) and a cast event (when the spell is actually cast). What I was doing was saying "If the timestamp of begincast is within 100ms of the timestamp on cast, then its an instant cast so ignore it ... if its not, then it was hard casted". Then separately i was using a boolean variable to track when the buff was up or down and discarding any casts while the buff was down. What remained was the buffed hard casts.

It worked great for Nighthold and ToS, then I went and tried it on an Antorus fight and noticed that the instant cast Pyroblasts didnt have begincast events anymore ... and it completely broke everything i wrote. So i was trying to come up with a new way to determine whether it was a instant cast or not so i could fix it, and then i got busy with other stuff and just havent circled back to it. Ill see if I still have the code i wrote (i really hope i do).

Sharrq:
Here is what I had previously, along with some stuff i tried. It doesnt seem to quite be working yet. The console is claiming that most fights had more procs than they actually did (when you compare the console vs the WCL Log)

https://pastebin.com/230LpLNd

Sharrq:
@fuzzys 
So while i was sending you that code that i wrote previously but couldnt get to work, I had an idea on how i could fix it, so i tried it and it seems to have worked. I spent some additional time bug testing it on a bunch of fights and everything looks to match what is in WCL.

So i have an option for you. 
Would you prefer that I submit my code as a separate PR and disregard this one?
or would you prefer that I send you my code and you can look at it and add onto it/adjust it and submit it?

To give you an idea, mine counts the damage as well as wasted procs, total procs, etc. It just has no suggestions tied to it.

So if you were primarily interested in the suggestions & the damage bonus, then i think it would be best if i submit mine and get it approved, then you can make a new PR to add the suggestions and analysis onto mine (since mine has none of that). But if you would prefer to make your own or think you can make mine better, i can send you my working code.

Let me know what you would prefer.

kfinch:
@fuzzys and @Sharrq let me know when this PR is in a state where you guys are satisfied that it works as intended

fuzzys:
@Sharrq

You understand this a lot better than I do so I trust your version a whole lot more than mine. Submit away! And if nobody else is up for adding suggestions then maybe I'll look into doing them sometime.

Sharrq:
Ok sounds good. I submitted it.

kfinch:
As other PR went in, I'm closing this one.

PR #1131, opened 2018-01-03
[Windwalker] Fixing Windwalker GCD and adding Channeling

Juko8:
Added Gol'ganneths Vitality proc to Casts that aren't casts for this.


src/Parser/Monk/Windwalker/Modules/Core/Channeling.js
```diff
@@ -0,0 +1,72 @@
...
+    if (event.ability.guid !== SPELLS.FISTS_OF_FURY_CAST.id) {
+      return;
+    }
+    if (!this.isChannelingSpell(SPELLS.FISTS_OF_FURY_CAST.id)) {
+      // This may be true if we did the event-order fix in begincast/cast and it was already ended there.
+      return;
+    }
+    const fistsTime = event.timestamp - this.fistsCastStart;
+    if (fistsTime < 1000)  {
+      return;
```
MartijnHols: I'm not a fan of "hacks" like this as there are no guarantees it doesn't actually make things worse. The only way I can live with this is with two additions:

1. in development (`process.env.NODE_ENV === 'development'`) throw an error
2. always do a `console.error` when this occurs, explaining why it's bad when it happens

Alternatively if you think this will rarely occur you can remove this if statement.

src/Parser/Monk/Windwalker/Modules/Features/AlwaysBeCasting.js
```diff
@@ -7,6 +7,9 @@ import { formatPercentage } from 'common/format';
 import { STATISTIC_ORDER } from 'Main/StatisticBox';
 
 class AlwaysBeCasting extends CoreAlwaysBeCasting {
+    // Windwalker GCD is 1 second by default and static in almost all cases, 750 is lowest recorded GCD
+  static BASE_GCD = 1000;
+  static MINIMUM_GCD = 1000;
```
MartijnHols: I recommend looking at how I configured this recently for Brewmaster; there's no reason to configure a minimum equal to base if you're setting `STATIC_GCD_ABILITIES` and it will actually bug out spells that *are* affected by Haste.

src/Parser/Monk/Windwalker/Modules/Core/Channeling.js
Juko8: Yeah, it was something i had used to workaround, just forgot to remove it after fixing gol'ganneths being seen as a cast. Thanks for noticing.

src/Parser/Monk/Windwalker/Modules/Features/AlwaysBeCasting.js
Juko8: I did try it out with minimum GCD set to 750, but all abilities get hasted GCDs then, regardless if they're being set with `STATIC_GCD_ABILITIES` or not

MartijnHols: Ohh, it was working for brewmaster...

PR #1138, opened 2018-01-05
[Fire Mage] Support for Marquee Bindings of the Sun King

Sharrq:
Adds support for Marquee Bindings. Counts the damage, proc counts, wasted procs, used procs, and the average bonus damage.

Use this PR instead of the one fuzzys submitted. He will look at adding suggestions and anything that is missing after this gets approved.


src/Parser/Mage/Fire/Normalizers/KaelthasUltimateAbility.js
```diff
@@ -0,0 +1,37 @@
+import SPELLS from 'common/SPELLS';
+
+import EventsNormalizer from 'Parser/Core/EventsNormalizer';
+
+class KaelthasUltimateAbility extends EventsNormalizer {
+  /**
+   * @param {Array} events
+   * @returns {Array}
+   */
+  normalize(events) {
```
kfinch: Please include in the documentation comment above what the issue is with unmodified event ordering, and what this Normalizer does to remedy the issue.

It's often difficult to discern exactly what a Normalizer is doing (and why), which is why I'm insisting on commenting in this case.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,151 @@
...
+class MarqueeBindingsOfTheSunKing extends Analyzer {
+
+  static dependencies = {
+		combatants: Combatants,
+	};
+
+  damage = 0;
+  beginCastTimestamp = 0;
+  castTimestamp = 0;
+  hasBindingsBuff = false;
```
kfinch: So far as I can tell, this flag is true iff the player has the bindings buff... so why not remove this flag entirely and just use `this.combatants.selected.hasBuff(...` instead?

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,151 @@
...
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.PYROBLAST.id || this.beginCastFound === false) {
+      return;
+    }
+    this.beginCastFound = false;
+    debug && console.log("Pyroblast Casted @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
+    this.castTimestamp = event.timestamp;
+    const castTime = this.castTimestamp - this.beginCastTimestamp;
+    if (castTime >= CAST_BUFFER && this.hasBindingsBuff === true) {
```
kfinch: I'm not sure I understand what this is about... I'm guessing this is to discern a hardcast pyro from an instant pyro, but would an instant pyro even have a `begincast` event? If Hot Streak Pyro does indeed have a begincast, please add a brief comment here explaining this to a future dev looking at this code.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,151 @@
...
+  get avgBonusDamage() {
+    return this.damage / this.usedProcs;
+  }
+
+  on_finished() {
+    if (this.combatants.selected.hasBuff(SPELLS.KAELTHAS_ULTIMATE_ABILITY.id)) {
+      const adjustedFightEnding = this.owner.currentTimestamp - 5000;
+      if (this.buffAppliedTimestamp < adjustedFightEnding) {
+        this.wastedProcs += 1;
+        debug && console.log("Fight Ended with Unused Proc @ " + formatMilliseconds(this.owner.currentTimestamp - this.owner.fight.start_time));
```
kfinch: This seems a little harsh, and possibly confusing that this counts as wasted... Still I guess it's arguably the right thing to do.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
```diff
@@ -0,0 +1,151 @@
...
+      item: ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING,
+      result: (
+        <dfn data-tip={`${this.totalProcs} Total Procs
+        <ul>
+          <li>${this.usedProcs} Procs Used</li>
+          <li>${this.wastedProcs} Procs Wasted</li>
+          <li>${formatNumber(this.avgBonusDamage)} Average Bonus Damage</li>
+        </ul>
+        `}>
+          {this.owner.formatItemDamageDone(this.damage)}
```
kfinch: It's very misleading to include the bonus damage from proc pyros as being the damage boost due to the legendary, because without the legendary players do not hardcast Pyroblast. The actual damage boost due to the bindings is the "damage increase vs. replacement", e.g. the actual damage from casting a proc pyro *minus the damage the player would have done if they'd used those 3.5 seconds to do their normal rotation*.

Furthermore, this doesn't account for Ignite.

Given the inherit inaccuracy of this number, I think it shouldn't go live displayed as is. I'm open to brainstorming possible solutions.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
Sharrq: So when I wrote this initially, Hot Streak Pyroblast did have a begincast that was the same timestamp as the cast event. But now hot streak pyro does not have a begincast event. Nonetheless, the code still worked so i didnt change it. 

Ill take another look at it and see if i can clean it up

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
Sharrq: Yea i gave 5 seconds at the end of the fight so you could reasonably get the buff, realize you have it, and cast it before the fight ends ... but i might increase that.

Essentially that disregards any procs that were applied within the last x seconds of the fight.

src/Parser/Mage/Fire/Modules/Items/MarqueeBindingsOfTheSunKing.js
Sharrq: i mean, its not really possible to know how much more damage it did vs casting Fireballs instead ... so im not exactly sure how i would make it more accurate in that sense. 

Maybe i could figure out how to account for Ignite, but idk if i could reasonably figure that out either.

kfinch: If "it's not really possible to calculate", then don't display a damage boost number. Show PPM or something instead.

Basically, it's better to show nothing than to show a number that you know is wrong.

Sharrq:
Made suggested changes

PR #1141, opened 2018-01-05
[Balance] Added most legendaries

Gebuz:
- Added wasted cooldown reduction to IFE
- Added Soul of the Archdruid
- Added Lady and the Child
- Added Oneth's Intuition
- Added Promise of Elune

### DIFF
--- a/src/Parser/Druid/Balance/CHANGELOG.js
+++ b/src/Parser/Druid/Balance/CHANGELOG.js
@@ -10,7 +10,7 @@ import SpellLink from 'common/SpellLink';
 export default [
   { 
     date: new Date('2018-1-5'),
-    changes: <Wrapper>Added most legendaries.</Wrapper>,
+    changes: <Wrapper>Added the following legendaries: <ItemLink id={ITEMS.IMPECCABLE_FEL_ESSENCE.id} icon />, <ItemLink id={ITEMS.ONETHS_INTUITION.id} icon />, <ItemLink id={ITEMS.LADY_AND_THE_CHILD.id} icon />, <ItemLink id={ITEMS.PROMISE_OF_ELUNE.id} icon />, and <ItemLink id={ITEMS.SOUL_OF_THE_ARCHDRUID.id} icon /> and updated <ItemLink id={ITEMS.THE_EMERALD_DREAMCATCHER.id} icon />.</Wrapper>,
     contributors: [Gebuz],
   },
   { 
--- a/src/Parser/Druid/Balance/CHANGELOG.js
+++ b/src/Parser/Druid/Balance/CHANGELOG.js
@@ -9,6 +9,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   { 
+    date: new Date('2018-1-5'),
+    changes: <Wrapper>Added most legendaries.</Wrapper>,
+    contributors: [Gebuz],
+  },
+  { 
     date: new Date('2018-1-2'),
     changes: <Wrapper>Added Astral Power usage tab.</Wrapper>,
     contributors: [Gebuz],
--- a/src/Parser/Druid/Balance/Modules/Items/LadyAndTheChild.js
+++ b/src/Parser/Druid/Balance/Modules/Items/LadyAndTheChild.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import SpellLink from 'common/SpellLink';
+import { formatNumber , formatPercentage } from 'common/format';
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Wrapper from 'common/Wrapper';
+
+class LadyAndTheChild extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  moonfireCasts = 0;
+  moonfireHits = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasShoulder(ITEMS.LADY_AND_THE_CHILD.id);
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.MOONFIRE_BEAR.id || event.tick === true) {
+      return;
+    }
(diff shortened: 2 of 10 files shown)


Gebuz:
I forgot to mention, this is what Oneth's Intuition currently looks like:
Image
I can't make more space between the icons and they are pretty small, I'm guessing the items aint supposed to have icons, but I'd like to use them if possible. If anybody has a suggestion for a better look  or how to format it better it would be appreciated :)

src/Parser/Druid/Balance/Modules/Items/LadyAndTheChild.js
```diff
@@ -0,0 +1,51 @@
...
+  get percentTwoHits() {
+    return (this.moonfireHits - this.moonfireCasts) / this.moonfireCasts;
+  }
+
+  item() {
+    return {
+      item: ITEMS.LADY_AND_THE_CHILD,
+      result: (
+        <dfn data-tip={`You hit ${this.moonfireHits} times with ${this.moonfireCasts} casts.`}>
+          <Wrapper>{formatPercentage(this.percentTwoHits)}% of your <SpellLink id={SPELLS.MOONFIRE_BEAR.id}/> casts hit two targets.</Wrapper>
```
kfinch: While you don't have to implement this now, it would be nice to also see the damage gained due to LatC's moonfire damage buff.

src/Parser/Druid/Balance/Modules/Items/OnethsIntuition.js
```diff
@@ -0,0 +1,41 @@
...
+    if (spellId === SPELLS.ONETHS_OVERCONFIDENCE.id) {
+      this.freeStarfallProcs += 1;
+    }
+  }
+
+  item() {
+    return {
+      item: ITEMS.ONETHS_INTUITION,
+      result: (
+          <Wrapper>{this.freeStarsurgeProcs} <SpellIcon id={SPELLS.ONETHS_INTUITION.id}/> {this.freeStarfallProcs} <SpellIcon id={SPELLS.ONETHS_OVERCONFIDENCE.id}/></Wrapper>
```
kfinch: For better display, you could try something like the stats box in Frost Mage's Winter's Chill module (https://github.com/WoWAnalyzer/WoWAnalyzer/blob/b568ef41a82cc30e4c96f129f44b53e3af63a971/src/Parser/Mage/Frost/Modules/Features/WintersChill.js#L171)

Also, it would be nice to include a tooltip to more clearly explain to the user what the symbols mean, as they might not recognize them by icon.

Finally, as "extra credit", it would also be nice to show the player how much Astral Power this leggo saved them.

src/Parser/Druid/Balance/Modules/Items/PromiseOfElune.js
```diff
@@ -0,0 +1,57 @@
...
+  on_byPlayer_damage(event) {
+    if (event.ability.guid === SPELLS.SOLAR_WRATH_MOONKIN.id) {
+      this.solarWrathDamage += this.damageContribution(event, DAMAGE_MODIFIER);
+    }
+    if (event.ability.guid === SPELLS.LUNAR_STRIKE.id) {
+      this.lunarStrikeDamage += this.damageContribution(event, DAMAGE_MODIFIER);
+    }     
+  }
+
+  damageContribution(event, mod) {
```
kfinch: This function is already provided in core @ https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/calculateEffectiveDamage.js

No need to reimplement here.

src/Parser/Druid/Balance/CHANGELOG.js
```diff
@@ -8,6 +8,11 @@ import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 
 export default [
+  { 
+    date: new Date('2018-1-5'),
+    changes: <Wrapper>Added most legendaries.</Wrapper>,
```
kfinch: List which ones you added.

src/Parser/Druid/Balance/Modules/Items/LadyAndTheChild.js
Gebuz: In a lot of situations the biggest DMG benefit is the saved GCD from not having to cast moonfire on the other target it hits so it isn't really possible to estimate the DPS gain. Showing only the increase from moonfire would be misleading in my opinion, that's why I left it out.

src/Parser/Druid/Balance/Modules/Items/OnethsIntuition.js
Gebuz: It would definitely look nicer as a stat box, I just avoided it since it's an item. Probably makes more sense as a stat box and then throwing the procs/casts as a percent on the item.  I don't think it should display Astral Power saved as you usually wouldn't be casting the opposite spender without the proc, meaning the proc is worth less than the AsP it saves.

src/Parser/Druid/Balance/Modules/Items/LadyAndTheChild.js
Gebuz: It's pretty much impossible to estimate the gain because a big part of it is the saved GCDs in an AOE situation. On top of that Moonfire becomes the AOE filler instead of Solar Wrath/Lunar Strike when you have it equipped, so it would be inflated by that.

kfinch: I was suggesting more as part of the tooltip include an "also gained X% damage from buffed moonfire" ... but I can see actual rotation differences complicating accuracy.

src/Parser/Druid/Balance/Modules/Items/OnethsIntuition.js
kfinch: No, keep it in Items... I guess you would need to modify more about that code in order to get it to display in Item format what you want. Really, I think the small icons are ok as long as you include a tooltip saying what's going on.

As for the AsP bit, it's a good point I wasn't thinking about how you might Starfall when you otherwise wouldn't.

Gebuz:
Oneth's Intuition now has a tooltip:
Image

kfinch:
Thanks!

PR #1142, opened 2018-01-06
[Balance] Tier 21

Gebuz:
- 2p: damage contribution
- 4p: damage contribition and uptime


Gebuz:
Good Idea, I'll take a look at that over the weekend.

PR #1144, opened 2018-01-06
[Fire Mage] Enhancements and Additions for Hot Streak

Sharrq:
Hot Streak has been expanded to once again check for direct damage crits while Hot Streak is active (but now also disregards crits that happen when Pyromaniac procs and immediately re-activated Hot Streak).

Also added checks to make sure each use of hot streak has a hard cast ability like Fireball, Scorch, or Pyroblast (only if you have a bracer proc) immediately before it to increase the chance of getting a Heating Up or Hot Streak ... unless Combustion is active.

### DIFF
--- a/src/Parser/Mage/Fire/Modules/Features/HotStreak.js
+++ b/src/Parser/Mage/Fire/Modules/Features/HotStreak.js
@@ -56,7 +56,7 @@ class HotStreak extends Analyzer {
     if (!HOT_STREAK_CONTRIBUTORS.includes(spellId) || !this.combatants.selected.hasBuff(SPELLS.HOT_STREAK.id) || event.hitType !== HIT_TYPES.CRIT) {
       return;
     }
-    //If Pyromaniac caused the player to immediately get a new hot streak after spending one, then dont count the damage crits that were cast before Pyromaniac Proc's since the user cant do anything to prevent this
+    //If Pyromaniac caused the player to immediately get a new hot streak after spending one, then dont count the damage crits that were cast before Pyromaniac Proc's since the user cant do anything to prevent this.
     if ((spellId === SPELLS.FIREBALL.id || spellId === SPELLS.SCORCH.id || spellId === SPELLS.PYROBLAST.id) && this.pyromaniacProc) {
       debug && console.log("Wasted Crit Ignored @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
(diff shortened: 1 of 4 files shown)


src/Parser/Mage/Fire/Modules/Features/HotStreak.js
```diff
@@ -30,115 +32,182 @@ class HotStreak extends Analyzer {
...
           .icon(SPELLS.HOT_STREAK.icon)
-          .actual(`${formatPercentage(expiredProcsPercent)}% expired`)
-          .recommended(`Letting none expire is recommended`)
-          .regular(.00).major(.05);
+          .actual(`${formatPercentage(this.expiredProcsPercent)}% expired`)
+          .recommended(`<${formatPercentage(recommended)}% is recommended`);
+      });
+      when(this.wastedCritsThresholds)
+        .addSuggestion((suggest, actual, recommended) => {
+          return suggest(<span>You crit with {formatNumber(this.wastedCrits)} direct damage abilities while <SpellLink id={SPELLS.HOT_STREAK.id} /> was active. This is a waste since those crits could have contibuted towards your next Hot Streak. Try to use your procs as soon as possible to avoid this.</span>)
```
kfinch: `<Wrapper>` instead of `<span>` please.

src/Parser/Mage/Fire/Modules/Features/HotStreak.js
```diff
@@ -30,115 +32,182 @@ class HotStreak extends Analyzer {
...
+        average: 0.10,
+        major: 0.20,
+      },
+      style: 'percentage',
+    };
+  }
+
+  get wastedCritsThresholds() {
+    return {
+      actual: this.wastedCrits,
```
kfinch: Don't make suggestions based on absolute number... this looks like an appropriate time to use wasted crits per minute.

Sharrq:
I did test it, ill try to test it a little more in depth but on the 3-4 fights i tried it on it worked as it should. 

I should have time tonight so ill make the other suggested changes and get a little more testing in.

kfinch:
I think I get largely what is going on, fix the `<span>`s and I can merge

Sharrq:
I actually found another issue with it, so I'm trying to fix that first

Sharrq:
I fixed the issue I found with Scorch Damage Events happening at the exact same timestamp as Hot Streak being removed

Also added a check for Firestarter.

PR #1155, opened 2018-01-08
[Balance] Tier 20, L90 Astral Power gain, Updated suggestion Thresholds.

Gebuz:
- Updated the suggestion thresholds to the new format aswell as updating some of the values.
- Added Tier 20 2p & 4p.
- Added a statistic box that shows the potential Astral Power Generation from all the L90 talents:
image
Feedback on the layout for the stat box is very welcome, I was thinking about possibly displaying the Actual value for the encounter under the big icon on the left but I have no idea how to add it in that position. I also thought about having it as a line of icons with the values below instead of having the value to the right of the icon. (but again, I have no idea how to do that :D)
I will likely also add a suggestion if the value of the talent chosen is not the highest but that is a thing for the future, as there are some complications there I have to consider.


src/Parser/Druid/Balance/Modules/Features/AlwaysBeCasting.js
```diff
@@ -17,16 +17,13 @@ class AlwaysBeCasting extends CoreAlwaysBeCasting {
...
-      .addSuggestion((suggest, actual, recommended) => {
+  suggestions(when) {
+    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
         return suggest(<span> Your downtime can be improved. Try to Always Be Casting (ABC)...</span>)
           .icon('spell_mage_altertime')
           .actual(`${formatPercentage(actual)}% downtime`)
-          .recommended(`${Math.round(formatPercentage(recommended))}% or less is recommended`)
-          .regular(0.075).major(0.1);
+          .recommended(`${Math.round(formatPercentage(recommended))}% or less is recommended`);
       });
```
Pewtro: Why are we rounding here, wouldn't formatPercentage be enough?

src/Parser/Druid/Balance/Modules/Features/UnempoweredLs.js
```diff
@@ -76,16 +77,29 @@ class UnempoweredLS extends Analyzer {
...
-          .actual(`${formatNumber(actual)} Unempowered LS per minute`)
-          .recommended(`${recommended} Unempowered LS that hits less than 4 targets are recomended`)
-          .regular(recommended + 2).major(recommended + 4);
-      });
+    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
+      return suggest(<Wrapper>You casted {this.suboptUmempLS} unempowered and non instant cast <SpellLink id={SPELLS.LUNAR_STRIKE.id} /> that hit less than 4 targets. Always prioritize Solar Wrath as a filler if there are less than 4 targets.</Wrapper>)
+        .icon(SPELLS.LUNAR_STRIKE.icon)
+        .actual(`${formatNumber(actual)} Unempowered LS per minute`)
+        .recommended(`${recommended} Unempowered LS are recomended`);
+    });
```
Pewtro: ``recomended``should be ``recommended``

kfinch:
Regarding the display of the L90 talent stats box, I don't think showing the potential AsP gain from each option really makes sense. I'd prefer to see the AsP gain from just the chosen talent, and then maybe the "what could have been" numbers in the tooltip.

Also, this seeems to be prime territory for a suggestion... is there a reason you haven't implemented a suggestion for if the player picks a L90 talent but could have gotten more from another option?

Let me know what you think of these concerns.

Gebuz:
Reason I want to show the potential AsP gain from the other talents is mostly just because it's a fun stat to have and it can be useful to determine roughly how close the talents are in value on a given encounter. Maybe it's better to hide the "what could have been" info in the tooltip as you say, and then just have the number for the actual talent chosen being displayed in the stat box. If it's confusing what the numbers actually represent or if it isn't relevant enough to be shown I can hide it. I'm obviously biased as I made it so I know exactly what the numbers mean and it isn't an issue for me.

The talents aren't completely comparable one to one, as shooting stars also deals a tiny amount of damage, Astral Communion doesn't actually fit very well into the opening rotation as it will usually cause you to overcap either empowerments or moon spells, and Blessing of Elune and Blessing of An'she can technically be switched mid encounter for a minor benefit, it's just extremely rare that people ever do it as the gain is extremely small compared to the amount of micro management it requires.

This is why I haven't added suggestions yet, but I suppose I could just make the threshold relatively generous and note in the suggestion that it's more of a "consider using X instead, as it could potentially have been better".

kfinch:
I completely understand the value of seeing the potential strength of other options, I just think it's odd to put all 4 together without even any indication of which you took. Stat boxes are supposed to be for what actually happened, not what could have happened.

I think it's extra weird that you even show the average AsP gain for the talent you did pick as opposed to the actual AsP gain.

I believe the potential gain should be best in the tooltip of the stats box, along with perhaps a little bit on how those potential gains were calculated.

A generous buffer for suggestion is a good idea, it's up to you how you want to implement that.

Gebuz:
This was my original idea:
l90
Anyway you're probably right that showing potential doesn't belong there, so I'll move the potential gains to the tooltip and make the stat box only display the actual value + add suggestions based on both actual/expected (for Blessing of the Ancients & Astral Communion, doesn't make sense for Shooting Stars since it's RNG) and actual/maxPotential (expected/maxPotential for Shooting Stars instead) next time I work on it.

kfinch:
Yeah that does look better, but putting it in a nice tooltip probably best.

PR #1156, opened 2018-01-08
[BeastMastery] Aspect of the wild and Dire Frenzy updates

Pewtro:
Noticed some bugs with Dire Frenzy and its uptime where removebuff events aren't in the combatlog for whatever reason, this accounts for that. 

Also adjust what we deem a good cast of Aspect of the wild.


Pewtro:
Got working on a golganneth update while adding a bit to this as I noticed the buffs for the regular and empowered procs, so they made their way into the same commit here - woops. 

Couldn't get to use formatItemDamageDone, so I just went with a flat damage done + DPS of the individual procs.

image

Issue #1059, opened 2017-12-25
Update deprecated changelogs to new changelog format


MartijnHols:
Old format:

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/23a0f3401215e67861c2b792f5f21f7ceea536fc/src/Parser/Warlock/Destruction/CHANGELOG.js

New format:

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/23a0f3401215e67861c2b792f5f21f7ceea536fc/src/Parser/Paladin/Holy/CHANGELOG.js

Notable differences:

* Date format MUST be y-m-d
* Contributors must be taken from the `MAINTAINERS` file
*  Wowhead spell/item links should be replaced with their respectable ItemLink / SpellLink components (preferably with the `icon` property set to true).
* It's preferred to make all spell/item links actual links instead of textual versions (as seen in the old format example)
* Entries should not include the spec name 
* `class` in the old format has to be `className` in the new format
* I used regular expressions to quickly convert all the entries in the changelogs I converted to the new format, writing a good regular expression can save you most of the work

Gebuz:
I believe the last missing ones were updated in #1535

joshinat0r:
```
src/Parser/Shaman/Enhancement/CHANGELOG.js
src/Parser/Shaman/Elemental/CHANGELOG.js
```

Those files might need an update since both files had entrys without contributors (I just coppied the last available contributor).
Probably something that @fasib  and @Nighteyez07 need to check out and update.

PR #1157, opened 2018-01-08
[WIP] Update changelogs

avilene:
For issue #1059


kfinch (changes requested):
Don't put `<Wrapper>` around everything, it's only actually needed if you wish to include elements like a `SpellLink`. Changes that are text-only should be surrounded by quotation marks.

fyruna:
I might start from scratch and just copy some of the changes. :D

fyruna:
Closing in favor of #1185

PR #1158, opened 2018-01-09
Maelstrom-Chart and partly Focus-Chart

HawkCorrigan:
Started out as just putting ABC into a submodule, which was already done tho, so it became just deleting old code.

Stole @Pewtro's implementation of the focus chart. Fixed a bug where it showed double the intended value for Aspect of the wild, put all charts in submodules.

Bug was caused by all generating abilites which were caused by and targetting the same player getting registered twice, so deleting the on_by_Player_energize should solve the issue.

New Maelstrom-Tab gets used by both Elemental and Enhancement and works for both, so seperate implementations became obsolete.
  
Sorry for that PR becoming quite messy.


Gebuz:
Is there anything special about focus or maelstrom or could this be made a core module for all resources? It would avoid a lot of code duplication and make it easier to implement new resources in the future. I'm just wondering since when I made the Astral Power tab for balance druid I ended up just moving the rogue resourcetracker modules to core and then it didn't take much more than 10 lines to implement it for balance.

HawkCorrigan:
Ye that was kinda the plan in the long run
I basically just wanted to play around with the tabs first, so I only looked into these two.

HawkCorrigan:
So rogue tracker is already in core?

Gebuz:
yeah, it doesn't have a chart though, just tracker, tab and suggestion.

Pewtro:
oh yeah okay that makes sense for the Aspect of the wild stuff monkaS
Thanks for moving it into a standalone module tho 😃 
 
however I think running the risk of classResources being different for specs might provide an issue that could prevent this from being core

src/Parser/Hunter/Shared/Modules/Features/FocusChart/FocusTracker.js
```diff
@@ -81,7 +77,7 @@ class FocusTracker extends Analyzer {
   }
 
   checkActiveWaste(event) {
-    if ((event.sourceID === this.owner.player.id || event.targetID === this.owner.player.id) && event.classResources && event.classResources[0].type === RESOURCE_TYPES.FOCUS) {
+    if ((event.sourceID === this.owner.player.id || event.targetID === this.owner.player.id) && event.classResources && event.classResources[0].type === RESOURCE_TYPES.FOCUS.id) {
       this.tracker++;
       if (this.generatorCasts[event.ability.guid]) {
```
Pewtro: Why does this line appear changed into this? It's already like this on live 🤔

HawkCorrigan: ye i think i fetched it a second before you fixed it, no idea why it gets highlighted tho

HawkCorrigan:
I might look into making it a core module next, i just expect some issues with passive ressource generation.

PR should be fine as is right now though.

Pewtro:
With a mental note to take a look at making a core tab in the future possibly

PR #1161, opened 2018-01-10
[Resto Druid] Added preliminary HoT Tracker, enabling several module updates

### LINKED ISSUE #1160: [BUG] Tearstone of Elune triggered ~25 times, only one was caught.
https://wowanalyzer.com/report/yWB68j2PqaZkpKMV/42/53

kfinch:
The meat of this PR is a new 'HotTracker' module which tracks attribution of application, boosts, and extensions of HoTs. This allows several modules to be able to report attributable healing with much better accuracy.

Currently the 'extension' part is NYI, I intend to add that in the future. For now, I want to run this design by everyone, because I'm uncomfortable with how complex and monolithic the HotTracker module is. Any design suggestions would be appreciated.

I've implemented updated modules for PotA, Tearstone (fixes #1160), and T21, and I've added a new module implementing 4pc T19 using this framework.

I also added a couple events normalizers that make events go in an easier to handle order. Details are in their comments.

-----------------------

A little more about the design of HotTracker, which might not be fully apparent from looking it over: HotTracker tracks all HoTs applied by the player, and at the moment of application calls `_getAttributions` to go through some logic to figure out where the HoT came from (eg a hardcast or a Tearstone proc or a PotA proc etc etc). When an attribution is made, an accumulator object is passed to the HoT, and stats about that HoT are updated as the HoT heals. This means at any time we can access some abilities "attributable healing" through its accumulator. Handling extensions and boosts will use similar logic.


src/Parser/Druid/Restoration/Modules/Core/HotTracking/HotTracker.js
```diff
@@ -0,0 +1,205 @@
...
+    hot.boosts = [];
+  }
+
+  on_byPlayer_removebuff(event) {
+    if(event.ability.guid === SPELLS.POWER_OF_THE_ARCHDRUID_BUFF.id) {
+      this.potaFallTimestamp = event.timestamp; // FIXME temp until figure out hasBuff problem
+    }
+
+    const spellId = event.ability.guid;
+    const target = this._validateAndGetTarget(event);
```
Pewtro: why don't you just move ``const spellId = event.ability.guid;`` that initial check so that could be if ``spellID ===`` aswell?

kfinch:
I was making an effort to make this as readable as possible, but I guess I can only go so far given the complexity of what I'm doing. I made several updates since you last looked, but not sure if it will make things more clear.

kfinch:
I actually don't want to merge this just yet, There's a few more tweaks I'll put in first, mostly formalizing the interactions between the HotTracker and the Attributors

kfinch:
Alright, putting this in...

PR #1170, opened 2018-01-12
[BrM Monk] Basic Purifying Brew Stat

emallson:
Shows the amount, mean, min, and max amount of damage purified with Purifying Brew. Also tracks the number of times you purified w/o being in heavy (red) stagger. 

Generally speaking not purifying when below heavy stagger is a good rule of thumb, but I'm not sure where to put the suggestion threshold yet (so there isn't an associated suggestion).

### DIFF
--- a/src/Parser/Monk/Brewmaster/Modules/Spells/PurifyingBrew.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Spells/PurifyingBrew.js
@@ -7,8 +7,6 @@ import Combatants from 'Parser/Core/Modules/Combatants';
 import Analyzer from 'Parser/Core/Analyzer';
 import SharedBrews from '../Core/SharedBrews';
 
-const HEAVY_STAGGER_LEVEL = 2/3;
-
 class PurifyingBrew extends Analyzer {
   static dependencies = {
     brews: SharedBrews,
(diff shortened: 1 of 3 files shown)


src/Parser/Monk/Brewmaster/Modules/Spells/PurifyingBrew.js
```diff
@@ -1,17 +1,94 @@
...
+  }
+
   on_byPlayer_cast(event) {
     if(event.ability.guid === SPELLS.PURIFYING_BREW.id) {
       this.brews.consumeCharge(event);
     }
   }
+
+  get meanPurify() {
+    if(this.purifyAmounts.length === 0) {
```
MartijnHols: Using `this.totalPurifies` here would reduce duplicate code

src/Parser/Monk/Brewmaster/Modules/Spells/PurifyingBrew.js
```diff
@@ -1,17 +1,94 @@
...
+
+    return this.totalPurified / this.totalPurifies;
+  }
+
+  get minPurify() {
+    if(this.purifyAmounts.length === 0) {
+      return 0;
+    }
+
+    return this.purifyAmounts.reduce((prev, cur) => (prev < cur) ? prev : cur, Infinity);
```
MartijnHols: Could be simplified; Math.min(min, current)

src/Parser/Monk/Brewmaster/Modules/Spells/PurifyingBrew.js
```diff
@@ -1,17 +1,94 @@
...
+  }
+
+  get maxPurify() {
+    return this.purifyAmounts.reduce((prev, cur) => (prev > cur) ? prev : cur, 0);
```
MartijnHols: Could be simplified: Math.max(max, current)

PR #1175, opened 2018-01-12
[Windwalker] Tiny correction in Abilities

Juko8:


src/Parser/Monk/Windwalker/Modules/Features/Abilities.js
```diff
@@ -132,7 +132,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.CRACKLING_JADE_LIGHTNING,
         category: Abilities.SPELL_CATEGORIES.OTHERS,
-        enabled: combatant.hasTalent(ITEMS.THE_EMPERORS_CAPACITOR.id),
+        enabled: combatant.hasChest(ITEMS.THE_EMPERORS_CAPACITOR.id),
```
MartijnHols: Wasn't Crackling Jade Lightning always available to players? If so it should not have an `enabled` toggle.

Juko8: It's always there, but not relevant to know amount of casts without the chest, since they do close to 0 damage then

MartijnHols: The abilities config is used by a lot more than just showing the total amount of casts in cast efficiency. For all modules relying on Abilities to work optimally, it needs to be able to see all spells that are available to a player (listed in the spellbook), so if the spell is available at all times `enabled` should be omitted.

Juko8: Alright, removing the `enabled` instead then

PR #1177, opened 2018-01-12
Fix current Haste percentage interaction with Haste rating buff changes (this improves ABC and spell cooldown accuracy).

MartijnHols:
Also improved the code for stacked haste buffs.

This fixes SpellUsable errors for haste rating trinkets.


src/Parser/Core/Modules/Haste.js
```diff
@@ -179,29 +185,29 @@ class Haste extends Analyzer {
...
+    const fabricatedEvent = {
       timestamp: event ? event.timestamp : this.owner.currentTimestamp,
       type: 'changehaste',
       sourceID: event ? event.sourceID : this.owner.playerId,
       targetID: this.owner.playerId,
       reason: event,
       oldHaste,
       newHaste,
-      hasteGain,
-      hasteLoss,
```
kfinch: Just to be completely sure, did you do a search to check that these fields aren't used anywhere?

MartijnHols: Yes

PR #1178, opened 2018-01-12
[Retribution Paladin] Refactoring

CollCrom:
Mostly changed a bunch of spans to Wrappers. I fixed a few suggestion tooltips and also removed some duplicate spells from the ABC module.


src/Parser/Core/Modules/Haste.js
```diff
@@ -41,7 +41,6 @@ class Haste extends Analyzer {
     [209166]: 0.3, // DEBUFF - Fast Time from Elisande
     [209165]: -0.3, // DEBUFF - Slow Time from Elisande
     // [208944]: -Infinity, // DEBUFF - Time Stop from Elisande
-    [SPELLS.BONE_SHIELD.id]: 0.1, // Blood BK haste buff from maintaining boneshield
```
MartijnHols: Why did you remove this?

CollCrom: It was repeated. I don't have line numbers on my phone but it's right below Icy Veins

Issue #1171, opened 2018-01-12
[Fire Mage] Heating Up Bug


Sharrq:
The Heating Up Module that counts how many Fire Blasts you cast without Heating Up (i.e. they were used without Heating Up or Hot Streak) is miscounting during Firestarter and Combustion.

Its essentially doing the following

Applied Hot Streak
Cast Pyroblast
Removed Hot Streak
Cast Fire Blast
Damage Fire Blast (without Heating Up)
Applied Heating Up
Damage Pyroblast
Applied Hot Streak

Playing this way is correct since its a guarantee that the Pyroblast will Crit. Need to adjust the logic to disregard this during Firestarter and Combustion, or to get more detailed, to disregard it if a Pyroblast was cast x ms before that.

Sharrq:
Yes it can be closed, i didnt know about the fixes thing.

PR #1180, opened 2018-01-13
[Fire Mage] Fixed Bug with Heating Up

Sharrq:
The Heating Up Module was not taking into account the guaranteed crits during Firestarter and Combustion ... Per #1171


src/Parser/Mage/Fire/Modules/Features/HeatingUp.js
```diff
@@ -10,7 +10,7 @@ import Analyzer from 'Parser/Core/Analyzer';
 import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
 import EnemyInstances, { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';
 
-const debug = false;
+const debug = true;
```
CollCrom: Just turn this off and it looks good to me

PR #1181, opened 2018-01-14
[Fire Mage] Cinderstorm Support

Sharrq:
Added support for Cinderstorm Talent


src/Parser/Mage/Fire/Modules/Features/Cinderstorm.js
```diff
@@ -0,0 +1,107 @@
...
+  get missedCinders() {
+    return this.totalCinders - this.totalHits;
+  }
+
+  get missedCindersPercent() {
+    return this.missedCinders / this.totalCinders;
+  }
+
+  get averageHitsPerCast() {
+    return 6 - ((this.totalHits / 6) / (this.totalCinders / 6));
```
kfinch: I don't understand how this could be the correct calculation.

Lets take a simple example: you cast Cinderstorm once and completely miss. This calculation is then `6 - ((0 / 6) - (6 / 6)) === 6 - (0 / 6) === 6` ... that seems obviously incorrect. Wouldn't the right calculation be much simpler, something like `this.totalHits / this.abilityTracker.getAbility(SPELLS.CINDERSTORM_TALENT.id).casts` ?

src/Parser/Mage/Fire/Modules/Features/Cinderstorm.js
```diff
@@ -0,0 +1,107 @@
...
+    this.castTimestamp = event.timestamp;
+    this.cinderHits = 0;
+  }
+
+  on_byPlayer_damage(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.CINDERSTORM_DAMAGE.id) {
+      return;
+    }
+    if (this.castTimestamp - event.timestamp < DAMAGE_WINDOW_MS && this.cinderHits < 6) {
```
kfinch: Why are you capping hits per cast at 6? When engaging multiple targets each cinder can strike more than once, allowing the possibility of well over 6 hits per cast. I assume this is to avoid showing an over 100% hit rate during AoE, but I think capping like this is a very misguided choice. In fact, my understanding is that the main reason one even specs for Cinderstorm is for AoE encounters. When I've played Fire in the past and specced Cinderstorm, the stat I was most interested in was how effectively I AoE'd with it.

src/Parser/Mage/Fire/Modules/Features/Cinderstorm.js
Sharrq: yea i was overthinking it and still got it wrong. ill fix it.

src/Parser/Mage/Fire/Modules/Features/Cinderstorm.js
Sharrq: So originally i was only counting the cinders that hit the target, but Cinderstorm has no target. So i was going to use the Hit % for the statistic, but i didnt want a Hit % of over 100% cause that would just be weird, so i capped it at 6. But then I decided that Hits per Cast worked a lot better and wouldnt be awkward if the average was like 8 per cast ... but then i forgot to uncap the hits.

Sharrq:
Made the changes and got rid of some stuff that wasnt needed since focusing more on hits per cast instead of missed/hit %

src/Parser/Mage/Fire/Modules/Features/Cinderstorm.js
```diff
@@ -0,0 +1,71 @@
...
+    return this.cinderHits / this.abilityTracker.getAbility(SPELLS.CINDERSTORM_TALENT.id).casts;
+  }
+
+  get suggestionThreshold() {
+    return {
+      actual: this.averageHitsPerCast,
+      isLessThan: {
+        minor: 6,
+        average: 5.7,
+        major: 5.4,
```
kfinch: This seems like a really narrow window of thresholds IMO. This is saying that always getting 6 is no issue, but getting a mix of 5s and 6s could potentially be a major issue. Getting 9 perfect hits and 1 miss is a major issue.

I dunno how much you've played with Cinderstorm, but it's really hard and sometimes up to chance if you can actually hit with all 6 cinders. My opinion is this should be spaced out more.

Sharrq:
I can widen it a little, I just made it that narrow because my understanding from the mage discord is that the only way the talent is useful is if you hit every cinder and if every cinder hits every mob. The average is the equivalent of 95% and the major is the equivalent of 90%.

Maybe I can do 95, 90, 80

kfinch:
Hmm, yeah I could easily be behind the times. Clearly you should make the minor threshold at 10 targets hit, because it's only useful in AoE! (I'm joking)

Sharrq:
i made the change to the suggestion thresholds, so it should be good now once travis finishes

PR #1183, opened 2018-01-14
[resto druid] Fixed some minor issues in RejuvenationAttributor and Tearstone

kfinch:
In RejuvenationAttributor, fixed an issue where procs that happen at around the timestamp as a PotA consumption could be spuriously attributed to PotA.

In Tearstone, fixed an issue where it was assumed that the maximum number of HoTs were applied per cast, which could sometimes deflate the reported proc rate.


kfinch:
Also pushed an update to the Tearstone tooltip to better explain proc rate and to include a "bonus rejuvs per cast" stat.

PR #1188, opened 2018-01-15
[General] Aman'Thul's Vision

CollCrom:
Added in a module for Amanthuls. Also removed friendly fire in all Retribution modules that were accounting for it.

capture

### DIFF
--- a/src/common/SPELLS/OTHERS.js
+++ b/src/common/SPELLS/OTHERS.js
@@ -778,4 +778,15 @@ export default {
     name: 'Felshield',
     icon: 'ability-vehicle-shellshieldgenerator-green',
   },
+  //Aman'Thul's Vision
+  AMANTHULS_GRANDEUR: { //Pantheon Proc
+    id: 256832,
+    name: 'Aman\'Thul\'s Grandeur',
+    icon: 'ability_shaman_repulsiontotem',
+  },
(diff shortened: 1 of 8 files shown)


src/common/SPELLS/OTHERS.js
```diff
@@ -778,4 +778,15 @@ export default {
...
   },
+  //Aman'Thul's Vision
+  AMANTHULS_GRANDEUR: { //Pantheon Proc
+    id: 256832,
+    name: 'Aman\'Thul\'s Grandeur',
+    icon: 'ability_shaman_repulsiontotem',
+  },
+  GLIMPSE_OF_ENLIGHTENMENT: { //Regular Proc
+    id: 256818,
+    name: 'Glimpse of Enlightenment',
```
Pewtro: These two `` //Regular Proc `` and ``//Pantheon Proc`` should be opposite right?

CollCrom: http://www.wowhead.com/spell=256818/glimpse-of-enlightenment

http://www.wowhead.com/spell=256832/amanthuls-grandeur

I don't think so.

CollCrom: done and done friend

Pewtro:
Oi travis speed up

PR #1203, opened 2018-01-17
[Core] Added eye of command support 

Pewtro:
It looks like this: 
image

### DIFF
--- a/src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
+++ b/src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
@@ -1,0 +1,143 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import { calculateSecondaryStatDefault } from 'common/stats';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import { formatPercentage } from 'common/format';
+
+/**
+ * Eye of Command
+ * Your melee auto attacks increase your Critical Strike by X for 10 sec, stacking up to 10 times. This effect is reset if you auto attack a different target.
+ */
+class EyeOfCommand extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  eyeOfCommand = {
+    [SPELLS.LEGIONS_GAZE.id]: {
+      1: {
+        start: 0,
+        uptime: 0,
+      },
+      2: {
+        start: 0,
+        uptime: 0,
+      },
+      3: {
+        start: 0,
+        uptime: 0,
+      },
+      4: {
+        start: 0,
+        uptime: 0,
+      },
+      5: {
+        start: 0,
+        uptime: 0,
+      },
+      6: {
+        start: 0,
+        uptime: 0,
+      },
+      7: {
+        start: 0,
+        uptime: 0,
+      },
+      8: {
+        start: 0,
+        uptime: 0,
+      },
+      9: {
+        start: 0,
+        uptime: 0,
+      },
+      10: {
+        start: 0,
+        uptime: 0,
+      },
+    },
+  };
+
+  critPerStack = 0;
+  currentStacks = 0;
+  buffIsUp = false;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTrinket(ITEMS.EYE_OF_COMMAND.id);
+    if (this.active) {
+      const itemDetails = this.combatants.selected.getItem(ITEMS.EYE_OF_COMMAND.id);
+      this.critPerStack = calculateSecondaryStatDefault(865, 151, itemDetails.itemLevel);
+    }
+  }
+
+  on_byPlayer_applybuff(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.LEGIONS_GAZE.id) {
+      return;
+    }
+    this.currentStacks = 1;
+    this.eyeOfCommand[spellId][this.currentStacks].start = event.timestamp;
+    this.buffIsUp = true;
+  }
+  on_byPlayer_applybuffstack(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.LEGIONS_GAZE.id) {
+      return;
+    }
+    this.eyeOfCommand[spellId][this.currentStacks].uptime += event.timestamp - this.eyeOfCommand[spellId][this.currentStacks].start;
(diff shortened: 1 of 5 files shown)


src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Eye of Command
+ * Your melee auto attacks increase your Critical Strike by X for 10 sec, stacking up to 10 times. This effect is reset if you auto attack a different target.
+ */
+class EyeOfCommand extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  eyeOfCommand = {
```
MartijnHols: Why is this an object allowing for multiple spells? Can't there only be one buff as a result of this item?

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
```diff
@@ -0,0 +1,143 @@
...
+    [SPELLS.LEGIONS_GAZE.id]: {
```
MartijnHols: What do you need the individual start/uptimes for?

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
Pewtro: Oh woops - fixing.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
Pewtro: It was my original idea of how to calculate average critGain by getting the uptime of each stack level and summing in the end and then it worked and was accurate.

MartijnHols: Aha, in that case I recommend keeping a total instead like here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/bef2745c470bc4d1c5993d121731be76ce5ddb7b/src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/GarothiFeedbackConduit.js#L46
With that you can use a single getter to calculate the average:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/bef2745c470bc4d1c5993d121731be76ce5ddb7b/src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/GarothiFeedbackConduit.js#L22-L24

I believe this to be much simpler code, this will reduce the amount of code you need to calculate the average and be easier to track.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Eye of Command
+ * Your melee auto attacks increase your Critical Strike by X for 10 sec, stacking up to 10 times. This effect is reset if you auto attack a different target.
+ */
+class EyeOfCommand extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  [SPELLS.LEGIONS_GAZE.id] = {
```
MartijnHols: How about making this a statically named property instead? Since the buff and its ID don't change, and you don't have any other buffs, I think putting it in an array will only make things slightly more complicated.

MartijnHols: See my older comment for an even conciser way of calculating the average. In general when you are making giant pre-defined objects like this you need to ask yourself if there's not a way to do it runtime. What would you do if there were 1000 entries instead of 10? Adding properties is also make error-sensitive.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
```diff
@@ -0,0 +1,141 @@
...
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTrinket(ITEMS.EYE_OF_COMMAND.id);
+    if (this.active) {
+      const itemDetails = this.combatants.selected.getItem(ITEMS.EYE_OF_COMMAND.id);
+      this.critPerStack = calculateSecondaryStatDefault(865, 151, itemDetails.itemLevel);
+    }
+  }
+
+  on_byPlayer_applybuff(event) {
```
MartijnHols: Have you looked at the `changebuffstack` fabricated event? I think you should be able to reduce the total amount of code as well as the amount of duplicate code with that.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
Pewtro: is changebuffstack something we've fabricated?

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
```diff
@@ -0,0 +1,141 @@
...
+  on_finished() {
+    if (this.buffIsUp === true) {
+      this[SPELLS.LEGIONS_GAZE.id][this.currentStacks].uptime += this.owner.fight.end_time - this[SPELLS.LEGIONS_GAZE.id][this.currentStacks].start;
+    }
+  }
+
+  get averageCritGain() {
+    let averageCritGain = 0;
+    const spellId = SPELLS.LEGIONS_GAZE.id;
+    for (let i = 0; i < 10; i++) {
```
MartijnHols: Please use one of the array methods instead of `for` loops, in this case a `reducer` should do the job.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
MartijnHols: Yes, it was created to normalize buff stack handling.

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
Pewtro: The problem is that the buffs don't clear on ending combat, leaving me with inaccurate numbers if I was to simply adopt the Garothi Feedback direction.

MartijnHols: I suppose that's a bug in GFC, I'll see if I can fix

src/Parser/Core/Modules/Items/Legion/EyeOfCommand.js
Pewtro: I mean with a 1000, I'd obviously do something different, with 10 I thought it was fine.

Pewtro:
new commit should fix all the issues

PR #1204, opened 2018-01-18
[Balance] Added RM, merged moon spells and changed SpellUsable.reduceCooldown to return correct value for charge spells.

Gebuz:
Also changed balance spec completeness to GREAT. Sorry it's all in one PR, should've done the reduceCooldown on it's own, but without it the cooldown reduction on Radiant Moonlight is incorrect.

The fix to cooldown reduction was that any cooldown reduction that overlapped a charge would only return the value from replenishing the rest of the charge. For instance when at 2 charges and 2 seconds from finishing cd, reducing it by 5 would only return 2, even thought you went to 3 charges and 3 seconds recharged. It should now return the correct value.

### DIFF
--- a/src/Parser/Core/Modules/SpellUsable.test.js
+++ b/src/Parser/Core/Modules/SpellUsable.test.js
@@ -143,6 +143,14 @@ describe('Core/Modules/SpellUsable', () => {
       expect(result).toBe(7500);
       expect(instance.isOnCooldown(SPELLS.FAKE_SPELL.id)).toBe(false);
     });
+    it('reducing a spell with multiple charges on cooldown reduces the CD time on the next charge if it fully recharges the first charge', () => {
+      abilitiesMock.getMaxCharges = jest.fn(() => 2);
+      triggerCast(SPELLS.FAKE_SPELL.id);
+      triggerCast(SPELLS.FAKE_SPELL.id);
+      parserMock.currentTimestamp = 6000; //Leaves 1500ms cooldown remaining of the total 7500ms of the first charge recharging.
+      const reduction = instance.reduceCooldown(SPELLS.FAKE_SPELL.id, 5000); 
+      expect(reduction).toBe(5000);    
+    });
     it('reduceCooldown on a spell not on cooldown throws', () => {
       // We throw instead of returning something like null so that implementers *have* to take this into consideration.
       expect(() => {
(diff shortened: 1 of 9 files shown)


src/Parser/Core/Modules/SpellUsable.test.js
```diff
@@ -360,6 +360,15 @@ describe('Core/Modules/SpellUsable', () => {
         });
       }
     });
+    it('A charge spell that has the cooldown of more than a single charge reduced should return the full reduction value', () => {
```
MartijnHols: I think a better position for this test would be underneath the `reducing a cooldown beyond its duration finishes the cooldown` test.

src/Parser/Core/Modules/SpellUsable.test.js
```diff
@@ -360,6 +360,15 @@ describe('Core/Modules/SpellUsable', () => {
...
+      abilitiesMock.getMaxCharges = jest.fn(() => 3);
+      triggerCast(SPELLS.FAKE_SPELL.id);
+      triggerCast(SPELLS.FAKE_SPELL.id);
+      triggerCast(SPELLS.FAKE_SPELL.id);
+      parserMock.currentTimestamp = 6000; //Leaves 1500ms cooldown remainig
```
MartijnHols: This comment is kinda putting me on the wrong path, how about `leaves 1500ms cooldown remaining of the total 7500ms of the first charge recharging`?

src/Parser/Core/Modules/SpellUsable.test.js
MartijnHols: For name I recommend to state how it should work, preferably like it works in-game. e.g. `reducing a spell with multiple charges on cooldown reduces the CD time on the next charge if it fully recharges the first charge`. The full reduction value should always be returned by this method and is tested separately.

Gebuz:
Added the suggested changes

src/Parser/Monk/Brewmaster/Modules/Core/BrewCDR.test.js
```diff
@@ -30,6 +30,6 @@ describe('BrewCDR', () => {
   });
 
   it("should report 427.09s of CDR", () => {
```
emallson: This test title is no longer accurate (but it is a bit of a brittle title).

PR #1205, opened 2018-01-18
[BrM Monk] Implemented High Tolerance Talent

emallson:
This implements the haste gain from the High Tolerance for BrM monks. The haste is integrated into the core `Haste` module. I also changed the `BrewCDR` module to use the *mean* haste instead of the haste at the last event. This more accurately reflects the amount of additional CDR you need over the course of a fight, especially in conjuction with the extra 12%+ haste from using HT on an appropriately high-damage fight. A couple logs I've looked at it reduced the CDR suggestion threshold by almost 5% (additive).

### DIFF
--- a/src/Parser/Monk/Brewmaster/Modules/Spells/HighTolerance.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Spells/HighTolerance.js
@@ -81,7 +81,7 @@ class HighTolerance extends Analyzer {
 
   on_finished() {
     if(this._staggerLevel !== null) {
-      this.staggerDurations[this._staggerLevel] += this.owner.currentTimestamp - this._lastDebuffApplied;
+      this.staggerDurations[this._staggerLevel] += this.owner.fight.end_time - this._lastDebuffApplied;
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/Parser/Monk/Brewmaster/Modules/Spells/HighTolerance.js
```diff
@@ -0,0 +1,106 @@
...
+      return;
+    }
+    this.staggerDurations[event.ability.guid] += event.timestamp - this._lastDebuffApplied;
+    this._staggerLevel = null;
+  }
+
+  on_finished() {
+    if(this._staggerLevel !== null) {
+      this.staggerDurations[this._staggerLevel] += this.owner.currentTimestamp - this._lastDebuffApplied;
+    }
```
Pewtro: Not sure if ``this.owner.currentTimestamp`` works in on_finished, what does this return? 

Else you can use ``this.owner.fight.end_time``, that should return what you want atleast

emallson: It was working when I tried it, but using `end_time` is clearer so I'll change to that

PR #1202, opened 2018-01-17
[Survival] Initial Survival commit

Pewtro:

PR #1209, opened 2018-01-18
[Survival] Tier20 and Tier21

### LINKED ISSUE #1191: [Survival] Tier 20 support
2p:
Lacerate Deals 15% additional damage and lasts 6.0 sec longer.
4p:
Mongoose Bite deals 10% increased damage to targets bleeding from your Lacerate.

### LINKED ISSUE #1192: [Survival] Tier 21 support
2p:
Flanking Strike has a 50% chance to increase the critical strike chance of your next Raptor Strike by 100% and the critical strike damage of Raptor Strike by 100% within the next 20 sec.
--- Applies http://www.wowhead.com/spell=252094/exposed-flank

4p: 
Each cast of Mongoose Bite increases the damage of your next Raptor Strike by 20%. Stacks up to 6 times.
--- Applies http://www.wowhead.com/spell=252095/in-for-the-kill

Pewtro:
Relies on #1202 

Fixes #1191 
Fixes #1192


src/Parser/Hunter/Survival/Modules/Items/Tier21_2p.js
```diff
@@ -0,0 +1,106 @@
...
+
+  on_byPlayer_damage(event) {
+    if (!this.isApplicable(event)) {
+      return;
+    }
+    const amount = event.amount;
+    const absorbed = event.absorbed || 0;
+    const raw = amount + absorbed;
+    this.buffedRaptorStrikes++;
+    this.bonusDmg += (raw / this.critEffectBonus.getBonus(event));
```
Gebuz: Shouldn't this be `this.bonusDmg += raw * T21_2P_CRIT_DMG_BONUS / this.critEffectBonus.getBonus(event);` ? What is calculated currently is just the raw value assuming no crit if I'm reading it correctly.

Gebuz: Also, wouldn't it be possible to include the value of the 100% crit chance somehow? Otherwise the tooltip is going to be missing a lot of damage I'd assume.

Pewtro: It just guarantees the next raptor strike to be a crit, just worded weirdly by blizzard. 
I took the calculations from one of the existing modules using criftEffectBonus

Gebuz: I think you might have missed the line that multiplied by the critbonus then, this is from DrapeOfShame.js:

```
const rawNormalPart = raw / this.critEffectBonus.getBonus(healEvent);
const rawDrapeHealing = rawNormalPart * DRAPE_OF_SHAME_CRIT_EFFECT;
```

In your case it ends up actually returning the same value since the crit bonus is exactly 100%.

Pewtro: That is only because drape of shame is double dipping from said bonus, this doesn't doubledip anywhere

Gebuz: Maybe I misunderstood how the critEffectBonus works. My understanding was that it returns the value to multiply damage by, so that critDamage = nonCritDamage * critEffectBonus <=> nonCritDamage = critDamage /  critEffectBonus.

Pewtro: I'll revisit this tonight or tomorrow, see if I can get something for crit chance in aswell as ensuring the critdmg is correct.

Pewtro:
Did some in-game testing and changed my approach to this so it's just a getDamageBonus and using 50% as modifier, since it onyl affects the crit damage (so we go from 200% damage to 300% damage in total). 
Decided to not implement a crit chance modifier as it would never be accurate, but opted to talk a bit about it in the tooltip instead.

PR #1219, opened 2018-01-19
Add a DamageTakenTable component (and implementation for BrM)

emallson:
This stems from the discussion we had today in discord. tldr its a table that shows the abilities you, as a tank, took damage from, categorized by whether it was mitigated by physical or magical mitigation, the total amount you took, and the maximum single hit you got.

It looks [like](https://imgur.com/a/p8rsi) [this](https://imgur.com/a/xiT3q). Importantly, (at least for BrMs) it doesn't rely on the spell data so it catches things like Jagged Abrasion (Harjatan) being mitigated by physical reduction unlike *literally every other bleed effect.*

The component itself can be used by other tank specs by importing the component and then feeding it your own data. The way I classified abilities for BrM depends on stagger, and so won't work for other tanks. Spell data isn't reliable (see also: Jagged Abrasion) but can be a good starting point if you don't have an easy way to distinguish them (like BrMs do)

### DIFF
--- a/src/Main/DamageTakenTable.js
+++ b/src/Main/DamageTakenTable.js
@@ -13,6 +13,7 @@ class DamageTakenTable extends React.Component {
   static propTypes = {
     data: PropTypes.array.isRequired,
     spec: PropTypes.object.isRequired,
+    total: PropTypes.number.isRequired,
   };
 
   mitigationNames = {
@@ -24,21 +25,20 @@ class DamageTakenTable extends React.Component {
 
   render() {
     const specClassName = this.props.spec.className.replace(' ', '');
-    const sumTotalDmg = this.props.data.reduce((sum, cur) => sum + cur.totalDmg, 0);
     const row = (abilityData)  => {
       const { ability, totalDmg, largestSpike } = abilityData;
       return (
         <tr key={ability.guid}>
           <td>
             <div className="flex performance-bar-container"
-                 data-tip={`Total Damage Taken: ${formatNumber(totalDmg)} of ${formatNumber(sumTotalDmg)}.`} >
+                 data-tip={`Total Damage Taken: ${formatNumber(totalDmg)} of ${formatNumber(this.props.total)}.`} >
               <div
                 className={`flex-sub performance-bar ${specClassName}-bg`}
-                style={{ width: `${(totalDmg - largestSpike) / sumTotalDmg * 100}%` }}
+                style={{ width: `${(totalDmg - largestSpike) / this.props.total * 100}%` }}
               />
               <div
                 className="flex-sub performance-bar Hunter-bg"
-                style={{ width: `${(largestSpike / sumTotalDmg * 100)}%`, opacity: 0.4 }}
+                style={{ width: `${(largestSpike / this.props.total * 100)}%`, opacity: 0.4 }}
               />
             </div>
           </td>
--- a/src/Main/DamageTakenTable.js
+++ b/src/Main/DamageTakenTable.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 import Icon from 'common/Icon';
 import { formatNumber } from 'common/format';
@@ -24,7 +23,7 @@ class DamageTakenTable extends React.Component {
 
   render() {
     const row = (abilityData)  => {
-      const { ability, mitigatedAs, totalDmg, largestSpike } = abilityData;
+      const { ability, totalDmg, largestSpike } = abilityData;
       return (
         <tr key={ability.guid}>
           <td></td>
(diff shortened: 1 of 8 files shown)


kfinch:
Could you provide a screenshot of what this looks like?

Pewtro:
He has provided 2. 
Like: https://imgur.com/a/p8rsi
This: https://imgur.com/a/xiT3q

kfinch:
Oh reading is really difficult

Gebuz:
Wouldn't it be easier to get mitigation type (phys/magic) by using event.ability.type instead of having to do the stagger calculations? That way it would also work for other tank specs. Also it might be worth extending/using this module as it collects similar data: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/Modules/DamageTaken.js

emallson:
The problem with using `event.ability.type` is that it may not reflect how damage it's actually mitigated. Going back to the example in the PR: Harjatan's Jagged Abrasions is listed as physical (type 1) and mitigated as physical, while Imonar's Sever is listed as the same type but mitigated as magical.

It does look like I could extend `DamageTaken` to simplify this module, though.

Gebuz:
Is Sever being mitigated as magic consistent for all specs or is it just using magic stagger reduction?

emallson:
Consistent for all specs afaik. Sever is in fact the normal way bleeds are
mitigated (listed as physical, mitigated as magic) -- Jagged Abrasions is
actually the outlier, presumably because blizzard wanted to avoid another
Krosus. Moroes bleed, for instance, is listed as physical but mitigated as
magic.

On Jan 22, 2018 9:13 AM, "Gebuz" <notifications@github.com> wrote:

Is Sever being mitigated as magic consistent for all specs or is it just
using magic stagger reduction?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/WoWAnalyzer/WoWAnalyzer/pull/1219#issuecomment-359434573>,
or mute the thread

Gebuz:
I agree it makes a lot of sense to have it listed as magic for brewmasters then, since you don't Diffuse Magic or anything else that reduces magic dmg taken (and I presume, not affect a bleed), so for all intends and purposes it might aswell be magic for a BrM.

One thing I just noticed, the initial hit of Sever is mitigated by armor (bears take almost the same dmg from the initial hit as the dot), but is still staggered as magic.

emallson:
iirc the initial hit that applies the dot is mitigated as physical in general, but the way i've got it rigged up if *any* hit by an ability is treated as magic it lists the whole thing under magic. Both are under the same spell ID for this ability (blizz why)

Gebuz:
It looks like the initial hit isn't mitigated as physical by stagger to me?
udklip

I was going to suggest that you split each ability into physical or magical per hit like you say until I saw that it doesn't appear to be consistent with what is reduced by armor/stagger :( It might still be the way to go if the primary aim for you is to display whether it's mitigated by stagger, and then just rename the tab to reflect that.

emallson:
Huh, i just checked with `#brewcraft` and that is right, which is doubly weird as both parts are listed as physical.

emallson:
just updated it to use the `DamageTaken` module

@Gebuz i've left the generic column names for now. idk if it is even possible for other tank specs to use this without building a list of abilities by mitigation type, but other tanks do care about magic vs physical mitigation (e.g. BDKs have AMS, VDH has empower wards, Prot has Blessing of Spellwarding (talent)).

Gebuz:
I think you're right that keeping it BrM specific is best because of how central Stagger is. Other tanks can still use the table if they wish and just use the values from the DamageTaken Module to make the split between magic/physical based on the type.

At some point in the future I think adding a manual list for bleeds and how they are mitigated (armor? stagger? tick/initial) would be cool, and it shouldn't be too much work to maintain since there is usually only a few bleeds per tier. Afaik everything except for bleeds generally follow the normal rules of damage mitigation by type?

If you think it makes sense to you, you could rename the titles "Physical" and "Magical" to "Staggered as physical" and "Staggered as magical" for BrM. If I didn't know already, I would be a bit confused seing Sever listed as "Magical", but I'm no Brewmaster so you probably know better.

MartijnHols:
How about adding bars to the table? Like the *Atonement sources* tab bars, it could relativize the total damage taken with each other and for the largest spike it might be interesting to show a health bar like the *Low health healing* tab to show the actual danger the spike put you in.

emallson:
Looks like this now

Gebuz:
If you want to exclude absorbs from biggest hit wouldn't it make sense to exclude it from the total dmg taken aswell?

emallson:
Technically the total damage column also includes absorbed damage, but negative absorbs are a thing. I'm gonna try to track down wtf happened with that this weekend I guess.

emallson:
As mentioned in the commit msg, the negative absorb issue has been sniffed out. This also resolves the issue of sometimes largest spike being larger than total damage.

kfinch:
So what's the status of this, it's difficult to follow from all the discussion. @emallson do you believe this is ready for merge?

emallson:
I believe so, yes.

PR #1220, opened 2018-01-20
[Fire Mage] Add Rune of Power & Adjust Cast Efficiency

Sharrq:
Rune of Power was made previously and put in the shared folder for Mage, but was never added to the Fire CombatLogParser.

Also went through Abilities and added suggestions and recommended efficiency to some talents and abilities that didnt have it previously. Also removed cooldowns from some utility abilities to remove the Cast Efficiency % bar from them.


src/Parser/Mage/Fire/Modules/Features/Abilities.js
```diff
@@ -123,40 +139,34 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.ICE_BLOCK,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
-        cooldown: 240,
```
sMteX: Wouldn't be better to add 
```jsx
castEfficiency: {
   suggestion: false,
}
```
maybe together with `importance: ISSUE_IMPORTANCE.MINOR,` instead of removing the cooldown?

kfinch: In a bunch of Mage things we'd been skipping showing CD entirely because didn't like seeing the maximum casts in the cast efficiency tab at all .... really should be a settable flag somewhere instead of leaving out the CD tho.

kfinch:
I wrote the RuneOfPower module, and I deliberately excluded it from Fire because it doesn't handle Ignite properly.

Sharrq:
Oh right, it checks damage.... Crap. I'll find a way to hide that for fire so it still has the time spent outside of the circle.

src/Parser/Mage/Fire/Modules/Features/Abilities.js
Sharrq: Yea this like "you used your interupt 3% of the time" or "you didn't cast ice block at all" just looks bad with the bar and the % imo.

I wanted it to show the number of casts, but not the efficiency %. I'll see if I can find a way to add a setable flag, but this is exactly how frost has it setup.

Sharrq: added a castEfficiency flag to disable castEfficiency and therefore hide the percentage, bar, and max casts.

Sharrq:
I made a new RoP Module just for fire that has no damage bits since the time spent inside/outside the circle was the only thing i was really interested in from that module.

Sharrq:
Alright, i made that change.

PR #1222, opened 2018-01-20
Misc Stuff for Elemental

HawkCorrigan:
- Echoes of the Great Sundering Support
- Smoldering Heart Support
- EB buff fix
- ES spending checks

Putting it up for review
Pls don't merge before slanderman looked over it tomorrow and gives his okay too


src/Parser/Shaman/Elemental/Modules/Talents/ElementalBlast.js
```diff
@@ -14,6 +14,26 @@ class ElementalBlast extends Analyzer {
...
   };
 
+  currentBuffAmount=0;
+  lastFreshApply=0;
+  resultDuration=0;
+
+  on_removebuff(event) {
+    if (event.ability.guid === SPELLS.ELEMENTAL_BLAST_CRIT.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_MASTERY.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_HASTE.id){
+      this.currentBuffAmount--;
+      if (this.currentBuffAmount===0)
```
sMteX: Please use braces even around single-line statements. So in this case, 
```jsx
if (this.currentBuffAmount === 0) {
   this.resultDuration += event.timestamp - this.lastFreshApply;
}
```
Same with the following function.

src/Parser/Shaman/Elemental/Modules/Talents/ElementalBlast.js
```diff
@@ -14,6 +14,26 @@ class ElementalBlast extends Analyzer {
...
+
+  on_applybuff(event) {
+    if (event.ability.guid === SPELLS.ELEMENTAL_BLAST_CRIT.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_MASTERY.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_HASTE.id){
+      if (this.currentBuffAmount===0)
+        this.lastFreshApply=event.timestamp;
+      this.currentBuffAmount++;
+    }
+  }
+
   on_initialized() {
```
sMteX: I would prefer having that as a first method in the class for readability but it's just a suggestion, not an issue.

src/Parser/Shaman/Elemental/Modules/Talents/ElementalBlast.js
```diff
@@ -14,6 +14,26 @@ class ElementalBlast extends Analyzer {
     combatants: Combatants,
   };
 
+  currentBuffAmount=0;
+  lastFreshApply=0;
+  resultDuration=0;
+
+  on_removebuff(event) {
+    if (event.ability.guid === SPELLS.ELEMENTAL_BLAST_CRIT.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_MASTERY.id ||event.ability.guid === SPELLS.ELEMENTAL_BLAST_HASTE.id){
```
sMteX: You can simplify this by making an array of the ids (preferably a constant outside the class)
```jsx
const ELEMENTAL_BLAST_IDS = [
   SPELLS.ELEMENTAL_BLAST_CRIT.id,
   SPELLS.ELEMENTAL_BLAST_MASTERY.id,
   SPELLS.ELEMENTAL_BLAST_HASTE.id,
];
```
and then using `if (ELEMENTAL_BLAST_IDS.includes(event.ability.guid))` (or if you prefer, you can make that a `Set` but it's nearly the same).

src/Parser/Shaman/Elemental/Modules/Talents/ElementalBlast.js
```diff
@@ -14,6 +14,26 @@ class ElementalBlast extends Analyzer {
     combatants: Combatants,
   };
 
+  currentBuffAmount=0;
+  lastFreshApply=0;
+  resultDuration=0;
+
+  on_removebuff(event) {
```
sMteX: Would be better to prefix it also with `toPlayer` - `on_toPlayer_removebuff`. These would fire only when the target is the player. Same with `applybuff`.

src/Parser/Shaman/Elemental/Modules/Items/EchoesOfTheGreatSundering.js
```diff
@@ -0,0 +1,71 @@
...
+
+class EchoesOfTheGreatSundering extends Analyzer {
+    static dependencies = {
+        combatants: Combatants,
+    };
+
+    BuffedEarthquakeCasts = 0;
+    EchoesProcsCounter = 0;
+    TotalEarthquakeDamage = 0;
+    TotalEarthquakeCasts = 0;
```
sMteX: Class variables begin with lowercase letters - `totalEarthquakeCasts` etc.

src/Parser/Shaman/Elemental/Modules/Items/EchoesOfTheGreatSundering.js
```diff
@@ -0,0 +1,71 @@
...
+    
+    estimate_bonus_damage() {
```
sMteX: This can be made a getter - `get estimatedBonusDamage()` and then used with `this.estimatedBonusDamage`.

src/Parser/Shaman/Elemental/Modules/Items/SmolderingHeart.js
```diff
@@ -0,0 +1,79 @@
...
+
+class SmolderingHeart extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  SmolderingProcs = 0;
+  SmolderingDuration = 0;
+  MaelstromSpent = 0;
+  AscendanceCastEventIds = new Set();
```
sMteX: Lowercase variables please - `ascendanceCastEventIds` etc.

src/Parser/Shaman/Elemental/Modules/Items/SmolderingHeart.js
```diff
@@ -0,0 +1,79 @@
...
+
+  smoldering_uptime() {
```
sMteX: Both uptime functions can be made into getters
```jsx
get smolderingUptime() {
   return this.smolderingDuration / this.owner.fightDuration;
}
```
and used as `this.smolderingUptime`.

src/Parser/Shaman/Elemental/Modules/Items/SmolderingHeart.js
```diff
@@ -0,0 +1,79 @@
...
+    return this.SmolderingDuration / this.owner.fightDuration;
+  }
+
+  expected_smoldering_uptime() {
+    // Smoldering Heart (Elemental) gives a 0.10% (0.001 decimal) chance per point of Maelstrom spent to grant Ascendance for 10 seconds (10000ms).
+    const EXPECTED_PROC_COUNT = this.MaelstromSpent * 0.001;
+    const EXPECTED_UPTIME = (EXPECTED_PROC_COUNT * 10000) / this.owner.fightDuration;
+    return EXPECTED_UPTIME;
```
sMteX: Would be nice if you extracted the magic numbers outside the class as constants.
```jsx
const SMOLDERING_HEART_CHANCE = 0.001;
const SMOLDERING_HEART_ASCENDANCE_DURATION = 10000;
```
Also the variables here should be lowercase - `expectedProcCount`, `expectedUptime`.

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  maelstromSpendOutAscendence = [];
+  maelstromSpendInAscendence = [];
+  latershock = false;
+  gambling = false;
+
+  on_initialized() {
+    this.gambling = (this.combatants.selected.hasHands(ITEMS.SMOLDERING_HEART.id) && this.combatants.selected.hasFeet(ITEMS.THE_DECEIVERS_BLOOD_PACT.id));
+    this.pseudogambling = (this.combatants.selected.hasShoulder(ITEMS.ECHOES_OF_THE_GREAT_SUNDERING.id) && this.combatants.selected.hasFeet(ITEMS.THE_DECEIVERS_BLOOD_PACT.id));
```
sMteX: Please add this variable also to the list of variables few lines up, makes it easier to read.

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  get avgMaelstromSpendOutOfAsc() {
+    let spend = this.maelstromSpendOutAscendence.reduce((prev, cur) => {
```
sMteX: Can be shortened to 
```jsx
let spend = this.maelstromSpendOutAscendence.reduce((prev, cur) => prev + cur, 0);
```

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -0,0 +1,126 @@
...
+        .addSuggestion((suggest) => {
+          return suggest(<span>Try to cast <SpellLink id={SPELLS.EARTH_SHOCK.id} /> at 111 Maelstrom or more.</span>)
+            .icon(SPELLS.EARTH_SHOCK.icon)
+            .actual(`Average Maelstrom spent: ${formatNumber(this.avgMaelstromSpendOutOfAsc)}`)
+            .recommended(`More than 111 is recommended`)
+            .regular(5).major(15);
+        });
+    }
+
+    when(this.avgMaelstromSpendInAsc).isLessThan(111)
```
sMteX: This is duplicate of the suggestion above - is that a typo or do you *really* want to have 2 identical suggestions when both `gambling` and `pseudogambling` are `false`?

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -0,0 +1,126 @@
...
+    },0);
+    spend /= this.maelstromSpendInAscendence.length||1;
+    return spend;
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.EARTH_SHOCK.id && !this.latershock) {
+      this.latershock = true;
+      if (event.classResources[0].type === RESOURCE_TYPES.MAELSTROM.id) {
+        if (this.combatants.selected.hasBuff(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id)) {
```
sMteX: I assume you want to check if the player has Ascendance at the moment of cast. If yes, you need to supply additional parameter to `hasBuff` like this: 
```jsx
if (this.combatants.selected.hasBuff(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id, event.timestamp))
```

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -0,0 +1,126 @@
...
+          this.maelstromSpendOutAscendence.push(event.classResources[0].amount);
+        }
+      }
+    }
+    if (event.classResources[0].amount < 40 && event.classResources[0].type === RESOURCE_TYPES.MAELSTROM.id) {
+      this.latershock = false;
+    }
+  }
+
+  suggestions(when) {
```
sMteX: (forgot to submit this comment)

This section needs overhaul in my opinion. Please use the `actual` and `recommended` variables when you can. But more importantly, I would advise you to use the "suggestion objects" some of us are using.

Basically means you'll make 3 getters for each `gambling`, `pseudogambling` and "default", which return an object representing the suggestion thresholds and values, and pass it into the `when` function which takes care of the rest. It's easier to maintain if you decide to change the values, and as a bonus - it's also much more convenient to use when you get to implementing a Checklist in the future (see any implemented one for example).

An example for the "default" would be:
```jsx
get defaultSuggestionThresholds() {
    return {
      actual: this.avgMaelstromSpendOutOfAsc,
      isLessThan: {
        minor: 111,   // these numbers don't make a lot of sense either, please take a look at them
        average: 5,
        major: 15,
      },
      style: 'number',
    };
  }

suggestions(when) {
   .....
   else {
       when(this.defaultSuggestionThresholds)
            return suggest(<span>Try to cast <SpellLink id={SPELLS.EARTH_SHOCK.id} /> at 111 Maelstrom or more.</span>)
            .icon(SPELLS.EARTH_SHOCK.icon)
            .actual(`Average Maelstrom spent: ${formatNumber(actual)}`)
            .recommended(`> ${recommended} is recommended`);
       });
   }
}
```

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
HawkCorrigan: This one checks for spending outside of Ascendence.

So it should be displayed independently from the other ones and regardless of gambling or not

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
HawkCorrigan: Ah i thought the average and minor values were deltas from the suggested value, that makes things a little different then ye.

HawkCorrigan: Don't know if that would help out a lot tho because the wording changes in every case, so i would need to  make that dynamic too and also it get's weird with recommended being a range.

sMteX: The average and minor values were always a concrete number (that's why people wrote it like `regular(recommended + 5)`). Also, I think you misunderstood me - the object is there just to abstract the numbers away, you will still have 3 different suggestions with different texts, just the numbers are elsewhere.

HawkCorrigan: okay that makes sense then thx

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
sMteX: Oh, my bad, I didn't see the in/out. Still, the suggestion for this whole block is correct even for this block.

kfinch:
Poking this, what's this PR currently blocked on?

HawkCorrigan:
Not anymore no
before that commit it just crashed

Should be working right now, slanderman is quite busy but it should be fine to commit as is.

It's not quite flawless but for it to be better i have to do some extended stuff.

So if you guys give your okay, it should be okay to merge it.

PR #1223, opened 2018-01-20
[Sub rogue] ABC

TsaboTavok:
Rogue GCD always 1 second, except for outlaw during AR.
Thresholds expected to be high, as rogue is a class that does not have free abilities - everything costs energy or combo points to cast.

### DIFF
--- a/src/Parser/Rogue/Subtlety/Modules/Features/AlwaysBeCasting.js
+++ b/src/Parser/Rogue/Subtlety/Modules/Features/AlwaysBeCasting.js
@@ -1,0 +1,43 @@
+import CoreAlwaysBeCasting from 'Parser/Core/Modules/AlwaysBeCasting';
+import SPELLS from 'common/SPELLS';
+
+class AlwaysBeCasting extends CoreAlwaysBeCasting {
+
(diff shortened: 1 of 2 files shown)


src/Parser/Rogue/Subtlety/Modules/Features/AlwaysBeCasting.js
```diff
@@ -0,0 +1,45 @@
...
+        SPELLS.EVISCERATE.id,
+        SPELLS.NIGHTBLADE.id,
+        SPELLS.KIDNEY_SHOT.id,
+
+        SPELLS.CRIMSON_VAIL.id,
+        SPELLS.FEINT.id,
+    ]
+
+    static STATIC_GCD_ABILITIES = {
+        [SPELLS.DEATH_FROM_ABOVE_TALENT.id]: 2000,
```
kfinch: A static 2 second GCD? Blizz pls.

Pewtro: You're jumping in the air and such anyways

PR #1229, opened 2018-01-21
[Havoc] Refactoring

CollCrom:
Changed the ABC to the suggestions thresholds. Added in a few haste buffs to get the downtime correct. Updated the cast efficiency a little to account for talents and tier and such.

### DIFF
--- a/src/Parser/Core/Modules/Haste.js
+++ b/src/Parser/Core/Modules/Haste.js
@@ -37,8 +37,8 @@ class Haste extends Analyzer {
     [SPELLS.ICY_VEINS.id]: 0.3,
     [SPELLS.BONE_SHIELD.id]: 0.1, // Blood BK haste buff from maintaining boneshield
     ...HIGH_TOLERANCE_HASTE_FNS,
-    [SPELLS.METAMORPHOSIS_HAVOC_BUFF.id]: 25,
-    [SPELLS.HAVOC_T21_4PC_BUFF.id]: 25,
+    [SPELLS.METAMORPHOSIS_HAVOC_BUFF.id]: 0.25,
+    [SPELLS.HAVOC_T21_4PC_BUFF.id]: 0.25,
     // Haste RATING buffs are handled by the StatTracker module
 
     // Boss abilities:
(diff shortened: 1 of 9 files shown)


src/Parser/Core/Modules/Haste.js
```diff
@@ -37,6 +37,8 @@ class Haste extends Analyzer {
     [SPELLS.ICY_VEINS.id]: 0.3,
     [SPELLS.BONE_SHIELD.id]: 0.1, // Blood BK haste buff from maintaining boneshield
     ...HIGH_TOLERANCE_HASTE_FNS,
+    [SPELLS.METAMORPHOSIS_HAVOC_BUFF.id]: 25,
+    [SPELLS.HAVOC_T21_4PC_BUFF.id]: 25,
```
jjs451: Shouldn't these be `.25` instead of `25`?  I think `25` would get read as 2500% haste

src/Parser/DemonHunter/Havoc/Modules/Abilities.js
```diff
@@ -44,17 +51,7 @@ class Abilities extends CoreAbilities {
           suggestion: true,
           recommendedEfficiency: 0.95,
           extraSuggestion: `This plus Nemesis and Metamorphosis make up your huge windows.`,
-        },
-      },
-      {
-        spell: SPELLS.MOMENTUM_TALENT,
```
jjs451: Did momentum fall out of favor?  Why remove it entirely?

CollCrom: Momentum is a buff you gain from casting fel rush and vengeful retreat. It is not an ability you cast. It should never have been in abilities in the first place.
http://www.wowhead.com/spell=206476/momentum

src/Parser/Core/Modules/Haste.js
CollCrom: fixed

src/Parser/DemonHunter/Havoc/Modules/Abilities.js
jjs451: Gotcha, I know nothing about Havoc

PR #1230, opened 2018-01-21
Reworked the Unending Thirst Module.

yajinni:
As i noted in my comment:
Death Strike causes the buff blood shield. If blood shield is refreshed it means death strike was casted when the blood shield buff was active.

http://www.wowhead.com/spell=192567/unending-thirst

My previous way wasnt working right because in the logs it was showing the buff first then the cast. Making every cast show it had the buff up.

image


src/Parser/DeathKnight/Blood/Modules/Features/UnendingThirstTracker.js
```diff
@@ -12,28 +12,35 @@ class UnendingThirstTracker extends Analyzer {
...
+
+
   statistic() {
     return (
       <StatisticBox
         icon={<SpellIcon id={SPELLS.UNENDING_THIRST.id} />}
-        value={this.RecentDSCounter}
-        label="Empowered Death Strike"
-        tooltip="If you're looking to increase your dps as the cost of defense try increasing this."
+        value={`${this.bloodShieldRefreshed} out of ${this.totalDSCounter}`}
```
emallson: Consider either showing the percentage of empowered death strikes here, or showing the total additional damage (or dps) gained through unending thirst. If you do either, you can put the exact "X out of Y" in the tooltip.

src/Parser/DeathKnight/Blood/Modules/Features/UnendingThirstTracker.js
```diff
@@ -12,28 +12,35 @@ class UnendingThirstTracker extends Analyzer {
...
+        label="Empowered Death Strikes"
+        tooltip="If you're looking to increase your dps at the cost of defense try increasing this number more."
```
emallson: This tooltip should be expanded upon. What is the cost to defense? Maybe something like:
 
> Fraction of Death Strikes empowered by Unending Thirst. It is possible to trade healing done for damage done by using Death Strikes in sequence to take advantage of Unending Thirst.

src/Parser/DeathKnight/Blood/Modules/Features/UnendingThirstTracker.js
yajinni: This isnt a bad idea. I will look into seeing if i can show the additional damage as a dps gain.

src/Parser/DeathKnight/Blood/Modules/Features/UnendingThirstTracker.js
yajinni: Was trying to keep the tooltip short. But if you are new to blood i can see them not understanding how to do it properly. Will rewrite.

yajinni:
If possible i would like this pushed to live now. Currently the module returns wrong numbers which is pretty bad.  I will change the module to show dps increase over its static how many ds over total ds.

PR #1232, opened 2018-01-21
[resto druid] Fixed a bug where the ToL suggestion wasn't displaying correctly

kfinch:


kfinch:
I also noticed that the mana saved number wasn't being displayed right for Chameleon Song (I accidentally didn't copy the format function I used for ToL's statistic to CS's) .. just pushed fix for that too

PR #1234, opened 2018-01-22
Skullflowers haemostasis

yajinni:
Added blood dk shoulder legendary.

image

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -27,5 +27,4 @@
       <option name="m_ignoreAnnotatedVariables" value="false" />
     </inspection_tool>
   </profile>
-  
 </component>
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -27,4 +27,5 @@
       <option name="m_ignoreAnnotatedVariables" value="false" />
     </inspection_tool>
   </profile>
+  
 </component>
--- a/src/Parser/DeathKnight/Blood/CombatLogParser.js
+++ b/src/Parser/DeathKnight/Blood/CombatLogParser.js
@@ -21,6 +21,7 @@ import RunicPowerTracker from './Modules/RunicPower/RunicPowerTracker';
 import T20_2pc from './Modules/Items/T20_2pc';
 import T20_4pc from './Modules/Items/T20_4pc';
 import SkullflowersHaemostasis from './Modules/Items/SkullflowersHaemostasis';
+import ShacklesofBryndaor from './Modules/Items/ShacklesofBryndaor';
 
 class CombatLogParser extends CoreCombatLogParser {
   static specModules = {
@@ -62,6 +63,7 @@ class CombatLogParser extends CoreCombatLogParser {
     skullflowersHaemostasis: SkullflowersHaemostasis,
 
 
+    shacklesofBryndaor:ShacklesofBryndaor,
   };
 }
 
--- a/src/Parser/DeathKnight/Blood/CombatLogParser.js
+++ b/src/Parser/DeathKnight/Blood/CombatLogParser.js
@@ -20,6 +20,7 @@ import RunicPowerTracker from './Modules/RunicPower/RunicPowerTracker';
 
 import T20_2pc from './Modules/Items/T20_2pc';
 import T20_4pc from './Modules/Items/T20_4pc';
+import SkullflowersHaemostasis from './Modules/Items/Skullflowers_Haemostasis';
 
 class CombatLogParser extends CoreCombatLogParser {
   static specModules = {
@@ -58,6 +59,9 @@ class CombatLogParser extends CoreCombatLogParser {
     // Items:
     t20_2pc: T20_2pc,
     t20_4pc: T20_4pc,
+    skullflowersHaemostasis: SkullflowersHaemostasis,
+
+
   };
 }
 
--- a/src/Parser/DeathKnight/Blood/CombatLogParser.js
+++ b/src/Parser/DeathKnight/Blood/CombatLogParser.js
@@ -20,7 +20,7 @@ import RunicPowerTracker from './Modules/RunicPower/RunicPowerTracker';
 
 import T20_2pc from './Modules/Items/T20_2pc';
 import T20_4pc from './Modules/Items/T20_4pc';
-import SkullflowersHaemostasis from './Modules/Items/Skullflowers_Haemostasis';
+import SkullflowersHaemostasis from './Modules/Items/SkullflowersHaemostasis';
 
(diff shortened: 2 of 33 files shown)


src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
```diff
@@ -0,0 +1,86 @@
...
+  on_initialized() {
+    this.active = this.combatants.selected.hasShoulder(ITEMS.SKULLFLOWERS_HAEMOSTASIS.id);
+  }
+
+  buffStack = 0;
+  wastedBuff = 0;
+  maxBuffStacks = 5;
+  damage=0;
+  heal=0;
+  percentBuff=1/6;
```
Gebuz: Shouldn't this be 1/5 (20%)? Also, I'd suggest you make it a global const (define it before the class and name it in all caps). maxBuffStacks could also be made a global const.

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
```diff
@@ -0,0 +1,86 @@
...
+  dsCastWithBuff=false;
+
+
+  on_byPlayer_applybuff(event) {
+    if (event.ability.guid === SPELLS.Haemostasis_Buff.id) {
+      this.buffStack += 1;
+      this.savedBuffStack=this.buffStack;
```
Gebuz: Please define savedBuffStack with the other variables.

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
```diff
@@ -0,0 +1,86 @@
...
+    this.active = this.combatants.selected.hasShoulder(ITEMS.SKULLFLOWERS_HAEMOSTASIS.id);
+  }
+
+  buffStack = 0;
+  wastedBuff = 0;
+  maxBuffStacks = 5;
+  damage=0;
+  heal=0;
+  percentBuff=1/6;
+  dsCastWithBuff=false;
```
Gebuz: This bool is never used. I assume you meant to use it to set the savedBuffStacks back to 0? Right now it is never set to 0. Since you have to wait for both the healing and damage event to occur before you can set it to 0 I suggest you make 2 bools for this, one for the damage event and one for the healing, and once they're both true set them both to false and savedBuffStacks to 0.

src/common/SPELLS/DEATH_KNIGHT.js
```diff
@@ -589,4 +595,11 @@ export default {
     name: 'Cold Heart',
     icon: 'spell_frost_chainsofice',
   },
-};
+
+  //Skullflower's Haemostasis blood dk legendary buff
+  Haemostasis_Buff: {
```
Gebuz: Please change this to all caps.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -27,4 +27,4 @@
       <option name="m_ignoreAnnotatedVariables" value="false" />
     </inspection_tool>
   </profile>
-</component>
\ No newline at end of file
+</component>
```
Gebuz: Please add the newline back in here, purely a style thing.

src/Parser/DeathKnight/Blood/CombatLogParser.js
```diff
@@ -20,6 +20,7 @@ import RunicPowerTracker from './Modules/RunicPower/RunicPowerTracker';
 
 import T20_2pc from './Modules/Items/T20_2pc';
 import T20_4pc from './Modules/Items/T20_4pc';
+import SkullflowersHaemostasis from './Modules/Items/Skullflowers_Haemostasis';
```
Gebuz: Please move the underscore from the .js file name, just to match naming scheme.

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
```diff
@@ -0,0 +1,86 @@
...
+    if (event.ability.guid === SPELLS.Haemostasis_Buff.id) {
+      this.buffStack = 0;
+    }
+  }
+
+  on_byPlayer_heal(event) {
+    if (event.ability.guid !== SPELLS.DEATH_STRIKE_HEAL.id) {
+      return;
+    }
+    this.heal += event.amount * this.percentBuff * this.savedBuffStack;
```
kfinch: Should be `(event.amount + (event.absorbed || 0))` instead of `event.amount`, because absorbed healing is almost always still effective healing.

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.DEATH_STRIKE.id) {
+      return;
+    }
+    this.damage += event.amount * this.percentBuff * this.savedBuffStack;
```
kfinch: Should be `(event.amount + (event.absorbed || 0))` instead of `event.amount`, because absorbed damage is almost always still effective damage.

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
yajinni: If i am doing this wrong then sorry. But my understanding of that is as follows. Yes i increases the damage and healing by 20%. But when i see the damage, the buff has already been applied. So i am working backwards to find out from that damage how much of that damage came from the buff.

ex: 100 damage with 20% (20 extra damage then) buff = 120 total damage. 
So working backwards 120 *(1/6) = 20 which gives me the portion that was the damage increased part that i want to track.

.idea/inspectionProfiles/Project_Default.xml
yajinni: no idea why that happened. must have been my ide lol

src/common/SPELLS/DEATH_KNIGHT.js
yajinni: Fixed

src/Parser/DeathKnight/Blood/Modules/Items/Skullflowers_Haemostasis.js
yajinni: fixed

src/Parser/DeathKnight/Blood/CombatLogParser.js
yajinni: fixed

PR #1235, opened 2018-01-22
Down time update

yajinni:


Gebuz:
Looks fine, but looking at the AlwaysBeCasting.js file I noticed that there is a lot of abilities listed as being on the GCD even though they are not. This will mess with the downtime, so I'll wait with merging this if you want to fix that first. Otherwise let me know and I'll just merge this. 

Ideally it should be fixed by updating to the new format where all abilities are listed in abilities.js and those that are on the GCD have the prop `isOnGCD: true,` (see holy paladin for an up to date example), but if you want a quick fix you could just remove the abilities not on GCD from the ABILITIES_ON_GCD array.

Just at a quick glance, all the cooldowns, all the CC except for Gorefiends, and about half the talents shouldn't be in the list. Also Asphyxiate is listed as the UH/frost talent instead of the blood one (different ID cause it aint a talent).

Gebuz:
looks I was writing too slowly :D

CollCrom:
Ah darn it I just saw your comment @gebuz

Gebuz:
No worries, wasn't something that had to be in this PR.

PR #1238, opened 2018-01-23
[Havoc] Fury Tracker and Delusions of Grandeur Legendary

CollCrom:


src/Parser/DemonHunter/Havoc/Modules/Features/AlwaysBeCasting.js
```diff
@@ -20,16 +20,24 @@ class AlwaysBeCasting extends CoreAlwaysBeCasting {
...
     SPELLS.FEL_BARRAGE_TALENT.id,
     SPELLS.FEL_ERUPTION_TALENT.id,
+
+    //Utility
+    SPELLS.FEL_RUSH.id,
+    SPELLS.VENGEFUL_RETREAT.id,
   ];
 
+  static STATIC_GCD_ABILITIES = {
+    [SPELLS.FEL_RUSH]: 250,
```
kfinch: I assume this is also not a real GCD, but the animation time?

src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
```diff
@@ -0,0 +1,66 @@
...
+  			.recommended(`Wasting less than ${formatPercentage(recommended)}% is recommended.`);
+  	});
+  }
+
+	item() {
+		return {
+			item: ITEMS.DELUSIONS_OF_GRANDEUR,
+			result:(
+				<dfn data-tip={`You wasted ${formatNumber(this.furyTracker.cooldownReductionWasted)} second of cooldown reduction.`}>
+					<Wrapper>Reduced the cooldown of <SpellLink id={SPELLS.METAMORPHOSIS_HAVOC.id} icon/> by {formatNumber(this.furyTracker.cooldownReduction / this.owner.fightDuration * 60000)} seconds per minute.</Wrapper>
```
kfinch: I kinda think 'seconds per minute' is an awkward metric, 'seconds per cast' makes more sense to me. Let me know what you think about that.

src/Parser/DemonHunter/Havoc/Modules/ResourceTracker/FuryTracker.js
```diff
@@ -0,0 +1,67 @@
...
+		this.reduceCooldown(cost);
+		return cost;
+	}
+	
+
+	reduceCooldown(cost) {
+		if(!this.combatants.selected.hasShoulder(ITEMS.DELUSIONS_OF_GRANDEUR.id)){
+			return;
+		}
+		const COOLDOWN_REDUCTION_MS = 1000/30;
```
kfinch: This should probably a `const` outside the class (or even read from the legendary's file)

src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	get percentCDRWasted(){
+		return this.furyTracker.cooldownReductionWasted / this.furyTracker.cooldownReduction;
+	}
+
+	get suggestionThresholds() {
+    return {
+      actual: this.percentCDRWasted,
+      isGreaterThan: {
```
kfinch: I don't really like 'CDR wasted' suggestions / stats, because in my opinion they're just a more confusing version of cast efficiency. Doesn't this suggestion do basically the exact same thing as a cast efficiency suggestion?

src/Parser/DemonHunter/Havoc/Modules/Features/AlwaysBeCasting.js
CollCrom: http://www.wowhead.com/spell=195072/fel-rush
Real gcd in fact vs 

http://www.wowhead.com/spell=198793/vengeful-retreat

Which is off the gcd but has a 1s duration

src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
CollCrom: Per cast of the ability? What about reduced the cool down from X to Y. I was honestly unsure about what to display here

emallson: You could show it the way CDR is shown [for Brewmasters](https://wowanalyzer.com/report/nxzcqJKmYwHbdWMp/8-Heroic+Felhounds+of+Sargeras+-+Kill+(4:02)/Eisenpelz). The calculation for that [is here](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Monk/Brewmaster/Modules/Core/BrewCDR.js#L48-L50).

This number is (in aggregate) the same as CDR in league of legends (so 50% CDR means your cooldowns are effectively cut in half over the course of the fight, 100% means no cooldown (but is unreachable fwiw)).

kfinch: Yeah, % CDR is reasonable display, and so is "abilities average effective cooldown was X (reduced from Y)"

src/Parser/DemonHunter/Havoc/Modules/ResourceTracker/FuryTracker.js
CollCrom: I tried putting it in the DoG file and passing it into fury tracker but got a call stack error

src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
```diff
@@ -0,0 +1,44 @@
...
+		SpellUsable: SpellUsable,
+		furyTracker: FuryTracker,
+	}
+
+	on_initialized() {
+		this.active = this.combatants.selected.hasShoulder(ITEMS.DELUSIONS_OF_GRANDEUR.id);
+	}
+
+	get cooldownReductionRatio(){
+		return this.furyTracker.cooldownReduction / (this.owner.fightDuration + this.furyTracker.cooldownReduction);
```
kfinch: Uhh, I don't think that's what he meant by %CDR. I'll use concrete examples to explain the suggestion I made and the one emallson made:

**Percent CDR**
Meta's CD is 5 minutes, but with DoG you proc enough cooldown reduction that it comes off CD after 4 minutes. `(5 - 4) / 5 = 0.20` ... You display something like "This reduced Meta's CD by 20%". If there were multiple casts you would average the CDR.

**Cooldown Reduced per Cast**
Meta's CD is 5 minutes. After your first cast, DoG reduces the cooldown enough that it comes up after 4 minutes (so reduced CD by 60 seconds). After the second, it comes up after 3:30 (so reduced by 90 seconds). Average reduction was 75 seconds. You display something like "This reduced Meta's CD by an average of 75 seconds".

CollCrom: I can do that. Is I included abilities.js can I the cooldown of meta? Or is there a better way to do that since it's affected by traits

kfinch: You can just build it in the on_initialized

emallson: The way he has it is actually how I do it in the brewmaster code. Semantically, the numerator is the amount of time gained, the denominator is the total "time" that passed for this abilities cooldown. Note that:

1. It is 0% when no CDR was applied
2. It is 50% if time gained = fight time
3. It is 100% only if infinity time was gained (which can't happen, lining up with the fact that we can't get 100% CDR with "reduces CD by X seconds" tied to separate abilities)

This also lines up with what you wrote: if you gained 1 minute of reduction over 4 minutes of fight time, then `1 / (4 + 1) = 0.2`. If the total fight length was 6 minutes, you would gain an additional 30 seconds of time *if the CDR gain were perfectly continuous* at that rate, giving a CDR of `1.5 / (6 + 1.5) = 0.2`. The way the item is worded, the continuity assumption looks not unreasonable.

**However** my use case is for abilities with very short (< 20s) cooldowns. After thinking about it, this formula may not make the most sense for Meta because even if you do reduce the CD by 1.5 minutes total, you still may not get an extra cast. Not sure how to deal with that problem.

src/Parser/DemonHunter/Havoc/Modules/ResourceTracker/FuryDetails.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Tab from 'Main/Tab';
+import ResourceBreakdown from 'Parser/Core/Modules/ResourceTracker/ResourceBreakdown';
+
+import FuryTracker from './FuryTracker';
+
+class AstralPowerDetails extends Analyzer {
```
Gebuz: Looks like you forgot to rename this class (and the export at the end aswell)

CollCrom:
This PR has 10 file changes =[. Sorry!

kfinch:
Thanks!

PR #1242, opened 2018-01-23
[Affliction] Implemented Tier 21 set bonuses

sMteX:
The 2 piece required some awkward tracking but there are no events that indicate the proc whatsoever so.. This worked

Example:
image


src/Parser/Warlock/Affliction/Modules/Items/Tier21_2set.js
```diff
@@ -0,0 +1,103 @@
...
+import SpellLink from 'common/SpellLink';
+import Wrapper from 'common/Wrapper';
+import ItemDamageDone from 'Main/ItemDamageDone';
+import { formatNumber } from 'common/format';
+import { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';
+
+import { UNSTABLE_AFFLICTION_DEBUFF_IDS } from '../../Constants';
+
+const TICKS_PER_UA = 4;
+const UA_IDS_SET = new Set(UNSTABLE_AFFLICTION_DEBUFF_IDS);
```
kfinch: My understanding is that for small collections, just keeping this as an array and using `includes` is actually more performant than using a Set.

src/Parser/Warlock/Affliction/Modules/Items/Tier21_4set.js
```diff
@@ -0,0 +1,63 @@
...
+import Enemies from 'Parser/Core/Modules/Enemies';
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+import Wrapper from 'common/Wrapper';
+import ItemDamageDone from 'Main/ItemDamageDone';
+import { formatNumber, formatPercentage } from 'common/format';
+import getDamageBonus from 'Parser/Warlock/Affliction/Modules/WarlockCore/getDamageBonus';
+
+const AFFECTED_ABILITIES = new Set([
```
kfinch: My understanding is that for small collections, just keeping this as an array and using `includes` is actually more performant than using a Set.

sMteX:
I've checked - yes, additional haste just shortens the duration of UA, it doesn't add ticks. Each proc adds 2 ticks but from logs I've checked sometimes it ticked just once (maybe the mob died). What do you mean by partial ticks? Also, I've used Set because I use it in other places in the code for the same reason, but there's no harm in switching back to arrays.

kfinch:
Ahh, it could easily be different for DoTs vs HoTs.

For Resto Druid, our HoTs tick faster, but the overall duration remains the same (meaning haste actually gets us increased throughput per cast). If Haste doesn't speed things up enough to get to another full tick, we end up with a 'partial tick' right before the HoT expires. For example, consider a HoT that at 0% haste ticks 5 times over 10 seconds. If the player had 50% haste, the HoT would still last 10 seconds, but would tick "7.5 times" - basically 7 regular ticks and then a half strength tick right as the HoT expires. There's also some weird ass "tick clamping" things that happen when a HoT is extended.

ANYWAY, all of this means that tick counting is basically completely ineffective for Resto Druids. It sounds like things don't work this way for Affli.... so you can go ahead and ignore all of this, I'll approve the PR.

PR #1248, opened 2018-01-24
[Affliction] Miscellaneous updates

sMteX:
- updated few thresholds in the Checklist
- split the UA buff module into Reap Souls and Malefic Grasp/Haunt, to better showcase how those things buffed UA instead of shoving it into one statistic box
- added Soul Conduit to show shards refunded on a more visible place instead of just the tab
- added a dirty fix for the utility spells in the Checklist as I didn't like "time on cooldown" for them, but just the raw casts (as was previously used)


src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -135,6 +135,12 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.UNENDING_RESOLVE,
         category: Abilities.SPELL_CATEGORIES.UTILITY,
         cooldown: 180,
+        castEfficiency: {
```
Gebuz: you could add `importance: ISSUE_IMPORTANCE.MINOR` to castEfficiency if you only want defensive and utility spells to be minor issues. Then you could adjust the average and major thresholds specifically for the checklist and avoid the dirty fix.

src/Parser/Warlock/Affliction/Modules/Features/ReapBuffTracker.js
```diff
@@ -0,0 +1,77 @@
...
+
+  get unbuffedTicks() {
+    return this.totalTicks - this.buffedTicks;
+  }
+
+  get suggestionThresholds() {
+    return {
+      actual: (this.unbuffedTicks / this.totalTicks) || 1,  // if no UAs were cast (totalTicks and unbuffedTicks = 0), it should return NaN and thus be 1 (100% unbuffed ticks)
+      isGreaterThan: {
+        // TODO
```
Gebuz: Is this comment left in intentionally? Do you want to update the numbers before it's merged or is it a thing for the future? (same comment is also in Haunt & MG suggestionThresholds)

src/Parser/Warlock/Affliction/Modules/Talents/MaleficGrasp.js
```diff
@@ -42,46 +47,81 @@ class MaleficGrasp extends Analyzer {
...
     const target = this.enemies.getEntity(event);
     if (!target) {
       return;
     }
-    const buffedByDrain = target.hasBuff(SPELLS.DRAIN_SOUL.id, event.timestamp);
-    if (!buffedByDrain) {
-      return;
+
+    const drained = target.hasBuff(SPELLS.DRAIN_SOUL.id, event.timestamp);
+    if (AFFECTED_ABILITIES.includes(spellId)) {
```
Gebuz: I assume this is supposed to be `UNSTABLE_AFFLICTION_DEBUFF_IDS` like the other modules?

src/Parser/Warlock/Affliction/Modules/Features/ReapBuffTracker.js
```diff
@@ -0,0 +1,77 @@
...
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import { formatPercentage } from 'common/format';
+import Wrapper from 'common/Wrapper';
+import SpellLink from 'common/SpellLink';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+import { UNSTABLE_AFFLICTION_DEBUFF_IDS } from '../../Constants';
+
+const UA_IDS_SET = new Set(UNSTABLE_AFFLICTION_DEBUFF_IDS);
```
Gebuz: Is there any advantage to using a set for this instead of just the array? Otherwise it seems unnecessary. (same in MG and Haunt)

src/Parser/Warlock/Affliction/Modules/Features/ReapBuffTracker.js
sMteX: I plan on doing another round of getting feedback from the person that's helping me, so if it's ok with you, I'll leave the comment in. I'll remove it in next PR.

src/Parser/Warlock/Affliction/Modules/Talents/MaleficGrasp.js
sMteX: No, this is correct, there's an array of affected abilities (apart from UA it also contains Agony, Corruption and Siphon Life).

Gebuz: Isn't it only UA ticks you want to count though? The statbox label says "UA ticks buffed by Malefic Grasp" so I'd assume one of them needs to be changed.

sMteX: That's a good point. The stat box is oriented around buffed UA ticks, which matters the most, but the tooltip also shows total damage contributed by the talent. Might be more sensible to split that into two modules but I can't help but think that's a bit redundant.

Gebuz: I think it's fine to just display either percent of UA ticks buffed or DPS contributed, which ever you think is more important, and then show the other in the tooltip. If you show the DPS contributed you could also add the percent of buffed dot ticks for each of the ticks individually in the tooltip. 

Alternatively you could also make a module for Drain Soul that shows time spent channeling as that should also give a pretty good indication of how many Agony/Corruption ticks were buffed by MG.

sMteX: I would prefer the first approach which is what's currently impletented if I understood you correctly - percentage of buffed UA ticks (which is more important when running the talent), and in the tooltip damage done (+ % of total damage done), and then how much damage it contributed for each of the affected dots.

image

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
Gebuz: I tend to agree that time on CD isn't ideal, but max cast suffer similar issues. For instance if the encounter is 3 min. and 1 sec long and your cooldown is 3 min. max casts will be twice what it actually should be. Maybe the solution is to use something like cooldownDuration/encounterLength for cooldowns that are not supposed to be used on cooldown? Anyway we should probably discuss that on discord or in an issue instead of in this PR.

If you want to keep the dirty fix that's okay with me, although I'd prefer if you didn't since neither solution is ideal and the recommended thresholds are pretty forgiving anyway.

src/Parser/Warlock/Affliction/Modules/Talents/MaleficGrasp.js
Gebuz: Just remember to change it so it's only the UA ticks being counted! :D

sMteX: Woops, I misunderstood you in the beginning.. Aggressive refactoring 😅

PR #1249, opened 2018-01-24
 [Windwalker] Adding statistic showing actual casts vs expected casts during SEF or Serenity

Juko8:


CollCrom:
Does this account for moments of high haste like bloodlust and/or CD extension due to the legendary wrists?

src/Parser/Monk/Windwalker/Modules/Spells/StormEarthAndFire.js
```diff
@@ -36,6 +46,57 @@ class StormEarthAndFire extends Analyzer{
...
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+
+    if (!this.combatants.selected.hasBuff(SPELLS.STORM_EARTH_AND_FIRE_CAST.id) && !this.combatants.selected.hasBuff(SPELLS.SERENITY_TALENT.id)) {
+      return;
+    }
+    if (SPELLS.RISING_SUN_KICK.id === spellId) {
```
sMteX: This could be simplified to a switch statement:
```jsx
switch (spellId) {
   case SPELLS.RISING_SUN_KICK.id:
      this.risingSunKicks += 1;
      break;
   case SPELLS.FISTS_OF_FURY_CAST.id:
      this.fistsOfFuries += 1;
      break;
   case SPELLS.STRIKE_OF_THE_WINDLORD.id:
      this.strikeOfTheWindlords += 1;
      break;
   case SPELLS.WHIRLING_DRAGON_PUNCH_TALENT.id:
      this.whirlingDragonPunches += 1;
      break;
   default:
      break;
}

```

src/Parser/Monk/Windwalker/Modules/Spells/StormEarthAndFire.js
```diff
@@ -36,6 +46,57 @@ class StormEarthAndFire extends Analyzer{
...
+          <div>
+          During your {this.sefCount} {icon.name}s you cast: 
+          <li>{this.risingSunKicks}/{this.sefCount*2+1} Rising Sun Kicks</li> 
+          <li>{this.fistsOfFuries}/{this.sefCount} Fists of Furies</li>
+          <li>{this.strikeOfTheWindlords}/{this.sefCount} Strikes of the Windlord</li>
+              {this.whirlingDragonPunches > 0
+              ? <li>{this.whirlingDragonPunches}/{this.sefCount} Whirling Dragon Punches</li> : ''
+              }
+        </div>
+       }
```
sMteX: Seems too large for the label of a statistic box. How about rewording that and moving that to the tooltip so the `label` is smaller, or making use of `substatistic`s like here? 
image

src/Parser/Monk/Windwalker/Modules/Spells/StormEarthAndFire.js
```diff
@@ -36,6 +46,57 @@ class StormEarthAndFire extends Analyzer{
     const reducedCooldownWithTraits = 90 - this.traitsCDReduction;
     return reducedCooldownWithTraits;
   }
+
+  on_toPlayer_applybuff(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS.STORM_EARTH_AND_FIRE_CAST.id === spellId || SPELLS.SERENITY_TALENT.id === spellId) {
+      this.sefCount += 1;
```
sMteX: You could use the `AbilityTracker` module to get the number of casts instead of counting them yourself.

Juko8:
You need a lot of haste to get any extra casts, 1 extra Rising Sun Kick is expected for a use during heroism/lust. I have some more plans with the legendary wrists, and their extension isn't always consistent, so it's hard to expect more FoF/WDP casts with them even though it is possible. The right side number is also intended to be a "expected" number of casts, not a maximum possible number of casts nor an absolute must have amount of casts or, since SEF use can change a lot depending on the encounter.

src/Parser/Monk/Windwalker/Modules/Spells/StormEarthAndFire.js
Juko8: I had it as a label with the list as a tooltip, but i didn't like the way it looked in the statistics box as a greyed out text without also having a value. I had been looking around and wanted to make something that looks like the overload statistic for elemental shamans, which is also a quite large box.

kfinch:
@sMteX have your suggestions been resolved? If so, please mark approved.

sMteX:
Sorry for holding this back. I guess the stat box design is up to you, how you like it, I just wanted to give another opinion.

PR #1261, opened 2018-01-26
[Havoc] added in t21 2pc

CollCrom:
image


src/Parser/DemonHunter/Havoc/CHANGELOG.js
```diff
@@ -4,6 +4,16 @@ import { Hewhosmites, Mamtooth } from 'MAINTAINERS';
```
sMteX: Missing imports for `SpellLink, SPELLS, ItemLink` and `ITEMS` causing Travis to fail.

src/Parser/DemonHunter/Havoc/CHANGELOG.js
```diff
@@ -4,6 +4,16 @@ import { Hewhosmites, Mamtooth } from 'MAINTAINERS';
 
 
 export default [
+	{
+		date: new Date('2018-1-26'),
```
sMteX: I think the date has to be in yyyy-mm-dd format, `2018-01-26`

kfinch:

CollCrom:
I fixed all my changelog dates since I had the wrong format for all of them as well

PR #1262, opened 2018-01-26
[Fire Mage] Added Checklist

Sharrq:
Added the Checklist to Fire Mage.


src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+          }),
+          new Requirement({
+            name: 'Cancelled Casts',
+            check: () => this.cancelledCasts.cancelledCastSuggestionThresholds,
+          }),
+        ];
+      },
+    }),
+    new Rule({
+      name: 'Managing  your procs',
```
kfinch: Less passive name maybe, like "Manage your procs"

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+          new Requirement({
+            name: "Hard Cast Before Hot Streak Util",
+            check: () => this.hotStreak.castBeforeHotStreakThresholds,
+          }),
+          new Requirement({
+            name: "Expired Hot Streaks",
+            check: () => this.hotStreak.expiredProcsThresholds,
+          }),
+          new Requirement({
+            name: "Wasted Crits",
```
kfinch: It might not be clear from this what a "wasted crit" is... I should really add something so requirements can have tooltips...

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+        ];
+      },
+    }),
+    new Rule({
+      name: 'Managing  your procs',
+      description: <Wrapper>Fire Mage is heavily dependent on correctly using your <SpellLink id={SPELLS.FIRE_BLAST.id}/> and <SpellLink id={SPELLS.PHOENIXS_FLAMES.id}/> guaranteed crit abilities to properly convert <SpellLink id={SPELLS.HEATING_UP.id}/> to <SpellLink id={SPELLS.HOT_STREAK.id}/>. Remember to only use your guaranteed crit spells to convert Heating Up to Hot Streak, and also remember to precede each instant cast <SpellLink id={SPELLS.PYROBLAST.id}/> with a hardcast of <SpellLink id={SPELLS.FIREBALL.id}/>, <SpellLink id={SPELLS.SCORCH.id}/>, or <SpellLink id={SPELLS.PYROBLAST.id}/> (If you have <ItemLink id={ITEMS.MARQUEE_BINDINGS_OF_THE_SUN_KING.id}/>) to increase the chances of getting a new Heating Up or Hot Streak</Wrapper>,
+      requirements: () => {
+        return [
+          new Requirement({
+            name: "Fire Blast used without Heating Up",
```
kfinch: This stat might be better as an 'affirmative' stat, like the inverse number and "Fire Blasts used with Heaing Up"

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+            check: () => this.heatingUp.fireBlastUtilSuggestionThresholds,
+          }),
+          new Requirement({
+            name: "Phoenix's Flames used without Heating Up",
```
kfinch: This stat might be better as an 'affirmative' stat, like the inverse number and "Phoenix's Flames used with Heaing Up"

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+          }),
+          new Requirement({
+            name: "Wasted Crits",
+            check: () => this.hotStreak.wastedCritsThresholds,
+          }),
+        ];
+      },
+    }),
+    new Rule({
+      name: 'Managing Combustion',
```
kfinch: Less passive name, like "Maximize Combustion Effectiveness"

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+      requirements: () => {
+        return [
+          new Requirement({
+            name: "Combustion Used with < 2 Phoenix Flames Charges",
```
kfinch: If possible, I think it would make more sense to show how many PFs were actually cast per combustion, rather than the number available at the start of it.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+            check: () => this.combustionCharges.phoenixFlamesThresholds,
+          }),
+          new Requirement({
+            name: `Combustion Used with < ${this.combatants.selected.hasTalent(SPELLS.FLAME_ON_TALENT.id) ? 2 : 1} Fire Blast Charges`,
```
kfinch: If possible, I think it would make more sense to show how many Fire Blasts were actually cast per combustion, rather than the number available at the start of it.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,178 @@
...
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.METEOR_TALENT,
+            when: combatant.hasTalent(SPELLS.METEOR_TALENT.id),
+          }),
+        ];
+      },
+    }),
+
+    new Rule({
+      name: 'Utilize your talents',
```
kfinch: Is it possible to hide this if talent choices make there be no Requirements?

src/Parser/Mage/Fire/Modules/Features/Checklist.js
Sharrq: The point of that module is to make sure the player is banking charges so that they have them during Combustion. The number of uses during Combustion doesnt particularly matter because one Combustion could involve hard casting Pyroblast 3 times because they got lucky on Bracer Procs, and the next might be entirely full of instant casts because they got no Bracer Procs.

But the goal of Combustion is (with the exception of Bracer Procs) to get as many instant cast pyroblasts as possible, so if a player casts Combustion and they have 0 PF Charges and 1 Fire Blast charge ... that combustion is going to suck for them cause its going to be almost entirely full of hard casts.

So i could add the stat for the number used during combustion .. though idk that it would actually be meaningful info for the player. 

Same for Fire Blast.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
Sharrq: i have no idea ... is there someone else doing something like that? I can try to figure it out.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
kfinch: Oh ok, yeah I wasn't thinking about the Bracers. Sounds fine to leave as is then.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
Sharrq: i thought about trying to figure out how to add tooltips to it, but didnt end up doing it. Especially after the extremely long description i put on a couple of the rules. Tooltips would be better.

src/Parser/Mage/Fire/Modules/Features/Checklist.js
sMteX: The only thing that comes into mind is wrapping it somehow in a ternary operator, or providing another prop `when` to the Rule class. Might be a good thing to add sometimes as having an empty Rule is confusing.

kfinch:

Sharrq:
Alright, made the following changes

- Resolved the merge conflict
- Fixed a calculation issue in Rune of Power (Was showing Nan in Checklist)
- Added Tooltip support for the Core Checklist
- Added Tooltips to the Fire Mage Checklist
- Adjusted the Core Checklist to not show a rule if it has no active requirements
- Made the remaining suggested changes

sMteX:
I was thinking about the tooltip - wouldn't it be better to add the tooltip (not necessarily in your case but generally) to the displayed value instead? Because you can show the tooltip for the text in the exact way as you did it without making changes in Core (`<Wrapper><dfn data-tip={...}>Text</dfn></Wrapper>`) but you can't do that for the value. Would be very nice for displaying 2 values (for me it's for example "wasted shards per minute", when on mouseover it would show "total wasted shards").

PR #1263, opened 2018-01-26
[Feral] Ferocious Bite energy check

Anatta336:
Created an analyzer and suggestion to detect Ferocious Bite being used at low energy, and so missing out on bonus damage.
biteenergyexample


src/Parser/Druid/Feral/Modules/Features/FerociousBiteEnergy.js
```diff
@@ -0,0 +1,90 @@
...
+        average: 25,
+        major: 60,
+      },
+      style: 'number',
+    };
+  }
+
+  suggestions(when) {
+    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
+      let actualText = '';
```
Gebuz: I would suggest that you move this text to the suggest text instead (Just add it infront of your current suggest text) and then use the actual variable you get from the threshold as the actual text. That way your specific text on how many bites that was messed up and by how much is more in focus.

kfinch:

Anatta336:
Thank you very much for the feedback.

For the Suggestion threshold, I wonder if I'm needlessly complicating it. The rule widely stated for Feral druids is to never use Ferocious Bite at less than 50 energy. So the minor threshold and recommended value could just be 0% of all bites at low energy (good point to make it relative, thank you).

A 49 energy bite is much less of a problem than a 25 energy Bite. I'd like to convey how severe or minor a player's low energy bite use was. I worry that my measure of _"Your average energy below 50 on bites with below 50 energy was X"_ is difficult for a human to parse. While a simple average of energy spent on bites would disguise problems that the player should address.

Showing DPS lost by the mistake of biting at low energy seems the most useful feedback to give a player. They can decide for themselves how significant a particular DPS difference is.

I've reworked the analyser to also track bite damage lost from not being at 50+ energy. This will be an overestimate of the overall damage lost because the player will most likely have used the energy not put into the bite to do damage in other ways. Accurately accounting for that would be difficult.

biteenergyexample2

kfinch:
@Anatta336 I've encountered a similar issue with how to do suggestion thresholds. It is an issue that only counting above or below 50 energy loses some detail of how well the player did. I actually think that 'average amount below 50' is the correct metric to use. It seems like what you were struggling with was how to display it.

How about this: "You used an average of X energy for each Ferocious Bite. Ferocious Bite does increased damage at higher energy, capping at 50 energy. You should always aim to cast Ferocious Bite when above that cap.". You could then set the suggestion thresholds using the average FB energy.

I'm a little skeptical of the DPS lost due to not using enough energy thing, because as you admit yourself it doesn't factor in the opportunity cost of using the extra energy.

kfinch:
P.S. sorry to jerk you around with conflicting directions, at this point I'd say do the suggestion how you think is best, and we can get this merged in 😄

Anatta336:
I have a sinking feeling that I messed up the changelog merge conflict. Although the branch seems to have ended up with a correct (new style) CHANGELOG.js

The suggestion thresholds is definitely tricky. A player who consistently uses Ferocious Bite at 25 to 30 energy should be reminded of the mechanic (even if they also bite at 80 energy). But making a fuss over the occasional 47 energy bite is a needless distraction. I suppose the "average energy used when biting at below 50" is probably the best metric we can get. With a minor suggestion if any bites fall in that range, increasing in severity with a lower average. Averaging over all bites would risk showing a player who had both low and high energy bites an average that indicates they handled the mechanic correctly.

I agree showing a "damage lost" number that doesn't account for total damage lost is not ideal. I like that it gives some indication of the damage difference being dealt with though. Perhaps something like a tooltip explaining that it doesn't account for damage done with the saved energy, but with a reminder that theory crafters agree it's better to bite at 50+.

kfinch:
and yeah, it's odd that the changes on GitHub are showing like that after the merge... I dunno what the deal is. What did you do to merge from master? What I usually do that works well is just `git pull upstream master`, and then I go in to the conflicted file and fix manually.

Anatta336:
That capped average sounds perfect for the suggestion threshold, brilliant!

I think I messed up by trying to bring master on my fork up to date, and somehow dragging all the commits into this branch. I'll have some free time later to try to unpick that and tidy this branch up.

kfinch:
It might be GitHub not rendering the diff right. Try pushing a trivial commit to this branch and see if it unfucks itself

Anatta336:
The extra commits seem to be sticking around in this pull request. It was definitely my fault doing something unintelligible with Git. My apologises for the mess.

I don't want to risk messing with it further and potentially losing this discussion history, nor would I want to waste anyone else's time fixing it. So I've gone ahead and opened a new pull request #1275 from a fresh branch with a copy of the latest files.

I'll close this pull request with #1275 acting as its replacement.

PR #1268, opened 2018-01-26
[Resto Druid] Added support for HoT extension tracking

kfinch:
This is a big PR and there wasn't a whole lot I could do about it, sorry.

There is still some weirdness present where I'm not able to perfectly track how long a HoT is going to last, but I'm pretty close. Fortunately that doesn't matter much yet (I'll need to lock it down once I implement a "no early refresh" module).

The meat of the changes are in HoTTracker, where I've added an API for tracking/attributing HoT extensions. The general method is to add a tag to a HoT showing that it's been extended by N seconds, and then when the HoT drops I attribute the healing done by that HoT in its last N seconds. Smart handling is implemented to check for extension clipping due to early refresh, etc... More info found in the comments.

I also implemented all the relevant modules that use extensions, their implementations are pretty straightforward, and leverage HoTTracker's API.

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,1 +1,0 @@
-[0117/190541.817:ERROR:crash_report_database_win.cc(427)] unexpected header
--- a/debug.log
+++ b/debug.log
@@ -1,0 +1,1 @@
+[0117/190541.817:ERROR:crash_report_database_win.cc(427)] unexpected header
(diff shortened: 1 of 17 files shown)


kfinch:
I figure it was just too much to be easily reviewed, it's overall not too complicated what's going on, just needs more documentation. I'll fill that in going forward, as I move towards using this to implement an "early rejuv refresh" module.

MartijnHols:
automated tests based around requirements of the module are the best documentation! 😉

PR #1274, opened 2018-01-27
[Affliction] Checklist updates

sMteX:
Added few things to Affliction Checklist, also added another format for displaying values in Checklist.

EDIT: Also added `valueTooltip` prop to Requirement


src/Parser/Warlock/Affliction/Modules/SoulShards/SoulShardTracker.js
```diff
@@ -3,13 +3,20 @@ import Combatants from 'Parser/Core/Modules/Combatants';
...
 import SPELLS from 'common/SPELLS';
 
+const MAX_SHARDS = 5;
+
 class SoulShardTracker extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
 
+  _current = 3;   // you regenerate 3 shards out of combat so you should be starting combat always with 3 shards
```
Gebuz: If it's possible to manipulate shards prepull somehow this value could be thrown off for the whole encounter. You could get the amount of shards from the energize and cast events like you do with mana in LowMana.js instead of manually counting them to ensure the value is correct.

src/Parser/Warlock/Affliction/Modules/Features/LowMana.js
```diff
@@ -0,0 +1,56 @@
+import Analyzer from 'Parser/Core/Analyzer';
+import RESOURCE_TYPES from 'common/RESOURCE_TYPES';
+
+const MAX_MANA = 1100000;   // should be constant
+const LOW_MANA_THRESHOLD = 0.05;
```
Gebuz: Minor thing, I think it might make more sense to set it to a specific value based on the cost of your spells instead of a percent. If your dots cost 33k per cast, going to 54k mana (<5%) isn't really different from 56k mana (>5%) since you'll only be able to cast 1 more spell anyway. If the threshold was 66k instead it would be exactly enough to cast 2 extra spells.

src/Parser/Warlock/Affliction/Modules/Features/LowMana.js
```diff
@@ -0,0 +1,56 @@
...
+            this._lowManaTimestamp = null;  // reset timestamp so this branch doesn't fire when we're repeatedly > 5%
+            this._currentlyLow = false;
+          }
+        });
+    }
+  }
+
+  get suggestionThresholds() {
+    return {
+      actual: this.timeOnLowMana / 1000,
```
Gebuz: as I mentioned on discord I think you should use percent of time spent on low mana.

src/Parser/Warlock/Affliction/Modules/SoulShards/SoulShardTracker.js
sMteX: Unfortunately no. Cast events don't contain Soul Shards as a resources (`classResources` only contain mana for us) and energize events don't show current amount of shards (they only show `resourceChange` as in how many shards were gained, and `waste`). 

I agree with you that the starting shards can be manipulated prepull (summoning pet for example) but I don't know if there's any reliable way to track how many did you have on fight start. But usually even when you wipe, you summon your pet, use your Rune and then spend 10 seconds getting food buff. So unless you arrive at the boss, summon pet and *immediately* pull the boss, you should have enough time to regenerate the 3rd shard.

src/Parser/Warlock/Affliction/Modules/Features/LowMana.js
sMteX: Fixed, showing the seconds in tooltip instead.

src/Parser/Warlock/Affliction/Modules/Features/LowMana.js
sMteX: I've talked to Motoko about this and basically the 5% threshold was chosen because all of our mana-based spells use at most 5% mana (Haunt), so the threshold is pretty much "to have enough mana for any *one* mana-based spell". I've decided to leave that as a percentage so it can be correct even if mana (and the cost) goes up.

src/Parser/Warlock/Affliction/Modules/SoulShards/SoulShardTracker.js
Gebuz: That's weird that agony soulshard gains show mana instead of souldshards. But the UA cast events does look like they have the info, so you should be able to atleast check on UA cast to make sure the _current number is correct. 

image

sMteX: Fixed, I currently assume the fight started with 3 shards but correctly set that info on nearest shard-spending-spell-cast (which should have the class resources related to Soul Shards). If something goes wrong, I subtract 1 shard as usual.

PR #1275, opened 2018-01-27
[Feral] Ferocious Bite energy check

Anatta336:
_This is a replacement for pull request #1263 after I did something mysterious to that branch which made a mess of the commit history. My apologises again for any disruption._

This is a module to check for Feral druids using Ferocious Bite with less energy than the recommended 50. After a very useful discussion, its suggestion threshold is based on average energy spent by the player on Bite.

It also calculates how much Ferocious Bite damage was lost due to the low energy. This measure is not the same as overall damage lost, as the "saved" energy likely contributed damage in other ways during the fight. But it does provide a general guide to how much (or how little) damage is in play.

biteenergyexample3

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Features/FerociousBiteEnergy.js
+++ b/src/Parser/Druid/Feral/Modules/Features/FerociousBiteEnergy.js
@@ -13,7 +13,9 @@ import { formatNumber } from 'common/format';
 const ENERGY_MIN_USED_BY_BITE = 25;
 const ENERGY_FOR_FULL_DAMAGE_BITE = 50;
 const MAX_DAMAGE_BONUS_FROM_ENERGY = 1.0;
-const LEEWAY_BETWEEN_CAST_AND_DAMAGE = 100; // in thousandths of a second
+const LEEWAY_BETWEEN_CAST_AND_DAMAGE = 500; // in thousandths of a second
+
+const debug = false;
 
 /**
  * Although Ferocious Bite costs 25 energy, it does up to double damage if the character has more.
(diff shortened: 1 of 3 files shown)


src/Parser/Druid/Feral/Modules/Features/FerociousBiteEnergy.js
```diff
@@ -0,0 +1,160 @@
...
+      this.lowEnergyBiteCount++;
+      this.lowEnergyBiteCastEvents.push(event);
+    }
+    else {
+      this.energySpentOnBiteTotal += ENERGY_FOR_FULL_DAMAGE_BITE;
+    }
+  }
+
+  on_finished() {
+    this.lowEnergyBiteCastEvents.forEach((castEvent) => {
```
emallson: Is there a reason you're searching through the bite events after the fact rather than doing it during parsing? Can bite damage events occur before casts? If not, then you can simply:

1. record the energy spent on the bite in `on_byPlayer_cast`
2. calculate the damage lost using that number in `on_byPlayer_damage`

This avoids doing the extra searching at the end. Admittedly, it isn't *expensive* but it is unnecessary and unconventional.

Anatta336:
Good point, thank you.
In all the logs I've tested the cast event is received before the damage event (some were quite delayed, so upped the `LEEWAY_BETWEEN_CAST_AND_DAMAGE`). As suggested I've removed the `on_finished` and now do that processing during parsing. I've put in a check for out-of-order events just in case.

src/Parser/Druid/Feral/Modules/Features/FerociousBiteEnergy.js
```diff
@@ -32,77 +34,56 @@ class FerociousBiteEnergy extends Analyzer {
...
-
-    timestampedArray.forEach((item) =>{
-      const difference = Math.abs(item.timestamp - targetTime);
-      if (difference < closestDifference) {
-        closest = item;
-        closestDifference = difference;
-      }
-    });
-    return closest;
+  getEnergyUsedByBite(event) {
```
emallson: Minor style comment: this should probably be private (prefixed with `_`) and you can omit the `get` as the first word in the name, e.g. `_energyUsedByBite` or `_biteEnergySpent`.

PR #1280, opened 2018-01-27
[BrM Monk] Mastery Evaluation

emallson:
This PR adds a statistic for Brewmasters that shows both an estimate of the actual damage mitigated by [their mastery](www.wowhead.com/spell=117906/mastery-elusive-brawler) and the expected damage that would be mitigated if they repeated this fight a bunch of times.

Much of the code is dedicated to that latter part, which I opted to highlight over the actual because it depends only on player performance and gear, and not on RNG (if you didn't click the link, brewmaster mastery boosts dodge chance).

The math for calculating the expectation is relatively complicated, but I tried to comment well enough that it is reasonably clear what is going on. There are a couple of improvements that could tighten things up but aren't essential for giving useful output. I'd be fine with merging this without addressing either.

- [ ] Find the formula for the base (pre-mastery) dodge chance. I've reached out to the good fellows in `#brewcraft` but haven't gotten anything yet, and haven't tried to reverse-engineer it myself yet. Due to diminishing returns, the value of 15% is going to be reasonably close to the true value (fwiw my monk has 19% base dodge). The base dodge chance is mostly drowned out by mastery at this point in the expansion (each stack on my monk atm gives almost 40% dodge).
- [ ] Figure out how to determine mob level. We don't currently have any sub-boss-level mobs that are dangerous enough to worry about the extra 1.5-3% dodge chance, but it may become relevant in the future.

it looks like this

### DIFF
--- a/src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
@@ -119,9 +119,8 @@ class MasteryValue extends Analyzer {
   _hitCounts = {};
   _dodgeCounts = {};
 
-  get baseDodge() {
-    return 0.15;
-  }
+  // in the future, this will be determined based on player agility
+  baseDodge = 0.15;
 
   dodgePenalty(_source) {
     return 0.045; // 1.5% per level, bosses are three levels over players. not sure how to get trash levels yet -- may not matter
--- a/src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
+++ b/src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
@@ -124,13 +124,14 @@ class MasteryValue extends Analyzer {
   }
 
   dodgePenalty(_source) {
-    return 0.045; // 1.5% per level, bosses are three levels over players. not sure how to get trash levels yet
+    return 0.045; // 1.5% per level, bosses are three levels over players. not sure how to get trash levels yet -- may not matter
   }
 
   // returns the current chance to dodge a damage event assuming the
   // event is dodgeable
-  dodgeChance(masteryStacks, event) {
-    return (event._masteryPercentage || this.stats.currentMasteryPercentage) * masteryStacks + this.baseDodge - this.dodgePenalty(event.sourceID);
+  dodgeChance(masteryStacks, masteryRating, sourceID) {
+    const masteryPercentage = this.stats.masteryPercentage(masteryRating, true);
+    return masteryPercentage * masteryStacks + this.baseDodge - this.dodgePenalty(sourceID);
   }
 
 
@@ -141,7 +142,6 @@ class MasteryValue extends Analyzer {
   }
 
   on_toPlayer_damage(event) {
-    event._masteryPercentage = this.stats.currentMasteryPercentage;
     event._masteryRating = this.stats.currentMasteryRating;
     if(event.hitType === HIT_TYPES.DODGE) {
       this._addDodge(event);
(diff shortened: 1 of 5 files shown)


src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
```diff
@@ -0,0 +1,298 @@
...
+    stats: StatTracker,
+  };
+
+  _dodgeableSpells = {};
+  _timeline = [];
+  _hitCounts = {};
+  _dodgeCounts = {};
+
+  get baseDodge() {
+    return 0.15;
```
kfinch: Not clear on why this is in a getter instead of as a static const.

src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  dodgePenalty(_source) {
+    return 0.045; // 1.5% per level, bosses are three levels over players. not sure how to get trash levels yet -- may not matter
```
kfinch: I take it this is for future proofing, so can implement formula based on source level if it becomes available?

src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
```diff
@@ -0,0 +1,298 @@
...
+    return this._expectedValues.meanExpectedDodge;
+  }
+
+  get actualDodgeRate() {
+    const totalHits = Object.keys(this._dodgeableSpells).reduce((sum, spellId) => sum + (this._hitCounts[spellId] || 0), 0);
+    const totalDodges = Object.keys(this._dodgeableSpells).reduce((sum, spellId) => sum + this._dodgeCounts[spellId], 0);
+    return totalDodges / (totalHits + totalDodges);
+  }
+
+  get estimatedActualMitigation() {
```
kfinch: I see the main metric you're using as a point of comparison between 'no mastery', 'expected', and 'actual' is Reduced DTPS. I understand that you need to use some number in order to produce an effective comparison, but I don't like Reduced DTPS.

Due to wildly differing tank damage done by bosses (and differing times that the Monk will actually have aggro), the Reduced DTPS numbers seem like they are only meaningful in relation to each other. There's also the inaccuracy added by occasionally having to guess the amount mitigated. Instead, why not just report the dodge rate, again comparing 'no mastery', 'expected', and 'actual'? It will be an easier number to read, and will require less guesswork.

src/Parser/Monk/Brewmaster/Modules/Normalizers/GarothiWorldbreakerMelee.js
```diff
@@ -0,0 +1,37 @@
+import EventsNormalizer from 'Parser/Core/EventsNormalizer';
+import HIT_TYPES from 'Parser/Core/HIT_TYPES';
+import SPELLS from 'common/SPELLS';
+
+const GAROTHI_WORLDBREAKER = 2076;
```
kfinch: Arguably you could import https://github.com/WoWAnalyzer/WoWAnalyzer/blob/9e55fec255bac851822c2ce9636429ad2d963399/src/Raids/AntorusTheBurningThrone/GarothiWorldbreaker.js instead of making a new const, but as this number will never change I'm not sure it will ever matter.

src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
emallson: I was hoping the other brm theorycrafters would have the formula for it (its a function of agility). Turns out nobody bothered computing it for legion because it hasn't really mattered. For right now I'm going to change it to a const and then when I get the formula i'll revisit it.

src/Parser/Monk/Brewmaster/Modules/Core/MasteryValue.js
emallson: Dodge rate doesn't tell the whole story either. Some bosses have dangerous mechanics that can be dodged or have very hard-hitting melees, other bosses don't (see, for example, Starboi, on whom brm mastery does almost 0 mitigation).

You're right that DTPS only makes sense in comparison to another number. However, in this case we have another number available from the Damage Taken module. From the fight in the PR:

screenshot from 2018-01-29 16-47-13

so mastery reduced total DTPS by ~1/3 (disregarding all the complexity that stagger brings into that analysis).

src/Parser/Monk/Brewmaster/Modules/Normalizers/GarothiWorldbreakerMelee.js
emallson: Missed that module, I'll import just to keep everything in one place.

emallson:
Just pushed with revisions based on your comments. I've looked at every fight in antorus on at least a couple different logs (except Eonar, because 🤷‍♂️ ) and everything looks ok. The estimated dodge rate is typically within a couple % of the actual dodge rate, and when it isn't the difference can be pretty clearly explained by the fact that my `baseDodge` is a constant and doesn't take into account the extra dodge from agility (the large differences are almost exclusively on logs of 960+ equipped players that should have close to an extra 5% dodge from agility after diminishing returns). I'll be getting data for that this week (probably) so I should be able to address that relatively soon.

emallson:
I actually just got data together to get the true base dodge chance so don't merge til I push that

PR #1282, opened 2018-01-27
Updated ColdHeartEfficiency

Bicepspump:
Added tracking of concordance and khazgoroth shaping.
Updated a criteria for holding at 20 stacks for more than 4 seconds.
Changed the minimum stacks for an allowed cast to 14.

Most of these changes are marginal dps increases but some players track these buffs. Giving them failed casts would be wrong and I therefore implemented these new changes.

### DIFF
--- a/src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
+++ b/src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
@@ -9,6 +9,16 @@ import { formatPercentage } from 'common/format';
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
 import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+  
+const unholyStrengthDuration = 15000;
+const concordanceDuration = 10000;
+const khazgorothDuration = 15000;
+  
+const coldHeartMaxStack = 20;
+const coldHeartOptimalMinStack = 14;
+	
+const remainingDurationAllowed = 4000;
+const maxDurationAtMaxStacksAllowed = 4000;
 
 class ColdHeartEfficiency extends Analyzer {
   static dependencies = {
@@ -20,7 +30,7 @@ class ColdHeartEfficiency extends Analyzer {
 
   totalColdHeartCasts = 0;
   correctColdHeartCasts = 0;
-  buffColdHeart = 20; //Stacks start at 20 at start of fight
+  buffColdHeart = coldHeartMaxStack; //Stacks start at 20 at start of fight
   
   unholyStrengthStart = 0;
   unholyStrengthRemaining = 0;
@@ -33,16 +43,6 @@ class ColdHeartEfficiency extends Analyzer {
   
   timeAtMaxStacksStart = this.owner.fight.start_time;
   timeAtMaxStacks = 0;
-  
-  unholyStrengthDuration = 15000;
-  concordanceDuration = 10000;
-  khazgorothDuration = 15000;
-  
-  coldHeartMaxStack = 20;
-  coldHeartOptimalMinStack = 14;
-	
-  remainingDurationAllowed = 4000;
-  maxDurationAtMaxStacksAllowed = 4000;
   
   castsTooEarly = 0;
   castsTooLate = 0;
@@ -91,7 +91,7 @@ class ColdHeartEfficiency extends Analyzer {
     const spellID = event.ability.guid;
     if (spellID === SPELLS.COLD_HEART_BUFF.id) {
       this.buffColdHeart = event.stack;
-	  if (this.buffColdHeart == 20) {
+	  if (this.buffColdHeart === coldHeartMaxStack) {
 		  this.timeAtMaxStacksStart = event.timestamp;
 	  }
     }
@@ -104,22 +104,21 @@ class ColdHeartEfficiency extends Analyzer {
 		this.totalColdHeartCasts++;
 	  
 		this.timeAtMaxStacks = event.timestamp - this.timeAtMaxStacksStart;
-	
-		this.unholyStrengthRemaining = this.unholyStrengthDuration - (event.timestamp - this.unholyStrengthStart);
-		this.concordanceRemaining = this.concordanceDuration - (event.timestamp - this.concordanceStart);
-		this.khazgorothRemaining = this.khazgorothDuration - (event.timestamp - this.khazgorothStart);
+
+		this.unholyStrengthRemaining = unholyStrengthDuration - (event.timestamp - this.unholyStrengthStart);
+		this.concordanceRemaining = concordanceDuration - (event.timestamp - this.concordanceStart);
+		this.khazgorothRemaining = khazgorothDuration - (event.timestamp - this.khazgorothStart);
 	  
 		//This checks wether or not any of the three buffs are about to fall off within the next 4 seconds.
-		if ((this.unholyStrengthRemaining > 0 && this.unholyStrengthRemaining < this.remainingDurationAllowed) || (this.concordanceRemaining > 0 && this.concordanceRemaining < this.remainingDurationAllowed) || (this.khazgorothRemaining > 0 && this.khazgorothRemaining < this.remainingDurationAllowed)){
-			if (this.buffColdHeart < this.coldHeartMaxStack && this.buffColdHeart >= this.coldHeartOptimalMinStack) {
+		if ((this.unholyStrengthRemaining > 0 && this.unholyStrengthRemaining < remainingDurationAllowed) || (this.concordanceRemaining > 0 && this.concordanceRemaining < remainingDurationAllowed) || (this.khazgorothRemaining > 0 && this.khazgorothRemaining < remainingDurationAllowed)){
+			if (this.buffColdHeart < coldHeartMaxStack && this.buffColdHeart >= coldHeartOptimalMinStack) {
 				this.correctColdHeartCasts++;
-			}else if (this.buffColdHeart < 20) {
-				this.castsTooEarly++;
 			}
-		}			
-		if (this.buffColdHeart === 20 && this.timeAtMaxStacks <= this.maxDurationAtMaxStacksAllowed) {
+		}else if (this.buffColdHeart < coldHeartMaxStack) {
+			this.castsTooEarly++;
+		}else if (this.buffColdHeart === coldHeartMaxStack && this.timeAtMaxStacks <= maxDurationAtMaxStacksAllowed) {
 			this.correctColdHeartCasts++;
-		}else if(this.buffColdHeart === 20){
+		}else if(this.buffColdHeart === coldHeartMaxStack){
 			this.castsTooLate++;
 		}
 		//This is only here for double casting Cold Heart. If Cold Heart is casted again before it reaches 2 stacks, the event won't update.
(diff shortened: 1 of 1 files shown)


src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
     const unholyStrengthDuration = 15000;
+	const concordanceDuration = 10000;
+	const khazgorothDuration = 15000;
+	
     const coldHeartMaxStack = 20;
-    const coldHeartOptimalMinStack = 17;
-    const remainingUnholyStrengthAllowed = 4000;
+    const coldHeartOptimalMinStack = 14;
+	
+    const remainingDurationAllowed = 4000;
```
kfinch: The above consts should all be set as class constants at the top of this file, because they don't ever change.

I'm talking here about `remainingDurationAllowed` `coldHeartOptimalMinStack` `coldHeartMaxStack` `khazgorothDuration` `concordanceDuration` `unholyStrengthDuration`

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
-      if (this.unholyStrengthRemaining > 0) {
-        if (this.unholyStrengthRemaining < remainingUnholyStrengthAllowed) {
-          if (this.buffColdHeart < coldHeartMaxStack) {
-            if (this.buffColdHeart >= coldHeartOptimalMinStack) {
-              this.correctColdHeartCasts++;
-            }
-          }
-        }
+	  //This checks wether or not any of the three buffs are about to fall off within the next 4 seconds.
+      if ((this.unholyStrengthRemaining > 0 & this.unholyStrengthRemaining < remainingDurationAllowed) || (this.concordanceRemaining > 0 & this.concordanceRemaining < remainingDurationAllowed) || (this.khazgorothRemaining > 0 & this.khazgorothRemaining < remainingDurationAllowed) & this.buffColdHeart < coldHeartMaxStack & this.buffColdHeart >= coldHeartOptimalMinStack) {
```
kfinch: It looks like you're using `&` for logical-and here, should be `&&`

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
+         this.correctColdHeartCasts++;
       }
-      if (this.buffColdHeart === 20) {
+      if (this.buffColdHeart === 20 & this.timeAtMaxStacks <= 4000) {
```
kfinch: It looks like you're using `&` for logical-and here, should be `&&`

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
   unholyStrengthRemaining = 0;
+  
+  concordanceStart = 0;
+  concordanceRemaining = 0;
+  
+  khazgorothStart = 0;
+  khazgorothRemaining = 0;
+  
+  timeAtMaxStacksStart = 0;
+  timeAtMaxStacks = 0;
```
kfinch: Several of the above variables are only assigned and then immediately used in the `on_byPlayer_cast` block. As such, they should be defined in that block (as consts) instead of here. 

I'm talking about `timeAtMaxStacks` `khazgorothRemaining` `concordanceRemaining` `unholyStrengthRemaining`

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
       this.buffColdHeart = event.stack;
+	  if (this.buffColdHeart === 20) {
+		  this.timeAtMaxStacksStart = event.timestamp;
+	  }
     }
   }
 
   on_byPlayer_cast(event) {
     const spellId = event.ability.guid;
+
```
jjs451: Add a check here to see if chains of ice is being cast: `if (spellId !== SPELLS.CHAINS_OF_ICE.id) { return; }`  This saves the analyzer from executing the whole function for every cast event

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
jjs451: Couple of notes, first make sure to use `&&`.  `&` is a bitwise operator, `&&` is the logical operator.  This should also be broken into two if statements to make it more readable.  Something `if (this.buffColdHeart < coldHeartMaxStack & this.buffColdHeart >= coldHeartOptimalMinStack)` on one line and then under that do the rest.  It'll make it much easier to see whats going at a glance.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -21,49 +21,93 @@ class ColdHeartEfficiency extends Analyzer {
...
+	  }
     }
   }
 
   on_byPlayer_cast(event) {
     const spellId = event.ability.guid;
+	
     const unholyStrengthDuration = 15000;
+	const concordanceDuration = 10000;
+	const khazgorothDuration = 15000;
```
jjs451: You can move these consts up to near line 20 where you declare the other variables.  Since the values never change they don't need to be declared on every cast.

Gebuz:
2 things I noticed that might be worth fixing aswell:
* Chains of Ice current has a cooldown set to 40s in cast efficiency which it probably shouldn't have if this module covers Cold Heart?
* Cold Heart shows 0 dmg done as Frost at the moment. I'm guessing it uses a different spell ID.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
...
+	if (spellID === SPELLS.KHAZGOROTHS_SHAPING.id) {
+		this.khazgorothStart = 0;
+	}
   }
 
   on_byPlayer_applybuffstack(event) {
     const spellID = event.ability.guid;
     if (spellID === SPELLS.COLD_HEART_BUFF.id) {
       this.buffColdHeart = event.stack;
+	  if (this.buffColdHeart == 20) {
```
kfinch: This is causing the build to fail, double equals isn't allowed by our build rules, must be triple equals.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
 
   totalColdHeartCasts = 0;
   correctColdHeartCasts = 0;
-  buffColdHeart = 0;
+  buffColdHeart = 20; //Stacks start at 20 at start of fight
+  
   unholyStrengthStart = 0;
   unholyStrengthRemaining = 0;
```
kfinch: My previous comment still stands after the changes, `unholyStrengthRemaining` `concordanceRemaining`  `khazgorothRemaining` and `timeAtMaxStacks` all should not be class variables because they are only set and immediately used within one function.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
...
+  
+  unholyStrengthDuration = 15000;
+  concordanceDuration = 10000;
+  khazgorothDuration = 15000;
+  
+  coldHeartMaxStack = 20;
+  coldHeartOptimalMinStack = 14;
+	
+  remainingDurationAllowed = 4000;
+  maxDurationAtMaxStacksAllowed = 4000;
```
kfinch: All 7 of the above constants should be static consts, not class variables. For example this one should be `const MAX_DURATION_AT_MAX_STACKS_ALLOWED = 4000`, defined like that above the class declaration.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
...
+		if ((this.unholyStrengthRemaining > 0 && this.unholyStrengthRemaining < this.remainingDurationAllowed) || (this.concordanceRemaining > 0 && this.concordanceRemaining < this.remainingDurationAllowed) || (this.khazgorothRemaining > 0 && this.khazgorothRemaining < this.remainingDurationAllowed)){
+			if (this.buffColdHeart < this.coldHeartMaxStack && this.buffColdHeart >= this.coldHeartOptimalMinStack) {
+				this.correctColdHeartCasts++;
+			}else if (this.buffColdHeart < 20) {
+				this.castsTooEarly++;
+			}
+		}			
+		if (this.buffColdHeart === 20 && this.timeAtMaxStacks <= this.maxDurationAtMaxStacksAllowed) {
+			this.correctColdHeartCasts++;
+		}else if(this.buffColdHeart === 20){
```
kfinch: These should be use the cold heart max stack const rather than hardcoded '20'

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
...
     when(castEfficiency).isLessThan(0.8)
       .addSuggestion((suggest, actual, recommended) => {
-        return suggest(<span>You are casting <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at non optimal times. You either want to cast <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> when at 20 stacks of <SpellLink id={SPELLS.COLD_HEART_BUFF.id} /> or when you are above 16 stacks and your buff <SpellLink id={SPELLS.UNHOLY_STRENGTH_BUFF.id} /> is about to run out.</span>)
+        return suggest(<span> You are casting <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at non optimal times. {this.castsTooLate} cast(s) were made too late and {this.castsTooEarly} cast(s) were made too early. You either want to cast <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> when at 20 stacks of <SpellLink id={SPELLS.COLD_HEART_BUFF.id} /> ASAP, or when you are above 13 stacks and any of your buffs <SpellLink id={SPELLS.UNHOLY_STRENGTH_BUFF.id}/> or <SpellLink id={SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_STRENGTH.id} />  or <SpellLink id={SPELLS.KHAZGOROTHS_SHAPING.id} /> are about to run out. You also don't want to hold <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at 20 stacks for too long.</span>)
```
kfinch: Use `<Wrapper>` pls

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +20,117 @@ class ColdHeartEfficiency extends Analyzer {
...
+	  
+		//This checks wether or not any of the three buffs are about to fall off within the next 4 seconds.
+		if ((this.unholyStrengthRemaining > 0 && this.unholyStrengthRemaining < this.remainingDurationAllowed) || (this.concordanceRemaining > 0 && this.concordanceRemaining < this.remainingDurationAllowed) || (this.khazgorothRemaining > 0 && this.khazgorothRemaining < this.remainingDurationAllowed)){
+			if (this.buffColdHeart < this.coldHeartMaxStack && this.buffColdHeart >= this.coldHeartOptimalMinStack) {
+				this.correctColdHeartCasts++;
+			}else if (this.buffColdHeart < 20) {
+				this.castsTooEarly++;
+			}
+		}			
+		if (this.buffColdHeart === 20 && this.timeAtMaxStacks <= this.maxDurationAtMaxStacksAllowed) {
```
kfinch: While I understand by how you've built the if statements here that it shouldn't be possible to hit both this 'if' and the above 'if', you should make this explicit by making this an `else if`. I spent a few minutes trying to decipher how these blocks worked, and it would have been a lot easier with the explicit `else if`.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +30,106 @@ class ColdHeartEfficiency extends Analyzer {
...
     when(castEfficiency).isLessThan(0.8)
       .addSuggestion((suggest, actual, recommended) => {
-        return suggest(<span>You are casting <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at non optimal times. You either want to cast <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> when at 20 stacks of <SpellLink id={SPELLS.COLD_HEART_BUFF.id} /> or when you are above 16 stacks and your buff <SpellLink id={SPELLS.UNHOLY_STRENGTH_BUFF.id} /> is about to run out.</span>)
+        return suggest(<wrapper> You are casting <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at non optimal times. {this.castsTooLate} cast(s) were made too late and {this.castsTooEarly} cast(s) were made too early. You either want to cast <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> when at 20 stacks of <SpellLink id={SPELLS.COLD_HEART_BUFF.id} /> ASAP, or when you are above 13 stacks and any of your buffs <SpellLink id={SPELLS.UNHOLY_STRENGTH_BUFF.id}/> or <SpellLink id={SPELLS.CONCORDANCE_OF_THE_LEGIONFALL_STRENGTH.id} />  or <SpellLink id={SPELLS.KHAZGOROTHS_SHAPING.id} /> are about to run out. You also don't want to hold <SpellLink id={SPELLS.CHAINS_OF_ICE.id} /> at 20 stacks for too long.</wrapper>)
```
kfinch: It's `<Wrapper>`, and you'll need to `import Wrapper from 'common/Wrapper';` at the top... as written I'm pretty sure this will crash.

src/Parser/DeathKnight/Unholy/Modules/Items/ColdHeartEfficiency.js
```diff
@@ -20,63 +30,106 @@ class ColdHeartEfficiency extends Analyzer {
 
   totalColdHeartCasts = 0;
   correctColdHeartCasts = 0;
-  buffColdHeart = 0;
+  buffColdHeart = coldHeartMaxStack; //Stacks start at 20 at start of fight
+  
   unholyStrengthStart = 0;
   unholyStrengthRemaining = 0;
```
kfinch: My previous comment still stands after the changes, `unholyStrengthRemaining` `concordanceRemaining` `khazgorothRemaining` and `timeAtMaxStacks` all should not be class variables because they are only set and immediately used within one function.

Please delete them here, and then below in `on_byPlayer_cast` (starting at line 106), define them with `const` as you set them.

kfinch:
Sorry for the odyssey this was, and thanks for the changes.

PR #1285, opened 2018-01-27
[Havoc] Channeling

CollCrom:
Made eye beam appear as a channeled ability. I know it looks weird I look for both the cast and the applybuff event but one caught eye beams during meta and one caught eye beams outside of meta. Also had to normalize the demonic meta buff to get this to work.

image


src/Parser/DemonHunter/Havoc/Normalizers/EyeBeam.js
```diff
@@ -0,0 +1,39 @@
...
+		const fixedEvents = [];
+		events.forEach((event, eventIndex) => {
+			fixedEvents.push(event);
+
+			if(event.type === 'applybuff' && event.ability.guid === SPELLS.EYE_BEAM.id) {
+				const castTimestamp = event.timestamp;
+
+				for (let previousEventIndex = eventIndex; previousEventIndex >= 0; previousEventIndex -= 1) {
+          const previousEvent = fixedEvents[previousEventIndex];
+          if ((castTimestamp - previousEvent.timestamp) > 50) { // the max delay between the heal and cast events never looks to be more than this.
```
kfinch: Looks like a copy-pasta'd comment that doesn't make sense any more.

src/Parser/DemonHunter/Havoc/Normalizers/EyeBeam.js
```diff
@@ -0,0 +1,39 @@
+import SPELLS from 'common/SPELLS';
+
+import EventsNormalizer from 'Parser/Core/EventsNormalizer';
+
+class EyeBeam extends EventsNormalizer {
+	/**
+	* The applybuff from demoic is logges before the cast of Eye Beam. This makes it harder to deal with so reorder it.
```
kfinch: A couple typos here, `demoic` and `logges`

Also I think it's very important to be as clear as possible about what the Normalizer is going to do, in this case something like "This normalizes events so that the Eye Beam applybuff always comes _before_ the Meta Havoc buff" (assuming what I just wrote is a correct summary of what this does)

PR #1286, opened 2018-01-28
[Havoc] Eye beam CDR

CollCrom:
Added in the talent and trait that account for all eye beam cooldown reduction.

image

### DIFF
--- a/src/Parser/DemonHunter/Havoc/CHANGELOG.js
+++ b/src/Parser/DemonHunter/Havoc/CHANGELOG.js
@@ -9,6 +9,11 @@ import { Hewhosmites, Mamtooth } from 'MAINTAINERS';
 
 export default [
 	{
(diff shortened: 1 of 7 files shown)


CollCrom:
The talent turns souls picked up into energize events which always get picked up in the log. Not every heal event from the souls gets picked up.

https://www.warcraftlogs.com/reports/T68JHYX7faDPW3jv#fight=18&type=healing&source=67
33 heal events

Vs

https://www.warcraftlogs.com/reports/T68JHYX7faDPW3jv#fight=18&type=resources&source=67&spell=117
1295/35 = 37 energize events

kfinch:
Does this mean that if the player doesn't have the talent, some CDR will be missed because of this?

CollCrom:
Good question. I'll try and find some logs to test this

PR #1288, opened 2018-01-28
[Havoc] Tier 21 4PC

CollCrom:
image
image

The ret suggestion is a thing I meant to change a while ago but forgot.

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Items/Tier21_4set.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Items/Tier21_4set.js
@@ -1,0 +1,77 @@
+import React from 'react';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import SpellIcon from 'common/SpellIcon';
+import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
+import Wrapper from 'common/Wrapper';
+import { formatPercentage, formatNumber } from 'common/format';
+
+class Tier21_4set extends Analyzer {
+	static dependencies = {
+		combatants: Combatants,
+		abilityTracker: AbilityTracker,
+	};
+
+	betrayersFuryBuffCount = 0;
+
+	on_initialized() {
+		this.active = this.combatants.selected.hasBuff(SPELLS.HAVOC_T21_4PC_BONUS.id);
+	}
+
+	get eyeBeamCasts() {
+		return this.abilityTracker.getAbility(SPELLS.EYE_BEAM.id).casts;
+	}
+
+	on_byPlayer_applybuff (event) {
+		const spellId = event.ability.guid;
+		if(spellId !== SPELLS.HAVOC_T21_4PC_BUFF.id) {
+			return;
+		}
+		this.betrayersFuryBuffCount++;
+	}
+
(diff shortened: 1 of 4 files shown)


src/Parser/Paladin/Retribution/Modules/Items/WhisperOfTheNathrezim.js
```diff
@@ -78,7 +78,7 @@ class WhisperOfTheNathrezim extends Analyzer {
     when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
       return suggest(<Wrapper>Your usage of <ItemLink id={ITEMS.WHISPER_OF_THE_NATHREZIM.id} icon/> can be improved. Make sure to save up five holy power before your next <SpellLink id={SPELLS.JUDGMENT_CAST.id} icon/> window to get more time on the buff.</Wrapper>)
         .icon(ITEMS.WHISPER_OF_THE_NATHREZIM.icon)
-        .actual(`${formatPercentage(actual)}% damage contributed`)
+        .actual(<Wrapper>{formatPercentage(actual)}% of spenders with <SpellLink id={SPELLS.WHISPER_OF_THE_NATHREZIM_BUFF.id} icon/></Wrapper>)
```
kfinch: What's with the random Ret change? Was this intended?

src/Parser/DemonHunter/Havoc/Modules/Items/Tier21_4set.js
```diff
@@ -0,0 +1,77 @@
...
+		this.betrayersFuryBuffCount++;
+	}
+
+	get suggestionThresholds() {
+    return {
+      actual: 1 - this.betrayersFuryBuffCount / this.eyeBeamCasts,
+      isGreaterThan: {
+        minor: 0,
+        average: 0.05,
+        major: 0.1,
```
kfinch: You could disagree, but I think these thresholds are way too tight. Consider that on most normal length fights, the player will Eye Beam less than 10 times, and so clipping even one would immediately be a major issue.

I know "Havoc players don't do mechanics" is a meme, but I could see some occasional times where player might be forced to interrupt early to avoid boss mechanics. It's probably a failure on the player's part to not plan ahead for that, but is it major issue bad if it happens only once? I like to conceptualize thresholds as minor = "the player was less than perfect", average = "the player made a clear error", major = "does the player even understand this mechanic???"

src/Parser/Paladin/Retribution/Modules/Items/WhisperOfTheNathrezim.js
CollCrom: I changed the suggestion from % damage contributed to % of spenders with the buff from the legendary a while back but forgot to switch the text

kfinch:
Also CHANGELOG conflict lul

src/Parser/DemonHunter/Havoc/Modules/Items/Tier21_4set.js
CollCrom: I think its a minor issue if they clip one of the buffs but the thresholds could definitely be extended out a little. I think the fights I'm basing this off of have little movement which means there is less reason to clip. DH's also average 2-3 casts per minute now on Eye Beam. So clipping one on a 4 minute fight would be a minor/average issue

kfinch: Oh I didn't realize the CD was compressed so much, but yeah the outline of my point stands.

kfinch:
That commit says it changed suggestions, but did not.

Issue #1273, opened 2018-01-26
Changelog date is off by one day


CollCrom:
![image](https://user-images.githubusercontent.com/19510201/35462239-bc1949ee-02b0-11e8-8d7e-219a10ed57e9.png)
  {
		date: new Date('2018-01-13'),
		changes: 'Updated to new checklist format,',
		contributors: [Hewhosmites],
	},
	{
		date: new Date('2017-10-22'),
		changes: 'Changed the active time icon to something more befitting.',
		contributors: [Zerotorescue],
	},
	{
		date: new Date('2017-01-18'),
		changes: 'Added in the Tier 21 Modules',
		contributors: [Hewhosmites],
	},
	{
		date: new Date('2017-09-14'),
		changes: 'Added a Blessing of the Ashbringer Module.',
		contributors: [Hewhosmites],
	},

Anatta336:
Viewing the site from my machine the dates appear correct. e.g. "Updated to new checklist format" on the Retribution changelog shows as 1/13/2018.

My computer's clock is set equal to UTC (Coordinated Universal Time). I believe the dates are assumed to be stated in UTC, and get automatically converted to local timezone somewhere along the line. If so any timezone west of London would see the dates offset by one. That's supported by setting my computer's timezone to UTC-1 and reloading the page, which now displays the error you see.

It seems like this should be fixable with `date.toLocaleDateString(undefined, { timezone: 'UTC' })`. But that doesn't seem to work, perhaps because the `date` objects we deal with have their `.timezone` left as `undefined.` (There looks to be a whole mess of conflicting stuff about how `Date` objects created without specified timezones should be handled.)

It's less elegant but constructing the formatted date from `.getUTCMonth()` (January is month 0, so with a +1 offset) `.getUTCDate()` and `.getUTCFullYear()` works to get consistent results regardless of what timezone my computer is set to.

I'll submit a pull request with this one-line change.

MartijnHols:
A fallback option would be to change the changelog-provided dates to strings, but I'd prefer avoiding that.

I think we should be able to override the timezone without changing the date/time. I think that would be the best solution. An alternative that is kind of a hack would be to just add 12 hours to the date.

Anatta336:
I was mistaken! `toLocaleDateString()` can still be used, I just wasn't passing the timezone correctly. It's a `timeZone` not a `timezone`.

Gebuz:
fixed in #1294

[...]

PR #1293, opened 2018-01-28
Changelog off-by-one error in some timezones

Anatta336:
See issue #1273 
This change forces the date to be displayed in terms of UTC rather than attempting to convert to a user's local timezone, which was showing different dates depending on the viewer's location.
For consistency the date is still displayed in the US style (month/day/year), matching the defaults that were used by `date.toLocaleDateString()`.


MartijnHols:
Sorry, but we can not merge this. We need `toLocaleDateString` so that dates are shown in the user's native date format. I am confident this issue can be resolved without removing that.

PR #1305, opened 2018-01-29
Added most common haste proc trinkets to stat tracker

Gebuz:
Missing Khaz'goroth's Shaping which for some reason uses the same buff ID for all 4 secondary procs and I didn't have time to implement it tonight.


kfinch:
@Gebuz by the way, if Khaz'goroth's Shaping can't be solved using StatTracker's base API, I left an opening for handling strange buffs using the "forceChangeStats" function. You can see it used here: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/Modules/Items/Legion/TombOfSargeras/CharmOfTheRisingTide.js

Gebuz:
cheers, will take a look at it

PR #1307, opened 2018-01-30
[Havoc] Delusions of Grandeur Suggestion

CollCrom:
capture
https://www.warcraftlogs.com/reports/yqKaGptwXPRYr8zm#fight=6&type=summary&source=19

Updated abilites.js to have the isOnGCD value. Added in a suggestion if you are wearing shoulders and didn't get an extra cast of Metamorphosis

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
@@ -39,7 +39,7 @@ class DelusionsOfGrandeur extends Analyzer {
 	}
 
 	get metaCooldownWithShoulders(){
-		return this.metaCooldown * this.cooldownReductionRatio || 0;
+		return this.metaCooldown * this.cooldownReductionRatio || 1;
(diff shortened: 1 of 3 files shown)


src/Parser/DemonHunter/Havoc/Modules/Items/DelusionsOfGrandeur.js
```diff
@@ -40,8 +42,25 @@ class DelusionsOfGrandeur extends Analyzer {
 		return this.metaCooldown * this.cooldownReductionRatio || 0;
 	}
```
Pewtro: This should probably be || 1 instead so it doesn't return a 0 minute cooldown with shoulders, just in case.

CollCrom:
I don't think those situations really come into play. At 3:45 you would get a second meta without shoulders but you would have already had a whole meta duration if youre wearing them. Even though you get two casts regardless. It is still beneficial to wear them in that case.

Pewtro:
All fair, then I think it should just be the ``|| 1`` instead of ``|| 0`` 👌

PR #1309, opened 2018-01-30
Added Khazgoroth's Courage stat tracking

Gebuz:
I know _makeStatChange is ugly, idk why I can't make the statChangeObj with the string :(


src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/KhazgorothsCourage.js
```diff
@@ -54,6 +80,36 @@ class KhazgorothsCourage extends Analyzer {
...
+    } else if(crit >= mastery && crit >= versatility){
+      return 'crit';
+    } else if(mastery >= versatility){
+      return 'mastery';
+    } else{
+      return 'versatility';
+    }
+  }
+
+  _makeStatChange(ratingChange, statName, eventReason) {
```
kfinch: You can make an object field programatically with [ ], so here you could skip the if/else by just doing `const statChangeObj = { [statName]: ratingChange };`

Gebuz: Cheers, fixed now.

PR #1310, opened 2018-01-30
[Death Knight] Added rune speed up to RuneTracker and added it to unholy

Gebuz:
The runetracker now accounts for the rune regeneration acceleration effects Runic Corruption and Rune Master (4p21 blood). This means it should now also be accurate for unholy.

Also fixed a few bugs I encountered:
* holyshock cooldown wasn't properly checking for sanctified wrath
* Chronoshard was using wrong spellID (It was already in the file when I added it yesterday so I didn't check if the ID was correct)


src/Parser/DeathKnight/Blood/Modules/Abilities.js
```diff
@@ -265,21 +265,30 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.RUNE_1,
         category: Abilities.SPELL_CATEGORIES.HIDDEN,
-        cooldown: haste => 10 / (1 + haste),
+        cooldown: (haste, combatantCurrent) => {
+          const multiplier = combatantCurrent.hasBuff(SPELLS.RUNE_MASTER.id) ? 0.4 : 0;
+          return 10 / (1 + haste) / (1 + multiplier);
```
jjs451: Looks good, my only comment is wondering why you need to update the cooldown in Abilities and then again with SpellUsable.  I'm not as familiar with both files but it seems like if you're updating the CD with SpellUsable that should be enough

Gebuz: The reduction I do in SpellsUsable only affects the charge currently recharging. This is needed to ensure that when a rune starts recharging during it starts with the correct cooldown. I could do this by listening to on_updatespellusable and set it there instead but since I use ExpectedCooldown from abilities I thought this would be cleaner. It wouldn't break anything since all cooldowns would be wrong by the same percent, but might be confusing in the future.

jjs451: Gotcha.  It won't be a big deal to update rune cooldowns in the future just gotta remember to update it in two places.  I'm gonna ping for a more experienced reviewer to check this PR just to make everything is good to go

src/Parser/DeathKnight/Shared/RuneTracker.js
```diff
@@ -60,6 +64,44 @@ class RuneTracker extends Analyzer {
...
+  }
+
+  changeCooldown(spellId, multiplier){
+    if(!this.spellUsable.isOnCooldown(spellId)){
+      return;
+    }
+    const remainingCooldown = this.spellUsable.cooldownRemaining(spellId);
+    const newCooldown = remainingCooldown * multiplier;
+    const reduction = remainingCooldown - newCooldown;
+    this.spellUsable.reduceCooldown(spellId, reduction);
```
kfinch: Won't this throw an error if one of the runes isn't on cooldown?

Gebuz: Should be cought in line 97

kfinch: Woooow my eyes slid right over that

src/Parser/DeathKnight/Blood/Modules/Abilities.js
```diff
@@ -265,21 +265,30 @@ class Abilities extends CoreAbilities {
...
+        },
         charges: 2,
       },
 
       {
         spell: SPELLS.RUNE_2,
         category: Abilities.SPELL_CATEGORIES.HIDDEN,
-        cooldown: haste => 10 / (1 + haste),
+        cooldown: (haste, combatantCurrent) => {
+          const multiplier = combatantCurrent.hasBuff(SPELLS.RUNE_MASTER.id) ? 0.4 : 0;
```
kfinch: Should the runic corruption bit be handled here too?

Gebuz: Runic Corruption is Unholy only and Rune Master is blood only

kfinch: Oh ok, missed that too.

I was debugging all day at my job, I don't think I can look at code any more today ☹️

PR #1311, opened 2018-01-30
[Hunter] Mongoose Fury and small things

Pewtro:
I know it looks big, but it's not. 😓 

I think the change in DeathTracker sounds better, but if it doesn't I'll revert it. 
The insignia fixes an issue in formatting not being consistent in that one row as opposed to the other rows.
Small update in the focus usage modules to also include how many casts of a given spell in the tooltip. 
Small typos in a few files. 

The only ones that need review (imo obviously) is DeathTracker (on wording) and the two Mongoose Fury modules. 

Report: https://www.warcraftlogs.com/reports/GNTgBZLD8rnhvcpk/#fight=5&source=28

image

image


src/Parser/Hunter/Survival/Modules/Features/MongooseFury/SixBiteWindows.js
```diff
@@ -0,0 +1,70 @@
...
+    if (spellId !== SPELLS.MONGOOSE_FURY.id) {
+      return;
+    }
+    this._currentStacks = event.stack;
+    if (this._currentStacks === MAX_STACKS) {
+      this.sixBiteWindows++;
+    }
+  }
+
+  on_byPlater_removebuff(event) {
```
kfinch: P L A T E R

src/Parser/Hunter/Survival/Modules/Features/MongooseFury/SixStackBites.js
```diff
@@ -0,0 +1,70 @@
...
+    if (spellId !== SPELLS.MONGOOSE_FURY.id) {
+      return;
+    }
+    this._currentStacks = event.stack;
+    if (this._currentStacks === MAX_STACKS) {
+      this.sixBiteWindows++;
+    }
+  }
+
+  on_byPlater_removebuff(event) {
```
kfinch: P L A T E R

src/Parser/Hunter/Survival/Modules/Features/MongooseFury/SixBiteWindows.js
Pewtro: monkaS

kfinch:
Also why do you need to track both 'max stack windows' and 'max stack bites'? Seems like two different ways of tracking the same thing.

MartijnHols:
Screenshots + report links please.

Pewtro:
Added screenshots and a report link. 

You want to achieve as many max stack windows over the course of the fight, and then you want to hit as many max stacked mongoose bites into those windows. Once you hit 6 stacks, you can continue hitting mongoose bite for the duration of the buff which lasts 14 seconds from the first Mongoose Bite cast. This is the overall goal for the spec (to hit as many 6 stack windows with as many 6 stacked bites in each).

Showcasing average stacks would be something really low like this, and isn't really clearly telling the player how well they played the encounter, because the number is so much lower than anything they strive for. 
image

Pewtro:
Fixed for PR, removed some unnecessary code in one of the modules. 
I would have preferred having it all in one module since I know it is a lot of repeating code, but I wanted them in two different statistic boxes.

PR #1313, opened 2018-01-31
[Frost Death Knight] Frost t21

jjs451:
Added bonuses from T21.  I'd especially like a check of the 2pc file to make sure I did everything right.  If GitHub works like I hope it will, commit 1061f09 won't show in this PR as soon as #1312 is merged.  Otherwise I can go back and try to get the cold heart stuff out of the branch.    
image


src/Parser/DeathKnight/Frost/Modules/Items/Tier21_2p.js
```diff
@@ -0,0 +1,63 @@
...
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.OBLITERATE_MAIN_HAND_DAMAGE.id || spellId === SPELLS.OBLITERATE_OFF_HAND_DAMAGE){
+      this.obliterateDamage += getDamageBonus(event, DAMAGE_BONUS);
+      return;
+    }
+    if(spellId === SPELLS.HOWLING_BLAST.id){
+      this.howlingblastDamage += getDamageBonus(event, DAMAGE_BONUS);
+      return;
+    }
+    if(spellId === SPELLS.FROSTSCYTHE_TALENT.id){
```
kfinch: It's fairly common for talent spells to have the talent ID be the cast ID but *not* the damage ID, in which case this would be the wrong ID. I'm going to assume that you've checked and they are the same, but if you haven't checked I recommend doing so.

PR #1314, opened 2018-01-31
[Survival] Way of the Mok'Nathal talent

Pewtro:
Report: https://www.warcraftlogs.com/reports/GNTgBZLD8rnhvcpk/#fight=5&source=28

image

I plan on expanding this at a later point to include damage contribution, but I need to figure out the attack power scaling formulas a bit more and which abilities are affected and how much - but for now this provides good information to the user.

### DIFF
--- a/src/Parser/Hunter/Survival/Modules/Talents/WayOfTheMokNathal.js
+++ b/src/Parser/Hunter/Survival/Modules/Talents/WayOfTheMokNathal.js
@@ -1,0 +1,108 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import Analyzer from 'Parser/Core/Analyzer';
+import StatisticBox from 'Main/StatisticBox';
+import SpellIcon from 'common/SpellIcon';
+import STATISTIC_ORDER from 'Main/STATISTIC_ORDER';
+import Combatants from 'Parser/Core/Modules/Combatants';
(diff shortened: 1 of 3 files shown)


src/Parser/Hunter/Survival/Modules/Talents/WayOfTheMokNathal.js
```diff
@@ -0,0 +1,108 @@
...
+
+class WayOfTheMokNathal extends Analyzer {
+
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  _currentStacks = 0;
+  fourStackUptime = 0;
+  fourStackStart = 0;
```
emallson: should this be a private variable? don't see a use for it outside of this module.

Pewtro: Have we ever done that in any modules? 😮

emallson: "private" meaning "stick a _ in front so other people know it may break underneath them"

Pewtro: Done 👌

emallson:
Didn't necessarily mean all these props -> private but it can always be changed later

PR #1315, opened 2018-02-01
[BrM Monk] Stagger Plot

emallson:
A plot of damage in stagger pool (and HP) over time, with annotations showing deaths and purifies. Purifies from things other than PB aren't shown because they just add a ton of noise (they happen frequently and semi-passively).

Here's what it looks like on **Heroic Imonar**:

Heroic Imonar

**Mythic Argus**:
Mythic Argus

There is also a minor bug-fix for stagger tracking in here that only really affects Argus -- i wasn't clearing the stagger pool properly on death and with multiple deaths on Argus it starts to add up. The effect on other stat modules is small; it was only noticeable in the plot because stagger would jump up high right after a combat rez.


src/Parser/Monk/Brewmaster/Modules/Features/StaggerPoolGraph.js
```diff
@@ -0,0 +1,276 @@
...
+
+      if(hpBySeconds[label] === undefined) {
+        hpBySeconds[label] = lastHpContents;
+      } else {
+        lastHpContents = hpBySeconds[label];
+      }
+
+      if(!!deaths.find(event => event.seconds === Number(label))) {
+        lastPoolContents = 0;
+        lastHpContents = { hitPoints: 0, maxHitPoints: lastHpContents.maxHitPoints };
```
sMteX: When a person dies, doesn't it make sense to set his maximum HP to 0 as well? Considering you're doing this as a default action anyway (when there's no information about HP *and there was none before to copy*).

MartijnHols:
Would it be possible to make the deaths vertical lines? Would be consistent with other graphs.

emallson:
screenshot from 2018-02-02 14-41-13

Deaths look like this now. I tried different ways of showing max HP while dead, and ultimately I like leaving it alone. It is clear enough both that you died (HP + death bar) and when (if) you were rezzed (HP > 0). For reference, here is that same fight with no max HP plotted while dead and with it set to 0 while dead.

screenshot from 2018-02-02 14-51-55
screenshot from 2018-02-02 14-54-03

Both of these options look passable on fights with lengthy deaths (aka Argus) but I don't think it makes sense to make design decisions based on how it looks on that one fight, whereas tank death -> immediate brez is much more common.

Gebuz:
Would you want purifiy to be vertical lines like the deaths? Just a thing to consider. Otherwise it looks good!

emallson:
M Argus
This is what it looks like with vertical lines (old SS so some other things are different). IMO in periods with frequent purifies, the vertical lines are distracting (which is why I've gone with the points for them).

That label change seems fine. I'll do that this afternoon and then hopefully we can merge this this evening.

sMteX:
If I still may have something to say - I do agree that showing maxHp even when dead looks more "clean" (the first screen after you've been dealing with it). I'd also stay with dots instead of lines on Purify, makes the chart clearer (but the dots are a little hard to see initially, don't know what to do with that). Either way, looks fine to me

Gebuz:
Yeah you're right it looks pretty cluttered with the lines, I agree that dots looks better.

PR #1318, opened 2018-02-01
[Survival] Caltrops cooldown update

Pewtro:
i am bad at numbers apparently. reverting my fix that wasn't a fix on this.

### DIFF
--- a/src/Parser/Hunter/Survival/Modules/Abilities.js
+++ b/src/Parser/Hunter/Survival/Modules/Abilities.js
@@ -55,7 +55,7 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.EXPLOSIVE_TRAP_CAST,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
         isOnGCD: true,
-        cooldown: 24,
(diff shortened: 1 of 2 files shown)


src/Parser/Hunter/Survival/Modules/Abilities.js
```diff
@@ -87,7 +87,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.CALTROPS_TALENT,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: 15,
+        cooldown: 12,
```
sMteX: I may be wrong but how about this? http://www.wowhead.com/spell=194277/caltrops 
Shows cooldown of 15 seconds? 🤔

Pewtro: That is what prompted my fix - but then I was in-game testing something and noticed
image

sMteX: Any leggos or other talents that can affect this?

Pewtro: I found it 🤔 updating this and some others post raid, so just leave it up for now 😨

Pewtro:
Already there we go 😄

PR #1325, opened 2018-02-02
[Core] Update sephuz

Pewtro:
See here: 

image

26.69% - 24.21% = 2.48%. This means I gain 0.48% haste additional from Sephuz before the 2% are added. 

It can also be shown through this formula 24,21*0,02 = 0,4842. 

Additionally this effect also occurs on the proc which I've also accounted for now showcased here:

image
55,83% - 24,21% = 31,62%, which means I gained more than the advertised 25%.

However, 24,21`*`1,25 +25 = 55,2625 - so we're still a few decimals off. 
However if you do 24,21`*`1,02`*`1,25 + 25 = 55,86775 - which is closer than anything else.

Showcasing formula using my stats:
``(1 + startHaste * PASSIVE_HASTE + PASSIVE_HASTE) + (uptimePercent * ACTIVE_HASTE) - 1``

(1 + 0,2421 * 0,02+0,02 ) + (0 * 0,25)-1 = 0,02842 * 100 = 2,4842 (same as above)
Assuming 10% uptime (We'd expect `` 0,9*2+0,1*25 = 4,3`` using the old calculation):

New calculation: (1 + 0,2421 * 0,02+0,02 ) + (0,1* 0,25) - 1 = 0,049842 * 100 = 4,9842

The old calculation would have been almost 0.7% average haste off target, whereas my above calculation shows us merely being 0.03% off 

Log here:
Report: 
https://www.warcraftlogs.com/reports/dJkh3NZX76zwAM4r#fight=11&type=damage-done&source=4

Live: 
image

New version:
image

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-02-02'),
+    changes: <Wrapper>Updated <ItemLink id={ITEMS.SEPHUZS_SECRET.id} icon /> to also showcase the multiplicative effect that is added as part of the baseline and the proc on top of the flat bonuses</Wrapper>,
+    contributors: [Putro],
+  },
+  {
     date: new Date('2018-01-28'),
     changes: <Wrapper>Added <i>How It's Made</i>, <i>How To Change</i>, <i>Spec Maintainer</i> and <i>State Of The Spec</i> boxes to the results page.</Wrapper>,
     contributors: [Zerotorescue],
--- a/src/Parser/Core/Modules/Items/Legion/Legendaries/SephuzsSecret.js
+++ b/src/Parser/Core/Modules/Items/Legion/Legendaries/SephuzsSecret.js
@@ -28,13 +28,13 @@ class SephuzsSecret extends Analyzer {
   item() {
     const uptimePercent = this.combatants.selected.getBuffUptime(SPELLS.SEPHUZS_SECRET_BUFF.id) / this.owner.fightDuration;
     const startHaste = this.statTracker.startingHasteRating / 37500;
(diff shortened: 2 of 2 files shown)


Gebuz:
Most haste percentage increases are multiplicative and not additive, I don't think changing it to show additive is correct. It would also be wrong as soon as you have any haste buffs, rating or percent.

Pewtro:
Oh yeah, could just change it to be currentHasteRating instead of starting and that would fix. 
I feel as if it needs an update of some sorts and showcase that it probably added closer to 3% haste for the majority of users and not 2% 🤷‍♂️

Gebuz:
It's important to differentiate between percent (multiplicative) and percent point (additive). The passive adds exactly 2 percent (multiplicative). I understand why you want to display the percent point increase, but  I don't think that is correct since it makes it harder to relate the gain to a percent DPS increase and a lot of people don't realize the difference between percent and percent point increases. If you want a non relative value I would much prefer that you calculate it's average value in haste rating and put that in the tooltip.

kfinch:
I took a look at this thread, and for what it's worth I'm in agreement with Gebuz on this one.

Pewtro:
I'm well aware of the need to differentiate between multiplicative and additive numbering... 

I took this one because I've received questions about Sephuz and the values we show because we're not clear that it's multiplicative values we're showing for haste (and even then the effect calculation is wrong regardless, because it's not 25% multiplicatively as shown above). Furthermore I doubt that that it's well known by your average WoW player how blizzard applies various % modifiers (and who can blame them, they aren't exactly consistent on this manner), so that's not where their mind would wonder upon reading our shown information.

I fundamentally disagree, I think showing the additive number is infinitely better for the end user because it's way more relatable to your character and his/her stats. 

I'll be closing this and someone else can take up the mantle and rework it into average haste rating or something else entirely.

PR #1326, opened 2018-02-02
[Hunter] Call of the Wild

Pewtro:
This updates the shared legendary Call of the Wild to include information on the various Aspects it reduced the cooldown of and how many extra casts was gained through this.

Report: https://www.warcraftlogs.com/reports/92nNqC1gQaRhXmkx/#fight=13&source=10

image

Report: https://www.warcraftlogs.com/reports/YFcH8P27mtGnzLMA#fight=8&type=damage-done&source=5
image

### DIFF
--- a/src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
+++ b/src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
@@ -20,15 +20,16 @@ const EMBRACE_CDR = 0.2;
 const PATHFINDER_CDR = 1 / 3;
 const COTW_CDR = 0.35;
 
-let BASE_EAGLE_COOLDOWN = 120000;
-let BASE_CHEETAH_COOLDOWN = 180000;
-let BASE_TURTLE_COOLDOWN = 180000;
-const BASE_WILD_COOLDOWN = 120000;
-
 class CallOfTheWild extends Analyzer {
   static dependencies = {
     combatants: Combatants,
   };
+
+  //the base values that we can change as needed
+  baseEagleCooldown = 120000;
+  baseCheetahCooldown = 180000;
+  baseTurtleCooldown = 180000;
+  baseWildCooldown = 120000;
 
   //used later when I set their CDs with the bracers equipped
   eagleCooldownWithBracers = 0;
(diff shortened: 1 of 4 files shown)


src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
```diff
@@ -1,24 +1,124 @@
...
+    }
+    if (spellId === SPELLS.ASPECT_OF_THE_WILD.id) {
+      this.wildCDR += BASE_WILD_COOLDOWN * COTW_CDR;
+      this.wildCastTime = event.timestamp;
+    }
+  }
+
+  on_finished() {
+    //checks if any given ability was cast at a point in the fight where the CDR granted by the bracers doesn't allow for another cast, in which case the CDR is ineffective
+    if (this.eagleCastTime > (this.owner.fight.end_time - this.eagleCooldownWithBracers && this.eagleCDR > 0)) {
```
Gebuz: The parentheses are wrong here. I'm a bit confused why you do the time check anyway. If I cast my last aspect 10 min. before the encounter ends, surely that shouldn't get any credit for effective CDR either?

Pewtro: That's a case where cast efficiency takes over imo. It did still provide the CDR effectively gaining you (partly) additional possible casts

Gebuz: Ah okay that makes sense. I still think using the time between casts to calculate the effective CDR would be better but if you prefer this approach I'd be okay with that too.

src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
```diff
@@ -1,24 +1,124 @@
...
+  SPELLS.ASPECT_OF_THE_WILD.id,
+];
+
+const EMBRACE_CDR = 0.2;
+const PATHFINDER_CDR = 1 / 3;
+const COTW_CDR = 0.35;
+
+let BASE_EAGLE_COOLDOWN = 120000;
+let BASE_CHEETAH_COOLDOWN = 180000;
+let BASE_TURTLE_COOLDOWN = 180000;
```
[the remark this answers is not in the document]
Pewtro: That seems like a good way to just add unnecessary code and variables into the fray? 
I'd be adding 4 additional variables and additional spec checks throughout the code for SV vs BM vs MM, are you sure?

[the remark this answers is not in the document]
Gebuz: I agree it's a lot of variables, but afaik it's bad style to have non constants outside classes. You could also just move them inside the class. There might also be cleaner way to do it with getters.

Pewtro:
Did a rewrite of how things are handled, the on_finished is gone and it's overall cleaner (and better) now imo.

PR #1333, opened 2018-02-03
Make Resto Great Again

kfinch:
Implemented Checklist for Resto Druid, enabling me to return the spec to 'great' status.

Also removed most of the custom Resto Druid ABC code, as it was just duplicating core functionality. We now fallback on the core version.

Also tweaked WG/rejuv ratio, as it was a little strict for a "guessing" suggestion.

Here's me healing poorly:
LOG: wowanalyzer.com/report/4hqAvJXRyg1kFPd2/3-Heroic+Portal+Keeper+Hasabel+-+Wipe+3+(7:27)/Aslann

SCREENSHOTS:
resto-checklist-1
resto-checklist-2
resto-checklist-3


src/Parser/Druid/Restoration/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,166 @@
...
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.ESSENCE_OF_GHANIR,
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.TRANQUILITY_CAST,
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.INNERVATE,
+          }),
```
buimichael: Should also include ToL if specced.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,166 @@
...
+          new Requirement({
+            name: <Wrapper>Use your <SpellLink id={SPELLS.CLEARCASTING_BUFF.id} icon /> procs</Wrapper>,
+            check: () => this.clearcasting.clearcastingUtilSuggestionThresholds,
+            tooltip: `This is the percentage of Clearcasting procs that you used. Regrowth is normally very expensive, but it's completely free with a Clearcasting proc. Use your procs in a timely fashion for a lot of free healing.`,
+          }),
+          new Requirement({
+            name: <Wrapper>Don't overuse <SpellLink id={SPELLS.REGROWTH.id} icon /></Wrapper>,
+            check: () => this.clearcasting.nonCCRegrowthsSuggestionThresholds,
+            tooltip: `This is the number of no-clearcasting Regrowths you cast per minute. Regrowth is very mana inefficient, and should only be used in emergency situations (and when you've already expended Swiftmend). Usually, you should rely on other healers in your raid to triage.`,
+          }),
```
buimichael: Could also include "Use all of your mana effectively" i.e. making sure the player has used up most of their mana at the end of the encounter.

buimichael: I would probably move the "Used a second potion" here as it's directly tied with mana management.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,166 @@
...
+        ];
+      },
+    }),
+    new Rule({
+      name: 'Use your defensive / emergency spells',
+      description: <Wrapper>Restoration Druids unfortunately do not have many tools to deal with burst damage, but you should take care to use the ones you have. The below percentages represent the percentage of time you kept each spell on cooldown.</Wrapper>,
+      requirements: () => {
+        return [
+          new GenericCastEfficiencyRequirement({
```
buimichael: I think this could be excluded. But if you're using t20 then it should appear under "User your healing CDs"

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,166 @@
...
+            spell: SPELLS.IRONBARK,
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.BARKSKIN,
+          }),
+        ];
+      },
+    }),
+    new PreparationRule(),
+  ];
```
buimichael: Consider having  an "Avoid overhealing" section as well?

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,166 @@
...
+}
```
buimichael: I like the "Pick the right tools for the fight" which the holy paladin checklist has. We should probably include that aswell, for ToL/Cult (sotf LUL) and legendaries. Also give "error" if they are playing prosperity without t20.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
kfinch: Yeah, got similar feedback from #resto. I'll probably exclude it entirely from this area, and add it to the cooldowns dropdown if player has SotF, SotA, or 4T20.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
kfinch: Good catch, I'll add that.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
kfinch: Oh yeah, forgot about the mana use rate one, I'll put that in. The second potion thing is tied to a generic rule, I don't think it's worth trying to mess with that.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
kfinch: I'm kinda skeptical about that, because it's so unavoidable for much of what Resto does.

src/Parser/Druid/Restoration/Modules/Features/Checklist.js
kfinch: Good ideas, I wasn't sure what to do about the talent things. Probably group talent choice metrics together like that is good idea.

kfinch:
@buimichael Ok, pushed some updates. I kept Swiftmend in for now, but lowered the thresholds and also put a disclaimer in the description saying it should only be used in an emergency. I added Velens and ToL to the "use your cooldowns" rule. I added mana usage check to the "manage mana" rule. I added the existing talent throughput checks we have to a new "Pick the right tools for the fight" rule.

I'd like to add some more talent usage checkers, but I think that's better saved for a later PR.

PR #1334, opened 2018-02-03
[Death Knight] Added Rune chart and usage and updated statbox

### LINKED ISSUE #1255: [Death Knight] Runetracker
I'm working on implementing a Rune tracker. The way I do it is make 3 fake spells for the runes with 2 charges each to simulate how they recharge in game (I suspect that this is also how it works behind the scenes in game since it used to work like that). I have got the basics working but I have a few considerations concerning the display I would like to discuss.

How should it be displayed on the timeline (if at all)? I had an idea of making a bar that would colour code the number of runes on cooldown (>=3 green, = 4 orange, = 5 red, = 6 darker red for example), but I'm not sure if I should add something like this to the timeline (I don't know how complicated it would be to extend it for one class).

Another way to do it is just displaying the 3 rune pairs cooldown, but that might be confusing since that is not how it is displayed in game. If we choose to do this it would probably make sense to also show them in the cast efficency tab.

Since I plan to make this work for all 3 specs, I'll need to implement Runic Corruption. Are buffs that dynamically reduce cooldown (not through haste) supported? EDIT: looks like they aint, I would probably need to listen to on_toPlayer_buffapplied/removed similar to on_changehaste.

Gebuz:
image
image
image

Example logs:
Blood: https://www.warcraftlogs.com/reports/tRdqNKw4zA1hmWx6#fight=22&type=damage-done&source=17
Unholy: https://www.warcraftlogs.com/reports/HqgQV8TjMCARBnwD/#fight=1&source=7&type=damage-done
Frost: https://www.warcraftlogs.com/reports/8bmazGHTqFYJ4jrB#fight=52&type=damage-done&source=6

RuneBreakdown.js only has 4 new lines, rest is copied from ResourceBreakdown.js, I just needed to overwrite a single method but it is ~100 lines long.

RuneTracker.js now extends ResourceTracker to show runes spent/generated tables.  `addPassiveRuneRegeneration` and `addPassiveAccelerator` are responsible for adding passive rune regeneration gains to the table. They currently only called in on_finished() because of how their contribution is split between passive regeneration and Accelerators. In the future I might try to add them whenever a rune finishes recharging based on buff uptimes while it was recharging instead.

fixes #1255


Gebuz:
Added Frost & Unholy specific Thresholds and more comments.

src/Parser/DeathKnight/Shared/RuneBreakdown.js
```diff
@@ -0,0 +1,153 @@
...
+          <thead>
+            <tr>
+              <th>Ability</th>
+              <th colSpan="2"><dfn data-tip="Runes generated from passive regeneration and abilities that accelerate it are estimates.">{resourceName} generated</dfn></th>
+              <th colSpan="2"><dfn data-tip="This is the amount of resources that were generated while you were already at cap.">{resourceName} wasted</dfn></th>
+            </tr>
+          </thead>
+          <tbody>
+            {generated && generated
+              .filter(ability => ability.abilityId === SPELLS.RUNE_1.id)
```
jjs451: I might be misreading but wouldn't this just show the data for rune 1?

Gebuz: I just use that as the ID because it needs an ID to filter by, it has no functional use other than that. I merge all the runes into one in runetracker

src/Parser/DeathKnight/Shared/RuneBreakdown.js
```diff
@@ -0,0 +1,153 @@
...
+              .map(ability => (
+                <tr>
+                  <td style={{ width: '30%' }}>
+                    <SpellIcon id={ability.abilityId} noLink={true} />{' '}
+                    {'Passive Rune regeneration'}
```
MartijnHols: This can just be a string

src/Parser/DeathKnight/Shared/RuneBreakdown.js
```diff
@@ -0,0 +1,153 @@
...
+                    <div
+                      className="performance-bar"
+                      style={{ width: `${(ability.wasted / totalWasted) * 100}%` }}
+                    />
+                  </td>
+                </tr>
+              ))}
+          </tbody>
+        </table>
+        {showSpenders &&
```
MartijnHols: Please use parens around multiline blocks of JSX.

src/Parser/DeathKnight/Shared/RuneTracker.js
```diff
@@ -64,36 +87,93 @@ class RuneTracker extends Analyzer {
...
       });
     }
   }
+  on_updatespellusable(event){ //track when a rune comes off cooldown
+    if(!RUNE_IDS.includes(event.spellId)){
+      return;
+    }
+    let change = 0;
+    if(event.trigger === 'endcooldown' || event.trigger === 'restorecharge'){ //gained a rune
+      change++;
```
MartijnHols: It's preferred to use += 1 and -= 1 instead of ++ or --.

PR #1335, opened 2018-02-03
[Havoc] Nemesis Update

CollCrom:
Old
image
New
image
This fixes a bug where the uptime was not accounting for if the enemy died with the debuff on them.  Also adds in how much dps you gained from the talent.

Test log: https://www.warcraftlogs.com/reports/T68JHYX7faDPW3jv#fight=8&type=damage-done

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
@@ -48,7 +48,6 @@ class Nemesis extends Analyzer {
       return;
     }
     const enemy = this.enemies.getEntity(event);
-    console.log(enemy)
     if(enemy.hasBuff(SPELLS.NEMESIS_TALENT.id) || this.hasNemesisBuff) {
       this.bonusDmg += calculateEffectiveDamage(event, NEMESIS_DAMAGE_MODIFIER);
     }
--- a/src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
@@ -48,15 +48,15 @@ class Nemesis extends Analyzer {
       return;
     }
     const enemy = this.enemies.getEntity(event);
+    console.log(enemy)
     if(enemy.hasBuff(SPELLS.NEMESIS_TALENT.id) || this.hasNemesisBuff) {
       this.bonusDmg += calculateEffectiveDamage(event, NEMESIS_DAMAGE_MODIFIER);
     }
   }
 
   get nemesisUptimePercent() {
-    let playerUptime = 0;
     const enemyUptime = this.enemies.getBuffUptime(SPELLS.NEMESIS_TALENT.id);
-    NEMESIS_BUFF_IDS.forEach( spellId => playerUptime += this.combatants.selected.getBuffUptime(spellId));
+    const playerUptime = NEMESIS_BUFF_IDS.reduce((uptime, spellId) => uptime + this.combatants.selected.getBuffUptime(spellId), 0);
     return (enemyUptime + playerUptime) / this.owner.fightDuration;
   }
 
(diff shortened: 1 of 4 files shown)


src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
```diff
@@ -7,29 +7,66 @@ import Enemies from 'Parser/Core/Modules/Enemies';
...
-    const nemesisUptime = this.enemies.getBuffUptime(SPELLS.NEMESIS_TALENT.id);
-    const nemesisUptimePercentage = nemesisUptime / this.owner.fightDuration;
+  get hasNemesisBuff() {
+    const buffs = this.combatants.selected.activeBuffs();
+    return buffs.some(buff => NEMESIS_BUFF_IDS.includes(buff.ability.guid));
+  }
 
+  on_byPlayer_damage(event) {
+    if(event.targetIsFriendly) {
+      return;
```
kfinch: What's with the friendly check? What's the situation where this is needed?

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
```diff
@@ -7,29 +7,66 @@ import Enemies from 'Parser/Core/Modules/Enemies';
...
+    const enemy = this.enemies.getEntity(event);
+    if(enemy.hasBuff(SPELLS.NEMESIS_TALENT.id) || this.hasNemesisBuff) {
+      this.bonusDmg += calculateEffectiveDamage(event, NEMESIS_DAMAGE_MODIFIER);
+    }
+  }
+
+  get nemesisUptimePercent() {
+    let playerUptime = 0;
+    const enemyUptime = this.enemies.getBuffUptime(SPELLS.NEMESIS_TALENT.id);
+    NEMESIS_BUFF_IDS.forEach( spellId => playerUptime += this.combatants.selected.getBuffUptime(spellId));
```
kfinch: technically this should be done with a reduce instead, but I think reduces often look ugly so this is fine :stuck_out_tongue:

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
```diff
@@ -7,29 +7,66 @@ import Enemies from 'Parser/Core/Modules/Enemies';
...
+  }
+
+  statistic() {
     return (
       <StatisticBox
         icon={<SpellIcon id={SPELLS.NEMESIS_TALENT.id} />}
-        value={`${formatPercentage(nemesisUptimePercentage)}%`}
-        label="Nemesis Uptime"
-        tooltip={`The Nemesis buff total uptime was ${formatDuration(nemesisUptime / 1000)}.`}
+        value={`${formatNumber(this.bonusDmg / this.owner.fightDuration * 1000)} DPS`}
```
kfinch: For statistic boxes, it's standard to put percent of total damage as the front facing value, and relegating DPS to the tooltip.

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
```diff
@@ -7,29 +7,66 @@ import Enemies from 'Parser/Core/Modules/Enemies';
...
+    return buffs.some(buff => NEMESIS_BUFF_IDS.includes(buff.ability.guid));
+  }
 
+  on_byPlayer_damage(event) {
+    if(event.targetIsFriendly) {
+      return;
+    }
+    const enemy = this.enemies.getEntity(event);
+    if(enemy.hasBuff(SPELLS.NEMESIS_TALENT.id) || this.hasNemesisBuff) {
+      this.bonusDmg += calculateEffectiveDamage(event, NEMESIS_DAMAGE_MODIFIER);
```
kfinch: It looks to me as though you aren't actually checking for matched creature type here... isn't that a fairly large oversight?

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
MartijnHols: friendly fire

MartijnHols: image

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
MartijnHols: should definitely be a reducer

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
kfinch: Oh so like "Aura of Sacrifice can cause any damage to FF, so it needs to be checked for everywhere"?

MartijnHols: if you're not checking for a specific spell id doing the damage event then yes

src/Parser/DemonHunter/Havoc/Modules/Spells/Nemesis.js
CollCrom: image
I dont think you can get the enemy type from the log =/

Pewtro: image

CollCrom: nah
image

kfinch:
Sorry I haven't had the time to get back to this until now.

It seems you can't check creature type, but presenting the damage value as truth when you know it could be wrong strikes me as disingenuous. If any part of the damage happens while player has the buff (instead of the debuff on mob), you should leave a note in the tooltip that you can't actually know for sure if you were getting the buff there.

PR #1337, opened 2018-02-04
[Core] Fix Death Tracker

### LINKED ISSUE #443: Add a "you died and were dead for x seconds" suggestion

### LINKED ISSUE #1207: [core] DeathTracker not seeing a resurrection
See the following log: https://wowanalyzer.com/report/q6cnmrjRYwxZV93H/8-Mythic+Felhounds+of+Sargeras+-+Kill+(4:58)/Dustbunny

Player died, was resurrected, but the DeathTracker did not catch it. There didn't appear to be a resurrect event, and in fact from the log it's not clear to me how the player was resurrected... events around the moment of death are a little weird.

Sharrq:
Death Tracker was not counting Soulstone as a ressurect and there are no events to indicate that the player used their soulstone. So instead, I made it so that if you are healed, take damage, start casting something, or cast something while you are marked as dead, then it will count as a ressurect.

Fixes #1207 
Fixes #443

### DIFF
--- a/src/Parser/Core/Modules/DeathTracker.js
+++ b/src/Parser/Core/Modules/DeathTracker.js
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import SPELLS from 'common/SPELLS';
 import { formatMilliseconds, formatNumber, formatPercentage } from 'common/format';
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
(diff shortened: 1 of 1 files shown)


Gebuz:
I don't know if it's worth changing just for Argus, but while in ghost form some classes can cast abilities 
 such as Ghost Wolf, Glide and Wraith Walk. That could throw this off.

PR #1338, opened 2018-02-05
[Core] Updated Norgannons

CollCrom:
Instead of showing the percent uptime on the intellect buff it now shows the average amount of int gained.
new
image

old
cZhfdNGR4LQBwj32
image
test log
https://www.warcraftlogs.com/reports/cZhfdNGR4LQBwj32#fight=4&type=damage-done&source=2

### DIFF
--- a/src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/NorgannonsProwess.js
+++ b/src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/NorgannonsProwess.js
@@ -3,7 +3,8 @@ import React from 'react';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
-import { formatPercentage } from 'common/format';
+import { formatPercentage, formatNumber } from 'common/format';
+import { calculatePrimaryStat } from 'common/stats';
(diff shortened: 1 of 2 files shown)


src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/NorgannonsProwess.js
```diff
@@ -32,11 +33,15 @@ class NorgannonsProwess extends Analyzer {
...
 
   intProc = 0;
+  intBuff = 0;
   pantheonProc = 0;
   damage = 0;
 
   on_initialized() {
     this.active = this.combatants.selected.hasTrinket(ITEMS.NORGANNONS_PROWESS.id);
+    if(this.active) {
+      this.intBuff = calculatePrimaryStat(940, 3257, this.combatants.selected.getItem(ITEMS.NORGANNONS_PROWESS.id).itemLevel);
```
Gebuz: The int value should be 11483 at 940 ilvl. The value in the proc on wowhead is usually way too low, but the tooltip value on the trinket should be correct: http://www.wowhead.com/item=154177/norgannons-prowess

CollCrom:
Yeah the wowhead link I used was wrong. Will update tomorrow

PR #1339, opened 2018-02-05
[Core] Added Patch Compat Version

### LINKED ISSUE #884: Add a prop to modules allowing them to specify patch compatibility
Add a static prop to modules to allow them to specify patch compatibility. Nothing fancy, just date + message, if logs is before specified date the custom message shows that could specify what is wrong.

We considered giving modules versions but this got messy quick.

This prop could also be an object { date, warning, disabled } with disabled indicating if the module should be disabled if it's assumed to be incompatible.

Sharrq:
Added a Patch Compatibility Option to the CONFIG to allow the developer to indicate what patch the analyzer is certified/verified for. This will help locate outdated analyzers and let the users know whether patch changes are included or not.

Also adjusted the warning display order so feedbackWarning is higher up than the needs work warning (since only the first one will actually get displayed)

Fixes #884


MartijnHols:
When making a pull request please try to always include a screenshot. If you added something visual show what is new. If you changed a calculation screenshot the old value versus new. If you added a check similar to that of legendaries/enchants/deaths please include a link to an example report in the PR.

Sharrq:
ill get some pictures when i get home.

Sharrq:
image

image

MartijnHols:
I'd like to hear more insights about this. Is it fair to ask of contributors to push a patch compatibility update each patch and if not display a prominent message, or should it be more subtle?

Gebuz:
I personally wouldn't mind aslong as it is only major patches. Most of the time all it would take is changing the version number and maybe a tier modifier or two. 

It might be better to have a warning message similar to when a spec isn't implemented, so instead of "ask the maintainer to update it" it could be "Something from last patch that is missing? Ask the maintainer or help us add it!"

jjs451:
I second Gebuz.  It might be helpful if a message showing if a spec is updated for the most recent tier/major patch.  Most specs don't get changed in minor patches but something checking the first 2 numbers in a patch would make it obvious to users if a module hasn't been updated in a while

kfinch:
I concur with Gebuz, only update on major patches, and the lighter touch "warning"

Sharrq:
Ill make those changes probably tonight. Per jjs's comment though, checking the first 2 numbers wont work because blizzard went to the patch model of making major patches that are both 7.x and 7.x.5. 

But yea, major versions only is fine.

MartijnHols:
Since we had to remove the analyzer completion from the results header because it led to confusion I'd like an alternative that also incorporates patch compatibility properly. How about we add this to the headers after the maintainer name:

* Accurate and complete for 7.3.5 (platinum + updated)
* Accurate for 7.3.5 (gold + updated)
* Accurate for 7.2 (gold + outdated)
* Incomplete (incomplete)
* Not actively maintained

Sharrq:
I tried something a little different. I brought back the completion but renamed it "Analysis Accuracy". I know that isnt perfectly representative of what it actually is, but im hoping its less confusing from the player side.

I also changed the dividers to consolidate them into a single divider (by getting rid of the now duplicated listing of the spec maintainers), and i changed the boxes into headings instead. I decided to use the 2nd divider because its after almost everything the player cares about, so less scrolling.

MartijnHols:
I like the info box divider changes apart from the border underneath the titles, I think it would look cleaner without that. What do you think?

I think showing this as "analysis accuracy" will lead to similar confusion as the reason it was removed entirely earlier; it's easy to confuse with some sort of performance/player accuracy rating. I don't think we can make it work in the current format, hence the completely different suggestions earlier. At *incomplete* the patch compatibility doesn't matter; it's wrong in undetermined areas anyway so I left it out completely as to not mislead players. At *gold* it's supposed to be accurate enough all around so patch compatibility might indicate it has been outdated by a patch... maybe calling it "accurate for <old patch>" wouldn't have been right. `Outdated (accurate for 7.2)` might be better. Platinum was meant to indicate a spec being complete.

Sharrq:
Im not a fan of the lines either, ill see if i can find out where they are coming from. Ran into the same issue in the Character Page PR, not sure where they are coming from, but ill see if i can find them

As for completeness, I guess i can see where the player might misunderstand that. Im just a little shaky on combining patch compatibility with completeness .... Ill give it a shot I just kinda think they should be two different things.... Ill play with it though, i have some ideas on something that might help.

Sharrq:
alright, i removed the lines in the divider, redid the completeness to say "accurate" for Good/Gold and "complete" for Great/Platinum.

Platinum with Patch version
image

Platinum without Patch version
image

Gold without Patch version
image

Not actively maintained
image

Needs Work
image

MartijnHols:
Looks good, I think that can work, but I'll have to think some more about it and do a more thorough review later.

Sharrq:
closing this since its being taken care of in another PR

Issue #824, opened 2017-11-18
[Core] Add Scrollbar to Fight and Player Selectors


fasib:
There is now way right now to select all the players/fights, if your window is to small.

![2017-11-18_09-31-02](https://user-images.githubusercontent.com/317216/32978483-515370c4-cc43-11e7-8ad6-dce8d702a5c3.gif)
^End of Browser

Sinaloit:
This issue is still present even after the Grouping by Role change.  That change actually exacerbates the issue as there are now additional lines used up by the roles.  Zooming out your view will allow you to see more/all people but is not an optimal solution.  Collapsible roles groupings (Click on role) and maybe fight grouping (By sections listed in achievements?)  would work if a scroll enabled selector isn't feasible/ideal.

MartijnHols:
This was fixed some time ago, thanks for the report!

PR #1340, opened 2018-02-05
[Core] Separate Player list into Columns

### LINKED ISSUE #828: Player selection redesign
The player selection as it is is a pretty minimal solution. I think it should be prettier.

- How about making player cards with role / icon more prominent and for example listing 3 per row (or 4 so we get exactly 5 rows for Mythic). 
- Grouping players by role? So maybe 2 per row instead so it works well with tanks (amount of people in the other roles are random anyway)
- How hard would it be to get the player realm name and show player avatars? 

Users may be used to the list view as it is so it may be hard to get something that's prettier **and** works just as well.

Sharrq:
Separate Player Selector List into columns so some of the players arent going off the screen.

Part of #824 
Fixes #828

### DIFF
--- a/src/Main/PlayerSelectionList.js
+++ b/src/Main/PlayerSelectionList.js
@@ -159,7 +159,7 @@ export class PlayerSelectionList extends React.PureComponent {
         return (
           <div key={roleID === null ? -1 : roleID} className="card">
             {this.renderRoleHeader(roleID, friendlies.length)}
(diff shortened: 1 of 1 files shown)


MartijnHols:
Could you share with us some screenshots please?

Sharrq:
image

Sharrq:
I can get better ones once i get home, but that is the gist of it.

PR #1344, opened 2018-02-05
Misc layout changes

MartijnHols:
I removed the boss-specific background since it made styling the results page harder and inconsistent.
Added icons to the main tabs. Changed highlighting of tabs, I thought the border looked a bit weird since it's not used as a highlighter anywhere else so this is much easier to recognize. Slightly lightened the background to make it more visible and increase contrast. Blackened the panel background since the previous blueish background color was kinda weird and random. Made the panel heading fonts the same as tabs for more consistency. 

Old:
image
New:
image


MartijnHols:
One thing that still bugs me is the checklist item font size being bigger than the tab header font size, I think that's wrong. Reducing the font size didn't seem trivial though since then the item bonuses list will never match height anymore. Since I also kinda decided to make a new character tab and move the stats on pull and talents there in the future, this would also free up the space to allow us to reduce the checklist font size.

Issue #1271, opened 2018-01-26
Can't view two players with the same name


ACmage0042:
When trying to view a report for a log that has two players with the same name, it will always pick one of the players in the list and never show you the other player. 

So for example - https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)
When i click on Swoof (Windwalker) it will show me the analysis for him. then when you go to click Swoof (Demonology) It loads the Windwalker again. 

The url when you click on a player only shows the name not the realm data as well, and i'm not sure if the log data carries that over.

kfinch:
We don't get realm data, but we do get a unique ID for telling them apart. This should be solvable.

MartijnHols:
We can solve this by prepending the player id in front of the player name like we do for fights. A PR would be welcome. Please do ensure the server's path regex also works.

CubeLuke:
I started looking into this as a nice issue to get started with. @MartijnHols , how concerned should we be about allowing pre-existing urls to work for this? Meaning, should we just allow [https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/Swoof](https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/Swoof) to work similarly to [https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4-Swoof](https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4-Swoof) as a default? Or should we just move forward with the format of `playerId-playerName` as the new format?

MartijnHols:
Backwards compatibility would be greatly preferred.

PR #1351, opened 2018-02-08
[Core]Updated urls to pass the player id instead of just the name

CubeLuke:
#1271 Updated urls to allow for player name, player id, or a combination of both with a hyphen to separate.

The following log has two players by the name of 'Swoof'. This update will allow for the following
[https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4-Swoof](https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4-Swoof)
[https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4](https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/4)
[https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/Swoof](https://wowanalyzer.com/report/YkpZHwDvT7mc61fq/7-Mythic+Kin'garoth+-+Kill+(7:44)/Swoof)


CubeLuke:
Wasn't sure if this required a change log addition since it's really nothing that users should be bothered to know/care about.

server/index.js
```diff
@@ -47,7 +47,7 @@ app.get('/', (req, res) => {
 app.get('/news/:article', (req, res) => {
   res.send(index);
 });
-app.get('/report/:reportCode([A-Za-z0-9]+)/:fightId([0-9]+)?:fightName(-[^/]+)?/:playerName([^/]{2,})?/:tab([A-Za-z0-9-]+)?', (req, res) => {
+app.get('/report/:reportCode([A-Za-z0-9]+)/:fightId([0-9]+)?:fightName(-[^/]+)?/:playerId([0-9]+)?:playerName(-[^/]{2,})?/:tab([A-Za-z0-9-]+)?', (req, res) => {
```
MartijnHols: Did you check the playername regex? I think this regex will incorrectly include the dash in the player name param.

src/selectors/url/report/getPlayerName.js
```diff
@@ -2,5 +2,8 @@ import getMatch from './getMatch';
 
 export default state => {
   const match = getMatch(state);
-  return match ? match.params.playerName : null;
+  if(match && match.params.player){
+    return match.params.player.indexOf('-') > -1 ? match.params.player.split('-')[1] : match.params.player;
```
MartijnHols: Split might break the result in more than 2 parts, which would break this. How about:

```jsx
const index = match.params.player.indexOf('-');
return index !== -1 ? match.params.player.substr(index) : match.params.player;
```

server/index.js
CubeLuke: I see my mistake here, updated the regex so the dash isn't matched with the player name.

src/selectors/url/report/getPlayerName.js
CubeLuke: That's a much better solution than what I had, got it updated!

PR #1354, opened 2018-02-08
[Hunter] Minor abilities update

Pewtro:
Removing the cooldown of Dire Beast, Dire Frenzy and Mongoose Bite as there is no way to track the cooldown resets on these spells. I kept the code in with a little note just for clarity to anyone poking around in the files 
Sets a number of abilities to ROTATIONAL_AOE as deemed relevant
Adds a minor check to Trickshot modules to ensure it doesn't show always show in the talent box
Removes Dire Beast and Dire Frenzy from checklist because they have their cooldown removed. 
Changes the name of the Volley buff to just be Volley so it doesn't show up like this
image


Pewtro:
Added fixes for the recent hotfix to t214p. 
Added additional calculation into Call of the Wild to account for t214p for BM and also a small message regarding Convergence of Fates for BM as a build using it is gaining ground again

src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
```diff
@@ -93,9 +100,22 @@ class CallOfTheWild extends Analyzer {
...
+      this.tier21WildCDR = 0;
+    }
+    if (spellId === SPELLS.KILL_COMMAND.id) {
+      const aspectIsOnCooldown = this.spellUsable.isOnCooldown(SPELLS.ASPECT_OF_THE_WILD.id);
+      if (aspectIsOnCooldown) {
+        const remainingMs = this.spellUsable.cooldownRemaining(SPELLS.ASPECT_OF_THE_WILD.id);
+        if (remainingMs > TIER_COOLDOWN_REDUCTION_MS) {
+          this.tier21WildCDR += TIER_COOLDOWN_REDUCTION_MS;
+        } else {
+          console.log("under 3s : ", remainingMs);
```
Gebuz: Please remove or hide behind a debug

src/Parser/Hunter/BeastMastery/Modules/Abilities.js
```diff
@@ -117,7 +119,7 @@ class Abilities extends CoreAbilities {
       },
       {
         spell: SPELLS.BARRAGE_TALENT,
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
```
Gebuz: Since it has a cast efficiency, I'm assuming it is used in single target? If so I think it's better to keep it as rotational. Otherwise I think the cast efficiency should be removed (afaik we don't show cast efficiency for AOE only spells, but I could be wrong).

src/Parser/Hunter/Marksmanship/Modules/Abilities.js
```diff
@@ -88,7 +88,7 @@ class Abilities extends CoreAbilities {
       },
       {
         spell: SPELLS.BARRAGE_TALENT,
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
```
Gebuz: Same as other Barrage comment.

src/Parser/Hunter/Shared/Modules/Items/CallOfTheWild.js
Pewtro: oh I missed that one when removing them

Issue #1218, opened 2018-01-19
[General] Pre Potions are not always correctly accounted for


CollCrom:
In this log kidsmoke used a pre pot as shown in his buffs but the wow analyzer thinks that he did not use one.
https://wowanalyzer.com/report/9qhta2r4T3nFpvPx/20-Heroic+Varimathras+-+Kill+(3:04)/Kidsmoke
https://www.warcraftlogs.com/reports/9qhta2r4T3nFpvPx#fight=20&type=auras&source=6

Gebuz:
Looks like the issue is that he managed to prepot after the encounter started (iirc this is gonna happen a lot on Aggramar as he is slow to put the rest of the raid in combat), so solution is likely to just check if multiple pots were used during the encounter and if one of them were during the first few seconds assume it was a prepot. It wouldn't be ideal for bosses like Argus where you usually end up using 3 pots in total as it then wouldn't be able to tell the difference between prepotting too late or prepotting successfully but if we don't check for a 3rd pot anyway that is probably fine.

abelito75:
This is done closing.

PR #1357, opened 2018-02-08
Add a 1 second window for prepots

avilene:
Could be useful due to Aggramar (or other bosses) being slow at putting players in combat. #1218

### DIFF
--- a/src/Parser/Core/Modules/Items/PrePotion.js
+++ b/src/Parser/Core/Modules/Items/PrePotion.js
@@ -48,10 +48,15 @@ class PrePotion extends Analyzer {
   usedPrePotion = false;
   usedSecondPotion = false;
   neededManaSecondPotion = false;
+  maxPrepotWindow = 0;
+
+  on_initialized() {
+    this.maxPrepotWindow = this.owner.fight.start_time + 1000;
+  }
 
   on_toPlayer_applybuff(event) {
     const spellId = event.ability.guid;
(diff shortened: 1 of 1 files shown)


Gebuz (changes requested):
There needs to be a check in `on_byPlayer_cast` if the timestamp is less than `maxPrepotWindow` to ensure that a pot isn't counted as both a prepot and second pot.

In the case of a messed up prepot resulting in the player only being able to use one pot, should we display it as missing a prepot or missing a second pot? I'm not sure what's better. Either way it's probably safe to be pretty generous with the buffer window to ensure it catches all potential prepots (I'm not sure how long you actually have on aggramar).

MartijnHols:
I don't think we should increase the buffer time as my experience is that I immediately get in combat when someone else pulls (damn people landing hits on the boss at 1 or 2 seconds before end of countdown).

Gebuz:
couldn't we just assume that it is always a normal pot until a second one is used? I suppose deaths and potting again would mess with that though.

MartijnHols:
Some fights like Gul'dan offer combat resets at important moments. I think we should support that properly as well. A 1 second buffer could be solve the issue, but maybe 500ms would too (looking at the issue report it would).

MartijnHols:
Since this PR has been stale for over a month I'll close it. Please make a new PR if anyone wishes to continue work on this.

PR #1360, opened 2018-02-09
Frost shock

HawkCorrigan:
Added a Suggestion if Frost Shock is cast without the necessary buffs or resources

### DIFF
--- a/src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
+++ b/src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
@@ -16,25 +16,36 @@ class FrostShock extends Analyzer {
   badFrostShockCount = 0;
   suboptimalFrostShockCount = 0;  //not used yet, wanna talk to tc-people for thresholds
 
+  on_initialized() {
+    this.active = false;
+  }
+
   on_byPlayer_cast(event) {
-    if (event.ability.guid === SPELLS.FROST_SHOCK.id) {
-      this.active = true;
-      if (!this.combatants.selected.hasBuff(TALENTS.ICEFURY_TALENT.id, event.timestamp))
-        this.badFrostShockCount++;
+    if (event.ability.guid !== SPELLS.FROST_SHOCK.id) {
+      return;
+    }
+    this.active = true;
+    if (!this.combatants.selected.hasBuff(TALENTS.ICEFURY_TALENT.id, event.timestamp))
+      this.badFrostShockCount++;
 
-      const resource = event.classResources[0];
-      if (resource.type === RESOURCE_TYPES.MAELSTROM.id)
-        if (resource.amount < 20)
-          this.badFrostShockCount++;
+    const resource = event.classResources[0];
+    if (resource.type !== RESOURCE_TYPES.MAELSTROM.id) {
+      return;
+    }
+    
+    if (resource.amount < 20) {
+      this.badFrostShockCount++;
+    }
 
-      if (!this.combatants.selected.hasBuff(SPELLS.ELEMENTAL_FOCUS.id, event.timestamp))
-        this.suboptimalFrostShockCount++;
+    if (!this.combatants.selected.hasBuff(SPELLS.ELEMENTAL_FOCUS.id, event.timestamp)) {
+      this.suboptimalFrostShockCount++;
     }
   }
+
   suggestions(when) {
     when(this.badFrostShockCount).isGreaterThan(0)
       .addSuggestion((suggest) => {
-        return suggest(<span>Your <SpellLink id={SPELLS.FROST_SHOCK.id} /> can be improved.</span>)
+        return suggest(<span>Your <SpellLink id={SPELLS.FROST_SHOCK.id} /> usage can be improved.</span>)
           .icon(SPELLS.FROST_SHOCK.icon)
           .actual(`${this.badFrostShockCount} bad Casts. Only cast Frost Shock when you have Ice Fury-Stacks left and at least 20 Maelstrom`)
           .recommended(`0 is recommended`)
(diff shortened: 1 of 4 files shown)


HawkCorrigan:
screenshot-2018-2-9 wowanalyzer 1

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,46 @@
...
+          this.badFrostShockCount++;
+
+      if (!this.combatants.selected.hasBuff(SPELLS.ELEMENTAL_FOCUS.id, event.timestamp))
+        this.suboptimalFrostShockCount++;
+    }
+  }
+  suggestions(when) {
+    when(this.badFrostShockCount).isGreaterThan(0)
+      .addSuggestion((suggest) => {
+        return suggest(<span>Your <SpellLink id={SPELLS.FROST_SHOCK.id} /> can be improved.</span>)
```
Gebuz: Feels like it's missing a word, "Your Frost Shock _usage_ can be improved."

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,46 @@
...
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  badFrostShockCount = 0;
+  suboptimalFrostShockCount = 0;  //not used yet, wanna talk to tc-people for thresholds
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.FROST_SHOCK.id) {
+      this.active = true;
```
Gebuz: The default for `this.active` is true so no need to set it here.

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,46 @@
...
+class FrostShock extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+
+  badFrostShockCount = 0;
+  suboptimalFrostShockCount = 0;  //not used yet, wanna talk to tc-people for thresholds
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.FROST_SHOCK.id) {
```
Gebuz: It's preferred to avoid nested if statements when possible. Instead check for the opposite and return early.

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,46 @@
...
+      this.active = true;
+      if (!this.combatants.selected.hasBuff(TALENTS.ICEFURY_TALENT.id, event.timestamp))
+        this.badFrostShockCount++;
+
+      const resource = event.classResources[0];
+      if (resource.type === RESOURCE_TYPES.MAELSTROM.id)
```
Gebuz: Same as above, check for opposite and return early instead.

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,46 @@
...
+    combatants: Combatants,
+  };
+
+  badFrostShockCount = 0;
+  suboptimalFrostShockCount = 0;  //not used yet, wanna talk to tc-people for thresholds
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.FROST_SHOCK.id) {
+      this.active = true;
+      if (!this.combatants.selected.hasBuff(TALENTS.ICEFURY_TALENT.id, event.timestamp))
```
Gebuz: Please use curly brackets even if the following code is only one line. Same thing with the next few if statements.

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
Gebuz: Oh I just realized, I assume this line is supposed to be here and you forgot to hide it by default?

HawkCorrigan:
Alright should be fixed now.

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,57 @@
...
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.FROST_SHOCK.id) {
+      return;
+    }
+    this.active = true;
+    if (!this.combatants.selected.hasBuff(TALENTS.ICEFURY_TALENT.id, event.timestamp))
+      this.badFrostShockCount++;
+
+    const resource = event.classResources[0];
```
Gebuz: I missed this the first time around, but you should Ideally check if `classResources` is null. You can also use a filter to find the correct resource like this: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/Modules/ManaValues.js#L23
Anyway I don't think this should cause problems as Frost Shock always have a resource event afaik, so it's mostly a thing for the future :)

src/Parser/Shaman/Elemental/Modules/ShamanCore/FrostShock.js
```diff
@@ -0,0 +1,57 @@
...
+    if (resource.amount < 20) {
+      this.badFrostShockCount++;
+    }
+
+    if (!this.combatants.selected.hasBuff(SPELLS.ELEMENTAL_FOCUS.id, event.timestamp)) {
+      this.suboptimalFrostShockCount++;
+    }
+  }
+
+  suggestions(when) {
```
Gebuz: In the future I'd recommend that you use the newer format for suggestions. Example here: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Druid/Balance/Modules/Features/AlwaysBeCasting.js

PR #1361, opened 2018-02-09
make it so the ES suggestion doesn't show when you didn't use ES

HawkCorrigan:
Title basically


HawkCorrigan:
screenshot-2018-2-9 wowanalyzer mythic antoran high command - kill 7 17 by murshock
before

screenshot-2018-2-9 wowanalyzer
after i guess

src/Parser/Shaman/Elemental/Modules/ShamanCore/EarthShock.js
```diff
@@ -135,23 +137,26 @@ class EarthShock extends Analyzer {
         });
     }
     else {
-      when(this.outAscSuggestionThreshold)
+      if (this.avgMaelstromSpendOutOfAsc > 0) {
```
Gebuz: should merge this with the else above

PR #1362, opened 2018-02-09
Power Word Radiance Normalizer

kpralow:
Because of latency, after casting Power Word: Radiance, the atonement applications
  don't always register right away and an instant cast spell can be recorded before
  the atonements. Example: Evangelism casted right after a Power Word Radiance will
  be before the applications of the atonements. For this reason we reorder the
  events so that the applications are always right after the cast.


MartijnHols:
Is it possible for a PW: R `applybuff` to occur prior to the `cast` event? If not you might want to remove the buffer time as it is likely to only cause issues, or better yet increase it to like 3sec.

kpralow:
> it possible for a PW: R applybuff to occur prior to the cast event? 
Doubt it, don't know how it could happen

> or better yet increase it to like 3sec.
If I increased the buffer to 3 secondes, a player applying atonement via anything else would be considered. The max I could put is 1 sec (minimal GCD). What do you think?

PR #1366, opened 2018-02-10
[core][feral druid] Enhancements to ResourceTracker, Feral CP tracker now uses core ResourceTracker

kfinch:
Wow, this one kinda got away from me. I was intending to add 'natural resource regeneration' to ResourceTracker while at the same time hooking in the core ResourceTracker to Feral, and I didn't even get to the natural regen bit.

 I haven't included screenshots because everything looks the same lul.

The most changes are to the core ResourceTracker, which are as follows:
- Lots of comment instructions added to help new extenders of the class
- Rather than taking a resourceId and resourceName, implementers now just pass in the 'resource' object from RESOURCE_TYPES. All the changes to the non-Feral ResourceTrackers are just updating for this refactor. This will allow in the future adding automatic icons for the tracking, etc...
- 'generated', 'wasted', 'spent', and 'spendersCasts' are now got from getter methods that generate the values from the buildersObj and spendersObj. This allows these numbers to automatically reflect custom changes made there (this happens in the Feral implementation).
- 'spentByCast' obj added to spendersObj, which adds a history of how much was spent by a spell per cast. This allows implementations to generate more detailed stats about spenders. The use case I had particularly in mind was implementing the "casts at max CPs" feature for Feral.
- Added 'processInvisibleEnergize' function that can be used by implementing specs that have energizes that don't show in events (like Feral).

Other changes:
- Total replacement of the Feral Druid ComboPoint tracking classes to use the core version instead. I still had to use a custom implementation of the CPs tab, because for Feral we want a "casts at max CPs" column instead of a "total casts" column. This technically preserving previous functionality, as the previous Feral CPs tab implementation had this.
- Other classes ResourceTracker implementations switched over to extending 'resource' instead of 'resourceType' and 'resourceName'
- Added Primal Fury and Umbral Glaive Storm to CASTS_THAT_ARENT_CASTS, and also added SPELLS entries for a couple spells that were included only by id.
- While it looks like src/Parser/Druid/Feral/Modules/ComboPoints/ComboPointDetails.js has a lot of changes, it functions the exact same, it's just reading from a different API (the core ResourceTracker).
- src/Parser/Druid/Feral/Modules/ComboPoints/ComboPointDetails.js is basically deleted, as it now just extends the core ResourceTracker. It still implements a couple special Feral features, which can be seen documented in its new form here https://github.com/kfinch/WoWAnalyzer/blob/ee562527e3ec11c64f24a2415f699fe72c0ad3f0/src/Parser/Druid/Feral/Modules/ComboPoints/ComboPointTracker.js


kfinch:
Two more things I forgot about:
- Added Swipe to list of invisible CP energizes (wasn't listed in version on live, but confirmed it doesn't energize)
- Added Berserk to cooldowns to show in cooldown tab (Incarnation was there, but no Berserk for some reason)

Gebuz:
Looks good to me. Mind checking if this breaks anything for the DK runetracker? It shouldn't but since I manually add the passive rune regn. it might. It's not major, I'd just like to know so I can fix it asap. I plan to switch to using your `processInvisibleEnergize` for it anyway, I just need to figure out how to contribute parts of the rune regeneration to accelerators first.

kfinch:
Apparently my search/replace for the changed properties didn't hit DK for some reason... in any case yeah I need to make some updates to RuneTracker.

kfinch:
@Gebuz Ok with the updates to RuneTracker everything appears to work as intended. I verified by comparing the Rune Usage tab on a live version vs my dev version, and all numbers / bars / graph points appeared identical.

As you can see from the most recent commits, there are two fairly minor API changes in ResourceTracker: `resourceName` is now `resource.name`, `resourceType` is now `resource.id`, and the class wide `generated` and `wasted` are now no longer settable, they're populated automatically from the sum of all the spell's generated and wasted amounts.

Relatedly, I may have caught a very minor bug in `addPassiveRuneRegeneration()` ... Generated and wasted were added to but the global amount never was. I've left the function unchanged because now the global values are added to automatically.

kfinch:
@Gebuz Tsabo and Teace signed off on it, so want to do an actual review? Or approve if you're already satisfied with the state of it.

PR #1375, opened 2018-02-11
[Resto Shaman] 7.3.5 bug fixes and general updates

niseko:


Pewtro:
We generally don't include cast time into the cooldowns of spells, alter the cast-efficiency instead if necessary.

niseko:
I'm not a fan of 83% being the highest possible effectiveness reachable in a scale of 0-100% (for healing rain). if the best thing to do in a fight is to have 100% rain uptime, its kinda misleading to see 8 missed casts of 25 "maximum possible casts" when in reality its more about 3 missed out of 20.
i'm not too familiar with the code yet so maybe theres a different way to go about it

Pewtro:
Our cast efficiency is about "time spent on cooldown", which is why we use the cooldown as is, and not fidget with it. None of our other casttime abilities do what you want to my knowledge either, and that would make this an anomaly and could lead to confusion amongst users if things were functioning differently in baseline modules across classes/specs. 
You mentioned Healing Rain uptime being the important metric, and that could be its own module, from where you can make suggestions to the player.
As I said you could also change the castEfficiency thresholds to alter when it will suggest the player to cast it more (but it already seems to be at a low number before triggering from what I can see)

kfinch:
@niseko I know it's awkward, but yeah for now don't add the cast time to cooldown. Fixes have been considered, but for now it's a very small offset.

For Healing Rain it's a large offset, but also it looks like cast efficiency isn't the right metric? Shouldn't the relevant thing to track be Healing Rain uptime? (unless you're allowed to and supposed to have multiple rains up at once, I don't know much about Shaman)

niseko:
alright, i see your points and i removed the Abilities.js changes. My issue was mostly the cast efficiency tab showing the wrong amount of maximum possible casts, as it ignores casting time - but i see that this is the wrong way to go about that.
image
 I'll look for a different solution later on.

kfinch:
You're right, a spell with a long cast time and a short cooldown doesn't look right at all in cast efficiency. I'll look again into providing a more durable solution for this.

MartijnHols:
I think this is your first PR to WoWAnalyzer @niseko, thanks a lot for your contribution and welcome on board! It looks like you're getting comfortable with the Restoration Shaman analyzer, I hope you'll figure out a solution to the cast time problem. Your changes should be live on wowanalyzer.com by now, please let us know if you need anything further.

PR #1376, opened 2018-02-11
[resto druid] Fixed some Tree of Life bugs

### LINKED ISSUE #1369: [Resto Druid] Chameleon Song uptime estimate incorrect
See the following:

https://wowanalyzer.com/report/bfXkVDtYRZd6HABj/25-Heroic+Aggramar+-+Kill+(6:06)/4-Hatsya
https://www.warcraftlogs.com/reports/bfXkVDtYRZd6HABj#fight=25&type=auras&source=4

The total duration of Incarnation is approximately 84 seconds, and the two events are 42 seconds long each. That should be exactly 1 cast of Incarnation + 1 proc of Chameleon Song each, and Chameleon Song should be valued at 24 seconds of uptime.

Instead, the analyzer computes its uptime at 18 seconds.

While ToL buff extension appears supported in the code, I suspect that the issue is shifting out of ToL and back into ToL during one of the extended ToLs. This appears to trigger an extra cast of Incarnation in the logs.

kfinch:
In some circumstances, Tree of Life could generate extra casts events, so I added a Normalizer to remove them. For the gritty details, see my documentation in the Normalizer.

Also made the CS proc count calculator less dumb, it should no longer under count if the fight ends during a proc.

Finally, made the reported uptime be the amount of time the player was allowed to be in the form rather than the time the player was actually in form. This may seem odd, but some of a resto's movement / defensive abilities cause the player to briefly drop form, and it was making the numbers look confusing / weird.

Fixes #1369

### DIFF
--- a/src/Parser/Druid/Restoration/CHANGELOG.js
+++ b/src/Parser/Druid/Restoration/CHANGELOG.js
@@ -10,6 +10,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-02-11'),
+    changes: <Wrapper>Fixed a bug that could cause incorrect proc counts and uptimes for players using <SpellLink id={SPELLS.INCARNATION_TREE_OF_LIFE_TALENT.id} icon /> and <ItemLink id={ITEMS.CHAMELEON_SONG.id} icon /> together.</Wrapper>,
+    contributors: [sref],
+  },
+  {
     date: new Date('2018-02-03'),
     changes: 'Added a Checklist to the top of the page, which is a more organized and better explained version of the Suggestion system.',
     contributors: [sref],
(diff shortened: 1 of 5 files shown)


Gebuz:
Just a minor thing, since all the druid specs have a similar buff for their respective incarnations (I assume with different IDs) it might be better to name it something with tree of life instead of just Incarnation.

kfinch:
It sure would be nice to name it something else, but unfortunately that's what Blizzard called it. I didn't want to make it too easy to mix up with INCARNATION_TREE_OF_LIFE_TALENT either. I figured the comment on it would suffice.

Pewtro:
Couldn't it maybe be called something like ``INCARNATION_ABLE_TO_ASSUME_TOL`` ?

kfinch:
Ehh there we go, I guess it does make things more obvious on reading it.

PR #1377, opened 2018-02-12
[Frost Death Knight] Gathering Storm Talent

jjs451:
Adds a stat box for Gathering Storm

image


src/Parser/DeathKnight/Frost/Modules/Talents/GatheringStorm.js
```diff
@@ -0,0 +1,102 @@
...
+    debug && console.log(`removed buff`);
+  }
+
+  on_byPlayer_cast(event){
+    const resource = event.classResources[0];
+    if(event.ability.guid === SPELLS.HOWLING_BLAST.id && this.combatants.selected.hasBuff(SPELLS.RIME.id)){
+      this.extendedDuration += DURATION_BOOST;
+      return;
+    }    
+    if(resource.type !== 5){
```
Gebuz: You could use RESOURCE_TYPES.RUNES.id instead of 5 here.

src/Parser/DeathKnight/Frost/Modules/Talents/GatheringStorm.js
```diff
@@ -0,0 +1,102 @@
...
+      return;
+    }
+    this.extendedDuration = resource.hasOwnProperty("cost") ? this.extendedDuration + (DURATION_BOOST * resource.cost): this.extendedDuration;
```
Gebuz: I assume that obliterate would extend it by 1 sec even if it was cast with obliteration since HB with rime also extends it but just in case it doesn't: Resource cost reduction isn't accounted for in `resource.cost` so if this isn't the case you'd have to check for obliteration and manually subtract one rune.

Gebuz:
There is a changelog merge conflict because I merged your other PR, otherwise it should be good to merge!

src/Parser/DeathKnight/Frost/Modules/Talents/GatheringStorm.js
jjs451: Yeah Obliteration counts as two runes spent the same way Rime counts for its cost

PR #1380, opened 2018-02-12
[Survival] 2 talents and minor fixes

Pewtro:
Creates a module for Butchery and Carve and one for Mortal Wounds

Fixes a console.log always appearing. 
Adds some suggestions to Butchers Bone Apron legendary. 

Mortal Wounds: 
Log: https://www.warcraftlogs.com/reports/vfqmzPjG9J7Kt3aC#fight=4&type=damage-done
image

Butchery: 
Log: https://www.warcraftlogs.com/reports/vfqmzPjG9J7Kt3aC#fight=4&type=damage-done
image

Carve: (also shows the tooltip without anything because not using butchers bone apron)
Log: https://www.warcraftlogs.com/reports/qdKTjCMhDp3WBRVb#fight=3&type=damage-done
image

Butchers Bone Apron suggestions: 
Log: https://www.warcraftlogs.com/reports/vfqmzPjG9J7Kt3aC#fight=4&type=damage-done
image


src/Parser/Hunter/Survival/Modules/Talents/ButcheryCarve.js
```diff
@@ -0,0 +1,103 @@
...
+        average: 1.75,
+        major: 1.5,
+      },
+      style: 'number',
+    };
+  }
+
+  suggestions(when) {
+    const spellLink = this.combatants.selected.hasTalent(SPELLS.BUTCHERY_TALENT.id) ? SPELLS.BUTCHERY_TALENT : SPELLS.CARVE;
+    when(this.averageTargetsThreshold).addSuggestion((suggest, actual, recommended) => {
```
Gebuz: Assuming butcher's bone apron makes it worth casting butchery/carve single target, I think it would be a good idea to use another threshold when it's equipped.

src/Parser/Hunter/Survival/Modules/Talents/ButcheryCarve.js
```diff
@@ -0,0 +1,103 @@
...
+        .icon(spellLink.icon)
+        .actual(`${this.averageTargetsHit} targets hit on averages`)
+        .recommended(`>${recommended} targets hit on average is recommended`);
+    });
+  }
+
+  statistic() {
+    if (this.bonusDamage > 0) {
+      const spellLink = this.combatants.selected.hasTalent(SPELLS.BUTCHERY_TALENT.id) ? SPELLS.BUTCHERY_TALENT : SPELLS.CARVE;
+      const tooltipText = this.combatants.selected.hasChest(ITEMS.BUTCHERS_BONE_APRON.id) ? `You had an average of ${(this.totalStacksUsed / this.casts).toFixed(1)} stacks of the Butchers Bone Apron buff when casting ${spellLink.name}` : ``;
```
Gebuz: To me it would make more sense to have this in the items tooltip, but I'm no hunter so if you think it fits better here keep it.

src/Parser/Hunter/Survival/Modules/Items/ButchersBoneApron.js
```diff
@@ -78,11 +84,64 @@ class ButchersBoneApron extends Analyzer {
...
+  }
+
+  get unusedStacks() {
+    return this.totalStacks - this.usedStacks - this.wastedStacks;
+  }
+  get percentUnusedStacks() {
+    return this.unusedStacks / this.totalStacks;
+  }
+
+  get unusedStacksThreshold() {
```
Gebuz: This is one of the few cases where I think it makes sense to use a specific number instead of a percentage since it only ever happens once per encounter and a shorter/longer encounter shouldn't change how many stacks it's worth casting it at (I assume at X number of stacks it's worth casting single target? might also be different depending on carve/butchery?).

src/Parser/Hunter/Survival/Modules/Talents/ButcheryCarve.js
Pewtro: Yeah I contemplated both - since you suggested it aswell, I'll move it over.

src/Parser/Hunter/Survival/Modules/Items/ButchersBoneApron.js
Pewtro: My idea was that it would be less of an issue (since it would be less impactful) over the course of a longer fight than over a shorter fight, but I'll change it to numbers

src/Parser/Hunter/Survival/Modules/Talents/ButcheryCarve.js
Pewtro: Added a note in the suggestion to clarify possible usage if using Butchers Bone Apron on ST, but recommending to try and use something else.

PR #1391, opened 2018-02-15
[Assassination Rogue] Added support for T21 2P

adilasif:
Also slightly refactored Duskwalker's Footpads for clarity and adjusted rogue spells list to allow for proper display of T21 2P information on results page.


src/Parser/Rogue/Assassination/Modules/Legendaries/DuskwalkersFootpads.js
```diff
@@ -35,13 +35,12 @@ class DuskwalkersFootpads extends Analyzer {
...
+    const reduction = spent * VENDETTA_CDR_PER_ENERGY;
 
     if (!this.spellUsable.isOnCooldown(SPELLS.VENDETTA.id)) {
-      this.wastedReduction += spent * VENDETTA_CDR_PER_ENERGY;
+      this.wastedReduction += reduction;
     } else {
 
-      const reduction = spent * VENDETTA_CDR_PER_ENERGY;
-
       this.spellUsable.reduceCooldown(SPELLS.VENDETTA.id, reduction * 1000);
```
Gebuz: The return value of this is the effective reduction (if you're at 2 sec cd remaining and reduce it by 3 sec this would return 2), so you could add this value instead of `reduction` to `this.wastedReduction` for a slightly more accurate result.

src/Parser/Rogue/Assassination/Modules/Items/T21_2P.js
```diff
@@ -0,0 +1,34 @@
...
+    return this.combatants.selected.getBuffUptime(SPELLS.ASSA_ROGUE_T21_2SET_BONUS_BUFF.id) / this.owner.fightDuration;
+  }
+
+  item() {
+    return {
+      id: SPELLS.ASSA_ROGUE_T21_2SET_BONUS_BUFF.id,
+      icon: <SpellIcon id={SPELLS.ASSA_ROGUE_T21_2SET_BONUS.id} />,
+      title: <SpellLink id={SPELLS.ASSA_ROGUE_T21_2SET_BONUS.id} />,
+      result: (
+        <dfn data-tip={`It's worth keeping this in mind and trying to maximize your uptime, but not at the expense of your primary rotation. This should, most of the time, naturally happen in the process of clipping your Envenom debuff.`}>
```
Gebuz: I would consider removing the tooltip as it doesn't provide any extra information. If there is a certain uptime you should aim for it would be more appropriate to have as a suggestion.

adilasif: The uptime for this (and the Envenom debuff) should ideally be around ~79-80% in a perfectly played Patchwerk scenario with well-itemized gear. I'll amend the tooltip to mention that. Thanks for the tip!

PR #1395, opened 2018-02-16
[Windwalker] Spinning crane kick module

Juko8:
image
The "Bad Casts" part is still expandable on some more complicated calculations, but i believe it's useful as long as it isn't telling anyone they're making mistakes if they aren't. The bad casts tracking will spot people using Spinning Crane Kick on single target or without tagging anything, which is always a mistake, but using it with worse DPChi isn't always strictly a mistake and also way less impactful on the players damage.


src/Parser/Monk/Windwalker/Modules/Spells/SpinningCraneKick.js
```diff
@@ -0,0 +1,178 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  badCastsStatistic() {
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <dfn data-tip={`Bad casts is currently only counting casts with lower DPET (Damage Per Execute Time) than Blackout Kick.`}>
```
MartijnHols: This can be just a string (`data-tip="..."`).

src/Parser/Monk/Windwalker/Modules/Spells/SpinningCraneKick.js
```diff
@@ -0,0 +1,178 @@
...
+  statistic() {
+    if (this.abilityTracker.getAbility(SPELLS.SPINNING_CRANE_KICK.id).casts > 0) {
+      return (
+        <StatisticsListBox
+          title={
+            <SpellLink id={SPELLS.SPINNING_CRANE_KICK.id}>
+              <SpellIcon id={SPELLS.SPINNING_CRANE_KICK.id} noLink /> Spinning Crane Kick
+          </SpellLink>
+          }
+          tooltip=""
```
MartijnHols: This can be omitted.

src/Parser/Monk/Windwalker/Modules/Spells/SpinningCraneKick.js
```diff
@@ -0,0 +1,178 @@
...
+        .regular(3).major(5);
+      });
+  }
+
+  averageHits() {
+    const averageHits = this.spinningCraneKickHits / this.abilityTracker.getAbility(SPELLS.SPINNING_CRANE_KICK.id).casts;
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <dfn data-tip={`Spinning Crane Kick hits all nearby enemies 4 times over 1.5 seconds`}>
```
MartijnHols: This can be just a string (`data-tip="..."`).

src/Parser/Monk/Windwalker/Modules/Spells/SpinningCraneKick.js
```diff
@@ -0,0 +1,178 @@
...
+    if (event.timestamp - this.lastSpinningCraneKickTick > this.spinningCraneKickDuration) {
+      this.totalMarksDuringHits += this.markoftheCraneStacks;
+      this.lastSpinningCraneKickTick = event.timestamp;
+    }
+  }
+
+  suggestions(when) {
+    when(this.badCasts).isGreaterThan(0).addSuggestion(
+      (suggest, actual, recommended) => {
+        return suggest(<span>You have ineffecient casts of <SpellLink id={SPELLS.SPINNING_CRANE_KICK.id} /> </span>)
```
MartijnHols: Please use the `Wrapper` component from `common/Wrapper` instead of `span` when the container has no function.

src/Parser/Monk/Windwalker/CHANGELOG.js
```diff
@@ -1,4 +1,5 @@
 export default `
+16-02-2018 - Added Spinning Crane Kick statistics and suggestion on bad use of Spinning Crane Kick (Juko)
```
MartijnHols: The string based changelog style is deprecated. Please consider updating the changelog format to the new style when you are able to.

MartijnHols:
Thanks for your continued contributions @Juko8! I left some comments above that you could consider for improved maintainability and to make it easier to understand. Please consider them for a future PR.

PR #1405, opened 2018-02-18
[Core] Stats, Talents, Gear Layout Changes

### LINKED ISSUE #211: Add a character overview (gear/talents/traits/precombat buffs/stats)
Might also want to show the throughput value for items we know.

Might be interesting for later where we start putting throughput gain values to stats.

### LINKED ISSUE #1260: Interface updates
I've received a suggestion for little updates on the interface. 

1) The items, stats on pull and talents could be collapsible (just as Checklist Rules are), if the person doesn't want to see them
image
image

2) Another idea would be to join them + the gear tab to something similar to this (which could also be collapsible):
image

I personally like option 1 a lot, and option 2 as well, my only concern would be that it would take up a lot of space, or suddenly a lot of space would be empty if it was collapsed. Just leaving it as a suggestion.

Sharrq:
Moves the Stats on Pull, Talents, and Gear into a Character Panel alongside Checklist and Suggestions. This also reduces scrolling by removing Stats and Talents from the side Panel on the left side.

Additionally, I played around with the talents and stats layout and hid tertiary stats that the user doesnt have (if they have 0 rating)

I left the Gear Layout mostly the same and just added a heading to it, im sure more can be done to make this look better, perhaps to mirror the character sheet layout in game, but without item names or the realm name to get the character picture, this is probably good for now.

Fixes #1260
Fixes #211


Sharrq:
Default on load
image

Character Panel selected
image

Pewtro:
This looks super slick to me and removes a personal pet peeve of mega scrolling. 👍

Gebuz:
Agreed, looks awesome!

Sharrq:
oops, missed the divider when i submitted it. Ill get that fixed.

Sharrq:
Alright, the patch compat and divider changes should be removed now.

Sharrq:
added the changelog entry

MartijnHols:
Thanks! This is really neat, great job! :D

PR #1406, opened 2018-02-19
[Frost Death Knight] Add Frostscythe module

jjs451:
Adds a statistic box and suggestion related to Frostscythe usage

image
image

### DIFF
--- a/src/Parser/DeathKnight/Frost/Modules/Talents/Frostscythe.js
+++ b/src/Parser/DeathKnight/Frost/Modules/Talents/Frostscythe.js
@@ -1,0 +1,98 @@
+import React from 'react';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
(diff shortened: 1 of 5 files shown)


src/Parser/DeathKnight/Frost/Modules/Talents/Frostscythe.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+
+class Frostscythe extends Analyzer{
+  static dependencies = {
+    combatants: Combatants,
+  }
+
+  casts = 0;
+  hits = -1; // need to initialize negative to make sure first cast isn't counted as bad
```
Gebuz: Just a bit confusing comment, I assume you used to count bad casts and forgot to change the comment? From what I can see it is to prevent adding an extra "good" cast on first cast event (or finished with no casts)?

jjs451: Maybe I should've changed the comment to say "good" instead of "bad" but yeah, its to prevent the if block in `on_cast` from firing on the first cast of frostscythe

PR #1411, opened 2018-02-19
[Havoc] Fury suggestion

CollCrom:
Added in a wasted fury suggestion and a stat box to show how much you wasted.
image
Also updated the Delusions of Grandeur Module suggestion. Since it takes the average cooldown reduction  it didn't account for very short fights with lust where you actually got a 2nd Meta off.

test log: https://www.warcraftlogs.com/reports/BHyhvwRqYjrMVfk8#fight=13&type=damage-done


Gebuz:
Wouldn't it be easier to read if you swapped the % actual and the exact number as to have the percentages in the same line?

Also one thing to consider about the meta check, a cast 1 sec before the boss dies would hide the suggestion. Maybe you could check how close to the encounter end it was cast and if more than X% was wasted it was still a bad desicion to use shoulders?

CollCrom:
Good point. I updated it so that if you are getting less than 50% of your Meta buff it throws the suggestion.
here are some logs I used to test:
short fight 1 meta: https://www.warcraftlogs.com/reports/yqKaGptwXPRYr8zm#fight=6&type=summary&source=19
short fight but only got the 2nd meta for 12s: https://www.warcraftlogs.com/reports/JckrzBH2hYXQ197K#fight=6&type=casts&source=3&ability=200166
short fight but had 2nd meta for 20s: https://www.warcraftlogs.com/reports/BHyhvwRqYjrMVfk8#fight=1&type=casts&source=1&view=timeline&ability=200166

PR #1412, opened 2018-02-20
[Havoc] Added in Anger of the Half Giants

CollCrom:
added in a legendary stat box for [Anger of the Half Giants](http://www.wowhead.com/item=137038/anger-of-the-half-giants). Also since I changed the holy power tracker there was a bug with the Liadrin's Module that I also fixed.
image
and 
image
test logs:
https://www.warcraftlogs.com/reports/BHyhvwRqYjrMVfk8#fight=13&type=damage-done
https://www.warcraftlogs.com/reports/2QMTZNPmaWkwX1Az#fight=9&type=summary

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Items/AngerOfTheHalfGiants.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Items/AngerOfTheHalfGiants.js
@@ -4,7 +4,8 @@ import Wrapper from 'common/Wrapper';
 import SPELLS from 'common/SPELLS';
 import ITEMS from 'common/ITEMS';
 import SpellLink from 'common/SpellLink';
-import { formatNumber } from 'common/format';
+import ItemLink from 'common/ItemLink';
+import { formatNumber, formatPercentage } from 'common/format';
 
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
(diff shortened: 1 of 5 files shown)


src/Parser/DemonHunter/Havoc/Modules/Items/AngerOfTheHalfGiants.js
```diff
@@ -0,0 +1,82 @@
...
+		}
+		return 0;
+	}
+
+	get furyWasted() {
+		if(this.angerFury){
+			return this.angerFury.wasted;
+		}
+		return 0;
+	}
```
Pewtro: You never use this bit of code here?

src/Parser/DemonHunter/Havoc/Modules/Items/AngerOfTheHalfGiants.js
```diff
@@ -0,0 +1,82 @@
...
+	get totalFury() {
+		return this.furyGenerated + this.furyWasted;
+	}
+
+	item() {
+		const builderId = this._hasDemonBlades ? SPELLS.DEMON_BLADES_TALENT.id : SPELLS.DEMONS_BITE.id;
+		return {
+			item: ITEMS.ANGER_OF_THE_HALF_GIANTS,
+			result: (
+				<dfn data-tip={`Total Fury Gained: <b>${formatNumber(this.totalFury)}</b>.`}>
```
kfinch: Should show actual gain, not the amount would have gotten with perfect play (so show `this.furyGenerated` instead of `this.totalFury`)

src/Parser/DemonHunter/Havoc/Modules/Items/AngerOfTheHalfGiants.js
CollCrom: Yeah I meant to add a suggestion but I forgot. Fixed!

PR #1413, opened 2018-02-20
[Havoc] Moarg Bionic Stabiliziers

CollCrom:
added in [Moarg](http://www.wowhead.com/item=137090/moarg-bionic-stabilizers)

image
Worn by Angrydomerr in [this log](https://www.warcraftlogs.com/reports/2gLFBpMJXtcRajwr#fight=37&type=damage-done)

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Items/MoargBionicStabilizers.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Items/MoargBionicStabilizers.js
@@ -1,0 +1,70 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import ItemDamageDone from 'Main/ItemDamageDone';
+
+const MOARG_MODIFIER = 0.25;
+
+const MS_BUFFER = 1000;
+
+/**
+* Equip: Throw Glaive deals 25% increased damage for each enemy hit.
+**/
+
+class MoargBionicStabiliziers extends Analyzer {
+	static dependencies = {
+		combatants: Combatants,
+	};
+	rank = 0;
+	bonusDamage = 0;
+	lastCastTimestamp = 0;
+	enemiesHit = 0;
+	damagePreCalc = 0;
+	modifier = 0;
+
+	on_initialized() {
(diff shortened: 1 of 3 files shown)


src/Parser/DemonHunter/Havoc/Modules/Items/MoargBionicStabilizers.js
```diff
@@ -0,0 +1,70 @@
...
+	lastCastTimestamp = 0;
+	enemiesHit = 0;
+	damagePreCalc = 0;
+	modifier = 0;
+
+	on_initialized() {
+		this.active = this.combatants.selected.hasWrists(ITEMS.MOARG_BIONIC_STABILIZERS.id);
+	}
+
+	get averageTargetsHit() {
```
Gebuz: This isn't used and neither of the variables exists, I assume you forgot about it? :P would be nice to have in the tooltip.

src/Parser/DemonHunter/Havoc/Modules/Items/MoargBionicStabilizers.js
```diff
@@ -0,0 +1,70 @@
...
+		}
+		this.lastCastTimestamp = event.timestamp;
+    this.enemiesHit = 0;
+    this.damagePreCalc = 0;
+	}
+
+	on_byPlayer_damage(event) {
+		const spellId = event.ability.guid;
+		if (this.lastCastTimestamp && (this.lastCastTimestamp + MS_BUFFER < event.timestamp) && this.damagePreCalc > 0) {
+			this.modifier = MOARG_MODIFIER * this.enemiesHit;
```
Gebuz: you could make this a constant since it is only used within this scope

CollCrom:
It does add to bloodlet damage but I am currently not sure how to calculate that. I added a note in the tooltip stating as such however.

src/Parser/DemonHunter/Havoc/Modules/Items/MoargBionicStabilizers.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+* Equip: Throw Glaive deals 25% increased damage for each enemy hit.
+**/
+
+class MoargBionicStabiliziers extends Analyzer {
+	static dependencies = {
+		combatants: Combatants,
+		abilityTracker: AbilityTracker,
+	};
+	rank = 0;
```
Gebuz: this var isn't used

Gebuz:
Yeah bloodlet is a pain because of the ignite effect. I think it's alright aslong as it is mentioned in the tooltip, so if you fix the merge conflict in the changelog it should be fine to merge.

PR #1419, opened 2018-02-20
[Marksmanship] lnl update

Pewtro:
Removes some magic numbers from Lock and Load module
Adjust cancelled casts threshold
Wording in Aspect of the Wild module

Now to the fun bit: 
Includes normal distribution calculations for Lock and load procs so that we can estimate how lucky/unluck a player was within a certain degree of certainty.  

Standard amount of procs: 
image
Log: https://www.warcraftlogs.com/reports/B3GJkAKLh6bnVf2Q/#fight=3&source=3

Low amount of procs:
image
Log: https://www.warcraftlogs.com/reports/aTPgVxjZA3BvmGLM/#fight=9&source=89

High amount of procs: 
image
Log: https://www.warcraftlogs.com/reports/tC182dPncafY6Bmh#fight=15&type=damage-done&source=49

Since taking the pictures I updated when it says "proc" or "procs", but that's the only thing.


src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -81,14 +84,77 @@ class LockAndLoad extends Analyzer {
...
+
+    let factK = 1;
+    let sum = 0;
+    let term = 1;
+    let k = 0;
+    const loopStop = Math.exp(-23);
+
+    while (Math.abs(term) > loopStop) {
+      term = 0.3989422804 * Math.pow(-1, k) * Math.pow(z, k) / (2 * k + 1) /
+        Math.pow(2, k) * Math.pow(z, k + 1) / factK;
```
kfinch: https://i.imgur.com/TrymOlU.png

Pewtro: I got help from stackoverflow 😅 but verified it works

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -81,14 +84,77 @@ class LockAndLoad extends Analyzer {
...
+      sum += term;
+      k++;
+      factK *= k;
+    }
+    sum += 0.5;
+
+    return sum;
+  }
+
+  binomialCalculation(procs, tries, proc_chance) {
```
emallson: statistics :yum:

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -81,14 +84,77 @@ class LockAndLoad extends Analyzer {
...
+    //Correcting for continuity we add 0.5 to procs, because we're looking for the probability of getting at most the amount of procs we received
+    // if P(X <= a), then P(X<a+0.5)
+    const correctedProcs = procs + 0.5;
+    const nonProcChance = 1 - proc_chance;
+
+    const stddev = Math.sqrt(proc_chance * nonProcChance * tries);
+    //zScore is calculated by saying (X - M) / stddev
+    const zScore = (correctedProcs - this.pn) / stddev;
```
emallson: Why use the z-score and not the [CDF](https://en.wikipedia.org/wiki/Cumulative_distribution_function) of the binomial distribution?  The CDF of the binomial distribution has [simpler closed form](https://en.wikipedia.org/wiki/Binomial_distribution#Cumulative_distribution_function). From your PR comments it looks like you use the binomial distribution to model it (correctly, i think) and then use the normal distribution to compute the probability of observing a number of procs this-good-or-less? It should be the binomial distribution for both.

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
```diff
@@ -81,14 +84,77 @@ class LockAndLoad extends Analyzer {
...
+
+  //pn is the mean value of non-procs
+  get qn() {
+    return (1 - PROC_CHANCE) * this.autoShots;
+  }
+
   statistic() {
-    const expectedProcs = this.autoShots * 0.08;
+    const binomCalc = this.binomialCalculation(this.totalProcs, this.autoShots, PROC_CHANCE);
+    let tooltipText = `You had ${this.noGainLNLProcs} ${this.noGainLNLProcs > 1 ? `procs` : `proc`} with 2 lnl stacks remaining and ${this.halfLNLProcs} ${this.halfLNLProcs > 1 ? `
```
emallson: should the tooltip use "LnL" instead of "lnl"?

src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js
Pewtro: Calculation of the factorials makes it crash, as it far exceeds call stack size.

emallson: OK so I did some more reading and it looks like your normal distribution method is probably fine. I was unfamiliar with the use of the normal distribution to approximate the binomial distribution. I [implemented the CDF directly](https://github.com/emallson/WoWAnalyzer/blob/8fa0fff249324d7bbb13cc3860b3113b57ed4d2a/src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js#L91-L109) instead and it seems to hold up, although the (n choose k) bit may break as n and k grow with larger fights (i tested it on some ~7-8 minute fights and it still seems fine, but WCL is down right now so I can't do much testing).

There's also an [alternate implementation](https://github.com/emallson/WoWAnalyzer/blob/8bc7fcc0c6a9e2912e64ddf30161a4b9d76a5ac5/src/Parser/Hunter/Marksmanship/Modules/Talents/LockAndLoad.js#L91-L106) I have that trades risk of overflow in `choose(n ,k)` for risk of loss-of-precision from small floats (which is fairly low risk for the range of floats we'll ever realistically see). (This impl is based on decomposing the `choose(n, k)` term into the individual elements of its product). Both give identical values on every log I've tested (and all are within the 2% margin given by your normal distribution-based implementation).

So, ultimately, which implementation to use (normal, direct, alternate) is up to you. If you want to use the normal one you have right now lemme know and I can approve (maybe once you adjust that tooltip thing?)

PR #1421, opened 2018-02-22
[Retribution] Checklist Refactor

CollCrom:
new:
image
old:
image

[Test log](https://www.warcraftlogs.com/reports/CxgM2KP8rbLqQ43R#fight=27&type=summary&source=125)


Gebuz:
A few suggestions for future changes: 

If you inverse the 3 resource wasted to show efficiency instead it makes it easier to understand the relation between the bar and the percentage.

I also think you should consider revising the order of the rules so it goes from most important to least important. For instance most specs has the preperation rule last since the impact is pretty minor.

CollCrom:
Good idea. I'll fix it tomorrow

PR #1423, opened 2018-02-22
[Destruction] Eradication

lpoorthuis:
Currently spells benefitting from Eradication are only added to the statistics if the debuff Eradication is applied to the target. But the dmg bonus of Eradication is added at the end of a cast.
Just calculating the statistics on _cast and not on _damage will fix this.


lpoorthuis:
Alright i already thought this was a bit too easy. Is someone already working on this? I will invest more time otherwise.

sMteX:
Not that I'm aware of, feel free to look at it, then open another PR, I'll close this one for now.

PR #1427, opened 2018-02-23
[Mistweaver] Drape of Shame effective ilvl

CollCrom:
image
[test log](https://www.warcraftlogs.com/reports/DLRvBJX4x1Ft9pnh#fight=12&type=healing&source=30)

Added in a display to show the effective itemlevel of Drape of Shame for mistweaver monks.


MartijnHols:
Thanks! I was thinking of making this a part of the base module, with some sort of "optional dependency", but until then this is good. @kfinch how about doing this for Resto Dudu too?

CollCrom:
Yeah I was considering that as well I'll look into it today.

PR #1428, opened 2018-02-23
[BrM] Fix Stagger Pool Plot Hitboxes

emallson:
This is a super simple PR. tl;dr the plot as it is now on the site is extremely difficult to mouseover the data points (aside from Purifies) since the point radius for them is set to 0. This sets the hit/hover radius to 4 so that you can actually see the exact values by mousing over the lines.

There is nothing I can really screenshot about this change. On live you *can* get the tooltip, its just a pain in the ass.


MartijnHols:
Thanks! 🍺

PR #1432, opened 2018-02-24
Added Armor to stat tracker

emallson:
Fairly self-explanatory. Armor gets tracked. The only buffs currently represented are from DMD:I. This is going to be more important as we ramp up for BfA as a bunch of tank AM is getting switched to bonus armor.

I worked around a bug (?) in the `combatantinfo` armor value. It doesn't include the bonuses from artifacts or passives. For BrM this nearly doubles the amount of armor I have pre-pull. However, while these passives do multiply together they do not multiply other temporary buffs and so cannot be used with the overall stattracker infrastructure to update stats.

I was waffling over whether to do this in `StatTracker` or adding a normalizer for a bit and ended up just doing it here. If it'd be better in a normalizer it's not much effort for me to move it before the PR is accepted.

### DIFF
--- a/src/Parser/Core/Modules/StatTracker.js
+++ b/src/Parser/Core/Modules/StatTracker.js
@@ -239,7 +239,7 @@ class StatTracker extends Analyzer {
   }
 
   applySpecModifiers() {
-    const modifiers = this.constructor.SPEC_MULTIPLIERS[this.combatants.selected.spec.id];
+    const modifiers = this.constructor.SPEC_MULTIPLIERS[this.combatants.selected.spec.id] || {};
     Object.entries(modifiers).forEach(([stat, multiplier]) => this._pullStats[stat] *= multiplier);
   }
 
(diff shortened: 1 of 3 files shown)


src/Parser/Core/Modules/StatTracker.js
```diff
@@ -203,14 +225,39 @@ class StatTracker extends Analyzer {
...
 
+  applySpecModifiers() {
+    const modifiers = this.constructor.SPEC_MULTIPLIERS[this.combatants.selected.spec.id];
+    this.applyMultipliers(this._pullStats, modifiers);
+  }
+
+  applyArtifactModifiers() {
+    Object.entries(this.constructor.ARTIFACT_MULTIPLIERS).forEach(([spellId, modifiers]) => {
+      if(this.combatants.selected.traitsBySpellId[spellId] > 0) {
+        this.applyMultipliers(this._pullStats, modifiers);
```
MartijnHols: Shouldn’t this multiply the value by the number of traits?

emallson: All the ones I've seen are 0 or 1 but thats a reasonable adjustment.

PR #1435, opened 2018-02-26
[Fire Mage] Support for Koralon's Burning Touch

Sharrq:
Added a suggestion for casting Fireball instead of Scorch when the target is under 30% health when using Koralon's Burning Touch. Also adjusted Heating Up to take Scorch guaranteed crits into account when using Koralons.


Sharrq:
Intentionally left that out because damage effectiveness isn't accurate without ignite handling which is borderline impossible because of the way it spreads.

Sharrq:
I've been meaning to remove the damage numbers for the ones that are already live as well because of this.

PR #1440, opened 2018-03-01
VP Effiency, Apocalypse change to suggestiontresh and some rogue Console.logs

Bicepspump:
Removed some console.logs in Coldheart. Reworked the suggestiontreshold in Apocalypse so that they work correctly. Added a new module that tracks the reapplication of Virulent Plague.

### DIFF
--- a/src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueEfficiency.js
+++ b/src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueEfficiency.js
@@ -20,54 +20,57 @@ class VirulentPlagueEfficiency extends Analyzer {
   targets = {};
   
   totalOutBreakCasts = 0;
-  totalVirulentDuration = 0;
   totalTimeWasted = 0;
   
   get VirulentDuration(){
-	if (this.combatants.selected.hasTalent(SPELLS.EBON_FEVER_TALENT.id)) {
-		return(19.65);
-	} else{
-		return(33.3);
-	}
+	  if (this.combatants.selected.hasTalent(SPELLS.EBON_FEVER_TALENT.id)) {
+		  return(13.65);
+	  } else{
+		  return(27.3);
+	  }
   }
   
   on_byPlayer_refreshdebuff(event){
     const spellId = event.ability.guid;
     if(spellId === SPELLS.VIRULENT_PLAGUE.id){
-		this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp;
-	}
+		  this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp + 1000*this.VirulentDuration;
+	  }
   }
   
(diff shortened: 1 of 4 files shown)


src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueEfficiency.js
```diff
@@ -0,0 +1,102 @@
...
+		return(19.65);
+	} else{
+		return(33.3);
+	}
+  }
+  
+  on_byPlayer_refreshdebuff(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.VIRULENT_PLAGUE.id){
+		this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp;
```
Gebuz: I think you might have mixed up `refreshdebuff` and `applydebuff`, the pandemic duration should be added here in `refreshdebuff`. I would suggest that you save a timestamp for when you expect the dot to expire instead of when it was last cast, that way you can also account for pandemic effects properly: 
```
const remainingDuration = this.targets[encodeTargetString(event.targetID, event.targetInstance)] - event.timestamp;
if (remainingDuration < 1000*0.3*this.VirulentDuration) { //less than full pandemic
  this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp + this.VirulentDuration + remainingDuration;
} else { //full pandemic
  this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp + this.VirulentDuration * 1.3;
}
```

and then in `applydebuff` you just do `this.targets[encodeTargetString(event.targetID, event.targetInstance)] = event.timestamp + this.VirulentDuration;`

src/Parser/DeathKnight/Unholy/Modules/Features/VirulentPlagueEfficiency.js
```diff
@@ -0,0 +1,102 @@
...
+  
+
+  
+  on_byPlayer_cast(event) {
+	const spellID = event.ability.guid;
+	if(spellID === SPELLS.OUTBREAK.id){	
+		this.totalOutBreakCasts += 1;
+		if (this.targets[encodeTargetString(event.targetID, event.targetInstance)]) {
+			//We subtract 6 seconds from the total duration since this is the time left after Outbreak finishes.
+			if (((this.VirulentDuration - 6) - (event.timestamp -  this.targets[encodeTargetString(event.targetID, event.targetInstance)]) / 1000) >= 0) {
```
Gebuz: If you do what I suggested above and save the timestamp for when you expect the dot to expire instead, you can simply do this here:
```
if (this.targets[encodeTargetString(event.targetID, event.targetInstance)] > event.timestamp) {
  this.totalTimeWasted += this.targets[encodeTargetString(event.targetID, event.targetInstance)] - event.timestamp;
}
```

PR #1441, opened 2018-03-02
[Havoc] Blind Fury

CollCrom:
[Blind Fury](http://www.wowhead.com/spell=203550/blind-fury) does not have an energize event so I added a process fake events for it. Also increased the tolerance since it usually wastes 300 over the fight but it is more important to keep eye beam on cooldown.

image
[Test Log](https://www.warcraftlogs.com/reports/BHyhvwRqYjrMVfk8#fight=18&type=damage-done&source=1)


Gebuz:
May I suggest having a different threshold specifically for when the Blind Fury talent is used?

PR #1442, opened 2018-03-02
[Assassination] Zoldyck Family Training Shackles

CollCrom:
Added support for the [Bracers](http://www.wowhead.com/item=137098/zoldyck-family-training-shackles)

image

[Test Log](https://www.warcraftlogs.com/reports/BJYkxXH2mrpNtK7n#fight=21&type=damage-done&source=14)

Also Updated the T21 Icon from the DH Eye Beam Icon to the buff icon to make more sense.


src/Parser/Rogue/Assassination/Modules/Items/ZoldyckFamilyTrainingShackles.js
```diff
@@ -0,0 +1,76 @@
...
+    this.active = this.combatants.selected.hasWrists(ITEMS.ZOLDYCK_FAMILY_TRAINING_SHACKLES.id);
+  }
+
+  on_byPlayer_damage(event) {
+  	const spellId = event.ability.guid;
+  	if(!POISONS_AND_BLEEDS.includes(spellId)) {
+  		return;
+  	}
+  	this.currentHealth = event.hitPoints;
+  	this.maxHealth = event.maxHitPoints;
```
Gebuz: It would be prettier if you either passed these numbers as parameters to a healthPercent function or did the calculation here instead (it's pretty simple so doesn't really need it's own function). No need to save them as class variables.

CollCrom: I just pulled from the fire mage leggo belt. Wasn't sure what the convention should be. I'll fix it

PR #1448, opened 2018-03-05
[Windwalker] Updating some suggestions and TWB item breakdown

Juko8:
Added a check on "The Wind Blows" so it now only attributes a free Blackout Kick if Strike of the Windlord was used without the buff already active.

### DIFF
--- a/src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
+++ b/src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
@@ -48,10 +48,11 @@ class FistsofFury extends Analyzer {
 
     suggestions(when) {
       const averageTicksRecommended = this.combatants.selected.hasBuff(SPELLS.WW_TIER20_4PC.id) ? 4.5 : 5;
+      const tier20Text = this.combatants.selected.hasBuff(SPELLS.WW_TIER20_4PC.id) ? ". This is not always true while using T20 4pc" : "";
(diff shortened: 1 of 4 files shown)


src/Parser/Monk/Windwalker/Modules/Spells/FistsofFury.js
```diff
@@ -48,10 +48,11 @@ class FistsofFury extends Analyzer {
 
     suggestions(when) {
       const averageTicksRecommended = this.combatants.selected.hasBuff(SPELLS.WW_TIER20_4PC.id) ? 4.5 : 5;
+      const tier20Text = this.combatants.selected.hasBuff(SPELLS.WW_TIER20_4PC.id) ? ". This is not always true while using T20 4pc" : "";
       when(this.averageTicks).isLessThan(averageTicksRecommended).addSuggestion((suggest, actual, recommended) => {
             return suggest(<span> You are cancelling your <SpellLink id={SPELLS.FISTS_OF_FURY_CAST.id} /> casts early and losing ticks </span>)
                 .icon(SPELLS.FISTS_OF_FURY_CAST.icon).actual(`${this.averageTicks.toFixed(2)} average ticks on each Fists of Fury cast`)
-                .recommended(`Aim to get 5 ticks with each Fists of Fury cast (not always true with T20)`)
+                .recommended(`Aim to get 5 ticks with each Fists of Fury cast`+tier20Text)
```
emallson: This should either be a normal string using concatenation or a template string using `${tier20Text}` substitution.

PR #1450, opened 2018-03-05
[Boomkin & BDK] Bug fixes

Gebuz:
- AsP suggestion was using an inverted threshold and wrong var for percent in tooltip.
- L90 talents was including BoE value even when BoA was active
- fixed ms -> s in 2p21 blood tooltip
- fix BDK bracers % of max RP (was broken by RP module update)
- fix BDK shoulders on vari (no heal event when absorbed) and wasted stacks.

### DIFF
--- a/src/Parser/DeathKnight/Blood/Modules/Features/CrimsonScourge.js
+++ b/src/Parser/DeathKnight/Blood/Modules/Features/CrimsonScourge.js
@@ -3,10 +3,14 @@ import Analyzer from 'Parser/Core/Analyzer';
 import SPELLS from 'common/SPELLS';
 import SpellIcon from 'common/SpellIcon';
 import { formatPercentage } from 'common/format';
+import SpellLink from 'common/SpellLink';
+import Wrapper from 'common/Wrapper';
 import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
 import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
 import Combatants from 'Parser/Core/Modules/Combatants';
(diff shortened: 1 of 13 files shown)


Gebuz:
- fixed RP suggestion threshold using < instead of > and made it more strict
- removed tooltips that didn't add extra info.
- removed Crimson Scourge suggestion (should be covered by DnD cast efficiency).
- fixed styling a few places.
- Made DnD suggestionThreshold less strict to account for reaction time on reset procs.

src/Parser/DeathKnight/Blood/Modules/Features/CrimsonScourge.js
```diff
@@ -50,34 +50,12 @@ class CrimsonScourge extends Analyzer {
...
-        average: 0.1,
-        major: 0.15,
-      },
-      style: 'percentage',
-      text: 'You had unspent Crimson Scourge procs. Make sure you always use them.',
-    };
-  }
-
-  suggestions(when) {
-    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
```
CollCrom: Why remove this suggestion?

Gebuz: Don't need it now that DnD has cast efficiency. Although saying that I should probably enable it if Rapid Decomposition talent isn't used and disable the DnD efficiency in that case.

Gebuz:
Added Crimson Scourge suggestion back when not using rapid decomposition, and disabled DnD efficiency when not using rapid decomposition. Also added a buffer to not include unspent procs when the boss is about to die.

CollCrom:
looks good! thanks =]

PR #1451, opened 2018-03-05
[Enhancement/Elemental] Eye of the Twisting Nether

CollCrom:
image
Added [Eye of the Twisting Nether](http://www.wowhead.com/item=137050/eye-of-the-twisting-nether)

[Enhancement](https://www.warcraftlogs.com/reports/yPqv481A7cYdzXBa#fight=25&type=damage-done&source=193)

[Elemental](https://www.warcraftlogs.com/reports/7Zp4GLmFW9NRrzck#fight=21&type=damage-done&source=12)

### DIFF
--- a/src/Parser/Shaman/Elemental/CHANGELOG.js
+++ b/src/Parser/Shaman/Elemental/CHANGELOG.js
@@ -1,4 +1,5 @@
 export default `
+03-05-2018 - Added Eye of the Twisting Nether
 03-02-2018 - Fixed Fire Elemental Precast not counting towards Casting Efficiency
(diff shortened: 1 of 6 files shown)


Gebuz:
This probably has a pretty minor impact on the final number since the percentages are so small but technically you should sum the modifiers before calculating the effective damage they provided. So either multiplicative: `(1 + MODIFIER)^stacks - 1` or additive: `MODIFIER*stacks`. I would assume they are multiplicative since it is 3 different buffs, but might be worth checking if you haven't already.

PR #1457, opened 2018-03-08
[Havoc] Artifact Traits

CollCrom:
image

[Test Log](https://www.warcraftlogs.com/reports/BHyhvwRqYjrMVfk8#fight=13&type=damage-done&source=1)

Didn't account for the traits that did not have a direct damage increase since I am currently unsure how I would calculate that.

also changed the ordering on the ret paladin stat boxes to make it look nicer.

New:
image

Old: 
image

### DIFF
--- a/src/Parser/Core/calculateEffectiveDamageStacked.js
+++ b/src/Parser/Core/calculateEffectiveDamageStacked.js
@@ -1,0 +1,8 @@
+export default function getDamageBonusStacked(event, increase, stacks) {
+	const raw = (event.amount || 0) + (event.absorbed || 0);
+	const relativeDamageIncreaseFactor = 1 + (increase * stacks);
+	const totalIncrease = raw - raw / relativeDamageIncreaseFactor;
+	const oneStackIncrease = totalIncrease / stacks;
+
+	return Math.max(0, oneStackIncrease);
+}
--- a/src/Parser/Core/calculateEffectiveDamageStacked.js
+++ b/src/Parser/Core/calculateEffectiveDamageStacked.js
@@ -1,4 +1,4 @@
-export default function getDamageBonusStacked(event, increase, stacks) {
+export default function calculateEffectiveDamageStacked(event, increase, stacks) {
 	const raw = (event.amount || 0) + (event.absorbed || 0);
 	const relativeDamageIncreaseFactor = 1 + (increase * stacks);
 	const totalIncrease = raw - raw / relativeDamageIncreaseFactor;
--- a/src/Parser/DemonHunter/Havoc/Modules/Traits/CriticalChaos.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Traits/CriticalChaos.js
@@ -34,7 +34,7 @@ class CrticalChaos extends Analyzer {
 
 	rank = 0;
 	damage = 0;
-  lastCritTimestamp = 0;
+  chaosStrikeCrit = false;
 
 	on_initialized() {
  		this.rank = this.combatants.selected.traitsBySpellId[SPELLS.CRITICAL_CHAOS.id];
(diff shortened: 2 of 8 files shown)


Pastaaa:
I'm not realy familiar with programming but i think your chaosstrike calculation is wrong. 

AT your test log Annihilation Main and OH Crit did 132,14m Damage, 
annihilation
Chaos strike did 61,55M
chaos strike
, so the total Crit damage is 193,69M, which is rough 50% of the total damage the player did. 

If i understood it right, you are calculating only the last trait, so the 193,69M Crit Damage is 106% of the Base value (6% More Crit strike damage per trait), so the last trait did:  193,69/100*6 = 10,96 Million Damage. 

431,47M Damage did The Player on this fight, so 10,96M from the Trait is a 2,54% Damage increase.

That Calculation was without the Talent "Chaos Cleave" http://www.wowhead.com/spell=206475/chaos-cleave but since its Damage depends on the Damage Chaosstrike did, the "Critical chaos" trait is even stronger.

Hope i could help a little bit :)

Gebuz:
The traits are additive, not multiplicative so you do have to take the other traits into account, you can't just do /1.06. Saying that, the current calculation gets the average value of all your traits. If you want the value of the last point specifically it would be:
```
LastTraitDMGIncrease = DMG -  DMG * (1 + (traitsCount - 1) * modifier) / (1 + traitsCount * modifier)
Example:
LastTraitDMGIncrease = 193.69M - 193.69M * (1 + (5 - 1) * 0.03) / (1 + 5 - * 0.06) = ~5.1M
```
Granted, the percentages are so small the difference is pretty minor between the last point and the average. Also, I remember Critical chaos calculation being changed a while back (or maybe it was just the tooltip value), so make sure to check if the damage bonus applies to all the damage (my calculation above assumes that) or if it only applies to the crit bonus damage which would cut the actual value in half. If you haven't already, that is!

src/Parser/DemonHunter/Havoc/Modules/Traits/CriticalChaos.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  on_byPlayer_damage(event) {
+  	const spellId = event.ability.guid;
+  	if (!CRITICAL_CHAOS_SPELLS.includes(spellId) || event.hitType !== HIT_TYPES.CRIT) {
+  		return;
+  	}
+
+  	const raw = (event.amount || 0) + (event.absorbed || 0);
+  	const rawNormalPart = raw / this.critEffectBonus.getBonus(event);
```
Gebuz: Right now you're getting the non crit damage (which matches the var name), I assume you'd want only the crit bonus damage instead? `critBonusDamage = raw - raw / this.critEffectBonus.getBonus(event);`

src/Parser/Core/calculateEffectiveDamageStacked.js
```diff
@@ -0,0 +1,8 @@
+export default function getDamageBonusStacked(event, increase, stacks) {
```
Gebuz: not used! :D

CollCrom:
I think they condense down the same formula no?
` (DMG * Modifier)/(Modifier * Stacks + 1)`

Gebuz:
You're right, that's way shorter, I didn't even realize it could be shortened down to that. I shouldn't do math when I'm tired. Do you want to add chaos cleave? Otherwise I think it's good to merge.

CollCrom:
I suck at math thats why I had wolfram do it for me.

I'm thinking about how to add chaos cleave in right now. If I get the last crit timestamp would the extra damage contributed from chaos cleave be the same?
`damage * mod / (mod * stacks +1)`

Gebuz:
I think you can just save a bool for whether the last Chaos Strike was a crit or not, and then only add those that are. I think just doing the same calc you do for the chaos strike crit dmg should be correct.

src/Parser/DemonHunter/Havoc/Modules/Traits/CriticalChaos.js
```diff
@@ -0,0 +1,96 @@
...
+    return rawCritDamage;
+  }
+
+  on_byPlayer_damage(event) {
+  	const spellId = event.ability.guid;
+    //Chaos Cleave Attribution
+    if(this.combatants.selected.hasTalent(SPELLS.CHAOS_CLEAVE_TALENT.id) && 
+      event.timestamp < this.lastCritTimestamp + MS_BUFFER && 
+      spellId === SPELLS.CHAOS_CLEAVE_DAMAGE.id) 
+    {
```
CollCrom: This is a bit ugly but I am not sure how to make it look nicer

Gebuz: I would increase the buffer (sometimes there is up to 100ms between CS and chaos cleave dmg), or just replace the timestamp with a bool that you change based on the last Chaos Strike being a crit or not. Otherwise it looks fine.

CollCrom: I would still need some sort of window if i used a bool right? To account for the cleave damage on AoE? Unless I am thinking about this wrong

Gebuz: Shouldn't be necessary.  There should always be a chaos strike damage event before a chaos cleave event you can link it too. Even if the log is bugged and there is chaos cleave events without a chaos strike just before (say eonar range issues), I think it's more correct to just guess based on the most recent chaos strike being a crit or not, rather than just discard it.

CollCrom:
Whoops I'll fix the lint error later today when I have time l

PR #1461, opened 2018-03-10
[Blood DK] added DS timing graph

joshinat0r:
Adds a plot to the BDK analysis thats showing Death Strike casts and the HP.
Goal is to help spotting low HP situations where the BDK relied on external healing instead of DS-selfhealing.

ds_timing

### DIFF
--- a/src/Parser/DeathKnight/Blood/CombatLogParser.js
+++ b/src/Parser/DeathKnight/Blood/CombatLogParser.js
@@ -16,6 +16,7 @@ import CrimsonScourge from './Modules/Features/CrimsonScourge';
 import BlooddrinkerTicks from './Modules/Features/BlooddrinkerTicks';
 import UnendingThirstTracker from './Modules/Features/UnendingThirstTracker';
 import Checklist from './Modules/Features/Checklist';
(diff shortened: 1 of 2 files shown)


Gebuz:
Thanks for the PR! Would it be possible to get the DS dots match the height of current health like the purify ones do for monk? Makes it easier to see at what HP it was used. Also, since DK max HP fluctuates so much, would it make sense to show the health as percent of max health instead? If you think raw numbers are better I would suggest that you make the Max HP curve more obvious, it's a bit hard to see at the moment.

joshinat0r:
Updated the PR, showing the graph based on percent is actually better.
Displaying the DS dots at about 25% height of the graph was on purpose since the tooltip has more infos about the DS cast (how much it healed and overhealed for and at what HP), but hovering over the DS dot would show me the HP-graphs tooltip instead. Changed the tooltip now, so that shouldn't be an issue anymore.

ds_timing2

Gebuz:
Oh I forgot to mention, remember to add a changelog entry in the future!

joshinat0r:
Should have read the guideline till the end :)
I'll add a changelog for this PR in my next PR.

PR #1463, opened 2018-03-11
Remove completeness, add patch compatibility status

MartijnHols:
In an effort to make it easier to understand a spec's analysis implementation status, the completeness rating was removed. Instead we now show the config's description on the results page which can have a custom explanation of what the spec does and doesn't have. Hopefully this will allow everyone to get their spec to a point that they're comfortable without discrediting their work when it's not entirely complete.

The goal is to allow specs to grow incrementally without any pressure. We do still want to inform our users if the analysis is complete or good enough, and thus you have full freedom to explain the status of the spec in the description.

Accuracy is no longer a part of completeness/rating, as everything that is active should be accurate given it was released for the current patch. Patch incompatibility is handled through the new patch compatibility settings, which gives an indicator about accuracy if something becomes outdated.

It will be our responsibility to disable modules that are/become inaccurate with the goal of only ever having things that are accurate.

image
image
image

Going to move the item bonuses down at a later time. PRing already to somewhat keep the changes limited.

Also disables timeline features if they're inaccurate:
image
The idea is that core features will be disabled / downgraded if they'd otherwise be inaccurate. The goal is to have full accuracy.

Fixes #1339


Sharrq:
I like it, but I would like you also incorporate the change I made in my pr to remove the blocks of text in the section dividers (the source, completeness, maintainers, etc) if that isn't already done here.

I also kinda of think the items panel should be moved at the same time but I also agree that keeping the changes smaller can be beneficial

MartijnHols:
I did that

Sharrq:
Cool. Then i think its good.

MartijnHols:
Sweet, thanks. Since I won't be home all of tonight I think this a great time to merge this. Here we go!

PR #1464, opened 2018-03-11
[BDK] Marrowrend-Module

joshinat0r:
- fixed an issue with the DeathstrikeTimingGraph (death event wasn't based on %-HP)
- added a RP counter for Ossuary (showing a more usefull metric than uptime)
- added a MR module for tracking bad MR casts ("bad MR cast" = cast at 7 or more stacks of BS (ignoring casts that were made to refresh its duration))

os
mr


Gebuz:
I noticed that the time in the Death Strike chart is doubled, I assume it's cause you have a point every 500ms instead of 1000ms? I'll take a look at this PR tomorrow!

joshinat0r:
Should have noticed that earlier, is pretty obvious 🤔 
Reverted the 500ms interval, initially went with that because the DS-dots were on a fixed position. Changing the interval to something smaller helped spotting the DS impact on the HP-bar, but that shouldn't be necessary anymore.

src/Parser/DeathKnight/Blood/Modules/Features/MarrowrendUsage.js
```diff
@@ -0,0 +1,104 @@
...
+
+  on_toPlayer_removebuffstack(event) {
+    if (event.ability.guid === SPELLS.BONE_SHIELD.id){
+      this.currentBoneShieldBuffer = 0;
+      this.currentBoneShieldStacks -= 1;
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid === SPELLS.MARROWREND.id) {
```
Gebuz: It's preferred that you avoid nested if statements when possible. You can instead check for the opposite and return early.

src/Parser/DeathKnight/Blood/Modules/Talents/Ossuary.js
```diff
@@ -19,6 +23,17 @@ class Ossuary extends Analyzer {
     return this.combatants.getBuffUptime(SPELLS.OSSUARY.id) / this.owner.fightDuration;
   }
 
+  on_byPlayer_cast(event) {
+
+    if (event.ability.guid === SPELLS.DEATH_STRIKE.id) {
```
Gebuz: It's preferred that you avoid nested if statements when possible. You can instead check for the opposite and return early.

PR #1465, opened 2018-03-11
[Death Knight] Fix Rapid Decomposition sometimes using wrong spell ID in logs for RP

Gebuz:


Gebuz:
Caused by 2p19, which doesn't have anything to do with Death and Decay or Rapid Decomposition.

PR #1467, opened 2018-03-12
Added new module tracking dark transformation and wound count

Bicepspump:
Staying at 0 wounds with Dark Transformation will make you miss out on wound pops from pet. This module tracks the total amount of time you spent at 0 wounds with dark transformation active.


src/Parser/DeathKnight/Unholy/Modules/Features/DarkTransformationAndWounds.js
```diff
@@ -0,0 +1,125 @@
...
+
+  on_byPlayer_removedebuff(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FESTERING_WOUND.id){
+		  this.targets[encodeTargetString(event.targetID, event.targetInstance)] = 0;
+	  }
+  }
+
+  on_byPlayer_cast(event){
+    if(this.targets.hasOwnProperty(encodeTargetString(event.targetID, event.targetInstance))) {
```
Gebuz: you can have the null check inside the same if statement as the check for 0 stacks below with a logical OR (aslong as the null check is first), since if it's null it means you have 0 stacks on the target aswell.

src/Parser/DeathKnight/Unholy/Modules/Features/DarkTransformationAndWounds.js
```diff
@@ -0,0 +1,125 @@
...
+class DarkTransformationAndWounds extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+  };
+  // used to track how many stacks a target has
+  targets = {};
+
+  darkTransformationActive = 0;
+  darkTransformationCasts = 0;
+  stopWatchActive = 0;
```
Gebuz: you should probably use `true` and `false` instead of `1` and `0` to make it more obvious that `darkTransformationActive` and `stopWatchActive` they are bools.

src/Parser/DeathKnight/Unholy/Modules/Features/DarkTransformationAndWounds.js
```diff
@@ -0,0 +1,125 @@
...
+  on_byPlayer_cast(event){
+    if(this.targets.hasOwnProperty(encodeTargetString(event.targetID, event.targetInstance))) {
+	    if (this.targets[encodeTargetString(event.targetID, event.targetInstance)] === 0) {
+		    if (this.darkTransformationActive === 1) {
+			    if (this.stopWatchActive === 0) {
+				    this.stopWatch = event.timestamp;
+				    this.stopWatchActive = 1;
+			    }
+		    } 
+		    else if (this.stopWatchActive === 1) {
```
Gebuz: You can avoid some of the duplicate code and the nested if statements if you do this check first thing in `on_byPlayer_cast`, assuming you always wants to add the time if the stopwatch is active. Then you can stop/start it depending on whether the buff or debuffs are up afterwards.

Bicepspump: I'm not too sure I can do that. I only add the time when the conditions no longer hold, so only when the stopwatch turns from active to inactive.

Gebuz: I was thinking you could just start it again afterwards. Anyway I guess this works, it's only two extra lines after all.

src/Parser/DeathKnight/Unholy/Modules/Features/DarkTransformationAndWounds.js
```diff
@@ -0,0 +1,123 @@
...
+
+  on_byPlayer_removedebuff(event){
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FESTERING_WOUND.id){
+		  this.targets[encodeTargetString(event.targetID, event.targetInstance)] = 0;
+	  }
+  }
+
+  on_byPlayer_cast(event){
+    if(this.targets.hasOwnProperty(encodeTargetString(event.targetID, event.targetInstance)) && this.targets[encodeTargetString(event.targetID, event.targetInstance)] === 0) {
```
Gebuz: This wont catch the case where your target has never had any stacks and thus aint in the array. If you instead invert the first check and use `||` instead of `&&` it will do what you want:
not in array -> first part of the condition evaluates to true and executes the body without evaluating the second part of the if condition.
in array -> first part of the condition evaluates to false and it continues to second part and executes the body if stacks === 0.
```
if(!this.targets.hasOwnProperty(encodeTargetString(event.targetID, event.targetInstance)) 
|| this.targets[encodeTargetString(event.targetID, event.targetInstance)] === 0)
```

src/Parser/DeathKnight/Unholy/Modules/Features/DarkTransformationAndWounds.js
```diff
@@ -0,0 +1,123 @@
...
+		  if (this.darkTransformationActive === true) {
```
Gebuz: no need to do `=== true`, `if (this.darkTransformationActive)` is the same thing. Same with `=== false` below, you can do `if (!this.stopWatchActive)`. `!` inverts the boolean value.

PR #1468, opened 2018-03-12
[Elemental] Only warn about elemental focus for >1 Flames Shock casts

mpolednik:
Instead of warning when a single flame shock was cast without elemental
focus, we suggest no more than 1 flame shocks are cast without it.
Reason for the change is elemental shaman's opener, as it is possible
to cast

prepot/fire elemental -> precast EB -> FS -> LB and on

leading to unbuffed initial (short) flame shock.

### DIFF
--- a/src/Parser/Shaman/Elemental/Modules/ShamanCore/FlameShock.js
+++ b/src/Parser/Shaman/Elemental/Modules/ShamanCore/FlameShock.js
@@ -50,7 +50,7 @@ class FlameShock extends Analyzer {
 
   suggestions(when) {
     const flameShockCasts = this.elementalFocusCasts;
-    when(flameShockCasts.unbuffed).isGreaterThan(0)
+    when(flameShockCasts.unbuffed).isGreaterThan(1)
       .addSuggestion((suggest, actual, recommended) => {
         return suggest(<span>Try to cast <SpellLink id={SPELLS.FLAME_SHOCK.id}/> when you have <SpellLink id={SPELLS.ELEMENTAL_FOCUS.id}/>.</span>)
           .icon(SPELLS.ELEMENTAL_FOCUS.icon)


Pewtro:
Instead of this PR, wouldn't it be better to fix the underlying problem in the ElementalFocus class, which is that it doesn't use ``this.combatants.selected.hasBuff()`` but instead switches a bool between true and false.

mpolednik:
This isn't about the buff not being correctly tracked, rather about the fact that Flame Shock cast in the opener should (or could, because with absolutely perfect timing and RNG the EB might arrive+crit right in time) be cast *without* the buff. Therefore the analyzer should permit 1 FS (perfect case) cast without it. Checking my logs with perfect opener/flame shock casts, the tracking seems to work correctly.

I don't think I have the code/js knowledge (at the moment) to fix the buff tracking infra.

HawkCorrigan:
Switching the EF module to use hasBuff wouldn't solve the issue that is solved here. Both are completely unrelated. 
EF module prolly needs an overhaul too but afaik they use the weird boolean version to keep track of how many stacks you lost. 

Only thing I would maybe disagree with is that if you do a "wrong" opener it would still not complain if you would have had the option to cast it with EF.

Let's say you play without eb. 
That means your precast is a lb with close to zero travel time and you have the possibility to get a EF stack in the first global and it wouldn't track another "missed" empowered flameshock.

So maybe just ignore all flamesshocks that were cast before you got an EF stack?

Idk I would be fine with the change tbh. It is definitely "nicer" than it is right now, but it might then ignore something that would be improvable.

mpolednik:
> So maybe just ignore all flameshocks that were cast before you got an EF stack?

Agree, that sounds somewhat robust.

MartijnHols:
@mpolednik Thanks for your contribution. Unfortunately as discussed above we think this requires a different solution before we would be able to merge this into our project. The most important thing is to ensure it stays accurate and not misleading.

Because this PR has had no activity for a month I will close it now. If you wish to take a stab at it you can still post a comment in this PR or reopen it, or create a new PR when you're ready.

Thank you for your time.

PR #1476, opened 2018-03-14
[Restoration Shaman] GCD fix for totems

niseko:
Totems are not configured correctly, throwing the timeline off, this changes them to have a 1 second static GCD like they have in game.
image

Earthbind is technically 1 second minus haste, but no idea how to handle that.


Gebuz:
If you want to reduce the GCD of earthbind with haste you can import GlobalCooldown and override `getCurrentGlobalCooldown` to reduce it like here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Hunter/BeastMastery/Modules/Core/GlobalCooldown.js
```
getCurrentGlobalCooldown(spellId = null) {
  if (spellId === SPELLS.EARTHBIND_TOTEM.id){
    const gcd = TOTEM_GCD / (1 + this.haste.current);
    return Math.max(gcd, this.owner.modules.alwaysBeCasting.constructor.MINIMUM_GCD); //assuming that the min GCD is 750ms for earthbind, I don't know if it is.
  } else {
    return super.getCurrentGlobalCooldown(spellId);
  }
}
```

PR #1478, opened 2018-03-14
[BDK] added Souldrinker

joshinat0r:
Tracking for the Souldrinker-Trait
(tracking the % is done manually since the combatlog event doesn't contain the exact values)

sd


Gebuz:
Looks good, is the tooltip necessary though? It doesn't really add any extra information. 
I assume adding a suggestion isn't needed here since effective healing is more important than Soul Drinker?
Remember changelog!

joshinat0r:
Not quite sure if it's necessary, some people could think that this stat is to show that they could get higher values.
Just want to avoid confusions, but I'll remove it as it doesn't really contain information

PR #1481, opened 2018-03-14
[BDK] added BDK DPS traits

joshinat0r:
relics


src/Parser/DeathKnight/Shared/getDamageBonusStacked.js
```diff
@@ -0,0 +1,8 @@
+export default function getDamageBonusStacked(event, increase, stacks) {
```
Gebuz: You should import this from core instead of duplicating it here just for dks:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/calculateEffectiveDamageStacked.js

PR #1485, opened 2018-03-15
[Restoration Shaman] Timeline and Checklist changes

niseko:
image
I added talents to the checklist in the respective sections as they're just as important as core spells, hid the chain heal module if there are no casts as the module just showing N/A always annoyed me, and there are lots of logs without any chain heal cast.
(image is before change)
image

EchoReset "handles" riptide getting random resets without combat log events by just cutting the CD short if you casted it without being able to. Knowing the spell efficiency of riptide is too important and resets are too rare to just disable the spell. Cast efficiency is a lot harsher if echo is selected to account for the resets. Lava Surge is simpler as you have a buff event for that.

Non healing time was broken since it was using a variable that didn't exist, i just removed it so it falls back on the standard function.


src/Parser/Shaman/Restoration/Modules/Features/AlwaysBeCasting.js
```diff
@@ -59,18 +59,6 @@ class AlwaysBeCasting extends CoreAlwaysBeCastingHealing {
...
-      actual: nonHealingTimePercentage,
-      isGreaterThan: {
-        minor: 0.3,
-        average: 0.4,
-        major: 0.45,
-      },
-      style: 'percentage',
-    };
-  }
   get downtimeSuggestionThresholds() {
```
Gebuz: You can remove the downtimeSuggestionThreshold and the suggestions aswell since it is all inherited anyway.

src/Parser/Shaman/Restoration/Modules/Talents/EchoReset.js
```diff
@@ -0,0 +1,44 @@
...
+    this.active = this.combatants.selected.hasTalent(SPELLS.ECHO_OF_THE_ELEMENTS_TALENT.id) || this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_FARSEER.id);
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.RIPTIDE.id) {
+      return;
+    }
+    if (this.spellUsable.chargesAvailable(SPELLS.RIPTIDE.id) === 0) {
+        if(event.timestamp - this.lastRiptideTimestamp <= 6000 && this.lastRiptideCharges <= 1) {
```
Gebuz: Isn't this if condition always true if the above if statement is true? If you have 0 charges it means it must be < CooldownDuration since last cast and last cast must have had 1 charge?

src/Parser/Shaman/Restoration/Modules/Spells/LavaSurge.js
```diff
@@ -0,0 +1,22 @@
...
+    spellUsable: SpellUsable,
+  };
+
+  on_applybuff(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.LAVA_SURGE.id) {
+      return;
+    }
+    if (this.spellUsable.isOnCooldown(SPELLS.LAVA_BURST.id)) {
+      this.spellUsable.reduceCooldown(SPELLS.LAVA_BURST.id, 8000);
```
Gebuz: I'd suggest you define the cooldown duration as a global constant before the class:
`const LAVA_BURST_CD_MS = 8000;`
Same with riptide CD in EchoReset.

src/Parser/Shaman/Restoration/Modules/Talents/EchoReset.js
```diff
@@ -0,0 +1,44 @@
...
+            this.spellUsable.endCooldown(spellId,false,event.timestamp,6000);
+            this.spellUsable.beginCooldown(spellId);
```
Gebuz: Any reason why you don't just use reduceCooldown instead of this and the above line like you do in LavaSurge? This will essentially reset both charges as it first ends the cooldown of the current charge, and then reduce the cooldown of the second charge by 6000 ms.

src/Parser/Shaman/Restoration/Modules/Spells/LavaSurge.js
Gebuz: Actually, you could also get the expected CD from abilities instead, that way you'd avoid the cooldown being hardcoded in two different places.

niseko:
I reread the riptide reset solution with your (@Gebuz) suggestions in mind and it indeed did not make any sense, so i scrapped it in favour of a different way. 

At first I tried to point the reset to the last Tidal Wave use (which has a chance for the reset), but what gave / gives me issues is that the reset can happen 3, 4 or more Riptides before the actual SpellUsable error (because multiple charges can delay the error a lot) - if you did not consume more tidal waves anymore after the actual reset, any solution i tried will have very weird behaviour if that happens (double CD bar on the timeline or no bar at all, or both).

What I commited now just ends the cooldown if it was casted while on cooldown, which is good enough to remove the errors at least, it just doesn't point to a potential reset.

PR #1487, opened 2018-03-15
[core] Sort timeline to match abilities.js ordering

Gebuz:
Currently it is sorted by Cooldown length and secondary is spellID, which is not ideal for utility spells with no cooldown as they will show up at the top of the timeline (Druid forms for instance).
This change will allow devs to customize the order as they see fit.

Before:
image
After:
image

### DIFF
--- a/src/Main/Timeline/SpellTimeline.js
+++ b/src/Main/Timeline/SpellTimeline.js
@@ -63,15 +63,27 @@ class SpellTimeline extends React.PureComponent {
 
   get spells() {
     const { spellId, historyBySpellId, abilities } = this.props;
-    const spellIds = spellId ? [spellId] : Object.keys(historyBySpellId).map(Number);
+    const spellIds = (spellId ? [spellId] : Object.keys(historyBySpellId).map(Number))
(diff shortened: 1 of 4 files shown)


MartijnHols:
```

Gebuz - Today at 19:04
@Zerotorescue about changing the order of the spells on the timeline, what would you like it to do? I figured just matching the order in abilities was the easiest way to let the dev decide on the ordering, but would you rather that it is sorted by category or something like that?
Zerotorescue - Today at 19:13
The abilities categorization doesn't really allow for re-ordering outside of categories. How about adding a sortIndex prop to abilities and using that instead of expected cooldown whenever it is set?
If not set it could still use the CD so that works the same
and you'd just set special abilities' sortIndex to a high value to order them at the bottom or where you want
also please don't call it sortIndex since that could suggest it's used elsewhere
timelineOrder or timelineSortOrder or timelineSortIndex might be better :sweat_smile:
Gebuz - Today at 19:15
alright, will do that
```

Gebuz:
Should now sort by timelineSortIndex and if they are the same expected cooldown.

PR #1488, opened 2018-03-15
[Balance Druid] Highlight bad LS & SW casts

Gebuz:
also rewrote unempoweredLS to avoid using onfinished.
image
image


MartijnHols:
I don't understand the tooltip. Now I don't know Balance so that probably has something to do with it, but you can probably write it in such a way that an idiot like me could understand it. How about including what is proper behavior? "You cast an unbuffed Solar Wrath while Lunar Empowerment was active. You should spend your empowerments first."

Gebuz:
Good idea, added that.

src/Parser/Druid/Balance/Modules/Features/UnempoweredSolarWrath.js
```diff
@@ -0,0 +1,57 @@
...
+        average: 1,
+        major: 2,
+      },
+      style: 'number',
+    };
+  }
+
+  suggestions(when) {
+    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
+      return suggest(<Wrapper>You casted {this.badCasts} unempowered <SpellLink id={SPELLS.SOLAR_WRATH.id} /> while you had <SpellLink id={SPELLS.LUNAR_EMP_BUFF.id} />. Try to always spent your empowerments before casting unempowered spells.</Wrapper>)
```
jjs451: Just a couple of minor grammatical things here.  Saying "casted" is technically correct but imo it is more natural to say "cast".  "You cast 5 unempowered Solar Wrath" vs "You casted 5 unempowered Solar Wrath".  Finally you could say "Try to always spend your empowerments" instead of "Try to always spent your empowerments".  Spent sounds unnatural there to me but its up to you.  I'll leave this approved and unmerged and you can decide if you want to change those.

Gebuz: cheers, I pushed the changes.

PR #1489, opened 2018-03-15
Protection Paladin: handle blocked damage, handle Grand Crusader, LotP/HotP scale off Haste

MartijnHols:


src/Parser/Core/HIT_TYPES.js
```diff
@@ -2,6 +2,10 @@ const HIT_TYPES = {
   NORMAL: 1,
   CRIT: 2,
   NOCLUEWHATTHISIS: 3, // seen at Aura of Sacrifice
+  BLOCKED_NORMAL: 4,
+  BLOCKED_CRIT: 5,
```
Gebuz: Does this mean that all the melee spec that checks for crit hits somewhere should be updated to check for both crit hits and crit blocked?

src/Parser/Paladin/Protection/Modules/Spells/Judgment.js
```diff
@@ -26,7 +27,8 @@ class Judgment extends Analyzer {
     }
 
     if (this.spellUsable.isOnCooldown(SPELLS.SHIELD_OF_THE_RIGHTEOUS.id)) {
-      const isCrit = event.hitType === HIT_TYPES.CRIT;
+      // Nope, I did not verify if blocked crits count as crits for this trait, I just assumed it. Please do test if you can and report back or fix this comment.
```
Gebuz: afaik they count as crits for all intends and purposes.

PR #1490, opened 2018-03-16
Update modules that check for crit hits to also check for crit blocks.

Gebuz:
I intentionally left out the Tiger Palm module as it uses the hits to guess whether Face Palm procced or not.


Gebuz:
I checked Chaos Strike and Flanking Strike. Ranged abilities definitely used to be blockable, I'm not sure if they are anymore but I figured I'd just be on the safe side since the MM ones check for all damage events and not just one specific spell.

PR #1493, opened 2018-03-16
[Restoration Shaman] Resurgence Module

niseko:
This is a module showing your Resurgence mana gain, broken down by spell and how much the talent Bottomless Depths did or would have contributed.

image
image


src/Parser/Shaman/Restoration/Modules/Spells/Resurgence.js
```diff
@@ -0,0 +1,145 @@
...
+
+    if (spellId !== SPELLS.RESURGENCE.id) {     
+      return this.extraMana += event.resourceChange;
+    }
+
+    this.totalResurgenceGain += event.resourceChange;
+  }
+
+  get totalMana() {
+    this.regenedMana = ((this.owner.fightDuration/1000)/5) * 44000;
```
Gebuz: Does mana regn. continue while dead? Otherwise might be worth checking (can probably use the death module? haven't looked at it).

niseko: I've not found a 100% concluding answer on that yet looking through logs - i did not find one where the person didn't gain mana while dead though. but every death that i do find they revive with what seems to be about 36-40k MP5 during their death instead of the expected 44k

Gebuz: Alright, should be fine to just assume it continues then I suppose. Argus is probably a bit weird, I dunno if you want to do something specific with that, but I'll just merge and you can add something for Argus later if you think it makes sense (and if it is even detectable?)

PR #1500, opened 2018-03-18
[Restoration Shaman] Cloudburst Totem normalizer

niseko:
Cloudburst has odd behaviour, where, if there is no injured target in range after its done gathering healing, it will explode but create no healing events at all. That is throwing off the cooldown tracker badly, either not creating the CBT cooldown at all if its the first one in combat, or combining that and the next CBT into a mega-CBT.

The normalizer fixes that by creating a dummy heal event with 100% overheal (which is basically what happened anyway) if there aren't any healing events to be found within 20 seconds after casting.

Now i wasn't 100% sure what properties an event needs, so i added all that i thought of important enough, and hitpoints/maxhitpoints so the masteryeffectiveness module doesn't give it NaN.

Example 1 (before and after)
image
image

Example 2 (before and after)
image
image

### DIFF
--- a/src/Parser/Shaman/Restoration/CombatLogParser.js
+++ b/src/Parser/Shaman/Restoration/CombatLogParser.js
@@ -55,6 +55,8 @@ import HealingWave from './Modules/Spells/HealingWave';
 import LavaSurge from './Modules/Spells/LavaSurge';
 import Resurgence from './Modules/Spells/Resurgence';
 
+import CloudburstNormalizer from './Normalizers/CloudburstNormalizer';
+
 import { ABILITIES_AFFECTED_BY_HEALING_INCREASES } from './Constants';
 
 class CombatLogParser extends CoreCombatLogParser {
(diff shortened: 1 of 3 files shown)


Gebuz:
You could probably avoid having to use a normalizer if you instead rewrote the CooldownThroughputTracker to check for Cloudburst totem death events and Recall Cloudburst casts instead of checking for a heal event. If you stick to the normalizer I'd suggest you check for Recall casts and use that timestamp when it is found.

niseko:
Good ideas, i didn't think of either of them - but it looks like CBT doesn't consistently fire Death events, i'm finding multiple in every log without death or recall. (first death is HTT)
image
same in a different log (there were no recalls)
image

Gebuz:
I only had a look at a single log but it had a death event for every cloudburst without a recall cast. Looking at some more logs it seems there is indeed missing death events as you say. Should be fine if you adjust the normalizer to use recall casts when found then.

niseko:
probably should've left this one alone until i'm back home, so i don't commit 3 times and realize that something is off after each one

PR #1504, opened 2018-03-20
[BrM] Converted Breath of Fire from uptime to # of hits mitigated

### LINKED ISSUE #1211: Update Ironskin Brew uptime stat to match checklist item
Currently, the ISB uptime Statistic shows buff uptime, while the checklist item shows % of hits mitigated. The latter doesn't penalize for dropping ISB e.g. on transitions or while not actively tanking. The two should be made to match.

The actual buff uptime should probably then get added to the tooltip.

emallson:
The main idea of this change is to switch the breath of fire checklist item from uptime % to # of hits mitigated. This ends up being trickier than ISB because:

- lots of bosses do abilities during intermissions that you can't realistically mitigate with BoF
- lots of adds due pulsing AoE's (looking at you Aggy)
- lots of irrelevant ability hits are generated by "NPC"s (aka imonar traps, aggy adds)

While I was doing this, I also fixed #1211 

**Before (H Imonar):**
screenshot from 2018-03-19 23-36-31
screenshot from 2018-03-19 23-36-40

**After:**
screenshot from 2018-03-19 23-37-23
screenshot from 2018-03-19 23-37-31

### DIFF
--- a/src/Parser/Monk/Brewmaster/CHANGELOG.js
+++ b/src/Parser/Monk/Brewmaster/CHANGELOG.js
@@ -9,6 +9,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-03-19'),
+    changes: <Wrapper>Converted <SpellLink id={SPELLS.BREATH_OF_FIRE.id} icon /> from uptime to hit tracking and updated hit tracking for <SpellLink id={SPELLS.IRONSKIN_BREW.id} icon />.</Wrapper>,
+    contributors: [emallson],
+  },
+  {
     date: new Date('2018-01-29'),
     changes: 'Added plot of damage in Stagger pool over time.',
     contributors: [emallson],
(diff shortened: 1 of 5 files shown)


Gebuz:
Are you sure that you want to use the blacklist for Ironskin Brew aswell? Looks to me like most of the spells in the list and environmental damage are still things you'd want to mitigate with brew maybe with the exception of imonar transition stuff.

emallson:
That's a good question. Going back through the list, the only one I'd say is really a problem if you don't mitigate it with ISB is empowered bombs. (It is worthwhile to keep in mind that ISB doesn't mitigate anything, it smooths it out. if the damage is *already smooth*--like when you've got many small hits from a pulsing aoe--it doesn't do much. This is compounded by the fact that everything on the list atm is magical, which makes stagger/isb worse)

That said, I'm going to split the lists up, because there is at least one we want BoF to ignore but not ISB at the moment, and that number is only going to grow.

PR #1505, opened 2018-03-20
[Balance Druid] Update L90Talents and Soul of the Archdruid and add them to checklist.

Gebuz:
- Updated Soul of the Archdruid to match other classes version with suggestion & checklist entry. Also moved it to the shared folder and added it to the other specs.
- Changed AstralPowerTracker to display Blessing of Elune as a separate entry, and fixed a bug where lunar strike was giving an incorrect amount with both Incarnation and Blessing of Elune active (it is rounded down to 22 but it actually gives 22.5).
- Updated L90Talents to include total numbers in the tooltip, and added a suggestion for when it performed worse than expected (ei. using wrong blessing). Also added it to the checklist.

image
image

New tooltip:
image

Old tooltip:
image

New AsP tab:
image

Old AsP tab:
image


src/Parser/Druid/Balance/Modules/Features/L90Talents.js
```diff
@@ -91,45 +90,53 @@ class L90_talents extends Analyzer {
...
       return this.shootingStarsValue / maxPotential;
     }
-    if (this.activeTalent.id === SPELLS.ASTRAL_COMMUNION_TALENT.id) {
+    else if (this.activeTalent.id === SPELLS.ASTRAL_COMMUNION_TALENT.id) {
       return this.astralCommunionValue / maxPotential;
     }
-    return Math.max(this.generatedPerMinute, this.BlessingOfEluneValue, this.BlessingOfAnsheValue) / maxPotential;
+    else if (this.activeTalent.id === SPELLS.BLESSING_OF_THE_ANCIENTS_TALENT.id) {
+      return Math.max(this.actualValue, this.BlessingOfEluneValue, this.BlessingOfAnsheValue) / maxPotential;
+    }
```
MartijnHols: This doesn't have a consistent return value; missing a default.

PR #1509, opened 2018-03-21
Encounter based stats

joshinat0r:
Adds a button to the character-pane which will load statistics based on the current encounter, spec and difficulty.

stats


MartijnHols:
I think you should call the load method from the constructor so it automatically loads after opening the tab. You might also want to show a loading indicator while loading. I don't think we need a button to delay loading this as it's not unique per log and so will be cached globally anyway.

Gebuz:
Great idea! Have you thought about using Icons for trinkets and legendaries too?  I'm thinking something along the lines of the below picture might give a better visual. Just imagine the percentages and icons match the items, and maybe add a border to show the item quality.
Also, since you sort the talents it might not be necessary to also change the opacity; it does give a nice visual representation of the percentages but it makes it a bit hard to see the icons.

unavngivet

joshinat0r:
Using large icons for trinkets and legendaries would mean I'd have to use ItemLink, right?
`{
                    "name": "Gravewarden Visage",
                    "quality": "epic",
                    "id": 147124
                }`
is everything I get from the WCL-API, so I dont have the icon's name to work with.

Using the ItemLink icon becomes already an issue when using it with trinkets, legendaries do work (I think because we have those icons).
Is there another way to get the icon from an itemID?

MartijnHols:
I liked the initial view more where it was a simple list with item names. How about small icons in front of those items? I don't think we should only show icons, especially trinkets but legendaries too aren't that easy to recognize by icon only. In web design it's usually considered bad practice to use icons without text if the icons aren't always immediately clear.

joshinat0r:
You had something like this in mind? I personally liked the icons-only grid more, looked cleaner and I don't have an issue recognizing good trinkets and talents.
But I'm fine with either one, I can see why one would want the additional info.

stats

MartijnHols:
Yes, I think that layout looks a lot better.

Gebuz:
How about aligning the percentages somehow? Dunno how important it is since they are ordered anyway, but would be faster to read.

joshinat0r:
As suggested, the percentage is now aligned. Having them aligned to the right looked kinda odd, hence left from the icon & name.
Missing icons are fetched from the bnet-api (got cors-errors when trying to fetch() from wowdb/wowhead).

stats

src/Parser/Core/Modules/Features/EncounterStats.js
```diff
@@ -0,0 +1,243 @@
...
+                    )}
+                  </div>
+                )}
+              </div>
+            </div>
+          </div>
+        </div>
+      );
+    } else {
+      this.load();
```
MartijnHols: You should never do any actions such as querying an API from the render method as things might be re-rendered multiple times a second which could trigger the query several times. A better place for this would be in the constructor (which is preferred over componentWillMount in ES6 classes).

joshinat0r:
stats

Issue #719, opened 2017-11-01
Track if Bonestorm hits 2+ mobs


yajinni:
Track if Bonestorm hits 2+ mobs. Its a waste if used on single target.

joshinat0r:
Can be closed, was implemented with RP #1514

PR #1514, opened 2018-03-22
[BDK] Bonestorm

joshinat0r:
Basically does what #719 describes 

bs
sugg


src/Parser/DeathKnight/Blood/Modules/Talents/Bonestorm.js
```diff
@@ -0,0 +1,104 @@
...
+    this.cast += 1;
+    this.totalCost += event.classResources[0].cost;
+  }
+
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.BONESTORM_HIT.id) {
+      return;
+    }
+
+    this.bsCasts.forEach((bsCast, index) => {
```
Gebuz: You could avoid iterating through the whole array if you just used `this.bsCasts[cast]` instead (just remember to check if  it's null first).

src/Parser/DeathKnight/Blood/Modules/Talents/Bonestorm.js
```diff
@@ -0,0 +1,104 @@
...
+    if (event.ability.guid !== SPELLS.BONESTORM_TALENT.id) {
+      return;
+    }
+
+    this.bsCasts.push({
+      timestamp: event.timestamp,
+      expire: event.timestamp + (event.classResources[0].cost * 10),
+      cost: event.classResources[0].cost,
+      hits: [],
+      cast: this.cast,
```
Gebuz: You could drop this prop since it's always equal to the index and just use the index in it's place.

src/Parser/DeathKnight/Blood/Modules/Talents/Bonestorm.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.BONESTORM_TALENT.id) {
+      return;
+    }
+
+    this.bsCasts.push({
+      timestamp: event.timestamp,
+      expire: event.timestamp + (event.classResources[0].cost * 10),
```
Gebuz: `timestamp` and `expire` are never used, did you want to use them for something and forgot?

PR #1522, opened 2018-03-24
Encounter stats highlighting

joshinat0r:
Since we moved the stats to another screen its harder to compare talents. Highlights the players talents, legendaries and trinkets (and adds the talent if its not in the top100) with a green color. Makes it easier to see if the player had an "optimal" setup without switching between two tabs.

stats

### DIFF
--- a/src/Main/EncounterStats.js
+++ b/src/Main/EncounterStats.js
@@ -42,8 +42,6 @@ class EncounterStats extends React.PureComponent {
       message: 'Loading statistics...',
     };
 
-    console.info(this.props.player);
-
     this.load = this.load.bind(this);
     this.load();
   }
--- a/src/Main/EncounterStats.js
+++ b/src/Main/EncounterStats.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import TickIcon from 'Icons/Tick';
+
 import SPECS from 'common/SPECS';
 import ROLES from 'common/ROLES';
 import ITEMS from 'common/ITEMS';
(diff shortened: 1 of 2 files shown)


MartijnHols:
How about instead of the tick the text "EQUIPPED" or something like that? Could make it white to stand out more. I don't think we should use a tick as that suggests it's the recommended item since the ones being ranked a few weeks into a tier are mainly about parsing.

Also I don't think this is appropriate for talents since they're much more situational.

joshinat0r:
You're right about the tick-icon, changed that to the armor-icon. Fits better.
The white highlighting looks kinda odd, but I think red looks way better than green. 
Also slightly change the opactiy for the talents that are from the statistics (1 opacity for selected, .9 for talents from stats) to highlight it even further. Also thought about changing the grayscale, but not sure if thats a good idea at all (might cause the same "you shouldn't pick those"-feeling)

stats_red
stats_white

MartijnHols:
I meant to just make the equipped part red. I think as it is it might be more confusing than revealing, creating questions like why is this percentage red (is something wrong with it?), red=bad so why are the item links red? what does that chest icon mean? I think you could be most clear by leaving the item link standard and putting just `EQUIPPED` in white after it, or if that doesn't look decent `(equipped)`.

As for talents I don't think they should be grayed out since it's more important to show the rankings than it is what was equipped. Using the red border for selection could work, but it might also confuse people especially since that same info is more logically found in the character pane.

joshinat0r:
You can see the legendaries and trinkets aswell in the char pane, so both are not necessary on the stats pane.
Having just an "EQUIPPED" behind the item names in red (and the rest in the items color) looks decent, but the talent selection looks weird when you add text to it. A border alone isn't really clear enough.

Just thought it's a nice addition to see fast "hey, I used a good talent/trinket/legendary setup" without having to switch between two tabs when you check logs from yesterday or from a raidmember. But let's scrap the idea if there's no intuitive way to show that on this pane and keep it as it is to avoid confusion.

PR #1529, opened 2018-03-26
[Tanks] Antorus Trinkets

### LINKED ISSUE #1516: Add Antorus Tank Trinkets
- [ ] Diima's Glacial Aegis
- [ ] Eye of F'harg (appears as 151974 in logs? 152645 stam, 153544 agi/str)
- [ ] Riftworld Codex
- [ ] Smoldering Titanguard

joshinat0r:
Adds Antorus Trinkets (fixes #1516).

"Eye of F'harg" is not possible to implement as it appears as [Eye of'](http://www.wowhead.com/item=151974/eye-of&bonus=3610:1472) in the log. 
But the actual versions are [Eye of Shatug](http://www.wowhead.com/item=153544/eye-of-fharg&bonus=3610:1472) and [Eye of F'harg](http://www.wowhead.com/item=153544/eye-of-fharg&bonus=3610:1472), with the on-use to switch between both versions during combat. This makes it impossible to track unless the player used the trinket at least once (which happens very very rarely (looked into 20 logs, not a single cast)). The "empowerment" buff between both trinkets is also not trackable as the [buff](http://www.wowhead.com/spell=255769/eye-of-the-hounds) that grands the bonus armor/versatility is invisible and does not appear in the combat log.
TL;DR: it's a 🍝 -trinket with 🍝 -code and a waste of time to implement.

rift
diima
titan

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-03-26'),
+    changes: <Wrapper>Added <ItemLink id={ITEMS.SMOLDERING_TITANGUARD.id} icon />, <ItemLink id={ITEMS.DIIMAS_GLACIAL_AEGIS.id} icon /> and  <ItemLink id={ITEMS.RIFTWORLD_CODEX.id} icon />.</Wrapper>,
+    contributors: [joshinator],
+  },
+  {
     date: new Date('2018-03-24'),
     changes: 'Added statistics for talents, legendaries and trinkets based on the current encounter.',
     contributors: [joshinator],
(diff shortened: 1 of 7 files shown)


Gebuz:
Looks good, may I suggest that you sum the healing numbers from Riftworld Codex and move the breakdown to the tooltip and show average armor gained from Diima's?

As for invisible buffs, I do believe we add them as fabricated events (that's how we track tier bonuses for instance), but if it's impossible to detect which trinket is used I suppose it doesn't matter much anyway.

joshinat0r:
rift
diima

It doesn't looks like the fabricated get the stat-change from the trinkets, but I'll take a look into it (if I need to change anything), since thats the only number we could show anyways (they are just stat sticks that provide the other tank a buff while within 15yrd (based on the "state" of the trinket).

PR #1532, opened 2018-03-28
[Frost Death Knight] Properly tracks reduced cooldowns

jjs451:
Adds SpellUsable to Frost to address issues with tracking CDs for Pillar of Frost and Sindragosa's Fury.

### DIFF
--- a/src/Parser/DeathKnight/Frost/Modules/Features/SpellUsable.js
+++ b/src/Parser/DeathKnight/Frost/Modules/Features/SpellUsable.js
@@ -1,5 +1,4 @@
 import SPELLS from 'common/SPELLS';
-import ITEMS from 'common/ITEMS';
 import CoreSpellUsable from 'Parser/Core/Modules/SpellUsable';
 import Combatants from 'Parser/Core/Modules/Combatants';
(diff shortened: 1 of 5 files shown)


src/Parser/DeathKnight/Frost/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,67 @@
...
+        this.reduceCooldown(SPELLS.PILLAR_OF_FROST.id, ICECAP_COOLDOWN_REDUCTION_MS);
+        this.lastCritTime = event.timestamp;
+      }
+    }    
+  }
+
+  beginCooldown(spellId, timestamp) {
+    if (this.hasIcecap && spellId === SPELLS.PILLAR_OF_FROST.id) {
+      if (!this.isAvailable(spellId)) {
+        this.endCooldown(spellId);
```
Gebuz: You should only do this if the cooldown reduction is untrackable, going by the above calculations I assume that it is trackable?

jjs451: Oh yeah this should be commented out.  I shouldn't PR stuff at the end of the day when I'm tired and want to hurry to bed

PR #1533, opened 2018-03-28
[Restoration Shaman] Mastery effectiveness tab

niseko:
I'm using the paladin mastery effectiveness tab since their mastery works very similar, just changed it slightly so it works for both classes and added a column with healing done numbers.
image

Also since people were talking about resurgence being hard to read, i moved it lower and bolded the text which should help. Removing the transparency when expanded would probably be better but i couldn't quite make it work without also changing the non-expanded background, so i left it as is.


niseko:
since this is still open anyway, I noticed Velens wasn't accounting for totem healing and added a fix for it

PR #1535, opened 2018-03-28
Contributor pages

joshinat0r:
[Example](https://github.com/WoWAnalyzer/WoWAnalyzer/compare/master...joshinat0r:contributors?expand=1#diff-2fd27eaa6be3cdc1a4ce7d74ad1d5e49) for all fields allowed in CONTRIBUTORS.
Existing contributor-components are now links to the contributors page.

contributor with new props:
contrib1

contributor without new props:
contrib2

changelog/commit expandable per class
contrib3


MartijnHols:
Btw (changelog) dates should be YYYY-MM-DD.

joshinat0r:
The wrong date format was just the result of copy&pasting the old changelog-format into the new one. 
I'll add that and the code/style refactoring to my to-do list.

src/CONTRIBUTORS.js
```diff
@@ -3,10 +3,42 @@
...
- * @type {{nickname: string, github: string, avatar: *}}
+ * 
+    EXAMPLE
+
+    export const NICKNAME = {
+      nickname: 'NICKNAME',
+      github: 'GITHUB_NAME',
+      avatar: require('./Images/IMAGE'),
+      desc: 'DESC',
+      maintainer: [
```
Gebuz: This creates some duplication as maintainers are already defined in the specs' CONFIG file. Would it be possible to get the info from those instead? Alternatively we could define it here instead of in the CONFIG files, I'm not sure which would be prefered.

joshinat0r: I'll change that aswell.
Using the AVAILABLE_CONFIGS for importing the changelogs will allow me easily to access the specs maintainer. Would also mean we avoid going through all specs + files to make the maintainer contributor-based.

Issue #1266, opened 2018-01-26
Wrong icons from wow's api


Gebuz:
It's pretty confusing when some core abilities uses a completely different icon. For example:
Corruption icon according to wowhead (matches in game): 
http://www.wowhead.com/spell=172/corruption
Corruption according to the wow api (not even remotely close to in game):
https://render-us.worldofwarcraft.com/icons/56/spell_shadow_abominationexplosion.jpg

I know the bug isn't on our end but I don't think it's likely to be fixed on their end any time soon (I know feral and guardian affinity have always been wrong since they were added in legion beta). I don't know if it's possible to get the correct icons from the wow api using a different icon name (that would be ideal) or if the solution would be to just store the few icons that are wrong.

Of the top of my head the following icons are wrong from the specs I have looked at recently:
* Guardian Affinity
* Feral Affinity
* Corruption
* Control Undead
* Death Strike
* Blood Boil

yajinni:
I have noticed blood boil being fel green. Havent payed attention to death strike.
Was considering downloading the right icons and linking to them. 
This a wow or wowhead issue?

yajinni:
Blood boil was fine btw when i first started. A couple of months ago it suddenly changed.

Gebuz:
afaik wowhead has the correct icon names that they get from the game files and it's just https://render-us.worldofwarcraft.com giving us the wrong icons.

MartijnHols:
Not sure how to solve this. I don't know of any other CDN. Should we make a new GitHub repo with all the icons and manually manage them? I don't really want to solve this with code (apart from changing the URL).

Gebuz:
It might be possible to get them from wowhead, looks like they come without a border though: 
![image](https://user-images.githubusercontent.com/6773230/35758653-68c89fae-0875-11e8-9797-e7b1e36f509e.png)
http://wow.zamimg.com/images/wow/icons/large/talentspec_druid_feral_cat.jpg

MartijnHols:
I don’t think they would allow us to do that.

Gebuz:
fixed in #1537

PR #1537, opened 2018-03-29
fix for wrong icons from blizzards CDN

joshinat0r:
Fixes issues #1266 
Basically looks into the BAD_ICONS-array and replaces the blizzard-url with our local folder if the icons name is there.
The icons mentioned in the issue are already fixed, not sure if there are other bad ones.

**Note**: some icons from wowheads (and/or wowdbs) CDN have a light border in the corners, requires some editing in photoshop/gimp to remove them. Adding a border-radius to the icon-component would also fix that, but those icons look weird when they have a radius of 4px (which is necessary to hide the borders).

**Live**:
corruption_old

**Fixed**:
corruption_new


src/common/BAD_ICONS.js
```diff
@@ -0,0 +1,16 @@
+// icons in this array get our icons instead of the ones from blizzards API
+// make sure to add the correct icon in public/img/Icons/
```
MartijnHols: Could add a development time only check in this file to ensure the files exist. In fact importing the files could achieve this too and allow the images to be in a dir next to this file, which would be slightly cleaner.

PR #1544, opened 2018-04-04
[Hunter] Lacerate and other small updates

### LINKED ISSUE #1217: [Survival] General ideas
For my own sake of remembering and collecting suggestions and more:

- average time left on mongoose fury when aspect of the eagle is used 
- mok uptime, maybe average time left when you refresh, or a list of the times left when you refreshed it
- a blurb on the cobra section about saving it for lust because it scales really well with haste (just hitting dummy no lust snek hit 18 times, lust snek hit 37 times)
- check if the player is Aspect of the Skylord'ing before a MF window (and ensure MF window is at X stacks)
- check if the player lacerates too much
- check for On The Trail uptime

### LINKED ISSUE #1543: [Hunter] Average time left on refreshes
Things like:
Way of the moknathal for SV
BM chest legendary
probably more

Pewtro:
Not sure what happened with the package thing.. 

Created a lacerate module to track whether a person is lacerating too much (it's not a traditional DOT which you want to maximize uptime of)
-- considering to implement something for too low uptime in the future aswell. (but some people have had great success just completely removing this spell from their bars.. JustBlizzardBalancing)

Report: https://www.warcraftlogs.com/reports/gmVqNJjTcpxHKCG6/#fight=3&source=3
image

Created a On The Trail module for further suggestions down the line
Report: https://www.warcraftlogs.com/reports/gmVqNJjTcpxHKCG6/#fight=3&source=3
image

Minor updates around Beast Mastery and Survival modules, such as average time between refreshes on certain legendary buffs / abilities. 

fixes #1543 
fixes #1217


MartijnHols:
Could you revert the package-lock.json please? It's probably caused by the usage of an older version of NPM, but we can't really merge it with this change in since it might lead to dependency issues.

PR #1546, opened 2018-04-05
Contributors

joshinat0r:
Adresses suggested changes from the last version

- JSX returns wrapped in parenthesis
- removed unnecessary `== true` & `== undefined` & `else` checks
- changed paddings from 30px to the 20px for more layout-consistency
- using AVAILABLE_CONFIGS instead of manual importing all changelogs
- pulling maintained specs from AVAILABLE_CONFIGS instead of CONTRIBUTORS

Also changed to contributor-component to open a modal instead of a new link. 
The ability to link a contributor is still available, might be usefull for something.

Wrapping the modal inside of the portal-component was necessary due to CSS inheritance (e.g. the contributor-components in the changelog-tab would inherit the text-mute & text-align style). I think its cleaner than overwriting the styles that cause issues.
Also might be a usefull in the future for adding modals / alerts / overlays.

**Overlay triggered by contributor-components:**
contrib_overlay

**Old link:**
contrib_page

### DIFF
--- a/src/CONTRIBUTORS.js
+++ b/src/CONTRIBUTORS.js
@@ -9,6 +9,7 @@
     export const NICKNAME = {
       nickname: 'NICKNAME',
       github: 'GITHUB_NAME',
+      discord: 'DISCORD_NAME INCL #xxxx',
       avatar: require('./Images/IMAGE'),
       desc: 'DESC',
       mains: [{
@@ -304,6 +305,7 @@ export const Cloake = {
 export const joshinator = {
   nickname: 'joshinator',
   github: 'joshinat0r',
+  discord: 'joshinator#7267',
   mains: [{
     name: "Êxtêndêd",
     spec: SPECS.BLOOD_DEATH_KNIGHT,
--- a/src/CONTRIBUTORS.js
+++ b/src/CONTRIBUTORS.js
@@ -11,9 +11,6 @@
       github: 'GITHUB_NAME',
       avatar: require('./Images/IMAGE'),
       desc: 'DESC',
-      maintainer: [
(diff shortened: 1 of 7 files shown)


Gebuz:
Since discord is our primary tool of communication (other than github), I think it would be a good idea to add it below the github name. I think we also have a list of github:discord names we could quickly copy over to CONTRIBUTORS.js somewhere but I can't seem to find it just now.

joshinat0r:
Isn't the nickname always the discord-name (at least from a quick glance it seems like it is)? 
If so thats the text above the avatar, adding the same name another time seems redundant.
Might be usefull to add the complete discord-name ( with the #xxxx ) and add that so users can PM the contributor.

sMteX:
Not always, on GitHub I'm sMteX but on Discord I'm known as Chizu (added the github nickname for clarity).

joshinat0r:
Right, but your entry looks like

```
export const Chizu = {
  nickname: 'Chizu',
  github: 'sMteX',
  avatar: require('./Images/Chizu_avatar.jpg'),
};
```

which means your discord-name is already the title/name, and the correct GitHub name is listed below the avatar.

sMteX:
Fair enough, I've forgotten how the contributors look, thought you want to derive discord name from github name.

Gebuz:
Yeah that is what I meant, adding the discord tag so ppl can msg us.

joshinat0r:
technically already possible with
```
 others: {
        "Custom Item": "normal text",
}
```
which would be enough since there's no DM/profile link for discord (afaik)
but I could add a dedicated discord tag

PR #1547, opened 2018-04-05
Various improvements to Arms Warrior

Aelexe:
An initial commit for my contribution to the Arms Warrior analyzer. The improvements are more or less explained in the commit messages.


src/Parser/Warrior/Arms/Modules/Abilities.js
```diff
@@ -1,24 +1,23 @@
 import SPELLS from 'common/SPELLS';
 
 import CoreAbilities from 'Parser/Core/Modules/Abilities';
+import ISSUE_IMPORTANCE from 'Parser/Core/ISSUE_IMPORTANCE';
 
 class Abilities extends CoreAbilities {
   spellbook() {
     return [
```
Gebuz: It looks like this module is slightly outdated. I'd recommend defining the combatant locally here. This will be useful when you need to check if a talent or tier bonus is active.
```
spellbook() {
  const combatant = this.combatants.selected; //add this line
  return [
```

src/Parser/Warrior/Arms/Modules/Abilities.js
```diff
@@ -1,24 +1,23 @@
...
-        },
+        cooldown: haste => 6 / (1 + haste),
       },
       {
         spell: SPELLS.COLOSSUS_SMASH,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: 20,
-        castEfficiency: {
-          suggestion: true,
+        cooldown: (haste, combatant) => {
```
Gebuz: Is CS cd reduced by haste? Your changelog entry says it is, but you don't use it here.
If it isn't, you don't need to use a function here. You can instead use the local combatant (see comment above) to get talent info. The difference between the two is that the local version snapshots your character at encounter start, so the combatant passed by the function is only needed if you need to check for a buff at a specific time during the encounter.

src/Parser/Warrior/Arms/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,40 @@
+import SPELLS from 'common/SPELLS';
+import CoreSpellUsable from 'Parser/Core/Modules/SpellUsable';
+
+const TACTICIAN_ID = SPELLS.TACTICIAN.id;
+const MORTAL_STRIKE_ID = SPELLS.MORTAL_STRIKE.id;
+const COLOSSUS_SMASH_ID = SPELLS.COLOSSUS_SMASH.id;
```
Gebuz: no need to set these here, you can just use `SPELLS.COLOSSUS_SMASH.id` directly :)

src/common/SPELLS/TALENTS/WARRIOR.js
```diff
@@ -22,6 +22,7 @@ export default {
   TITANIC_MIGHT_TALENT: { id: 202612, name: "Titanic Might", icon: "ability_warrior_colossussmash" },
   DEADLY_CALM_TALENT: { id: 227266, name: "Deadly Calm", icon: "achievement_boss_kingymiron" },
   IN_FOR_THE_KILL_TALENT: { id: 248621, name: "In For The Kill", icon: "ability_blackhand_marked4death" },
+  IN_FOR_THE_KILL_TALENT_BUFF: { id: 248622, name: "In For The Kill", icon: "ability_blackhand_marked4death" },
```
Gebuz: The talent files are auto generated, so this would be deleted the next time they are generated. You should instead add it to src/common/SPELLS//WARRIOR.js.

src/Parser/Warrior/Arms/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,40 @@
...
+
+class SpellUsable extends CoreSpellUsable {
```
Gebuz: You don't need to change this now but in the future you might want a tactician module that tracks wasted procs. When doing that it would be appropriate to also move handling of the resetting to said module.

src/Parser/Warrior/Arms/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,36 @@
...
+  on_byPlayer_refreshbuff(event) {
+    if(super.on_byPlayer_refreshbuff) {
+      super.on_byPlayer_refreshbuff(event);
+    }
+    this.handleTacticianBuff(event);
+  }
+
+  handleTacticianBuff(event) {
+    // If the buff being applied/refreshed is tactician, refresh the cooldowns of mortal strike and colossus smash.
+    if (SPELLS.TACTICIAN.id === event.ability.guid) {
```
Gebuz: My bad for missing this the first time around but we prefer avoiding nested if statements. Instead check for the opposite and return early when possible.

PR #1551, opened 2018-04-07
[Arcane Mage] Evocation

Sharrq:
Checks to see if Evocation is available or has less than 20 seconds left on the cooldown when the user hits Arcane Power. This is because you pop Arcane Power and then go into a burn phase where you burn your mana until its empty. If Evocate isnt off cooldown when you bottom out your mana, then you are sitting there until it comes off cooldown.


Sharrq:
Don't merge. I'm going to adjust the thresholds

Sharrq:
alright, should be good to be reviewed now :)

Gebuz:
Looks good, just wondering if it shouldn't be called ArcanePower instead since it's bad Arcane Power casts you are checking for, not evocation?

Sharrq:
yea i bounced back and forth on the naming, but ultimately im checking the cooldown on evo, and ill be checking a bunch of other stuff for Arcane Power, so I opted for separating this so the Arcane Power one doesnt get overcrowded

PR #1554, opened 2018-04-07
[Protection Paladin] some stuff

joshinat0r:
- Adds basic checklist
- Fixes the Righteous Protector-Module by using the correct Wings & Hand/Light of the Protector ID
- Adds a selfheal-timing graph for Hand/Light of the Protector effectiveness
- Adds a Seraphim-Module for average uptime + stat-gain
- Adds Heathcliff's Immortality (Legendary)
- Adds Breastplate of the Golden Val'kyr (Legendary)

The old Deathstrike-Timing graph shared basically everything but the SpellID&Name with the LotP-Graph. So I made a more generic core-module out of it, just needs the correct spell for the class to work with other specs (for selfheal & no-HoT heals).

sera
cheklisrt
lego1
lego2
prot
selfheal


src/Parser/Paladin/Protection/Modules/Talents/Seraphim.js
```diff
@@ -0,0 +1,90 @@
...
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import { formatPercentage } from 'common/format';
+
+/**
+ * Consumes up to 2 SotR charges to provice 5500 Haste+Vers+Mastery+Crit for 8sec per consumed charge
+*/
+
+//ToDo: remove possible SotR casts from Ability-Tab that were used by Seraphim
```
Gebuz: You can just do `this.spellUsable.beginCooldown(SPELLS.SHIELD_OF_THE_RIGHTEOUS.id, event.timestamp);` for each charge spent by Seraphim

joshinat0r:
Maybe should explain the 100ms buffer, chargesAvailable would return the wrong amount of charges when I remove the SotR charges that were used for seraphim. That would cause up to 6 seconds of missing CD on a SotR charge that was actually used but still marked on cooldown.

The only issue I could imagine was some latency (or not exactly 2-4s CDR from Judgment) since haste calculation itself and SotR cooldown reduction are all implemented.

Gebuz:
It would definitely be better to find the source of the problem and make sure that SotR cooldown is tracked properly.

joshinat0r:
This was actually not fully ready for a merge 😅 
The tracking of SotR isn't 100% on point, which may cause inaccuracies for charges spend on seraphim (and later result in errors for SotR casts that were calculated as not available).

I'll make a new PR that removes the tooltip with the details about seraphim and removes 2 charges per sera cast to get the cast efficiency right.

PR #1560, opened 2018-04-09
[Arms Warrior] Create Battle Cry analyzer with Shattered Defense analysis

Aelexe:
Created the Battle Cry analyzer which currently will make suggestions regarding the preparation of Shattered Defenses.

The BattleCry data class is in place for future features that require collating a bit of information.

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
+++ b/src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
@@ -1,0 +1,84 @@
+import React from 'react';
+
+import Wrapper from 'common/Wrapper';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Analyzer from 'Parser/Core/Analyzer';
+
+const SHATTERED_DEFENSES_ICON = 'warrior_talent_icon_igniteweapon';
+
+/**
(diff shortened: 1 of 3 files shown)


src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
```diff
@@ -0,0 +1,84 @@
...
+        return suggest(<Wrapper>You should ensure <SpellLink id={SPELLS.SHATTERED_DEFENSES.id} icon/> is up before you use <SpellLink id={SPELLS.BATTLE_CRY.id} icon/> to maximize your burst potential.</Wrapper>)
+          .icon(SHATTERED_DEFENSES_ICON)
+          .actual(`Shattered Defenses was up for ${formatPercentage(actual)}% of Battle Cries.`)
+          .recommended(`${formatPercentage(recommended)}% is recommended`);
+      });
+  }
+}
+
+/** Class containing information regarding a single use of Battle Cry */
+class BattleCry {
```
Gebuz: Making a new class for this seems a bit overkill. Unless you intend to use it cross module (in which case it should also have it's own file) it would be simpler to just use an object for it: 
```
this.currentBattleCry = 
{ 
  shatteredSetup = true, 
  usedDuringAvatar = true,
  mortalStrikesCast = 0,
}
```

Pewtro:
As already mentioned by Gebuz, the class simply called BattleCry should be it's own file to keep it consistent with all other similiar modules, and then you just add it as a dependency if you need to use it elsewhere imo.

PR #1561, opened 2018-04-09
[Hunter] fix crash and update contributors

Pewtro:


joshinat0r:
Btw: the maintainer-prop isn't necessary anymore. The maintainer is pulled automatically from eachs specs config-file.

PR #1562, opened 2018-04-09
[Death Knight] Checklist update

jjs451:
I added a basic checklist for Frost and added the rune requirement to Unholy.  Frost's checklist feels a little sparse but I'm not sure what else to add at the moment.  The spec is so simple there's not a lot of advice beyond "press buttons as often as possible, don't waste resources". 

image


src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,125 @@
...
+      name: 'Use core spells as often as possible',
+      description: <Wrapper>Spells with short, static cooldowns like <SpellLink id={SPELLS.REMORSELESS_WINTER.id}/> and <SpellLink id={SPELLS.CHAINS_OF_ICE.id}/>(when using Cold Heart) should be used as often as possible</Wrapper>,
+      requirements: () => {
+        const combatant = this.combatants.selected;
+        return [
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.REMORSELESS_WINTER,
+          }),
+          new GenericCastEfficiencyRequirement({
+            spell: SPELLS.CHAINS_OF_ICE,
```
Gebuz: CoI isn't in abilities.js for frost (a lot of other abilities are probably missing aswell, it's quite short). I would also recommend tracking it using the cold heart module to get a requirement for this as setting CoI to having a cooldown would throw errors whenever it is used early (it does this for UH atm) and the Cold Heart module was updated to have a lot of extra rules for casting early/late iirc.

src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,125 @@
...
+            new GenericCastEfficiencyRequirement({
+              spell: SPELLS.OBLITERATION_TALENT,
+              when: combatant.hasTalent(SPELLS.OBLITERATION_TALENT.id),
+            }),
+            new GenericCastEfficiencyRequirement({
+              spell: SPELLS.BREATH_OF_SINDRAGOSA_TALENT,
+              when: combatant.hasTalent(SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id),
+            }),
+            new GenericCastEfficiencyRequirement({
+              spell: SPELLS.EMPOWER_RUNE_WEAPON,
```
Gebuz: missing hungering rune weapon here?

src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
```diff
@@ -0,0 +1,125 @@
...
+      },
+    }),
+    new Rule({
+      name: 'Try to avoid being inactive for a large portion of the fight',
+      description: <Wrapper>While some downtime is inevitable in fights with movement, you should aim to reduce downtime to prevent capping Runes.  You can reduce downtime by casting ranged abilities like <SpellLink id={SPELLS.FROST_STRIKE_CAST.id}/></Wrapper>,
+      requirements: () => {
+        return [
+          new Requirement({
+            name: 'Downtime',
+            check: () => this.alwaysBeCasting.downtimeSuggestionThresholds,
```
Gebuz: the downtime suggestion will likely be a bit off since ability.js seems pretty short atm.

jjs451: I thought downtime came from AlwaysBeCasting.js, not Abilities.js.

src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
jjs451: Whoops, forgot to add CoI to abilities.  The rules for casting CH are different for frost and uh and I'm inclined to forgo adding a new module for frost.  It's lazy of me but CH is rarely used with frost and with BFA in 4 months it'll become a moot point.  I should go and make sure UHs module updates the cast efficiency correctly, though.  CH is widely used so the majority of UH players who use the tool will get those errors.

src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
jjs451: Oh yeah that talent exists :P

src/Parser/DeathKnight/Frost/Modules/Features/Checklist.js
Gebuz: You're right, I didn't check but frost does indeed have the GCDs set in ABC so it should be correct. That way of doing it is deprecated though, ideally it should be set in Abilities with `isOnGCD: true,`

jjs451:
Updated the glaring issues.  Created a new issue to update Abilities.js to use the new method of downtime tracking.

PR #1563, opened 2018-04-09
[Arms Warrior] Add Battle Cry damage statistic block

Aelexe:
Added a Battle Cry damage statistic block that lets you see how much damage you dealt during your burst window. I likely won't add a suggestion for this as smaller suggestions I'll be adding would lead to this metric being reached anyway.

Also addressed a few code review comments from #1560.

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
+++ b/src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
@@ -1,13 +1,16 @@
 import React from 'react';
 
-import Wrapper from 'common/Wrapper';
-import SPELLS from 'common/SPELLS';
+import { formatNumber, formatPercentage } from 'common/format';
 import SpellLink from 'common/SpellLink';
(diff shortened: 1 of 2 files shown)


src/Parser/Warrior/Arms/Modules/Features/BattleCry.js
```diff
@@ -40,45 +44,95 @@ class BattleCryAnalyzer extends Analyzer {
...
+        label="Average damage during Battle Cry"
+        tooltip={`Damage dealt during Battle Cry contributed ${formatPercentage(battleCryDamage / totalDamage)}% of your total damage done.`}
+      />
+    );
+  }
+  statisticOrder = STATISTIC_ORDER.CORE(4);
 }
 
 /** Class containing information regarding a single use of Battle Cry */
 class BattleCry {
```
Gebuz: As mentioned previously, you should move this class to it's own file.

PR #1564, opened 2018-04-10
[Arms Warrior] Add Rend tracking and Rend/Execute range suggestion

Aelexe:
Added Rend to the ability list along with a suggestion to not use it during Execute range. 

I was limited by the cast and debuff events not having access to the enemy health, so I flagged the target id to be checked in the next damage event. This lead to an issue where damage from a good Rend would leak over into the Execute range.

As a workaround I would only flag a Rend as bad if it wasn't the first one whose first damage instance occurred in Execute range, which is probably fine as the player's triggering this suggestion are likely using Rend more than once in Execute range on the same target.

image

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Talents/Rend.js
+++ b/src/Parser/Warrior/Arms/Modules/Talents/Rend.js
@@ -5,6 +5,7 @@ import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 import Combatants from 'Parser/Core/Modules/Combatants';
 import Enemies from 'Parser/Core/Modules/Enemies';
+import { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';
 import Analyzer from 'Parser/Core/Analyzer';
 
 const EXECUTE_RANGE = 0.2;
@@ -26,28 +27,30 @@ class RendAnalyzer extends Analyzer {
   on_byPlayer_cast(event) {
     if(SPELLS.REND_TALENT.id === event.ability.guid) {
       // If Rend is used flag the target to check their health on the next damage event.
-      this.targets.push(event.targetID);
+      this.targets.push(encodeTargetString(event.targetID, event.targetInstance));
     }
   }
 
   on_byPlayer_damage(event) {
-    const targetId = event.targetID;
+    const targetString = encodeTargetString(event.targetID, event.targetInstance);
 
-    if(SPELLS.REND_TALENT.id === event.ability.guid && this.targets.indexOf(targetId) !== -1) {
+    if(SPELLS.REND_TALENT.id === event.ability.guid && this.targets.indexOf(targetString) !== -1) {
+      // Unflag the target for checking.
+      this.targets.splice(this.targets.indexOf(targetString), 1);
+
       // Check if the target's health was within execute range.
-      if(event.hitPoints / event.maxHitPoints < EXECUTE_RANGE) {
-        if(this.failedTargets.indexOf(targetId) === -1) {
-          // If this is the first time Rend was used on this target in execute range we flag that target.
-          // This is to prevent false positives from Rend being applied just as the target enters execute range.
-          this.failedTargets.push(targetId);
-        } else {
-          // Increment the counter
-          this.rendsInExecuteRange += 1;
-        }
+      if(event.hitPoints / event.maxHitPoints >= EXECUTE_RANGE) {
+        return;
       }
 
-      // Unflag the target for checking.
-      this.targets.splice(this.targets.indexOf(targetId), 1);
+      if(this.failedTargets.indexOf(targetString) === -1) {
+        // If this is the first time Rend was used on this target in execute range we flag that target.
+        // This is to prevent false positives from Rend being applied just as the target enters execute range.
+        this.failedTargets.push(targetString);
+      } else {
+        // Increment the counter
+        this.rendsInExecuteRange += 1;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


Pewtro:
Cast should have access to enemy health. See https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Hunter/Marksmanship/Modules/Talents/AMurderOfCrows.js for an example.

Aelexe:
When I tried that it seemed like it was accessing the player health instead, which seemed odd, but I'll check again. That would make this module much cleaner.

Pewtro:
Maybe some changes happened causing that, the MM crows module hasn't changed in ages - but I see what you're saying myself now aswell 🤔 
I'll get a few extra hours of sleep and look more into it, otherwise yeah it's going to be the wonky work arounds that are go-to 👍

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
```diff
@@ -0,0 +1,72 @@
...
+      this.targets.push(event.targetID);
+    }
+  }
+
+  on_byPlayer_damage(event) {
+    const targetId = event.targetID;
+
+    if(SPELLS.REND_TALENT.id === event.ability.guid && this.targets.indexOf(targetId) !== -1) {
+      // Check if the target's health was within execute range.
+      if(event.hitPoints / event.maxHitPoints < EXECUTE_RANGE) {
```
Gebuz: It's prefered to avoid nested if statements by checking for the opposite and returning early instead. Applies to both of the above if statements. (you can move `the this.targets.splice(this.targets.indexOf(targetId), 1);` line up to right after the first if statement passes since it is removed no matter what at that point)

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
```diff
@@ -0,0 +1,72 @@
...
+        if(this.failedTargets.indexOf(targetId) === -1) {
+          // If this is the first time Rend was used on this target in execute range we flag that target.
+          // This is to prevent false positives from Rend being applied just as the target enters execute range.
```
Gebuz: Are you sure these casts shouldn't just be counted as bad aswell? I have no idea what the threshold is as I don't play warrior, but since this check always happens on the first tick of the Rend I would imagine it might be just as bad as casting it druing execute range since it's rage you could've spent on execute anyway.

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      // Unflag the target for checking.
+      this.targets.splice(this.targets.indexOf(targetId), 1);
+    }
+  }
+
+  get executeRendsThresholds() {
+    return {
+			actual: this.rendsInExecuteRange,
```
Gebuz: I'd suggest using per minute or percent of total rend casts for this to keep the threshold relative to encounter duration.

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
```diff
@@ -0,0 +1,72 @@
...
+  failedTargets = [];
+
+  on_initialized() {
+		this.active = this.combatants.selected.hasTalent(SPELLS.REND_TALENT.id);
+	}
+
+  on_byPlayer_cast(event) {
+    if(SPELLS.REND_TALENT.id === event.ability.guid) {
+      // If Rend is used flag the target to check their health on the next damage event.
+      this.targets.push(event.targetID);
```
Gebuz: Multiple enemies can have the same ID, so you'll have to use `encodeTargetString(event.targetID, event.targetInstance)` (import: `import { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';`)

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
```diff
@@ -0,0 +1,72 @@
...
+			style: 'number',
+		};
+  }
+
+  suggestions(when) {
+    when(this.executeRendsThresholds).addSuggestion((suggest, actual, recommended) => {
+        return suggest(<Wrapper>You should avoid using <SpellLink id={SPELLS.REND_TALENT.id} icon/> on a target in <SpellLink id={SPELLS.EXECUTE.id} icon/> range.</Wrapper>)
+          .icon(SPELLS.REND_TALENT.icon)
+          .actual(`Rend was used ${actual} times on a target in execute range`)
+          .recommended('0 is recommended.');
```
Gebuz: you can use `${recommended}` instead of `0` here so it gets the value from the suggestionThreshold.

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
Aelexe: Avoiding using rend during execute range isn't a mechanic that gets more difficult with encounter length. I think it's best as a hard suggestion so if the player makes the mistake even once they're guaranteed to receive feedback about it.

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
Gebuz: alright sounds reasonable.

src/Parser/Warrior/Arms/Modules/Talents/Rend.js
Gebuz: What about encounters with adds or multiple bosses dying at different times? There's also the fact that messing up 1 rend is significantly more impact full on a 1 min. encounter than a 10 min. encounter.

PR #1573, opened 2018-04-12
[Arms Warrior] Add wasted Colossus Smash/Shattered Defenses suggestion

Aelexe:
Adds a suggestion to avoid using Colossus Smash with Shattered Defenses already active.

image

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Core/ColossusSmash.js
+++ b/src/Parser/Warrior/Arms/Modules/Core/ColossusSmash.js
@@ -7,18 +7,15 @@ import Wrapper from 'common/Wrapper';
 
 import Analyzer from 'Parser/Core/Analyzer';
 import Combatants from 'Parser/Core/Modules/Combatants';
-import Enemies from 'Parser/Core/Modules/Enemies';
 
 import SpellUsable from '../Features/SpellUsable';
 
 class ColossusSmashAnalyzer extends Analyzer {
   static dependencies = {
     combatants: Combatants,
-    enemies: Enemies,
     spellUsable: SpellUsable,
   };
 
-  // Statistics
   colossusSmashes = 0;
   shatteredColossusOverlaps = 0;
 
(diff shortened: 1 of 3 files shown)


src/Parser/Warrior/Arms/Modules/Core/ColossusSmash.js
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  colossusSmashes = 0;
+  shatteredColossusOverlaps = 0;
+
+  on_byPlayer_cast(event) {
+    if(SPELLS.COLOSSUS_SMASH.id === event.ability.guid) {
+      this.colossusSmashes += 1;
+      // Ignore the cast if the player didn't have shattered defenses.
+      if(!this.combatants.selected.hasBuff(SPELLS.SHATTERED_DEFENSES.id, event.timestamp - 1)) {
```
jjs451: Just out of curiosity, why `-1` on the timestamp?

Aelexe: If shattered defenses is applied using warbringer instead of colossus smash the buff occurs in the same frame, so I used a 1ms offset to avoid that.

jjs451: Oh gotcha, forgot that warbreaker applies the CS debuff

jjs451:
Looks good!  One style note: It's common in this project to use `SpellId !== SPELLS.SOME_SPELL.id` and return early instead of doing `SpellId === SPELLS.SOME_STUFF.id` and having to indent the important parts an extra level.  You can see an example of this in `Unholy/Modules/Items/Tier21_4p.js`.  It is a just style issue though so I'll merge as soon as Travis clears

src/Parser/Warrior/Arms/Modules/Core/ColossusSmash.js
```diff
@@ -0,0 +1,56 @@
...
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+
+import SpellUsable from '../Features/SpellUsable';
+
+class ColossusSmashAnalyzer extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    spellUsable: SpellUsable,
```
jjs451: Err, I spoke a little soon.  You should remove spellUsable since you don't use it here.

PR #1574, opened 2018-04-12
[Arms Warrior] Add wasted Executioner's Precision suggestion

Aelexe:
Adds a suggestion to avoid using execute if the target has two stack of executioner's precision and mortal strike is available.

I actually thought I'd coded it wrong, but it turns out my test subject might need a bit of practice on his execute phase execution.

image

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Core/Execute.js
+++ b/src/Parser/Warrior/Arms/Modules/Core/Execute.js
@@ -26,9 +26,13 @@ class ExecutionAnalyzer extends Analyzer {
 
     this.executes += 1;
     const enemy = this.enemies.getEntity(event);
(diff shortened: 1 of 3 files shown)


src/Parser/Warrior/Arms/Modules/Core/Execute.js
```diff
@@ -0,0 +1,57 @@
...
+  wastedExecutionersPrecisions = 0;
+
+  on_byPlayer_cast(event) {
+    if(SPELLS.EXECUTE.id !== event.ability.guid) {
+      return;
+    }
+
+    this.executes += 1;
+    const enemy = this.enemies.getEntity(event);
+    const executionersPrecision = enemy.getBuff(SPELLS.EXECUTIONERS_PRECISION.id, null, 0, 0, event.sourceID);
```
Gebuz: No need to include other parameters than the spellID, we don't track other players debuffs on enemies since we only parse events that include the player.

PR #1578, opened 2018-04-12
Added isOnGCD To Frost DK Abilities

scottymcraig:
Fixed Abilities.js for Frost DK to include isOnGCD.

### DIFF
--- a/src/Parser/DeathKnight/Frost/Modules/Features/Abilities.js
+++ b/src/Parser/DeathKnight/Frost/Modules/Features/Abilities.js
@@ -15,6 +15,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.SINDRAGOSAS_FURY_ARTIFACT,
         category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+        isOnGCD: true,
         cooldown: 300,
         castEfficiency: {
           suggestion: true,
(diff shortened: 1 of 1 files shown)


MartijnHols:
Hey @scottymcraig! I think this is your first PR to WoWAnalyzer, thanks a lot for your contribution and welcome on board! It looks like you're getting comfortable with the Frost Death Knight analyzer. As Khazak (@jjs451) has done some work on this recently, maybe you guys can team up? We're pretty active on Discord if you ever want to discuss anything.

If you're looking for more things to do, there is something related I can think of; any spells marked as being on the GCD in the abilities config can be removed from the `ABILITIES_ON_GCD` array of the [AlwaysBeCasting](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/DeathKnight/Frost/Modules/Features/AlwaysBeCasting.js#L12-L44) module to avoid duplication. The end goal would be to remove the array entirely (as defining GCDs like this is deprecated) and only rely on the `isOnGCD` property of the [Abilities](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/DeathKnight/Frost/Modules/Features/Abilities.js) config instead. I'd love to hear if you plan on picking this up!

Your changes should be live on wowanalyzer.com by now, please let us know if you need anything further.

PR #1583, opened 2018-04-13
[Arms Warrior] Add slam/whirlwind suggestions

Aelexe:
Add slam and whirlwind suggestions to not use them when better options are available. 

Adding a suggestion to not use them as filler during the execute phase still needs to be done, but I'm saving that for when I've created my health tracking module as a more graceful workaround for tracking enemy health through player cast events.

image
image


Gebuz:
Looks good! Just a question since I have no idea how to play the spec: Would you ever use Whirlwind over Mortal Strike for AOE? If so I would suggest that you also track the number of targets hit by WW, and only mark them as bad if it hit less than X targets. 

You can do this by saving the cast event (and the text to add if it should be marked as bad), count number of hits in on_byPlayer_damage and then on the next cast event (all hits should happen before that) you can check number of hits for the previous cast and mark it as bad if needed. You can then also set it as a bad cast if it only hit 1 target and playing without fervor of battle with the text "this WW only hit X targets, you should have cast slam instead" or similar.

Aelexe:
I didn't actually consider that scenario because generally sweeping strikes is taken, which makes mortal strike a better AOE move than whirlwind even without shattered defenses. The guide doesn't mention what to do if you don't have sweeping strikes, so I'll have to workout at what point whirlwind becomes more valuable and add a case for it.

Aelexe:
I've removed the whirlwind module for now whilst I consult a few sources on AOE rotations without sweeping strikes. The slam module should be fine as it is.

PR #1585, opened 2018-04-13
[Arms Warrior] Add Execute range tracker and refactor Rend module

Aelexe:
I've improved the execute range checking code and moved it into its own module. I'll likely move it to a shared folder once I begin working on the fury analyzer.

### DIFF
--- a/src/Parser/Warrior/Arms/Modules/Features/ExecuteRange.js
+++ b/src/Parser/Warrior/Arms/Modules/Features/ExecuteRange.js
@@ -38,7 +38,7 @@ class ExecuteRangeTracker extends Analyzer {
         timestamp: event.timestamp,
         inExecuteRange: inExecuteRange,
       });
-    } else if(!newEvent) {
+    } else {
       // If this event is the same as timestamp as the last event, modify the last record.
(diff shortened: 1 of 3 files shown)


src/Parser/Warrior/Arms/Modules/Features/ExecuteRange.js
```diff
@@ -0,0 +1,69 @@
...
+      return;
+    }
+
+    if(newEvent) {
+      // If this event is more recent than the last event, record it.
+      timeline.push({
+        timestamp: event.timestamp,
+        inExecuteRange: inExecuteRange,
+      });
+    } else if(!newEvent) {
```
Gebuz: the if statement is redundant here since you already have the else from the above if statement.

Aelexe:
I thought a module analyzes the entire fight before it's available to modules depending on it, but if that's not the case then I can probably refactor it as you've suggested.

Gebuz:
The analyzer runs through the log event by event and calls all active modules that are subscribed to those events. It's also why you rarely need to pass a timestamp when calling other modules, they are at the same point in the log. So yes, it is possible to do it like that :)

Issue #1525, opened 2018-03-25
Allow searching by region/realm/playername


MartijnHols:
Using the parses API endpoint we could make getting started easier.

https://www.warcraftlogs.com/v1/parses/character/Zerotorescue/AlAkir/EU?api_key=x

I'm not entirely sure if that's the best for our goal; WCL also has `/rankings/character/{characterName}/{serverName}/{serverRegion}` that might be more useful.

joshinat0r:
the majority of work for the frontend so far is done:

The general flow from the search to the overview is as following:

```
Player selects region/realm/charname
    call to blizzards-API
        when the char doesn't exist:
            alert
        when the char exists:
            save the path to the character-image in localStorage of the browser
            open the parse-overview
            fetch parses from WCL
                char doesn't exist on WCL:
                    message with link
```

the parse-overview then checks if the char exists on WCL, shows an error if the char doesn't exist at all or if the char has no parses for the selected raid tier. Checks also if the char has an entry in the localStore to display the image or makes a request to save it.

Saving the path of the image to the localStorage saves a few bnet-requests since there's no need to check again if the player exists (and the path of the image never changes for a character). 
Doing the bnet-call before the WCL one to check if a char has actually parses was done because the WCL-APIs response time can take up to 5-6 seconds (depending on the amount of parses). Not only makes the user experience smoother, but filters typing errors (or wrong region, thus saving WCL requests) and allows a more clean error message since the user is redirected to the parse-overview with reasons why the player doesn't have any parses and a link to get started with logging (instead of showing an alert).

**Stuff that's missing/not working**
- support for special-chars of our API
special chars cause issues with our API, which means parsing anything from the TW, KR, RU region or from any realm/character combo with anything beyond A-Z with is impossible to fetch right now. 
I've looked into the server folder but I have not a single idea what it's doing and what needs to be changed to support the characters without breaking anything.

- proper actorID for analyze link
not a major issue, but the parses-API doesn't include the actorID of the player for a log, just the name and the WoW-ID (the same ID thats partitially returned with the bnet-API for the char-image). But the combatants-info do not contain any of this information, just the actorID and the WCL-ID. I'm not quite sure why the actorID is in the URL since it works just fine without right now (the analyze-button submits a "+"-char instead of an actorID and does its job), but I guess its there for a reason. 

- character renames / realm transfer
[...]

PR #1597, opened 2018-04-19
[API] special character support?

joshinat0r:
This might already fix the special-character issue that #1525 needs.
Since calling our API from the app doesn't throw the "Corrupted response" error but "Character not found", which might be because our API takes the correctly encoded URL, decodes it and sends that request to WCL ( [Express does it apparently by default](https://github.com/expressjs/express/issues/1479)). But WCL requires the name and real uriencoded, hence the "Character not found" (the same error we get when we try to call the WCL URL manually with special-characters).

Might work, might not work. But it definitely shouldn't break anything since encodeURI doesn't touch alphanumerical values and `/`
So the normal `reports/XXXXXXXXX` endpoint shouldn't call a different URL.

### DIFF
--- a/server/api.js
+++ b/server/api.js
@@ -69,7 +69,7 @@ class ApiRequestHandler {
   }
   async fetchFromWcl(cachedWclApiResponse) {
     const query = Object.assign({}, this.req.query, { api_key: this.apiKey });
-    const path = `v1/${this.req.params[0]}?${querystring.stringify(query)}`;
+    const path = `v1/${encodeURI(this.req.params[0])}?${querystring.stringify(query)}`;
     console.log('GET', path);
     try {
       const wclStart = Date.now();


MartijnHols:
Hey thanks for looking into this. I think we should look for another solution for this as I don't think we should encode the entire path for all API requests we send to WCL. This seems very fragile for various unrelated calls and can not be influenced by the caller of the API. If WCL requires the values in the URL to be encoded, and Express auto decodes it, would it be a solution to double-encode the character and realm names in the initial request?

joshinat0r:
Encoding the parameters twice is so simple that it actually works.
Parsing from US & EU (with the exception of russian) works now, the regions TW/KR are not working at all (they cause a timeout, can't tell if it's WCL or our API). I guess it's due to the charset, not quite sure whats causing the issue but I'll take another look at it sometime this weekend.

I'll PR the char-parsing after I took a look at that, even if RU/TW/KR are not working (and removing those regions from the selection). Might draw attention to some devs out there that have experience with those languages and their encoding.

Closing this one.

MartijnHols:
Actually if you're right that the issue is Express auto-decoding, your original solution would be closer to the best solution I can think of which would be disabling the auto-decoding of parameters for this endpoint. This would fix the underlying problem, while double-encoding is just a quick hack mostly (but it has the advantage of not risking breaking anything else and being isolated to only the trouble area).

My issue with the approach is that it can break paths in unexpected ways if it interacts with the URL differently from Express, which I reckon it does. I think Express decodes the URL part by part (so it first splits by /, then decodes each part separately) and there would be no guarantees your approach does the reverse properly for all possible cases. Rebuilding/building a reverse implementation of Express' deserializer and auto-decoder would also seem fragile, but if there were some setting to disable this behavior for this endpoint or a property that gets the raw url, that would be very nice.

PR #1608, opened 2018-04-23
Relic stats update

yajinni:
Old look:
image

New look:

image


Pewtro:
Just needs to fix the formatPercentages being unused, and this should be good to go

Issue #1601, opened 2018-04-21
Add back the Mana usage chart


MartijnHols:
In https://github.com/WoWAnalyzer/WoWAnalyzer/commit/16b3f3ac4a66be892dd31052f13989d963e4f877 all remaining charts using the deprecated Chartist library were removed so the dependency on this package could finally be removed.

In order to provide this functionality again, it needs to be re-implemented using the Chart.js library instead. The differences in chart libraries shouldn't be big.
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=57402077)](https://www.bountysource.com/issues/57402077-add-back-the-mana-usage-chart?utm_source=57402077&utm_medium=shield&utm_campaign=ISSUE_BADGE)

YogeshSharma0201:
This issue was resolved in #1633.

PR #1625, opened 2018-04-26
Reimplement Mana pool graph with chart.js

### LINKED ISSUE #1600: Add back the Mana pool chart
In https://github.com/WoWAnalyzer/WoWAnalyzer/commit/16b3f3ac4a66be892dd31052f13989d963e4f877 all remaining charts using the deprecated Chartist library were removed so the dependency on this package could finally be removed.

In order to provide this functionality again, it needs to be re-implemented using the Chart.js library instead. The differences in chart libraries shouldn't be big.
[Bountysource](https://www.bountysource.com/issues/57402059-add-back-the-mana-pool-chart?utm_source=57402059&utm_medium=shield&utm_campaign=ISSUE_BADGE)

YogeshSharma0201:
I have implemented all the basic functionality. The legend labels and x-axis scales look a little different. Please take a look at it and let me know if they should be changed.

screenshot:
mana

closes #1600


src/Main/ManaStyles.js
```diff
@@ -0,0 +1,118 @@
+
+export const ManaStyles = {
```
MartijnHols: When there's only one export in a file, we prefer making it the default export.

src/Main/ManaLevelGraph.js
```diff
@@ -0,0 +1,226 @@
...
+        animationDuration: 0,
+      },
+      responsiveAnimationDuration: 0,
+      tooltips: {
+        enabled: false,
+      },
+      legend: ManaStyles.legend,
+    };
+
+    Chart.plugins.register({
```
MartijnHols: A better place for this would probably be in the constructor of this class, and even then we risk registering the same plugin multiple times. Also having this in the mana-specific class would encourage copy-pasting this to other modules. If you have any ideas how we could improve that, it would be much appreciated but as long as it doesn't actually lead to problems and so long as it hasn't been duplicated yet it's ok to leave it as is.

src/Main/ManaLevelGraph.js
```diff
@@ -0,0 +1,226 @@
...
+            max: 100,
+            stepSize: 25,
+            fontSize: 14,
+          },
+        }],
+        xAxes: [{
+          gridLines: gridLines,
+          ticks: {
+            callback: (value, index, values) => {
+              if(value%30 === 0)
```
MartijnHols: We prefer using curly brackets around all multi-line blocks.

MartijnHols:
Hey @YogeshSharma0201! I believe this was your first PR to WoWAnalyzer, thanks a lot for your contribution and welcome on board! It looks like you're getting comfortable with the project and the charts in particular. If you're looking for something new to work on I left some comments above that you could look into, and there are also some other similar issues that you could also consider implementing #1601, #1602 and #1603. Like this one they have a BountySource bounty assigned as well. We're pretty active on Discord if you ever want to discuss anything.

Your changes should be live on https://bfa.wowanalyzer.com within a few minutes of this comment, please let us know if you need anything further.

YogeshSharma0201:
@MartijnHols thanks for such detailed review. I would surely work on the other similar issues and will make the changes you mentioned above in my next PR.

PR #1626, opened 2018-04-26
[Blood Death Knight][BfA] some stuff

joshinat0r:
- removes current Ossuary
- adds Hemostasis (old legendary based on amount of targets hit by BB, no interaction with DRW right now, might be a bug, might be intentional)
- removes estimated absorb from Boneshield
- updated talents-file (old spells still included for changelogs)


src/Parser/DeathKnight/Blood/Modules/Features/Checklist.js
```diff
@@ -140,16 +138,12 @@ class Checklist extends CoreChecklist {
           new Requirement({
             name: <React.Fragment><SpellLink id={SPELLS.MARK_OF_BLOOD_TALENT.id}/> Uptime</React.Fragment>,
             check: () => this.markOfBloodUptime.uptimeSuggestionThresholds,
+            when: this.combatants.selected.hasTalent(SPELLS.MARK_OF_BLOOD_TALENT.id),
```
Gebuz: Shouldn't this change be made to live instead?

Gebuz: I clearly read the PRs in wrong order. Should be able to leave this out however, since zero periodically merges master into BFA.

joshinat0r:
Was the fastest way for now to start adding BfA stuff since generating a new one right now would cause issues with the changelog (spellLinks wouldn't find old talents).

I'll remove the MoB fix later today

PR #1630, opened 2018-04-28
[MW Monk] Checklist Updates 

anom0ly:
Adjusted some of the Checklist items for readability based on feedback.


src/Parser/Monk/Mistweaver/Modules/Spells/SoothingMist.js
```diff
@@ -33,18 +33,17 @@ class SoothingMist extends Analyzer {
   }
 
   get soomTicksPerDuration() {
-    return (this.soomTicks * 2 / this.owner.fightDuration * 1000) || 0;
+    const soomTicks = (this.soomTicks * 2 / this.owner.fightDuration * 1000) || 0;
+    if (soomTicks >= 1.5) {
+      return true;
+    }
```
MartijnHols: This doesn't have a consistent return path; it should return `false` when not passing the if. How about `return soomTicks >= 1.5;`?

anom0ly: Thanks Zero. I updated this along with the below suggestion. I realized swapping to a boolean, I didn't also update the suggestion construction. Should be all fixed up.

PR #1631, opened 2018-04-28
[BFA] Bloodworm Talent Analysis

yajinni:
image


Gebuz:
See #1634 for comments. There is a mergeconflict that needs to be solved but otherwise it looks good to go.

yajinni:
Did not know that live changes got merged into the BFA branch. Wont need to duplicate my pushes then. Thanks. Closing this.

PR #1633, opened 2018-04-28
Reimplement ManaUsage graph using chart.js

### LINKED ISSUE #1601: Add back the Mana usage chart

YogeshSharma0201:
I have implemented mana usage graph and also have made some changes as was asked in  #1625. Now extra bosses have different colors, width of the lines is reduced and width of the legend boxes is reduced.

closes #1601


MartijnHols:
When making a pull request please try to always include a screenshot. If you added something visual show what is new. If you changed a calculation screenshot the old value versus new. If you added a check similar to that of legendaries/enchants/deaths please include a link to an example report in the PR.

In this case I'd prefer if you used this report since the mana graphs have pretty much everything: https://wowanalyzer.com/report/hNqbFwd7Mx3G1KnZ/18-Mythic+Antoran+High+Command+-+Kill+(6:51)/7-Taffly/mana

src/Main/ManaLevelGraph.js
```diff
@@ -5,7 +5,7 @@ import {Line} from 'react-chartjs-2';
 
 import fetchWcl from 'common/fetchWcl';
 
-import {ManaStyles} from './ManaStyles.js';
+import ManaStyles from 'Main/ManaStyles.js';
```
MartijnHols: .js can be omitted.

src/Main/ManaUsageGraph.js
```diff
@@ -0,0 +1,167 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Line} from 'react-chartjs-2';
+
+import { formatThousands } from 'common/format';
+import ManaStyles from 'Main/ManaStyles.js';
+
+
+const formatDuration = (duration) => {
```
MartijnHols: Using format functions from `common/format` is preferred.

PR #1634, opened 2018-04-28
Blooddrinker Updates

yajinni:
Added dps/hps stuff to the tooltip and cleaned up the code by putting in early returns for the events.

image

### DIFF
--- a/src/Parser/DeathKnight/Blood/Modules/Talents/Bloodworms.js
+++ b/src/Parser/DeathKnight/Blood/Modules/Talents/Bloodworms.js
@@ -1,0 +1,97 @@
+import React from 'react';
+import Analyzer from 'Parser/Core/Analyzer';
+import SPELLS from 'common/SPELLS/index';
+import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SpellIcon from 'common/SpellIcon';
+import { formatThousands } from 'common/format';
+
+//Worms last 15 sec. But sometimes lag and such makes them expire a little bit early.
+const WORMLIFESPAN = 14900;
+class Bloodworms extends Analyzer {
+  static dependencies = {
+    abilityTracker: AbilityTracker,
+    combatants: Combatants,
+  };
+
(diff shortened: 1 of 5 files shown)


src/Parser/DeathKnight/Blood/Modules/Talents/Bloodworms.js
```diff
@@ -0,0 +1,97 @@
...
+  wormID=0;
+
+  bloodworm = [];
+
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.BLOODWORMS_TALENT.id);
+  }
+
+  poppedWorms(bloodworm) {
```
Gebuz: no need to have `bloodworm` as a parameter here since it can already access it.

src/Parser/DeathKnight/Blood/Modules/Talents/Bloodworms.js
```diff
@@ -0,0 +1,97 @@
...
+    this.totalDamage += event.amount + (event.absorbed || 0);
+  }
+
+
+  on_byPlayerPet_instakill(event) {
+    if (event.ability.guid !== SPELLS.BLOODWORM_DEATH.id) {
+      return;
+    }
+    let index = -1;
+    this.bloodworm.forEach((e, i) => {
```
Gebuz: You could use .find instead here to optimize it slightly.

src/Parser/DeathKnight/Blood/Modules/Talents/Bloodworms.js
```diff
@@ -0,0 +1,97 @@
...
+  statistic() {
+    return (
+
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.BLOODWORMS_TALENT.id} />}
+        value={this.owner.formatItemHealingDone(this.totalHealing)}
+        label="Bloodworm Stats"
+        tooltip={`<strong>Damage:</strong> ${formatThousands(this.totalDamage)} / ${this.owner.formatItemDamageDone(this.totalDamage)}<br>
+        <strong>Number of worms summoned:</strong> ${this.totalSummons}<br>
+        <strong># of worms popped early:</strong> ${this.poppedWorms(this.bloodworm)}
```
Gebuz: Should probably write "Number" instead of "#" here to make it consistent with the line above.

PR #1635, opened 2018-04-30
[Blood DK][BfA] updated BS & FB-Modules

joshinat0r:
The expandable on FB is maybe unnecessary since it relies on the Boneshields stacks.
But I think it's usefull to have it for both, doesn't force the user to check the number in two places.

I'll make a PR to add this on live aswell, just needs some minor changes.

bs
bs_big
fb

### DIFF
--- a/src/Parser/DeathKnight/Blood/Modules/Features/BoneShieldStacksBySeconds.js
+++ b/src/Parser/DeathKnight/Blood/Modules/Features/BoneShieldStacksBySeconds.js
@@ -1,4 +1,3 @@
-import React from 'react';
 import Analyzer from 'Parser/Core/Analyzer';
 import SPELLS from 'common/SPELLS';
 
@@ -45,7 +44,7 @@ class BoneShieldStacksBySeconds extends Analyzer {
   }
 
   get boneShieldStacksBySeconds() {
-    return this.boneShield
+    return this.boneShield;
   }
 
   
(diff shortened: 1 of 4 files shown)


src/Parser/DeathKnight/Blood/Modules/Talents/FoulBulwark.js
```diff
@@ -0,0 +1,69 @@
...
+            <tr>
+              <th>HP-bonus</th>
+              <th>Time (s)</th>
+              <th>Time (%)</th>
+            </tr>
+          </thead>
+          <tbody>
+            {Array.from({length: this.maxStacks + 1}, (x, i) => i).map((e, i) =>
+              <tr key={i}>
+                <th>{formatPercentage(i * HP_PER_BONE_SHIELD_STACK)}%</th>
```
Gebuz: Could round these to 0 decimals since it's always .00

src/Parser/DeathKnight/Blood/Modules/Features/BoneShieldStacksBySeconds.js
```diff
@@ -0,0 +1,84 @@
+import Analyzer from 'Parser/Core/Analyzer';
+import SPELLS from 'common/SPELLS';
+
+/* 
+  boneShieldStacksBySeconds() returns an array with the amount of stacks with an entry for each second
```
Gebuz: I would suggest using the exact timestamps to sum time between buff stack changes instead of only having one entry per second. It would make the uptimes more accurate.

joshinat0r:
I made an extra PR for live, the only real difference is that FB is in BfA only 1% per stack.
But already stated in the live PR that I should've made the live version first and then change it once Zero merged live into BfA.
I'll close this one and change the stuff for the live-version (and change the BfA version later on).

PR #1639, opened 2018-05-02
Bfa Tombstone

yajinni:
Added in the tombstone talent
image


yajinni:
Made all the changes you recommended

PR #1644, opened 2018-05-02
[BFA] Build 26530 - Berserking, Resto Shaman changes & Static Charge Talent

niseko:
Added downpour mastery scaling since it got fixed this build.

Adjusted berserking to be 20%

Implemented the Static Charge cooldown reduction which depends on how many targets you hit, and added it + lightning surge totem to the other 2 specs as it functions the same.
image


src/Parser/Shaman/Shared/Talents/StaticCharge.js
```diff
@@ -0,0 +1,35 @@
...
+const stunReduction = 5000;
+
+class StaticCharge extends Analyzer {
+  static dependencies = {
+    spellUsable: SpellUsable,
+    combatants: Combatants,
+  };
+
+  on_initialized() {
+    this.hasStaticCharge = this.combatants.selected.hasTalent(SPELLS.STATIC_CHARGE_TALENT.id);
```
Gebuz: You can set `this.active` here instead of `this.hasStaticCharge`, that way you don't need to check for the talent in the applydebuff below.

src/Parser/Shaman/Restoration/Constants.js
```diff
@@ -31,7 +31,7 @@ export const ABILITIES_AFFECTED_BY_MASTERY = [
   SPELLS.HEALING_RAIN_HEAL.id,
   SPELLS.UNLEASH_LIFE_TALENT.id,
   SPELLS.EARTH_SHIELD_HEAL.id,
-  //SPELLS.DOWNPOUR_TALENT.id, Not yet, but expected to get fixed
+  SPELLS.DOWNPOUR_TALENT.id,
   SPELLS.DOWNPOUR.id,
```
Gebuz: Is it intended that there are both `SPELLS.DOWNPOUR_TALENT.id` and `SPELLS.DOWNPOUR.id` here?

niseko: yes, Downpour_Talent is the newly added talent which is basically the artifact active, Downpour is T21 4 piece. i have no clue why they have the same name

PR #1647, opened 2018-05-02
[Core] Show buffs in the timeline tab

### LINKED ISSUE #792: Show important buffs in the timeline tab
It would be nice to show buff windows in the timeline tab.

Buffs like Ilterendi's (15% healing increase for 5 sec), Vulnerable (30% damage for 7 sec) etc

niseko:
This displays buffs and (your) debuffs on the timeline.
image
Debuff example:
image

To add a (de)buff to the timeline, its ID needs to be assigned to a spell with buffSpellId in the spellbook. Only added resto shaman ones (well, and velens) as i don't know other classes well enough.

I'm not great with styling so it could probably look better or have better colors, using transparency had a better look but you'd have different looking bars depending on if the spell is still on CD or not (since it has a different background).

closes #792


niseko:
image
like this? i think it looks odd and most stuff will start together with the CD

making it thinner might be too subtle? 
image

MartijnHols:
Thinner makes it harder to read, I think a bar next to it looks best but I don’t think a row should use space that isn’t the row’s. How about making the cd duration and icon smaller when there’s a buff bar?

niseko:
Couldn't really get the size changes working correctly, so i changed it in a temporary way to see how it'd look like
image

Also pushed some slight code change which hopefully makes it easier to read.

With earth shield on beta i've been thinking if there's a way to visually include the amount of stacks a buff has, might be interesting for some buffs but seems like it would be hard to make it readable.

MartijnHols:
I made a mockup of something like that for Atonement some time ago, maybe it helps:

image

niseko:
Apart from that I can't decide on visuals, so that the PR still looks like the 1st picture I posted, it should be ready. 

I actually like that atonement bar idea and might experiment with something like that for tidal waves, but separate from this.

MartijnHols:
Hmmm not sure if I was unclear. I think the last screenshot you posted with the buff bars within the row constrains and resizing the spell icon looks really neat 👍  If you could push that to the PR I'll start a review.

niseko:
Yeah, I didn't quite catch that.
I got it to work now so it only resizes if there's a buff for that spell (the screenshot before was doctored together).
image

I think it looks a bit odd when a spell without a buff is right next to those, but maybe that's just my eyes.
image

Should be ready for a review now, though!

Gebuz:
You could try to make the gray bars always have full height (and have the buff bar overlap) and only scale the icon instead. That way you avoid the different size gaps in your second img.

niseko:
That gets ugly in situations where the buff isn't in sync with the spell, like lava surge, or spells that have travel time before they activate the buff / debuff like judgement.
image

MartijnHols:
Would be nice if we could figure out a way to include buffs on other plays for things like Riptide, totem-like behavior for Healing Stream Totem, Healing Rain and Cloudburst totem.

We could also consider showing buff-consumers like Tidal Waves.

niseko:
Buff consumers like tidal waves probably need their own row as you have multiple generators and spenders, and also have multiple stacks. For Totems i'd probably need to throw fabricated buff events with a set duration. But yeah Cloudburst is why i initially wanted to do this since what you do during it is important.

PR #1649, opened 2018-05-02
Bfa rapid decomp

yajinni:
Added rapid decomp talent
image


src/Parser/DeathKnight/Blood/Modules/Talents/RapidDecomposition.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+    this.totalDamage = this.bpDamage + this.dndDamage;
+  }
+
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.RAPID_DECOMPOSITION_TALENT.id} />}
+        value={`${this.owner.formatItemDamageDone(this.totalDamage)}`}
+        label='Damage Increase'
```
Gebuz: How about calling it Rapid Decomposition instead? I think the "Damage Increase" part is obvious enough from the fact that it is DPS. Same for the tooltips below, could remove the "Damage Increase" and maybe add "Breakdown:" in a line at the top.

yajinni: Changed it. Decided the world break down wasnt needed.

PR #1651, opened 2018-05-02
Will of the Necropolis

yajinni:
image

### DIFF
--- a/src/Parser/DeathKnight/Blood/Modules/Talents/WillOfTheNecropolis.js
+++ b/src/Parser/DeathKnight/Blood/Modules/Talents/WillOfTheNecropolis.js
@@ -1,0 +1,73 @@
+import React from 'react';
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SPELLS from 'common/SPELLS/index';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber } from "common/format";
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+const MINIMUMABSORTBTHRESHOLD = 0.05;
+
+class WillOfTheNecropolis extends Analyzer {
+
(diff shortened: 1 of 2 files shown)


src/Parser/DeathKnight/Blood/Modules/Talents/WillOfTheNecropolis.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import SPELLS from 'common/SPELLS/index';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber } from "common/format";
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+const MINIMUMABSORTBTHRESHOLD = 0.05;
```
Gebuz: looks like an extra T managed to sneak it's way in there :P also please use underscores to split the words in constants.

src/Parser/DeathKnight/Blood/Modules/Talents/WillOfTheNecropolis.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  on_toPlayer_damage(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== this.spellDamageId || this.nextEvent === false) {
+      return;
+    }
+    this.nextEvent = false;
+    this.playerHealth = event.maxHitPoints;
+    this.absorbToHeathPercent = this.currentWotnAbsorbed / this.playerHealth;
```
Gebuz: You can make this and playerHealth above local constants since they aren't used anywhere else, no need to define them as class variables. (also there's missing an L in absorbToHeathPercent)
```
const playerHealth = event.maxHitPoints;
const absorbToHeathPercent = this.currentWotnAbsorbed / playerHealth;
```

PR #1653, opened 2018-05-03
DH Pain stat and resource tracker

yajinni:
The pain chart/resource tab was using an depreciated system.
Updated it to current standards. If i need to include pics let me know.


yajinni:
added update for painbringer, didnt mean to make it one big pr though...
image

yajinni:
Side note, painbringer is going away in BFA, so not worth it to keep working on it and showing mitigated damage, so probably my last update to that module.

src/Parser/DemonHunter/Vengeance/CombatLogParser.js
```diff
@@ -40,6 +39,9 @@ class CombatLogParser extends CoreCombatLogParser {
     alwaysBeCasting: AlwaysBeCasting,
     abilities: Abilities,
     cooldownThroughputTracker: CooldownThroughputTracker,
+    //Resource Tracker
+    painTracker :PainTracker,
```
MartijnHols: You misplaced your colon

PR #1655, opened 2018-05-03
DH Soul Barrier Talent

yajinni:
image

### DIFF
--- a/src/Parser/DemonHunter/Vengeance/Modules/Spells/SoulBarrier.js
+++ b/src/Parser/DemonHunter/Vengeance/Modules/Spells/SoulBarrier.js
@@ -59,11 +59,11 @@ class SoulBarrier extends Analyzer {
 
   get suggestionThresholdsEfficiency() {
     return {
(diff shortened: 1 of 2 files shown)


yajinni:
HPS is shown. I added that as a way for future proofing it because in BFA it will be variable. They are getting rid of its minium shield and letting it drop.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SoulBarrier.js
```diff
@@ -0,0 +1,99 @@
...
+    damageTracker: DamageTracker,
+    enemies: Enemies,
+  };
+
+  casts = 0;
+  totalAbsorbed = 0;
+  buffApplied = 0;
+  buffRemoved = 0;
+  buffLength = 0;
+  avgBuffLength = 0;
```
Gebuz: no need to define this here. (same with buffLength & buffRemoved)

PR #1665, opened 2018-05-04
Demon Spikes Code rework

yajinni:
image


src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.DEMON_SPIKES.id} />}
+        value={`${this.mitigatedUptime}%`}
+        label="Hits Mitigated by Demon Spikes"
+        tooltip={`Demon Spikes usage breakdown:
+          <ul>
+          <li>You were hit <b>${this.hitsWithDS}</b> times with your Demon Spikes buff.</li>
+          <li>You were hit <b>${this.hitsWithoutDS}</b> times <b><i>without</i></b> your Demon Spikes buff.</li>
+          <li>You were hit <b>${this.hitsWithDSOnCD}</b> times <b><i>with</i></b> Demon Spikes avalible for use but not used.</li>
```
Gebuz: You're counting hits with DS on cd, but text here says the opposite? Anyway I don't think this is really a needed stat to display considering you'll often be holding your AM for big damage intakes.

src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
```diff
@@ -0,0 +1,68 @@
...
+          </ul>
+          <b>${this.mitigatedUptime}%</b> of physical attacks were mitigated with Demon Spikes, and your overall uptime was <b>${formatPercentage(demonSpikesUptimePercentage)}%</b>.`}
```
Gebuz: Don't really need to have the mitigatedUptime here since it is already the primary value.

src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
yajinni: Whoops named the var wrong. 
As for AM its similar to bears iron fur. You obviously want to do both. Save it for big physical hits and also use it as much as possible to smooth damage out. The DH is pretty barebones so this is something to get them started. Might make updates to it down the line. But without fight specific stats hard to know if you used it at big AM moments or not. For now it simply encourages you to use it more without trying to force you to use it on CD.

src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
yajinni: Yea your right, took it out.

PR #1666, opened 2018-05-04
Updated DH abilities

yajinni:
Added some spells and updated gcd/cd times.


src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import SPELLS from 'common/SPELLS';
-import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS/index';
+import ITEMS from 'common/ITEMS/index';
```
Gebuz: Shouldn't need /index here.

Issue #1662, opened 2018-05-04
[Discipline][BFA Prepatch] Contrition Talent


rp4rk:
**Description:**
The Contrition talent is a new addition in Battle for Azeroth. The talent causes each Penance bolt that directly heals a friendly player to heal each player with Atonement for 20% spellpower. Beyond displaying the basic healing breakdown, which is quite misleading, it would be more valuable to display the real increase, as the heal incurs a large opportunity cost.

The "real" increase from Contrition will be worth the value of the Penance heal and Contrition heal, subtracting the value of an offensive Penance. 

**Ability Link:** https://beta.wowdb.com/spells/197419-contrition

**Acceptance Criteria:**
- [x] The module displays how much healing gained from Contrition.
- [x] The module accounts for the lack of an offensive Penance.
- [x] The module can approximate a loss of damage from each lost offensive Penance.

rp4rk:
Merged.

PR #1671, opened 2018-05-06
[Discipline][BFA Prepatch] Contrition Talent

rp4rk:
Relevant Issue: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/1662

---

**Description**
The Contrition talent is a new addition in Battle for Azeroth. The talent causes each Penance bolt that directly heals a friendly player to heal each player with Atonement for 20% spellpower. Beyond displaying the basic healing breakdown, which is quite misleading, it would be more valuable to display the real increase, as the heal incurs a large opportunity cost.

### DIFF
--- a/src/Parser/Priest/Discipline/Modules/Spells/Contrition.js
+++ b/src/Parser/Priest/Discipline/Modules/Spells/Contrition.js
@@ -22,7 +22,6 @@ class Contrition extends Analyzer {
 
   healing = 0;
   damagePenalty = 0;
-  player = null;
   penanceBoltEstimation;
 
   on_initialized() {
@@ -48,11 +47,11 @@ class Contrition extends Analyzer {
     this.healing += event.amount;
 
     // Get an estimated amount of damage and healing for a bolt
-    const { damage, healing } = this.penanceBoltEstimation();
+    const { boltDamage, boltHealing } = this.penanceBoltEstimation();
 
     // Calculate the difference between contrition and an offensive penance
     if (event.ability.guid === SPELLS.CONTRITION_HEAL.id) {
-      const overhealing = healing - event.overheal;
+      const overhealing = boltHealing - event.overheal;
       const estimatedDifference = event.amount - (overhealing > 0 ? overhealing : 0);
 
       this.healing -= estimatedDifference;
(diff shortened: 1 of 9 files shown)


MartijnHols:
What does WIP mean?

MartijnHols:
Please merge from the bfa branch so the files changed are accurate.

src/Parser/Priest/Discipline/Modules/Spells/Contrition.js
```diff
@@ -0,0 +1,87 @@
...
+
+    // Add the healing to our count
+    this.healing += event.amount;
+
+    // Get an estimated amount of damage and healing for a bolt
+    const { damage, healing } = this.penanceBoltEstimation();
+
+    // Calculate the difference between contrition and an offensive penance
+    if (event.ability.guid === SPELLS.CONTRITION_HEAL.id) {
+      const overhealing = healing - event.overheal;
```
Gebuz: Shouldn't overhealing only be considered if `healing > event.amount`, and in that case just set healing = event.amount?
for instance, if contrition heals for 50 and overheals for 50 and atonement would have healed for 75 (50 + 25 overheal), then it should just subtract the full value of heal for a net 0 gain from speccing contrition right?

`this.healing -= estimatedDifference;` unless I misunderstood it, you should probably rename estimatedDifference to atonmentHeal or something like that. I assume that's what it's meant to be atleast since you add the full contrition heal in l48, it would make sense to subtract the full atonement heal instead of the difference.

src/Parser/Priest/Discipline/Modules/Spells/Contrition.js
```diff
@@ -0,0 +1,87 @@
...
+  on_byPlayer_heal(event) {
+    if (event.ability.guid !== SPELLS.CONTRITION_HEAL.id && event.ability.guid !== SPELLS.PENANCE_HEAL.id) {
+      return;
+    }
+
+    // Add the healing to our count
+    this.healing += event.amount;
+
+    // Get an estimated amount of damage and healing for a bolt
+    const { damage, healing } = this.penanceBoltEstimation();
```
Gebuz: iirc Power of the Dark side was added baseline recently on Beta, so might need to account for that.

rp4rk: Yeah, need to decide on this at some stage. Thinking just adding a buff tracker.

rp4rk:
`/report/xnT6fv4kDLd3y7Xg/16-Heroic+G'huun+-+Wipe+9+(4:27)/23-Jaayyddee`

Here is an example log where this now hits the exact number provided by penance with no outside modifiers. Cheers!

Gebuz (approved):
Looks good, thanks!

PR #1673, opened 2018-05-07
[BfA][Windwalker] Further updating Blackout Kick module and SPELLS.MONK

Juko8:
Adding some energize ID's from talents to prevent crashing when loading chi tab. 

Blackout Kick module will have more work done before 8.0 goes live. Right now the statistic box is just a copy from Crusaders Might 

image


Juko8:
billede

Looks like this now

PR #1682, opened 2018-05-10
[Core][Protection Warrior] Death Log

joshinat0r:
**Prot**
Updated/recategorized protection warriors cooldowns.
Added cooldown-reductions from
- Thundergod's Vigor for Demo Shout CDR
- Angermanagement for Battle Cry, Demo Shout, Shield Wall, Last Stand
- T21 2p for Shield Slam

to allow proper cooldown tracking for the death recap.
Should've made that in a seperate RP but stumbled across the CD-tracking issue while working on the death recap and added that on the same branch.

items
angermanagement

**Death Tracker**
Adds a new tab if the player died during the fight. The tab contains all deaths of this player, with a detailed breakdown of the last 10sec of each death. Tracks the amounts (absorbs/overhealing) of events, a HP bar to visualize it and a list of defensive buffs that were up right next to the personal CDs (and if they were available or on cooldown).
Added for this a new spell category called `SEMI_DEFENSIVE` to display cooldowns that can be used offensively and defensively.
Slider on the top filters the events based on the amount relative to your HP (to remove minor heal/damage events like HoTs).
For now only active on Blood-DKs and Prot-Warriors, I'll add other tanks later on (function isn't limited to tanks tho, so adding other specs is possible)

death_recap


Gebuz:
May I suggest using the buffSpellId prop in abilities (added in #1647) for the class specific buffs? (purg, boneshield, etc.).

Gebuz:
The prot warrior stuff looks good, other than the time format being inconsistent in the wording (Thundergod's says 1:28 minutes, Anger management says 9:04 seconds)

joshinat0r:
Can do, the CDR of AM is actually 9:04 minutes, so its not inconsistent just a typo
/shrug

Where should I add the general externals (like IB / BoP)? Just leave them in the `DEFENSIVE_BUFFS`-file?

Gebuz:
Yeah I think leaving externals in that file is fine.

joshinat0r:
Is that a good idea? Just started removing the class specific buffs from the exta file and added it to the spellbook.
Issue is now that when I want to add `Bone Shield` I'd need to change the category of Marrowrend to `DEFENSIVE` or `SEMI_DEFENSIVE` because I only want to show those categories. But Marrowrend shouldn't be grouped with `Defensive Cooldown` in the ability tab.
Otherwise all spells with a `buffSpellId` would appear as active buff, which may or may not be an issue right now. But when other specs want to track some 4sec DPS-buffs in the timeline they'd be shown in the death recap (when we add the death recap for other specs that is).

The only options I can see are 
a) make a new prop in the spellbook just for the death recap
b) add new spell categories (like rotational-defensive)
c) keep the spec specifc buffs in the `DEFENSIVE_BUFFS`-file

Gebuz:
Hmm yeah I might not have thought that through, reason I suggested it is because some abilities have a buff ID that is different from the cast ID. Maybe add a check for `buffSpellId` and if it's present use that instead? But yeah buffs that aren't connected directly to an active spell can't really be added in abilities. Aslong as there isn't too many of those it should be fine to keep them in `DEFENSIVE_BUFFS` I think. If it gets out of hand each spec could extend it's own version and add the extra IDs.

Assuming most classes have relatively complete spellbooks at this point, it probably wouldn't hurt to just enable the deathlog baseline.

joshinat0r:
It's not like a lot of spells need to be added to the `DEFENSIVE_BUFFS` should almost be exclusive to tank specs, general defensive buffs. Most defensive cooldowns use the same id for buff and spell (at least afaik). So if a spell is already a defensive cooldown and is using the same ID its already added to the buffs for the recap.

Just enabling the recap by default might cause confusion if some cooldowns have different IDs for the buff. I'd rather avoid that have each spec checked first (like eg. rogue's feint, warriors die by sword are right now not implemented as defensive). I'd go through them spec by spec by myself since some specs are not really maintained.

So, best way to implemtent them for now is: use `DEFENSIVE`/`SEMI_DEFENSIVE` abilities with a `buffSpellId` by default and add the other buffs to `DEFENSIVE_BUFFS` and keep in mind that we could split it if the file gets too large?

src/Main/DeathRecapTracker.js
```diff
@@ -0,0 +1,94 @@
...
+  lastBuffs = [];
+
+  static dependencies = {
+    combatants: Combatants,
+    abilities: Abilities,
+    spellUsable: SpellUsable,
+  };
+
+  on_initialized() {
+    this.cooldowns = this.abilities.abilities.filter(e =>
```
rp4rk: This would read so much better if the logic for the filter was extracted, e.g.

```js
const hasCooldown = ability =>  (ability.category === Abilities.SPELL_CATEGORIES.DEFENSIVE || ability.category === Abilities.SPELL_CATEGORIES.SEMI_DEFENSIVE) && ability.enabled === true;

this.cooldowns = this.abilities.abilities.filter(hasCooldown);
```

src/Main/DeathRecapTracker.js
```diff
@@ -0,0 +1,94 @@
...
+    spellUsable: SpellUsable,
+  };
+
+  on_initialized() {
+    this.cooldowns = this.abilities.abilities.filter(e => 
+      (e.category === Abilities.SPELL_CATEGORIES.DEFENSIVE || e.category === Abilities.SPELL_CATEGORIES.SEMI_DEFENSIVE) &&
+      e.enabled === true
+    );
+    //add additional defensive buffs/debuffs to common/DEFENSIVE_BUFFS
+    this.buffs = DEFENSIVE_BUFFS.concat(this.cooldowns);
```
rp4rk: Might be more future proof if you used spread here. Not a blocker though.

src/Main/DeathRecapTracker.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  addEvent(event) {
+    const extendedEvent = event;
```
rp4rk: This would still mutate the event being passed in, please use:
```js
const extendedEvent = { ...event };
```
or another similar shallow/deep clone method.

src/Main/DeathRecapTracker.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  on_toPlayer_death(event) {
+    this.addEvent(event);
+    this.deaths.push(event.timestamp);
+  }
+
+  get secondsBeforeDeath() {
+    const deaths = new Array(this.deaths.length);
+    this.deaths.forEach((deathtime, index) => {
```
rp4rk: Please use a `map` here, the input and output of the `deaths` mutator are 1:1.

src/Main/DeathRecapTracker.js
```diff
@@ -0,0 +1,94 @@
...
+        deathtime: deathtime,
+        events: this.events,
+        open: false,
+      };
+    });
+    return deaths;
+  }
+
+  tab() {
+    if (this.deaths.length > 0) {
```
rp4rk: Please use a return-early approach, e.g.
```js
if (this.deaths.length === 0) {
  return;
}
```

src/Parser/Warrior/Protection/Modules/Items/ThundergodsVigor.js
```diff
@@ -0,0 +1,57 @@
...
+  on_byPlayer_damage(event) {
+    if (event.ability.guid !== SPELLS.THUNDER_CLAP.id) {
+      return;
+    }
+
+    if(this.spellUsable.isOnCooldown(SPELLS.DEMORALIZING_SHOUT.id)){
+      const reductionEffective = this.spellUsable.reduceCooldown(SPELLS.DEMORALIZING_SHOUT.id, COOLDOWN_REDUCTION_MS);
+      this.effectiveReduction += reductionEffective;
+      this.wastedReduction += COOLDOWN_REDUCTION_MS - reductionEffective;
+    }else{
```
rp4rk: Could return here also to avoid the else block.

src/Parser/Warrior/Protection/Modules/Talents/AngerManagement.js
```diff
@@ -0,0 +1,76 @@
...
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.ANGER_MANAGEMENT_TALENT.id);
+    COOLDOWNS_AFFECTED_BY_ANGER_MANAGEMENT.forEach(e => {
+      this.wastedReduction[e] = 0;
+      this.effectiveReduction[e] = 0;
+    });
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.classResources[0].type === RESOURCE_TYPES.RAGE.id && event.classResources[0].cost) {
```
rp4rk: Reverse the conditional logic here and return early please.

src/Parser/Warrior/Protection/Modules/Talents/AngerManagement.js
```diff
@@ -0,0 +1,76 @@
...
+        }
+      });
+      this.totalRageSpend += rageSpend;
+    }
+  }
+
+  get tooltip() {
+    let returnString = "";
+
+    COOLDOWNS_AFFECTED_BY_ANGER_MANAGEMENT.forEach(e => {
```
rp4rk: This isn't a blocker, but research `reduce` as it would be the best solution to this.

joshinat0r:
btw, made the requested changes

PR #1683, opened 2018-05-10
DH signil of flames

yajinni:
Reworked it. The uptime is no important. The amount of tickle damage/healing it provides is not important.
Double stacking the buff and the burst damage/healing it provides is more important to track.


src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -162,6 +162,19 @@ class Abilities extends CoreAbilities {
         castEfficiency: {
           suggestion: true,
           recommendedEfficiency: 0.90,
+          extraSuggestion: combatant.hasTalent(SPELLS.FLAME_CRASH_TALENT.id)?<span>Line this up with <SpellLink id={SPELLS.INFERNAL_STRIKE.id} />to double stack <SpellLink id={SPELLS.SIGIL_OF_FLAME.id} /> because of the <SpellLink id={SPELLS.FLAME_CRASH_TALENT.id} /> talent.</span>:<span>Cast on cooldown for a dps increase.</span>,
```
MartijnHols: Please use the `React.Fragment` component instead of `span` for wrapping elements without special styling.

MartijnHols: Please pad your code with sufficient white-spacing to keep it readable, also there's a space missing after Infernal Strike.

MartijnHols: `Cast on cooldown for a dps increase.` should be a string, since there's no JSX in it.

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -170,6 +183,13 @@ class Abilities extends CoreAbilities {
         cooldown: 20,
         charges: 2,
       },
+      {
+        spell: SPELLS.SIGIL_OF_SILENCE,
+        category: Abilities.SPELL_CATEGORIES.UTILITY,
+        enabled: combatant.hasTalent(SPELLS.QUICKENED_SIGILS_TALENT.id),
+        cooldown: 48,
```
MartijnHols: This cooldown looks to be incorrect, the tooltip says it's 60 seconds.

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -162,6 +162,19 @@ class Abilities extends CoreAbilities {
...
           suggestion: true,
           recommendedEfficiency: 0.90,
+          extraSuggestion: combatant.hasTalent(SPELLS.FLAME_CRASH_TALENT.id)?<span>Line this up with <SpellLink id={SPELLS.INFERNAL_STRIKE.id} />to double stack <SpellLink id={SPELLS.SIGIL_OF_FLAME.id} /> because of the <SpellLink id={SPELLS.FLAME_CRASH_TALENT.id} /> talent.</span>:<span>Cast on cooldown for a dps increase.</span>,
+        },
+      },
+      {
+        spell: SPELLS.SIGIL_OF_FLAME,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
+        enabled: combatant.hasTalent(SPELLS.QUICKENED_SIGILS_TALENT.id),
+        cooldown: 24,
```
MartijnHols: This cooldown looks to be incorrect, the tooltip says it's 30 seconds.

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -182,6 +202,13 @@ class Abilities extends CoreAbilities {
         cooldown: 60,
         isOnGCD: true,
       },
+      {
+        spell: SPELLS.SIGIL_OF_MISERY,
+        category: Abilities.SPELL_CATEGORIES.UTILITY,
+        enabled: combatant.hasTalent(SPELLS.QUICKENED_SIGILS_TALENT.id),
+        cooldown: 48,
```
MartijnHols: This cooldown looks to be incorrect, the tooltip says it's 60 seconds.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+  currentBuff = 0;
+  difference = 0;
+  stacked = 0;
+
+  on_byPlayer_applydebuff(event){
+    const spellId = event.ability.guid;
+
+    if(spellId !== SPELLS.SIGIL_OF_FLAME_DEBUFF.id) {
+      return;
+    } else{
```
MartijnHols: Please try to avoid nesting code in additional blocks. Instead you should early return. If you return early, you also won't need `else`, which can be omitted here and further down.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+      if(this.previousBuff === 0){
+        this.previousBuff = event.timestamp;
+        return;
+      }else{
+        this.currentBuff = event.timestamp;
+        this.difference = this.currentBuff - this.previousBuff;
```
MartijnHols: This value is not used outside this method or re-used at a later time (it's only overwritten), so please remove it as a class property and instead define a local variable.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
+import Enemies from 'Parser/Core/Modules/Enemies';
+
+import SPELLS from 'common/SPELLS/index';
+import SpellIcon from 'common/SpellIcon';
+
+import { formatPercentage, formatThousands, formatDuration } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+class SigilOfFlame extends Analyzer {
```
MartijnHols: I recommend including a comment above the class to explain the goal of this analyzer.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+      return;
+    } else{
+      if(this.previousBuff === 0){
+        this.previousBuff = event.timestamp;
+        return;
+      }else{
+        this.currentBuff = event.timestamp;
+        this.difference = this.currentBuff - this.previousBuff;
+      }
+      if((this.difference/1000) < 2 ) {
```
MartijnHols: Shouldn't this account for Quickened Sigils?

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+  stacked = 0;
+
+  on_byPlayer_applydebuff(event){
+    const spellId = event.ability.guid;
+
+    if(spellId !== SPELLS.SIGIL_OF_FLAME_DEBUFF.id) {
+      return;
+    } else{
+      if(this.previousBuff === 0){
+        this.previousBuff = event.timestamp;
```
MartijnHols: This variable name isn't crystal clear and suggests it has the entire buff event. How about `lastApplicationTimestamp`?

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+  difference = 0;
+  stacked = 0;
+
+  on_byPlayer_applydebuff(event){
+    const spellId = event.ability.guid;
+
+    if(spellId !== SPELLS.SIGIL_OF_FLAME_DEBUFF.id) {
+      return;
+    } else{
+      if(this.previousBuff === 0){
```
MartijnHols: Unset variables should start as `null` and be checked instead, 0 should be treated as a possible valid value.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+        this.previousBuff = event.timestamp;
+        return;
+      }else{
+        this.currentBuff = event.timestamp;
```
MartijnHols: This value is not used outside this method or re-used at a later time (it's only overwritten), so please remove it as a class property and instead define a local variable.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+class SigilOfFlame extends Analyzer {
+  static dependencies = {
+    abilityTracker: AbilityTracker,
+    enemies: Enemies,
+  };
+
+  previousBuff = 0;
+  currentBuff = 0;
+  difference = 0;
+  stacked = 0;
```
MartijnHols: Taking into account the usages of this class property, I think its name is probably fine but its meaning is somewhat unclear. I recommend leaving a comment here to explain what it tracks. e.g. "This tracks the amount of targets hit by a cast of Sigil of Flame."

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+      }else{
+        this.currentBuff = event.timestamp;
+        this.difference = this.currentBuff - this.previousBuff;
+      }
+      if((this.difference/1000) < 2 ) {
+        this.stacked += 1;
+      }
+      this.previousBuff = this.currentBuff;
+
+      console.log("stacked #", this.stacked);
```
MartijnHols: You should avoid leaving debug lines in code you're pushing to prod.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+  statistic() {
+    const sigilOfFlameUptime = this.enemies.getBuffUptime(SPELLS.SIGIL_OF_FLAME_DEBUFF.id);
+    const sigilOfFlameUptimePercentage = sigilOfFlameUptime / this.owner.fightDuration;
+    const sigilOfFlameDamage = this.abilityTracker.getAbility(SPELLS.SIGIL_OF_FLAME_DEBUFF.id).damageEffective;
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SIGIL_OF_FLAME.id} />}
+        value={`${this.stacked} Times`}
+        label="Sigil of Flame Double Stacked"
```
MartijnHols: Please do not capitalize all words, only those part of the name. This should be "Sigil of Flame double stacked" or just "Double stacked".

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,66 @@
...
+
+  statistic() {
+    const sigilOfFlameUptime = this.enemies.getBuffUptime(SPELLS.SIGIL_OF_FLAME_DEBUFF.id);
+    const sigilOfFlameUptimePercentage = sigilOfFlameUptime / this.owner.fightDuration;
+    const sigilOfFlameDamage = this.abilityTracker.getAbility(SPELLS.SIGIL_OF_FLAME_DEBUFF.id).damageEffective;
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.SIGIL_OF_FLAME.id} />}
+        value={`${this.stacked} Times`}
```
MartijnHols: Please do not capitalize all words, only those part of the name or the first character. How about "${this.stacked} times"?

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
MartijnHols: Is "double stacked" a common term used in this spec's community? If so can you point me to a guide where it's explained? If not, please explain what this shows, as it's not clear as-is.

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
niseko: Instead of having 2 entries for each Sigil, it's possible to use 1 and change the CD based on whether the talent is chosen - like this example from resto shaman (CD with : without talent).
> cooldown: combatant.hasTalent(SPELLS.GRACEFUL_SPIRIT_TALENT.id) ? 60 : 120,

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
MartijnHols: Ohh my bad, I hadn't seen the enabled check. Thanks for enlightening me!

It looks like the only difference with the non-talent-specific version (which is missing an inverted `enabled` property btw as Nisekoi pointed out) is the cooldown, so you could combine the two versions into one with the solution Nisekoi suggested. I personally like to exactly match in-game, which would make it:

`cooldown: 30 * (1 - combatant.hasTalent(SPELLS.GRACEFUL_SPIRIT_TALENT.id) ? 0.2 : 0),`

Then you can reduce some of the duplicate code. Yay! 🎉

yajinni: this looks like a great idea. Will implement it.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
```diff
@@ -0,0 +1,68 @@
...
+    const spellId = event.ability.guid;
+    let timeStampDifference = null;
+
+    if(spellId !== SPELLS.SIGIL_OF_FLAME_DEBUFF.id) {
+      return;
+    }
+    if(this.lastApplicationTimestamp === 0){
+      this.lastApplicationTimestamp = event.timestamp;
+      return;
+    }else{
```
MartijnHols: `else` can be omitted since you already return in the previous block

MartijnHols:
Looks like there's a conflict in `src/Parser/DemonHunter/Vengeance/Modules/Abilities.js`, after this is fixed we can merge this.

src/Parser/DemonHunter/Vengeance/Modules/Spells/SigilOfFlame.js
yajinni: Removed

yajinni:
Fixed merge conflicts

Issue #1661, opened 2018-05-04
[Discipline][BFA Prepatch] Mastery: Reverence


rp4rk:
**Description:**
Mastery: Reverence is the new mastery for Discipline Priests in Battle for Azeroth. Instead of directly increasing the amount of healing Atonement does, it increases healing done to targets with Atonement. 

The passive is currently tuned at 1.2% per point of mastery.

**Ability Link:** https://beta.wowdb.com/spells/271534-mastery-reverence

**Open Questions:**
~~Should this be built into an overall larger pull request including stat weights?~~

**Acceptance Criteria:**
- [x]  HPS from mastery is viewable.

karlpralow:
New Mastery: Your healing and absorption is increased by 10.0 (actually 9.6 in game) % on targets with Atonement.

MartijnHols:
Should this be built into an overall larger pull request including stat weights?

Multiple smaller please!

rp4rk:
Pull request opened.

rp4rk:
Merged, cheers fellas.

PR #1685, opened 2018-05-13
[Discipline][BFA Prepatch] Mastery: Reverence

rp4rk:
**Relevant Issue:** https://github.com/WoWAnalyzer/WoWAnalyzer/issues/1661

---

**Description:**
Mastery: Reverence is the new mastery for Discipline Priests in Battle for Azeroth. Instead of directly increasing the amount of healing Atonement does, it increases healing done to targets with Atonement.

The passive is currently tuned at 1.2% per point of mastery.

### DIFF
--- a/src/Parser/Priest/Discipline/Modules/Spells/Reverence.js
+++ b/src/Parser/Priest/Discipline/Modules/Spells/Reverence.js
@@ -65,32 +65,34 @@ class Reverence extends Analyzer {
   }
 
   statistic() {
+    const reverenceHealingPerc = this.owner.getPercentageOfTotalHealingDone(
+      this.reverenceHealing
+    );
+    const baseHealingPerc = this.owner.getPercentageOfTotalHealingDone(
+      this.baseHealing
+    );
+    const healingWithoutMastery = baseHealingPerc / (1 - reverenceHealingPerc);
+
     return (
       <StatisticBox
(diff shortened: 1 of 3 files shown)


src/Parser/Priest/Discipline/Modules/Spells/Reverence.js
```diff
@@ -0,0 +1,96 @@
...
+          <dfn
+            data-tip={`
+            Reverence contributed towards ${formatPercentage(
+              this.owner.getPercentageOfTotalHealingDone(this.reverenceHealing)
+            )}% of your healing.
+
+
+            ${formatPercentage(
+              this.owner.getPercentageOfTotalHealingDone(this.baseHealing)
+            )}% of your healing benefitted from Reverence.
```
Gebuz: Should this be `baseHealing / (totalHealing - reveranceHealing)` instead of `baseHealing / totalHealing`? Personally I think that percentage would make more sense. For instance, assuming all healing benefitted from Reverance and Reverance did half the healing, I think showing 100% would be more useful than 50% since that way I know how well I utilized my mastery without having to take my mastery percentage into consideration.

rp4rk: Good shout.

Gebuz (approved):
Thanks!

Issue #11, opened 2017-04-30
Suggestion for mana usage


MartijnHols:
Might be hard to do since the mana info is fetched through a separate API call

MartijnHols:
I'm pretty sure mana is actually sent with each event, so doing this shouldn't be too hard.

![image](https://cloud.githubusercontent.com/assets/4565223/26035046/5b4a8f60-38c6-11e7-8bda-1bb93e5f453d.png)

PR #1704, opened 2018-05-16
[BFA] [MW] Rising Mist Implementation

anom0ly:
Implemented Rising Mist talent for MW.

The talent extends all active HOTs by 2 seconds on Rising Sun Kick cast. I used the current Flourish Module and HotTracker to build out this functionality.

@kfinch I wanted to ask if you could take a look since I believe you wrote the original druid HotTracker / Flourish modules for druid.


src/Parser/Monk/Mistweaver/Modules/Talents/RisingMist.js
```diff
@@ -0,0 +1,135 @@
...
+class RisingMist extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    hotTracker: HotTracker,
+  };
+
+  risingMistCount = 0;
+  risingMists = [];
+
+  efsExtended = 0; // tracks how many flourishes extended Wild Growth
```
rp4rk: 🤔

src/Parser/Monk/Mistweaver/Modules/Talents/RisingMist.js
```diff
@@ -0,0 +1,135 @@
...
+  efCount = 0;
+  evmCount = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.RISING_MIST.id);
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS.RISING_SUN_KICK.id !== spellId) {
```
rp4rk: 👍

src/Parser/Monk/Mistweaver/Modules/Talents/RisingMist.js
```diff
@@ -0,0 +1,135 @@
...
+      return;
+    }
+
+    this.risingMistCount += 1;
+    debug && console.log(`risingMist cast #: ${this.risingMistCount}`);
+
+    const newRisingMist = {
+      name: `risingMist #${this.risingMistCount}`,
```
rp4rk: Probably better to have this in english instead of camelCase right? "Rising Mist #11" for example?

src/Parser/Monk/Mistweaver/Modules/Talents/RisingMist.js
```diff
@@ -0,0 +1,135 @@
...
+      healing: 0,
+      procs: 0,
+      duration: 0,
+    };
+    this.risingMists.push(newRisingMist);
+
+    let foundEf = false;
+
+    Object.keys(this.hotTracker.hots).forEach(playerId => {
```
rp4rk: Oof, okay - this could be improved a good bit I think. Any time you find yourself assigning to an object from a forEach, you should probably consider reduce as a good way to make that code a bit more readable. To make this even more verbose, we can use `Object.entries` to avoid keying the object and using that to access the associated values. 

Here's some quick code to demonstrate a similar problem:
```js
const hotTest = {
  reglitch: ['3', '4'],
  oratio: ['2'],
  zero: ['1', '2', '4'],
};

// Should be obvious
const strToNumber = str => Number(str) || undefined;

// Log occurences of a variable to an object
const logToObject = (arr, obj = {}) =>
  arr.reduce((acc, curr) => {
    if (acc[curr]) {
      acc[curr] += 1;
      return acc;
    }

    acc[curr] = 1;
    return acc;
  }, obj);

// Build a k:v pair of how often spells are used given an array of spellIds assigned to playerIds
// ​​​​​Output here is: { '1': 1, '2': 2, '3': 1, '4': 2 }​​​​​
const spellUsageLog = Object.entries(hotTest).reduce((acc, curr) => {
  const [player, spellIds] = curr;
  const spellsUsed = spellIds.map(strToNumber);

  // Put code to add extension to hot tracker here lol
  const logSpells = logToObject(spellsUsed, acc);

  return logSpells;
}, {});
```

You'll need to filter for the set of spells you want to account for but that should be trivial.

src/Parser/Monk/Mistweaver/Modules/Core/HotTracker.js
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  _generateHotInfo() { // must be generated dynamically because it reads from traits
+    return {
+      [SPELLS.RENEWING_MIST_HEAL.id]: {
+        duration: 20000,
+        tickPeriod: 2000,
+      },
+      [SPELLS.ENVELOPING_MISTS.id]: {
+        duration: 6000 + (this.combatants.selected.hasTalent(SPELLS.MIST_WRAP_TALENT.id) ? 1 : 0),
```
kfinch: Should be ` ... ? 1000 : 0),` , right?

src/Parser/Monk/Mistweaver/Modules/Core/HotTracker.js
```diff
@@ -0,0 +1,366 @@
...
+        tickPeriod: 1000,
+      },
+      [SPELLS.ESSENCE_FONT_BUFF.id]: {
+        duration: 8000 + (this.combatants.selected.hasTalent(SPELLS.UPWELLING_TALENT.id) ? 4 : 0),
```
kfinch: Should be `... ? 4000 : 0),` , right?

src/Parser/Monk/Mistweaver/Modules/Talents/RisingMist.js
kfinch: I think he was copying my Flourish code... but yeah ok it's not the prettiest thing I've written.

PR #1706, opened 2018-05-16
[Protection Warrior] Into the Fray, Heavy Repercussions, Avatar, Shield Slam Resets update

joshinat0r:
- adds Into the fray
average haste gained + detailed stack uptime + stattracker takes stacks into account
- adds heavy repercussions
adds suggestion when refreshing the SS damage buff (by casting two SBs without an SS inbetween)
- adds avatar
- updates T21 2p & shield slam resets to take the GCD into account 
eg. TC cast would reset SS, SS is of CD but the GCD is still going, resulting in inaccurate SS efficiency

hr
intothefray
avatar

### DIFF
--- a/src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
+++ b/src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
@@ -1,0 +1,102 @@
+import React from 'react';
+import Analyzer from 'Parser/Core/Analyzer';
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import Enemies from 'Parser/Core/Modules/Enemies';
+import calculateEffectiveDamage from 'Parser/Core/calculateEffectiveDamage';
+import { formatNumber, formatPercentage } from 'common/format';
+
+const HEAVY_REPERCUSSIONS_SHIELD_BLOCK_EXTEND_MS = 1000;
+const HEAVY_REPERCUSSIONS_SHIELD_SLAM_DAMAGE_BUFF = 0.3;
+
+class HeavyRepercussions extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    enemies: Enemies,
+  };
+
+  sbExtended = 0;
+  ssBuffRefreshed = 0;
+  nextShieldSlamBuffed = false;
+  bonusDmg = 0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.HEAVY_REPERCUSSIONS_TALENT.id);
+  }
+
+  get shieldBlockuptime() {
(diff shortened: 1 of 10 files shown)


Gebuz:
Please make the title a bit more precise! :P If you're changing lots of different things just mention the most important onces.

joshinat0r:
thought "some stuff" is precise enough 🤔

src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
```diff
@@ -0,0 +1,102 @@
...
+
+class HeavyRepercussions extends Analyzer {
+  static dependencies = {
+    combatants: Combatants,
+    enemies: Enemies,
+  };
+
+  sbExtended = 0;
+  ssBuffRefreshed = 0;
+  nextShieldSlamBuffed = false;
```
Gebuz: I don't play prot, but reading the Heavy Repercussions tooltip it implies that all Shield Slams are buffed during Shield Block, not just the first one. Are you sure this is how it works?

src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
```diff
@@ -0,0 +1,102 @@
...
+            .recommended(`${recommended} recommended`);
+        });
+  }
+
+  statistic() {
+    const sbExtendedMS = this.sbExtended * HEAVY_REPERCUSSIONS_SHIELD_BLOCK_EXTEND_MS;
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.HEAVY_REPERCUSSIONS_TALENT.id} />}
+        value={`${formatPercentage(sbExtendedMS / (this.shieldBlockuptime - sbExtendedMS))}%`}
```
Gebuz: Would it make sense to make a suggestion based on this percent?

src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
```diff
@@ -0,0 +1,102 @@
...
+        label={`more Shield Block uptime`}
+        tooltip={`
+          You casted Shield Slam ${this.sbExtended} times during Shield Block, resulting in additional ${sbExtendedMS / 1000}sec uptime.<br/>
+          ${formatNumber(this.bonusDmg)} damage contributed by casting buffed Shield Slams<br/>
```
Gebuz: would be nice to have DPS and percentOfTotalDmg aswell!

src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
joshinat0r: You're right, it's buffing all SS's during SB. Not sure why I thought it's per SB cast.

src/Parser/Warrior/Protection/Modules/Talents/HeavyRepercussions.js
joshinat0r: Suggestion is now based on the amount of casts made during SB. Its kinda a mood point, someone might want to pool SB charges while offtanking. On the other hand its possible to geht 100% uptime on SB with HR (± haste), so casting it while offtanking is still a DPS increase.
Might adjust thresholds later on.

PR #1707, opened 2018-05-16
[Discipline Priest] Desperate Prayer

karlpralow:
image


src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+      </ExpandableStatisticBox>
+    );
+  }
+
+  suggestions(when) {
+    when(this.deathsWithDPReady).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>You died with <SpellLink id={SPELLS.DESPERATE_PRAYER.id} /> on cooldown.</span>)
+          .icon(SPELLS.DESPERATE_PRAYER.icon)
+          .actual(`You died ${this.deathsWithDPReady} times with Desperate Prayer on cooldown.`)
```
rp4rk: time(s)

src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+    this.currentDesperatePrayerUsage.HPWhenUsed = event.hitPoints - event.amount;
+    this.currentDesperatePrayerUsage.MaxHPWhenUsed = event.maxHitPoints;
+  }
+
+  on_toPlayer_damage(event) {
+    if(!this.combatants.selected.hasBuff(SPELLS.DESPERATE_PRAYER.id)) {
+      return;
+    }
+    this.currentDesperatePrayerUsage.DamageTaken += event.amount;
```
MartijnHols: This should include absorbed healing, such as from shields.

src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+
+class DesperatePrayer extends Analyzer {
+
+  static dependencies = {
+    combatants: Combatants,
+    spellUsable: SpellUsable,
+  };
+
+  desperatePrayerUsages = [];
+  currentDesperatePrayerUsage;
```
MartijnHols: Please explicitly set a default value for a class property, in this case setting it to `null` would probably be best.

src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+      return;
+    }
+    this.currentDesperatePrayerUsage.DamageTaken += event.amount;
+  }
+
+  on_byPlayer_removebuff(event) {
+    if (event.ability.guid !== SPELLS.DESPERATE_PRAYER.id) {
+      return;
+    }
+    this.desperatePrayerUsages.push(this.currentDesperatePrayerUsage);
```
MartijnHols: Activations at the end of the fight will be missed with this. You might not even need the `currentDesperatePrayerUsage` property and could instead refer to the last item in the array.

src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+                ))
+            }
+          </tbody>
+        </table>
+      </ExpandableStatisticBox>
+    );
+  }
+
+  suggestions(when) {
+    when(this.deathsWithDPReady).isGreaterThan(0)
```
MartijnHols: Please use the `disableDeathSuggestion` property to avoid unnecessarily giving suggestions for fights like Argus.

src/Parser/Priest/Shared/Modules/Features/DesperatePrayer.js
```diff
@@ -0,0 +1,105 @@
...
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span>You died with <SpellLink id={SPELLS.DESPERATE_PRAYER.id} /> on cooldown.</span>)
```
MartijnHols: Please use the `React.Fragment` component instead of `span`s when you don't need the wrapper for any other reason than to make React happy.

MartijnHols: Shouldn't this be the opposite; "You died with DP available"?

MartijnHols: The `DeathTracker` module already complains about dying, how about extending that with a suggestion to use DP when available instead of making an all new suggestion?

karlpralow: I can see the redundancy here, but I feel like having the "You were dead for x % of the fight" statistic and the deaths with DP available portrays 2 different things. 

I'm open to suggestion on how both informations can be displayed in a clear way

MartijnHols: DeathTracker could reveal this information, it doesn't have to be plain text.

PR #1709, opened 2018-05-16
[Blood Death Knight] Deaths Caress Suggestion

joshinat0r:
Adds a suggestion for the special kind of people when using DC, there is only very little reason to cast it over HS, DnD or BD.

checklist
suggestion

### DIFF
--- a/src/Parser/DeathKnight/Blood/Modules/Core/DeathsCaress.js
+++ b/src/Parser/DeathKnight/Blood/Modules/Core/DeathsCaress.js
@@ -22,13 +22,13 @@ class DeathsCaress extends Analyzer {
 
   get averageCastSuggestionThresholds() {
     return {
-      actual: this.dcCasts,
-      isGreaterThan: {
+      actual: 1 - (this.dcCasts / (this.tracker.runesMaxCasts - this.tracker.runesWasted)),
+      isLessThan: {
         minor: 1,
-        average: 3,
-        major: 5,
+        average: .95,
+        major: .9,
       },
-      style: 'number',
+      style: 'percentage',
     };
   }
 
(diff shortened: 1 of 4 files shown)


Gebuz:
I'd suggest making it casts per minute or percent of runes spent to make it relative to encounter length, otherwise it looks fine. Also seems to be missing a "casting" or "using" in the checklist (Avoid >casting< Death's Caress)

joshinat0r:
Updated the suggestion. The actual threshold is still based on the total amount of casts but the suggestion-text is using a percentage of total runes spend.
Making the suggestion based on percent would lead to inaccurate suggestions (like a player used DC 20 times during a 15min fight might not get a suggestion (even if its just a minor one) but one that had one DC cast in a 2min fight would get a major one), thats why its still based on total amount of casts.

Gebuz:
That's the point of making it relative. If the encounter is 15 min. long casting 5 DC should be a minor issue since it is a rather small dps loss. On a 1 min. encounter casting 5 DCs is a major dps loss, thus should be a major issue.

joshinat0r:
But casting DC 5 times during a 15min encounter is just as wrong as casting it 5 times during a 1min encounter. It should be in both cases a major issue, at least in my opinion. Even if its not as bad relative to the total amount of runes spend.

Gebuz:
It's not a matter of if it's wrong or not, it's a matter of how impactful it is. If you had 2 buttons A and B, and A did 1dps and B did 0.99 dps, it would always be wrong to press B, but it would still be a minor issue, particularly when compared to pressing nothing at all. 

If you have casting DC 5 times over a 15 min. encounter as a major issue, a player might see that and focus on that instead of focusing on a medium issue like "you overcapped 10% of your runes" (which on a 15 min. encounter would be waaaaaay more detrimental to ones output), when in reality making sure that they spent all their runes is much more impactful than not casting 5 DCs over 15 min.

src/Parser/DeathKnight/Blood/Modules/Core/DeathsCaress.js
```diff
@@ -0,0 +1,46 @@
...
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.DEATHS_CARESS.id) {
+      return;
+    }
+    this.dcCasts += 1;
+  }
+
+  get averageCastSuggestionThresholds() {
+    return {
+      actual: 1 - (this.dcCasts / (this.tracker.runesMaxCasts - this.tracker.runesWasted)),
```
Gebuz: Threshold or the suggestion needs to be inverted (threshold is runes not spent on DC, suggestion is runes spent on DC). Otherwise good to merge.

joshinat0r: That's actually done on purpose.
I think having in the checklist "avoid casting DC" and display 100% with 0 casts is better than having "DC casts" and a full green bar with 0 casts.
Having the suggestion based on runes "wasted" by casting DC is better for the suggestions, at least in my opinion.

If you don't think so, tell me which one should he inverted and I'll update it.

PR #1712, opened 2018-05-18
Deathlog for DH

yajinni:
image

### DIFF
--- a/src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
+++ b/src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
@@ -22,28 +22,14 @@ class Abilities extends CoreAbilities {
       },
       {
         spell: SPELLS.DEMON_SPIKES,
-        enabled: combatant.hasTalent(SPELLS.RAZOR_SPIKES_TALENT.id),
         buffSpellId: SPELLS.DEMON_SPIKES_BUFF.id,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
         cooldown: haste => 15 / (1 + haste),
-        charges: 3,
+        charges: combatant.hasTalent(SPELLS.RAZOR_SPIKES_TALENT.id) ? 3 : 2,
         castEfficiency: {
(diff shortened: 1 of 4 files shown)


src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -144,14 +146,18 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.SHEAR,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
       },
+      {
+        spell: SPELLS.PAINBRINGER_Buff,
```
Gebuz: Wouldn't this cause it to show up in the cast efficiency tab? I think we came to the conclusion that we should just put buffs that doesn't belong to a specific skill in `src/common/DEFENSIVE_BUFFS.js`.

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
```diff
@@ -23,7 +23,8 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.DEMON_SPIKES,
         enabled: combatant.hasTalent(SPELLS.RAZOR_SPIKES_TALENT.id),
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        buffSpellId: SPELLS.DEMON_SPIKES_BUFF.id,
+        category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
         cooldown: haste => 15 / (1 + haste),
         charges: 3,
```
Gebuz: Not related to the PR, but you could remove one of the demon spike entrances and replace this line with `charges: combatant.hasTalent(SPELLS.RAZOR_SPIKES_TALENT.id) ? 3 : 2,`

src/Parser/DemonHunter/Vengeance/Modules/Abilities.js
yajinni: True. Forgot about that

PR #1713, opened 2018-05-18
[BFA Resto Druid] Photosynthesis

karlpralow:
Implementation of the new resto druid talent Photosynthesis

image

### DIFF
--- a/src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
+++ b/src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
@@ -71,7 +71,7 @@ class Photosynthesis extends Analyzer {
         value={`${formatPercentage(totalPercent)} %`}
         label={'Photosynthesis'}
         tooltip={`
-            Photosynthesis healing contribution
+            Healing contribution
             <ul>
               <li>Rejuvenation: <b>${formatPercentage(this.owner.getPercentageOfTotalHealingDone(this.rejuvenationIncrease))} %</b></li>
               <li>Lifebloom: <b>${formatPercentage(this.owner.getPercentageOfTotalHealingDone(this.lifebloomIncrease))} %</b></li>
--- a/src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
+++ b/src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
@@ -58,16 +58,30 @@ class Photosynthesis extends Analyzer {
 
   statistic() {
     const totalPercent = this.owner.getPercentageOfTotalHealingDone(this.rejuvenationIncrease + this.lifebloomIncrease);
+    const sourceID = this.combatants.selected._combatantInfo.sourceID;
+    const selfUptime = this.combatants.selected.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id, sourceID);
+    const totalUptime =
+      Object.keys(this.combatants.players)
+          .map(key => this.combatants.players[key])
+          .reduce((uptime, player) => uptime + player.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id),  sourceID);
+
     return (
       <StatisticBox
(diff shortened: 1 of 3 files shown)


src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
```diff
@@ -0,0 +1,75 @@
...
+      return;
+    }
+    this.lastRealBloomTimestamp = event.timestamp;
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    const amount = event.amount + (event.absorbed || 0);
+
+    if(spellId === SPELLS.REJUVENATION.id && this.combatants.selected.hasBuff(SPELLS.LIFEBLOOM_HOT_HEAL.id)) {
```
Gebuz: Remember to check if it is the player's own lifebloom and not another resto druid's.

src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
```diff
@@ -0,0 +1,75 @@
...
+      this.rejuvenationIncrease += amount * (1 - (1 / (1 + PHOTOSYNTHESIS_REJUV_INCREASE)));
```
Gebuz: please use `calculateEffectiveHealing(event, PHOTOSYNTHESIS_REJUV_INCREASE)` for this (`import calculateEffectiveHealing from 'Parser/Core/calculateEffectiveHealing';`)

src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if(spellId === SPELLS.LIFEBLOOM_BLOOM_HEAL.id && (this.lastRealBloomTimestamp === null || (event.timestamp - this.lastRealBloomTimestamp) > 32)){
```
Gebuz: What's the magic number 32? I don't understand the logic of `lastRealBloomTimestamp`. I assume it's mean to track the last bloom heal that was not from Photosynthesis, but why is it set on debuff refreshed/removed? iirc the bloom happens every 15 sec after initial application (so if you refresh it it doesn't affect the bloom timing) reduced by haste, but it's been a while since I read up on it so could've changed. This makes it pretty hard to track due to dynamic haste scaling.

karlpralow: Lifebloom also heal in the pandemic window. This has been taken from the legendary belt module, I assume it has been discussed why 32 is a fair window for accurate measuraments.

karlpralow:
@Gebuz modification we're made accordingly. thank you

src/Parser/Druid/Restoration/Modules/Talents/Photosynthesis.js
Gebuz: hmm, might be better if @kfinch takes a look at it then, idk how that works. Unless lifebloom refreshes itself somehow when the belt procs (which seems weird) it doesn't make sense to me. Anyway should throw the 32 into a constant and call it MS_BUFFER or something so we know what it is  :P

karlpralow: If you cast Lifebloom and your lifebloom has 12 seconds remaining, it triggers the bloom effect and refreshes the debuff.

I will throw the 32 ms into a constant

kfinch: The 32 ms was basically arbitrary, I think Zero came up with it months ago when he was looking for the gap between some two events that should happen together, and the biggest gap he saw was 32 ms. In practice it varies by spell/event pair. Some are always on the same timestamp, while for other effects I've observed gaps of as much as 150ms.

When picking a buffer I've personally often defaulted to 50, but will make it larger if it's something where there is almost no chance of a false positive.

Anyway, 32 should be fine for this.

PR #1714, opened 2018-05-18
guardian deathlog

yajinni:


Gebuz:
I made it ignore abilities categorized as DEFENSIVE on the "Personals Available" in #1715

Gebuz:
@yajinni Remember Rage of the Sleeper! Otherwise good to merge.

joshinat0r:
Yeah, adding those "rotational" defensives to the cooldown list wasn't inteded (thought the "defensive" category was just used for cooldowns, not rotational defensives). I think I would have added IF in that case to the defensive_buffs-file.
Will add a check for spells that have no cooldown and hide them in the "cooldowns available" column.

yajinni:
I mean something like iron fur. If you die and you dont have it up you screwed up though. So shouldnt that be showing? Or are you guys talking about something else.

joshinat0r:
It wasn't about if it should be displayed as a buff, totally should. But giving it the defensive-category would add it to the "cooldown available"-tab, which should not happen (its less of a cooldown, more a thing you maintain). Gebuz merged already a fix for that.

PR #1715, opened 2018-05-19
[Balance Druid] Deathlog and buffs/debuffs on timeline

Gebuz:
image
image
(yeah I know we don't have many defensives <.<)


Gebuz:
image

PR #1718, opened 2018-05-19
[BFA] [BrM] Basic Updates

emallson:
Update some CDs, remove some stuff to do with traits, work around a bug with stagger logging. Disabled the `DamageTakenTable` since we don't know the actual amount thats supposed to be staggered yet.

No visible changes, aside from misc. changes to tooltips that can be read.


src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -67,12 +62,29 @@ class StaggerFabricator extends Analyzer {
...
+        this._lastMelee = null;
+      }
+
       this._staggerPool -= amount;
       // sometimes a stagger tick is recorded immediately after death.
       // this ensures we don't go into negative stagger
       this._staggerPool = Math.max(this._staggerPool, 0);
       debug && console.log("triggering stagger pool update due to stagger tick");
       this.owner.fabricateEvent(this._fab(EVENT_STAGGER_POOL_REMOVED, event, amount), event);
+    } else if (event.ability.guid === SPELLS.MELEE.id) {
```
Gebuz: Could avoid this else if by checking if it is a melee first and returning early if it is. Would avoid the indentation on the main logic above.
```
if (event.ability.guid === SPELLS.MELEE.id) {
  this._lastMelee = event;
  return;
}
if (event.ability.guid !== SPELLS.STAGGER_TAKEN.id) {
  return;
}
//handle stagger
```

src/Parser/Monk/Brewmaster/Modules/Core/StaggerFabricator.js
```diff
@@ -82,12 +94,13 @@ class StaggerFabricator extends Analyzer {
       this._staggerPool -= amount;
       debug && console.log("triggering stagger pool update due to purify");
       this.owner.fabricateEvent(this._fab(EVENT_STAGGER_POOL_REMOVED, event, amount), event);
-    } else if (this._hasQuickSipTrait && event.ability.guid === SPELLS.IRONSKIN_BREW.id) {
-      const amount = this._staggerPool * ISB_QUICK_SIP_PURIFY;
-      this._staggerPool -= amount;
-      debug && console.log("triggering stagger pool update due to ISB + Quick Sip trait");
-      this.owner.fabricateEvent(this._fab(EVENT_STAGGER_POOL_REMOVED, event, amount), event);
-    }
+    } // leaving this as an IF because i expect an ELSE from azerite traits
```
Gebuz: Better to return early.

src/Parser/Monk/Brewmaster/Modules/Features/Abilities.js
```diff
@@ -43,14 +43,6 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.TIGER_PALM,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
       },
-      {
-        spell: SPELLS.EXPLODING_KEG,
```
Gebuz: You could have the `enabled:` check if the artifact is equipped instead of removing it.

PR #1719, opened 2018-05-19
[Core] debuffs in Deathrecap 

joshinat0r:
Adds a new column to the table with all defensive debuffs up on the target for the damage event. Still takes the IDs from all defensive cooldowns (with the same spellID as the ability, the buffSpellId or the `DEFENSIVE_BUFFS`-file).
Might add debuffs for the player aswell (like tank major debuffs that increase damage taken) but I'm right now not quite sure how to add that without making the recap too cluttered.

image


joshinat0r:
Debuffs on the target appear now in the same column as the buffs, with a red border in a new line.
I hope its enough to distinguish buffs on the player and debuffs on the target.

image

yajinni:
I like it. In the end its still a defensive ability, whether its on you or the target.  I dont think there is much benefit to keeping the 2 separate.

PR #1725, opened 2018-05-20
[Restoration Shaman] Healing Rain Interaction - Deluge & Elemental Rebalancers

### LINKED ISSUE #1648: [Restoration Shaman] Find out Deluge healing contribution
#### [Deluge](http://www.wowhead.com/spell=200076/deluge) increases Chain Heal healing on people in Healing Rain or affected by Riptide by 20%.

Issue about this is that everyone in healing rain gets increased healing, not only the ones getting healed by rain, so tracking people getting healed would be inaccurate (and they could move out of rain).
Thought about finding out the rain position by the coordinates of people getting healed, and checking chain heal healing event coordinates against that, but not sure if thats even feasible.

With how the talents change in BFA we will see a lot more deluge logs.

niseko:
image
imageimage

As proposed in #1648 I'm figuring out the position of Healing Rain by using the coordinates of people getting healed by it, and using those to compare the coordinates of healing events during healing rain against an ellipse.
It seems to be pretty accurate, rare that a rain is bigger than the expected 20 yards, and if so, it gets filtered out (likely the odd mage blinking out while receiving a tick). 

closes #1648


src/Parser/Shaman/Restoration/Modules/ShamanCore/HealingRainLocation.js
```diff
@@ -0,0 +1,119 @@
...
+    if (spellId !== SPELLS.HEALING_RAIN_CAST.id || event.isCancelled) {
+      return;
+    }
+
+    // Setting a variable to reset the array on the next heal
+    // If we'd reset here it would sometimes clear the array before the healing is calculated
+    this.newHealingRain = true;
+  }
+
+  processLastRain(eventsDuringRain, healIncrease) {
```
MartijnHols: This method name is unclear and I think it does one too many things. How about splitting it in two; filterEventsInHealingRain which act like a filter for Array.prototype.filter and sumHealing which could perhaps be a reducer?

MartijnHols:
just some conflicts and then good to go

niseko:
fixed, and me sneaking in a small change from the last PR apparently dismisses the review 💯

Issue #1724, opened 2018-05-19
Trinket: Darkmoon Deck: Blockades


MartijnHols:
https://bfa.wowhead.com/item=159128/darkmoon-deck-blockades

![image](https://user-images.githubusercontent.com/4565223/40273554-347bbccc-5bc2-11e8-912f-2fcf98aeb4f7.png)

Issue #1723, opened 2018-05-19
Trinket: Darkmoon Deck: Tides


MartijnHols:
https://bfa.wowhead.com/item=159127/darkmoon-deck-tides

![image](https://user-images.githubusercontent.com/4565223/40273549-273ced06-5bc2-11e8-8709-17d916d43edc.png)

Issue #1722, opened 2018-05-19
Trinket: Darkmoon Deck: Squalls


MartijnHols:
https://bfa.wowhead.com/item=159126/darkmoon-deck-squalls

![image](https://user-images.githubusercontent.com/4565223/40273544-0e9ea2e4-5bc2-11e8-9fee-fc64f81ce43c.png)

abelito75:
Are these trinkets still needed?

Issue #1721, opened 2018-05-19
Trinket: Darkmoon Deck: Fathoms


MartijnHols:
https://bfa.wowhead.com/item=159125/darkmoon-deck-fathoms

![image](https://user-images.githubusercontent.com/4565223/40273539-fdcf5daa-5bc1-11e8-93c5-eb2df85efa30.png)

PR #1730, opened 2018-05-21
[BFA] Add missing trinkets to items

avilene:
Should make working on the following issues a bit easier:

- #1724
- #1723
- #1722
- #1721
- #1720
- #1703
- #1701
- #1700
- #1699
- #1698
- #1697
- #1696
- #1695
- #1694
- #1693
- #1692
- #1691


src/common/ITEMS/BFA/Others.js
```diff
@@ -0,0 +1,34 @@
+export default {
```
MartijnHols: I'd like to avoid an "others" file as much as possible for the BFA folders. How about naming this "Crafted"?

MartijnHols (changes requested):

MartijnHols (approved):
Thank you

PR #1731, opened 2018-05-22
[BfA][Blood Death Knight] added Rune Strike, updated Hemostasis, Ossuary and spells on the GCD

joshinat0r:
- Updated Hemostasis (from 10% -> 8%)
- Removed Blood Tap and added Rune Strike
- readded Ossuary (removed it because Ossuary got an overhaul, but Blizzard went back to the old Ossuary)
- added DRW and Tombstone to the GCD (to reflect the current beta-server)

**Statistic Box**
image
**Timeline**
image
**Checklist**
image
The first checklist-item for Rune Strike displays the generic cast efficiency, the second one in "dont overcap your resources" is for the wasted cooldown-reduction on Rune Strike.


joshinat0r:
Updated it, its not a constant outside of the class.

PR #1734, opened 2018-05-23
[BFA] [Frost Mage] General Spec Changes

Sharrq:
Adjusted Abilities and ABC. Also changed changelog to get rid of item links for legendaries, tier sets, and removed spells.

Icy Veins added to GCD
Removed Frost Bomb
Removed Water Jet
Pointed to EBONBOLT_TALENT instead of EBONBOLT
adjusted Ebonbolt Cast Time


Sharrq:
fixed that error before travis finished.

src/Parser/Mage/Frost/Modules/Features/Abilities.js
```diff
@@ -12,9 +12,10 @@ class Abilities extends CoreAbilities {
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
       },
       {
-        spell: SPELLS.EBONBOLT,
+        spell: SPELLS.EBONBOLT_TALENT,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: haste => 45 + (3 / (1 + haste)), // 45 Second Cooldown with a 3 Second Cast time (Reduced by Haste). Temp until CastEfficiency gets a redo
+        cooldown: haste => 45 + (2.5 / (1 + haste)), // 45 Second Cooldown with a 2.5 Second Cast time (Reduced by Haste). Temp until CastEfficiency gets a redo
```
MartijnHols: Cast efficiency not including cast time is intended. The cooldown should match in-game cd as closely as possible. To compensate for casting time you can reduce the recommended efficiency.

src/Parser/Mage/Frost/Modules/Features/AlwaysBeCasting.js
```diff
@@ -19,20 +19,20 @@ class AlwaysBeCasting extends CoreAlwaysBeCasting {
...
     // talents
     SPELLS.RAY_OF_FROST_TALENT.id,
     SPELLS.MIRROR_IMAGE_TALENT.id,
     SPELLS.RUNE_OF_POWER_TALENT.id,
     SPELLS.ICE_NOVA_TALENT.id,
     SPELLS.RING_OF_FROST_TALENT.id,
-    SPELLS.FROST_BOMB_TALENT.id,
 	  SPELLS.GLACIAL_SPIKE_TALENT.id,
-	  SPELLS.COMET_STORM_TALENT.id,
+    SPELLS.COMET_STORM_TALENT.id,
```
MartijnHols: Please use the isOnGCD prop instead. This array is deprecated.

src/Parser/Mage/Frost/Modules/Features/Abilities.js
Sharrq: The reason I added it months and months ago was because I was frequently getting complaints from people who were indicating that they literally were hitting ebonbolt the second it came off cooldown, and cast efficiency said they had 7/8 casts.

When I investigated, I found that the cast time on it, since it is longer, added up to where the analyzer thought they could fit another cast in, but they couldn't.

So I added that to make the results more accurate since Ebonbolt is a hard hitting spell with a long cast that is pretty much required to get the max casts in per fight. Without compensating for the cast time, players were lead to believe they screwed up when they didn't.

I can remove it, but I would prefer to keep it for the above reasons

PR #1738, opened 2018-05-26
[Restoration Shaman] Death recap & Cloudburst buff fabrication

niseko:
image

Fabricated buff and removebuff events for cloudburst, so that i can work with them for the timeline, as what you do during that buff is important. This also finally shows how the totem has a delay after expiring before it heals, which varies between 50ms and 900ms 👎 
image
image


src/Parser/Shaman/Restoration/Modules/Abilities.js
```diff
@@ -40,8 +40,8 @@ class Abilities extends CoreAbilities {
       },
       {
         spell: SPELLS.HEALING_STREAM_TOTEM_CAST,
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        charges: 2,
+        category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
```
MartijnHols: I just pushed a commit so you don't have to change the category of this spell to include it as a defensive and can instead just set the `isDefensive` prop to `true` - if you want.

src/Parser/Shaman/Restoration/Modules/Talents/CloudburstTotem.js
```diff
@@ -26,10 +36,46 @@ class CloudburstTotem extends Analyzer {
...
+    this.owner.fabricateEvent({
+      ...event,
+      ability: {
+        ...event.ability,
+        guid: SPELLS.CLOUDBURST_TOTEM_TALENT.id,
+      },
+      type: type,
+      targetID: event.sourceID,
+      targetIsFriendly: event.sourceIsFriendly,
+      __fabricated: true,
```
MartijnHols: `fabricateEvent` already sets `__fabricated`, so you don't have to do this here as well

src/Parser/Shaman/Restoration/Modules/Talents/UnleashLife.js
```diff
@@ -251,7 +251,7 @@ class UnleashLife extends Analyzer {
 
   unleashLifeCastRatioChart() {
     const totalUses = this.buffedChainHeals + this.buffedHealingWaves + this.buffedHealingSurges + this.buffedRiptides;
-    const unusedUL = totalUses - this.unleashLifeCasts;
+    const unusedUL = this.unleashLifeCasts - totalUses;
```
MartijnHols: This looks like a significant change, was this intended?

niseko: I noticed logs where Unleash Life unused casts showed as negative, on top of never seeing it higher than 0. After investigating, this being the wrong way around was the cause since `11 Buffs used - 12 Casts = -1 Unused`
The change fixes it to `12 Casts - 11 Buffs used = 1 Unused`

PR #1742, opened 2018-05-27
[Core] Death Recap

joshinat0r:
Enables the death recap for all specs.
Updated all specs and their abilities (or added if missing) to include defensives with `buffSpellId` and updated the categories to `DEFENSIVE` or `SEMI_DEFENSIVE` or `isDefensive: true` to show them in the recap.

Does not include highly situational spells (such as Hand of freedom, dispells or externals(since they're ususally not used as personal, or shouldn't).
Pretty sure I missed some spells that gain defensive benefits from the artifact weapon, but I'll leave that to the spec maintainers.

### DIFF
--- a/src/Parser/Rogue/Assassination/Modules/Features/Abilities.js
+++ b/src/Parser/Rogue/Assassination/Modules/Features/Abilities.js
@@ -96,6 +96,12 @@ class Abilities extends CoreAbilities {
         cooldown: 30,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
       },
+      {
+        spell: SPELLS.EVASION,
+        buffSpellId: SPELLS.EVASION.id,
+        cooldown: 120,
+        category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
+      },
     ];
   }
 }
--- a/src/Parser/Rogue/Outlaw/Modules/Features/Abilities.js
+++ b/src/Parser/Rogue/Outlaw/Modules/Features/Abilities.js
@@ -92,6 +92,12 @@ class Abilities extends CoreAbilities {
         cooldown: 30,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
       },
+      {
+        spell: SPELLS.RIPOSTE,
+        buffSpellId: SPELLS.RIPOSTE.id,
+        cooldown: 120,
+        category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
+      },
     ];
   }
 }
--- a/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
+++ b/src/Parser/Rogue/Subtlety/Modules/Features/Abilities.js
@@ -122,6 +122,12 @@ class Abilities extends CoreAbilities {
         cooldown: 30,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
       },
+      {
+        spell: SPELLS.EVASION,
+        buffSpellId: SPELLS.EVASION.id,
+        cooldown: 120,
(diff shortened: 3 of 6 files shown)


src/common/DEFENSIVE_BUFFS.js
```diff
@@ -52,6 +56,22 @@ const DEFENSIVE_BUFFS = [
   {
     spell: SPELLS.GHOST_IN_THE_MIST_BUFF,
   },
+  //Brewmaster Monk
+  {
+    spell: SPELLS.IRONSKIN_BREW_BUFF,
```
Gebuz: You already added the buffID to abilities so shouldn't be needed here?

src/common/DEFENSIVE_BUFFS.js
```diff
@@ -41,6 +41,10 @@ const DEFENSIVE_BUFFS = [
   {
     spell: SPELLS.SHROUD_OF_PURGATORY,
   },
+  //Frost Death-Knight
+  {
+    spell: SPELLS.PERMAFROST_TALENT,
```
Gebuz: I don't think absorbs like this and Soul Leech are worth tracking since they are completely passive.

src/common/DEFENSIVE_BUFFS.js
joshinat0r: Ironskinbrew shares a cooldown with Purifyingbrew.
While ISB has already a buffId, it doesn't show up (I guess because the ability is an array).
Adding it to the `DEFENSIVE_BUFFS` was a quick fix.

src/common/DEFENSIVE_BUFFS.js
joshinat0r: removed both

src/common/DEFENSIVE_BUFFS.js
```diff
@@ -52,6 +56,22 @@ const DEFENSIVE_BUFFS = [
...
+  {
+    spell: SPELLS.IRONSKIN_BREW_BUFF,
+  },
+  //Affliction Warlock
+  {
+    spell: SPELLS.SOUL_LEECH_BUFF,
+  },
+  //Arms Warrior
+  {
+    spell: SPELLS.DEFENSIVE_STANCE_TALENT,
```
Gebuz: Defensive Stance is an active spell so should be tracked in abilities instead

src/common/DEFENSIVE_BUFFS.js
Gebuz: Interesting, I thought that would've just worked with the way array are implemented.

MartijnHols: It probably only picks the first spell when looking for what to put on cd

joshinat0r:
Changed commanding shouts category to utility
Added evasion and reposte for rogues.

src/common/DEFENSIVE_BUFFS.js
joshinat0r: It doesn't show up because ISB has to either have the `isDefensive`-prop or a defensive category.
`buffSpellId`s that dont match those criterias wont show up as defensive buffs.
Setting ISB as defensive would cause an error, I assume because `spellUsable.isAvailable()` doesn't like arrays as parameter.
I could add an exeption for that in the tracker, but not quite sure if its actually usefull to show ISB as "cooldown".

Gebuz: good point, we do expect 100% uptime on it. Only thing would be that it could indicate if the monk was purifying too much if it was not ready. Anyway it sounds like there might be a bug that needs fixing anyway, it should definitely be possible to call `spellUsable.isAvailable()` for spells that use array aswell.

PR #1745, opened 2018-05-27
[Restoration Shaman] Healing Rain tick calculation change

niseko:
Currently the healing rain module divides ticks by looking at the exact millisecond, but looking at logs it appears that ticks can be split up a bit (seen up to 83ms). Often you have 1 hit followed by 5 hits within 5ms, which it currently interprets as 2 different sets of ticks even though its just 1.
I gave it a buffer of 85ms to correct that behaviour, and average ticks went up in all logs as you'd expect, and checking the array i haven't found wrong groupings.

Varimathras rank 1 log where you'd expect 6 hits per tick (before & after):
imageimage
Argus (again you'd expect high amount of hits):
imageimage
Kingaroth:
imageimage

### DIFF
--- a/src/Parser/Shaman/Restoration/Modules/Spells/HealingRain.js
+++ b/src/Parser/Shaman/Restoration/Modules/Spells/HealingRain.js
@@ -8,16 +8,20 @@ import Combatants from 'Parser/Core/Modules/Combatants';
 import SPELLS from 'common/SPELLS';
 import Analyzer from 'Parser/Core/Analyzer';
 
+// 50 was too low, 100 was too high
+// had no issues with 85ms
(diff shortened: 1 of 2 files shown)


MartijnHols:
Modules like this and behavior like that could really benefit from some tests. Would you be willing to add a set for this PR?

niseko:
Seems like its time to learn how tests work, I'll read up and try it out.

niseko:
I added some kind of test, feedback would be appreciated since I've never done something like that

MartijnHols:
Looks good. I recommend trying to test all requirements so that all code paths are hit, in particular the change in this PR.

PR #1746, opened 2018-05-28
[Feral] Snapshot Tracking for Rake

Anatta336:
The snapshotting of DoTs is a unique feral mechanic. This is a snapshot analyzer that keeps track of what applicable buffs were active when rake is cast, marks "bad" casts on the timeline and offers suggestions to fix lost damage from poor management of snapshots.

The generic tracking is handled in Snapshot while the rake-specific parts are in RakeSnapshot. Additional analyzers for the important feral DoTs can be made that also extend Snapshot, but I didn't want to cram too much into a single PR.

RakeSnapshot enforces 2 rules that are recommended by the feral theorycraft community. In order of precedence:
- Only ever refresh a rake buffed by prowl/incarnation if the one you apply would have that buff too.
- When you have a less powerful set of buffs than is already present on the DoT, only refresh within the pandemic window (i.e. less than 30% of the DoT's base duration remaining.)

rakesnapshot01
rakesnapshot02
rakesnapshot03

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
+++ b/src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
@@ -1,0 +1,125 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+import Snapshot, { JAGGED_WOUNDS_MODIFIER } from '../FeralCore/Snapshot';
+
+/*
+Identify inefficient refreshes of the Rake DoT:
+  Refresh of a Rake doing double damage due to Prowl with one that will not do double damage.
+  Early refresh of a Rake doing bonus damage from snapshot buffs with one that will do less damage.
+*/
+
+const RAKE_BASE_DURATION = 15000;
+class RakeSnapshot extends Snapshot {
+  rakeCastCount = 0;        // count of rake casts of all kinds
+  prowlLostCastCount = 0;   // count of rake buffed with Prowl ending early due to refresh without Prowl buff
+  prowlLostTimeSum = 0;     // total time cut out from the end of prowl-buffed rake bleeds by refreshing early (milliseconds)
+  downgradeCastCount = 0;   // count of rake DoTs refreshed with weaker snapshot before pandemic
+
+  hasBloodtalonsTalent = false;
+
+  on_initialized() {
+    this.spellCastId = SPELLS.RAKE.id;
+    this.durationOfFresh = RAKE_BASE_DURATION;
+    this.isProwlAffected = true;
+    this.isTigersFuryAffected = true;
+
+    const combatant = this.combatants.selected;
+    if (combatant.hasTalent(SPELLS.BLOODTALONS_TALENT.id)) {
+      this.isBloodtalonsAffected = true;
+      this.hasBloodtalonsTalent = true;
+    }
+    if (combatant.hasTalent(SPELLS.JAGGED_WOUNDS_TALENT.id)) {
+      this.durationOfFresh *= JAGGED_WOUNDS_MODIFIER;
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    if (SPELLS.RAKE.id === event.ability.guid) {
+      ++this.rakeCastCount;
+    }
+
+    super.on_byPlayer_cast(event);
+  }
+
+  checkRefreshRule(event, stateOld, stateNew) {
+    if (stateOld.prowl && !stateNew.prowl) {
+      // refresh removed a powerful prowl-buffed bleed
+      const timeLost = stateOld.expireTime - event.timestamp;
+      this.prowlLostTimeSum += timeLost;
+      ++this.prowlLostCastCount;
+
+      event.meta = event.meta || {};
+      event.meta.isInefficientCast = true;
+      event.meta.inefficientCastReason = `You lost ${(timeLost / 1000).toFixed(1)} seconds of a Rake empowered with Prowl by refreshing early.`;
+    } else if (stateOld.pandemicTime > event.timestamp &&
+      stateOld.power > stateNew.power &&
+      !stateNew.hasProwl) {
+      // refreshed with weaker DoT before pandemic window - but ignore this rule if the new Rake has Prowl buff as that's more important.
+      ++this.downgradeCastCount;
+
+      if (!event.meta || (!event.meta.isInefficientCast && !event.meta.isEnhancedCast)) {
+        // this downgrade is relatively minor, so don't overwrite output from elsewhere.
+        event.meta = event.meta || {};
+        event.meta.isInefficientCast = true;
+        event.meta.inefficientCastReason = `You refreshed with a weaker version of Rake before the pandemic window.`;
+      }
+    }
+  }
+
+  // seconds of double-damage Rake lost per minute
+  get prowlLostTimePerMinute() {
+    return (this.prowlLostTimeSum / this.owner.fightDuration) * 60;
+  }
+  get downgradeProportion() {
+    return this.downgradeCastCount / this.rakeCastCount;
+  }
+  get prowlLostSuggestionThresholds() {
+    return {
+      actual: this.prowlLostTimePerMinute,
+      isGreaterThan: {
+        minor: 0,
+        average: 1.0,
+        major: 3.0,
+      },
+      style: 'decimal',
+    };
+  }
+  get downgradeSuggestionThresholds() {
+    return {
+      actual: this.downgradeProportion,
+      isGreaterThan: {
+        minor: 0,
+        average: 0.15,
+        major: 0.30,
+      },
+      style: 'percentage',
+    };
+  }
+
+  suggestions(when) {
+    when(this.prowlLostSuggestionThresholds).addSuggestion((suggest, actual, recommended) => {
+      return suggest(
+        <React.Fragment>
+          You sometimes replaced a powerful <SpellLink id={SPELLS.RAKE.id} /> cast with the <SpellLink id={SPELLS.PROWL.id} /> effect with one without it. You ended {this.prowlLostCastCount} empowered <SpellLink id={SPELLS.RAKE.id} /> bleed{this.prowlLostCastCount!==1?'s':''} early.
+        </React.Fragment>
+      )
+        .icon(SPELLS.RAKE.icon)
+        .actual(`${actual.toFixed(1)} seconds of Prowl buffed Rake bleed was lost per minute.`)
+        .recommended(`${recommended} is recommended`);
+    });
+
+    when(this.downgradeSuggestionThresholds).addSuggestion((suggest, actual, recommended) => {
+      return suggest(
+        <React.Fragment>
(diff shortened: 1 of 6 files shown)


src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -0,0 +1,127 @@
...
+
+const TIGERS_FURY_MULTIPLIER = 1.15;
+const PROWL_MULTIPLIER = 2.00; // also applied by Incarnation: King of the Jungle, and Shadowmeld
+const BLOODTALONS_MULTIPLIER = 1.20;
+
+const JAGGED_WOUNDS_MODIFIER = 0.80;  // "[...]deal the same damage as normal but in 20% less time."
+const PANDEMIC_FRACTION = 0.3;
+
+// leeway in ms after loss of bloodtalons/prowl buff to count a cast as being buffed. Keep well below GCD to avoid false positives.
+const BUFF_WINDOW_TIME = 300;
```
Gebuz: Is it common that the delay is that high? Otherwise I'd suggest lowering it to 50 or 100ms, as you can still get false positives from the buffs expirering naturally.

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -0,0 +1,127 @@
...
+  // It's common for buffs that are consumed by an ability to be reported as not present when the ability's event is processed.
+  // So record when the buff wears off, and if that's sufficiently close to the ability being used we can assume it was active.
+  on_byPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if (this.isProwlAffected && (
+        SPELLS.PROWL.id === spellId ||
+        SPELLS.PROWL_INCARNATION.id === spellId || 
+        SPELLS.SHADOWMELD.id === spellId)) {
+      this.prowlFadedAt = event.timestamp;
+    } else if (this.isBloodtalonsAffected && SPELLS.BLOODTALONS_BUFF.id === spellId) {
```
Gebuz: no need to do `else` here

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -0,0 +1,127 @@
...
+      this.bloodtalonsFadedAt = event.timestamp;
+    }
+  }
+
+  on_byPlayer_cast(event) {
```
Gebuz: I recommend using `on_byPlayer_applydebuff` and `on_byPlayer_refreshdebuff` and using the debuffID instead of `on_byPlayer_cast` and castID. Cast might not always be accurate because the debuff might not be applied (immune, dodged, parry, etc.) or removed early (pretty rare in pve but might aswell cover that case). It would also support anything that allows you to apply the bleed to multiple targets like Thrash.

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -0,0 +1,127 @@
...
+
+    const stateNew = {
+      expireTime: expireNew,
+      pandemicTime: expireNew - this.durationOfFresh * PANDEMIC_FRACTION, // after this time the DoT is in the "pandemic window"
+      tigersFury: false,
+      prowl: false,
+      bloodtalons: false,
+      power: 1,
+    };
+    if (this.isProwlAffected && (combatant.hasBuff(SPELLS.INCARNATION_KING_OF_THE_JUNGLE_TALENT.id, event.timestamp) ||
```
Gebuz: You can use `combatant.hasBuff(SPELLS.INCARNATION_KING_OF_THE_JUNGLE_TALENT.id, null, BUFF_WINDOW_TIME)` instead of saving the timestamps for prowl and blood talons. You can read more about .hasBuff here: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/eb8fcfd489ac1bd7c0a695c6657cfe07a0c27885/src/Parser/Core/Entity.js#L46

src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
```diff
@@ -0,0 +1,125 @@
...
+  hasBloodtalonsTalent = false;
+
+  on_initialized() {
+    this.spellCastId = SPELLS.RAKE.id;
+    this.durationOfFresh = RAKE_BASE_DURATION;
+    this.isProwlAffected = true;
+    this.isTigersFuryAffected = true;
+
+    const combatant = this.combatants.selected;
+    if (combatant.hasTalent(SPELLS.BLOODTALONS_TALENT.id)) {
```
Gebuz: No need to do this check as `this.hasBloodtalonsTalent` isn't used and `this.isBloodtalonsAffected` can always be true (if you don't have the talent you just wont ever get the buff).

src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
```diff
@@ -0,0 +1,125 @@
...
+      const timeLost = stateOld.expireTime - event.timestamp;
+      this.prowlLostTimeSum += timeLost;
+      ++this.prowlLostCastCount;
+
+      event.meta = event.meta || {};
+      event.meta.isInefficientCast = true;
+      event.meta.inefficientCastReason = `You lost ${(timeLost / 1000).toFixed(1)} seconds of a Rake empowered with Prowl by refreshing early.`;
+    } else if (stateOld.pandemicTime > event.timestamp &&
+      stateOld.power > stateNew.power &&
+      !stateNew.hasProwl) {
```
Gebuz: if `stateOld.power > stateNew.power` isn't it still a bad refresh even though the new cast has Prowl? For instance if you do a rake with BT, TF and Prowl and then overwrite it right away with a rake only buffed by Prowl.

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
Gebuz: For marking the cast as inefficient on the timeline you can just save the most recent cast at all times (The cast should always come before the debuff apply event).

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
Anatta336: Good point on the expiring buffs causing false positives. Especially for Prowl and Shadowmeld which can be lost through taking damage, which is a real risk on some fights.
Highest delay I could find on a log was 36ms, so I set the window to a hopefully safe 60ms.

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
Anatta336: Thank you, the cast not actually applying a DoT is important to catch.

Using `on_byPlayer_applydebuff` and `on_byPlayer_refreshdebuff` instead of `on_byPlayer_cast` means it loses easy access to the cast event needed to mark inefficient casts on the timeline. So I made some further changes to link the cast event with the DoT's state information. Usually the debuff event occurs after the cast but in some logs it occurs before, so it attempts to make the link during both debuff and cast event handling.

As the "rules" should only be checked after we have the cast event, `checkRuleIfReady` has been added. This also makes sure each DoT application only gets checked once to avoid any double-counting issues. I packed all the information the rule check needs into the DoT's state object, making it easier to call the rule check whenever the cast event is added to it.

src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
Anatta336: Overwriting immediately like that would likely be a damage loss. But a downgrade while still maintaining Prowl is often an overall damage increase. The clearest case is when Incarnation (which gives the Prowl buff effect) is about to wear off.

Let's say the player already has a 2.6 (Prowl, BT, TF) Rake active, with`t`seconds remaining. Incarnation is about to wear off. Should they cast Rake at 2.0 (just Prowl)?

If they cast it, they'll have a 2.0 Rake for 15 seconds plus whatever pandemic they get from the existing, so `15 + min(4.5, t)`
If they don't cast it, they'll have a 2.6 Rake for`t`

We can model Rake for the next 20 second (long enough for the DoTs in question to expire), and assume when they wear off Rake will be immediately replaced by a 1.0 version. This interactive graph shows the scenario:
https://www.desmos.com/calculator/iovjnuz54k
The x axis is the`t`duration that was remaining when we refresh, and y is the total damage done over the modelled period.
A is the original powerful Rake. B is the new one we're considering casting. C is the non-prowl Rake that will fill in when they fade. 
The red line shows total Rake damage if we don't cast the new one, and the blue line shows if we do.

In this scenario it's a gain to refresh with a Rake weaker by both TF and BT so long as the existing one has less than 12 seconds remaining - and Rake's base duration is only 15!

With all that said, a player going from 2.6 to 2.0 on a long duration Rake when they have plenty of time left on Incarnation is probably a mistake. But it would take some more careful analysis to decide for certain.

Anatta336:
Thank you very much for taking the time to review. I've incorporated your suggestions.

> Should you always let your prowl buffed rake expire before applying a new non prowl buffed rake? I don't play feral so I have no idea, just figured it might make sense to refresh it before the last tick to avoid downtime.

Regulars on the Feral Discord have assured me that is it best to let it tick down entirely, but I've become less convinced since trying to confirm it for myself. I still think it's good to leave it ticking as long as possible, but my calculations suggest refreshing with a plain Rake t seconds before a Prowl-buffed Rake wears off (so long as t < pandemic window) is the same damage loss as refreshing t seconds after. So I don't feel confident in using the "always let it tick down" advice in a suggestion. Instead I've made it detect when the player refreshes with >1 second remaining, and suggest that they aim to refresh as close to the expiration time as possible (wording of suggestion, detection, and thresholds changed.) Remembering to refresh soon after it has expired should already be covered by the uptime analyzer.

The reason for the symmetry of effect is that refreshing early replaces a 2.0 DoT with a 1.0, so losing 1.0 damage per second. Refreshing late leaves a gap where there's no damage when there could have been 1.0 damage, so again a loss of 1.0 per second.

The advice to let it tick down is based on the fact that damage is applied in ticks every 3 seconds, not continuously. So refreshing 1 second early and downgrading the final tick has more effect than refreshing 1 second late and (effectively) losing 1/3 of a tick. But once a DoT has been extended through pandemic and especially in BfA where bleeds are hasted, the timing of individual ticks is far more fluid which makes me believe its best modelled as being continuous. (There is an "extra" tick when the DoT expires if the ticks aren't lined up neatly, but its damage is based on the time since the last tick so playing around that should not be significant.)

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+    this.lastDoTCastEvent = event;
+    
+    // attempt to associate with (very) recent debuff apply/refresh
+    // this cast could apply DoTs to multiple targets, so check them all
+    for (const state of this.stateByTarget) {
+      if (!state.castEvent &&
+          event.timestamp - state.startTime < CAST_WINDOW_TIME) {
+        event.castEvent = event;
+        this.checkRuleIfReady(event, state.prev, state);
```
Gebuz: method only takes 1 parameter, I assume you forgot to remove the last two?

Anatta336: Oooh, yikes. I actually forgot to remove the *first* two. It should be:
`this.checkRuleIfReady(state);`
I've pushed a fix into my branch. Will I need to make a separate pull request as this has been merged now?

Gebuz: Yeah you'll have to make a new PR for the fix, I'll merge in as soon as Travis is done :)

Anatta336: Thank you again. Should be ready now #1752

PR #1753, opened 2018-05-29
[Windwalker] Touch of Death / Gale Burst module

Juko8:
billede
There's some other small things added like changes to SPELLS.MONK and the vers food being added. This was just to make sure those things missing weren't causing wrong calculations by this module. There is some rounding error, but as long as decimals are avoided on the % it should be accurate almost all of the time.

### DIFF
--- a/src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
+++ b/src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import Analyzer from 'Parser/Core/Analyzer';
+import statTracker from 'Parser/Core/Modules/StatTracker';
+import StatisticBox, { STATISTIC_ORDER } from 'Main/StatisticBox';
+
+class TouchOfDeath extends Analyzer {
+  static dependencies = {
+    statTracker: StatTracker,
+  };
+
+  expectedBaseDamage = 0;
+  expectedGaleBurst = 0;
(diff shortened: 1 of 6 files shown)


src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
```diff
@@ -0,0 +1,82 @@
...
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS.TOUCH_OF_DEATH.id !== spellId) {
+      return;
+    }
+    this.expectedGaleBurst = 0;
+    const masteryPercentage = this.statTracker.currentMasteryPercentage;
+    const versatilityPercentage = this.statTracker.currentVersatilityPercentage;
+    this.expectedBaseDamage = event.maxHitPoints * 0.5 * (1 + masteryPercentage) * (1 + versatilityPercentage);
```
Gebuz: What's the 0.5? Should make it a constant so it has a name. Same with the 0.3 and 0.1 SEF multipliers below.

Juko8: Touch of Deaths base damage is half of the casters max health, amplified by mastery and versatility. Gale burst is 10% of damage done to the target with Touch of Death, but as mentioned in the comment, is raised to 30% while SEF is active. I can move the 0.5 to be a constant, although i feel that 'HP*0.5' is pretty self-explanatory about what it does. The Gale Burst 0.1 could be moved to a constant and then the SEF multiplier would be either 1x or 3x that.

PR #1759, opened 2018-06-03
[Core] Add healthstone/health pots to spellbook and death recap.

Gebuz:
Also fixed an issue with death recap not working with multi id spells in the spellbook (such as healthstone/pot).
image
image

### DIFF
--- a/src/Main/DeathRecap.js
+++ b/src/Main/DeathRecap.js
@@ -201,15 +201,15 @@ class DeathRecap extends React.PureComponent {
                         </td>
                         <td style={{ width: '20%' }}>
                           {event.buffsUp && event.buffsUp.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.spell instanceof Array ? e.spell[0].id : e.spell.id} />
+                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.id} />
                           )}<br />
                           {event.debuffsUp && event.debuffsUp.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ border: '1px solid red' }} id={e.ability.guid} />
+                            <SpellIcon style={{ border: '1px solid red' }} id={e.id} />
                           )}
                         </td>
                         <td style={{ width: '15%' }}>
                           {event.defensiveCooldowns.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ opacity: e.cooldownReady ? 1 : .2 }} id={e.spell instanceof Array ? e.spell[0].id : e.spell.id} />
+                            <SpellIcon style={{ opacity: e.cooldownReady ? 1 : .2 }} id={e.id} />
                           )}
                         </td>
                       </tr>
--- a/src/Main/DeathRecap.js
+++ b/src/Main/DeathRecap.js
@@ -201,15 +201,15 @@ class DeathRecap extends React.PureComponent {
                         </td>
                         <td style={{ width: '20%' }}>
                           {event.buffsUp && event.buffsUp.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.spell ? e.spell.id : e} />
+                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.spell instanceof Array ? e.spell[0].id : e.spell.id} />
                           )}<br />
                           {event.debuffsUp && event.debuffsUp.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ border: '1px solid red' }} id={e.ability ? e.ability.guid : e} />
+                            <SpellIcon style={{ border: '1px solid red' }} id={e.ability.guid} />
                           )}
                         </td>
                         <td style={{ width: '15%' }}>
                           {event.defensiveCooldowns.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ opacity: e.cooldownReady ? 1 : .2 }} id={e.spell.id} />
+                            <SpellIcon style={{ opacity: e.cooldownReady ? 1 : .2 }} id={e.spell instanceof Array ? e.spell[0].id : e.spell.id} />
                           )}
                         </td>
                       </tr>
--- a/src/Main/DeathRecapTracker.js
+++ b/src/Main/DeathRecapTracker.js
@@ -44,17 +44,17 @@ class DeathRecapTracker extends Analyzer {
     extendedEvent.time = event.timestamp - this.owner.fight.start_time;
 
     const cooldownsOnly = this.cooldowns.filter(e => e.cooldown);
-    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => {
-      return ({...e, cooldownReady: this.spellUsable.isAvailable(e.spell instanceof Array ? e.spell[0].id : e.spell.id)});
-    });
+    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => ({id: e.primarySpell.id, cooldownReady: this.spellUsable.isAvailable(e.primarySpell.id)}));
     if (event.hitPoints > 0) {
-      this.lastBuffs = this.buffs.filter(e => this.combatants.selected.hasBuff(e.buffSpellId) || this.combatants.selected.hasBuff(e.spell instanceof Array ? e.spell[0].id : e.spell.id));
+      this.lastBuffs = this.buffs.filter(e => this.combatants.selected.hasBuff(e.buffSpellId) || this.combatants.selected.hasBuff(e.spell.id))
+        .map(e => ({id: e.buffSpellId || e.spell.id}));
     }
     extendedEvent.buffsUp = this.lastBuffs;
 
     if (!event.sourceIsFriendly && this.enemies.enemies[event.sourceID]) {
-      const sourceHasDebuff = debuff => (!debuff.end || event.timestamp <= debuff.end) && event.timestamp >= debuff.start && debuff.isDebuff && this.buffs.some(e => e.buffSpellId === debuff.ability.guid || (e.spell instanceof Array ? e.spell[0].id : e.spell.id) === debuff.ability.guid);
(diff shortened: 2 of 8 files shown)


src/Main/DeathRecap.js
```diff
@@ -201,15 +201,15 @@ class DeathRecap extends React.PureComponent {
                         </td>
                         <td style={{ width: '20%' }}>
                           {event.buffsUp && event.buffsUp.sort(sortByTimelineIndex).map(e =>
-                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.spell ? e.spell.id : e} />
+                            <SpellIcon style={{ border: '1px solid rgba(0, 0, 0, 0)' }} id={e.spell instanceof Array ? e.spell[0].id : e.spell.id} />
```
MartijnHols: Please use primarySpell instead https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/Modules/Ability.js#L108

src/Main/DeathRecapTracker.js
```diff
@@ -44,14 +44,16 @@ class DeathRecapTracker extends Analyzer {
     extendedEvent.time = event.timestamp - this.owner.fight.start_time;
 
     const cooldownsOnly = this.cooldowns.filter(e => e.cooldown);
-    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => ({...e, cooldownReady: this.spellUsable.isAvailable(e.spell.id)}));
+    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => {
+      return ({...e, cooldownReady: this.spellUsable.isAvailable(e.spell instanceof Array ? e.spell[0].id : e.spell.id)});
```
MartijnHols: Return can be omitted when there’s only a single statement.

src/Parser/Core/Modules/Items/Healthstone.js
```diff
@@ -0,0 +1,42 @@
...
+        suggestion: true,
+      },
+    });
+  }
+
+  on_toPlayer_death(event) {
+    if (!this.spellUsable.isOnCooldown(HEALTHSTONE_SPELLS[0].id)){
+      return;
+    }
+    this.spellUsable.reduceCooldown(HEALTHSTONE_SPELLS[0].id, COOLDOWN_MS - OUT_OF_COMBAT_COOLDOWN_MS);
```
MartijnHols: Death doesn’t drop combat (not since wotlk)

src/Parser/Core/Modules/Items/Healthstone.js
```diff
@@ -0,0 +1,42 @@
+import SPELLS from 'common/SPELLS';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Abilities from 'Parser/Core/Modules/Abilities';
+import SpellUsable from 'Parser/Core/Modules/SpellUsable';
+
+const HEALTHSTONE_SPELLS = [
```
MartijnHols: Could this or prepotion somehow work together? I don’t want to have to maintain two identical lists.

src/Parser/Core/CombatLogParser.js
```diff
@@ -269,6 +271,9 @@ class CombatLogParser {
 
     infernalCinders: InfernalCinders,
     umbralMoonglaives: UmbralMoonglaives,
+
+    //added last so that it includes items that are added to the spellbook on initialization.
+    deathRecapTracker: DeathRecapTracker,
```
MartijnHols: This doesn’t work like this. Import order is determined by dependencies, anything with a single dependency not loaded will be loaded on a later iteration.

src/Parser/Monk/Brewmaster/Modules/Features/Abilities.js
```diff
@@ -69,13 +69,13 @@ class Abilities extends CoreAbilities {
         // id
         spell: [SPELLS.IRONSKIN_BREW, SPELLS.PURIFYING_BREW],
         buffSpellId: SPELLS.IRONSKIN_BREW_BUFF.id,
-        category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+        category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
```
MartijnHols: These seems like a miscategorization, I recommend using isDefensive instead.

src/Parser/Core/Modules/Items/Healthstone.js
```diff
@@ -0,0 +1,42 @@
...
+      category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
+      cooldown: COOLDOWN_MS / 1000, // The cooldown does not start while in combat.
+      castEfficiency: {
+        suggestion: true,
+      },
+    });
+  }
+
+  on_toPlayer_death(event) {
+    if (!this.spellUsable.isOnCooldown(HEALTHSTONE_SPELLS[0].id)){
```
MartijnHols: You can’t expect the first index to always be the same thing.

src/Parser/Core/Modules/Items/Healthstone.js
Gebuz: Deaths still starts Healthstone/pot cooldowns however.

src/Parser/Core/Modules/Items/Healthstone.js
Gebuz: HS/healthpots don't share cooldowns with DPS/mana pots. Shouldn't be any overlaps in the lists.

src/Parser/Core/CombatLogParser.js
Gebuz: The DeathRecapTracker sets the list on IDs to track in it's 'on_initialize'. There's probably a better way to handle it but I didn't want to mess around too much with it, and it seemed to consistently not include Healthstone if the Healthstone module was after it in the dependency list, and likewise always include it when HS module was before it.

src/Parser/Monk/Brewmaster/Modules/Features/Abilities.js
Gebuz: Why? The ability is purely defensive.

src/Main/DeathRecap.js
Gebuz: Not sure why, but this doesn't work :( will try to figure out why tomorrow.

Gebuz:
Turns out the reason .primarySpell didn't work was because sometimes it was only the `spell` that was added instead of the full `abilities` entrance. I changed it to only add the ID (and for cooldowns, also `cooldownReady`) to the buffs, debuffs and defensives array instead of the full abilities/spell/debuff event. Still missing some other stuff.

Gebuz:
Not correctly starts the cooldown at 60s on death:
image
Also calculates max casts correctly:
image

src/Parser/Core/CombatLogParser.js
Gebuz: Okay I'm dumb, just added healthstone as a dependency to the death recap tracker <.<

src/Main/DeathRecapTracker.js
```diff
@@ -44,15 +46,17 @@ class DeathRecapTracker extends Analyzer {
     extendedEvent.time = event.timestamp - this.owner.fight.start_time;
 
     const cooldownsOnly = this.cooldowns.filter(e => e.cooldown);
-    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => ({...e, cooldownReady: this.spellUsable.isAvailable(e.spell.id)}));
+    extendedEvent.defensiveCooldowns = cooldownsOnly.map(e => ({id: e.primarySpell.id, cooldownReady: this.spellUsable.isAvailable(e.primarySpell.id)}));
     if (event.hitPoints > 0) {
-      this.lastBuffs = this.buffs.filter(e => this.combatants.selected.hasBuff(e.buffSpellId) || this.combatants.selected.hasBuff(e.spell.id));
+      this.lastBuffs = this.buffs.filter(e => this.combatants.selected.hasBuff(e.buffSpellId) || this.combatants.selected.hasBuff(e.spell instanceof Array ? e.spell[0].id : e.spell.id))
```
MartijnHols: Inconsistent spell id usage

src/Parser/Core/Modules/Items/Healthstone.js
```diff
@@ -0,0 +1,55 @@
...
+import SpellUsable from 'Parser/Core/Modules/SpellUsable';
+
+const HEALTHSTONE_SPELLS = [
+  SPELLS.HEALTHSTONE,
+  SPELLS.ANCIENT_HEALING_POTION,
+  SPELLS.ASTRAL_HEALING_POTION,
+];
+
+const COOLDOWN_MS = 3600000; // one hour
+const OUT_OF_COMBAT_COOLDOWN_MS = 60000; // one minute
```
MartijnHols: In that case please leave a comment explaining how this works ingame (or preferably add a test for this requirement).

Also if it’s triggered by dying, how about  naming it so that it refers to that instead of leaving combat?

src/Parser/Core/Modules/Items/Healthstone.js
```diff
@@ -0,0 +1,55 @@
...
+    if (!this.spellUsable.isOnCooldown(SPELLS.HEALTHSTONE.id)){
+      return;
+    }
+    const cooldownRemaining = this.spellUsable.cooldownRemaining(SPELLS.HEALTHSTONE.id);
+    //only start cooldown if not already started.
+    if (cooldownRemaining < OUT_OF_COMBAT_COOLDOWN_MS){
+      return;
+    }
+    this.spellUsable.reduceCooldown(SPELLS.HEALTHSTONE.id, cooldownRemaining - OUT_OF_COMBAT_COOLDOWN_MS);
+    if(event.timestamp + OUT_OF_COMBAT_COOLDOWN_MS < this.owner.fight.end_time){
```
MartijnHols: How about putting the future time in a var? With good naming this might become easier to understand. A comment explaining why might also clarify things.

Gebuz:
Changed `this.buffs` to be consistent in what it contains to remove the last `e.spell instanceof Array ? e.spell[0].id : e.spell.id`. It now only contains an object with `id`. I decided to leave it as an object so that more info like stacks etc. can be added in the future.

Also renamed the constants in Healthstone and added more comments.

Gebuz:
Aaaand it should now also calculate maxCasts correctly when HS was not used before a death. Also changed the efficiency thresholds.

PR #1760, opened 2018-06-06
[BFA] [Frost Mage] Frost Mage Updates

Sharrq:
Frost Mage changes for BFA.

-Updated spellbook in Abilities to include cooldown info for everything (even if i dont want a suggestion or the cast counter) - mostly for completeness
- Removed Icy Veins CDR
- Renamed Arctic Gale to Freezing Rain and adjusted its functionality as it only works while frozen orb is up as opposed to always
- Removed Frost Bomb
- Adjusted suggestions in Winter's Chill for precasting to remove the ternary operators now that both Glacial Spike and Ebonbolt are talents (seemed to overcomplicate things trying to adjust the suggestions for both talents)
- Removed the T20 4pc initialization in FrozenOrb CDR, looked to be redundant with the t204pc module
- Adjusted Splitting Ice to also include Ebonbolt in abilities that can split (Still need to figure out the Glacial Spike calculation its doing, but thats for a later time)


src/Parser/Mage/Frost/Modules/Features/FreezingRain.js
```diff
@@ -44,22 +44,22 @@ class ArcticGale extends Analyzer {
   statistic() {
     return (
       <StatisticBox
-        icon={<SpellIcon id={SPELLS.ARCTIC_GALE_TALENT.id} />}
+        icon={<SpellIcon id={SPELLS.FREEZING_RAIN_TALENT.id} />}
         value={`${formatPercentage(this.damagePercent)} %`}
-        label="Arctic Gale damage"
-        tooltip="This is the portion of your total damage attributable to Arctic Gale. This number only considers the damage bonus to Blizzard, and does not factor in extra targets you may have hit due to the increased area of effect."
+        label="Freezing Rain damage"
+        tooltip="This is the portion of your total damage attributable to Freezing Rain. This number only considers the damage bonus to Blizzard, and does not factor in extra targets you may have hit due to the increased area of effect."
```
Gebuz: no longer increases radius afaik

src/Parser/Mage/Frost/Modules/Features/FreezingRain.js
```diff
@@ -44,22 +44,22 @@ class ArcticGale extends Analyzer {
   statistic() {
     return (
       <StatisticBox
-        icon={<SpellIcon id={SPELLS.ARCTIC_GALE_TALENT.id} />}
+        icon={<SpellIcon id={SPELLS.FREEZING_RAIN_TALENT.id} />}
         value={`${formatPercentage(this.damagePercent)} %`}
-        label="Arctic Gale damage"
-        tooltip="This is the portion of your total damage attributable to Arctic Gale. This number only considers the damage bonus to Blizzard, and does not factor in extra targets you may have hit due to the increased area of effect."
+        label="Freezing Rain damage"
```
Gebuz: I don't think this would give a reasonable estimate of the damage contribution in BFA since it does not account for the reduced cast time on AOE, and for single target casting it would replace another cast. I would suggest displaying something else, maybe proc chance & count?

src/Parser/Mage/Frost/Modules/Features/WintersChill.js
```diff
@@ -39,6 +39,7 @@ class WintersChillTracker extends Analyzer {
 
   on_initialized() {
     this.hasGlacialSpike = this.combatants.selected.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id);
+    this.hasEbonbolt = this.combatants.selected.hasTalent(SPELLS.EBONBOLT_TALENT.id);
```
Gebuz: `this.hasGlacialSpike` and `this.hasEbonbolt` are not used

src/Parser/Mage/Frost/Modules/Features/FreezingRain.js
Sharrq: yep, you are right, i didnt see that in the suggestion

src/Parser/Mage/Frost/Modules/Features/FreezingRain.js
Sharrq: what proc are you talking about?

Freezing Rain is a buff that is up while Frozen Orb is up, there are no procs.

Sharrq: what proc are you talking about?

Freezing Rain is a buff that is up while Frozen Orb is up, there are no procs.

Gebuz: whoops for some reason I thought it was only a chance on frozen orb cast. Not sure what to display then, but DPS definitely aint accurate for the reasons mentioned above.

Sharrq: Hmm, the instant cast thing while frozen orb is down is there in legion too, so its not really any different from legion aside from only giving the damage boost while frozen orb is down.

I might just remove the module entirely, ill take another look at it later.

Sharrq:
fixed

PR #1763, opened 2018-06-06
[Feral] Snapshot tracking for Rip

Anatta336:
This analyzer keeps track of the power of the player's Rip bleeds. It provides suggestions when they are downgrading a snapshot or failing to use Bite when they could have extended a good snapshot.

Feral's Rip bleed benefits from snapshotting the effect of Tiger's Fury and Bloodtalons over its full duration, even after the buffs have worn off. Unlike Feral's other DoTs Rip's base damage also varies by how many combo points were used when it was cast. At <25% health, or always with the Sabertooth talent, Rip's duration can be extended with Ferocious Bite. Extending with Bite maintains the existing snapshot effect.

As well as adding RipSnapshot, there's some refactoring in Snapshot (which is now used by RipSnapshot and RakeSnapshot) to expose the power calculation so RipSnapshot can add to it. Plus a few small fixes and style guide things like using `.forEach` when possible.

ripsnapshot01
ripsnapshot02
ripsnapshot03
ripsnapshot04
ripsnapshot05

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
+++ b/src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
@@ -57,7 +57,7 @@ class RipSnapshot extends Snapshot {
    * So it's safe to store the comboLastRip on spendresource then use it in calcPower,
    * which is called by Snapshot in response to debuffapply and debuffrefresh.
    */
-  comboLastRip;
+  comboLastRip = 0;
 
   hasSabertooth = false;
   healthFraction = {};
@@ -122,7 +122,7 @@ class RipSnapshot extends Snapshot {
       return;
     }
 
-    // Bite "refreshes the duration of your Rip on the target" which means setting to the base duration.
+    // Bite sets an existing Rip bleed to the base duration regardless of the current duration - even if that reduces it
     existing.expireTime = event.timestamp + this.durationOfFresh;
     existing.pandemicTime = event.timestamp + this.durationOfFresh * (1.0 - PANDEMIC_FRACTION);
     debug && console.log(`${this.owner.formatTimestamp(event.timestamp)} bite extended rip to ${this.owner.formatTimestamp(existing.expireTime)}`);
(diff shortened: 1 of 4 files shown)


src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * Order of processed events when player casts rip is always:
+   *   on_byPlayer_cast
+   *   on_byPlayer_spendresource
+   *   on_byPlayer_debuffapply, or on_byPlayer_debuffrefresh
+   * So it's safe to store the comboLastRip on spendresource then use it in calcPower,
+   * which is called by Snapshot in response to debuffapply and debuffrefresh.
+   */
+  comboLastRip;
```
Gebuz: Please always give variables an initial value (0 or null is fine).

src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
```diff
@@ -0,0 +1,237 @@
...
+    const target = encodeTargetString(event.targetID, event.targetInstance);
+    if (!this.biteCanRefresh(target)) {
+      return;
+    }
+    const existing = this.stateByTarget[target];
+    if (!existing || existing.expireTime < event.timestamp) {
+      return;
+    }
+
+    // Bite "refreshes the duration of your Rip on the target" which means setting to the base duration.
```
Gebuz: Is this always the case? Say I cast a rip before a pandemic window so I have a 1.3\*duration rip up and it's duration is still > 1\*duration, will it reduce the duration to 1\*duration?

src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
```diff
@@ -0,0 +1,237 @@
...
+    if (this.biteCanRefresh(target) &&
+        stateOld.power >= stateNew.power) {
+      this.shouldBeBiteCount += 1;
+      event.meta = event.meta || {};
+      event.meta.isInefficientCast = true;
+      const strengthComment = (stateOld.power > stateNew.power) ? 'a stronger' : 'the same strength';
+      event.meta.inefficientCastReason = `Used Rip when you could have extended using Ferocious Bite and kept ${strengthComment} snapshot.`;
+    }
+    
+    if (stateOld.pandemicTime > stateNew.startTime &&
```
Gebuz: can check for the opposite and return early here to avoid the nested if statements

src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
```diff
@@ -0,0 +1,237 @@
...
+        stateOld.power > stateNew.power) {
+      this.downgradeCount += 1;
+
+      // this downgrade is relatively minor, so don't overwrite cast info from elsewhere
+      if (!event.meta || (!event.meta.isInefficientCast && !event.meta.isEnhancedCast)) {
```
Gebuz: can check for the opposite and return early here to avoid the nested if statements

Anatta336:
Thanks very much for taking the time to review. I've fixed those issues.

Surprisingly the Bite "extend" can reduce the duration if it's currently above the base. I just double-checked on target dummies <25% health and at higher health with the Sabertooth talent. I've amended the comment there to point out the counterintuitive behaviour.

In practice it's rarely something that needs to be played around, at least on single target. If you're able to extend with Bite usually you'll be extending it again before it gets close to expiring so losing a few seconds of duration ends up not mattering.

Anatta336:
Thanks, Travis. We got there in the end.

PR #1764, opened 2018-06-06
[Core][Bug] Firefox dropdown fix

joshinat0r:
firefox has apparently some issue with rendering, not quite sure whats causing that
Replaced the native <select> element with the same library for the realm-search


joshinat0r:
Zero fixed it already

Issue #1668, opened 2018-05-06
Send people to another subdomain when parsing a log from a different patch


MartijnHols:
E.g.
on www: -> bfa for BFA dungeons
on bfa -> www for Legion dungeons

In the future:
www -> legion for Legion dungeons

Extra:
Also recognize having an artifact weapon or whatever will be unique for the prepatch and send people to a future pre-prepatch subdomain.

MartijnHols:
I think this should be solved by looking at talents and / or traits;

- Legion logs have Legion traits and talents active
- Prepatch logs have BFA traits and talents active
- BFA logs have BFA traits and talents active

Looking at instance would be unreliable since you can go back and do content.
Looking at date would be unreliable since different zones get different patch release dates.

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=58014862)](https://www.bountysource.com/issues/58014862-send-people-to-another-subdomain-when-parsing-a-log-from-a-different-patch?utm_source=58014862&utm_medium=shield&utm_campaign=ISSUE_BADGE)

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=58014862)](https://www.bountysource.com/issues/58014862-send-people-to-another-subdomain-when-parsing-a-log-from-a-different-patch?utm_source=58014862&utm_medium=shield&utm_campaign=ISSUE_BADGE)

ndudnicz:
Since BFA will be released at the exact same date and time in every zones [(battle net post)](https://worldofwarcraft.com/en-us/news/21849061/battle-for-azeroth-one-launch-to-rule-them-all), BFA logs timestamps would also be > 1534197600 (08/14/2018 12:00AM CEST).

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=58014862)](https://www.bountysource.com/issues/58014862-send-people-to-another-subdomain-when-parsing-a-log-from-a-different-patch?utm_source=58014862&utm_medium=shield&utm_campaign=ISSUE_BADGE) (increased)

sahalsaad:
Hi, I interested working on this. May I know where should I start? Thanks!

sahalsaad:
Hi @MartijnHols ,

Sorry, I'm not wow player. I don't think I can work on this since I lack of testing data(report link) unless there is somewhere I can get the report link.

[...]

PR #1766, opened 2018-06-08
[BfA] Add Tracker for First Mate's Spyglass Trinket

HighTechLowLife:
My first time doing this, GitHub and all, please be gentle.


MartijnHols:
Hmmm Travis has been having issues with npm these last few days, restarted it to see if we can get it green.

src/Parser/Core/Modules/Items/BFA/FirstMatesSpyglass.js
```diff
@@ -0,0 +1,73 @@
...
+  
+  casts = 0;
+  timesDamageCrit = 0;
+  timesHealCrit = 0;
+    
+  on_initialized() {
+    this.active = this.combatants.selected.hasTrinket(
+      ITEMS.FIRST_MATES_SPYGLASS.id
+    );
+  }
```
MartijnHols: For activateable items please also add an entry to the abilities module so it shows in the timeline, cast efficiency and as a possible suggestion. Here's an example of where this is done: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/c9f720358b547ebf2d407efc421ed6bdf3176a29/src/Parser/Core/Modules/Items/Legion/TombOfSargeras/DeceiversGrandDesign.js#L43-L54

src/Parser/Core/Modules/Items/BFA/FirstMatesSpyglass.js
```diff
@@ -0,0 +1,73 @@
...
+  
+  get totalBuffUptime() {
+      return this.combatants.selected.getBuffUptime(SPELLS.SPYGLASS_SIGHT.id) / this.owner.fightDuration;
+  }
+  
+  item() {
+    return {
+      item: ITEMS.FIRST_MATES_SPYGLASS,
+      result: (
+        <dfn data-tip={`You critically damaged ${this.timesDamageCrit} times with this buff up<br/>
```
MartijnHols: Did you have any purpose in mind for this information? I think users can't really do anything with this information and so it seems unnecessary to me. We already present a huge amount of data (in most specs), so we need to keep it focused on actionable and interesting info.

HighTechLowLife: I was thinking that seeing the number of critical hits that happened while using an item that buffed your crit would be an interesting thing to note, kind of showing an incentive to use the item as much as possible when you want crits. 

Maybe changing it to "x times out of y hits," or a straight up percentage, would make it better? I'm not sure. I was just looking to see if there was any extra information someone might want out of using the buff.

MartijnHols: Showing the average amount of critical strike hits as a percentage of total hits could be interesting. 👍 

The most information (and the goal for all items) would be to show actual dps/hps gain, but unfortunately for items that provide stat rating buffs that requires the StatValues module to be implemented for specs and only a few have that. The next best thing we have for a bunch of items is the average stat gain, such as for [Aggramar's Conviction](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/c19e5ffa3404213059cd17f86a23c3dfe501f0e0/src/Parser/Core/Modules/Items/Legion/AntorusTheBurningThrone/AggramarsConviction.js#L82) and [Sephuz's Secret](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/c19e5ffa3404213059cd17f86a23c3dfe501f0e0/src/Parser/Core/Modules/Items/Legion/Legendaries/SephuzsSecret.js#L34).

src/Parser/Core/Modules/Items/BFA/FirstMatesSpyglass.js
```diff
@@ -14,16 +15,29 @@ import { formatPercentage } from 'common/format';
...
   on_initialized() {
     this.active = this.combatants.selected.hasTrinket(
       ITEMS.FIRST_MATES_SPYGLASS.id
     );
+    if (this.active) {
+      this.abilities.add({
+        spell: SPELLS.SPYGLASS_SIGHT,
+        name: ITEMS.FIRST_MATES_SPYGLASS.name,
+        category: Abilities.SPELL_CATEGORIES.ITEMS,
+        charges: 1,
```
MartijnHols: This property can be omitted when there are no extra charges (1 is default).

src/Parser/Core/Modules/Items/BFA/FirstMatesSpyglass.js
```diff
@@ -38,19 +52,21 @@ class FirstMatesSpyglass extends Analyzer {
       return;
     }
     if (event.hitType !== HIT_TYPES.CRIT) {
-      return;
+      this.timesHit += 1;
```
MartijnHols: The way you're calculating the crit percentage will lead to incorrect numbers; it will mark spells as crit when it didn't crit, and count crits as 2 hits and 1 crit.

MartijnHols:
Hey @HighTechLowLife! I think this is your first PR to WoWAnalyzer, thanks a lot for your contribution and welcome on board! It looks like you're getting comfortable with adding analyzers, if you're interested there are a few more in the [BFA](https://github.com/WoWAnalyzer/WoWAnalyzer/issues?q=is%3Aissue+is%3Aopen+label%3ABFA) label that you could work on. If you're looking for a challenge, the Darkmoon Deck trinkets are probably not easy to implement or #1668 might peak your interest.

Your changes should be live on wowanalyzer.com by now, please let us know if you need anything further.

PR #1767, opened 2018-06-09
[Feral] Snapshot tracking for Moonfire

Anatta336:
Moonfire (granted to Feral only with the Lunar Inspiration talent) is the simplest of the snapshot-benefiting abilities as it only benefits from the damage bonus of Tiger's Fury, and unlike Rip has no other mechanics affecting its base damage.

The same rule applies for Moonfire as the rest: only refresh before pandemic if you would upgrade the snapshot. Using the Snapshot class shared with RakeSnapshot and RipSnapshot, this analyzer checks for the combatant "downgrading" their Moonfire DoTs. Inefficient casts are marked on the timeline and a suggestion is generated if they do it often.

moonfiresnapshot01
moonfiresnapshot02

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
+++ b/src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
@@ -1,0 +1,99 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+
+import Snapshot, { PANDEMIC_FRACTION } from '../FeralCore/Snapshot';
+
+/**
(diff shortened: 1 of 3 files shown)


src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
```diff
@@ -0,0 +1,99 @@
...
+    
+    if (stateNew.startTime >= stateOld.pandemicTime ||
+        stateNew.power >= stateOld.power) {
+      // good refresh
+      return;
+    }
+    
+    ++this.downgradeCastCount;
+    
+    // this downgrade is relatively minor, so don't overwrite cast info from elsewhere
```
Gebuz: Doesn't look like you're setting it anywhere else so you could leave this part out

PR #1768, opened 2018-06-09
[Death Knight] Fix runes generated/wasted from passive regn.

Gebuz:
It was not correctly accounting for runes that was instantly refunded through abilities such as Runic Empowerment. The percent shown in the stat box is not affected.
Before:
image
After:
image


src/Parser/DeathKnight/Shared/RuneTracker.js
```diff
@@ -247,15 +245,24 @@ class RuneTracker extends ResourceTracker {
...
-      runeMaxCasts.push(this.castEfficiency.getCastEfficiencyForSpellId(spellId).maxCasts);
-    });
-    const totalCasts = runeMaxCasts.reduce((accumulator, currentValue) => accumulator + currentValue);
-    return totalCasts;
+    let totalCasts = 0;
+    for (const spender in this.spendersObj) { //add runes spent
+      totalCasts += this.spendersObj[spender].spent;
+    }
+    for (const builder in this.buildersObj) { //subtract runes generated by other sources
+      const asInt = parseInt(builder, 10);
```
MartijnHols: Please use Number(x) instead of parseInt.

PR #1770, opened 2018-06-10
[Feral] Rake normalizer, minor changes in Snapshots

Anatta336:
This adds a RakeBleed normalizer which ensures the `applydebuff` or `refreshdebuff` for Rake's bleed always appear after that Rake's `cast` event. With this normalizer in place the code in Snapshot that handled them being out of order is now removed. After checking with a lot of logs it's only ever Rake that had the events in the wrong order, and only when used from stealth.

I also made some minor changes to the Snapshot, RakeSnapshot, RipSnapshot, and MoonfireSnapshot modules from my recent pull requests.
- `castCount` is now maintained on Snapshot rather than having near identical code on each subclass.
- The threshold for "major" suggestions about snapshot downgrading has been significantly pushed up. With current Feral tuning even completely messing up the snapshot mechanic is a fairly small overall damage loss. This is expected to change in BfA, when the thresholds should be adjusted again. 
- Various tidying up I couldn't resist, initialising variables and so on.

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
+++ b/src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
@@ -15,7 +15,6 @@ import Snapshot, { PANDEMIC_FRACTION } from '../FeralCore/Snapshot';
 const MOONFIRE_FERAL_BASE_DURATION = 14000;
 
 class MoonfireSnapshot extends Snapshot {
-  moonfireCastCount = 0;
   downgradeCastCount = 0;
 
   on_initialized() {
@@ -34,13 +33,7 @@ class MoonfireSnapshot extends Snapshot {
 
     // bloodtalons only affects melee abilities
     this.isBloodtalonsAffected = false;
-  }
-
-  on_byPlayer_cast(event) {
-    if (SPELLS.MOONFIRE_FERAL.id === event.ability.guid) {
-      ++this.moonfireCastCount;
-    }
-    super.on_byPlayer_cast(event);
+    super.on_initialized();
   }
 
   checkRefreshRule(stateNew) {
@@ -56,7 +49,7 @@ class MoonfireSnapshot extends Snapshot {
       return;
     }
     
-    ++this.downgradeCastCount;
+    this.downgradeCastCount += 1;
     
     // this downgrade is relatively minor, so don't overwrite cast info from elsewhere
     const event = stateNew.castEvent;
@@ -69,14 +62,14 @@ class MoonfireSnapshot extends Snapshot {
   }
 
   get downgradeProportion() {
-    return this.downgradeCastCount / this.moonfireCastCount;
+    return this.downgradeCastCount / this.castCount;
   }
   get downgradeSuggestionThresholds() {
     return {
       actual: this.downgradeProportion,
       isGreaterThan: {
         minor: 0,
-        average: 0.30,
+        average: 0.15,
         major: 0.60,
       },
       style: 'percentage',
--- a/src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
+++ b/src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
@@ -4,23 +4,29 @@ import SpellLink from 'common/SpellLink';
 import { formatPercentage } from 'common/format';
 import Snapshot, { JAGGED_WOUNDS_MODIFIER, PANDEMIC_FRACTION } from '../FeralCore/Snapshot';
 
-/*
-Identify inefficient refreshes of the Rake DoT:
-  Early refresh of a Rake doing double damage due to Prowl with one that will not do double damage.
-  Pre-pandemic refresh of a Rake doing bonus damage from snapshot buffs with one that will do less damage.
-*/
+/**
+ * Identify inefficient refreshes of the Rake DoT:
+ *  Early refresh of a Rake doing double damage due to Prowl with one that will not do double damage.
+ *  Pre-pandemic refresh of a Rake doing bonus damage from snapshot buffs with one that will do less damage.
+ */
 
-// When you cannot refresh a prowl-buffed rake with prowl, ideally you'd let it tick down. At the
-// moment that it expires apply your new one.
-// Looking for exact timing is unrealistic, so give some leeway.
+/**
+ * When you cannot refresh a prowl-buffed rake with prowl, ideally you'd let it tick down.
+ * Then at the moment that it expires you'd apply a fresh DoT.
+ * But exact timing is unrealistic, so give some leeway.
+ */
 const FORGIVE_PROWL_LOSS_TIME = 1000;
 
 const RAKE_BASE_DURATION = 15000;
 class RakeSnapshot extends Snapshot {
-  rakeCastCount = 0;        // count of rake casts of all kinds
-  prowlLostCastCount = 0;   // count of rake buffed with Prowl ending early due to refresh without Prowl buff
-  prowlLostTimeSum = 0;     // total time cut out from the end of prowl-buffed rake bleeds by refreshing early (milliseconds)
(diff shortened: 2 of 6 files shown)


src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
```diff
@@ -34,13 +33,7 @@ class MoonfireSnapshot extends Snapshot {
...
     // bloodtalons only affects melee abilities
     this.isBloodtalonsAffected = false;
-  }
-
-  on_byPlayer_cast(event) {
-    if (SPELLS.MOONFIRE_FERAL.id === event.ability.guid) {
-      ++this.moonfireCastCount;
-    }
-    super.on_byPlayer_cast(event);
+    super.on_initialized();
```
MartijnHols: super calls should generally be the first statement of a method.

src/Parser/Druid/Feral/Modules/Bleeds/MoonfireSnapshot.js
```diff
@@ -56,7 +49,7 @@ class MoonfireSnapshot extends Snapshot {
       return;
     }
     
-    ++this.downgradeCastCount;
+    this.downgradeCastCount += 1;
```
MartijnHols: 👍

src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
```diff
@@ -33,13 +39,7 @@ class RakeSnapshot extends Snapshot {
...
       this.durationOfFresh *= JAGGED_WOUNDS_MODIFIER;
     }
-  }
-
-  on_byPlayer_cast(event) {
-    if (SPELLS.RAKE.id === event.ability.guid) {
-      ++this.rakeCastCount;
-    }
-    super.on_byPlayer_cast(event);
+    super.on_initialized();
```
MartijnHols: super calls should generally be the first statement of a method.

src/Parser/Druid/Feral/Modules/Bleeds/RipSnapshot.js
```diff
@@ -75,14 +75,11 @@ class RipSnapshot extends Snapshot {
     if (combatant.hasTalent(SPELLS.SABERTOOTH_TALENT.id)) {
       this.hasSabertooth = true;
     }
+    super.on_initialized();
```
MartijnHols: super calls should generally be the first statement of a method.

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
-        event.castEvent = event;
-        this.checkRuleIfReady(state);
-      }
-    });
+  }
+
+  on_initialized() {
+    if (!this.spellCastId || !this.debuffId) {
+      this.active = false;
+      debug && console.warn('Snapshot missing spellCastId or debuffId.');
```
MartijnHols: This looks like a coding mistake, not something that can randomly occur. If that's the case I believe this should `throw new Error(...)` instead so it's more easily caught by developers.

src/Parser/Druid/Feral/Modules/Bleeds/RakeSnapshot.js
```diff
@@ -4,23 +4,29 @@ import SpellLink from 'common/SpellLink';
...
+  prowlLostCastCount = 0;
+
+  // total time cut out from the end of prowl-buffed rake bleeds by refreshing early (milliseconds)
+  prowlLostTimeSum = 0;
+
+  // rake DoTs refreshed with weaker snapshot before pandemic
+  downgradeCastCount = 0;
 
   on_initialized() {
     this.spellCastId = SPELLS.RAKE.id;
```
MartijnHols: I think you should set this as a class property instead. Also since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
   };
 
   // extending class should fill these in:
-  spellCastId;
-  debuffId;
-  isProwlAffected;
-  isTigersFuryAffected;
-  isBloodtalonsAffected;
-  durationOfFresh;
+  spellCastId = null;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
+  debuffId = null;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
+  isProwlAffected = false;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
+  isTigersFuryAffected = false;
+  isBloodtalonsAffected = false;
+  durationOfFresh = false;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
-  debuffId;
-  isProwlAffected;
-  isTigersFuryAffected;
-  isBloodtalonsAffected;
-  durationOfFresh;
+  spellCastId = null;
+  debuffId = null;
+  isProwlAffected = false;
+  isTigersFuryAffected = false;
+  isBloodtalonsAffected = false;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
```diff
@@ -42,30 +40,31 @@ class Snapshot extends Analyzer {
...
-  spellCastId;
-  debuffId;
-  isProwlAffected;
-  isTigersFuryAffected;
-  isBloodtalonsAffected;
-  durationOfFresh;
+  spellCastId = null;
+  debuffId = null;
+  isProwlAffected = false;
+  isTigersFuryAffected = false;
```
MartijnHols: Since it never changes, how about making it static instead?

src/Parser/Druid/Feral/Modules/FeralCore/Snapshot.js
MartijnHols: Please avoid changing types of vars as much as possible. A better default value for this would be `null`.

src/Parser/Druid/Feral/Modules/Normalizers/RakeBleed.js
```diff
@@ -0,0 +1,45 @@
+import SPELLS from 'common/SPELLS';
+import EventsNormalizer from 'Parser/Core/EventsNormalizer';
+
+const CAST_WINDOW = 100;
+class RakeBleed extends EventsNormalizer {
```
MartijnHols: This looks good, but please consider adding a test for this class. Classes like Normalizers are very important to work well and so a test suite ensuring their behavior could prevent issues in the future.

Anatta336:
Thank you very much for taking the time to the review. I'm putting together a PR now with a test for the normalizer and the suggested changes elsewhere.

PR #1773, opened 2018-06-11
[BFA] Updates for Builds 26734 - 26806

niseko:
Some broader changes as its been quite a few builds.

- Reverted Berserking because blizzard likes to revert stuff
- Added the first few Azerite trait stat scalings that i was able to test
- Added the Spirit Wolf buff as a defensive
- Added versatility scaling to Healthstones and Healthpots after confirming it

#### Resto Shaman
- Removed Crashing Waves module as the talent got removed
- Added a Flash Flood module for the new talent
- Added a statistic box to Downpour to display the average cooldown time, which pretty much shows how effective the target selection was

image
image


src/Parser/Shaman/Restoration/Modules/Abilities.js
```diff
@@ -13,7 +12,7 @@ class Abilities extends CoreAbilities {
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
         charges: 2,
```
Gebuz: instead of having 2 entries for riptide you could set the number of charges based on the enabled logic (like you do with healing stream below)

src/Parser/Shaman/Restoration/Modules/Talents/EarthenWallTotem.js
```diff
@@ -79,6 +79,7 @@ class EarthenWallTotem extends Analyzer {
 
   statistic() {
     const casts = this.abilityTracker.getAbility(SPELLS.EARTHEN_SHIELD_TOTEM_TALENT.id).casts;
+    this.earthenShieldEfficiency = this.healing / this.potentialHealing;
```
Gebuz: Should make earthenShieldEfficiency a getter instead of a var since it relies on two other vars.

src/Parser/Shaman/Restoration/Modules/Talents/FlashFlood.js
```diff
@@ -0,0 +1,106 @@
...
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatTracker from 'Parser/Core/Modules/StatTracker';
+import calculateEffectiveHealing from 'Parser/Core/calculateEffectiveHealing';
+
+import Abilities from '../Abilities';
+
+const BUFFER = 750;
+const flashFloodHaste = 0.2;
+const tidalWavesHaste = 0.3;
```
Gebuz: constants should be all caps with underscores: `TIDAL_WAVES_HASTE`. Same goes for `flashFloodHaste` & `spellsConsumingFlashFlood`

src/Parser/Shaman/Restoration/Modules/Talents/FlashFlood.js
```diff
@@ -0,0 +1,106 @@
...
+
+// All heal spells with cast time...
+const spellsConsumingFlashFlood = [
+  SPELLS.HEALING_WAVE.id,
+  SPELLS.HEALING_SURGE_RESTORATION.id,
```
Gebuz: Does Flash Flood reduce the GCD aswell? Otherwise it would be wasted on surge. If it does, remember to also extend the GlobalCooldown module and reduce the GCD.

src/Parser/Shaman/Restoration/Modules/Talents/FlashFlood.js
```diff
@@ -0,0 +1,106 @@
...
+    this.active = this.combatants.selected.hasTalent(SPELLS.FLASH_FLOOD_TALENT.id) || this.combatants.selected.hasFinger(ITEMS.SOUL_OF_THE_FARSEER.id);
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+
+    if(this.healingRainTimestamp && spellId === SPELLS.HEALING_RAIN_HEAL.id && 
+      this.healingRainTimestamp + this.abilities.getExpectedCooldownDuration(SPELLS.HEALING_RAIN_CAST.id) <= event.timestamp) {
+      const flashFloodContribution = flashFloodHaste / (flashFloodHaste + this.healingRainHastePercent);
+      this.healing += calculateEffectiveHealing(event, flashFloodContribution);
```
Gebuz: Does Flash Flood increase tick rate? Tooltip would suggest that it doesn't (cast time reductions usually doesn't affect number of ticks).

src/Parser/Shaman/Restoration/Modules/Talents/FlashFlood.js
```diff
@@ -0,0 +1,106 @@
...
+import { formatPercentage } from 'common/format';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Combatants from 'Parser/Core/Modules/Combatants';
+import StatTracker from 'Parser/Core/Modules/StatTracker';
+import calculateEffectiveHealing from 'Parser/Core/calculateEffectiveHealing';
+
+import Abilities from '../Abilities';
+
+const BUFFER = 750;
```
Gebuz: This seems like a way too big buffer, any reason it is so high? Having it at GCD cap could catch a lot of false positives at high haste.

src/Parser/Shaman/Restoration/Modules/Talents/FlashFlood.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  subStatistic() {
+    return (
+      <div className="flex">
+        <div className="flex-main">
+          <SpellLink id={SPELLS.FLASH_FLOOD_TALENT.id} />
+        </div>
+        <div className="flex-sub text-right">
+          {formatPercentage(this.owner.getPercentageOfTotalHealingDone(this.healing))} %
```
Gebuz: Considering the implications of mana, cooldown spells and tidal waves not scaling with cast time reduction I would consider if showing "time saved" (in seconds) would be a more appropriate stat to show instead of healing. Assuming any time saved would be filled with equally strong spells is heavily overestimating its value.

src/Parser/Shaman/Restoration/Modules/Talents/Downpour.js
```diff
@@ -49,12 +55,43 @@ class Downpour extends Analyzer {
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.DOWNPOUR_TALENT.id} />}
+        value={`${downpourAverageCooldown.toFixed(1)} seconds`} 
+        label={(
+          <dfn data-tip={`
+            You cast a total of ${downpourCasts} Downpours, which on average hit ${(downpourAverageHits + downpourAverageOverhealedHits).toFixed(1)} out of 6 targets. <br /> 
+            Of those hits, ${downpourAverageHits.toFixed(1)} had effective healing and increased the cooldown.
+          `}>
+            Average Downpour cooldown
```
Gebuz: Would it be more intuitive to show average number of effective hits as the primary value instead? Maybe it's just cause I don't play shaman but to me the average cooldown doesn't say much about how effective the usage was (other than higher -> better)

niseko:
Fixed all suggestions apart from the one i commented on.

New Flash Flood: 
image

src/Parser/Shaman/Restoration/Modules/Talents/Downpour.js
niseko: My issue with Downpour is that higher isn't necessarily better, apart from mana usage and the small base cooldown punishment. Just showing raw hits wouldn't tell you the full story as **effective** hits is what impacts the cooldown time. Showing only effective hits could be confusing as all others (chain heal, healing rain, gift of the queen) show raw hits, and showing both is hard to read.
Having it be the average cooldown time with more information on hover is the best compromise i was able to come up with.

PR #1777, opened 2018-06-12
[Feral] Statistic for uptime of snapshot on DoTs

Anatta336:
Adds tick and damage counting to the existing Snapshot analyzer, and uses that to generate statistics for how each DoT benefited from the snapshot buffs.

How much a DoT benefited is measured by counting the damage tick events and checking what snapshot buffs were active for the DoT on that target at the time. The increase in damage due to each buff is also calculated from the tick. It only deals with the DoT ticks, not the initial damage that Rake and Moonfire also apply.

The % of ticks benefiting from each snapshot buff could be a useful measure in a future checklist for the spec. Although care should be taken as there's quite a lot of variability in what's possible even on a simple fight depending on legendaries and talents.

snapshotuptimestats01
snapshotuptimestats02
snapshotuptimestats03
snapshotuptimestats04
snapshotuptimestats05

### DIFF
--- a/src/Parser/Druid/Feral/CHANGELOG.js
+++ b/src/Parser/Druid/Feral/CHANGELOG.js
@@ -6,6 +6,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-06-11'),
+    changes: <React.Fragment>Added statistics breaking down snapshot uptime by buff for <SpellLink id={SPELLS.RAKE.id} />, <SpellLink id={SPELLS.RIP.id} />, and <SpellLink id={SPELLS.MOONFIRE_FERAL.id} />.</React.Fragment>,
+    contributors: [Anatta336],
+  },
+  {
     date: new Date('2018-06-09'),
     changes: <React.Fragment>Added snapshot tracking for <SpellLink id={SPELLS.MOONFIRE_FERAL.id} /></React.Fragment>,
     contributors: [Anatta336],
--- a/src/Parser/Druid/Feral/CombatLogParser.js
+++ b/src/Parser/Druid/Feral/CombatLogParser.js
@@ -12,7 +12,6 @@ import RipUptime from './Modules/Bleeds/RipUptime';
 import FerociousBiteEnergy from './Modules/Features/FerociousBiteEnergy';
 import RakeSnapshot from './Modules/Bleeds/RakeSnapshot';
 import RipSnapshot from './Modules/Bleeds/RipSnapshot';
-import MoonfireSnapshot from './Modules/Bleeds/MoonfireSnapshot';
 
 import ComboPointTracker from './Modules/ComboPoints/ComboPointTracker';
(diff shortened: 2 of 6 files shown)


Gebuz:
Looks good! There is one thing I would suggest changing and that is how you calculate the damage contributed by each buff. 
WARNING: MATH BELOW. 

Since they are multiplicative contributing the damage to each buff as if it were the last will inflate the numbers substantially. For instance if you had both prowl and blood talons active, blood talons would be noted down for 16.67% dmg and prowl for 50% dmg, leaving only 33.33% for base damage (which should be 1/2.4 = 41.67%).

A better way to approach this were the damage is more evenly split between the modifiers would be to contribute the total bonus dmg based on the percent it contributed. So if we have both BT and prowl for a total of 120% (additive), BT should get 20/120 = 16.67% and prowl should get 100/120 = 83.33%.

This would give the following numbers:
BT: 9.72% = (1.4 * 20 / 120) / 2.4
prowl: 48.61% (1.4 * 100 / 120) / 2.4
base dmg: 41.67% = 1 / 2.4
which sums to 100%

```
// moved the modifier calculation out so rip can overwrite it to include combo points without altering 
// the modifier calculation which we need for getting the modifier without combo points.
  calcPower(stateNew) {
    return calcPowerModifiers(stateNew);
  }

// without combo points so it can be used for just the modifier
  calcPowerModifiers(stateNew) {
    let power = 1.0;
    if (stateNew.prowl) {
      power *= PROWL_MULTIPLIER;
    }
    if (stateNew.tigersFury) {
      power *= TIGERS_FURY_MULTIPLIER;
    }
    if (stateNew.bloodtalons) {
      power *= BLOODTALONS_MULTIPLIER;
    }
    return power;
  }

// additive bonus damage for splitting damage contribution between modifiers
  additivePowerSum(stateNew) {
    let power = 0;
    if (stateNew.prowl) {
      power += PROWL_MULTIPLIER - 1;
    }
    if (stateNew.tigersFury) {
      power += TIGERS_FURY_MULTIPLIER - 1;
    }
    if (stateNew.bloodtalons) {
      power += BLOODTALONS_MULTIPLIER - 1;
    }
    return power;
  }
...

// line 95
const increase = calcPowerModifiers(state) - 1; // the one is added in calculateEffectiveDamage
const bonusDamage =  calculateEffectiveDamage(event, increase); // does the bonus dmg calculation for you, see link below
const additiveBonuses = additiveBonusSum(state)

// line 109
const percentContribution = (PROWL_MULTIPLIER - 1) / additiveBonuses;
this.damageFromProwl += bonusDamage * percentContribution;
```

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/43db0c73087681bf0a1fb3c77c187ee8532a6978/src/Parser/Core/calculateEffectiveDamage.js#L1

Anatta336:
Thank you again for the review, and the well explained improvement to the maths!

PR #1779, opened 2018-06-13
[BFA] [Fire Mage] Fire Mage Changes

Sharrq:
Here is the first round of Fire Mage Changes for BFA Prepatch

- Updated Spellbook with GCD Information
- Added Arcane Intellect and Remove Curse to Spellbook
- Removed Cinderstorm
- Removed Phoenix Flames Cooldown Reduction (Was Artifact Trait)
- Renamed instances of PHOENIXS_FLAMES to PHOENIX_FLAMES_TALENT
- Removed link to Unstable Magic in Changelog
- Updated CombustionCharges to account for Phoenix Flames being a talent
- Removed PreIgnited Trait from CombustionMarqueeBindings


src/Parser/Mage/Fire/Modules/Features/Abilities.js
```diff
@@ -33,6 +46,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.FIRE_BLAST,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        isOnGCD: false,
```
Gebuz: it defaults to false so no need to add it

src/Parser/Mage/Fire/Modules/Features/Abilities.js
```diff
@@ -33,6 +46,7 @@ class Abilities extends CoreAbilities {
...
         cooldown: haste => 12 / (1 + haste),
         charges: 2,
```
Gebuz: instead of having 2 entries for Fireblast you could set the number of charges based on the current enabled logic.

src/Parser/Mage/Fire/Modules/Features/Abilities.js
```diff
@@ -73,26 +90,22 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.FLAMESTRIKE,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
+        isOnGCD: true,
+        castEfficiency: {
```
Gebuz: it is disabled by default so no need to include that

src/Parser/Mage/Fire/Modules/Features/Abilities.js
Sharrq: i know it defaults to false, but for the sake of completeness i opted for including it anyway so at a glance you could look at the spellbook and get all the info about that spell including that its off the gcd.

Granted you could just assume that its off the gcd if the item was missing, but this just seemed to be more clear and doesnt really hurt anything to have it explicitly marked false.

src/Parser/Mage/Fire/Modules/Features/Abilities.js
Sharrq: its only disabled by default if you dont list a cooldown. If a cooldown is set, then its enabled by default.

That said, ill remove it from the ones where there is no cooldown set.

src/Parser/Mage/Fire/Modules/Features/Abilities.js
Sharrq: how would i do that?? Every other instance of this I have seen has 2 entries. If you point me in the direction of a spec that has what you are referring to then ill change it.

Gebuz: `charges: combatant.hasTalent(SPELLS.FLAME_ON_TALENT.id) || combatant.hasFinger(ITEMS.SOUL_OF_THE_ARCHMAGE.id) ? 3 : 2;`

Sharrq: ahh, didnt think about that. Will do.

Sharrq:
And im pretty sure it was confirmed that traits were dead as of prepatch

image

Sharrq:
made the suggested changes

PR #1790, opened 2018-06-17
[BFA][Restoration Shaman] - EWT pre-combat fix & [Mag'har Orc] - basic support

niseko:
- Added the Mag'har Orc active racial
- Suport for Earthen Wall Totem pre-combat casts & accounting for Mag'har Orcs (if they use their racial)
- Moved talents over to the talents file so I won't have to fix double spell conflicts when the file gets overwritten later on
- Renamed EWT and LST
- Made the Earth Shield math more accurate
- Added a Changelog as I'm not expecting anymore changes


src/Parser/Shaman/Restoration/Modules/Talents/EarthenWallTotem.js
```diff
@@ -12,79 +12,84 @@ import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
...
 
-    if (event.targetID === this.activeEST) {
-      this.healing += (event.amount || 0) + (event.overheal || 0) + (event.absorbed || 0);
+    if (this.prePullCast) {
+      this.potentialHealing += event.maxHitPoints; // this is taking the totems max HP, which is the same result as the players unless Mag'har Orc
+      this.prePullCast = false;
+      this.prePullCastHealth = event.maxHitPoints;
     }
 
+    this.healing += (event.amount || 0) + (event.overheal || 0) + (event.absorbed || 0);
```
Gebuz: Damage events shouldn't have an overheal prop, and I don't think totems can get absorb shields (and if they can I don't think you'd want to count it as the totems healing anyway).

src/Parser/Shaman/Restoration/Modules/Talents/EarthenWallTotem.js
```diff
@@ -12,79 +12,84 @@ import AbilityTracker from 'Parser/Core/Modules/AbilityTracker';
...
 
-    if (!(spellId === SPELLS.EARTHEN_SHIELD_TOTEM_SELF_DAMAGE.id)) {
+    if (spellId !== SPELLS.EARTHEN_SHIELD_TOTEM_SELF_DAMAGE.id) {
       return;
     }
 
-    if (event.targetID === this.activeEST) {
-      this.healing += (event.amount || 0) + (event.overheal || 0) + (event.absorbed || 0);
+    if (this.prePullCast) {
+      this.potentialHealing += event.maxHitPoints; // this is taking the totems max HP, which is the same result as the players unless Mag'har Orc
```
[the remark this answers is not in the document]
niseko: Thought about that, but what if the totem placement is so miserable that there never is an damage event

src/Parser/Shaman/Restoration/Modules/Talents/EarthenWallTotem.js
niseko: yeah, when its a module that someone else wrote I often keep the oddities in case it does something I don't think about. I'll remove it in a minute so be ready for another travis error :P

PR #1791, opened 2018-06-18
[BFA] Remove Concordance & Netherlight Crucible

niseko:
As its going away with the prepatch, also changed Insignia to state that NLC traits are inactive.


MartijnHols:
Did you confirm that the NLC traits are not working in the prepatch?

niseko:
The NLC forge is quite literally not even click-able anymore

PR #1793, opened 2018-06-18
[BFA][Warlocks] Initial changes for BFA

sMteX:
This will be.. not quite big, rather large, but bear with me please.

Things changed:
- Revised talents, spells, abilities
- Removed old modules
- Occasionally repurposed some modules (Soul Harvest merged into The Master Harvester, Backdraft, Grimoire of Service -> Grimoire: Felguard etc.)
- Fixed references in Changelogs and other texts

This is only the first part in these changes, as you can see quite a lot of TODOs:
- fix the Demonology pets somehow, probably needs some dynamic tracking (because we have 2 new talents that have the possibility to summon **random** demons (I've heard stories of Prince Malchezaar popping out of Inner Demons)).. I tried to make it at least not crash on undefined properties etc.
    - this also gives way to fixing Kazzak's Final Curse 
- revisit Destruction Fragment events, if the numbers are still correct or not
- revisit the Affliction and Demonology Checklists, probably remove some rules or reword them to make sense (particularly Affliction and the buffing UA ticks part)
- mana costs (and `maxMana` in Affliction/LowMana) are probably wrong - I've got the numbers directly from PTR, but base mana (and the costs as well) probably increase with level, hence the console errors
- verify Felstorm functionality (when activated by player)
- verify T20 2p resetting Call Dreadstalkers (can't find anyone on PTR who has T20 2p, I only have 1)

Notes:
- I have no clue what happened there with `package_lock.json`, it happened when I updated my branch from `upstream/bfa`. Advice appreciated.
- Abilities and CooldownThroughoutputTracker will probably be easier to review when viewed side-by-side instead of inline, that's just a mess.

Tested on all 3 specs, I hope it doesn't crash anywhere, even if it's incomplete yet.


Gebuz:
Assuming you haven't added any new dependencies you can just discard any changes made to the package_lock.json

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -8,135 +8,119 @@ class Abilities extends CoreAbilities {
...
-        enabled: combatant.hasTalent(SPELLS.PHANTOM_SINGULARITY_TALENT.id),
-        isOnGCD: true,
-        castEfficiency: {
-          suggestion: true,
-        },
-      },
       {
         spell: SPELLS.AGONY,
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        category: Abilities.SPELL_CATEGORIES.DOTS,
```
Gebuz: This category does not exist at the moment: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ecbb94d9e63f7a2263872522addf975417be9243/src/Parser/Core/Modules/Abilities.js#L14
Any reason to make the distinction between rotational and dot? dots seems pretty rotational to me :P

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -8,135 +8,119 @@ class Abilities extends CoreAbilities {
...
       },
-
-      // Cooldowns
       {
-        spell: SPELLS.SOUL_HARVEST_TALENT,
-        category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
-        cooldown: 120,
-        enabled: combatant.hasTalent(SPELLS.SOUL_HARVEST_TALENT.id),
+        spell: SPELLS.PHANTOM_SINGULARITY_TALENT,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
```
Gebuz: If you're supposed to be casting this on cooldown on single target I'd suggest changing it to `ROTATIONAL` (or if you feel it is more appropriate, `COOLDOWN`) instead. If not, I'd suggest disabling castEfficiency suggestion.

This goes for all the other AOE rotational spells aswell.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -8,135 +8,119 @@ class Abilities extends CoreAbilities {
...
-      },
+      // Defensive
       {
         spell: SPELLS.UNENDING_RESOLVE,
         buffSpellId: SPELLS.UNENDING_RESOLVE.id,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
         cooldown: 180,
+        isOnGCD: false,
         castEfficiency: {
           suggestion: false,
```
Gebuz: This should be true if you want it to show up as a suggestion (which I assume is the case since you have importance set below). It's also the case for a lot of the other defensive or utility spells.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -8,135 +8,119 @@ class Abilities extends CoreAbilities {
...
-          SPELLS.GRIMOIRE_IMP,
-          SPELLS.GRIMOIRE_SUCCUBUS,
-          SPELLS.GRIMOIRE_FELHUNTER,
-          SPELLS.GRIMOIRE_VOIDWALKER,
-        ],
+        spell: SPELLS.DARK_SOUL_MISERY_TALENT,
         category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
-        cooldown: 90,
-        enabled: combatant.hasTalent(SPELLS.GRIMOIRE_OF_SERVICE_TALENT.id),
-        isOnGCD: true,
```
Gebuz: isn't Dark Soul on the GCD? I thought it was.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -8,135 +8,119 @@ class Abilities extends CoreAbilities {
...
-        enabled: combatant.hasTalent(SPELLS.BURNING_RUSH_TALENT.id),
-        isOnGCD: true,
-      },
+      // Defensive
       {
         spell: SPELLS.UNENDING_RESOLVE,
         buffSpellId: SPELLS.UNENDING_RESOLVE.id,
         category: Abilities.SPELL_CATEGORIES.DEFENSIVE,
         cooldown: 180,
+        isOnGCD: false,
```
Gebuz: `isOnGCD` defaults to false so you can leave it out if you want.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
```diff
@@ -146,68 +130,75 @@ class Abilities extends CoreAbilities {
...
+          majorIssueEfficiency: 0.01,
         },
       },
       {
-        spell: SPELLS.MORTAL_COIL_TALENT,
+        spell: SPELLS.SOULSTONE,
         category: Abilities.SPELL_CATEGORIES.UTILITY,
-        cooldown: 45,
-        enabled: combatant.hasTalent(SPELLS.MORTAL_COIL_TALENT.id),
+        cooldown: 600,
```
Gebuz: I would recommend simply not setting a cooldown. You can't get the info from other people using resses.

src/Parser/Warlock/Affliction/Modules/WarlockCore/Channeling.js
```diff
@@ -13,15 +13,15 @@ import { formatMilliseconds } from 'common/format';
  */
 class Channeling extends CoreChanneling {
   on_byPlayer_cast(event) {
-    if (event.ability.guid === SPELLS.DRAIN_SOUL.id) {
+    if (event.ability.guid === SPELLS.DRAIN_SOUL_TALENT.id) {
```
Gebuz: Since Drain Life is its own ability now, I'd suggest also adding handling for that (other specs might need that too).

src/Parser/Warlock/Demonology/Modules/SoulShards/SoulShardTracker.js
```diff
@@ -60,22 +48,16 @@ class SoulShardTracker extends Analyzer {
   spent = {
```
Gebuz: If you find the time at some point in the future it might be a good idea to update this module to extend the core `ResourceTracker` module (it would be significantly less work to maintain then since it automatically generates all the spenders and generators without you having to define them).

sMteX: I'll try to look into that (for a future PR to not bloat this one even more), might not be a lot of problem, but I can see some troubles in Destruction, I think that one *still* hasn't got energize events for the Soul Shard Fragments (but I haven't explicitly looked into the logs).

src/Parser/Warlock/Affliction/Modules/WarlockCore/Channeling.js
sMteX: Could probably make it a shared Warlock module, and have Affliction one extend the shared one, adding Drain Soul as well. I'll add it as TODO for the future since I am little confused by the code.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
sMteX: I prefer to see it explicitly (although it would make sense nowadays to have it inverted - isOnGCD to be true by default with the BFA changes 😅 )

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
sMteX: I meant this to be like "you should use your defensives at least once" (to help healers during fights), but I probably messed it up. Same goes for Demonic Circle - you should use it probably if you have it talented, but it's very situational (so not cast on cooldown). I'll turn the suggestion off and perhaps make a module to track that instead.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
sMteX: It depends. On Legion, many of us were running PS even on single target, casting it on cooldown, but it *is* better suited for recurring add spawns, picture something like Portal Keeper. So it isn't generally a cooldown, but I wouldn't necessarily call it inherently rotational. Disabling the suggestion might be a good thing though.

src/Parser/Warlock/Affliction/Modules/Features/Abilities.js
Gebuz: all you need to do is set `suggestion` to true and it will work like you want it to.

src/Parser/Warlock/Affliction/Modules/Features/Checklist.js
```diff
@@ -73,23 +67,15 @@ class Checklist extends CoreChecklist {
...
     new Rule({
       name: <React.Fragment>Buff your <SpellLink id={SPELLS.UNSTABLE_AFFLICTION_CAST.id} /> as much as possible</React.Fragment>,
-      description: <React.Fragment><SpellLink id={SPELLS.UNSTABLE_AFFLICTION_CAST.id} /> is your biggest source of damage and you should try to buff its damage as much as possible with <SpellLink id={SPELLS.REAP_SOULS.id} /> and <SpellLink id={SPELLS.MALEFIC_GRASP_TALENT.id} /> or <SpellLink id={SPELLS.HAUNT_TALENT.id} /> (if talented). <br />
-        If you don't have <SpellLink id={SPELLS.WARLOCK_TORMENTED_SOULS.id} />, it's ok to wait a bit (as long as you're not wasting Soul Shards while waiting) and cast Unstable Affliction when you can buff it with Reap Souls.
+      description: <React.Fragment><SpellLink id={SPELLS.UNSTABLE_AFFLICTION_CAST.id} /> is your biggest source of damage and you should try to buff its damage as much as possible with Reap Souls and Malefic Grasp or <SpellLink id={SPELLS.HAUNT_TALENT.id} /> (if talented). <br />
+        If you don't have Tormented Souls, it's ok to wait a bit (as long as you're not wasting Soul Shards while waiting) and cast Unstable Affliction when you can buff it with Reap Souls.
```
Gebuz: I'm guessing this line doesn't make sense anymore with the artifact ability being removed?

sMteX: Yeah, I mentioned that in the description of this PR, that I need to revisit the texts. For now it was just to fix it from crashing.

PR #1794, opened 2018-06-18
[Blood Death Knight][Protection Warrior][BfA] Dragon Roar & Consumption

joshinat0r:
- adds proper breakdowns for Consumption & Dragon Roar (since tracking only cooldown time is kinda meh, its actually benefitial to miss out on a cast if you can hit multiple targets with it)
- removes consumption as artifact ability
- replaces Battle Cry with Avatar for Angermanagement 
- adjusts DRW from 3min to 2min

image
image
image
Consumption has the same suggestions & checklist items


joshinat0r:
Not ready for a review, just realized that it's not that simple. A player that could cast Consumption 6x in a fight wouldn't receive a suggestion right now if he made a single cast with 6 hits even when he could've made additional casts. 
Would close this one and open a new PR when I get more accurate numbers.

Issue #1747, opened 2018-05-28
Better cast efficiency for spells with cast time and resets


Gebuz:
Currently cast time spells like Healing Rain and Ebonbolt can not reach 100% cast efficiency because they will always have a time off cooldown between the spell coming off cooldown and the cast completing. It can probably be solved by adding a cast time prop to abilities and accounting for that when calculating the cast efficiency. To be completely accurate it would have to account for the haste at the time of the cast.

Spells like Mangle that has a chance to reset on other casts can not reach 100% cast efficiency either because it always needs the full GCD from the cast that procced it to run its course before it can be cast. I'm not sure exactly what the best way to solve this would be, but it would likely include summing up the duration of all the GCDs after a proc and accounting for those same way as the cast times above.

This changes would mean that cast efficiency is no longer equal to 100%-timeOnCooldown, so we need to make sure that we don't display it as "Time on cooldown" when using the efficiency.

Gebuz:
I've been thinking about how this could be done.

Channel time:
 - if a spell does not have multiple charges add averageCastTime to averageCooldown when calculating the cast efficiency here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/eb8fcfd489ac1bd7c0a695c6657cfe07a0c27885/src/Parser/Core/Modules/CastEfficiency.js#L129
 - Make a function to get the average cast time for a given ability. Should be as simple as adding the times between any casts and the previous begincast events and dividing by the total number of begincast events.
 - Consider copying the cast bar down to the cooldown line on the timeline to visually indicate the efficiency.

GCD and resets:
 - if a spell is on the GCD and does not have multiple charges, add AverageTimeOnGCD to averageCooldown when calculating the cast efficiency. (Note that a spell could have both a cast time and resets like Lava Burst)
 - we need: Reset times ('endcooldown' events where start + expectedDuration > end), GCD going at the time (need a buffer here in case the "reset buff" is applied before the cast event).
 - how to handle partial GCDs (off the GCD procs like lava surge from dot ticks). Maybe have a prop for it  in ability.js so we can handle it differently. Check if there was a GCD active at the time of the proc and if there was add only the time from proc to end of GCD. Could also ignore the prop and just use this whenever it is outside the buffer window, but would add +-bufferMS in inaccuracy.
 - consider copying the GCD (or partial GCD) bar down to the cooldown line on the timeline to visually indicate the efficiency.

[...]

PR #1797, opened 2018-06-20
[Core] Improved cast efficiency

Gebuz:
Takes the cast time and the GCDs after a reset proc into account when calculating max casts and time on cooldown/efficiency.

How to name and display it is still up for debate, see #1747

before:
image
image

after:
image
image


src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -81,6 +81,56 @@ class CastEfficiency extends Analyzer {
     };
   }
 
+  /*
+   * Time spent casting the ability
```
MartijnHols: This comment doesn't add any information and can be omitted or changed to explain why or summarize what it does

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -101,16 +151,20 @@ class CastEfficiency extends Analyzer {
     const cooldown = ability.castEfficiency.disabled ? null : ability.cooldown;
     const cooldownMs = !cooldown ? null : cooldown * 1000;
     const cdInfo = this._getCooldownInfo(ability);
+    const timeSpentCasting = cooldown && ability.charges < 2 ? this._getTimeSpentCasting(ability) : 0;
```
MartijnHols: When a ternary has multiple conditions it's often unclear if this results in either:
``` 
(cooldown && ability.charges < 2) ? this._getTimeSpentCasting(ability) : 0
```
or 
```
cooldown && (ability.charges < 2 ? this._getTimeSpentCasting(ability) : 0)
```

It requires advanced language knowledge to figure out the behavior of such a statement (&& taking priority over ternaries, which is different in some other languages). The fix also has little cost (imo improves readability), therefore it's recommended to use parenthesis in ternaries when you're checking more than one thing.

MartijnHols: Why are spells with multiple charges excluded? I reckon if Healing Rain had 2 charges you'd still want to add the casting time when it first goes on cooldown.

src/Parser/Core/Modules/CastEfficiency.js
```diff
@@ -101,16 +151,20 @@ class CastEfficiency extends Analyzer {
...
+    const timeWaitingOnGCD = cooldown && ability.charges < 2 && ability.isOnGCD ? this._getTimeWaitingOnGCD(ability) : 0;
```
MartijnHols: Why are spells with multiple charges excluded? I reckon if Riptide had two charges you'd still want to add the GCD time when it first goes on cooldown.

src/Parser/Core/Modules/SpellHistory.js
```diff
@@ -42,6 +47,22 @@ class SpellHistory extends Analyzer {
...
 
+  _addTimeWaitingOnGCD(event) {
+    if (!this.lastGlobalCooldown) {
+      return event;
+    }
+    const resetBufferMS = RESET_BUFFER_PERCENT * this.lastGlobalCooldown.duration;
+    const earlyByMS = event.start + event.expectedDuration - event.timestamp;
+    // check if the ability was reset early. If the reset was less than a percentage of the GCD don't consider it as a reset.
+    // check if the reset happened as a result of the previous cast
+    if (event.trigger === 'endcooldown' && resetBufferMS < earlyByMS
```
MartijnHols: This if-statement has a bunch of conditionals making it harder to read that could be separated into multiple if-statements with early returns instead.

src/Parser/Core/Modules/SpellHistory.js
```diff
@@ -42,6 +47,22 @@ class SpellHistory extends Analyzer {
     }
   }
 
+  _addTimeWaitingOnGCD(event) {
+    if (!this.lastGlobalCooldown) {
+      return event;
+    }
+    const resetBufferMS = RESET_BUFFER_PERCENT * this.lastGlobalCooldown.duration;
+    const earlyByMS = event.start + event.expectedDuration - event.timestamp;
+    // check if the ability was reset early. If the reset was less than a percentage of the GCD don't consider it as a reset.
```
MartijnHols: How would this handle Divine Purpose?

src/Parser/Core/Modules/SpellHistory.js
```diff
@@ -60,7 +81,10 @@ class SpellHistory extends Analyzer {
   }
   on_toPlayer_updatespellusable(event) {
     const spellId = event.ability.guid;
-    this._append(spellId, event);
+    this._append(spellId, this._addTimeWaitingOnGCD(event));
```
MartijnHols: Nit pick: this suggests `_addTimeWaitingOnGCD` is a pure function while it mutates event. I think it would be cleaner if this were two lines without changing the value of `event`.

src/Parser/Core/Modules/CastEfficiency.js
Gebuz: The reason spells with charges is not included is because you can always have one of them recharging if you play optimally, except for, as you mention, the very first cast. It has a pretty minor impact overall though, and precasting would mess with it, so I'm not sure if it should be included.

src/Parser/Core/Modules/CastEfficiency.js
Gebuz: The GCD time is after a proc resets the cooldown, but because the proc happens on a cast you are always locked out of actually casting the spell again until the next GCD. For instance, take mangle and gore procs (other abilities has an X% chance to reset the CD o mangle):
image
The max possible casts is achieved here (as soon as another ability resets the CD it is cast the next GCD), but the time on cooldown is significantly lower.

If a spell has multiple charges, you could always have one recharging, even if the other one resets.

src/Parser/Core/Modules/SpellHistory.js
Gebuz: It would work the same way as mangle (explained above)
before:
image
after:
image

Sharrq:
out of curiosity (i didnt look at the actual code) when it takes the cast time into account, does that include cancelled casts? Would the time spent casting that cast that was cancelled count towards the efficiency? 

It probably shouldnt if it does.

Gebuz:
Good question, I asked myself that aswell ;) It wont take cancelled casts into account because it when it goes through the events it just saves the most recent `begincast` and use that whenever it hits a `cast`, so any cancelled casts will simply be overwritten by a new `begincast` before hitting a `cast` event. As for cast time spells that are made instant (like pyro & lava burst), they still have a `begincast` event.

src/Parser/Core/Modules/SpellHistory.js
```diff
@@ -81,7 +81,8 @@ class SpellHistory extends Analyzer {
   }
   on_toPlayer_updatespellusable(event) {
     const spellId = event.ability.guid;
-    this._append(spellId, this._addTimeWaitingOnGCD(event));
+    const extendedEvent = this._eventWithTimeWaitingOnGCD(event);
```
MartijnHols: I meant to just have it on a separate line. Returning the event suggests the returned value is different, but it's not since objects are passed by reference and the original object is mutated.

I'll change it quickly to what I meant so this can be merged.

PR #1803, opened 2018-06-21
[BfA][Ret] Updated abilities and talents and removed artifact traits

Juko8:
Also added myself as contributor since i'm planning to do more work with this going forward


Juko8:
I haven't set patchCompatibility to 8.0 since Judgment and Blade of Wrath modules are currently providing some misleading analysis that needs updating.

src/Parser/Paladin/Retribution/Modules/Abilities.js
```diff
@@ -60,21 +60,11 @@ class Abilities extends CoreAbilities {
...
-        castEfficiency: {
-          suggestion: true,
-        },
-      },
       {
         spell: SPELLS.CRUSADER_STRIKE,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
         charges: 2,
-        cooldown: haste => 3.5 / (1 + haste),
+        cooldown: haste => 6 * 0.85 / (1 + haste),
```
Gebuz: Instead of having two entries for Crusader strike depending on talent choice you could instead set the cooldown based on the talent:
`cooldown: haste => (combatant.hasTalent(SPELLS.THE_FIRES_OF_JUSTICE_TALENT.id) ? 5 : 6) / (1 + haste),`
btw 6*0.85 is 5.1 and not 5, idk if that's intentional?

src/Parser/Paladin/Retribution/Modules/Abilities.js
```diff
@@ -131,21 +110,16 @@ class Abilities extends CoreAbilities {
...
         castEfficiency: {
           suggestion: true,
           recommendedEfficiency: 0.75,
         },
       },
       {
-        spell: SPELLS.DIVINE_HAMMER_TALENT,
+        spell: SPELLS.HAMMER_OF_WRATH_TALENT,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: haste => 12 / (1 + haste),
```
Gebuz: Hammer of Wrath has a cooldown right? :D

src/common/SPELLS/PALADIN.js
```diff
@@ -23,6 +23,12 @@ export default {
     icon: 'spell_holy_righteousfury',
     manaCost: 600,
   },
+  JUDGMENT_CAST_ALT_2: {
```
Gebuz: this name is a bit confusing, Judgment cast alternative 2? Please add a note explaining what this spellID is used for.

src/Parser/Paladin/Retribution/Modules/Abilities.js
Juko8: Will change this to one entry. Blizzard changed the talent to reduce the cooldown by 15% rather than 1 second, so 5.1 seconds is intentional.

PR #1804, opened 2018-06-21
[Protection Warrior][Blood Death Knight][BfA] Dragon Roar/Consumption module, IP update

joshinat0r:
**Shared**
- adds module for Dragon Roar & Consumption (checks now for hits done (and doesn't complain when a player delayed a cast for less than the spells cooldown if it hit multiple targets) instead of cast efficiency)

image
image

**Blood Death Knight**
- adds Consumption as talent
- adds Wraith Walk as talent
- updates DRW cooldown (its only a 2min CD in BfA)
- adds CDR from AMB for AMS
- adds Death's Advance

**Protection Warrior**
- updates Ignore Pain (from 20-60 rage cost to static 40)
- removes Battle Cry from Anger Management CDR and adds Avatar instead

(and sorry for having multiple specs in one PR 😅 )


joshinat0r:
Extracted the logic for it in an extra module.
Also updated the logic behind it, checking if the last cast had multiple hits was indeed wrong. It's now using the correct cast.

src/Parser/Core/Modules/Features/AoESpellEfficiency.js
```diff
@@ -0,0 +1,104 @@
...
+    return Math.ceil(calculateMaxCasts(cooldown, this.owner.fightDuration));
+  }
+
+  get possibleHits() {
+    const cooldown = this.abilities.getAbility(this.ability.id).cooldown;
+    let lastCast = this.owner.fight.start_time;
+    let missedCasts = 0;
+
+    this.casts.forEach((element, index) => {
+      if (element.timestamp - lastCast >= cooldown * 1000 && element.hits === 1) {
```
Gebuz: Sorry about the slow response time, been busy this week. I'm going on vacation Saturday so you might need to poke someone else to finish up this PR.

I still don't think this would achieve the result you want. The time between any two casts will always be >= cooldown, so `element.timestamp - lastCast >= cooldown * 1000` will always evaluate to true. As I tried to explain in my last comment (my apologies if it was a poor explanation) I think what you want to look at is the time between the 2+ hit casts and see how many one hit casts you could've fit between them. I imagine something like this:

```
const cooldownMS = this.abilities.getAbility(this.ability.id).cooldown * 1000;
let lastCast = this.owner.fight.start_time;
let missedCasts = 0;
let timeSum = 0;

this.casts.forEach(e => {
  timeSum += e.timestamp - lastCast - cooldownMS;
  lastCast = e.timestamp;
  // Check if the time sum would have allowed to fit any extra casts and reduce it accordingly.
  missedCasts += Math.floor(timeSum / cooldownMS);
  timeSum %= cooldownMS;
  // reset the time sum if a cast hit more than one target (we have to assume this cast was at an optimal time)
  if (e.hits > 1) {
    timeSum = 0;
  }
});

// add any spare time between last cast and end of encounter 
// (could use current time instead of end of encounter so it updates dynamically while loading)
timeSum += this.owner.fight.end_time - lastCast;
missedCasts += Math.floor(timeSum / cooldownMS);
timeSum %= cooldownMS;
```

Other than that everything else looks fine.

joshinat0r: Tested my version initially with two different logs, and they both showed the numbers I expected. So didn't really bother to check the time between 2+ hit casts since the numbers looked right.
Should've tested with more logs to realize that my numbers are off 😅
Your version works just fine after a few minor adjustments.

PR #1808, opened 2018-06-21
Removing obsolete Abilities and Adding new Spells

HawkCorrigan:
Second attempt at deleting everything that does not apply when prepatch/bfa hits.


src/Parser/Shaman/Elemental/Modules/ShamanCore/FlameShock.js
```diff
@@ -76,7 +38,7 @@ class FlameShock extends Analyzer {
         value={`${formatPercentage(this.uptime)} %`}
         label="Uptime"
         tooltip={
```
MartijnHols: When not joining strings you can just provide a regular (double quoted) string instead of a template field.

PR #1814, opened 2018-06-22
Elemental Shaman: Implemented Exposed Elements

HawkCorrigan:


src/Parser/Shaman/Elemental/Constants.js
```diff
@@ -16,4 +16,9 @@ export const ECHOES = {
   TRESHOLD_PERCENTAGE : 0.3,
 };
 
+export const EXPOSED_ELEMENTS  = {
```
Gebuz: If these constants aren't used in other modules you can just define them in the ExposedElements module

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,56 @@
...
+class ExposedElements extends Analyzer {
+  removeDebuffTimestamp = null;
+  damageGained=0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.EXPOSED_ELEMENTS_TALENT.id);
+  }
+
+  on_byPlayer_cast(event){
+    if(this.removeDebuffTimestamp === null)
```
Gebuz: please always use curly brackets to enclose statements, even if it is one liners

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,56 @@
...
+      return;
+
+    if((event.ability.guid === SPELLS.LIGHTNING_BOLT.id) && (enemy.hasBuff(SPELLS.EXPOSED_ELEMENTS_TALENT.id))){
+      this.removeDebuffTimestamp=event.timestamp;
+      this.damageGained+=(1/EXPOSED_ELEMENTS.MULTIPLIER)*event.amount;
```
Gebuz: please use the helper for this:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/c81403a4dba387b3e984642cbf14fda4b9913228/src/Parser/Core/calculateEffectiveDamage.js#L1
(note that it takes increase and not multiplier as a parameter, so a 100% increase would just be `1`)

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,56 @@
...
+
+class ExposedElements extends Analyzer {
+  removeDebuffTimestamp = null;
+  damageGained=0;
+
+  on_initialized() {
+    this.active = this.combatants.selected.hasTalent(SPELLS.EXPOSED_ELEMENTS_TALENT.id);
+  }
+
+  on_byPlayer_cast(event){
```
Gebuz: Cast events does contain any damage information. You'll want to use `on_byPlayer_damage(event)` instead.

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,56 @@
...
+    if(this.removeDebuffTimestamp === null)
+      return;
+
+    if((event.ability.guid === SPELLS.LIGHTNING_BOLT.id) && (enemy.hasBuff(SPELLS.EXPOSED_ELEMENTS_TALENT.id))){
```
Gebuz: Have you checked if the damage events always happens before the debuff is removed? Otherwise it might be a good idea to add a buffer.

HawkCorrigan:
image

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
HawkCorrigan: I checked for the cast events so I will just continue checking the debuff with those.

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
HawkCorrigan: Yup forgot about that

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
HawkCorrigan: Didn't know that one existed. I just changed the const to work like this now.

HawkCorrigan:
The number is still kinda bullshit because the talent is bugged on beta but that would be the damage it is supposed to be doing.

src/common/SPELLS/TALENTS/SHAMAN.js
```diff
@@ -43,6 +43,7 @@ export default {
   HIGH_VOLTAGE_TALENT: { id: 260890, name: "Unlimited Power", icon: "ability_vehicle_electrocharge"},
   MASTER_OF_THE_ELEMENTS_TALENT: { id: 16166, name: "Master of the Elements", icon: "spell_nature_elementalabsorption"},
   EXPOSED_ELEMENTS_TALENT: { id: 260694, name: "Exposed Elements", icon: "spell_nature_callstorm"},
+  EXPOSED_ELEMENTS_DEBUFF: { id: 269808, name: "Exposed Elements Debuff", icon: "spell_nature_callstorm"},
```
Gebuz: Please add the debuff id in `/SPELLS/SHAMAN` instead, otherwise it will be removed once we auto generate the talent files. (The auto generate only generates the ID of the talents, not any related debuffs or buffs).

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,80 @@
...
+import calculateEffectiveDamage from 'Parser/Core/calculateEffectiveDamage';
+
+
+class ExposedElements extends Analyzer {
+  static dependencies = {
+    combatants : Combatants,
+    enemies : Enemies,
+  };
+
+  EXPOSED_ELEMENTS  = {
```
Gebuz: you can define between imports and class definition

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  on_byPlayer_cast(event) {
+
+    if(event.ability.guid !== SPELLS.LIGHTNING_BOLT.id){
+      return;
+    }
+
+    const enemy = this.enemies.getEntity(event);
+    console.log(enemy);
```
Gebuz: Please remove console.log :)

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,80 @@
...
+      return;
+
+    if(event.timestamp>this.removeDebuffTimestamp+this.EXPOSED_ELEMENTS.WINDOW_DURATION){
+      return;
+    }
+
+    if((event.ability.guid !== SPELLS.LIGHTNING_BOLT_OVERLOAD.id) && (event.ability.guid !== SPELLS.LIGHTNING_BOLT.id)){
+      return;
+    }
+    console.log("OI");
```
Gebuz: Please remove console.log :)

src/Parser/Shaman/Elemental/Modules/Talents/ExposedElements.js
```diff
@@ -0,0 +1,80 @@
...
+
+  on_byPlayer_cast(event) {
+
+    if(event.ability.guid !== SPELLS.LIGHTNING_BOLT.id){
+      return;
+    }
+
+    const enemy = this.enemies.getEntity(event);
+    console.log(enemy);
+    if((enemy) && (enemy.hasBuff(SPELLS.EXPOSED_ELEMENTS_DEBUFF.id))){
```
Gebuz: you could remove some of these parentheses:
`if(enemy && enemy.hasBuff(SPELLS.EXPOSED_ELEMENTS_DEBUFF.id)) {`

HawkCorrigan:
should be fine now

PR #1818, opened 2018-06-22
[BFA] Azerite Trait Parsing

emallson:
Two main things:

1. Change `traitsBySpellId` from `SpellID -> Rank` map to `SpellID -> [Rank]` map
2. Workaround for spell ids not being reported yet. Looks up spell IDs based on the simc azerite data in `common/TraitIdMap.js`. This will *not* abruptly break whenever kihra addresses that particular logging quirk, but can be safely removed once that has been done.

No idea how azerite trait scaling works yet as a function of ranks, but ranks do appear to be related to the ilvl of their source item. It looks like the SimC people may have figured that out but I can't follow their code. 

(Also a couple misc SPELLS changes that got wrapped up in here and that I didn't bother reverting)


emallson:
This shouldn't be merged yet, although Travis will probably say its fine. There are still a LOT of lingering uses of `traitsBySpellId` that will cause client-facing crashes.

emallson:
OK, so I figured out azerite trait scaling (based off simc). The formula is at once simple and a huge pain in the ass:

```budget * avg. coeff```

Easy, right? **WRONG**

Depending on the trait, `budget` is either (a) the amount of primary stat on your (azerite) helm or (b) that value multiplied by the armor CR multiplier. The only way to find out is to try both and see which matches.

Then, the `avg. coeff` is an arbitrary real value that can only be found in the DBC. *There can be one for each effect!* These can be looked up in [this file](https://raw.githubusercontent.com/simulationcraft/simc/bfa-dev/engine/dbc/generated/sc_spell_data.inc) (WARNING ~25MB FILE).

As two examples, consider [Elusive Footwork](http://bfa.wowhead.com/spell=278571/elusive-footwork) and [Gemhide](http://bfa.wowhead.com/spell=268596/gemhide).

- Elusive Footwork: uses *primary* scaling and has an avg coefficient of 2.2 for its one effect.
- Gemhide: uses *secondary* (terminology?) scaling and has separate avg coefficients of ~0.42 and ~1.86 for each of its two effects.

Now, for the final wrench in the works: the rank given by WCL doesn't include the +5 ilvls from the final trait in each piece, but those +5 ilvls *affect scaling*. So a 325 piece with the final trait is treated as 330 for calculating scaling. Currently, I'm not sure if there is a way of matching upgrades to items. Will have to investigate further.

*P.S.* this might ought be made a separate issue

PR #1819, opened 2018-06-22
[Bug][Blood Death Knight] fix for .reduce on empty array

joshinat0r:


src/Parser/DeathKnight/Blood/Modules/Talents/FoulBulwark.js
```diff
@@ -27,7 +27,7 @@ class FoulBulwark extends Analyzer {
   get averageFoulBullwark() {
     let avgStacks = 0;
     this.boneShieldTimesByStack.forEach((elem, index) => {
-      avgStacks += elem.reduce((a, b) => a + b) / this.owner.fightDuration * index;
+      avgStacks += elem.length > 0 ? elem.reduce((a, b) => a + b) / this.owner.fightDuration * index : 0;
```
MartijnHols: The error is caused by the reducer not having a default value, I think giving it a default value might lead to simpler code.

joshinat0r: Definitely cleaner, updated it

PR #1824, opened 2018-06-24
[BFA] Replacing this.combatants.selected with this.selectedCombatant and removing unused combatants imports

niseko:
image
Only 696 files in the end :cry: 

I double checked stuff a lot and I'm pretty sure that no imports that are still needed got removed.


niseko:
Not sure why the IronFur test keeps failing, adding the dependency back doesn't fix it unless i revert the whole file, but the file itself is working fine when trying logs.

PR #1831, opened 2018-06-25
[Feral] Predator talent

Anatta336:
"The cooldown on Tiger's Fury resets when a target dies with one of your Bleed effects active, and Tiger's Fury last 4 additional seconds."

Set up SpellUsable for Feral which attempts to track when an enemy last died with a bleed active, and uses that timestamp to reset Tiger's Fury cooldown when it's used early with the Predator talent. Possible deaths are based on bleed debuffs fading early. There can be other causes of DoTs being removed early so it doesn't make any firm claims about how many enemies died with bleeds on them.

Made a Predator analyzer that assesses how many extra Tiger's Fury casts Predator has given the player, providing a statistic and suggestion. It tracks both overall cast count increase and how many casts were earlier than would have been possible without the talent. There are stronger single target talents available as alternatives to Predator so when few Tiger's Fury resets are detected the suggestion will recommend switching talents.

With Predator it's possible to have Tiger's Fury reset extremely often, so that keeping it on cooldown all the time would have little benefit. So when Predator is active the suggestion threshold for cast efficiency on Tiger's Fury is reduced.

predator01
predator02
predator03

The tracking of bleed durations needed some constants that were already being used in various Feral modules, so I moved them and some others out to a shared Constants file.

The talent isn't changed in BfA, and this will continue to work without any changes (although there's a couple of on_initialised that can be turned into constructors, and combatants dependencies will be able to be removed.)

### DIFF
--- a/src/Parser/Druid/Feral/Modules/Features/SpellUsable.js
+++ b/src/Parser/Druid/Feral/Modules/Features/SpellUsable.js
@@ -5,6 +5,11 @@ import { encodeTargetString } from 'Parser/Core/Modules/EnemyInstances';
 import { RAKE_BASE_DURATION, RIP_BASE_DURATION, THRASH_FERAL_BASE_DURATION, JAGGED_WOUNDS_MODIFIER, PANDEMIC_FRACTION } from '../../Constants';
 
 const EARLY_BLEED_EXPIRE_TO_COUNT_AS_DEATH = 500;
+const BLEED_BASE_DURATIONS = {
+  [SPELLS.RAKE.id]: RAKE_BASE_DURATION,
+  [SPELLS.RIP.id]: RIP_BASE_DURATION,
+  [SPELLS.THRASH_FERAL.id]: THRASH_FERAL_BASE_DURATION,
+};
 
 /**
  * Predator:
@@ -22,12 +27,6 @@ class SpellUsable extends CoreSpellUsable {
     ...CoreSpellUsable.dependencies,
     combatants: Combatants,
   };
-
-  static bleedBaseDurations = {
-    [SPELLS.RAKE.id]: RAKE_BASE_DURATION,
-    [SPELLS.RIP.id]: RIP_BASE_DURATION,
-    [SPELLS.THRASH_FERAL.id]: THRASH_FERAL_BASE_DURATION,
-  }
 
   earlyCastsOfTigersFury = 0;
 
@@ -60,7 +59,7 @@ class SpellUsable extends CoreSpellUsable {
     if (!this.activeBleedsExpire[target]) {
       this.activeBleedsExpire[target] = {};
     }
-    const duration = this.constructor.bleedBaseDurations[spellId] * (this.hasJaggedWounds ? JAGGED_WOUNDS_MODIFIER : 1);
+    const duration = BLEED_BASE_DURATIONS[spellId] * (this.hasJaggedWounds ? JAGGED_WOUNDS_MODIFIER : 1);
     this.activeBleedsExpire[target][spellId] = event.timestamp + duration;
   }
 
@@ -78,8 +77,8 @@ class SpellUsable extends CoreSpellUsable {
     }
     // existingExpire may be null if combat log missed the original applydebuff
     const existingExpire = this.activeBleedsExpire[target][spellId];
-    const remainingOnPrevious = Math.max(0, existingExpire ? (event.timestamp - existingExpire) : 0);
-    const durationWithoutPandemic = this.constructor.bleedBaseDurations[spellId] * (this.hasJaggedWounds ? JAGGED_WOUNDS_MODIFIER : 1);
+    const remainingOnPrevious = Math.max(0, existingExpire ? (existingExpire - event.timestamp) : 0);
+    const durationWithoutPandemic = BLEED_BASE_DURATIONS[spellId] * (this.hasJaggedWounds ? JAGGED_WOUNDS_MODIFIER : 1);
     const pandemic = Math.min(durationWithoutPandemic * PANDEMIC_FRACTION, remainingOnPrevious);
     this.activeBleedsExpire[target][spellId] = event.timestamp + durationWithoutPandemic + pandemic;
   }
@@ -104,7 +103,7 @@ class SpellUsable extends CoreSpellUsable {
   }
 
   isBleed(spellId) {
-    return !!Object.keys(this.constructor.bleedBaseDurations).includes(spellId.toString());
+    return !!Object.keys(BLEED_BASE_DURATIONS).includes(spellId.toString());
   }
 
   beginCooldown(spellId, timestamp) {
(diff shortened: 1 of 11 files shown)


src/Parser/Druid/Feral/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,121 @@
...
+    if (!this.hasPredator || !this.isBleed(spellId)) {
+      return;
+    }
+    const target = encodeTargetString(event.targetID, event.targetInstance);
+    if (!this.activeBleedsExpire[target]) {
+      this.activeBleedsExpire[target] = {};
+    }
+    // existingExpire may be null if combat log missed the original applydebuff
+    const existingExpire = this.activeBleedsExpire[target][spellId];
+    const remainingOnPrevious = Math.max(0, existingExpire ? (event.timestamp - existingExpire) : 0);
```
sMteX: I'm a little confused here, shouldn't it be the opposite? My understanding:
- when you cast a bleed on a target, it calculates the "expected" timestamp when it should fall off on its own (line 64)
- but when you refresh it, then the bleed must still be on the target, so `existingExpire` should be larger than `event.timestamp` here, no?

Example:
- casted Rake at timestamp 120,000
- `this.activeBleedsExpire[target][SPELLS.RAKE.id]` should be 120,000 + 15,000 = 135,000 (assuming no Jagged Wounds)
- but when I *refresh* the Rake, it must have happened before timestamp 135,000, otherwise it wouldn't be a refresh but a new application
- so if I refresh Rake at timestamp let's say 130,000, 5 seconds before it should expire, `event.timestamp - existingExpire` = -5,000, so this line should evaluate *always* to 0

Am I missing something or understood it incorrectly?

src/Parser/Druid/Feral/Modules/Features/SpellUsable.js
```diff
@@ -0,0 +1,121 @@
...
+    const durationWithoutPandemic = this.constructor.bleedBaseDurations[spellId] * (this.hasJaggedWounds ? JAGGED_WOUNDS_MODIFIER : 1);
+    const pandemic = Math.min(durationWithoutPandemic * PANDEMIC_FRACTION, remainingOnPrevious);
+    this.activeBleedsExpire[target][spellId] = event.timestamp + durationWithoutPandemic + pandemic;
+  }
+
+  on_byPlayer_removedebuff(event) {
+    if (super.on_byPlayer_removedebuff) {
```
sMteX: Same as before.

src/Parser/Druid/Feral/Modules/Talents/Predator.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  }
+
+  statistic() {
+    // There may be early casts without any extra casts overall
+    const earlyCastsComment = this.earlyCasts > 0 ? `<br/>Thanks to Predator <b>${this.earlyCasts}</b> of your Tiger's Fury casts ${this.earlyCasts !== 1 ? 'were' : 'was'} before when the cooldown would have been ready.` : '';
+    const tooltip = this.extraCasts > 0 ?
+        `Your Predator talent allowed you to use Tiger's Fury at least <b>${this.extraCasts}</b> extra time${this.extraCasts !== 1 ? 's' : ''}. Without it you would have had time for <b>${this.baseCasts}</b> cast${this.baseCasts !== 1 ? 's' : ''} but with it you were able to use Tiger's Fury <b>${this.totalCasts}</b> time${this.totalCasts !== 1 ? 's' : ''}.${earlyCastsComment}` : 
+        `Your Predator talent didn't allow you to cast more Tiger's Fury overall than you would have been able to without it, with the fight lasting long enough for all <b>${this.totalCasts}</b> of your cast${this.totalCasts !== 1 ? 's' : ''}. Either there were no enemies dying with your bleeds on them during this fight or you didn't make use of Tiger's Fury when it came off cooldown.${earlyCastsComment}`;
```
sMteX: Might be a little easier to read to split into 2 variables, let's say `noExtraCasts` and `hadExtraCasts` instead of ternary. In the tooltip you could then do `this.extraCasts > 0 ? hadExtraCasts : noExtraCasts`.

src/Parser/Druid/Feral/Modules/Talents/Predator.js
```diff
@@ -0,0 +1,92 @@
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.PREDATOR_TALENT.id} />}
+        value={`${this.extraCastsPerMinute.toFixed(2)}`}
+        label="Extra Tiger's Fury casts per minute"
+        tooltip={`${tooltip}`}
```
sMteX: Could be simplified to `tooltip={tooltip}` I think. Or `tooltip={this.extraCasts > 0 ? hadExtraCasts : noExtraCasts}`, should you decide to split it as I suggested.

Anatta336: You're right, thanks! Also did the same for the 'value' line just above.

src/Parser/Druid/Feral/Modules/Features/SpellUsable.js
Anatta336: Thank you for that catch! Duration remaining on existing should definitely be `existingExpire - event.timestamp`

Anatta336:
Thank you for taking the time to review. I've made most of the suggested changes.
The one thing I haven't changed yet is the unnecessary super calls of the form:
```
if (super.on_byPlayer_applydebuff) {
  super.on_byPlayer_applydebuff(event);
}
```
My reasoning is that although Parser/Core/Modules/SpellUsable doesn't handle those events at the moment, if in the future it does I wouldn't want my class to stop it receiving them. It's pretty minor and I'm happy to remove them as unnecessary if you'd like.

A note on exactly how Predator resets Tiger's Fury: the cooldown is reset as soon as any enemy dies with a bleed on it. But this analyzer will assign the "blame" for the cooldown reset to the enemy death event closest to when Tiger's Fury is used early. An example timeline:
```
t:00 cast Tiger's Fury (cooldown due to finish at t = 30)
t:10 A appears to die
t:20 B appears to die
t:21 cast Tiger's Fury
```
Because the reset is assumed to happen at `t = 20` the player's cast efficiency will be reported as higher than if the cooldown actually reset at `t= 10` and they just didn't use it until later. I err on the side of being generous to the player because these "appears to die" events are uncertain. It might be that at `t = 10` a Monk removed all the player's debuffs from A by using Paralysis. It would be frustrating for the player to be told they weren't using Tiger's Fury often enough due to our inaccuracy.

Ideally it would detect enemy death with certainty, but as our logs are filtered to only include player-relevant events I don't think that's currently possible.

sMteX:
Fair enough, I mean it doesn't break anything, the conditions will evaluate to `false` and not call anything nonexistent, just seemed redundant to me at this time :) I imagine if someone were to rework it somehow, they would make sure to update any subclasses too anyway. As for the situation you have described, it sure is tricky and I'm not sure if there even *is* a correct way of deciding since as you said, we don't have the full picture so we can't be sure what really happened with the enemy.

PR #1834, opened 2018-06-27
[Core] character-selection with partition-selection

joshinat0r:
Adds a new dropdown for "partition" if the currently selected zone has a `partitions`-prop.

The initial request will still omit the `partition` (latest partition is selected by default in the dropdown), so we'll always receive first the parses of the latest partition.
If the dropdown is changed it'll send the correct `partition`-id.

Also removed the unnecessary brackets from zones to make the file slightly smaller.

image

### DIFF
--- a/src/Interface/Character/Parses.js
+++ b/src/Interface/Character/Parses.js
@@ -26,6 +26,7 @@ const ORDER_BY = {
   DPS: 1,
   PERCENTILE: 2,
 };
+const OMIT_PARTITION = 0;
 const ZONE_DEFAULT_ANTORUS = 17;
 const BOSS_DEFAULT_ALL_BOSSES = 0;
 const TRINKET_SLOTS = [12, 13];
@@ -59,6 +60,7 @@ class Parses extends React.Component {
       activeZoneID: ZONE_DEFAULT_ANTORUS,
       activeEncounter: BOSS_DEFAULT_ALL_BOSSES,
       sortBy: ORDER_BY.DATE,
+      partition: OMIT_PARTITION,
       metric: 'dps',
       image: null,
       parses: [],
@@ -73,7 +75,7 @@ class Parses extends React.Component {
     this.load = this.load.bind(this);
     this.changeParseStructure = this.changeParseStructure.bind(this);
     this.iconPath = this.iconPath.bind(this);
-    this.updateZoneMetricBoss = this.updateZoneMetricBoss.bind(this);
+    this.updateZoneMetricBossPartition = this.updateZoneMetricBossPartition.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


MartijnHols:
As I said on Discord I don't like the idea of having to maintain such a partitions array, and looking at this it already doesn't work for the older raids.

Does the beta API also require a partition filter? We could ask for that specifically since filtering by partition isn't really relevant for us. Is there any reason we would want to show partition info? I can't think of anything. I think it just makes the interface more complex.

If not then I don't think we actually need this at all. If it does then how about previous/next partition buttons or something like that? Anything that doesn't require manual work would be greatly preferred.

joshinat0r:
We don't need the partitions to filter, we need them to show older partitions.
Eg. we just got the new `empowered` partition, if we omit the partition in our requests we'll the the `empowered` parses & rankings only.
Means all people without logs on the new partition wont see any logs, and no one is able to access logs from last week right now.
I'll ask Kihra if we can get partition infos in the /zone-endpoint, but even then, we'd need to export the zones again once a new patch went live.

Simply showing a next/prev button wouldn't be a good solution either, since some raids have 2 partitions, some have 3.
Which means it would default to the current partition (we omit it), then on a "next"-click we'd show partition 1, on the next click partition 2 (if it exists) and no one would know what the current partition is (since partitions count from 1 upwards).

MartijnHols:
We omit the partition by default, so by default it will look broken to users. They have to discover and learn about partitions and what it means/does before they can continue, but by that time some users will have given up already (caching has that same effect but that's another discussion). Just look at WCL's Discord, class Discords and Twitter to see some of the confusion.

Basically the issue I am describing is that partitioning is a shitty system for our users and I think we shouldn't bother them with it as much as possible.

The second issue is having to maintain a partitions list. It seems important to update the partitions list when a partition is added as soon as possible, but having to wait for a developer to do this won't work in the long term as it relies on someone to both become aware of the issue in a timely fashion and know what to do about it, that he has to do something, and actually doing it. Too many awkward steps to make this a reliable process.

One way to make the partitioning easier would be "older/newer patch" buttons or something like that. That translates well into the reality of the partitions where it's based around patches, and seems like something that most of our users will understand without explanation. This indeed isn't ideal in that we don't know how many partitions there are and it may be weird if someone has logs in partition 0 and then 2, or only in partition 2. We could just allow going 10 partitions back and it shouldn't break any time soon.

Or at the bottom of the list of parses we could show a "look for older parses" button that just goes further into the history (like some apps have "show more results" buttons at the bottom of lists).

Another possible solution would be to load the next partition by default if less than for example 10 parses were found. Then users probably wouldn't even notice the existence of partitions.

Best case I think would be an API endpoint that just gives a list of all fights for a character regardless of partition. The WCL beta character parses endpoint might already do that, I haven't investigated it yet.

joshinat0r:
Making an "older"/"newer" button would result in a lot of trial and error.
The thing is, we still don't know how many partitions a given tier has. Most right now have 2 - 3.
So we'd first make a request without a partition, see that a player has none/very few parses. Now we try it with parition 3 (maybe even get the same results back because 3 is the lastest partition), then rinse and repeat until we're at partition 1?

Asking for all reports regardless of partition might sound like a good idea, but I doubt that. Getting all parses from an active raider right now already takes a decent amount of time (sometimes 5+sec), now add the new logs from `empowered` and then probably the bfa-prepatch partition. Not quite sure if Kihra is a fan of that.

beta.wcl does still the same, omit partition and get the lastest one, or provide one to get data from it.

I'll close this one, don't think there's a good solution to it. We either have to update the partitions manually or increase the amounts of requests we do by trial and erroring us to the correct partition.
And it's not like we get a new partition every other week, mostly happens with the release of a new patch (or like now with a major power increase).

PR #1835, opened 2018-06-28
[BfA][Protection Warrior] Rage generation

joshinat0r:
Warriors dont generate rage from damage taken anymore in BfA.
Instead they'll generate 2 rage per melee swing, and 3 rage for every melee hit taken (with an 1sec ICD).

Adjusts the IP cost to the static 40 rage from the old 20-60 range.

image


joshinat0r:
Yep, if a hit was avoided (parry/dodge) you wont generate rage.
Blocking or fully absorbing the hit does generate rage.

Issue #1786, opened 2018-06-16
Allow users to paste WCL character or even battle.net link in the report address bar


MartijnHols:
And send them to the character page

PR #1836, opened 2018-06-28
[Core] Character links in report-selecter

joshinat0r:
Does what #1786 suggests.

image


joshinat0r:
Note quite sure why I had to update both CSS-files for the path, but that was the reason Travis failed.

PR #1839, opened 2018-06-29
[Prepatch] [Frost Mage] Glacial Spike Changes

Sharrq:
The rotation for Glacial Spike was slightly changed.

Now it is a DPS increase (something like 2-3%) to hold Glacial Spike until you have a Brain Freeze Proc to cast with it as opposed to getting as many Glacial Spike's out as possible.

Additionally, because of this, if the player is spec'd into Ebonbolt, they should hold Ebonbolt and purposely not cast it just incase they have Glacial Spike and need a proc in a pinch. Because of this, it is acceptable (if they player has good luck) for them to never cast Ebonbolt in the fight.

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Features/Abilities.js
+++ b/src/Parser/Mage/Frost/Modules/Features/Abilities.js
@@ -20,6 +20,7 @@ class Abilities extends CoreAbilities {
         cooldown: 45,
         enabled: combatant.hasTalent(SPELLS.EBONBOLT_TALENT.id),
         castEfficiency: {
+          //If using Glacial Spike, it is recommended to hold Ebonbolt as an emergency proc if GS is available and you dont have a Brain Freeze Proc. Therefore, with good luck, it is possible to go the entire fight without casting Ebonbolt.
           disabled: combatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id) ? true : false,
           suggestion: true,
           recommendedEfficiency: 0.90,
--- a/src/Parser/Mage/Frost/Modules/Features/Abilities.js
+++ b/src/Parser/Mage/Frost/Modules/Features/Abilities.js
@@ -20,6 +20,7 @@ class Abilities extends CoreAbilities {
         cooldown: 45,
         enabled: combatant.hasTalent(SPELLS.EBONBOLT_TALENT.id),
         castEfficiency: {
+          disabled: combatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id) ? true : false,
           suggestion: true,
           recommendedEfficiency: 0.90,
         },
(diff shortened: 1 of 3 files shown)


src/Parser/Mage/Frost/Modules/Features/GlacialSpike.js
```diff
@@ -6,48 +6,57 @@ import SpellIcon from 'common/SpellIcon';
...
+      return;
     }
-  }
 
-  on_toPlayer_refreshbuff(event) {
-    if (event.ability.guid === SPELLS.GLACIAL_SPIKE_BUFF.id) {
-      this.overcapped += 1;
+    const damageTarget = encodeTargetString(event.targetID, event.targetInstance);
+    const enemy = this.enemies.getEntity(event);
+    if (this.castTarget === damageTarget && !enemy.hasBuff(SPELLS.WINTERS_CHILL.id)) {
```
Gebuz: Doesn't Frost mage check for winter's chill on cast and not on damage?

Sharrq: Winters chill has a very short duration on the target, and isn't applied until the brain freeze flurry hits the target. So when the glacial spike is cast, winters chill won't be on the target cause flurry hasn't been cast yet, but flurry travels faster than glacial spike so it will land before glacial spike and thus gs will land into winters chill (as long as they aren't standing on top of the boss)

src/Parser/Mage/Frost/Modules/Features/Abilities.js
```diff
@@ -20,6 +20,7 @@ class Abilities extends CoreAbilities {
         cooldown: 45,
         enabled: combatant.hasTalent(SPELLS.EBONBOLT_TALENT.id),
         castEfficiency: {
+          disabled: combatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id) ? true : false,
```
kfinch: Leave a comment (like the one in the PR) explaining why cast effic should be disabled in this case

src/Parser/Mage/Frost/Modules/Features/Checklist.js
```diff
@@ -167,7 +167,7 @@ class Checklist extends CoreChecklist {
             when: combatant.hasTalent(SPELLS.THERMAL_VOID_TALENT.id),
           }),
           new Requirement({
-            name: "All Icicles into Glacial Spike",
+            name: "Glacial Spike Utilization",
```
kfinch: Maybe "Glacial Spike into Winter's Chill"? I've worried some about checklist items not being clear enough.

src/Parser/Mage/Frost/Modules/Features/GlacialSpike.js
kfinch: I'm pretty sure it actually does check on damage

Sharrq:
I made the suggested changes

PR #1841, opened 2018-06-29
[BfA][Protection Paladin] Initial update

joshinat0r:
- updates Consecration
adjusts cooldown
adds suggestion for uptime and its CDR
removes the "buff" component for SotR & HotP/LotP
- updates GCD & cooldown
- adds SotR efficiency (by counting physical damage taken during and outside of SotR)
- adds "mitigate checklist"
- adds proper Judgement-ID for prot
- adds Unbreakable Spirit CDR for Ardent Defender

image
image
image


src/Parser/Paladin/Protection/Modules/Features/Consecration.js
```diff
@@ -0,0 +1,49 @@
...
+            .actual(`${formatPercentage(actual)}% Consecration uptime`)
+            .recommended(`>${formatPercentage(recommended)}% is recommended`);
+        });
+  }
+
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.CONSECRATION_CAST.id} />}
+        value={`${formatPercentage(this.uptime)} %`}
```
Gebuz: Why not use the same efficiency as you do for Shield of the Righteous?

joshinat0r: Same efficiency as in "count the hits taken like SotR"?
Thats kinda pointless, Pallys [new Mastery](http://bfa.wowhead.com/spell=76671/mastery-divine-bulwark) reduces all damage taken, not just phyiscal.
And there's no reason not to have 100% uptime with Consecration lasting 12sec on a 4.5sec (-haste) CD

src/Parser/Paladin/Protection/Modules/Features/ShieldOfTheRighteous.js
```diff
@@ -0,0 +1,71 @@
...
+
+  physicalHitsWithShieldOfTheRighteous = 0;
+  physicalDamageWithShieldOfTheRighteous = 0;
+  physicalHitsWithoutShieldOfTheRighteous = 0;
+  physicalDamageWithoutShieldOfTheRighteous = 0;
+
+  on_toPlayer_damage(event) {
+    if (event.ability.type === MAGIC_SCHOOLS.ids.PHYSICAL && this.selectedCombatant.hasBuff(SPELLS.SHIELD_OF_THE_RIGHTEOUS_BUFF.id)) {
+      this.physicalHitsWithShieldOfTheRighteous += 1;
+      this.physicalDamageWithShieldOfTheRighteous += event.amount + (event.absorbed || 0) + (event.overkill || 0);
```
Gebuz: Might make sense to calculate the damage pre SotR mitigation and use that for calculating the efficiency instead (if you take 100 total dmg pre mitigation and mitigate half of it (and let's assume it's 50% DR), I think showing 50/100 = 50% mitigated would be more intuitive than 25mitigated / (50 unmit + 25 mitigated) = 33%. Might not be possible to calculate though because of the armor changes in BFA, I haven't though it through.

src/Parser/Paladin/Protection/Modules/Features/Consecration.js
Gebuz: It was mostly using percent of dmg taken instead of percent uptime. Could have downtime while not actively tanking and running around, but I'm not sure how relevant that is. I agree showing numbers of hits doesn't make sense for conc though.

src/Parser/Paladin/Protection/Modules/Features/Checklist.js
```diff
@@ -50,6 +58,32 @@ class Checklist extends CoreChecklist {
...
+        </React.Fragment>
+      ),
+      requirements: () => {
+        return [
+          new Requirement({
+            name: <React.Fragment><SpellLink id={SPELLS.SHIELD_OF_THE_RIGHTEOUS.id} /> efficiency</React.Fragment>,
+            check: () => this.shieldOfTheRighteous.suggestionThresholds,
+          }),
+          new Requirement({
+            name: <React.Fragment><SpellLink id={SPELLS.CONSECRATION_CAST.id} /> efficiency</React.Fragment>,
```
Gebuz: I'd suggest calling this 'uptime' instead to not confuse it with how SotR is calculated

src/Parser/Paladin/Protection/Modules/Features/ShieldOfTheRighteous.js
joshinat0r: Just checked it, SotRs DR is just added to the armor-field in WCLs API.
Could probably track back the amount of armor provided by SotR (since its 150% of Str), but that doesn't only require an accurate str tracking but also the proper math behind the armor (DR and armor cap and whatever).

Gebuz: Was there also something about harder bosses (as in later raid tiers) requiring more armor to get the same damage reduction in BFA? Or maybe that was just Stagger working like that, I can't remember. Anyway it's probably a thing for the future.

PR #1846, opened 2018-06-30
[Core][Bug] Update to new parses structure

joshinat0r:
With the update from WLCs API we have a new structure of the `character/rankings` endpoint.
It's right now missing some stuff (namely talents & gear), but PRing it for now so we have at least the basic functions working again.

Another issue is right now that heroic (not sure if normal & LFR have the same issue, but I'd guess so) don't return any percentile (rank 0 out of 0). Will talk with Kihra about it.

image

### DIFF
--- a/src/Interface/Character/Parses.js
+++ b/src/Interface/Character/Parses.js
@@ -144,7 +144,6 @@ class Parses extends React.Component {
       const encounter = Object.values(encounters).find(e => e.id === elem.encounter).name;
       const spec = Object.values(SPECS).find(e => e.className === charClass && e.ranking.spec === elem.spec).specName || 0;
 
-      console.info(elem);
       return {
         name: encounter,
         spec: spec,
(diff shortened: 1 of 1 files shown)


src/Interface/Character/Parses.js
```diff
@@ -144,57 +138,24 @@ class Parses extends React.Component {
...
-          });
-      }
-      return null;
+  changeParseStructure(rawParses, charClass) {
+    const parses = rawParses.map(elem => {
+      const encounters = Object.values(ZONES).find(e => e.id === this.state.activeZoneID).encounters;
+      const encounter = Object.values(encounters).find(e => e.id === elem.encounter).name;
+      const spec = Object.values(SPECS).find(e => e.className === charClass && e.ranking.spec === elem.spec).specName || 0;
+
+      console.info(elem);
```
MartijnHols: Left a console command here

joshinat0r: Indeed, just pushed

PR #1850, opened 2018-07-01
[BFA] Add Azerite Trait ranks from +5 ilvl trait

emallson:
Basically what it says on the tin. Instead of listing the number of empowerment traits, we process them during `_parseTraits` and add +5 ranks to the relevant traits (ranks are 1:1 with ilvls).

I also cleared out the artifact field (replaced with an empty list) from the `COMBATANTINFO.json`. This JSON file should probably be replaced with a BFA version.

After this is merged, it should be possible to start working with traits from beta logs. We do still need a bit more info on how some traits work (e.g. procs: does 2x the trait buff the proc or allow two copies of it? i haven't checked yet), but once we know it'll be possible to do stat calculations and damage calculations.


emallson:
rip this PR, somebody just showed me azerite with 4 traits sigh

PR #1855, opened 2018-07-04
Performance increase for player and boss fadein

rp4rk:
Changes the `left` property animation to be a transform, which eliminates the lag!


MartijnHols:
Could you revert the changes to package-lock.json please? It's probably caused by having a different version of npm (npm is shit like that), but afaik the changes could introduce modules with security issues that were fixed `npm audit`.

PR #1859, opened 2018-07-04
[BFA][Core] Track when a regenerating resource caps, with example for Feral

Anatta336:
The combat log (usually) tells us when resources are spent or energized, but doesn't tell us about natural regeneration. This analyzer keeps track of a resource and its regeneration rate to predict how much resource is available in between the combat log events that give us definite numbers, and whether the resource reached its maximum in that time.

This is mostly useful for detecting a spec wasting energy/focus by allowing it to regenerate to cap and sit there. For these specs time spent at capped energy may be a better measure of lost throughput due to inactivity than time spent not casting. It should be particularly useful for Feral druids early in BfA where it they may spend a significant amount of a fight not casting anything while waiting for energy to regenerate.

The analyzer maintains a record of the last known resource value, its rate of regen, and its maximum value. It listens for the events that might effect any of those values and updates them as needed. From that it works out if the resource reached the cap value since the last update. The bulk of work has been finding ways that the combat log reports resources incorrectly and compensating for that.

The example implementation for Feral druids has a good level of accuracy. The `debugActualVsPredicted` function compares actual combat log values to what it predicted based on the previous event, which consistently shows less than a 2 energy error (not bad considering it's based on rounded numbers.) I also ran a few fights with an addon recording timestamps for when I capped on energy, which showed an average error of 33ms between when the analyzer thought the energy capped and when it really did.

Originally I tried to build this into the existing Parser/Core/Modules/ResourceTracker (which has `FIXME` comments about adding natural regeneration tracking) but keeping compatibility with the many ways that has been extended was making it even longer and more convoluted. I think the two analyzers do different enough tasks to comfortably exist separately, even with a little overlap.

My focus was on energy-like resources but it should work for mana too. Just be ready to spend a while chasing down what abilities aren't reporting resource changes correctly.

regencap01
regencap02

An example BfA log:
https://www.warcraftlogs.com/reports/RzWAa74xLMBdJV8K/#fight=5&source=18&type=resources&spell=103
(Characters in 120 beta raids often wear legendaries that would affect resource regeneration but are inactive due to level. You can manually disable them by setting `debugIsPlayerAbove115` in Feral's `EnergyCapTracker`)


TsaboTavok:
Hi, this looks great, looking forward to using it!
I have a couple of suggestions:
1) I think that RegeneratingResourceTracker is a better name, because this does more then just track the cap. I can see myself using this to help track correct pooling.
2) I can see why you don't want to edit the existing Resource Tracker due to compatibility issues, however, have you considered inheriting from it? You would need to call _applyBuilder and extract a similar _applySpender method, and that should make this tracker compatible with anything that was using the old one. This way strange stuff like invisible events and hidden discounts only need to be defined in 1 tracker.
If you want to make it more clear, we can extract a base resource tracker with common functionality like builders and spenders lists and the spendresource event, and then have 2 sub classes of it defining things like current value and event processing.

MartijnHols:
If nobody reviews this PR before then, I'll review it no later than the end of this weekend. It looks pretty complicated that requires my full attention unlike most other PRs, which is why they have continued to be reviewed and merged. Your contribution definitely looks top notch.

Anatta336:
Thanks both for having a look. I'm happy to take it slow and make sure it's as useful as possible.

Something I'd like to improve on is a spec using both this and ResourceTracker needing to define invisible energizes and unaccounted for discounts in two places. I think that's avoidable with some inheritance arrangement as TsaboTavok suggests. Or the corrections could be dealt with by a normalizer for each spec rather than on the analyzers themselves.

The normalizer approach feels slightly better as all modules would benefit from the corrections. The downside is any modules which are already correcting those errors for themselves might "double-correct", so some care would be needed. Also converting every spec's corrections currently on ResourceTracker into normalizers would take some work.

When I have some time I'll try some experiments and see how much it falls apart.

src/Parser/Core/Modules/RegenResourceCapTracker.js
```diff
@@ -0,0 +1,607 @@
...
+    const current = (shouldAccumulate ? this.predictValue(event.timestamp) : eventResource.amount) - drain;
+    if (debug && !shouldAccumulate) {
+      this.debugActualVsPredicted(eventResource.amount, event.timestamp);
+    }
+    const max = this.currentMaxResourceWithFallback(event);
+    this.updateState(current, max);
+  }
+
+  isLastUpdateRecent(timestamp) {
+    return this.regenState.timestamp + this.constructor.cumulativeEventWindow >= timestamp;
```
MartijnHols: Using parens around multiple conditionals is recommended so the code is easier to understand

src/Parser/Core/Modules/RegenResourceCapTracker.js
```diff
@@ -0,0 +1,607 @@
...
+   */
+  combatantHasBuffActive(buffId, timestamp = null) {
+    if (!buffId || isNaN(buffId)) {
+      throw new Error(`combatantHasBuffActive called without required parameter. buffId: ${buffId}`);
+    }
+    if (!timestamp) {
+      timestamp = this.owner.currentTimestamp;
+    }
+    const buffHistory = this.selectedCombatant.getBuffHistory(buffId);
+    return !!buffHistory.find(buff => (buff.start <= timestamp && (!buff.end || buff.end > timestamp)));
```
MartijnHols: Afaik buffs are only removed from the history after the removebuff event, so the default behavior should already match the event order exactly.

src/Parser/Druid/Feral/Modules/Features/EnergyCapTracker.js
```diff
@@ -0,0 +1,138 @@
...
+    // What should be x.5 becomes x in-game.
+    return Math.floor(max);
+  }
+
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<Icon icon="spell_shadow_shadowworddominate" alt="Capped Energy"/>}
+        value={`${formatPercentage(this.cappedProportion)}%`}
+        label="Time with capped energy"
```
MartijnHols: How about “Watsed energy” instead? I think that would be easier to grasp, and match the goal of these analyzers even more.

PR #1861, opened 2018-07-05
[Core] Charparsing for CN

joshinat0r:
Is using the realm list from https://wow.gamepedia.com/Realms_list.
Not quite sure how up to date it is, footer says ` This page was last edited on 26 July 2017, at 01:05`. But thats for now the best list I could find until Kihra adds a realm endpoint.
Haven't tested all of the realm, but it looks like the slugs of each realm match with the name (slugs differentiate from the name only very few times for characters like `'` anyways), we'll have to rely on users reporting them if a slug is wrong.

CN realms have to skip all checks via the bnet API and get a fallback-image for the character-image (the same fallback is now used when a character render failed to load).

Currently hiding the wipefest link for CN characters since they just keep loading (I guess they can't handle the realm list).

Additionally, fixed the error when users paste an unknown language code into the report-selecter. Would previously crash when the language code is missing in the `REGION_CODES.js`-object. Also removes the hash from pasted URLs (eg. pasting a WCL URL from a specific partition will add it as URL-hash).

image


emallson:
This PR has been approved but not merged, is there a particular reason @rp4rk or just haven't done it yet?

PR #1873, opened 2018-07-09
[BfA] Trinket: Lingering Sporepods

Anatta336:
Lingering Sporepods
>Equip: Your attacks and attacks made against you have a chance to trigger spores to grow for 4 sec before bursting. When they burst, they restore [x] health to you and deal [y] damage split among enemies within 8 yds.

A trinket intended for tanks and physical DPS specs that drops from Sporecaller Zancha in The Underrot (dungeon). Nothing fancy just tracks the damage and healing it causes and count how often it procs.

The buff and damage/heal spells are named the way Blizzard names them, which doesn't quite match the trinket's name.

Example BfA log, a M+ run with the trinket worn by the Vengeance Demon Hunter tank:
https://www.warcraftlogs.com/reports/cGp6XRFjZd4zbaWv#fight=last&source=2

lingering-sporepods

### DIFF
--- a/src/Parser/Core/Modules/Items/BFA/Dungeons/LingeringSporepods.js
+++ b/src/Parser/Core/Modules/Items/BFA/Dungeons/LingeringSporepods.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import ItemDamageDone from 'Main/ItemDamageDone';
+import ItemHealingDone from 'Main/ItemHealingDone';
+
+/**
+ * Equip: Your attacks and attacks made against you have a chance to trigger spores to grow for 4 sec before bursting.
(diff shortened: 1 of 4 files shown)


src/Parser/Core/Modules/Items/BFA/Dungeons/LingeringSporepods.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+    this.healing += (event.amount || 0) + (event.absorbed || 0);
+  }
+  
+  item() {
+    return {
+      item: ITEMS.LINGERING_SPOREPODS,
+      result: (
+        <dfn data-tip={`Procced <b>${this.totalProcs}</b> time${this.totalProcs === 1 ? '' : 's'}.`}>
+        <React.Fragment>
```
MartijnHols: React.Fragment is a substitute wrapper for when you don't have any reason to use another element to wrap multiple children, like you're doing here already with dfn. So React.Fragment can be omitted here.

Anatta336: That's good to know, thank you!

Issue #1698, opened 2018-05-15
Trinket: Waycrest Manor: Balefire Branch


MartijnHols:
![image](https://user-images.githubusercontent.com/4565223/40075360-6c8527c8-587c-11e8-8246-00ea362a8852.png)
![image](https://user-images.githubusercontent.com/4565223/40075368-6ea27d8a-587c-11e8-8385-fca2563c9726.png)
![image](https://user-images.githubusercontent.com/4565223/40075373-70fda33e-587c-11e8-9ab3-99bcf735ead4.png)

Anatta336:
M+ log with a MW monk using the trinket:
https://www.warcraftlogs.com/reports/mtnTNQd3jYZcykCH#fight=4&source=38&type=summary
Buff appears to be 268998:
http://bfa.wowhead.com/spell=268998/kindled-soul

MartijnHols:
Do none of the stat scaling methods work? Trinket effects don’t consistently use the same formula for stats, sometimes it’s the primary stat, then secondary stat and other times a trinket specific formula. They should all be implemented.

PR #1876, opened 2018-07-10
[BFA] Trinket Balefire Branch

Anatta336:
Related issue: #1698 
**Balefire Branch** [(wowdb)](https://beta.wowdb.com/items/159630-balefire-branch)
>Use: Kindle your soul, gaining [x] Intellect, which decays over 20 sec or when taking damage. (1 Min, 30 Sec Cooldown)

Fortunately the decaying intellect is represented as a stacking buff that is mostly reported correctly in the combat log (the exception is when it's first applied at 100 stacks, which shows as 1.)

As with other "on use" trinkets an ability is added with a default cast efficiency suggestion. There's no cast event shown in the log when the trinket is used, so an event is fabricated when trinket use is detected through applybuff.

Using the same technique as the Eye of Command trinket from Legion, the analyzer tracks the average number of stacks and so the average intellect buff from the trinket. It also calculates what the average intellect contribution would have been if the player hadn't taken any damage. That should help give an idea of how much effect the "decays when taking damage" part has.

Example logs:
Frost Mage in M+ https://www.warcraftlogs.com/reports/Mz1pajtcfJmgGVhF#fight=1&source=2&type=summary&start=1354306&end=1373420
Balance Druid in M+ https://www.warcraftlogs.com/reports/LWhT1yjB8V32FH4C#fight=last&source=91&type=auras

balefire-branch-01
balefire-branch-02
balefire-branch-03

Also snuck in a small fix to remove a needless `<React.Fragment>` from the Lingering Sporepods trinket.

### DIFF
--- a/src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
+++ b/src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
@@ -12,7 +12,7 @@ const ACTIVATION_COOLDOWN = 90; // seconds
 
 const STACKS_START = 100;
 const DURATION_TO_DECAY = 20; // seconds
-const STACKS_LOST_PER_SEC = 5;
+const STACKS_LOST_PER_SEC = STACKS_START / DURATION_TO_DECAY;
 
 // sum of finite linear sequence {0, 5, 10, ... 100} to find sum stacks if no damage taken
 const EXPECTED_SUM_STACKS_PER_USE = (DURATION_TO_DECAY + 1) * STACKS_START * 0.5;
@@ -94,6 +94,17 @@ class BalefireBranch extends Analyzer {
       },
       type: 'cast',
     }, event);
+
+    /**
+     * This first applybuff event shows just 1 stack, but it's actually STACKS_START stacks.
+     * Fabricate an event to let things like StatTracker know the accurate stack number.
+     * (applybuffstack isn't listened for in this analyzer)
+     */
+    this.owner.fabricateEvent({
+      ...event,
+      stack: STACKS_START,
+      type: 'applybuffstack',
+    });
   }
 
   on_toPlayer_removebuffstack(event) {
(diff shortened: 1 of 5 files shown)


src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
```diff
@@ -0,0 +1,167 @@
...
+
+import Analyzer from 'Parser/Core/Analyzer';
+import Abilities from 'Parser/Core/Modules/Abilities';
+import SpellUsable from 'Parser/Core/Modules/SpellUsable';
+
+const ACTIVATION_COOLDOWN = 90; // seconds
+
+const STACKS_START = 100;
+const DURATION_TO_DECAY = 20; // seconds
+const STACKS_LOST_PER_SEC = 5;
```
MartijnHols: nit: Technically more correct would be `STACKS_START / DURATION_TO_DECAY` since that most likely matches the way it's implemented in-game

src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
```diff
@@ -0,0 +1,167 @@
...
+
+  on_finished() {
+    if (!this.lastChange || !this.currentStack) {
+      return;
+    }
+    const now = this.owner.fight.end_time;
+    this.stackChange(0, now);
+    this.totalUptime += (now - this.lastApply);
+
+    if (this.lastApply + DURATION_TO_DECAY > now) {
```
MartijnHols: nit: Using parenthesis around operators is preferred so it can be interpreted by humans without having to be aware of operator precedence.

src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
```diff
@@ -0,0 +1,167 @@
...
+      // buff still active when fight ends, so revert some of the expectedSumStacks
+      const remainingDuration = (this.lastApply + DURATION_TO_DECAY - now) / 1000;
```
MartijnHols: Even simple math rules as + and - could benefit

src/Parser/Core/Modules/Items/BFA/Dungeons/BalefireBranch.js
```diff
@@ -0,0 +1,167 @@
...
+      const wholeSeconds = Math.floor(remainingDuration);
+
+      // sum of finite linear sequence for the 'whole seconds' part
+      const stackSumFromWholeSeconds = wholeSeconds * STACKS_LOST_PER_SEC * (1 + wholeSeconds) * 0.5;
+      const fractionOfSecond = remainingDuration - wholeSeconds;
+      const stacksDuringFraction = (wholeSeconds + 1) * STACKS_LOST_PER_SEC;
+      this.expectedSumStacks -= (stackSumFromWholeSeconds + fractionOfSecond * stacksDuringFraction);
```
MartijnHols: or *

Anatta336:
Thank you for taking the time to review. I've applied those tidying up changes.

I noticed I'd missed StatTracker while you were reviewing, so snuck in a commit with that a little earlier. 
To make it give the full intellect gain for the initial buff I've fabricated an `applybuffstack` in response to the initial `buffapply` to make it appear as 100 stacks. After that the log gives accurate stack counts and StatTracker applies the correct intellect changes as the stacks decay.

PR #1877, opened 2018-07-10
[BFA][Core] Azerite Rank change

emallson:
WCL now reports ranks as the full ilvl rather than the difference from ilvl 251. Updated the scaling function and tests to match this.


emallson:
@MartijnHols halp

PR #1878, opened 2018-07-12
[BFA] Gemhide Trait

emallson:
screenshot from 2018-07-11 20-58-56
basically self-explanatory

also updates the trait scaling data from the latest simc dbc dump


src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/Gemhide.js
```diff
@@ -0,0 +1,83 @@
...
+  get pctHitsMitigated() {
+    return this._hitsMitigated / this._totalHits;
+  }
+
+  on_toPlayer_damage(event) {
+    if(event.sourceIsFriendly) {
+      return;
+    }
+    this._totalHits += 1;
+    this._hitsMitigated += this.selectedCombatant.hasBuff(SPELLS.GEMHIDE_BUFF.id);
```
MartijnHols: Using js’ automatic type changing is not recommended to be used since it can lead to bugs more easily and is less obvious. Please use a ternary.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/Gemhide.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  statistic() {
+    return (
+      <StatisticBox 
+        icon={<SpellIcon id={SPELLS.GEMHIDE.id} />}
+        value={`${formatNumber(this.avgArmor)} & ${formatNumber(this.avgAvoidance)}`}
```
MartijnHols: Showing the value like this makes it hard to read, how about showing the stat names as well?

emallson:
screenshot from 2018-07-15 21-12-11
new look

PR #1885, opened 2018-07-14
[BFA][Feral] ComboPointTracker fixes

Anatta336:
This fixes a couple of bugs in Feral's `ComboPointTracker` and tweaks how wasted combo points from Primal Fury are displayed.

Feral has 3 AoE abilities which generate combo points if they hit something, but that generation is not shown in the combat log. Previously a `processInvisibleEnergize` was called whenever such an ability is cast. But that would trigger an energize even if the ability hit nothing at all. An example of Thrash being cast and hitting nothing is at 2:01.058 in this log (it's a log from 7.3.5, but the relevant mechanics are unchanged)
https://www.warcraftlogs.com/reports/VhHqJQ3M89GfcbRD/#fight=4&source=10

I solved this with a normalizer that generates energize events backdated to immediately after the cast event, but only for those casts that resulted in a hit. There's also a (rather verbose) test for that normalizer. This feels like a big fix for a small problem but I couldn't find a way to do it with fabricated events during parsing or applying then reverting combo points that wouldn't fail in some situations, mainly related to identifying when a Primal Fury waste is unavoidable.

The second bug was that the player reaching 5 combo points and marking Primal Fury waste as unavoidable was being detected in `on_toPlayer_energize`, which doesn't get called when the energize is applied through `processInvisibleEnergize`. There are many instances of this happening in that same log. Technically this was fixed by the normalizer as well, but I moved the logic over to `_applyBuilder` so it will continue to work if something else needs to use `processInvisibleEnergize` in the future.

I have also changed how unavoidable waste of Primal Fury procs is displayed. The total number of wasted combo points used for the statistic and suggestion is unchanged and still doesn't count unavoidable Primal Fury procs (the screenshot below shows a different value because of the bug fixes). But now the waste is shown in the Combo Point Usage tab. I feel providing the full information is more clear to a user so they can see that Primal Fury is wasting combo points, even if that's not something they should try to avoid.

combo-fix-01
combo-fix-02
Previous results are on the left, new results on the right. Note Thrash generating 1 fewer combo point due to it being used and hitting nothing. The apparent increase in wasted Primal Fury is because the unavoidable waste is now being shown in that panel. The improved detection of unavoidable waste can be seen in the reduced "wasted combo points per minute" statistic.
(The unknown spell is Feral's artifact weapon ability.)


src/Parser/Druid/Feral/Modules/Normalizers/ComboPointsFromAoE.js
```diff
@@ -0,0 +1,147 @@
...
+    let castEventIndex = null;
+    let comboAtCast = null;
+
+    // When adding new events the indexing of original events and fixedEvents will become offset
+    let fixedEventIndex = 0;
+
+    events.forEach(event => {
+      const eventComboResource = this.getResource(event, RESOURCE_TYPES.COMBO_POINTS.id);
+
+      if ((event.type === 'energize') && (event.targetID === this.playerId) &&
```
MartijnHols: These parenthesis can be omitted

PR #1892, opened 2018-07-15
[BFA][Restoration Shaman] Nature's Guardian

niseko:
- Added support for the Nature's Guardian talent by creating a cast event for every heal event. Probably doesn't add much of value but might be interesting for some people.
- Changed Tremor Totem's GCD to be reduced by haste.

https://bfa.wowhead.com/spell=30884/natures-guardian
>When your health is brought below 35%, you instantly heal for 20% of your maximum health.  Cannot occur more than once every 45 sec.

image
image image

### DIFF
--- a/src/Parser/Shaman/Restoration/Modules/Talents/NaturesGuardian.js
+++ b/src/Parser/Shaman/Restoration/Modules/Talents/NaturesGuardian.js
@@ -1,0 +1,61 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+
+import Analyzer from 'Parser/Core/Analyzer';
+
+import CooldownThroughputTracker from '../Features/CooldownThroughputTracker';
+/**
+ * When your health is brought below 35%, you instantly heal for 20% of your maximum health.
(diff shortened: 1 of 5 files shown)


src/Parser/Shaman/Restoration/Modules/Talents/NaturesGuardian.js
```diff
@@ -0,0 +1,61 @@
...
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+
+    if (spellId !== SPELLS.NATURES_GUARDIAN_HEAL.id) {
+      return;
+    }
+
+    const castEvent = {...event};
+    castEvent.ability.guid = SPELLS.NATURES_GUARDIAN_TALENT.id;
```
MartijnHols: This mutates the original event! While you duplicated the event, the ability object will continue to reference to the same object.

Objects like this could be written like this to prevent such issues:

```jsx
this.owner.fabricateEvent(castEvent, {
  ...event,
  type: 'cast',
  ability: {
    ...event.ability,
    guid: SPELLS.NATURES_GUARDIAN_TALENT.id,
  },
});
```

niseko: Oh, guess i misunderstood how {...event} works. Fixed

PR #1894, opened 2018-07-15
[Restoration Shaman] Prepatch

niseko:
- Adjusted mana costs for level 110
- Adjusted Resurgence module for level 110
- Added Earth Elemental
- Hide healing rain module instead of showing NaN when there were no hits
- Tidecallers / ShamanAbilityTracker / ChainHeal are structural changes, they still work the same way
- Adjusted Earth Shield buff as they changed the value once again


src/Parser/Shaman/Restoration/Modules/Spells/HealingRain.js
```diff
@@ -62,6 +62,10 @@ class HealingRain extends Analyzer {
   }
 
   statistic() {
+    if (isNaN(this.averageHitsPerTick)) {
+      return false;
```
MartijnHols: When you don't want to render anything you should return `null` rather than `false` (I think false triggers a React warning).

PR #1917, opened 2018-07-18
Penance Bugfix

rp4rk:
Penance has broken due to a regression in how the game handles Penance cast events. Previously, one was generated for each bolt, now this only seems to happen for heals. This PR will fix that and hopefully future proof it a little.

It's currently impossible to accurately account for Penance in the timeline because of this though :(

---

* Fixed Penance
* Updated the spec compatibility
* Added a changelog message to indicate bfa readiness
* Added a new EventGrouper tool that will automatically group incoming events to a previous stem event, this is based off a provided threshold. (It's also iterable, which is amazing)

### DIFF
--- a/src/Parser/Core/EventGrouper.js
+++ b/src/Parser/Core/EventGrouper.js
@@ -1,0 +1,44 @@
+/**
+ * Groups events based on a given timestamp threshold
+ * e.g.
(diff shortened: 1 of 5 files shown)


rp4rk:
Merging this due to the severity of the bug and lack of review.

PR #1923, opened 2018-07-18
Mitigation check

hpabst:
-Add checking for "soft" mitigation checks to Antorus bosses (soft mitigation checks are hard-hitting tank abilities that differ from "true" mitigation checks) for protection paladins. This includes adding some ability data to Antorus bosses.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4609,13 +4609,11 @@
         },
         "balanced-match": {
           "version": "1.0.0",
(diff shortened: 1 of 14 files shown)


hpabst:
Good suggestion @emallson. I've changed the way I get boss data to reduce the number of imports. As a bonus, it should make it easier to keep using this Analyzer when future raids come out.

PR #1924, opened 2018-07-18
Remove old traits and abilities

yajinni:


yajinni:
Boneshield DR Removal

The talents/traits dont exist anymore.
And bonshields DR behaved very differently now, only applies to physical damage and is based on armor calculations. So removing everything in relation to that till we get a standardized module for these calcs since it affects all tanks.

src/Parser/DeathKnight/Blood/Modules/Features/BoneShield.js
```diff
@@ -91,13 +50,11 @@ class BoneShield extends Analyzer {
...
   statistic() {
 
-    if (this.hasSD) {
-
       return (
         <ExpandableStatisticBox
           icon={<SpellIcon id={SPELLS.BONE_SHIELD.id} />}
           value={`${formatPercentage(this.uptime)} %`}
-          label="Bone Shield Uptime"
+          label="Bone Shield uptime"
```
emallson: i'd probably leave this uppercased (labels should generally be title-cased).

PR #1926, opened 2018-07-18
[BrM] Added support for Light Brewing (Talent)

emallson:
This talent has existed for a while, but was so bad relative to anything else that I didn't bother implementing support. This PR introduces support for it into my brew CDR tracking code and to `Abilities`.

**Old Checklist**
screenshot from 2018-07-18 18-19-48

**New Checklist**
screenshot from 2018-07-18 18-18-06

There are also two secondary things in here that i touched while working on this PR that I didn't feel quite warranted separate PRs:

- I updated the buff values for Strength of the Sea/Sky for Argus. They were still at 2000 per stat per stack, which fucks up cooldown tracking on argus if you use the haste buff.
- I cut out a module that was only intended for use on beta til we worked out the stagger formula. We know it now, and this module doesn't serve a good purpose outside of that. This is a simple deletion


src/Parser/Monk/Brewmaster/Modules/Core/BrewCDR.js
```diff
@@ -71,6 +76,7 @@ class BrewCDR extends Analyzer {
     this._totalHaste += event.oldHaste * (event.timestamp - this._lastHasteChange);
     this._lastHasteChange = event.timestamp;
     this._newHaste = event.newHaste;
+    console.log('haste', this.meanHaste, event.timestamp);
```
Gebuz: remember to remove this console.log. Otherwise everything looks good.

emallson:
@Gebuz fixed

anom0ly:
👍

PR #1928, opened 2018-07-19
[WW] Adding EnergyCapTracker

Juko8:
There didn't seem to be any errors when using discounted crackling jade lighting casts with the legendary chest, so skipping `getReducedCost`

billede


Juko8:
Rushing Jade Winds energy drain is not creating any events, so i'm taking down this PR and making another one once i've gotten a fix for that.

PR #1936, opened 2018-07-20
[Arcane Mage] Mana Chart

Sharrq:
Im not 100% sure if this is accurate or not. I pretty much just took the healing Mana chart and adjusted it to accomodate damage, but that involved copying and editing a lot of stuff into DamageDone and I am not sure if I missed anything.

Im in no rush to get this in, so I would appreciate any feedback on what I might have missed here to make this more accurate.

image

### DIFF
--- a/src/Parser/Mage/Arcane/Modules/ManaUsageGraph.js
+++ b/src/Parser/Mage/Arcane/Modules/ManaUsageGraph.js
@@ -1,0 +1,158 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Line} from 'react-chartjs-2';
+
+import {formatThousands } from 'common/format';
+import {formatDuration} from 'common/format';
+
+import ManaStyles from 'Main/ManaStyles.js';
+
+const baseConfig = {
+  responsive: true,
+  scales: {
+    yAxes: [{
+      gridLines: ManaStyles.gridLines,
+      ticks: {
+        callback: (percentage, index, values) => {
+          return formatThousands(percentage);
+        },
+        fontSize: 14,
+      },
+    }],
+    xAxes: [{
+      gridLines: ManaStyles.gridLines,
+      ticks: {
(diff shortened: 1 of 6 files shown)


src/Parser/Mage/Arcane/Modules/ManaUsageGraph.js
```diff
@@ -0,0 +1,158 @@
...
+          data: Object.keys(damagePerFrame).map(key => damagePerFrame[key] !== null ? damagePerFrame[key] / interval : null),
+        },
+      ],
+    };
+
+    return (
+      <div>
+        <div className="flex">
+          <div className="flex-main">
+            This shows you your mana usage in correlation with your throughput. Big spikes in mana usage without increases in throughput may indicate poor mana usage. The scale for both mana lines is 0-100% where 100% is aligned with the max HPS throughput.
```
Anatta336: The explanatory text could be customised for Arcane mages.

src/Parser/Mage/Arcane/Modules/ManaUsageGraph.js
```diff
@@ -0,0 +1,158 @@
...
+      .map(k => damagePerFrame[k])
+      .forEach((damageDone) => {
+        const current = damageDone.effective;
+        if (current > max) {
+          max = current;
+        }
+      });
+    max /= interval;
+
+    const manaUsagePerFrame = {
```
Anatta336: `manaUsagePerFrame` is calculated but doesn't end up getting used for anything.

Sharrq:
Im starting to think that the DPS Throughput on this isnt really that important or useful, so i think im going to redo this with the Mana Level Graph that shows the boss's health instead of the DPS Throughput ... I think that will be more beneficial seeing as with arcane you want to start burning everything to the ground regardless of cooldowns at the end of the fight (so you should end the fight with low mana).

So im going to close this and will submit a new PR once I get that going.

PR #1939, opened 2018-07-20
[Tanks] Extend Existing Mitigation Check

hpabst:
- Extends the existing analyzer for mitigation checks to support all tank specs rather than just protection paladin.

- Extends existing mitigation check statistic box to include information on pass/failure for each checked
ability.

image

- Adds encounter spell and damage data for Antorus to facilitate the new statistic box.


src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -0,0 +1,150 @@
...
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber, formatPercentage } from 'common/format';
+
+import SpellLink from 'common/SpellLink';
+import Enemies from 'Parser/Core/Modules/Enemies';
+
+import {findByBossId} from 'Raids/index';
+import ExpandableStatisticBox from 'Main/ExpandableStatisticBox';
+
+const mitigationBuffs = {'Paladin' : [SPELLS.SHIELD_OF_THE_RIGHTEOUS_BUFF.id,
```
MartijnHols: Apostrophes can be omitted in object keys

MartijnHols: We generally don't want to require spec maintainers to have to configure things in the core modules, could you make these constants static properties of the class instead please? Then each spec can make its own version of this module by extending it, and by overriding the prop they can configure the correct spells.

Then you also wouldn't need the class names.

src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -0,0 +1,150 @@
...
+const mitigationDebuffs = {'Paladin' : [],
+                            'Monk' : [SPELLS.BREATH_OF_FIRE_DEBUFF.id],
+                            'Warrior': [SPELLS.DEMORALIZING_SHOUT.id],
+                            'DemonHunter': [SPELLS.FIERY_BRAND_DEBUFF.id],
+                            'Druid': [],
+                            'DeathKnight': []};
+
+
+class MitigationCheck extends Analyzer{
+  static checks;
```
MartijnHols: These properties should not be marked as static, they're being used as regular class properties.

A static property is for something that couldn't different between instances (e.g. is the same over multiple fights). If something can be different per fight (or instance of the module), it should be a regular class property, which is just `checks;` without `static`.

MartijnHols: Giving properties a default value of at least `null` would be preferred. Without doing this the value of vars is `undefined`, which is generally a value that should be left for things that are truly undefined. A class property is essentially a definition, and if something doesn't have a default value `null` is a great indicator of that.

src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -0,0 +1,150 @@
...
+  static buffCheck;
+  static debuffCheck;
+  static checksPassedMap;
+  static checksFailedMap;
+
+  static dependencies = {
```
MartijnHols: When making the other properties non-static, please consider moving the dependency definition to the top of the class.

src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -0,0 +1,150 @@
...
+    enemies: Enemies,
+  };
+
+
+  constructor(...args) {
+    super(...args);
+    this.checksPassedMap = new Map();
+    this.checksFailedMap = new Map();
+    const boss = findByBossId(this.owner.boss.id);
+    if(boss !== null && boss.fight.hasOwnProperty('softMitigationChecks')){
```
MartijnHols: Rather than `hasOwnProperty` we prefer checking for property existence with either `boss.fight.softMitigationChecks !== undefined` or just `boss.fights.softMitigationChecks` if `null` should also be considered falsey.

src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -0,0 +1,150 @@
...
+      this.checks = Object.values(boss.fight.softMitigationChecks);
+      if(typeof this.checks === "undefined"){
```
MartijnHols: Could be simplified to `this.checks === undefined`. Alternatively just `if (this.checks) {` would be sufficient.

PR #1942, opened 2018-07-21
Furious Slash Modules

jtufts:
I created two new Modules for the Fury Warrior Talent, Furious Slash, under Parser\Warrior\Fury\Modules\Talents. The code works similar to the code currently used to keep track of Bone Shield stacks for Blood DK's. The FuriousSlashTimesByStacks module returns an array of arrays; the sum of each smaller array gives the total time in ms that the Furious Slash buff had that number of stacks throughout the fight. The FuriousSlashUptime Module takes this data, and presents the results as a table, with the percentage of time at maximum stacks shown at the top, and both the durations and percentages for all stacks below that. This table is necessary because Furious Slash is a maintenance buff, and should be tracked as such.
furiousslashstacktracker
furiousslashsuggestion


jtufts:
Closed due to bugs in the code found upon further testing.

PR #1947, opened 2018-07-21
[Fury] Furious Slash Talent Modules

jtufts:
I have created two new modules that work together to keep track of the maintenance buff Fury Warrior talent, Furious Slash. The module works similar to the Bone Shield modules for the Blood DK's, but I had to use on_byPlayer_damage with restrictions instead of on_byPlayer_applybuff due to it not being able to keep track of when buffs are refreshed. Keeping track of when buffs are refreshed is necessary for another suggestion module I am planning to implement that keeps track of how much time is clipped off of Furious Slash at maximum stacks throughout the fight. I attempted to use refreshbuff, but that did not work either, forcing me to resort to this solution. 

The FuriousSlashTimesByStacks module calculates how long each single Furious Slash buff lasted, and how many stacks that buff had before either being dropped, refreshed, or increasing in stacks. The module then outputs the results as an array of smaller arrays, with each larger array corresponding to a number of stacks and each smaller array keeping track of the buff uptimes. The FuriousSlashUptime module then takes this data, and inputs it into a table, showing both the duration and uptime of each stack count. These modules are necessary because currently, we have no suggestions on this talent, and it is used often enough and is important enough to the rotation to be worth our users' attention.
furiousslashstacktracker
furiousslashsuggestion


src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
```diff
@@ -0,0 +1,102 @@
...
+      event.stack = 0;
+    }
+	/*
+    if (event.type === 'applybuff') {
+      event.stack = 1;
+    }
+	*/
+	
+	if (event.type === 'damage') {
+		if(!this.lastFuriousSlashStack)
```
MartijnHols: I'm having a hard time following what you're doing here. Should this trigger on every single damage event that is dodged?

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
```diff
@@ -0,0 +1,102 @@
...
+		{
+			event.stack = 1;
```
MartijnHols: Please try to avoid mutating the event as much as possible. Especially props such as `stack` are not meant to be messed with as much as possible since that could break other modules.

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+	
+    if (stack) {
+      event.stack = stack;
+    }
+
+    this.furiousSlashStacks[this.lastFuriousSlashStack].push(event.timestamp - this.lastFuriousSlashUpdate);
+    this.lastFuriousSlashUpdate = event.timestamp;
+    this.lastFuriousSlashStack = event.stack;
```
MartijnHols: Does this only occur once per fight? If I am following the code correctly this looks like it only works once.

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashUptime.js
```diff
@@ -0,0 +1,91 @@
...
+  suggestions(when){
+	  const{
+		  isLessThan:{
+			  minor,
+			  average,
+			  major,
+		  },
+	  }=this.uptimeSuggestionThresholds;
+	  
+	  if(this.selectedCombatant.hasTalent(SPELLS.FURIOUS_SLASH_TALENT.id))
```
MartijnHols: Is this module useful when the talent is not selected? If not, the entire module should be deactivated if the talent wasn't selected, you can do this by setting the `active` prop in the constructor.

src/Parser/Warrior/Fury/CombatLogParser.js
```diff
@@ -35,6 +37,8 @@ class CombatLogParser extends CoreCombatLogParser {
     rampageFrothingBerserker: RampageFrothingBerserker,
     rampageCancelled: RampageCancelled,
     angerManagement: AngerManagement,
+	furiousSlashTimesByStacks: FuriousSlashTimesByStacks,
```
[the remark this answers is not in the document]
jtufts: I believe I have fixed this in a new commit. Would you please review again and see if the indentation meets your standards?

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
jtufts: The code in question should trigger if and only if:
a. The spellId matches the id of the Furious Slash Talent
b. The attack is NOT dodged
c. The attack is NOT parried

Pewtro:
Hey - it seems you've run into a travis error, once you have it fixed up I'll be happy to review the code again.

jtufts:
I have fixed the Travis error, and my code is ready to be reviewed again

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashUptime.js
Pewtro: This bit here still needs fixing. 
Seeing as you're already setting the active prop, you can simply remove the  if(this.selectedCombatant.hasTalent(SPELLS.FURIOUS_SLASH_TALENT.id)) { }

Pewtro: Also the suggestions seems to follow our old way of doing them. 

You can simply do 
```suggestions(when) {
when(this.uptimeSuggestionsThresholds).addSuggestion(xxxx)
``` 

No need for the redundant adding of 
```when(this.maxStackUptime / this.owner.fightDuration).isLessThan(minor)```
or 
```
const{
	  isLessThan:{
    minor,
	  average,
	  major,
 },
 }=this.uptimeSuggestionThresholds;```

PR #1949, opened 2018-07-22
[Ret] Quick fix for Divine Purpose

Juko8:
The statistic wasn't counting back to back procs with `getBuffTriggerCount` since that only tracks when the buff is applied and not when it's refreshed.


MartijnHols:
When you fix bugs that for things that are used by more specs, please consider making sure the other specs work properly.

Juko8:
I'm not sure what you mean. This module is to my knowledge not used outside Retribution, since Holy has its own DP implementation.

PR #1951, opened 2018-07-22
[Ret] Fixed Retribution and Inquisition damage calculations

Juko8:
Created some constants to keep track of which Retribution spells are affected by damage increases, one of which won't be used until later.


src/Parser/Paladin/Retribution/Constants.js
```diff
@@ -0,0 +1,29 @@
+import SPELLS from 'common/SPELLS';
+
+// Based on spelldata for Avenging Wrath, Retribution(buff) and Inquisition
+export const ABILITIES_AFFECTED_BY_DAMAGE_INCREASES = [
```
Anatta336: Should SPELLS.MELEE.id also be increased? It wouldn't be listed as an affected spellID by the spelldata dump, but as a separate effect on auto-attack damage.

Juko8: That's right, i'll include melees when those buffs are looked at. Inquisition actually reduces attack speed by 7%, so there's no net increase there.

Anatta336:
👍

PR #1958, opened 2018-07-22
VDH Reworked Demon Spikes suggestion

yajinni:
Previously the spell asked you to mindlessly cast it frequently. Regardless if you were taking or mitigating hits with it. So no point in casting it if the other tank is tanking and your waiting to swap. Now it sees if you took physical hits and had it up.

image

versus now:
image


yajinni:
Part of this pull: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/1959
Closing this.

PR #1959, opened 2018-07-23
VDH Spell Updates

yajinni:
Updated Fracture and Demon Spikes.

Fracture did not have a suggestion, also updated CD and charges numbers.

Previously Demon Spikes asked you to mindlessly cast it frequently. Regardless if you were taking or mitigating hits with it. So no point in casting it if the other tank is tanking and your waiting to swap. Now it sees if you took physical hits and had it up.

image

versus now:
image


src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
```diff
@@ -40,6 +41,18 @@ class DemonSpikes extends Analyzer {
     return formatPercentage(this.hitsWithDS / (this.hitsWithDS + this.hitsWithoutDS));
   }
 
+  suggestions(when) {
+    const hitsWithDSOffCDPercent = this.hitsWithDSOffCD / (this.hitsWithDS+ this.hitsWithoutDS);
+    when(hitsWithDSOffCDPercent).isGreaterThan(0.1)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span> Cast <SpellLink id={SPELLS.DEMON_SPIKES.id} /> more regularly while actively tanking the boss or when they use a big phsyical attack. You missed having it up for ${formatPercentage(hitsWithDSOffCDPercent)}% of physical hits.</span>)
```
emallson: extra $ here, looks like a holdover from template strings. after this it looks good

MartijnHols: Also that should technically be `React.Fragment` instead of `span`.

PR #1962, opened 2018-07-23
[Survival] Baseline BFA Survival

Pewtro:
Relies on #1960 because I checked out from there without thinking. 

Fixes BM Bloodlust spellID. 
Adds Celerity of the windrunners buff to haste module
Fixes issues around the survival module in general to ensure it works correctly and loads.


src/Parser/Hunter/Survival/Modules/Abilities.js
```diff
@@ -1,29 +1,49 @@
...
+      },
+      {
+        spell: SPELLS.RAPTOR_STRIKE,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        gcd: {
+          base: 1500,
+        },
+      },
+      {
+        spell: SPELLS.RAPTOR_STRIKE_2,
```
Gebuz: Why is there 2 Raptor Strikes? Probably a good idea to add a note explaining it to make future maintenance easier.

src/Parser/Hunter/Survival/Modules/Abilities.js
```diff
@@ -216,6 +232,30 @@ class Abilities extends CoreAbilities {
...
+          base: 1500,
+        },
+      },
+
+      /**
+       * Following are implemented to avoid annoying console messages.
+       */
+
+      {
+        spell: SPELLS.DEFILED_AUGMENT_RUNE,
```
Gebuz: this is already in https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/CASTS_THAT_ARENT_CASTS.js, does that not prevent the console messages? If not, we should probably make CASTS_THAT_ARENT_CASTS be ignored by the module that is printing the warnings. That way we would also avoid everybody having to make a normalizer just to avoid the console spam. @MartijnHols

src/Parser/Hunter/Survival/Modules/Abilities.js
```diff
@@ -216,6 +232,30 @@ class Abilities extends CoreAbilities {
...
+        category: Abilities.SPELL_CATEGORIES.HIDDEN,
+      },
+      {
+        spell: SPELLS.POTION_OF_PROLONGED_POWER,
```
Gebuz: I plan to make a module for pots so you can leave these out if you want (or you can leave them in until I have added the pots globally, up to you).

src/Parser/Hunter/Survival/Modules/Spells/ButcheryCarve.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.BUTCHERY_TALENT.id && spellId !== SPELLS.CARVE.id) {
+      return;
+    }
+    this.casts++;
+    if (event.timestamp > this.timeOfLastCast + MS_BUFFER) {
```
Gebuz: Is there multiple cast events from a single cast? Otherwise this if statement is redudant.

src/Parser/Hunter/Survival/Modules/Spells/ButcheryCarve.js
```diff
@@ -0,0 +1,82 @@
...
+      return;
+    }
+
+    if (this.spellUsable.cooldownRemaining(SPELLS.WILDFIRE_BOMB.id) > COOLDOWN_REDUCTION_MS) {
+      this.effectiveWFBCDR += this.spellUsable.reduceCooldown(SPELLS.WILDFIRE_BOMB.id, COOLDOWN_REDUCTION_MS);
+      return;
+    }
+
+    if (this.hasGT) {
+      const newChargeCDR = this.abilities.getExpectedCooldownDuration(SPELLS.WILDFIRE_BOMB.id) - this.spellUsable.cooldownRemaining(SPELLS.WILDFIRE_BOMB.id);
```
Gebuz: `spellUsable.reduceCooldown` automatically reduces the next charge if there is leftover cooldown reduction, so no need to do it here! You can leave out l65 to l75

src/Parser/Hunter/Survival/Modules/Spells/KillCommand.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  on_byPlayer_applybuff(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.FLANKERS_ADVANTAGE.id) {
+      return;
+    }
+    if (this.spellUsable.isOnCooldown(SPELLS.KILL_COMMAND_SV.id)) {
+      this.resets++;
+      if (this.hasAlphaPredator) {
```
Gebuz: same as above, no need for special handling for cooldown reduction overlapping multiple charges. You can just do `this.spellUsable.reduceCooldown(this.abilities.getExpectedCooldownDuration(SPELLS.KILL_COMMAND_SV.id));`

src/Parser/Hunter/Survival/Modules/Abilities.js
Pewtro: Nope, they're still in the console if I remove them from abilities.
image

src/Parser/Hunter/Survival/Modules/Spells/ButcheryCarve.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+    if (this.reductionAtCurrentCast === MAX_TARGETS_HIT) {
+      return;
+    }
+    this.reductionAtCurrentCast++;
+    if (!this.spellUsable.isOnCooldown(SPELLS.WILDFIRE_BOMB.id)) {
+      this.wastedWFBReductionMs += COOLDOWN_REDUCTION_MS;
+      return;
+    }
+    if (this.spellUsable.cooldownRemaining(SPELLS.WILDFIRE_BOMB.id) > COOLDOWN_REDUCTION_MS) {
```
Gebuz: you could remove this if statement aswell if you want (l56-l59) the handling for partial CDR below will also handle fully effective CDR correctly (effectiveReductionMs will be equal to COOLDOWN_REDUCTION_MS)

src/Parser/Hunter/Survival/Modules/Spells/KillCommand.js
```diff
@@ -0,0 +1,70 @@
...
+    if (spellId !== SPELLS.FLANKERS_ADVANTAGE.id) {
+      return;
+    }
+    if (!this.spellUsable.isOnCooldown(SPELLS.KILL_COMMAND_SV.id)) {
+      this.resetWhileNotOnCD++;
+      return;
+    }
+    this.resets++;
+    if (this.hasAlphaPredator) {
+      this.spellUsable.endCooldown(SPELLS.KILL_COMMAND_SV.id, this.abilities.getExpectedCooldownDuration(SPELLS.KILL_COMMAND_SV.id));
```
Gebuz: this needs to be `.reduceCooldown` instead of `.endCooldown` (reduce the cooldown by the cooldown duration is effectively refreshing a charge). You could also drop the check for alphaPredator if you want and use it for both cases.

PR #1967, opened 2018-07-23
[Bug] Fix crash when attempting to view report for a boss that cannot be found.

hpabst:
MitigationCheck currently causes a crash if the user attempts to analyze a boss that cannot be found. This can occur whenever a user attempts to analyze a non-Antorus boss.

### DIFF
--- a/src/Parser/Core/Modules/MitigationCheck.js
+++ b/src/Parser/Core/Modules/MitigationCheck.js
@@ -15,7 +15,7 @@ class MitigationCheck extends Analyzer{
   static dependencies = {
     enemies: Enemies,
   };
-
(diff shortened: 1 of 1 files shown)


src/Parser/Core/Modules/MitigationCheck.js
```diff
@@ -25,6 +25,10 @@ class MitigationCheck extends Analyzer{
 
   constructor(...args) {
     super(...args);
+    if(this.owner.boss == null){
+      this.disabled = true;
```
MartijnHols: You can disable a module by setting `this.active` to `false.`

PR #1981, opened 2018-07-24
[Havoc/BM] Fix crashes

Pewtro:
I'm not sure what's happening with the localization namechange, I've never touched it and I'm not allowed to rename any of it. (and if I remove that file, it also removes the lowercase file which I can't even find so...) If it needs to get removed, I'll try and purge everything.


Pewtro:
I'll fix localization after #1983 in this once it goes through (or gets declined)

PR #1983, opened 2018-07-24
Abilities update

Pewtro:
This fixes the issue raised by @Gebuz in #1962 .

It checks whether certain casts are in CASTS_THAT_ARENT_CASTS before throwing the "this spell isn't in spellbook" warning into console.


src/Localization/.gitignore
```diff
@@ -0,0 +1,3 @@
+# https://lingui.github.io/js-lingui/guides/excluding-build-files
```
MartijnHols: Hmmm

MartijnHols: O joy the good old capitalization issue. 

If you removed the dile try to revert it to the parent (original) version in a separate commit. That usually works.

Pewtro: We did it 😄

PR #1986, opened 2018-07-24
Add Potions to abilities and split Healthstones/Health pots

### LINKED ISSUE #1978: Add Potions to abilities and split Healthstones/Health pots
Should be pretty simple since the logic is already implemented for Healthstone/Health pot.
Health pots also no longer share cooldown with Heathstones, so they need their own aswell. Basically there needs to be 3 different abilities added:
 - Healthstone
 - Health pots
 - Output pots (DPS/Mana/mitigation)

Gebuz:
resolves #1978 
Since combat potion suggestion is already handled by another module it doesn't not have a suggestion. The Heathstone/health potion suggestion is the same as before (but split into one for each).

image

One concern I have is that the combat and healing pot will always just show the first one in the array, which might be confusing if you used another potion. Would it be better to just name them "Combat potion" and "Healing potion" and remove the tooltips?


src/Parser/Core/Modules/Items/CombatPotion.js
```diff
@@ -0,0 +1,27 @@
...
+  SPELLS.COASTAL_MANA_POTION,
+];
+const RECOMMENDED_EFFICIENCY = 0;
+
+/**
+ * Tracks combat potion cooldown (DPS, HPS, mana, mitigation).
+ */
+
+class Healthstone extends Potion {
+
```
Pewtro: Should be called ``CombatPotion``

src/Parser/Core/Modules/Items/CombatPotion.js
```diff
@@ -0,0 +1,27 @@
...
+  constructor(...args) {
+    super(...args, SPELL, RECOMMENDED_EFFICIENCY);
+  }
+}
+
+export default Healthstone;
```
Pewtro: Same as above

src/Parser/Core/Modules/Items/HealthPotion.js
```diff
@@ -0,0 +1,31 @@
...
+];
+const RECOMMENDED_EFFICIENCY = 0.6;
+
+/**
+ * Tracks health potion cooldown.
+ */
+
+class Healthstone extends Potion {
+
+  constructor(...args) {
```
Pewtro: Should be ``HealthPotion``

src/Parser/Core/Modules/Items/HealthPotion.js
```diff
@@ -0,0 +1,31 @@
...
+  suggestions(when) {
+		when(this.suggestionThresholds)
+			.addSuggestion((suggest, actual, recommended) => {
+				return suggest(<React.Fragment>You used an <SpellLink id={SPELL[0].id} /> {this.potionCasts > 1 || this.potionCasts === 0 ? this.potionCasts + ' times' : this.potionCasts + ' time'} but could have used it {this.maxCasts > 1 ? this.maxCasts + ' times' : this.maxCasts + ' time'}. If you are low on health, make sure you use your health potion and defensive abilities to stay alive and to help the healers. </React.Fragment>)
+					.icon(SPELL[0].icon);
+      });
+  }
+}
+
+export default Healthstone;
```
Pewtro: Same as above

src/Parser/Core/Modules/Items/Potion.js
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+    this.spellUsable.reduceCooldown(this.spellId, cooldownRemaining - COOLDOWN_MS);
+    this.increaseMaxCasts(event);
+  }
+
+  increaseMaxCasts(event) {
+    // If the death starts the cooldown and there is less than 60 seconds remaining of the encounter another cast was possible.
+    const nextAvailablePotionCast = event.timestamp + COOLDOWN_MS;
+    if (nextAvailablePotionCast < this.owner.fight.end_time) {
```
Pewtro: should be `` and there is more than 60 seconds`` right?

Gebuz:
Yeah that seems like a reasonable solution. Likely pretty rare that a player has multiple uses, let alone uses of different pots.

Gebuz:
Turns out there's a problem with that, can't change the entry in the spellbook after initialization (which is probably for the best). I'm not sure if there is a clean way to get around that.

Pewtro:
That's rather unfortunate. 
It would be more misleading for us to show Prolonged Power if they're using Deadly Grace than for it to just say 'Combat Potion'. So I guess the idea of calling them 'Combat Potion' and 'Healing Potion' is the second best option then.

Gebuz:
I'll take a look at renaming it tonight

Gebuz:
Wont have time tonight, probably wont get around to have a look at it until sunday or monday :(

Gebuz:
I was not able to find an elegant solution to the naming issue. I considered saving the first cast for every ability, but that would also cause other spells like New Moon and Ironskin Brew to change, and that might cause some confusion for users. I think it should be fine to merge for now, it will always show Prolonged power as the combat potion, but since all specs can use that potion I think it's fine for now.

PR #1988, opened 2018-07-25
Vdh soul fragments rework

yajinni:
Renamed and reworked a couple of the files related to soul fragments to make them more clear.
Then added a soul fragment consumption module.

image


src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
```diff
@@ -0,0 +1,98 @@
...
+      this.totalSoulsConsumed += 1;
+    }
+  }
+
+
+  statistic() {
+    this.soulsConsumedBySpell[1000000].souls = this.soulFragmentsTracker.soulsWasted;
+    this.soulsConsumedBySpell[1000000].name = 'Overcap';
+    this.soulsConsumedBySpell[1000001].souls = this.soulFragmentsTracker.soulsGenerated - this.soulFragmentsTracker.currentSouls - this.soulFragmentsTracker.soulsWasted - this.totalSoulsConsumed;
+    this.soulsConsumedBySpell[1000001].name = 'By Touch';
```
kfinch: Looks like you're trying to jam 'Overcap' and 'By Touch' into a map with spells, which is difficult to read and I'm not really on board with it. I see two possible solutions:

1) Set consts like `OVERCAP_ID = 1000000` and `TOUCH_ID = 1000001` and index by those in code so it's easier to understand what's going on.
2) Stop putting these stats in the map, and just add entries directly in the table below.

I prefer option (2) unless there's some technical limitation I'm not thinking of that would make this more annoying

src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
```diff
@@ -0,0 +1,98 @@
...
+    this.castTimestamp = event.timestamp;
+    this.trackedSpell = spellId;
+  }
+
+  on_byPlayer_removebuffstack(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.SOUL_FRAGMENT_STACK.id) {
+      return;
+    }
+    if (event.timestamp - this.castTimestamp < 100) {
```
kfinch: extract the 100 ms constant into a const above the class, like `const REMOVE_STACK_BUFFER = 100;`

src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
```diff
@@ -0,0 +1,98 @@
...
+      this.totalSoulsConsumed += 1;
+      }
+    }
+
+  on_byPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.SOUL_FRAGMENT_STACK.id) {
+      return;
+    }
+    if (event.timestamp - this.castTimestamp < 100) {
```
kfinch: you should be careful doing number comparison on a possibly undefined value (this.castTimestamp), as you can get unexpected results. In this case I'm pretty sure this statement will always evaluate false when this.castTimestamp is undefined, but in other cases you can get in to trouble. Good to start with an undefined check in such statements.

src/Parser/DemonHunter/Vengeance/Modules/Features/SoulFragmentsTracker.js
```diff
@@ -0,0 +1,48 @@
...
+        this.soulsWasted += 1;
+      }
+    }
+  }
+
+  on_byPlayer_removebuffstack(event) {
+    const spellId = event.ability.guid;
+    if (spellId === SPELLS.SOUL_FRAGMENT_STACK.id) {
+      this.soulsSpent += 1;
+      this.currentSouls -= 1;
```
kfinch: I can't recall, is removebuffstack guaranteed to always be exactly one stack?

src/Parser/DemonHunter/Vengeance/Modules/Talent/SpiritBombSoulsTracked.js
```diff
@@ -21,48 +24,15 @@ class SoulFragments extends Analyzer {
...
-      this.spent += 1;
-      this.actual -= 1;
-    }
-  }
-
   suggestions(when) {
-    const wasterPerGenerated = this.wasted / this.generated;
-    const maximumWaste = Math.floor(0.15 * this.generated);
+    const wasterPerGenerated = this.soulFragmentsTracker.soulsWasted / this.soulFragmentsTracker.soulsGenerated;
+    const maximumWaste = Math.floor(0.15 * this.soulFragmentsTracker.soulsGenerated);
```
kfinch: Looks like 0.15 is a recommendation threshold? Probably should be extracted into a const.

src/Parser/DemonHunter/Vengeance/Modules/Features/SoulFragmentsTracker.js
yajinni: in my case yes, i tested it out, for others i dont know...

src/Parser/DemonHunter/Vengeance/Modules/Talent/SpiritBombSoulsTracked.js
yajinni: Was left over from the person who previously worked on it so i left it, but will clean that up.

src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
yajinni: i initialize it to 0 for that reason. 
castTimestamp = 0;
So it will evaluate false incase there isnt any cast put into it yet. not good?

src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
yajinni: true, will do.

src/Parser/DemonHunter/Vengeance/Modules/Statistics/SoulFragmentsConsume.js
yajinni: I didnt think about doing it that way, since i was pulling all the values from the object i just decided to shove things into it. but yeah i should be able to add it to the table instead.

src/Parser/DemonHunter/Vengeance/Modules/Features/SoulFragmentsTracker.js
kfinch: alright as long as you've checked I think this is fine

Issue #1961, opened 2018-07-23
Arcane Mage Full Support


Sharrq:
to do list of things to add or modify for Arcane Mage Support

- [x] Mana Chart
- [x] Checklist
- [x] Fix suggestion for not casting Arcane Familiar on cooldown (since its a 1hr buff)
- [x] Remove Evo Suggestion for holding AP until Evo is available or will be in 20 seconds
- [x] Add Suggestion for making sure Mana is low (below 20%) on boss kill 
- [x] Add Suggestion for making sure AP is on cooldown at boss kill
- [x] Suggestion for dont cast anything other than Blast and Missiles during AP
- [x] Ensure you dont go OOM during AP
- [x] When casting Arcane Power, make sure you are above 40% mana (without Overpowered only), have 4 arcane charges, and have a RoP charge (if talented)
- [ ] With Time Anomaly stay below 70% mana incase evo proc
- [x] Dont Barrage with 3 stacks of Arcane Charge if you have Rule of Threes (since the 4th cast is free)
- [x] Ensure Arcane Familiar buff is up for the entire fight if talented into it.
- [x] Ensure Arcane Intellect is up for the entire fight (All Mage Specs)
- [x] Arcane Orb hits more than 1 mob per cast on average (totalHits/casts > 1) (With Arcane Orb Talent). Suggest different talent if not hitting more than 1 mob on average
- [ ] Only cast Arcane Missiles with Clearcasting

Will add more with suggestions from mage discord

Gebuz:
Looks like haste from clearcasting is missing, causing GCD/downtime to be hidden.

Sharrq:
Hmm, how would I add that in considering it only affects the channel time of Arcane Missiles?

Gebuz:
you can see here how I handle it for Empowerments for balance (reduces cast time of their respective spells by 15%):
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Druid/Balance/Modules/Core/GlobalCooldown.js

Sharrq:
unless im missing something, that looks like it adjusts the GCD? not the channel/cast time?

Gebuz:
Cast time is usually handled automatically (time between beginCast and CastSuccess), I'm not sure how that would work for a channel spell like AM. Penance had something similar afaik, maybe you can use that.

Dambroda:
This was completed with #2090 ; should be closed

PR #1994, opened 2018-07-26
[Arcane Mage] Arcane Charge Tracker

Sharrq:
Accidentally deleted the branch for the previous PR so here is a new one.

This adds an arcane charge tracker since combat logs dont have the ability to see how many charges you have or list the times you lose arcane charges.

Adding just this for now to make sure its solid before i refer to it from other modules.

Towards #1961


Sharrq:
Looking at some specs that use ResourceTracker, it seems like they all use the classResources info in the cast events, which isnt possible for Arcane Charges because they arent listed in classResources. 

Not sure if i can still use ResourceTracker or not. Im looking through it, but im not 100% sure how this all works.

PR #1995, opened 2018-07-26
[Arcane Mage] Mana Chart

Sharrq:
This adds a mana chart for Arcane. I decided against using the Mana Usage Chart as i didnt feel like the DPS Throughput was all that helpful, so i went with the Mana Level Chart instead as part of playing Arcane well is ending the fight with as little mana as possible (since you should burn your mana at the end of the fight regardless of cooldown state.

Towards #1961 

image

image

### DIFF
--- a/src/Parser/Mage/Arcane/Modules/ManaChart/Mana.js
+++ b/src/Parser/Mage/Arcane/Modules/ManaChart/Mana.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import Analyzer from 'Parser/Core/Analyzer';
+
+import TabComponent from './TabComponent';
+
+class ManaTab extends Analyzer {
+
+  tab() {
+    return {
+      title: 'Mana',
+      url: 'mana',
+      render: () => (
+        <TabComponent
+          parser={this.owner}
+        />
+      ),
+    };
+  }
+}
+
+export default ManaTab;
--- a/src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
+++ b/src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
@@ -1,0 +1,227 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Chart from 'chart.js';
+import {Line} from 'react-chartjs-2';
+
+import fetchWcl from 'common/fetchWclApi';
+
+import ManaStyles from 'Interface/Others/ManaStyles.js';
+
+const formatDuration = (duration) => {
+  const seconds = Math.floor(duration % 60);
+  return `${Math.floor(duration / 60)}:${seconds < 10 ? `0${seconds}` : seconds}`;
+};
+
+class Mana extends React.PureComponent {
+  static propTypes = {
+    reportCode: PropTypes.string.isRequired,
+    actorId: PropTypes.number.isRequired,
+    start: PropTypes.number.isRequired,
+    end: PropTypes.number.isRequired,
+    manaUpdates: PropTypes.array.isRequired,
+  };
+
+  constructor() {
+    super();
+    this.state = {
+      bossHealth: null,
+    };
+  }
+
+  componentWillMount() {
+    this.load(this.props.reportCode, this.props.actorId, this.props.start, this.props.end);
+  }
+  componentWillReceiveProps(newProps) {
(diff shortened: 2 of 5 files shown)


src/Parser/Mage/Arcane/Modules/ManaChart/Mana.js
```diff
@@ -0,0 +1,22 @@
...
+
+import TabComponent from './TabComponent';
+
+class ManaTab extends Analyzer {
+
+  tab() {
+    return {
+      title: 'Mana',
+      url: 'mana',
+      render: () => (
```
emallson: for a simple function+component like this, what about putting it on one line instead, eg

```jsx
render: () => <TabComponent parser={this.owner} />,
```

src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
```diff
@@ -0,0 +1,227 @@
...
+  load(reportCode, actorId, start, end) {
+    return fetchWcl(`report/tables/resources/${reportCode}`, {
+      start,
+      end,
+      sourceclass: 'Boss',
+      hostility: 1,
+      abilityid: 1000,
+    })
+      .then(json => {
+        console.log('Received boss health', json);
```
emallson: stray `console.log`?

src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
```diff
@@ -0,0 +1,227 @@
...
+      },
+      legend: ManaStyles.legend,
+    };
+
+    Chart.plugins.register({
+      id: 'specialEventIndiactor',
+      afterDatasetsDraw : (chart) => {
+        const ctx = chart.ctx;
+
+        chart.data.datasets.forEach(function(dataset, i) {
```
emallson: is there a reason you went with this approach for drawing deaths rather than the one used in the [StaggerPool](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Monk/Brewmaster/Modules/Features/StaggerPoolGraph.js#L148-L155)?

src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
Sharrq: i didnt add that, so it must be in the core manalevelgraph that healers use too. I can potentially remove it though.

src/Parser/Mage/Arcane/Modules/ManaChart/TabComponent/ManaLevelGraph.js
emallson: oh, gotcha. i can go ahead and approve if you like or you can fix, up to you

Sharrq: ill look at the staggerpool one to see what that looks like. 

Ideally id like to figure out how to draw stuff on the graph cause id like to be able to show where lust was cast and when arcane power was cast ... so if the staggerpool one is better then maybe later i can use that to add stuff.

Sharrq: So from what I can tell (with the very limited knowledge I have).

I would need to redo the entire thing because StaggerPool uses the on_toPlayer_death, on_toPlayer_damage, etc stuff which i believe wont work in the existing ManaLevelChart cause it extends React.PureComponent instead of Analyzer? Not sure if thats true or not, but the on_toPlayer_death stuff doesnt do anything if i put it in the ManaLevelGraph stuff.

So I think in order to do it that way I would need to completely start from scratch and build the chart myself (which would involve trying to figure out how to do the WCL Query that it is doing to get the Boss Health instead of Max HP like you are using and getting mana instead of hp. At the moment I just dont know enough to be able to do that without a ton of time to spend hours and hours ripping stuff out and redoing it to try and stumble my way through it. 

So I think for now I should stick to the one I have that works, and then once arcane is fully supported and bfa ready, then maybe i can tackle trying to make a chart from scratch to be able to accomodate things like showing the arcane power casts and lust and such.

emallson: 👍

PR #2001, opened 2018-07-27
Holy priest trail of light talent

yajinni:
image


src/Parser/Priest/Holy/Modules/Talents/TrailOfLight.js
```diff
@@ -0,0 +1,45 @@
...
+    this.active = this.selectedCombatant.hasTalent(SPELLS.TRAIL_OF_LIGHT_TALENT.id);
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.TRAIL_OF_LIGHT_HEAL.id) {
+      return;
+    }
+    this.overhealing += event.overheal || 0;
+    this.totalHealing+= (event.amount || 0) + (event.absorbed || 0);
```
rp4rk: IIRC `event.amount` will always be set to 0 so `|| 0` is unnecessary

PR #2005, opened 2018-07-27
[MMS Hunter] Fixed Hunter's Mark Bug

leapis:
The Hunter's Mark module was broken when the enemyID variable was changed from `event.targetID ` to `encodeTargetString(event.targetID, event.targetInstance)` because calculateMarkDamage was still using enemy.id, which corresponded to event.targetID. Now the sourcing for the enemyID variable (used primarily as an array index) is standardized across the module. 

Example bug: 
WoWA: https://wowanalyzer.com/report/Ck7VZnyRXfc4YjAz/17-Mythic+Portal+Keeper+Hasabel+-+Kill+(5:22)/40-Blaz/events

WCL Hunter's Mark Uptime: https://www.warcraftlogs.com/reports/Ck7VZnyRXfc4YjAz/#fight=17&type=auras&spells=debuffs&hostility=1&ability=257284


Pewtro:
My bad 😓

Issue #2006, opened 2018-07-27
Remove / Update artifact ability suggestions


emallson:
Notably Fire Mages and Havoc DHs are having their artifact ability show up as a suggestion when they haven't taken the relevant talent. Example:

- [x] [Fire](https://www.wowanalyzer.com/report/B28fkqv4hF3YZWCH/14-Mythic+Aggramar+-+Kill+(7:11)/15-Iitroxic)
- [x] [Havoc](https://www.wowanalyzer.com/report/B28fkqv4hF3YZWCH/14-Mythic+Aggramar+-+Kill+(7:11)/23-Archdh)

Looks like Havoc doesn't have the ability anymore period and also is lacking a maintainer, so I may hit that one myself.

ronaldpereira:
Havoc is now updated :)

PR #2012, opened 2018-07-28
[Havoc DH] Class BfA talents and spells updates

ronaldpereira:
I've updated some new spells, added some new talents (manually) and removed the artifact cast suggestion (#2006 ).
Also I've renamed some old "class" props to "className", so React doesn't complain about it anymore.

### DIFF
--- a/src/Parser/DemonHunter/Havoc/Modules/Abilities.js
+++ b/src/Parser/DemonHunter/Havoc/Modules/Abilities.js
@@ -29,7 +29,7 @@ class Abilities extends CoreAbilities {
           base: 1500,
         },
         castEfficiency: {
-          suggestion: true,
+          suggestion: false,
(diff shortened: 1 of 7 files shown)


src/Parser/DemonHunter/Havoc/Modules/Abilities.js
```diff
@@ -29,7 +29,7 @@ class Abilities extends CoreAbilities {
           base: 1500,
         },
         castEfficiency: {
-          suggestion: true,
```
emallson: isn't fury of the illidari just gone? so this could be removed from abilities entirely?

ronaldpereira: makes totally sense. Removing it right now :D

PR #2017, opened 2018-07-28
[Core] Generated Talents from the Blizzard API

niseko:
Had to fix / change / remove quite a lot of stuff to get rid all the errors caused by deleting old talents. 
Probably overlooked some talents that went baseline or similar stuff, as its quite a lot.


Pewtro:
Looks good overall to me.

Issue #1965, opened 2018-07-23
Show viewed characters in report history


MartijnHols:

joshinat0r:
added in #2024

PR #2024, opened 2018-07-30
[Core] add 'recently viewed characters' && add character link to report

joshinat0r:
Adds recently viewed characters to the 'recently viewed logs' div. (#1965 )
Pretty sure this:
```
const mapStateToProps = state => {
  return ({ });
};
 export default withRouter(connect(
  mapStateToProps,
  {
    appendReportHistory,
  }
)(Parses));
```
isn't how it's supposed to be done, but it doesn't work without mapStateToProps.

image

Also adds a link to our character-view to the report page if we have the realm & region info for that character.
image


joshinat0r:
**Don't merge**

Just realized that the realm-field has translated names (as in: our API saves for the Korean realm `Azshara` Azshara, but the character-URL want's the translated version 아즈샤라).

joshinat0r:
Removed the link to our character-page from the report-result-page.
Implementing this would either mean 
- that our API has save the slug or the untranslated name 
- that we add something like a `originalName` to the `REALMS.js` (but the export from the bnet-API doesn't include those)

joshinat0r:
- Readded character link to the report page
- Updated our character API to respect the encoding for each region

src/Interface/Character/Parses.js
```diff
@@ -544,4 +570,13 @@ class Parses extends React.Component {
   }
 }
 
-export default Parses;
+const mapStateToProps = state => {
+  return ({ });
+};
+
+export default withRouter(connect(
+  mapStateToProps,
```
MartijnHols: Since this is unused this should be null

src/Interface/Character/Parses.js
```diff
@@ -544,4 +570,13 @@ class Parses extends React.Component {
   }
 }
 
-export default Parses;
+const mapStateToProps = state => {
+  return ({ });
+};
+
+export default withRouter(connect(
```
MartijnHols: When using multiple decorators please use the `compose` function from the `redux` library.

src/Interface/Others/Report/Results/index.js
```diff
@@ -279,6 +279,17 @@ class Results extends React.PureComponent {
...
               </a>
+              {' '}
+              {characterProfile && characterProfile.realm && characterProfile.name && characterProfile.region && (
+                <Link 
+                  to={`/character/${characterProfile.region.toUpperCase()}/${characterProfile.realm}/${characterProfile.name}/`} 
+                  data-tip={`Show ${characterProfile.realm} - ${characterProfile.name}'s most recent logs`}
+                  className="btn"
+                  style={{ fontSize: 24 }}
+                >
+                  <img src="/favicon.png" alt="Wipefest logo" style={{ height: '1.4em', marginTop: '-0.15em' }} /> {characterProfile.name}
```
MartijnHols: "Wipefest logo"

src/Interface/Others/Report/Results/index.js
```diff
@@ -279,6 +279,17 @@ class Results extends React.PureComponent {
               >
                 <img src={WipefestLogo} alt="Wipefest logo" style={{ height: '1.4em', marginTop: '-0.15em' }} /> Wipefest
               </a>
+              {' '}
+              {characterProfile && characterProfile.realm && characterProfile.name && characterProfile.region && (
+                <Link 
+                  to={`/character/${characterProfile.region.toUpperCase()}/${characterProfile.realm}/${characterProfile.name}/`} 
+                  data-tip={`Show ${characterProfile.realm} - ${characterProfile.name}'s most recent logs`}
```
MartijnHols: Show -> View (view is what the users wants to do, show is what the system would do)

logs -> reports (we call them reports usually in line with WCL's naming)

joshinat0r:
Fixed the naming and updated the export to use compose

server/helpers/wowCommunityApi.js
```diff
@@ -1,20 +1,20 @@
...
+  'eu': 'ru_RU',
+  'us': 'en_US',
+  'tw': 'zh_TW',
+  'kr': 'ko_KR',
+};
 
 export function fetchCharacter(region, realm, name, fields) {
   region = region.toLowerCase();
-  if (!availableRegions.includes(region)) {
+  if (!Object.keys(availableRegions).includes(region)) {
```
MartijnHols: This could be just `availableRegions[region]`

src/Interface/Character/Parses.js
```diff
@@ -13,9 +16,11 @@ import ZONES from 'common/ZONES';
 import SPECS from 'common/SPECS';
 import DIFFICULTIES from 'common/DIFFICULTIES';
 import ITEMS from 'common/ITEMS';
+import REPORT_HISTORY_TYPES from 'Interface/Home/ReportHistory/REPORT_HISTORY_TYPES';
 
 import './Parses.css';
 import ParsesList from './ParsesList';
+import { compose } from '../../../node_modules/redux';
```
MartijnHols: Please change this to an absolute import

src/Interface/Character/Parses.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
+import { appendReportHistory } from 'Interface/actions/reportHistory';
```
MartijnHols: Please put project imports in a separate group from library imports.

joshinat0r:
Fixed

PR #2026, opened 2018-07-30
[Arcane Mage] Fixes & Tweaks

Sharrq:
- Arcane Power has an Unacceptable List now instead of an Acceptable List.
- Moved 40% mana threshold to constant as per previous PR feedback
- Added check for Arcane Power to see if it was available when the boss died
- Modified ManaLevels to not show the ending mana suggestion if the player was dead when the boss died.
- Added Arcane 4pc tier to Haste

Towards #1961


src/Parser/Mage/Arcane/Modules/ManaChart/ManaValues.js
```diff
@@ -58,6 +73,7 @@ class ManaValues extends Analyzer {
           .regular(this.suggestionThresholds.isGreaterThan.average)
```
Gebuz: you can remove .regular and .major

PR #2031, opened 2018-07-31
[Discipline Priest] Mastery

karlpralow:
Complete Rework of the old mastery module since it didn't count absorb effects. The main issue of absorb effects and mastery is we need to know when an absorb occurs if the initial application was buffed or not. 

Also added multiple information in the tooltipt (see screenshot)

image


src/Parser/Priest/Discipline/Modules/Spells/Grace.js
```diff
@@ -24,54 +25,100 @@ class grace extends Analyzer {
...
+      this.healingBuffedByMastery += event.amount;
+    } else {
+      this.healingUnbuffedByMastery += event.amount;
+    }
+  }
+
+  // Finds the absorb buff application event to verify if the initial event was mastery buffed
+  absorbApplicationWasMasteryBuffed(event){
+    let i;
+    for (i = this.applyAbsorbEvents.length - 1; i >= 0; i--) {
```
Gebuz: Instead of using a for loop, you can use `array.slice().reverse().find()` for this.

Issue #1908, opened 2018-07-17
Trinket: Uldir: Frenetic Corpulse


niseko:
![image](https://user-images.githubusercontent.com/2842471/42801125-c4cbe234-899e-11e8-948d-b280718ba164.png)
https://beta.wowdb.com/items/160648-frenetic-corpuscle

Pewtro:
Implemented in #2038

PR #2038, opened 2018-07-31
[Trinket] Frenetic Corpuscle

hpabst:
- Add Frenetic Corpuscle trinket from issue #1908 .

image


hpabst:
Sorry, it looks like merging in the other trinket pull request caused a merge conflict that needed to be fixed. Still getting used to using git here...

Issue #1083, opened 2017-12-26
Add a Mitigation/Defensive Cooldowns Tracker


FaideWW:
Effective mitigation usage is a key piece of data missing from tank analysis currently.  

What I'd like this module to capture is the following:

- Defensive abilities and passives and their DR values/affected damage schools (maybe this can be an extension of `Abilities`).  Somehow this should also count for absorbs like Ignore Pain and various trinkets with absorb effects.
- Incoming damage taken and calculation of pre-mitigation damage intake
- A breakdown of the contribution of individual casts of a defensive ability/passive (how much damage did spell X mitigate?)
- Detection of irregular damage spikes and whether they were properly defended against

I'm planning on implementing this myself; opening this issue to track progress and elicit feedback/suggestions from others on how they'd go about this or other things they'd like to see tracked.

FaideWW:
How I see "damage prevented per spell" working:

Maintain a list of passive DR effects (things like Thick Hide for guardian) and active DR spells (e.g. Barkskin for guardian).  

When a damage taken event occurs, record all the contributing DR effects that were present.  Use the DR formula `damageAfterMitigation = damageBeforeMitigation * (1 - totalDR%)` to obtain the raw damage value, and by extension the total amount of damage mitigated.

Each DR effect contributed a portion of the total reduction.  Determine the percentage of overall contribution for every DR effect, and distribute the amount of damage mitigated according to those percentages.  This is the "amount mitigated" of each DR effect for this particular damage event.

For example:
Thick hide is a 6% DR, Barkskin is a 20% DR. Together they provide 24.8% total DR.

A damage taken event occurs for 300k post-mitigation.  The raw damage value is `300k / (1 - 0.248) ~= 399k`.

Thick hide contributed 23% of the total DR (`6 / sum of DRs`), meaning it mitigated ~92k damage.
Barkskin contributed 77% of the total DR (`20 / sum of DRs`), meaning it mitigated ~307k damage.

yajinni:
This would be awesome for my blood dk: http://www.wowhead.com/spell=48792/icebound-fortitude
If possible i would like to add this to the cooldown tracker tab.

FaideWW:
In 8.0 the WCL event API will include "unmitigated damage taken" as a field on damage events, greatly simplifying this feature. I am going to put this on hold until then.  If someone else would like to pick this up for 7.3.5, feel free to do so.

MartijnHols:
[...]

PR #2041, opened 2018-08-01
[Core] Damage Mitigation Tab

Gebuz:
Early start on #1083. Splits all damage mitigated (excluding asorbs and blocks) evenly between all damage reductions using the formula explained in the issue.

#### Implemented
- All buffs (except Devo & Aura of Sac)
- All debuffs
- Supports both stacking buffs/debuffs and multiple instances (double Ironbark for instance)
- most passives, including armor and versatility.
- Fully absorbed blocks. They do not have a `blocked` property, and it aint likely to be added any time soon. I don't think it's possible to see if the combatant has a shield equipped, so for now it just assigns any unknown mitigation on fully absorbed hits for Prot pala & warror as block.
- Support for Avoidance/Feint. Still needs a complete whitelist to function properly.
- Mastery for Protection Paladin.
- Added armor tracker.
- Added Aura of Sac and Devo Aura.

#### Missing
- Avoidance. Needs a complete whitelist to function properly.
- Mastery for Destruction Warlock. Destro is impossible because it's random, so will just assign any unknown mitigation (up to the mastery maximum) to it.
- Write some text explaining the calculations (Link to Zero's article).

#### Screenshots
image
image
image


Gebuz:
Added a way to assign any unknown damage mitigation to Aura of Sacrifice and Devotion Aura. It doesn't quite work with multiple instances of the same aura yet (currently it just adds it as a single value so it gets a smaller portion of dmg mitigation assigned than it should).
image

Gebuz:
I'm not sure how much more time I'll have to work on this before BFA release, and I likely wont be working on it at all between that and clearing mythic Uldir. I think it's in an okay state for the most part, but it might not be complete enough that we want to show it yet.

If it's not in an acceptable state I will close it for now and come back to it when I find the time. If anybody else wants to pick it up in the mean time let me know you're welcome to, just PM me on discord.

It is currently missing the following:
#### Major
- Whitelist of AOE abilities that is reduced by avoidance (implementation is ready, but it only contains Garothi Worldbreaker abilities at the moment).
- Update stacktracker to include all temporary avoidance and versatility buffs.
- Destruction Mastery. Could be implemented to some extend by guessing, but it is impossible to tell this and other unknown sources like Devotion Aura apart.
- Spirit Link Totem's buff does not show up at all. Could contribute any ~10% unknown DR to it, but just like the above it's hard with multiple unknown sources.

#### Minor
 - Fully absorbed hits does not have a block event. Currently any unknown is contributed to block when this happens, but it might be a good idea to at least contribute the minimum value to devotion aura.
 - Debuffs fade before death events, causing the killing blow to look unmitigated.
 - Perseverence also triggers from other priests Renew if the combatant has the talent.
 - I don't know how Rugged Tenacity works yet. I'd assume it works like block, but the value is so low it mostly doesn't matter anyway.
 - The unknown mitigation assignement to Auras could be improved for the case where the paladin with Devotion Aura is on his own, or when multiple of the same aura are present and AM is active for at least one of them.

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
```diff
@@ -1,63 +1,63 @@
...
-			return;
-		}
-		
-		this.furiousSlashStacks[this.lastFuriousSlashStack].push(event.timestamp - this.lastFuriousSlashUpdate);
-		this.lastFuriousSlashUpdate = event.timestamp;
-		this.lastFuriousSlashStack = 0;
-	}
-}
-
+import Analyzer from 'Parser/Core/Analyzer';
```
MartijnHols: What did you change about this module? I can't see it because everything is marked as changed.

src/Parser/Core/Modules/EnemyInstances.js
```diff
@@ -5,11 +5,11 @@ const debug = true;
 
 class EnemyInstances extends Enemies {
   getEntity(event) {
-    if (event.targetIsFriendly) {
+    if (event.targetIsFriendly && event.sourceIsFriendly) {
```
MartijnHols: Why did you change this? I don't think this is supposed to include damage to friendly targets (e.g. other players or player pets), but your change would lead to including friendly targets given the source isn't friendly (e.g. another player hit by the boss).

src/Parser/Core/Modules/DamageMitigation/Reductions.js
```diff
@@ -0,0 +1,449 @@
+import SPELLS from 'common/SPELLS';
+import MAGIC_SCHOOLS from 'common/MAGIC_SCHOOLS';
+import SPECS from 'Game/SPECS';
+import RACES from 'Game/RACES';
+
+import AOE_TYPE from './AOE_TYPE';
+
```
MartijnHols: I'd prefer it if this file didn't include spells that are class/spec specific. I believe core shouldn't know about class/spec specific things unless they can be applied to others (e.g. externals).

src/Parser/Core/Modules/DamageMitigation/DamageMitigation.js
```diff
@@ -0,0 +1,380 @@
...
+      }
+      if (unknown < - BUFFER_PERCENT) {
+        debug && console.warn('The actual percent mitigated was lower than expected given the mitigations active at the time. Actual: ' + formatPercentage(percentMitigated) + '%, Expected: ' + formatPercentage(multiplicative) + '%');
+      }
+    }
+
+    // If the actual percent mitigated is higher than expected, try to assign the unknown mitigation.
+    if (unknown > BUFFER_PERCENT) {
+      // If the player is using a shield and the hit was fully absorbed, assume the remaining mitigation was block (Block amount is not in the log when the hit fully absorbed).
+      if (event.amount === 0 && (this.selectedCombatant.specId === SPECS.PROTECTION_PALADIN.id ||
```
MartijnHols: I'd prefer it if this file didn't include effects that are class/spec specific. I believe core shouldn't know about class/spec specific things unless they can be applied to others (e.g. externals).

src/Parser/Core/Modules/DamageMitigation/AOE_ABILITIES.js
```diff
@@ -0,0 +1,50 @@
+// White list of all AOE abilities (INCOMPLETE!)
+export default 
+[
+  // Garothi World Breaker
+  {
+    id: 244761,
```
MartijnHols: Please include the spell name in a comment.

src/Parser/Core/Modules/DamageMitigation/AOE_TYPE.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  BOTH: 'Both',
+  AOE_ONLY: 'AOE only',
+  NON_AOE_ONLY: 'Non AOE only',
```
MartijnHols: How about `DIRECT_ONLY`?

src/Interface/Others/DamageMitigationBreakdown.js
```diff
@@ -0,0 +1,87 @@
...
+    const mitigated = this.prepareMitigated(tracker.mitigated);
+
+    let totalMitigated = tracker.totalMitigated;
+    totalMitigated = (totalMitigated === 0) ? 1 : totalMitigated;
+
+    return (
+      <div>
+        {!tracker.isAccurate && (
+          <Danger style={{ marginTop: 10, marginBottom: 10 }}>
+            There is a large amount of damage mitigation which's source is unknown. This is likely caused by our avoidance or versatility tracker being incomplete, our list of abilities affected by avoidance being incomplete or one of the following abilities being partially untrackable in the log, and we can only guess at how much it actually mitigated: Devotion Aura, Aura of Sacrifice, Destruction Warlock Mastery and Spirit Link Totem.
```
MartijnHols: The usage of the word "which's" makes this somewhat hard to read

src/Parser/Core/Modules/DamageMitigation/Armor.js
```diff
@@ -0,0 +1,24 @@
+import Analyzer from 'Parser/Core/Analyzer';
+
+class Armor extends Analyzer {
```
MartijnHols: Why did you make this class instead of using StatTracker?

src/Parser/Core/Modules/DamageMitigation/DamageMitigation.js
```diff
@@ -0,0 +1,380 @@
...
+  initReduction(reduction) {
+    this.mitigated[reduction.id] = {
+      name: reduction.name,
+      amount: 0,
+    };
+  }
+
+  getMitigation(reduction, event) {
+    // Calculate Dampen Harm reduction. 
+    if (reduction.id === SPELLS.DAMPEN_HARM_TALENT.id) {
```
MartijnHols: I'd prefer it if this file didn't include effects that are class/spec specific. I believe core shouldn't know about class/spec specific things unless they can be applied to others (e.g. externals).

src/Parser/Core/Modules/DamageMitigation/Reductions.js
Gebuz: Agreed, dampen harm and the fully absorbed block should be moved to a class specific extension.

src/Parser/Core/Modules/DamageMitigation/AOE_ABILITIES.js
Gebuz: Will do in the future, was just being lazy with the test implementation. On that note, I later realized that there is a good chance that ticks are never reduced by avoidance, in which case the whitelist could simply be an array of ids. I never got around to test that theory though.

src/Parser/Core/Modules/DamageMitigation/Armor.js
Gebuz: Using the armor from the damage event is more accurate. Could update the statTracker to be updated based on that, but it might be confusing that the implementation is so different to the other stats. Also it would not update until a damage event, so if for some reason you tried to get the armor for something else (idk what that would be) between hits it might not return the correct value.

src/Parser/Warrior/Fury/Modules/Talents/FuriousSlashTimesByStacks.js
MartijnHols: Please try updating from master

Gebuz:
Closing for now, will return to it when I'm not spending all my free time playing the game (so likely after uldir mythic progress is over).

PR #2042, opened 2018-08-01
[Survival] Adding average targets hit and Birds of prey talent implementation + GCD

Pewtro:
Updates the changelog and updates Coordinated assault to include birds of prey information. 
Updates GCD for Harpoon.
Updates GCD for Arcane Torrent.
Adds average targets hit for Chakrams, Wildfire bomb and Butchery/Carve. 

Relies on #2019 as it further touches modules in there.


src/Parser/Hunter/Survival/Modules/Spells/WildfireBomb.js
```diff
@@ -0,0 +1,133 @@
...
+    this.active = !this.selectedCombatant.hasTalent(SPELLS.WILDFIRE_INFUSION_TALENT.id); //WFI changes WFB so much that a separate module will be added into talents folder to handle everything from that.
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.WILDFIRE_BOMB.id) {
+      return;
+    }
+    this.casts++;
+    this.currentGCD = (Math.max(1.5 / (1 + this.statTracker.currentHastePercentage), 0.75) * 1000);
```
Gebuz: You could let the GCD module do the calculation for you instead: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/8dde2d7c1d20502216aa303729da48f1d78dc588/src/Parser/Core/Modules/GlobalCooldown.js#L108

src/Parser/Hunter/Survival/Modules/Spells/WildfireBomb.js
```diff
@@ -0,0 +1,133 @@
...
+    if (this.spellUsable.cooldownRemaining(SPELLS.WILDFIRE_BOMB.id) < GCD_BUFFER + this.currentGCD || !this.spellUsable.isOnCooldown(SPELLS.WILDFIRE_BOMB.id)) {
```
Gebuz: The order of these two conditions should be reversed, otherwise it might throw an error from calling `.cooldownRemaining` on a spell that is not on cooldown.

src/Parser/Hunter/Survival/Modules/Spells/WildfireBomb.js
```diff
@@ -0,0 +1,133 @@
...
+    super(...args);
+    this.active = !this.selectedCombatant.hasTalent(SPELLS.WILDFIRE_INFUSION_TALENT.id); //WFI changes WFB so much that a separate module will be added into talents folder to handle everything from that.
+  }
+
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.WILDFIRE_BOMB.id) {
+      return;
+    }
+    this.casts++;
```
Gebuz: it's prefered to use `+=1` instead of `++`

src/Parser/Hunter/Survival/Modules/Spells/WildfireBomb.js
```diff
@@ -0,0 +1,133 @@
...
+    if (this.casts === 0) {
+      this.casts++;
+      this.spellUsable.beginCooldown(SPELLS.WILDFIRE_BOMB.id, this.owner.fight.start_time);
+    }
+    this.targetsHit++;
+    const enemy = this.enemies.getEntity(event);
+    if (this.acceptedCastDueToCapping || !enemy) {
+      return;
+    }
+    if (enemy.hasBuff(SPELLS.WILDFIRE_BOMB_DOT.id) && event.timestamp > this.lastRefresh + MS_BUFFER) {
```
Gebuz: The second part of the condition will basically always return true (unless the bomb can hit the same target multiple times from one cast?). Should it also be considered a bad refresh if you're AOEing and reduce the cooldown below the dot duration?

Pewtro: I only wanted to count it once during AoE, so first one will be true, the subsequent will be false until a new cast comes along. 

No AoE'ing and getting it below DoT duration should be fine (it can happen with Butchery), but should be more edge cases - and if you do so I can imagine Serpent Sting modules will start getting cranky because it'll come at the cost of that uptime.

Pewtro:
As to a single module (in core or shared) would probably be a good idea for a future PR 😅

PR #2045, opened 2018-08-01
Vdh suggestions

yajinni:
Added suggestions for spirit bomb souls consumed and soul cleave.


src/Parser/DemonHunter/Vengeance/Modules/Spells/DemonSpikes.js
```diff
@@ -43,7 +43,7 @@ class DemonSpikes extends Analyzer {
 
   suggestions(when) {
     const hitsWithDSOffCDPercent = this.hitsWithDSOffCD / (this.hitsWithDS+ this.hitsWithoutDS);
-    when(hitsWithDSOffCDPercent).isGreaterThan(0.1)
+    when(hitsWithDSOffCDPercent).isGreaterThan(0.15)
```
ronaldpereira: The recommended is going from 90% uptime to 85%. There's a practical or mathematical explanation to it?

yajinni: We get 2 charges and its a 20 sec cd, relaxing it because its our main active mitigation move. So giving people more leeway in when to time it in case they want to old it for a big spike of damage. Its not like iron fur or boneshield were it can be kept basically one 100% uptime. We need to make some decisions.

ronaldpereira: Understood it and I agree with that 'relaxing' :)

PR #2048, opened 2018-08-02
Realm character page link fix

### LINKED ISSUE #1966: Fix WCL/Armory/Wipefest link on character page: the realm should use the slug
See https://wowanalyzer.com/character/EU/Al'Akir/Zerotorescue/

CubeLuke:
Updated the character page to use the realm slug over the passed in realm via props. This will resolve #1966 

Before, the Warcraft Log, Armory, and Wipefest urls would look like this:
https://www.warcraftlogs.com/character/EU/Al'Akir/Zerotorescue
https://worldofwarcraft.com/en-EU/character/Al'Akir/Zerotorescue
https://www.wipefest.net/character/Zerotorescue/Al'Akir/EU

They will now appear as the following:
https://www.warcraftlogs.com/character/EU/alakir/Zerotorescue
https://worldofwarcraft.com/en-EU/character/alakir/Zerotorescue
https://www.wipefest.net/character/Zerotorescue/alakir/EU


src/Interface/Character/Parses.js
```diff
@@ -249,10 +250,12 @@ class Parses extends React.Component {
     if (!realm) {
       console.warn('Realm could not be found: ' + this.props.realm + '. This generally indicates a bug.');
     }
-    const realmSlug = realm ? realm.slug : this.props.realm;
+    this.setState({
```
MartijnHols: I don't think you should use state for this; the realmSlug/realm is based directly on the value of a property and as such it shouldn't be a part of the state. How about refactoring this method a bit so it's not necessary? For example you could move the code to determine the realm (including the warning) to its own method and then call that here and in the render to get the realm. If you add a second method on top of that to get the slug with the property based fallback, you could also eliminate the code duplication of `realm ? realm.slug : this.props.realm`.

src/Interface/Character/Parses.js
```diff
@@ -249,10 +250,12 @@ class Parses extends React.Component {
...
+      realmSlug: realm ? realm.slug : this.props.realm,
+    });
 
     const urlEncodedName = encodeURIComponent(this.props.name);
-    const urlEncodedRealm = encodeURIComponent(realmSlug);
+    const urlEncodedRealm = encodeURIComponent(this.state.realmSlug);
```
MartijnHols: `setState` is asynchronous, so this won't work reliably.

src/Interface/Character/Parses.js
CubeLuke: This makes sense, I had to leverage off the `componentDidMount` in order to interact with the render so that I didn't have to throw anything into state. Let me know how you feel about an `async` on `componentDidMount`, I imagine it's not the most common thing to do.

src/Interface/Character/Parses.js
```diff
@@ -73,8 +73,9 @@ class Parses extends React.Component {
     this.updateZoneMetricBoss = this.updateZoneMetricBoss.bind(this);
   }
 
-  componentDidMount() {
+  async componentDidMount() {
     this.fetchBattleNetInfo();
+    this.realmSlug = await this.getRealmSlug();
```
MartijnHols: This is basically the same as state, but looking at it again I don't think you can solve it without adding any kind of state. How about setting the realm to the value of `findRealm` in the `setState` on line 320? Then everything updates in the state at once.

src/Interface/Character/Parses.js
```diff
@@ -73,7 +74,7 @@ class Parses extends React.Component {
     this.updateZoneMetricBoss = this.updateZoneMetricBoss.bind(this);
   }
 
-  componentDidMount() {
+  async componentDidMount() {
```
MartijnHols: Is `async` still necessary?

PR #2049, opened 2018-08-02
[Blood Death Knight] "Bones of the Damned"-trait

joshinat0r:
adds a module to keep track of:
- procs and highlights them in the timeline
- average armor gained as statistic
- adds the armor from it to the stat tracker

image
image

**Bones of the Damned**
Armor gained does not increase with amount of stacks of Bone Shield (armor is tied to the BotD buff, not Bone Shield)
Armor gained does increase with multiple Bones of the Damned traits
Multiple traits do not increase the amount of BS charges per proc
Not sure if the % of proccing is increased but the module doesn't really care about that


src/Parser/DeathKnight/Blood/Modules/Features/MarrowrendUsage.js
```diff
@@ -29,10 +34,18 @@ class MarrowrendUsage extends Analyzer {
...
-  MR_GAIN = 3;
+  hasBonesOfTheDamned = false;
+
+  bonesOfTheDamnedProc = 0;
+  totalStacksGenerated = 0;
+
+  constructor(...args) {
+    super(...args);
+
+    if(this.selectedCombatant.hasTrait(SPELLS.BONES_OF_THE_DAMNED.id)) {
```
Gebuz: You can skip the if statement and just set `this.hasBonesOfTheDamned to the value it returns instead

src/Parser/DeathKnight/Blood/Modules/Spells/AzeriteTraits/BonesOfTheDamned.js
```diff
@@ -0,0 +1,86 @@
...
+import BoneShieldTimesByStacks from '../../Features/BoneShieldTimesByStacks';
+
+
+
+export function bonesOfTheDamnedStats(combatant){
+  if(!combatant.hasTrait(SPELLS.BONES_OF_THE_DAMNED.id)){
+    return null;
+  }
+  let armor = 0;
+  for(const rank of combatant.traitsBySpellId[SPELLS.BONES_OF_THE_DAMNED.id]){
```
Gebuz: It would be prefered to use .reduce instead of a for loop

PR #2051, opened 2018-08-02
[General][Azerite Traits] Crystalline Carapace

kcschoeler:
Shows:
Avg. Armor
Total Thorn Damage


src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+import { calculateAzeriteEffects } from 'common/stats';
+import SPELLS from 'common/SPELLS';
+
+export function crystallineCarapaceStats(combatant) {
+  if(!combatant.hasTrait(SPELLS.CRYSTALLINE_CARAPACE.id)) {
+    return null;
+  }
+  let armor = 0;
+  let thornDamage = 0;
+  for(const rank of combatant.traitsBySpellId[SPELLS.CRYSTALLINE_CARAPACE.id]) {
```
MartijnHols: Please use [Array.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) instead of the old for-loops whenever possible.

MartijnHols: This could actually be a `reduce` by making the default value the object you're returning anyway.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import SpellIcon from 'common/SpellIcon';
+import { formatNumber, formatPercentage } from 'common/format';
+import StatisticBox, { STATISTIC_ORDER } from 'Interface/Others/StatisticBox';
+import Analyzer from 'Parser/Core/Analyzer';
+import { calculateAzeriteEffects } from 'common/stats';
+import SPELLS from 'common/SPELLS';
+
+export function crystallineCarapaceStats(combatant) {
+  if(!combatant.hasTrait(SPELLS.CRYSTALLINE_CARAPACE.id)) {
```
MartijnHols: You should only check for the trait in the `constructor` of the module, and set `this.active` to false if the player doesn't have the required trait.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+    armor += ar;
+    thornDamage += td;
+  }
+
+  return {armor, thornDamage};
+}
+
+export const STAT_TRACKER = {
+  armor: (combatant) => crystallineCarapaceStats(combatant).armor,
+  thornDamage: (combatant) => crystallineCarapaceStats(combatant).thornDamage,
```
MartijnHols: This property is not supported in the StatTracker and should be omitted.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+  for(const rank of combatant.traitsBySpellId[SPELLS.CRYSTALLINE_CARAPACE.id]) {
+    const [td, ar] = calculateAzeriteEffects(SPELLS.CRYSTALLINE_CARAPACE.id, rank);
+    armor += ar;
+    thornDamage += td;
+  }
+
+  return {armor, thornDamage};
+}
+
+export const STAT_TRACKER = {
```
MartijnHols: This constant is never used and should either be omitted, or used in the StatTracker.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+  thornDamage: (combatant) => crystallineCarapaceStats(combatant).thornDamage,
+};
+
+class CrystallineCarapace extends Analyzer {
+  armor = 0;
+  thornDamage = 0;
+  constructor(...args) {
+    super(...args);
+    const resp = crystallineCarapaceStats(this.selectedCombatant);
+    console.log(resp);
```
MartijnHols: Please don't leave `console.log`s in production code.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+
+    if(resp === null) {
+      this.active = false;
```
MartijnHols: Please make it more obvious what the `active` check depends on to make the code easier to scan/follow.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+      return;
+    }
+    const {armor, thornDamage} = resp;
+    this.armor += armor;
```
MartijnHols: Why do you *add* the armor/thornDamage here, rather than set it?

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+
+  get uptime() {
+    return this.selectedCombatant.getBuffUptime(SPELLS.CRYSTALLINE_CARAPACE_BUFF.id) / this.owner.fightDuration;
+  }
+
+  get avgArmor() {
+    return this.uptime * this.armor;
+  }
+
+  get avgThornDamage() {
```
MartijnHols: This getter is never used and can be omitted (also it would be a strange stat when you already have the actual damage done).

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+    const resp = crystallineCarapaceStats(this.selectedCombatant);
+    console.log(resp);
+
+    if(resp === null) {
+      this.active = false;
+      return;
+    }
+    const {armor, thornDamage} = resp;
+    this.armor += armor;
+    this.thornDamage += thornDamage;
```
MartijnHols: The name of this property isn't very clear and can easily be confused with `_totalThornDamage`

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  _hitsMitigated = 0;
+  _totalHits = 0;
+  get pctHitsMitigated() {
+    return this._hitsMitigated / this._totalHits;
+  }
+
+  _totalThornDamage = 0;
+  _totalDamage = 0;
```
MartijnHols: This property is unused and can be omitted.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  statistic() {
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.CRYSTALLINE_CARAPACE.id} />}
+        value={(
+            <React.Fragment>
+            {formatNumber(this.avgArmor)} Avg. Armor<br />
+            {formatNumber(this._totalThornDamage)} Dmg
```
MartijnHols: Please use the `ItemDamageDone` component for showing damage values to make it easier to compare across different fights.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/CrystallineCarapace.js
```diff
@@ -0,0 +1,102 @@
...
+            </React.Fragment>
+        )}
+        label={"Stats from Crystalline Carapace"}
```
MartijnHols: The curly brackets can be omitted like regular HTML when the value is just a string.

MartijnHols: We generally only use the icon to indicate what a statistic is about and omit the name from the label. How about just "Gain"? I realize this gets kinda weird for Azerite Powers, we'll have to improve their rendering at a later time.

MartijnHols:
Oof I just noticed most of the code mentioned by the comments was based on the existing Gemhide module. I refactored that module to something cleaner, see here for the changes:

https://github.com/WoWAnalyzer/WoWAnalyzer/compare/a23a29910289ec104ba073426bdd05bd371f8a3b...49d59ced6202b764883defbf262a4cc312cbd1c3#diff-2b688288ec671f739852218504c193ae

kcschoeler:
As you noticed, I used the Gemhide Module to guide me.
Previously had no experience with this tool and JS6 in general.

Will fix a lot of stuff tonight(CET) and check if everything works.
Anyways, thanks for the review and the clues.

PR #2055, opened 2018-08-03
[Death Knight] "Bone Spiked Graveyard"-Trait

joshinat0r:
- adds the tooltip to TraitStatisticBox (not sure if it was missing on purpose or not but gemhide had a tooltip aswell that wasn't showing up)
- updates the "bad" code copied from Gemhide for Bones of the Damned 😄 
- adds Bone Spiked Graveyard statistic for Unholy and Blood

image

**Bone Spiked Graveyard**
Causes x damage and y healing when casting DnD or Defile
Damage and healing cause their own events, so no magic involved


joshinat0r:
You merged yesterday a trait that provides a whopping 248 armor... when the average DK has about 6k armor 😅 
All Blood specific T1 traits feel underwhelming af

PR #2060, opened 2018-08-05
[Havoc && Vengeance DH] Abilities Updated and Code Refactoring

ronaldpereira:
- Shared
  - Added Glide spell to abilities

- Havoc
  - Removed relic traits, as it won't be used anymore from now on
  - Statistic Order moved to position react property (deprecation refactoring)

- Vengeance
  - Removed tier20 piece module, as it won't be used anymore from now on (ToS tier)
  - Removed Demon Spikes checklist and cast efficiency recommendation if Feed the Demon talent is selected
    - Each time a Soul Fragment is consumed, the Demon Spikes CD is reduced by 0.5s, so when this talent was chosen, it was breaking the CD tracker, usage suggestions (it was casted more than suggested casts) and timeline was warning about CD not properly treated
  - Corrected a Sigil of Flame buffSpellID being passed as object. Now passes spell.id
  - Throw Glaive CD is now reduced by haste
  - Statistic Order moved to position react property (deprecation refactoring)


src/Parser/DemonHunter/Havoc/Modules/Abilities.js
```diff
@@ -7,11 +7,9 @@ import Haste from 'Parser/Core/Modules/Haste';
 
 import CoreAbilities from 'Parser/Core/Modules/Abilities';
 
-import UnleashedDemons from './Traits/UnleashedDemons';
 
 class Abilities extends CoreAbilities {
```
MartijnHols: I think you're still missing a bunch of spells from this, please check the console for errors/warnings and fix them.

ronaldpereira: Will fix asap.

Anatta336:
It's great to see all these updates coming in, thank you!

One of the old style `statisticOrder` assignments snuck past, it's hiding in Vengeance/Modules/Talents/RazorSpikes

Checking some random logs I found a couple of odd abilities you might want to add, if only to get rid of the console warnings:
* Chaos Blades (`211796` and `211797` for damage, `247938` for the buff) Although the ability was removed, a legendary item (https://www.wowhead.com/item=151798/chaos-theory) can still provide it.
Example log: https://www.warcraftlogs.com/reports/wHbk9K8NrmzBFMPQ/#fight=24&source=35
* Soul Fragment (`204062`) yet another thing called "Soul Fragment". This appears quite rarely, and I've only seen it on fights with adds. I suspect it might be related to the soul generated when the Demon Hunter gets the killing blow on an enemy.
Example log: https://www.warcraftlogs.com/reports/gDkdq4Z8cYyVraH2/#fight=1&source=11&type=casts

If you're interested in creating a module that applies the cooldown reduction from Feed the Demon you can see how a similar cooldown reduction works for the Raddon's Cascading Eyes legendary: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/DemonHunter/Havoc/Modules/Items/RaddonsCascadingEyes.js
Instead of damage Feed the Demon would be triggered through `on_byPlayer_changebuffstack` and checking for lost Soul Fragments buff stacks (with more cooldown reduction when more stacks are lost.) That'd be for a separate Pull Request though!

PR #2067, opened 2018-08-05
[Frost Death Knight] Bfa fdk abilities

jjs451:
Not finished updating FDK for 8.0.  This PR contains a subset of the changes in an effort to keep the PRs manageable.

### DIFF
--- a/src/Parser/DeathKnight/Frost/Modules/Abilities.js
+++ b/src/Parser/DeathKnight/Frost/Modules/Abilities.js
@@ -83,6 +83,29 @@ class Abilities extends CoreAbilities {
         gcd: null,
         cooldown: 90,
         isUndetectable: true,
+        castEfficiency: {
+          suggestion: true,
+        },
+      },
+      {
+        spell: SPELLS.BERSERKING,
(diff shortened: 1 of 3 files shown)


src/Parser/DeathKnight/Frost/Modules/Abilities.js
```diff
@@ -107,100 +159,123 @@ class Abilities extends CoreAbilities {
...
-          suggestion: true,
-          recommendedEfficiency: 0.90,
+          static: 500,
         },
+        cooldown: 25,
       },
       {
         spell: SPELLS.CHAINS_OF_ICE,
-        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL, //special case, this is rotational for the duration of prepatch
```
Pewtro: Is this only with the chest legendary? In which case you can create a check if it's equipped I assume

jjs451: There's a talent that does the same thing so now that I think about it a permanent check for the talent would probably be the best

PR #2068, opened 2018-08-06
[Hunter] deprecated statisticOrder, Example logs, frizzo crash and Careful Aim module

Pewtro:
This looks like a lot - but a lot of the changes are very minor. They merely add an example log, removes the deprecated statisticOrder or add buffSpellIds for use in the timeline. 

I've highlighted the interesting file changes below:
- Fixing a crash with Frizzos legendary that got changed: 
https://github.com/Pewtro/WoWAnalyzer/blob/aee716b73d617fa7f84dc38f40d29a7ead6d66d9/src/Parser/Hunter/Survival/Modules/Items/FrizzosFingertrap.js
Log: https://www.warcraftlogs.com/reports/XmxJf2w8NpYvPR3H#fight=9&type=damage-done&source=9
image

- The creation of a Careful Aim module: 
https://github.com/Pewtro/WoWAnalyzer/blob/aee716b73d617fa7f84dc38f40d29a7ead6d66d9/src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Log: https://www.warcraftlogs.com/reports/dZqKzB8fvhcAC1Y2#fight=25&type=summary&source=1
image
Curious about feedback for the CA module, the numbers I'm getting for the damage calculation are slightly off constantly, and I'm not sure if the difference is attributed to the newly changed tier212p/4p isn't accounted for (in which case it's fixed in BFA). 
I thought of making a module where I tracked non-crit damage, and if I got a hit that was 100 lower than the last I had received (and against the same target) I'd update aimed shot damage on a non-crit, non-CA hit and calculate from there on out. But damage reductions on bosses (Kin'garoth) would present an issue for that solution (and does for the current aswell I reckon).


src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Aimed Shot has a 50% chance to deal 100% bonus damage to targets who are above 80% health or below 20% health.
+ *
+ * Example log: https://www.warcraftlogs.com/reports/v6nrtTxNKGDmYJXy#fight=16&type=auras&source=6
+ */
+const HIGHER_HP_THRESHOLD = 0.8;
+const LOWER_HP_THRESHOLD = 0.2;
+const AIMED_AP_MOD = 2.07;
+const MASTERY_DMG_MODIFIER_PER_PERCENT = 2.24;
```
Gebuz: Is the mastery modifier different on AS than other shots? Otherwise it would probably be better to update the mastery coef instead.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
```diff
@@ -0,0 +1,107 @@
...
+    if (spellId !== SPELLS.AIMED_SHOT.id) {
+      return;
+    }
+    if (event.attackPower !== undefined && event.attackPower > 0) {
+      this.lastAttackPower = event.attackPower;
+    }
+  }
+
+  _carefulAimProc(event) {
+    const normalDamage = (this.lastAttackPower * AIMED_AP_MOD) * (1 + this.statTracker.currentVersatilityPercentage) * (1 + (this.statTracker.currentMasteryPercentage * MASTERY_DMG_MODIFIER_PER_PERCENT));
```
Gebuz: Shouldn't this include weapon DPS aswell? Iirc the new formula is (AP + WDPS * 6) * APcoef * dmgMods

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
```diff
@@ -0,0 +1,107 @@
...
+
+    const critDamage = normalDamage * 2;
```
Gebuz: You could avoid some code duplication if you checked for crit first by doing `* (event.hitType === HIT_TYPES.CRIT ? 2 : 1)` here and remove all the crit calculations below.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
```diff
@@ -0,0 +1,107 @@
...
+    const caNormalDamage = normalDamage * 2;
```
Gebuz: Should use that `CA_MODIFIER` :P

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Gebuz: Also I assume hunters no longer has a crit dmg multiplier (iirc they had one on AS or TS in Legion)?

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Pewtro: I just pulled from here: https://github.com/simulationcraft/simc/blob/bfa-dev/SpellDataDump/hunter.txt#L513 

Which is in line with https://www.wowhead.com/spell=19434/aimed-shot

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Pewtro: We don't - azerite traits might bring one in which case it'll have to account for that.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Pewtro: I used that in the ``calculateEffectDamage`` bit of the code, but I should get a const for it, I agree.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Pewtro: Our mastery % is range on abilities, our damage increase is 2.24 times as large.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Gebuz: I don't theorycraft weapon classes so I could be wrong, but I'm pretty sure weapon damage still affects the dmg unless it's included in the attack power somehow.

MartijnHols: How about moving this to its own method? It seems very isolated

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
MartijnHols: Please use constants for these magic numbers since right now it's easy to confuse all the 2s

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
MartijnHols: Please use constants for these magic numbers since right now it's easy to confuse all the 2s

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
```diff
@@ -0,0 +1,107 @@
...
+    const caCritDamage = critDamage * 2;
```
MartijnHols: Please use constants for these magic numbers since right now it's easy to confuse all the 2s

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
MartijnHols: I agree with @Gebuz about checking the `hitType`; I think this would make your code much cleaner.

src/Parser/Hunter/Marksmanship/Modules/Talents/CarefulAim.js
Pewtro: It should be included in attack power, but Marksmanship also has a -12% aura I forgot about - I'll get to updating this code today and verifying numbers 👍

Pewtro:
I'll try and think of a more elegant solution to the lone-wolf problem, maybe storing all damage hits until a pet event or something - but for now this should work just fine.

PR #2080, opened 2018-08-10
[BM Hunter] Azerite Trait: Feeding Frenzy

hpabst:
- Adds the BM Hunter azerite trait Feeding Frenzy.

image

### DIFF
--- a/src/Parser/Hunter/BeastMastery/CombatLogParser.js
+++ b/src/Parser/Hunter/BeastMastery/CombatLogParser.js
@@ -49,6 +49,7 @@ import TraitsAndTalents from './Modules/Features/TraitsAndTalents';
 
 //Azerite Traits
 import DanceOfDeath from './Modules/Spells/AzeriteTraits/DanceOfDeath';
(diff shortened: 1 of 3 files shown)


Anatta336:
Accuracy of the extra damage from the trait is tricky to get right. I assume the extra damage-per-tick which Feeding Frenzy provides would be multiplied by any damage buffs (or debuffs) on the player and target, as well as the player's Versatility. So assuming it always does the stated damage can give misleading numbers. See issue: #2088 

Knowing how much damage it provides is really useful information to give the user, but because it's so useful it's important we avoid inaccuracy.

Gebuz:
Just a heads up, the bonus damage from azerite traits is affected by damage modifiers and secondary stats, so simply adding the raw bonus dmg will undervalue the trait (or overvalue it if the target is taking reduced dmg). See #2088 for the discussion on how to handle this (a core solution has not been implemented yet).

hpabst:
Given that a core solution for handling bonus damage from azerite traits is still being worked on, I'm going to close this for the time being until said solution is available.

PR #2087, opened 2018-08-11
[Vengeance DH] Soul Bomb tracker bugfix

ronaldpereira:
- Soul Bomb tracker was showing wrong casts:
``` js
Soul Fragment buff removal counts +2 for the -1 stack with 1 buff stack
-1 for the on_byPlayer_removebuffstack(event)
-1 for the on_byPlayer_removebuff(event)
```
So, removing the on_byPlayer_removebuff(event) counter fixes the Soul Fragments consumed count.

Before:
image
image

After:
image
image

### DIFF
--- a/src/Parser/DemonHunter/Vengeance/Modules/Features/SoulFragmentsTracker.js
+++ b/src/Parser/DemonHunter/Vengeance/Modules/Features/SoulFragmentsTracker.js
@@ -1,34 +1,51 @@
 import Analyzer from 'Parser/Core/Analyzer';
 import SPELLS from 'common/SPELLS/index';
 
+const MAX_SOUL_FRAGMENTS = 5;
+
 class SoulFragmentsTracker extends Analyzer {
 
+  // includes wasted generation
   soulsGenerated = 0;
+
+  // souls generated above the maximum 5 that can be held
   soulsWasted = 0;
+
   soulsSpent = 0;
   currentSouls = 0;
 
   soulsConsumedBySpell = [];
 
-  on_byPlayer_cast(event) {
+  on_byPlayer_changebuffstack(event) {
     const spellId = event.ability.guid;
-    if (spellId === SPELLS.SOUL_FRAGMENT.id) {
-      this.soulsGenerated += 1;
-      if (this.currentSouls < 5) {
-        this.currentSouls += 1;
-      } else {
-        this.soulsWasted += 1;
-      }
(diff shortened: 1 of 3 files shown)


Anatta336:
Thank you for the fixes, these are definitely positive changes! I think there's still something going wrong with the tracker. I've got half way through sorting that out, but am getting called away. So hopefully will finish working that out in a couple of hours. Sorry for the delay!

Anatta336:
I think the best way to track the changing stacks of Soul Fragments is by listening for the `changebuffstack` event. That's a generated event (it doesn't occur in the log we get from WCL, but is added by the `Entities` module) which appears both when stacks change and when a buff is gained or lost. Effectively it combines `applybuffstack`, `removebuffstack`, `gainbuff`, and `removebuff`

This is an example log I looked at. On the events panel, turn on Fabricated Events so you can see `changebuffstack`
/report/JfXnwQTGztqL6cVg/6-Heroic+Antoran+High+Command+-+Kill+(4:05)/1-Aywadispel/events

It can be useful to look at the WCL tracking of the Souls Fragments buff as reference too:
https://www.warcraftlogs.com/reports/JfXnwQTGztqL6cVg#fight=6&type=auras&source=1&ability=203981

Example of using Spirit Bomb at 5 stacks:
```
0:05.122 cast Fracture, generating 2 Souls but there's a delay until the buff updates
0:06.062 gain a stack 	4 souls
0:06.112 gain a stack 	5 souls
0:06.122 cast Spirit Bomb
0:06.112 lost buff	0 souls
```

Example of going over the souls cap:
```
0:16.965 gain a stack	4 souls
0:17.715 gain a stack	5 souls
0:18.014 gain a stack 	6 souls
0:18.031 lose a stack 	5 souls
```
At `0:18.031` there's nothing spending soul fragments, that's the excess being automatically consumed.

To detect overcap we can look for changebuffstack going from 5 to 6. And be careful not to count the transition from 6 back to 5 as souls being spent by an ability.

As we talked about on Discord, I've pushed a commit to this branch to switch the tracking over to using `changebuffstack`.

With these changes, the results of the various trackers now represent what's shown on WCL's graph of Souls Fragments. In that example log it overcaps to 6 just once, and when it reaches 5 souls matches the 9 times Spirit Bomb was cast with 5 souls.

If you're interested I think the way Soul Fragments are tracked could be slimmed down, as currently it's a bit spread out between classes. You could create something similar to `ResourceTracker` where a single class keeps track of how the Souls are generated and spent, which can then be queried by other analzyers to produce suggestions.

Issue #2030, opened 2018-07-31
Subtlety Rogue: Add checklist


MartijnHols:
The IV guide has a great list that can be used as basis: https://www.icy-veins.com/wow/subtlety-rogue-pve-dps-common-worst-mistakes

TsaboTavok:
Hi Zero,

IcyVeins is viewed as a meme in the rogue community, so basing something on it is not a good idea in general. Also if you put the foot note in there, it will loose all credibility. 

In fact just this page of the guide has mistakes. For example: 
It is completely wrong to test if Shadow Dance was cast at 90+ energy, because the only thing that does matter is getting the maximum amount of casts during each dance.
Overcapping Combo Points is a huge mistake, but its not even mentioned .
His Nightblade example has only 1 case, that is not 100% correct. You can actually refresh Nightblade during Dance, if its outside of Symbols and you are not running DarkSahdow.

All of these things are handled by the current analysis. 

I have already started doing any new additions in the style required for checklist and will add it eventually (target is first raid week), so you can assign this issue to me. I will likely not have time to create checklists for Assassination and Outlaw in time tho, so help is more welcome in those specs =)

MartijnHols:
M'kay. Going through the two guides, the IV version looks a lot more comprehensive and worked out than Wowhead's, which is very bare-bones. It's definitely not a *meme*, as they are working hard to provide quality guides for all specs.

1. Hitting max casts is impossible with energy though right? You can't really tell someone "get max casts in SD" as that really doesn't help them one bit achieving that. A guideline such as "try to be at 90 energy or more" is something that is very clear and doesn't require the player to figure it out for themselves. Maybe it should be less than 90 Energy, but I think it would be a nice requirement to get users in the right direction.
2. Sorry, I forgot about that. It should be in the first rule.
3. If the recommendation is to stack cooldowns, then shouldn't symbols be pretty much always up during SD?

I haven't touched Assassination nor Outlaw yet, so not prepared to do a checklist for those. I'll leave sub to you then.

TsaboTavok:
Rogue theorycrafters put guides on http://www.ravenholdt.net/ or on discord pins. The IV rogue guides are a meme, because in the past they were always outdated and as a consequence wrong.

Regarding SD, its completely possible to get maximum amount of casts every dance, and the energy requirement for that depends on your talents. Under some conditions its actually impossible to run out of energy in Dance. I agree that the suggestion can hint at pooling energy.

PR #2093, opened 2018-08-12
[Sub Rogue] Checklist

TsaboTavok:
#2030

Added Subtlety checklist, most of the existing analysis has been integrated in to it.
Fixed Find Weakness analysis.

Fixed a copypasta prepot/secondpot error that was in every checklist v2, by removing the copypastas,

prepotcopypasts

### DIFF
--- a/server/config/config.js
+++ b/server/config/config.js
@@ -10,7 +10,6 @@ function config() {
   const env = process.env.NODE_ENV;
 
   return {
-    devtool: 'inline-eval-cheap-source-map',
     host: process.env.MYSQL_HOST,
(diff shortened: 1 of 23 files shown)


server/config/config.js
```diff
@@ -10,6 +10,7 @@ function config() {
   const env = process.env.NODE_ENV;
 
   return {
+    devtool: 'inline-eval-cheap-source-map',
```
MartijnHols: What does this do?

TsaboTavok: oops, was trying out something but it didnt work out. fogot to remove

PR #2094, opened 2018-08-12
[Hunter] Remove legendaries and tier

Pewtro:
Removes legendaries and tier gear from spelldata and modules.

Unsure if ``common/ITEMS/HUNTER.js`` should just be deleted in its entirety as it's pretty useless right now.


Pewtro:
Okay now I have no clue... no errors locally

src/common/ITEMS/HUNTER.js
```diff
@@ -1,5 +1,6 @@
 import ITEM_QUALITIES from '../ITEM_QUALITIES';
 
+//TODO - Repopulate with BFA items or just delete file?
```
MartijnHols: delete

PR #2097, opened 2018-08-17
[Fury] Fix crash

Pewtro:
It crashes due to the custom prepotion code.
I've commented it out since it's dated and checks for Legion potions, if it ever needs updating and reimplementing the code is still in the directory.


MartijnHols:
Thanks

PR #2105, opened 2018-08-28
Updated a large chunk of FDK for BFA and added Killing Machine Module

jjs451:
I need to update the checklist and then FDK should be fully functional for BFA outside of adding more in depth analysis.  The Rime and Killing Machine modules are for all intents identical, I should maybe have done a more slick approach with inheritance but I'd rather have something ready for Uldir than agonize over it  ¯\_(ツ)_/¯

image
image


src/Parser/DeathKnight/Frost/Modules/Features/HardHowlingBlastCasts.js
```diff
@@ -10,38 +10,26 @@ class HardHowlingBlastCasts extends Analyzer {
...
   statistic() {
     this.nonrimedHB = this.castsWithoutRime;
     return (
       <StatisticBox
         icon={<SpellIcon id={SPELLS.RIME.id} />}
-        value={this.nonrimedHB}
+        value={this.castsWithoutRime}
         label="Howling Blasts without Rime proc"
-        tooltip="You should aim to get this as close to 0 as possible."
+        tooltip="You should aim to get this as close to 0 as possible.  It is a DPS loss to cast Howling Blast without Rime."
```
Gebuz: What about on pull to apply Frost Fever and for AOE?

jjs451: There is almost no situation where it is worth it to cast Howling Blast without Rime.  Rime procs happen often enough there's no need to hardcast it just to apply the dot.  You will get a Rime proc in the first 10 seconds and on AoE it deals half damage to secondary targets.  You would need to be able to hit 10+ targets sustained to consider just doing HB spam.  Though now that you mention it, I should add that its okay to hardcast if you are forced to move off the boss for enough time that you're capping rune (e.g. Intermission phases on KJ you would spam it to generate RP)

PR #2107, opened 2018-08-28
Disc priest smite coefficient & sins fix

rubensayshi:
self explanatory fixes;

- Smite spellpower coefficient was incorrect (only used right now to offset Schism direct damage by damage of 1 Smite to account for the fact that you would be casting Smite instead)
 - Sins of the Many doesn't scale linear at all


Gebuz:
Are you sure that Sins of the Many doesn't follow another function? I'd expect it to have partial values instead of back to back the same values, for instance instead of 8: 4%, 9: 4%, I'd expect something like 8: 4.2%, 9: 3.8%

rubensayshi:
acording to disc community and guides this is the pattern.

rubensayshi:
ah there were some ppl on dummies in SW

6 atonements is tooltip 5% but it's really 5.5%,  
and 7 is 5%

8 is tooltip 4% but it's really 4.5%,  
and 9 is 4%

PR #2108, opened 2018-08-28
Disc priest azerite trait Gift of Forgiveness

rubensayshi:
Depends on https://github.com/WoWAnalyzer/WoWAnalyzer/pull/2107 (first 2 commits are from that PR) for:
- Smite spellpower coefficient was incorrect (only used right now to offset Schism direct damage by damage of 1 Smite to account for the fact that you would be casting Smite instead)
 - Sins of the Many doesn't scale linear at all

----------

Gift of Forgiveness adds to Smite base damage so I've added it to `SmiteEstimation` so that it's included into the Schism direct damage offset.

When you have no Gift traits it still displays a box to show how many of your Smites were casted with 3+ atonements,  
considering the Gift trait is a bit hyped atm I think it's interesting to know how many of your Smite casts would have boosted if you'd have the trait. 

I still need to review it 1 final time myself, but it should be ready for review, I've tested values manually to check that they're correct.

screenie; https://imgur.com/a/E1AjIzH

### DIFF
--- a/src/Parser/Priest/Discipline/Modules/AzeriteTraits/GiftOfForgiveness.js
+++ b/src/Parser/Priest/Discipline/Modules/AzeriteTraits/GiftOfForgiveness.js
@@ -1,0 +1,160 @@
+import React from 'react';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import { formatNumber } from 'common/format';
+import { calculateAzeriteEffects } from 'common/stats';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'Interface/Others/TraitStatisticBox';
+import StatTracker from 'Parser/Core/Modules/StatTracker';
+import SPELLS from 'common/SPELLS';
+import Enemies from 'Parser/Core/Modules/Enemies';
+import DualStatisticBox from 'Interface/Others/DualStatisticBox';
+import SpellIcon from 'common/SpellIcon';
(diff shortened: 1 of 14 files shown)


rubensayshi:
### DON'T MERGE YET

I want to make the azerite trait tooltip reflect the actual rank, figured out how to do it but need to implement it (2morrow)

Gebuz:
please include a screenshot of if when you're done! :D

rubensayshi:
fixed multiple traits and tooltip (requires wowhead tooltips, which will be in next release afaik).

still need to give it one final pass myself, but please do review already

rubensayshi:
should be ready for review, screenie; https://imgur.com/a/E1AjIzH

src/Parser/Priest/Discipline/Modules/AzeriteTraits/GiftOfForgiveness.js
```diff
@@ -0,0 +1,160 @@
...
+        if (giftedHealing > 0) {
+          this.giftedHealing += giftedHealing;
+        }
+      }
+    }
+  }
+
+  statistic() {
+    if (this.hasGift) {
+      // to make the tooltip reflect multiple ranks of the trait we use a bit of a hacky way to figure out which ilvl would result in displaying the sum of the bonus damage
```
rubensayshi: this is a bit funky I know, but I don't know any other way we could make the tooltip reflect the sum of multiple traits...

PR #2112, opened 2018-08-29
WIP/DISCUSS: Meticulous Scheming Azerite Trait through generic SecondaryStatProcTrait class.

rubensayshi:
#### This PR is purely opened for discussion on the subject 

So I want to implement a bunch of Azerite Traits that have secondary stat procs since it will be interesting to compare them to the more complicated Azerite Traits and I feel it would be nice if both are in wowanalyzer (plus StatTracker needs them).

As I mentioned in discord there are quite a few of these stat proc traits, many of them shared between various classes.  
It seems most useful to implement all the simple ones in one go, either by having 1 Module that does them all or by having a base class to extend, either way the could would be fairly similar but that's a discussion point.

I implemented it for 1 trait in this PR so we have something to look at / talk about; Meticulous Scheming,  
normally this is only a haste proc but I fiddled with `AZERITE_SCALING.generated.json` to make it think it's a proc for all 4 secondaries, just so that we can see it in action.

rest of my comments are inline in the PR.

screenie of the result; https://imgur.com/a/TJ5KfLX note that afaik there's no procs with more than 2 so it would look less funky in reality.


src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/SecondaryStatProcTrait.js
```diff
@@ -0,0 +1,143 @@
...
+import Analyzer from 'Parser/Core/Analyzer';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'Interface/Others/TraitStatisticBox';
+import StatTracker from 'Parser/Core/Modules/StatTracker';
+
+const CRIT = 'crit';
+const VERSA = 'versa';
+const MASTERY = 'mastery';
+const HASTE = 'haste';
+
+const STATS = {
```
rubensayshi: bit skeptical about defining function names as strings but at this point in the code there's no `statTracker` instance yet ...

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/SecondaryStatProcTrait.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  get uptime() {
+    return this.selectedCombatant.getBuffUptime(this.buffId) / this.owner.fightDuration;
+  }
+
+  /**
+   * to make the tooltip reflect multiple ranks of the trait we use a bit of a hacky way to figure out which ilvl would result in displaying the sum of the bonus damage
+   */
+  determineIlvlForSum() {
```
rubensayshi: I took this code from my other PR, it's a bit funky and off slightly but I don't see another way to come to an ilvl that when passed into the wowhead tooltip will be close to the sum of the traits effects.

src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/SecondaryStatProcTrait.js
```diff
@@ -0,0 +1,143 @@
...
+      return [
+        `${formatPercentage(avgPercentage)}% avg ${statInfo.name}`,
+        `<b>${statBuff} ${statInfo.name} Rating</b>`,
+        `<b>${formatNumber(avgStat)}</b> average Haste Rating (${formatPercentage(avgPercentage)}%)`,
+      ];
+    });
+
+    const valueStrings = statStrings.map((strings) => strings[0]).join(' and ');
+    const tooltip1Strings = statStrings.map((strings) => strings[1]).join(' and ');
+    const tooltip2Strings = statStrings.map((strings) => strings[2]).join(' and ');
```
rubensayshi: above code is a bit ... idk if someone can think of something cleaner? I'm not much of a map reduce flatten ninja.

src/Parser/Core/Modules/StatTracker.js
```diff
@@ -236,13 +236,17 @@ class StatTracker extends Analyzer {
     \****************************************/
 
     // region Azerite Traits
+    [SPELLS.GEMHIDE.id]: GEMHIDE_STATS,
+    [SPELLS.METICULOUS_SCHEMING.id]: combatant => Object.values(combatant.traitRanks(SPELLS.METICULOUS_SCHEMING.id)).reduce((haste, rank) => {
+      return haste + calculateAzeriteEffects(SPELLS.METICULOUS_SCHEMING.id, rank)[0];
+    }, 0),
```
rubensayshi: Gemhide uses a helper defined inside the Gemhide module, but because I use `statTracker` as dependancy in my Module that causes a circular import.

I think if I clean it up with an easy to use helper it would be fine to define it in here...

src/common/AZERITE_SCALING.generated.json
```diff
@@ -1390,13 +1390,13 @@
...
-        "avg": 0
+        "avg": 2.11156
       },
       "725608": {
-        "avg": 0
+        "avg": 2.11156
       },
       "725609": {
-        "avg": 0
+        "avg": 2.11156
```
rubensayshi: like I said in the PR description, ignore this, just to make it display all 4 stats for demo purpose for now.

Gebuz:
I appreciate the effort to make a general module for secondary stat azerite traits, but I'm not sure if it's the best solution. Most of the azerite traits have some kind of condition that we would likely want to include in the statisticbox or the tooltip. For instance, how many times did Meticulous scheming proc, and how many of those were not triggered? How many times did I extend blood rite by getting a kill? How many times did something proc, vs. passives not needing to display # of procs. 

About the stat box, I would suggest not showing percentage stat gains, as most players does not know the difference between multiplicative and additive and some classes have multipliers on them that would make the numbers wrong. Just showing ratings would be more accurate and less confusing, and it is generally easier to compare to other gear pieces just using ratings.

rubensayshi:
ye I realized that as well, but getting % uptime in the report would be a good start already I think?

MartijnHols:
Hey, since this PR has been stale for some time, I'll close this PR. Please make a new PR when you're ready.

PR #2114, opened 2018-08-31
Basic Unholy Death Knight support for BFA

jjs451:
The aim here was to remove dead modules and update existing ones so they still work when Uldir launches.  Still got a ways to go with adding new modules for new talents and updating the checklist but its a start.


src/Parser/DeathKnight/Unholy/Modules/Features/Abilities.js
```diff
@@ -194,39 +205,40 @@ class Abilities extends CoreAbilities {
...
         gcd: {
           base: 1500,
         },
       },
       {
-        spell: SPELLS.WRAITH_WALK_TALENT,
+        spell: SPELLS.RAISE_ALLY,
         category: Abilities.SPELL_CATEGORIES.UTILITY,
-        enabled: combatant.hasTalent(SPELLS.WRAITH_WALK_TALENT.id),
+        cooldown: 600,
```
Gebuz: I would suggest removing the cooldown since you are not limited by the normal cooldown in a raid setting.

jjs451: Good call, I'll add the change to whatever PR I do next.

PR #2116, opened 2018-08-31
[WW] Add Azerite traits to spells and remove leftover Legion stuff

Juko8:
-  Adds Azerite traits and corresponding buffs/debuffs/energize ids. Energize is important to prevent chi tab from crashing and the rest will be used later.

- I missed a lot of references to Tierpieces in previous PR. This should have all of them

- Storm, Earth and Fire module isn't as useful after the BFA changes and could sometimes be misleading, it's therefore being removed.


Juko8:
I'll do that in a coming PR

PR #2122, opened 2018-09-05
[Core] Removes most used legendaries from encounter stats 

Pewtro:
Should fix this showing: 
https://i.gyazo.com/thumb/1200/bbdc511ff8f9671102e11f2932a202b6-png.jpg

Can't test right now because WCL - but theoretically seems fine.


Pewtro:
image
Finally got WCL functioning again and it seems to work, but we're running into problems with all the names being "unknown item", but the tooltips are working correctly so I'll have to investigate a bit to find where it's going wrong.

PR #2132, opened 2018-09-06
[Bug] [Core] character-lookup error messages

joshinat0r:
Adds an alert/message to the character-lookup when the response status-code is 500 or the response itself is not okay.

image

image


jjs451:
Thanks!

PR #2115, opened 2018-08-31
Be well prepared BFA version

ArthurEnglebert:
Hello, first big MR ever in my life, as well as first participation in a pure JS oriented project, so please be gentle.

Do know that I still want to test it a bit more, squash commits into a nice one, etc. but I wanted to know if I'm going into the right direction or if there is a lot to change in my way of developing this part ( fixing #1982 )

Things I still need to check : 
- Test with a player report that does have sockets in his gear not socketted yet.
- Implementing the deprecated API ? (I only implemented the CheckList2 API, and i see it is not supported for all specs)
- Squash commits into a nice one after correcting what you guys might want me to.
- Be able to show in the recap something like "0 / 0" as green when there is no socket in the gear of the player (at the moment, I only show those when there is at least one socket, or we get a nasty division by 0 somewhere)

Waiting for your remarks, answers, etc.  :)

Have a nice day,
Arthur

PR #2133, opened 2018-09-06
Update the enchant checker to support BfA enchants.

TJMoats:
I don't know if there is a `Good` way to handle offhand weapon checks. Checking the icon string hit all of the off hand items that I found correctly.

First time contributing, I don't know if I missed anything.

### DIFF
--- a/src/Parser/Core/Modules/Items/EnchantChecker.js
+++ b/src/Parser/Core/Modules/Items/EnchantChecker.js
@@ -40,7 +40,16 @@ class EnchantChecker extends Analyzer {
 
   get enchantableGear() {
     return Object.keys(this.constructor.ENCHANTABLE_SLOTS).reduce((obj, slot) => {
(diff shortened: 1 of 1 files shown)


joshinat0r:
I hate to break it to you but there is already a PR #2115 open for it 😅  (+ other stuff for the well prep-rules).

Not even sure what's the best approach for this, merge the enchants now and keep the gems, flasks and foods outdated or merge them all in one go.

TJMoats:
I'm of the opinion that PR's should be smaller and self contained. gems, flasks and foods should each be a separate PR imo.

joshinat0r:
Yeah, that what we're also prefering. Approving a small PR is much easier.

Anyways, just looked into the other PR and it looks like its just updating the old enchants and adding gems + flasks (those are new modules).
This one should be good to go and thanks for the PR 👍

PR #2138, opened 2018-09-07
[MW] Spreadsheet Tab Updates

anom0ly:
Additional spreadsheet tab updates based on feedback from TC'ers.

image


src/Parser/Monk/Mistweaver/Modules/Spells/AzeriteTraits/UpliftedSpirits.js
```diff
@@ -1,18 +1,20 @@
...
 
 class UpliftedSpirits extends Analyzer {
   static dependencies = {
     combatants: Combatants,
+    statTracker: StatTracker,
   };
 
   healBuff = 0
-
+  critMod = 1
```
jjs451: Does this need to be a member variable?  It looks like you only use it in on_heal and it gets reset every call

anom0ly: No it doesn't. I modified to just use the variable within the on_heal event.

PR #2142, opened 2018-09-07
Reworked sanctify

yajinni:
OLD:
image

NEW:
image

Did this based on feedback from automatickjak, one of the guide writers


src/Parser/Priest/Holy/Modules/Spells/Sanctify.js
```diff
@@ -43,13 +47,22 @@ class Sanctify extends Analyzer {
...
         value={`${report.sancAvgHits.toFixed(2)}`}
-        label="Average hits"
-        tooltip={`A measure of how many targets were effectively healed by your Holy Word: Sanctify. Over 80% overhealing on a hit is considered a "miss". You missed ${report.sancMissedHits} of ${this.casts * 6} potential hits.`}
+        label="Avg effective hits per cast"
+        tooltip={`A measure of how many targets were effectively healed by your Holy Word: Sanctify. </br>
+                  Over 80% overhealing on a hit is considered a "miss". </br>
+                  <b>Total players hit:</b> ${this.totalHits} </br>
+                  <b>Total casts:</b> ${this.casts} </br>
+                  <b>Hits per cast :</b> ${hitsPerCast.toFixed(2)} </br>
+                  <b>Players overhealed for 80%+ hit:</b> ${this.overhealHit}</br>
```
emallson: This should probably be "Missed Hits" since you defined a miss above.

src/Parser/Priest/Holy/Modules/Spells/Sanctify.js
```diff
@@ -43,13 +47,22 @@ class Sanctify extends Analyzer {
...
+                  <b>Total of no hits:</b> ${noHits}`}
```
emallson: The wording of this is awkward, but i'm not really sure what to call it. However, it should be at least clarified what a "no hit" is (number of potential hits that didn't have targets and thus didn't hit?) up where you define what a "miss" is.

yajinni:
Made the requested updates

PR #2143, opened 2018-09-07
added statbox for fort buff

yajinni:
image


jjs451:
I'm a little hesitant to merge this PR, though the code looks good.  The suggestion from the module is very useful, as some newer players might not be aware of the buff but I'm not sure how useful a statistic box is.  A raid buff a very binary thing.  You either buffed the raid with fortitude or you didn't.  99% of the time, that will be 100%.  For the other 1% we have the suggestion to alert them.  We try to avoid as much unnecessary clutter as possible so users don't get desensitized by a bunch of information and then miss something important.  I've asked the other maintainers to take a look at this and see if they agree or if I'm totally wrong about this

PR #2163, opened 2018-09-09
[WW] Add Swift Roundhouse Azerite statistic

Juko8:
image


Juko8:
Yeah, the Touch of Death calculation update is intended, forgot to mention that.

PR #2169, opened 2018-09-10
Added context to prepotion checker

domainzero:
Prepotion checker will now suggest the proper potions based on class/spec. Recommendations for potion use were pulled from Icy Veins as a baseline since most of those guides are pretty well vetted. 

This code is kind of hideous and if anyone has suggestions to clean it up I would be more than happy to learn how to make it better.


Pewtro:
Isn't it possible to create a function where you simply pass the specId to check which spec we're analyzing. The function would do pretty much what you're doing already, but then it could adjust a variable (such as potionIcon, suggestionText etc) , so you don't have to run the de facto same code in both suggestions

domainzero:
Yeah, that makes a lot more sense. I'll see what I can do.

PR #2173, opened 2018-09-11
[Core] refresh encounter stats each day

joshinat0r:
Is currently only updating the stats each month, resulting for some specs in empty stats.


joshinat0r:
Shouldn't force push to a branch when there's already a PR on it 😅 
Updated it, the `cache`-value is now the current calendar-week (eg. its currently going to send 37)

PR #2179, opened 2018-09-12
Added support for Seabreeze

buimichael:
img

### DIFF
--- a/src/Parser/Core/Modules/Items/BFA/Dungeons/Seabreeze.js
+++ b/src/Parser/Core/Modules/Items/BFA/Dungeons/Seabreeze.js
@@ -1,0 +1,50 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import Analyzer from 'Parser/Core/Analyzer';
+import { formatPercentage, formatNumber } from 'common/format';
+import { calculateSecondaryStatDefault } from 'common/stats';
+
+/**
+ * Seabreeze -
+ * Equip: Your spells have a chance to grant you 89 Haste, increasing to 890 over 30 sec.
+ */
+class Seabreeze extends Analyzer {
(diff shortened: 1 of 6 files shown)


src/Parser/Core/Modules/Items/BFA/Dungeons/Seabreeze.js
```diff
@@ -0,0 +1,50 @@
...
+  statBuff = 0;
+  statBudget = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasMainHand(ITEMS.SEABREEZE.id);
+
+    if(this.active) {
+      this.statBuff = calculateSecondaryStatDefault(300, 84, this.selectedCombatant.getItem(ITEMS.SEABREEZE.id).itemLevel);
+      this.statBudget = calculateSecondaryStatDefault(340, 200, this.selectedCombatant.getItem(ITEMS.SEABREEZE.id).itemLevel);
```
jjs451: Why are there two different base ilvls?

buimichael: It's just calculating the Stat budget of a generic main hand weapon. In this case a generic ilvl 340 weapon has a Stat budget of 200.

jjs451: Right, I'm just wondering why you picked 340 as a base when Seabreeze has a base of 300.

buimichael: the statBudget is used to compare the Seabreeze with another generic main hand weapon of the same ilvl. I selected a random 340 main hand (not seabreeze) for comparison, it could be ilvl 300, 310, 315, doesn't really matter. It seems as Seabreeze doesn't have the exact same scalings as non-proc weapons.

PR #2180, opened 2018-09-12
Fix TouchOfDeath Meridian Strikes trait damage error

terndrup:
Error was flagged by @MartijnHols - basically, if the combatant didn't have any meridian strikes traits it was assumed that the helper getTraits rank function would return 0, but instead it returned undefined which caused the error.

Fixed by checking for undefined and manually setting modifier to 0.0 if undefined.

Also reduced some line lengths for readability.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,9 +16,9 @@
       "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0-beta.51.tgz",
       "integrity": "sha1-6IRK4loVlcz9QriWI7Q3bKBtIl0=",
       "requires": {
-        "chalk": "2.4.1",
-        "esutils": "2.0.2",
-        "js-tokens": "3.0.2"
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^3.0.0"
       },
       "dependencies": {
         "chalk": {
@@ -26,9 +26,9 @@
           "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.1.tgz",
           "integrity": "sha512-ObN6h1v2fTJSmUXoS3nMQ92LbDK9be4TV+6G+omQlGJFdcUX5heKi1LZ1YnRMIgwTLEj3E24bT6tYni50rlCfQ==",
           "requires": {
-            "ansi-styles": "3.2.1",
-            "escape-string-regexp": "1.0.5",
-            "supports-color": "5.4.0"
+            "ansi-styles": "^3.2.1",
+            "escape-string-regexp": "^1.0.5",
+            "supports-color": "^5.3.0"
           }
         }
       }
@@ -40,7 +40,7 @@
       "dev": true,
       "requires": {
         "@lingui/conf": "2.4.0",
-        "babel-generator": "6.26.1"
+        "babel-generator": "^6.26.1"
       }
     },
     "@lingui/babel-plugin-transform-js": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -16,19 +16,19 @@
       "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0-beta.51.tgz",
       "integrity": "sha1-6IRK4loVlcz9QriWI7Q3bKBtIl0=",
       "requires": {
-        "chalk": "^2.0.0",
-        "esutils": "^2.0.2",
-        "js-tokens": "^3.0.0"
+        "chalk": "2.4.1",
```
MartijnHols: Please revert changes to this file. You should only run `npm install` to avoid getting incompatible/dangerous dependencies.

src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
```diff
@@ -15,6 +15,7 @@ const GALE_BURST_VALUE = 0.1;
 
 // Meridian Strikes Azerite trait adds damage to Touch of Death
 const meridianStrikesDamage = traits => Object.values(traits).reduce((obj, rank) => {
+  console.log(obj, rank);
```
MartijnHols: Please don't leave console.logs in production code

src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
```diff
@@ -45,14 +46,29 @@ class TouchOfDeath extends Analyzer {
...
+
+    // Calculate Meridian Strikes Modifier Damage if combatant has MS traits
+    const merStrikesTraits = 
+      this.selectedCombatant.traitsBySpellId[SPELLS.MERIDIAN_STRIKES.id];
+
+    const merStrikesMod = 
+      (merStrikesTraits === undefined) ? 0.0 : meridianStrikesDamage(
+      this.selectedCombatant.traitsBySpellId[SPELLS.MERIDIAN_STRIKES.id]).damage;   
+
+    this.expectedBaseDamage =
```
MartijnHols: This code style makes the code very long

terndrup: You prefer the 281 long single line?

MartijnHols:
Because we want updates, but controlled and not in a random PR

src/Parser/Monk/Windwalker/Modules/Spells/TouchOfDeath.js
MartijnHols: Absolutely, most devs have a wide screen monitor, making the vertical space is limited

terndrup: I *strongly* disagree.. long lines are terrible for human code readability, in fact, almost every programming language style-guide enforce a 80 character column limit. To my knowledge, so do all major JavaScript ES6 style guides.

terndrup: That's a poor argument imo. Relying on client-side tools instead of writing readable code, by that logic you may as well not use any style guides at all. A better alternative is to introduce linting to the project to try and enforce consistentency by linting against battle tested linter guides (e.g. Google ES6, React ES6, AirBnb etc.).

MartijnHols: Unfortunately those read to code that is harder to read, and the max column limit (especially 80) is based on very outdated ideas that do not fit the project

terndrup: Sorry, very outdated? Literally every major 2018 JS Style Guide spec has the 80 column limit (100 absolute max) and if you look around at code samples you are not likely to find many 100+ lines. If you don't want that in this project for some reason that's one thing, but to say that it is "outdated" is just incorrect.

MartijnHols: iirc the 80 column limit dates back to time cards and have hardly changed even though we've begun using widescreen monitors up to 4 thousand pixels in width. With soft wrapping and other client side tools for your IDE, you can set it up locally the way you want. You can have things use the optimal space available and be as easy to read as the screen size you allocate, even if you have just one file open, multiple in a split screen or multiple windows dragged around the desktop. Those tools don't work well if the code already has every single variable on a separate line (how this becomes more readable is beyond me). 

But still anyone who takes ownership of a spec can decide for the most part how to do it themselves. We'd prefer consistency across the codebase, but in the end you are free to choose how to do opinionated/preference things that aren't enforced by eslint rules. It also complicates things when other people would want to help out, or when working on shared components, so long story short consistency is encouraged.

terndrup: You're right, it dates back to IBM Punchcards which had 80 columns. As for why it is still the industry standard in 2018, many discussions like ours now have taken place and the general consensus seems to be that shorter lines are still a lot more readable for humans, hence why the industry still enforces line length rules of less than 80-100 characters.

The tools you mentioned (soft-wrap in IDEs etc) are not usually encouraged because they typically simply cut lines on the last keyword that would violate the line-length, so they often make the readability worse. 

At the end of the day, it is a matter of personal opinion as you say, and if you want to enforce long lines in this project that's your prerogative, what bothered me was that you flagged it in the PR and then said that this was outdated practice, when in fact it is the industry standard for almost every programming language, including JavaScript ES6. As a professional in the industry, I would be crucified if I submitted a 281 char code line.

Either way, let's move on ;)

PR #2207, opened 2018-09-16
[Core] Meticulous Scheming trait

joshinat0r:
Adds Meticulous Scheming's haste to the statTracker and adds a suggestion when the player failed to trigger `Seize the Moment` every `Meticulous Scheming`-proc

image
image

### DIFF
--- a/src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/MeticulousScheming.js
+++ b/src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/MeticulousScheming.js
@@ -46,7 +46,7 @@ class MeticulousScheming extends Analyzer {
   }
 
   on_byPlayer_refreshbuff(event) {
-    // this.handleBuff(event);
+    this.handleBuff(event);
   }
 
   handleBuff(event) {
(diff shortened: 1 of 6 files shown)


src/Parser/Core/Modules/Spells/BFA/AzeriteTraits/MeticulousScheming.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import SpellLink from 'common/SpellLink';
+import { formatPercentage } from 'common/format';
+import { calculateAzeriteEffects } from 'common/stats';
+import Analyzer from 'Parser/Core/Analyzer';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'Interface/Others/TraitStatisticBox';
+
+const meticulousSchemingStats = traits => Object.values(traits).reduce((obj, rank) => {
```
emallson: why construct an object as the return value here? in both cases you use it it is immediately destructured or has the `haste` field used directly.

emallson:
(actually the conflict was trivial, so I handled it)

PR #2208, opened 2018-09-16
Brain Freeze munching change to check for icicle stacks

Dambroda:
With this change, icicle stacks will be taken into account when suggesting whether or not the player munched too many Brain Freeze procs. Previously, there was simply a comment that said it might be ok to overwrite procs if you had 3 icicles, this change should have the analyzer do that checking for the player


src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,123 @@ const debug = false;
 // Brain Freeze appears to always fall after Flurry cast, but not always on same timestamp. Giving a margin here.
 const PROC_WINDOW_MS = 100;
 
+// When Glacial Spike is talented, it's OK to overwrite Brain Freeze procs if you have at least this many icicles.
+const ICICLE_BREAKPOINT = 3;
```
Sharrq: I would have to look at events to figure out the order of things, but I would think it would only be acceptable for Brain Freeze to be refreshed if they have 3 or 4 icicles, not 1,2, or 5. Because at 5 they should be casting Glacial Spike if they have Brain Freeze and not Frostbolt (which can give more icicles).

Dambroda: Very true! But wouldn't that be more appropriately handled in the Glacial Spike module? Since it's a mistake to not be casting GS at that point, the BF proc munching is more of a side-effect?

Sharrq: it would technically be both. 

It is a overwrittenproc and therefore was a waste, and its a bad usage of Glacial Spike. Just because its a bad usage of Glacial Spike doesnt mean they didnt waste a Brain Freeze proc.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,123 @@ const debug = false;
...
-	}
+  get wastedPercent() {
+    return (this.wastedProcs / this.totalProcs) || 0;
+  }
 
-	get usedProcs() {
-		return this.totalProcs - this.wastedProcs;
-	}
+  get usedProcs() {
+    return this.totalProcs - this.wastedProcs - this.okOverwrittenProcs;
```
Sharrq: I might not be thinking straight because im sick .... but is this accurate??

If im reading this right, if you have GS and you refresh the buff with 1-2 stacks then you are marking it as an OverwrittenProc and if you have 3,4,5 then its an OKOverwrittenProc, therefore it wouldnt make sense to subtract the OKOverwrittenProcs from the totalprocs because thats going to lower the number of properly used procs.

So shouldnt usedProcs be totalProcs - wastedProcs?

Also, i dont think you need to count the number of OKOverwrittenProcs unless you are planning on listing that number somewhere. If not, it should be enough to just not count them as an overwrittenProc and therefore not have it count against them.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,123 @@ const debug = false;
...
-			return;
-		}
-		this.overwrittenProcs += 1;
-		this.totalProcs += 1;
-    debug && console.log("Brain Freeze proc overwritten @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
+      return;
+    }
+    this.totalProcs += 1;
+
+    if (!this.selectedCombatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id)) {
```
Sharrq: This one doesnt matter too terribly much, but i believe the standard is to set a variable within the constructor at the top of the file (Check out GlacialSpike, it should have this in it

constructor(...args) {
    super(...args);
    this.active = this.selectedCombatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id);
    this.hasEbonbolt = this.selectedCombatant.hasTalent(SPELLS.EBONBOLT_TALENT.id);
  }

this.active doesnt need to be included, but you could do
this.hasGlacialSpike = this.selectedCombatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id);

and then below do !this.hasGlacialSpike ... to shorten things a little. 

But again, its really minor and probably doesnt matter too terribly much

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: Actually now that I am thinking about it, it's slightly more complicated. Because I'm pretty sure you are not supposed to cancel your FB cast even if you get your BF proc while you have 5 icicles already (sub 100ms or so). So maybe for now it's simpler to just keep the 3 cutoff and figure out the best way to handle the event ordering with the procs w/ five icicles and getting the proc at the same time as your fifth icicle?

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,123 @@ const debug = false;
...
+  lastFlurryTimestamp;
+
+  overwrittenProcs = 0;
+  okOverwrittenProcs = 0;
+  expiredProcs = 0;
+  totalProcs = 0;
+  flurryWithoutProc = 0;
+  currentIcicleCount = 0;
+
+  handleIcicleStacks(event) {
```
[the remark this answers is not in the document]
Dambroda: Thank you! Not very familiar with the API yet

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: I am tracking OKOverwrittenProcs and OverwrittenProcs separately, but an OKOverwrittenProc is not a used proc which is why it gets subtracted. And yes I do display it somewhere, it shows up like this: 
image

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: I was following what was already in the file, in your opinion should I change it?

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Sharrq: Fair. I think its technically better to cancel the Frostbolt to cast GlacialSpike but I would have to talk to the Theorycrafters in the Mage Discord about it, but for now we can probably count a refresh at 5 stacks as ok

Dambroda: Later on I will work on adding the specific logic about checking to see when you got the proc. For more immediate concerns, I just realized that maybe 3 isn't the right number to use for the icicle cutoff. Because you get the icicle before the onRefresh I believe, (if BF procs) technically that means if you have 3 icicles when BF onRefreshes, you had 2 icicles when casting the FB. So that means the cutoff should be 4 because that would mean you had 3 when starting casting the FB. Correct me if I'm wrong there, but I think that's how the events go

Sharrq: You should be able to look at the events tab to see the order that it happens.

Though that might not be the worst case, cause i think the guides say 3 because if you have 2 icicles and are in the middle of casting Frostbolt (i.e. you will have 3 at the end of the cast), then you hold Brain Freeze.

Dambroda: Ok I checked, it looks like when applying a fresh buff it is on the same timestamp but before the icicle. When refreshing, brain freeze applies slightly after icicles (~150ms), probably due to the forgiveness window that allows you to flurry after frost bolt without munching. So I think I'll use the buffer option you mentioned to check and see how many icicles there were during the FB cast instead. I also double checked the mage discord and the consensus is you cast flurry->IL if you have 0, 1, or 2 icicles DURING the FB cast. The altered-time guide is worded incorrectly according to them. Luckily if they make up their minds and its different it should be pretty easy to change that const value

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,102 @@ const debug = false;
...
+    this.totalProcs += 1;
+
+    if (!this.glacialSpikeTalented) {
+      this.overwrittenProcs += 1;
+      debug && console.log("Brain Freeze proc overwritten w/o GS talented @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
+      return;
+    }
+
+    // Get the number of icicles the player had during the cast of the frostbolt
+    const stacksDuringCast = this.selectedCombatant.getBuff(SPELLS.ICICLES_BUFF.id, this.owner.currentTimestamp - 250).stacks || 0;
```
Sharrq: Alright, so I took a look at the events and it looks like Icicles gains a stack right before frostbolt successfully completes its cast, therefore the stack count on getBuff when Frostbolt is cast will include the stack it just gained.

image

Also for confirmation I asked the theorycrafters in the Mage Discord and they mentioned that it is actually at 4 Icicles that you should hold Brain Freeze. and that if you have 3 Icicles and are in the middle of casting Frostbolt, that you should hold Brain Freeze cause you will have 4 Icicles after Frostbolt completes.

So i think we can drop the buffer and we should change the breakpoint to 4.

The confusion on the 3 or 4 icicles is probably my fault. I was told to change it to 4 and I did, and then i started getting complaints that wowanalyzer was saying 4 but the Altered Time guides were saying 3 so I changed it back to 3 cause thats what Altered Time said. Turns out the Altered Time guide was wrong and we should be using 4 not 3.

Sharrq: Actually, scratch that. Looks like Brain Freeze is gained before Icicles. I should really reorder that, but leave the buffer in for now.

image

Sharrq: So one issue with his this works and how the events seem to stack up.

Lets say the player has 3 Icicles when they begin casting in each of these scenarios.

1. The player starts casting Ebonbolt. Ebonbolt doesnt give icicles, so getBuff is going to say that they had 3 Icicles when they gained the Brain Freeze Proc, and that is correct.

2. The player starts casting Frostbolt. When Frostbolt completes, the event order is as follows
image

So in this case, getBuff would also say that they had 3 Icicles, but they had actually just gained the 4th.

In Scenario 1, they should be using the Brain Freeze cause they have less than 4 Icicles whereas in Scenario 2 they should be holding it because they would be gaining their 4th Icicle from Frostbolt.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,102 @@ const debug = false;
 // Brain Freeze appears to always fall after Flurry cast, but not always on same timestamp. Giving a margin here.
 const PROC_WINDOW_MS = 100;
 
+// When Glacial Spike is talented, it's OK to overwrite Brain Freeze procs if you have at least this many icicles during the frost bolt cast
+const ICICLE_BREAKPOINT = 3;
```
Sharrq: I talked to the theorycrafters in the Mage Discord and as it turns out, the breakpoint is actually 4 not 3. This is likely a confusion on my part as during prepatch they told me to change it to 4 and I did ... and then there were complaints last week in the Mage Discord about how wowanalyzer contradicted Altered Time and Altered Time said 3. So I assumed Altered Time was right and changed it back to 3 so it matched.

But as it turns out, Altered Time was wrong and it should be 4.

Also confirmed that if they get Brain Freeze on the 3rd Frostbolt and are in the middle of casting Frostbolt, then they should hold Brain Freeze cause the current Frostbolt would give them 4.

image

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,104 @@ const debug = false;
 // Brain Freeze appears to always fall after Flurry cast, but not always on same timestamp. Giving a margin here.
 const PROC_WINDOW_MS = 100;
 
+// When Glacial Spike is talented, it's OK to overwrite Brain Freeze procs if you have at least this many icicles during the frost bolt cast
```
Sharrq: I talked to the theorycrafters in the Mage Discord and as it turns out, the breakpoint is actually 4 not 3. This is likely a confusion on my part as during prepatch they told me to change it to 4 and I did ... and then there were complaints last week in the Mage Discord about how wowanalyzer contradicted Altered Time and Altered Time said 3. So I assumed Altered Time was right and changed it back to 3 so it matched.

But as it turns out, Altered Time was wrong and it should be 4.

Also confirmed that if they get Brain Freeze on the 3rd Frostbolt and are in the middle of casting Frostbolt, then they should hold Brain Freeze cause the current Frostbolt would give them 4.

image

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -145,61 +163,57 @@ class BrainFreeze extends Analyzer {
...
-	statistic() {
+  }
+
+  suggestions(when) {
+
+    when(this.overwriteSuggestionThresholds)
+      .addSuggestion((suggest, actual, recommended) => {
+        let suggestBuilder;
+        if (this.glacialSpikeTalented) {
+          suggestBuilder = suggest(<React.Fragment>You overwrote {formatPercentage(this.overwrittenPercent)}% of your <SpellLink id={SPELLS.BRAIN_FREEZE.id} /> procs incorrectly. You should hold your proc for <SpellLink id={SPELLS.GLACIAL_SPIKE_TALENT.id} /> if you have {ICICLE_BREAKPOINT} or more <SpellLink id={SPELLS.ICICLES_BUFF.id} /> during your <SpellLink id={SPELLS.FROSTBOLT.id} /> cast, otherwise you should use it immediately.</React.Fragment>);
```
Sharrq: This suggestion isnt very clear, and it assumes they are already casting Frostbolt ... which they might not be.

I would suggest something along the lines of 

You should hold your proc for <SpellLink id={SPELLS.GLACIAL_SPIKE_TALENT.id} /> if you have {ICICLE_BREAKPOINT} or more <SpellLink id={SPELLS.ICICLES_BUFF.id} /> or if you will have {ICICLE_BREAKPOINT} icicles after your current cast completes, otherwise you should use it immediately.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,111 @@ const debug = false;
 // Brain Freeze appears to always fall after Flurry cast, but not always on same timestamp. Giving a margin here.
 const PROC_WINDOW_MS = 100;
 
+// When Glacial Spike is talented, it's OK to overwrite Brain Freeze procs if you have at least this many icicles *during* the frost bolt cast
+const ICICLE_BREAKPOINT = 3;
```
Sharrq: This should be 4

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,111 @@ const debug = false;
...
+      this.overwrittenProcs += 1;
+      debug && console.log("Brain Freeze proc overwritten w/o GS talented or by EB @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
+      return;
+    }
+
+    // Get the number of icicles the player had *after* the cast of the frostbolt
+    const stacks = this.selectedCombatant.getBuff(SPELLS.ICICLES_BUFF.id).stacks || 0;
+
+    // The breakpoint is based on the number of icicles the player had *during* the cast of frostbolt, so increment by 1 to check against current icicles
+    if (stacks < ICICLE_BREAKPOINT + 1) {
```
Sharrq: Just change the breakpoint to 4 instead of adding one for the sake of making the suggestion work right.

Its cleaner to just make the breakpoint correct instead of adding one here and complicating things. I would personally rather have the logic clean and easy for the next person that comes along to be able to understand it and not reference ICICLE_BREAKPOINT in the suggestion than to muddy up the logic just to be able to use ICICLE_BREAKPOINT in the suggestion.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
```diff
@@ -11,81 +11,111 @@ const debug = false;
...
+      this.okOverwrittenProcs += 1;
+      debug && console.log("Acceptable Brain Freeze proc overwritten w/ GS talented @ " + formatMilliseconds(event.timestamp - this.owner.fight.start_time));
+    }
   }
 
-	on_byPlayer_cast(event) {
-		const spellId = event.ability.guid;
+  on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.FROSTBOLT.id || spellId === SPELLS.EBONBOLT_TALENT.id) {
```
Sharrq: I believe the prefered style is to return on any spells that are not the ones you are looking for

i.e. 
if (spellId !== SPELLS.FROSTBOLT.id && spellId !== SPELLS.EBONBOLT_TALENT.id) {
    return;
}

I only mention it cause im sure someone else will when they go to approve it.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: The reason it mentions the cast of Frostbolt is because you only cast brain freeze after a Frostbolt to shatter it. Otherwise it's a mistake to just cast the flurry and ice lance if it wasn't before the FB. So the thought process the player should have is "do I have 3 or fewer icicles as I'm casting this FB?" If so, they cast flurry right as the FB finishes, otherwise they hold the proc. Personally I believe it's easier for most people to understand "do I have 3 icicles now as I'm casting" than "will I have 4 icicles when this cast finishes". Thoughts?

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: The reason it mentions Frostbolt is because you only cast brain freeze after a Frostbolt to shatter it. Otherwise it's a mistake to just cast the flurry and ice lance if it wasn't before the FB. So the thought process the player should have is "do I have 3 or fewer icicles as I'm casting this FB?" If so, they cast flurry right as the FB finishes, otherwise they hold the proc.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: Well yes, but neither of those are relevant to the munching as both are only at five icicles (at which point if you are casting flurry without one of those two its a different mistake).

And that's the point of the comment, it explains the variable. The same person might see 4 and be used to the "during the cast" breakpoint value and think it should be 3, not knowing the order that the events happen. I did the commenting and breakpoint value for the sake of consistency, not just for the suggestion text. In my opinion (and I could be wrong), I believe the easiest and clearest way to explain it to the player (and maintainer) is by putting it in terms of what you'll be thinking while playing the spec. "Ok, I'm casting frostbolt. If I have 3 or more icicles right now, I do not cast flurry after my frost bolt" vs "Will I have 4 or more after my cast?"

Dambroda: As I'm thinking, perhaps a compromise would be to move the +1 logic to another variable directly beneath the ICICLE_BREAKPOINT (something like ICICLE_AFTER_BREAKPOINT = ICICLE_BREAKPOINT +1), and then have a comment explaining the order of events and you can use either one depending on the context you're using the variable in. That way a maintainer, if the event order or breakpoint changes, can just change them at the top by seeing the comments and determining whether they need to change the breakpoint or remove the +1.

Dambroda: Well hypothetically, let's say you decide to use it in more than one place, it just makes it more difficult to maintain when you use magic numbers like that. By the same token I could just completely remove icicle_breakpoint and just use 4 in the if statement and it would work fine. But by adding in the const I can keep the entire module consistent and easy to maintain (as long as the variables are explained enough)

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: That doesn't work when you have two different conditionals though. It's also checking for flurry in there with two different paths. If I return early not matching FB or EB then the flurry conditional will never get checked

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Dambroda: I'm not sure if there is recommended max length on suggestions, but if there isn't I would be more than happy to explain more in-depth how BF holding works in the suggestion itself to try and make it clearer. EX:

You overwrote 5% of your Brain Freeze procs incorrectly. You should use Brain Freeze immediately following a Frostbolt if you have fewer than 3 Icicles during the cast. If you have 3 or more icicles during the Frostbolt cast, save the Brain Freeze proc to Shatter Glacial Spike.

Sharrq:
@MartijnHols For some background info on what he is trying to accomplish and how the spells interract with eachother.

Casting Frostbolt and using your Brain Freeze Procs on an instant cast Flurry each give you +1 Icicle (Icicles caps at 5). These are the only 2 spells that give Icicles.

When not playing Glacial Spike, you should be using your Brian Freeze Procs immediately to prevent them from getting overwritten (Brain Freeze is a proc chance on Frostbolt and a guaranteed proc on Ebonbolt).

However, when you are playing Glacial Spike (which can only be cast at 5 Icicles), it is beneficial to always cast Glacial Spike in conjunction with a Brain Freeze proc (So the rotation is always Glacial Spike > Flurry > Ice Lance. The "breakpoint" in this case is if you have 4 Icicles when you get your Brain Freeze Proc (or will have 4 icicles once your current cast completes) then you should hold the proc so you can cast it with Glacial Spike. If you have less than 4 icicles (or wont have 4 once your current cast completes), then you should use the proc asap.

On live currently, this module counts an overwritten proc as bad regardless of whether you are playing Glacial Spike or not and just has a note that this is sometimes acceptable when playing Glacial Spike. His PR is to not count overwrites as bad if the user has 4 or 5 icicles.

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Sharrq: good to know. I only suggested it cause it has been added as a suggestion on my PRs before so i figured someone would say it

src/Parser/Mage/Frost/Modules/Features/BrainFreeze.js
Sharrq: I dont have a problem with using the constant in the suggestion, i just dont really like that the constant doesnt represent the same value throughout the module. Cause in the suggestion it means 3 but in the if statement it means 4 (or 3+1) .... Its probably just a personal annoyance and doesnt matter in the grand scheme of things ....

Though the point i was trying to make earlier about assuming they are casting Frostbolt is that Yes I understand that they should not be casting Ebonbolt unless they have 5 Icicles and no Brain Freeze .... but people are going to do it anyway. Yes all the guides say to never cast Ebonbolt unless you have 5 icicles and need a brain freeze proc, but I can guarantee you that there will be people that choose to cast Ebonbolt on cooldown regardless. So I was wanting the suggestion to use the number 4 not 3 because I prefer to assume that not everyone is going to play correctly and i would rather the suggestions be accurate regardless of whether they are playing correctly or not.

Sharrq:
I added the comment in line but since its hidden now, ill repost here in response to why i was suggesting using 4 in the suggestion instead of 3.

I dont have a problem with using the constant in the suggestion, i just dont really like that the constant doesnt represent the same value throughout the module. Cause in the suggestion it means 3 but in the if statement it means 4 (or 3+1) .... Its probably just a personal annoyance and doesnt matter in the grand scheme of things .... also im not a developer/programmer, so me not understanding might just be because i dont do this for a living (i do some scripting and such but i dont know much about javascript and im not a developer ... which is why i am not allowed to review/approve/merge things :P )

Though the point i was trying to make earlier about assuming they are casting Frostbolt is that Yes I understand that they should not be casting Ebonbolt unless they have 5 Icicles and no Brain Freeze .... but people are going to do it anyway. Yes all the guides say to never cast Ebonbolt unless you have 5 icicles and need a brain freeze proc, but I can guarantee you that there will be people that choose to cast Ebonbolt on cooldown regardless. So I was wanting the suggestion to use the number 4 not 3 because I prefer to assume that not everyone is going to play correctly and i would rather the suggestions be accurate regardless of whether they are playing correctly or not.

Dambroda:
Totally see where you are coming from, the only reason I am using 3 instead of 4 is simply because I believe the wording "3 during the FB cast" is easier to understand and learn than "4 after the FB cast". With the rename of the variable it should now be clear what the const refers to. You are 100% correct that 4 is the breakpoint, so now the rename helps describe exactly what I was using 3 for and why it's in the suggestion.

However, the rest about using EB incorrectly should be in the EB or GS module IMO as it's not related to brain freeze munching (though EB->BF munching is handled now as well to make sure it's not ignored with any number of icicles). I'd even put it in as a major problem. I may run some sims later to see how bad the DPS loss is to cast EB on CD instead of saving to generate BF procs.

Sharrq:
My understanding from the TCs in the Mage Discord was that the difference between the old GS Playstyle of hit GS as soon as you get to 5 Icicles regardless of Brain Freeze + EB on cooldown vs the new playstyle was something like 2%, which was substantial enough for them to change the guides, but imo thats still a pretty small increase. I would be interested to see what the difference is between EB on CD vs holding for the guaranteed Brain Freeze .... 

Granted that was in prepatch, so it might be more substantial now. 

I dont really want to keep arguing about it back and forth though, so if you want to word it with 3 instead of 4 then fine, i can deal with it ... and if it causes any confusion then ill just change it.

Dambroda:
Turns out it's pretty dps neutral to use EB-on-CD vs saving it to generate BF procs for GS. Talking less than 100dps gain. The higher your mastery goes the more relevant it becomes. 

I'll just let you make the final call for the number. I have both options for the suggestion completed. They are:

1) You overwrote X% of your Brain Freeze procs incorrectly. You should use Brain Freeze immediately following Frostbolt if you have fewer than 3 Icicles during the Frostbolt cast. If you have 3 or more Icicles during the Frostbolt cast, save the Brain Freeze proc for Glacial Spike.

2) You overwrote X% of your Brain Freeze procs incorrectly. You should save Brain Freeze for Glacial Spike if you have 4 Icicles, or would have 4 Icicles after your current Frostbolt cast. Otherwise, use the proc immediately.

Sharrq:
👍

PR #2211, opened 2018-09-17
Defensive Stance statistic for damage prevented (DRPS/Total)

Cyniikal:
Defensive stance statistic module added to show DRPS prevented and total damage prevented.


Cyniikal:
Okay, I'll rewrite it.

Cyniikal:
Leaving the shield tooltip icon below the statistic in so I can also add a damage LOST due to Defensive Stance floating to the right of it.

src/Parser/Warrior/Arms/Modules/Talents/DefensiveStance.js
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+    this.addEventListener('damage', this.handleDamageTaken, {
+      toPlayer: true,
+    });
+  }
+
+  handleDamageTaken(event) {
+    if(this.selectedCombatant.hasBuff(SPELLS.DEFENSIVE_STANCE_TALENT.id)) {
+      this.totalDamageMitigated += (event.mitigated || 0);
```
joshinat0r: Are you sure about adding all the mitigated damage during def stance? I would assume that this stat shows me how much def stance did for me, not how much damage I mitigated during it. The mitigated-prop from WCL includes values from all sources (like externals, avoidance, armor, etc).

You could use something like this (just ripped that from an old version of boneshield, but you should get the idea on how we calculate the DR for just those spells):
```
const preMitigatedBoneShield = (event.amount + event.absorbed) / (1 - BONE_SHIELD_DR);
this.boneShieldMitigated += preMitigatedBoneShield * BONE_SHIELD_DR;
```

joshinat0r:
Thanks 👍

PR #2229, opened 2018-09-19
Resto Druid - Added azerite trait lively spirit

buimichael:


Gebuz:
You should probably add a note about it not tracking the int when cast on another target than the player.

buimichael:
Thanks. In addition I actually noticed a bug that will occur if the player gets an external innervate will cause this module so show wrong numbers. Closing this PR until fixed.

PR #2233, opened 2018-09-20
Defensive Stance Statistic (Damage-lost)

Cyniikal:
Added damage lost statistic to the Defensive Stance tooltip. Icons scale based on the tradeoff between Damage Prevented and Damage Lost.


joshinat0r:
Looks good 👍 
Just a minor thing for the future, we prefer to keep the else/ else if in the same row like this:
```
if(tradeoff > MAX_WIDTH) {
      tradeoff = MAX_WIDTH;
} else if(tradeoff < 1 - MAX_WIDTH) {
      tradeoff = 1 - MAX_WIDTH;
}
```
nothing major, just something to keep the code-base consistent.

Not sure if you're planning on doing more stuff for arms, but I think updating the suggestion for Def. Stance might be a good idea. I think Zero made that.
A new suggestion could check for large hits against the player and / or if a player had Def Stance the whole fight over active (or for low damage taken phases).

image

And don't forget to keep the changelog updated, just make a new maintainer entry in `src/CONTRIBUTOR.js` and add a new entry to the `CHANGELOG.js`-file in the arms-folder.

PR #2237, opened 2018-09-20
Resto druid - Added exception to bad regrowth usage if used on target…

buimichael:
… with low health

### DIFF
--- a/src/Parser/Druid/Restoration/CHANGELOG.js
+++ b/src/Parser/Druid/Restoration/CHANGELOG.js
@@ -4,6 +4,11 @@ import { Yajinni, blazyb } from 'CONTRIBUTORS';
 
 export default [
   {
+    date: new Date('2018-09-20'),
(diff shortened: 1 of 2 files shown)


joshinat0r:
Are you sure that the 123ms buffer is sufficient enough to check if the Regrowth was casted with/without a CC proc?

But the rest looks fine, should be good to go when you're sure about the buffer-time.

buimichael:
@joshinat0r No i'm not 100% sure, the few logs I tested with 123ms seemed sufficient (even lower seemed fine). Do you have any suggestions on what this buffer should be?

joshinat0r:
The 123 just seemed a little bit like a placeholder :D
I don't know, never had to use one. And searching for `buffer` in all files shows pretty much everything from 100ms - 200ms. Don't think there is the one and only buffer that works perfectly. I'll go ahead and merge this one.

PR #2239, opened 2018-09-20
[Beast Mastery] Feeding Frenzy and Primal Instincts

Pewtro:
image
image

The damage aspect of feeding frenzy is still NYI - since it follows the pattern of being a static amount of damage, and therefore had to track exact value. 

Logs:
https://www.warcraftlogs.com/reports/m9KrNBVCtDALZpzT#source=5&type=summary&fight=1
https://www.warcraftlogs.com/reports/QdrGMj1wFqnacXNW#fight=1&type=auras&source=13

### DIFF
--- a/src/Parser/Hunter/BeastMastery/CHANGELOG.js
+++ b/src/Parser/Hunter/BeastMastery/CHANGELOG.js
@@ -6,6 +6,11 @@ import SPELLS from 'common/SPELLS';
 
 export default [
   {
+    date: new Date('2018-09-20'),
+    changes: <React.Fragment>Added two azerite trait modules, one for <SpellLink id={SPELLS.PRIMAL_INSTINCTS.id} /> and an initial version for <SpellLink id={SPELLS.FEEDING_FRENZY.id} /></React.Fragment>,
+    contributors: [Putro],
+  },
+  {
     date: new Date('2018-08-12'),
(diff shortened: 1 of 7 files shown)


joshinat0r:
Looks good to me, but Primal Instincts' box looks a little bit filled imo. Is it necessary to show the uptime in the value and in the tooltip?
Also on Primal Instinct, `up to X charges gained`, would it be useful to show the user a `you wasted Y times by popping Apect while sitting at max charges`?

Other than that, should be ready for a merge.

Pewtro:
> Also on Primal Instinct, `up to X charges gained`, would it be useful to show the user a `you wasted Y times by popping Apect while sitting at max charges`?

I can't 😦 Barbed Shot resets aren't trackable, so I can never know if they're at max stacks or not.

joshinat0r:
My bad, didn't knew that Barbed Shot has a reset mechanic, Blizzard should really add them to the combatlog.

BTW: Looks like you didn't added the charges & cooldown to Barbed Shot because of the reset, [Avengers Shield](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Paladin/Protection/Modules/Features/SpellUsable.js) and [Shield Slam](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Warrior/Protection/Modules/Features/SpellUsable.js) have the same issue. The tracking isn't really 100% accurate, but might be useful.

PR #2248, opened 2018-09-20
[Frost Mage] Added two more spells that don't cancel casts, and added movement abilities to Cancelled Casts suggestion text

Dambroda:


Sharrq:
👍

PR #2252, opened 2018-09-21
Resto druid - Added stonebark talent tracker

buimichael:
img

report/7Pp8rQgCJmbTtFDz/71-Mythic+Fetid+Devourer+-+Wipe+16+(5:35)/4-Blazyb


src/Parser/Druid/Restoration/Modules/Talents/Stonebark.js
```diff
@@ -0,0 +1,51 @@
...
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.STONEBARK_TALENT.id);
+  }
+
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+
+    if(this.hotTracker.hotInfo[spellId] == null) {
+      this.healing += calculateEffectiveHealing(event, STONEBARK_HOT_INCREASE);
```
sMteX: I know the wording of the talent is a little ambiguous but shouldn't this increase be active only if the target has Ironbark active on them? Or is this condition somehow baked into the condition on the line above and I just don't understand?

PR #2256, opened 2018-09-22
[Frost Mage] Add statistics for trait Whiteout; add Dambroda to frost mage contributors

Dambroda:
image

With these commits, the azerite trait Whiteout will now have a statistics box that tracks how much bonus Ice Lance damage is dealt. It will also tally up the total Frozen Orb cooldown reduction applied to display to the user. 

Also, in the last commit I added myself to the contributors list for frost mage.

### DIFF
--- a/src/Parser/Mage/Frost/Modules/Traits/Whiteout.js
+++ b/src/Parser/Mage/Frost/Modules/Traits/Whiteout.js
@@ -1,0 +1,127 @@
+import React from 'react';
+
+import {formatNumber} from 'common/format';
+import SPELLS from 'common/SPELLS';
+import {calculateAzeriteEffects} from 'common/stats';
+
+import TraitStatisticBox, { STATISTIC_ORDER } from 'Interface/Others/TraitStatisticBox';
+
+import Analyzer from 'Parser/Core/Analyzer';
+import HIT_TYPES from 'Parser/Core/HIT_TYPES';
+import EnemyInstances from 'Parser/Core/Modules/EnemyInstances';
+import SpellUsable from 'Parser/Core/Modules/SpellUsable';
+import StatTracker from 'Parser/Core/Modules/StatTracker';
+
+// Enemy debuffs that provide the triple damage multiplier onto Ice Lance
+const SHATTER_EFFECTS = [
+  SPELLS.WINTERS_CHILL.id,
+  SPELLS.FROST_NOVA.id,
+  SPELLS.ICE_NOVA_TALENT.id,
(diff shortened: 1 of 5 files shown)


src/Parser/Mage/Frost/Modules/Traits/Whiteout.js
```diff
@@ -0,0 +1,127 @@
...
+
+    this.bonusDamagePerIceLance *= FROST_MAGE_AURA;
+  }
+
+  on_byPlayer_cast(event) {
+    if (event.ability.guid !== SPELLS.ICE_LANCE.id) {
+      return;
+    }
+    if (this.spellUsable.isOnCooldown(SPELLS.FROZEN_ORB.id)) {
+      this.frozenOrbReductions += 1;
```
joshinat0r: We usually try to provide more infos for CDRs, not just how much it effectively reduced the CD but also how much was wasted. Just to show players what the trait/talent could've done.

Not sure if its usefull in this scenario or if its usually good to delay frozen orb. Tell me if you think its pointless information for the trait and I'll merge it, the rest looks good.

Dambroda: The trait technically does not influence the APL of Frost mage at all; as in it's not worth using Frozen Orb earlier than you usually would (you normally use on CD anyway, but certain fights you hold it)  to get a few extra reductions in. 

However I am planning on an addition to this module anyway to estimate the Frozen Orb damage added via the trait. The math would look like this:

extraOrbs =  frozenOrbReducedCD / ORB_CD; // If I reduced 30 seconds total and FO is a 60s cd, I got 0.5 exta orbs)
// averageOrbDamage is the total damage provided by frozen orb over the fight divded by the number of casts
estimatedFrozenOrbDamage = averageOrbDamage * extraOrbs;

If that makes sense to add, I can commit that before the merge. Or I can put it in as a separate PR. If it doesn't make sense to add, then I would say just merge this one if approved :)

Dambroda: The idea was mainly geared towards just trying to remove some of the difference in value of what is shown on the analyzer vs what sims estimate. For example, at 355 ilvl, sims show 400 estimated DPS but only taking the ice lance damage into account shows 200dps. Obviously some of that discrepancy can be accounted for with just fight mechanics, etc, but part of it is missing the frozen orb damage. I will be looking more into this to see if I can make it more accurate (with a new PR), so thanks for merging this one!

PR #2257, opened 2018-09-22
Resto druid - Added abundance module

buimichael:
/report/q6wgn2AcByW9Cb4N/4-Mythic+Fetid+Devourer+-+Kill+(5:06)/73-Kmx
img
img

### DIFF
--- a/src/Parser/Druid/Restoration/Modules/Talents/Abundance.js
+++ b/src/Parser/Druid/Restoration/Modules/Talents/Abundance.js
@@ -1,0 +1,73 @@
+import React from 'react';
+import StatisticBox from 'Interface/Others/StatisticBox';
+import { formatPercentage } from 'common/format';
+import SpellIcon from 'common/SpellIcon';
+
+import SPELLS from 'common/SPELLS';
+import Analyzer from 'Parser/Core/Analyzer';
+
+const MS_BUFFER=200;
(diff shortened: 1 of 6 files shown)


src/Parser/Druid/Restoration/Modules/Talents/Abundance.js
```diff
@@ -0,0 +1,73 @@
...
+    const avgCritGains = this.critGains.reduce(function(a, b) { return a + b; }) / this.critGains.length;
+    const avgStacks = this.stacks.reduce(function(a, b) { return a + b; }) / this.stacks.length;
+
+    // TODO translate these values into healing/throughput.
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.ABUNDANCE_TALENT.id} />}
+        value={(
+          <React.Fragment>
+            <span style={{fontSize: 0.6 +'em'}}>
```
joshinat0r: Is it necessary to show all those infos in the box's value? Might be a good idea to move some of them in the tooltip, like just showing the average stacks and then in the tooltip the mana & crit.

Otherwise good to go.

buimichael: @joshinat0r Yes I agree. Fixed.

joshinat0r: Thanks 👍

PR #2261, opened 2018-09-23
[BrM] Updated Suggestions for ISB/BoF

emallson:
This has been popping up a lot on BrM discord because of the xpac launch. tl;dr the current implementation of the suggestion for ISB looks at uptime % (which is wrong) and the current suggestion for BoF says "uptime" but is actually "hits mitigated". This PR fixes both of those to say "hits mitigated". 

Example Log: https://www.warcraftlogs.com/reports/AbZ38fQBJYxPdGqp/#fight=6&source=104

New version:
screenshot from 2018-09-22 21-52-55
screenshot from 2018-09-22 21-52-43


emallson:
should be fixed @jjs451

PR #2266, opened 2018-09-24
[Core] Contextual prepotion

Pewtro:
Reopening to escape some nastyness happening in #2169 

Updated version of prepotion and context by @domainzero

image
image
image


domainzero:
Thanks for straightening this out @Pewtro!

PR #2268, opened 2018-09-24
[Blood Death Knight] [Bug] Eternal Rune Weapon fix

joshinat0r:
Fixes an issue where a player casted DRW before the combatlog started.


joshinat0r:
Thanks, didn't knew that `event.prepull` was a thing. Removed the boolean for that and starts the DRW cooldown when its used prepull.

PR #2279, opened 2018-09-25
Fixes #1724 Add Darkmoon Deck: Blockades healing component

### LINKED ISSUE #1724: Trinket: Darkmoon Deck: Blockades

viridis:
The items section should now correctly track the healing percentage and
HPS that Darkmoon Deck: Blockades provides when equiped.

### DIFF
--- a/src/Parser/Core/Modules/Items/BFA/Crafted/DarkmoonDeckBlockades.js
+++ b/src/Parser/Core/Modules/Items/BFA/Crafted/DarkmoonDeckBlockades.js
@@ -1,0 +1,53 @@
+import React from 'react';
+
+import ITEMS from 'common/ITEMS';
+import Analyzer from 'Parser/Core/Analyzer';
+import ItemHealingDone from 'Interface/Others/ItemHealingDone';
+
+const DARKMOON_DECK_BLOCKADES_CARDS = [
+  276204, // Ace
+  276205, // 2
+  276206, // 3
+  276207, // 4
+  276208, // 5
+  276209, // 6
+  276210, // 7
+  276211, // 8
+];
+/**
+ * Darkmoon Deck: Blockades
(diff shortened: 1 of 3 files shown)


joshinat0r:
Just as a small tip on how you could do that:

Should be possible by doing something similar to [this](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Core/Modules/Items/BFA/GildedLoaFigurine.js#L21).
You could create a new object with the buffIds of the cards and the corresponding stamina value for the base itemLevel ([this comment](https://www.wowhead.com/item=159128/darkmoon-deck-blockades#comments:id=2600174) has the values listed). Then listen to the `on_toPlayer_applybuff`-event, calculate the amount of stamina gained with `calculatePrimaryStat()` and average them out over the fights duration. You'd have to measue how long a given card was active to do that.

viridis:
Was mainly waiting for feedback and keep the iterations as small as possible as per the instructions in the contribution guide. Figured the healing component could be added as a standalone part. I'll follow this feedback and try to add the stamina information too later today.

src/Parser/Core/Modules/Items/BFA/Crafted/DarkmoonDeckBlockades.js
```diff
@@ -0,0 +1,53 @@
...
+      return;
+    }
+    this.healing += (event.amount || 0) + (event.absorbed || 0);
+  }
+
+  item() {
+    return {
+      item: ITEMS.DARKMOON_DECK_BLOCKADES,
+      result:(
+        <React.Fragment>
```
MartijnHols: React.Fragment can be omitted when it has only one child. The reason we use it is because it's necessary for multiple children since the compiler can not handle multiple JSX elements next to each other without a wrapper.

PR #2282, opened 2018-09-25
[BrM] Clarify purify usage in stagger plot

emallson:
Currently, the stagger plot uses the *right edge* of a bin for determining the y value. This can make purifies look bad when a player purifies very quickly after a hit, since the previous bin was small and we sometimes get an extra stagger tick within the current bin. This PR changes that by doing two things:

1. Bins are now plotted using the *max* value within the bin instead of the rightmost value.
2. (1) introduced an issue where if a hit occurred right after a purify, it would look like the purify did literally nothing. I fixed this by upping the resolution of the plot from 1s bins to 0.5s bins.

I also cleaned up how purifies are positioned. It should now always place them in the correct bin instead of occasionally putting them back an extra bin.

**M Fetid, before & after**
screenshot from 2018-09-25 14-33-42
screenshot from 2018-09-25 14-39-26

**M Vectis, before & after**
screenshot from 2018-09-25 14-34-09
screenshot from 2018-09-25 14-39-08

### DIFF
--- a/src/Parser/Monk/Brewmaster/CHANGELOG.js
+++ b/src/Parser/Monk/Brewmaster/CHANGELOG.js
@@ -8,6 +8,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-09-27'),
+    changes: 'Updated Stagger plot to show very quick purifies more accurately.',
+    contributors: [emallson],
+  },
+  {
     date: new Date('2018-09-22'),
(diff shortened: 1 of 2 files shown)


emallson:
actually don't merge this yet, i fucked up the resolution change

emallson:
OK disregard my previous comment :+1:

emallson:
uh oops

emallson:
screenshot from 2018-09-27 13-54-07
fixed @joshinat0r

joshinat0r:
👍

PR #2283, opened 2018-09-25
[Affliction] Summon Darkglare module

sMteX:
Added Summon Darkglare module. Little foreword - `package-lock.json` changed for me when I was working from laptop, so I copypasted it straight from `master` with the latest commit but the newline at the end stayed there and doesn't wanna go away 🤷‍♂️ 

It's a little longer so I tried to add comments where I felt it was necessary but the gist of it is this:
- manually track dots on targets, when they were cast and when they should drop off
- when Summon Darkglare is cast, record the current "state" of dots on targets, and extend the remaining time on them by 8 seconds
- after the "original" dot is expired, and we actually get to the extended part, count the bonus damage from those dots
- show some statistics for this

There's also some weird moments where things happen when they shouldn't (one example is I think this report `/report/kmJV8jPGCXFrxYDy/21-Heroic+Mythrax+-+Wipe+20+(7:15)/11-Chizu` where `removedebuff` event for Corruption on Vision of Madness add triggered `on_byPlayer_damage` event handler etc.), so that's why I had to put some safety check ifs here and there.

I originally wanted to show some kind of table showing each cast, each of the enemies affected by dots on those casts and the dots on each of those enemies (to visually see if the player casted it in the right moment), but that proved not so simple and I'm not good at CSS so I'll leave some kind of custom tab for another time.

image


MartijnHols:
There are some conflicts ([sorry](https://media.giphy.com/media/K1QnLV1caRpuw/giphy.gif)!), please merge from master.

sMteX:
Closing this for now, after I fix my stuff that broke with the renames, I'll reopen this (or open a new one)

PR #2290, opened 2018-09-27
Fixes #1724 Add Darkmoon Deck: Blockades Stamina Component

### LINKED ISSUE #1724: Trinket: Darkmoon Deck: Blockades

viridis:
The items section should now correctly track the stamina contribution
Darkmoon Deck: Blockades provides when equiped.
The information is broken down into the contribution and uptime of each
individual card, as well as the average stamina the deck provided during
the entire encounter.

### DIFF
--- a/src/parser/core/modules/items/bfa/crafted/DarkmoonDeckBlockades.js
+++ b/src/parser/core/modules/items/bfa/crafted/DarkmoonDeckBlockades.js
@@ -3,17 +3,47 @@ import React from 'react';
 import ITEMS from 'common/ITEMS';
 import Analyzer from 'parser/core/Analyzer';
 import ItemHealingDone from 'interface/others/ItemHealingDone';
+import { calculatePrimaryStat } from 'common/stats';
+import ExpandableStatisticBox from 'interface/others/ExpandableStatisticBox';
+import ItemIcon from 'common/ItemIcon';
+import STATISTIC_CATEGORY from 'interface/others/STATISTIC_CATEGORY';
+import { formatDuration, formatNumber, formatPercentage } from 'common/format';
 
-const DARKMOON_DECK_BLOCKADES_CARDS = [
-  276204, // Ace
-  276205, // 2
-  276206, // 3
-  276207, // 4
-  276208, // 5
-  276209, // 6
-  276210, // 7
-  276211, // 8
-];
+const DARKMOON_DECK_BLOCKADES_CARDS = {
+  276204: {
+    name: "Ace",
+    staminaIncrease: 72,
+  },
+  276205: {
+    name: "Two",
+    staminaIncrease: 146,
+  },
+  276206: {
+    name: "Three",
+    staminaIncrease: 218,
+  },
+  276207: {
+    name: "Four",
+    staminaIncrease: 291,
+  },
+  276208: {
+    name: "Five",
+    staminaIncrease: 364,
+  },
+  276209: {
+    name: "Six",
+    staminaIncrease: 437,
+  },
+  276210: {
+    name: "Seven",
+    staminaIncrease: 510,
+  },
+  276211: {
+    name: "Eight",
+    staminaIncrease: 583,
+  },
+};
+
 /**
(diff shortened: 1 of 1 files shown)


viridis:
Image of Darkmoon Deck: Blockades

src/Parser/Core/Modules/Items/BFA/Crafted/DarkmoonDeckBlockades.js
```diff
@@ -24,26 +51,125 @@ const DARKMOON_DECK_BLOCKADES_CARDS = [
...
+  on_toPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if (!this._isBlockadesCard(spellId)) {
+      return;
+    }
+    if (!this.cardTracker.hasOwnProperty(spellId)) {
+      return;
+    }
+    const cardsOfThisType = this.cardTracker[spellId];
+    cardsOfThisType.forEach((cardOccurance) => {
```
joshinat0r: This shouldn't be necessary, the last active card of this type is always the last pushed element (`cardsOfThisType[cardsOfThisType.length - 1]`).

joshinat0r:
Thanks for the update 👍 

Only see one issue with it, you're not handling the last card correctly. You get an `applybuff` event for the last card, but you wont get a `removebuff`. Which means that the last card application will always return 0, messing with the total stamina provided. 

You could either do that with the `on_finish` method, which is something that gets called once the fight finished parsing. But we usually try to avoid that (to avoid issues with partitially parsed reports).

Another (and better) option would be to default the cards `end`-prop to the fights duration, and update it once it gets removed.

MartijnHols:

src/parser/core/modules/items/bfa/crafted/DarkmoonDeckBlockades.js
```diff
@@ -24,30 +54,139 @@ const DARKMOON_DECK_BLOCKADES_CARDS = [
...
+
+  on_toPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if (!this._isBlockadesCard(spellId)) {
+      return;
+    }
+    if (!this.cardTracker.hasOwnProperty(spellId)) {
+      return;
+    }
+    const lastOccurrenceOfThisCard = this.cardTracker[spellId].pop();
```
joshinat0r: Is there a reason why you're poping and pushing the last item instead of just accessing it via the last index to update the `end`-prop?

src/parser/core/modules/items/bfa/crafted/DarkmoonDeckBlockades.js
```diff
@@ -24,30 +54,139 @@ const DARKMOON_DECK_BLOCKADES_CARDS = [
...
+    lastOccurrenceOfThisCard.end = event.timestamp;
+    this.cardTracker[spellId].push(lastOccurrenceOfThisCard);
+  }
+
+  _getStaminaSummary() {
```
joshinat0r: You could make this a getter when you're not passing any value into the function.

Simply changing it to `get staminaSummary()` and access it via `this.staminaSummary` like you would with a normal variable.

src/parser/core/modules/items/bfa/crafted/DarkmoonDeckBlockades.js
```diff
@@ -24,30 +53,136 @@ const DARKMOON_DECK_BLOCKADES_CARDS = [
...
+  on_toPlayer_removebuff(event) {
+    const spellId = event.ability.guid;
+    if (!this._isBlockadesCard(spellId)) {
+      return;
+    }
+    if (!this.cardTracker.hasOwnProperty(spellId)) {
+      return;
+    }
+    const lastOccurrenceOfThisCard = this.cardTracker[spellId][this.cardTracker[spellId].length -1];
+    console.log(lastOccurrenceOfThisCard);
```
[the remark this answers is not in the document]
viridis: bleh mb

PR #2293, opened 2018-09-28
[Blood Death Knight] update HB and MR-modules

joshinat0r:
Updates the Marrowrend usage module to default to a 7 stack refresh and only use 6 stacks when running Bones of the Damned.

Also removes the suggestion for running Hearth Breaker since it's now a more or less viable choice for pure defensive purpose (primarily due to the relative damage/healing loss on BD going into BfA).

### DIFF
--- a/src/parser/deathknight/blood/CHANGELOG.js
+++ b/src/parser/deathknight/blood/CHANGELOG.js
@@ -6,6 +6,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2018-09-28'),
+    changes: <React.Fragment>Updated the <SpellLink id={SPELLS.MARROWREND.id} />-usage module to account for <SpellLink id={SPELLS.BONES_OF_THE_DAMNED.id} /> and updated <SpellLink id={SPELLS.HEARTBREAKER_TALENT.id} />-Module.</React.Fragment>,
+    contributors: [joshinator],
+  },
+  {
     date: new Date('2018-08-03'),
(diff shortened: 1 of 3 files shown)


src/Parser/DeathKnight/Blood/Modules/Features/MarrowrendUsage.js
```diff
@@ -44,6 +46,7 @@ class MarrowrendUsage extends Analyzer {
 
     if(this.selectedCombatant.hasTrait(SPELLS.BONES_OF_THE_DAMNED.id)) {
       this.hasBonesOfTheDamned = true;
```
jjs451: I don't think hasBonesOfTheDamned is needed.  I don't see another usage of it other than setting it to true.

joshinat0r: It is, kind of. 
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/463776f08f4514ba4a4f8212fba9aa17965dee67/src/Parser/DeathKnight/Blood/Modules/Features/MarrowrendUsage.js#L104

I'm using it to determinate if I want to hightlight a cast in the timeline or not (in case that `this.currentBoneShieldBuffer` returns a value higher than 3 and the player has no BotD). Could technically remove that value and make it dependant on `this.refreshAtStacks`, but I'd be less readable.

MartijnHols:

jjs451:
Fair enough, an extra variable for readability in the if statement is a good enough reason.

joshinat0r:
Merge conflicts are now fixed

PR #2296, opened 2018-09-28
Add Havoc to specs that do not benefit from Arcane Torrent

Salmonhands:
Added Havoc Demon Hunter to the list of specs that do not benefit from Arcane Torrent.


jjs451:
What's the thought behind this?  Havoc gains fury from AT right?

Salmonhands:
15 Fury from that vs. 20 minimum from Demon's Bite. Even the builds that take the Demon Blades talent don't ever have time for it in their rotation as far as I can tell.

PR #2298, opened 2018-09-29
[Frost Mage] Move Frozen Orb CD reduction into whiteout module (SRP)

Dambroda:
Because a separate Whiteout trait module was created, the FO CD reduction should probably be moved into that module for SRP to remove duplicate logic checking


Dambroda:
If approved, please still do not merge until Sharrq takes a look and confirms whether or not the removed variables (cooldownReduction / baseCooldown) have usage elsewhere

MartijnHols:

PR #2308, opened 2018-09-30
[Affliction] Summon Darkglare module

sMteX:
Added Summon Darkglare module. 

It's a little longer so I tried to add comments where I felt it was necessary but the gist of it is this:
- manually track dots on targets, when they were cast and when they should drop off
- when Summon Darkglare is cast, record the current "state" of dots on targets, and extend the remaining time on them by 8 seconds
- after the "original" dot is expired, and we actually get to the extended part, count the bonus damage from those dots
- show some statistics for this

There's also some weird moments where things happen when they shouldn't (one example is I think this report `/report/kmJV8jPGCXFrxYDy/21-Heroic+Mythrax+-+Wipe+20+(7:15)/11-Chizu` where `removedebuff` event for Corruption on Vision of Madness add triggered `on_byPlayer_damage` event handler etc.), so that's why I had to put some safety check ifs here and there.

I originally wanted to show some kind of table showing each cast, each of the enemies affected by dots on those casts and the dots on each of those enemies (to visually see if the player casted it in the right moment), but that proved not so simple and I'm not good at CSS so I'll leave some kind of custom tab for another time.

image

### DIFF
--- a/src/parser/warlock/affliction/modules/features/Darkglare.js
+++ b/src/parser/warlock/affliction/modules/features/Darkglare.js
@@ -14,6 +14,8 @@ import StatisticListBoxItem from 'interface/others/StatisticListBoxItem';
 import { UNSTABLE_AFFLICTION_DEBUFF_IDS } from '../../constants';
 
 const BONUS_DURATION = 8000;
+const UNSTABLE_AFFLICTION_DURATION = 8000;
+const CREEPING_DEATH_COEFFICIENT = 0.85; // Creeping Death talent shortens duration and period of certain dots by 15%
 const DOTS_AFFECTED_BY_CREEPING_DEATH = [
   SPELLS.AGONY.id,
   SPELLS.CORRUPTION_DEBUFF.id,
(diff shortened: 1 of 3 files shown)


src/parser/warlock/affliction/modules/features/Darkglare.js
```diff
@@ -0,0 +1,256 @@
...
+
+  statistic() {
+    let totalExtendedDots = 0;
+    // for each cast, and each enemy in that cast, count the amount of dots on the enemy (disregard Corruption if player has Absolute Corruption)
+    Object.values(this.casts).forEach(cast => {
+      Object.keys(cast).filter(key => key !== 'timestamp').forEach(target => {
+        if (this._hasAC) {
+          totalExtendedDots += cast[target].dots.filter(id => id !== SPELLS.CORRUPTION_DEBUFF.id).length;
+        }
+        else {
```
joshinat0r: We prefer the `} else {` way

src/parser/warlock/affliction/modules/features/Darkglare.js
```diff
@@ -0,0 +1,256 @@
...
+
+  constructor(...args) {
+    super(...args);
+    // if player has Absolute Corruption, disregard the Corruption duration (it's permanent debuff then)
+    this._hasAC = this.selectedCombatant.hasTalent(SPELLS.ABSOLUTE_CORRUPTION_TALENT.id);
+    if (this._hasAC) {
+      delete this._dotDurations[SPELLS.CORRUPTION_DEBUFF.id];
+    }
+    UNSTABLE_AFFLICTION_DEBUFF_IDS.forEach(id => {
+      this._dotDurations[id] = 8000;
```
joshinat0r: This is the affliction duration if I read this code correct and not the dark glare's bonus duration.
Would make it cleaner if you put that in a const for better readability.

src/parser/warlock/affliction/modules/features/Darkglare.js
```diff
@@ -0,0 +1,256 @@
...
+    });
+    if (this.selectedCombatant.hasTalent(SPELLS.CREEPING_DEATH_TALENT.id)) {
+      // Creeping Death talent shortens the period and duration of dots by 15%
+      DOTS_AFFECTED_BY_CREEPING_DEATH.forEach(id => {
+          this._dotDurations[id] *= 0.85;
```
joshinat0r: You could put that into a const aswell, although its understandable what it does.

PR #2327, opened 2018-09-30
Update Rogue exampleReport urls

viridis:


Pewtro:
Done in https://github.com/WoWAnalyzer/WoWAnalyzer/pull/2317

PR #2331, opened 2018-10-01
[Havok DH] Removed fel eruption suggestion

yajinni:
The damage increase to it was removed in BFA so no longer worth using on CD
image


yajinni:
Yes my bad, fixed.

PR #2332, opened 2018-10-01
[Core] update prepots

Pewtro:
Fixes the error where some specs would get a wrong potion suggested as mentioned by Sharrq on discord. 
Specifically it was about frost mages and shadow priests, here is an example from frost mages. 
image

I made the code more readable imo aswell. 

Just to show that even if some specs are in multiple places (both bursting and agi for example), it'll still be accurate: 
image

### DIFF
--- a/src/parser/core/modules/items/PrePotion.js
+++ b/src/parser/core/modules/items/PrePotion.js
@@ -190,7 +190,7 @@ class PrePotion extends Analyzer {
       );
     when(this.secondPotionSuggestionThresholds)
       .addSuggestion((suggest) => {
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/items/PrePotion.js
```diff
@@ -137,24 +156,26 @@ class PrePotion extends Analyzer {
   }
 
   potionAdjuster(specID) {
-    this.alternatePotion = STR_SPECS.includes(specID) ? ITEMS.BATTLE_POTION_OF_STRENGTH.id : ITEMS.BATTLE_POTION_OF_AGILITY.id;
-    if (AGI_SPECS.includes(specID)) {
+    this.alternatePotion = STR_SPECS.includes(specID) ? ITEMS.BATTLE_POTION_OF_STRENGTH.id : AGI_SPECS.includes(specID) ? ITEMS.BATTLE_POTION_OF_AGILITY.id : ITEMS.BATTLE_POTION_OF_INTELLECT.id;
```
emallson: eww nested ternaries

Pewtro: Yeah, best option I could think of in the moment 😅

PR #2336, opened 2018-10-01
[HPriest] Holy word reduction styling

TJMoats:
It currently looks like this: https://i.imgur.com/KFQy0Dn.png
It will now look like this: https://i.imgur.com/1L0MyIh.png


yajinni:
Ah i was gonna get around to fixing that cuz i hated the icons only look. Nice !

Pewtro:
Do you not want the spell and base to be bolded, like it was before?

TJMoats:
It makes no difference to me. The <th> tag has a font-align: right on by default, and for some reason it wouldn't override it when I added the text-align class to it. It was easier to make it a td than to hunt down the css issue.

PR #2339, opened 2018-10-01
DH Havoc - Change fel eruption to utility

Salmonhands:
Fel Eruption is a utility spell as of 8.0.1


Salmonhands:
Whoops, should have checked that first!

PR #2344, opened 2018-10-02
Added Enchant masterful navigation

Gamewalker:


Pewtro:
While I can somewhat imagine what this looks like, a screenshot and a log link would be appreciated

Pewtro:
That's not what I'm seeing in quick nav at all, I've applied a small fix on my local version removing the possibility of a 5 stack, which makes it look this 
image

as opposed to 
image


What you're seeing must be some other code error, not related to how quick nav is handled.

Pewtro:
Issue you raised is fixed in #2352

Gamewalker:
should be obsolete with #2355

PR #2345, opened 2018-10-02
[HPriest] Mana Usage Chart

TJMoats:
I refactored the druid mana tracker to be usable by other classes. I made the priest version an expandable statistic box as it's a bit big otherwise.

What it looks like: https://i.imgur.com/FK0o802.png

I'd like to abstract one more level out and have a generic "Pie Chart" card, but I will save that for another PR.

### DIFF
--- a/src/common/ManaUsage.js
+++ b/src/common/ManaUsage.js
@@ -1,0 +1,152 @@
+import React from 'react';
+import { Doughnut as DoughnutChart } from 'react-chartjs-2';
+
+import SpellLink from 'common/SpellLink';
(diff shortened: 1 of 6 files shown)


Pewtro:
This doesn't seem useful information. There are too many bars to properly digest the information

TJMoats:
Are there other classes that have a mana usage tab that I can borrow from?

Pewtro:
Havoc DH has a fury usage tab you can look into, that's one I know for sure exists

TJMoats:
I'll emulate that style of resource tracking. Thanks!

PR #2346, opened 2018-10-02
Added azerite trait tidal surge

Gamewalker:
Dunno where the numbers in TraitIdMap.js comes from, it did not work with 280402 so I had to change it to 280404


Pewtro:
A log link and an image here aswell please 😄

Gamewalker:
tidal-surge

here we go

PR #2349, opened 2018-10-02
[Core] Be Prepared BFA edition

Pewtro:


Pewtro:
Some paths still capitalised 😡

PR #2354, opened 2018-10-02
[Hunter] Various changes and Bird of Prey module

Pewtro:
Adjusts dance of death trait to be 20% stronger as of the hotfix on the sept 26th.
Properly hides the focuscapped module only if you're a MM hunter with Steady Aim (some logs are a different spec, but still use MM azerite traits)
Moves some talents into the talent section.
Adjusts some cast efficiencies to more accurately represent what we're seeing in the raid & hides Raptor Strike when Mongoose Bite is talented
Adds Coordinated Assault contribution to the statisticlistbox and moves Bird of Prey handling into it's own module. 
Updates my Mongoose Bite MS_BUFFER to catch the last pesky few cases of the buggy interaction. 
image
image

Also fixes the <TalentStatisticBox> to not insert things into Azerite powers by default. 

This PR will be a part of a series of updates to SV that will allow me to implement the new checklist afterwards, since then I'm nearing a place where I think most of the functionality is there. 
Then the two other specs will receive the same amount of polishing and fine tuning prior to their checklist implementations.


src/parser/hunter/survival/modules/talents/BirdOfPrey.js
```diff
@@ -0,0 +1,118 @@
...
+
+/** Bird of Prey
+ * Attacking your pet's target with Mongoose Bite, Raptor Strike, Butchery or Carve extends the duration of Coordinated Assault by  1.5 sec.
+ *
+ * Example log: https://www.warcraftlogs.com/reports/pNJbYdLrMW2ynKGa#fight=3&type=damage-done&source=16&translate=true
+ */
+
+class BirdOfPrey extends Analyzer {
+
+  petTarget;
```
MartijnHols: Please do give props default values, even if it's just `null`.

PR #2356, opened 2018-10-02
Resource breakdown add key property to <tr>

Salmonhands:
Deals with the warning message presented in console when the resource breakdown tab is selected. Warning doesn't seem like a huge deal currently because the order of the list isn't going to change, but silencing it now also deals with any future changes (like filtering or ordering).


MartijnHols (approved):
Thanks!

PR #2357, opened 2018-10-02
[Destruction] Wasted Conflagrate 

sMteX:
Although I'm not exactly a fan of this, as it's duplicated from what is shown in the Abilities tab, apparently people are lazy to look for that, and from the viewpoint of analyzing logs, it's important to make it more visible, so here it is.

Example: https://www.warcraftlogs.com/reports/PdLX2rva3q8fN1Zw/#fight=29&source=23

image


sMteX:
A little shame, but I agree. I'll close this for now, will probably add once more when implementing Checklist. People need to be more mindful of the features we provide and look through other tabs 🤷‍♂️

PR #2367, opened 2018-10-03
[Core] Fix negative formatDuration; [Mage] Fix Time Warp GCD info

Dambroda:
formatDuration now will properly handle negative durations. This has no effect on how positive durations on displayed. Added to support potentially adding pre-pull features later on.

Also fixing time warp GCDs

### DIFF
--- a/src/parser/mage/fire/modules/features/Abilities.js
+++ b/src/parser/mage/fire/modules/features/Abilities.js
@@ -110,6 +110,15 @@ class Abilities extends CoreAbilities {
 
       // Cooldowns
       {
+        spell: SPELLS.TIME_WARP,
+        category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+        gcd: null,
+        cooldown: 300,
+        castEfficiency: {
+          disabled: true,
+        },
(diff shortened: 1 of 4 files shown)


src/parser/mage/fire/modules/features/Abilities.js
```diff
@@ -109,6 +109,15 @@ class Abilities extends CoreAbilities {
...
+      {
+        spell: SPELLS.TIME_WARP,
+        category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
+        gcd: null,
+        cooldown: 300,
+        castEfficiency: {
+          disabled: true,
+        },
+      },
       {
```
Pewtro:
```JS        
castEfficiency: {
          disabled: true,
        },
``` 
is unnecessary code that you can just remove from all 3 abilities.js

Dambroda: I think it is meant to do that because time warp is a spell that you do not care about the efficiency of at all, so having it say "can be improved" because you're not the designated lust caster or something like that would be incorrect

PR #2368, opened 2018-10-03
[Mage] Pre-pull cooldown normalizer

Dambroda:
This normalizer attempts to solve a problem throughout the WoWAnalyzer castefficiency module. While the castefficiency module can (with the ApplyBuff normalizer checking for event.prepull) properly deduce cooldown percentages, it does not accurately count the actual casts.

For example, the analyzer can tell I have _had_ the Icy Veins buff 3 times, but it can't tell that I have _cast the spell_ 3 times:
image
image

With this PR, cast events are fabricated pre-pull (with pre-pull timestamps, to avoid the "you cast spell X while still on the GCD from Y!" error) which will be accounted for castefficiency percentages AND counts.
Example:
image

image

The goal of this pull request is to test the concept for mages first, and then if it works and gets optimized it can be made global and all specs can remove their prepull buff checks because this will handle it automatically.

### DIFF
--- a/src/common/format.js
+++ b/src/common/format.js
@@ -36,12 +36,14 @@ export function formatPercentage(percentage, precision = 2) {
  * Ex: 317.3 => 5:17
  */
 export function formatDuration(duration, precision = 0) {
+  const neg = duration < 0 ? '-' : '';
+  duration = Math.abs(duration);
   const minutes = Math.floor(duration / 60);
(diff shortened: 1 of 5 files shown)


Gebuz:
This is really nice, would be awesome if you could turn it into a core module so we can extend it for other specs!

PR #2370, opened 2018-10-04
[HPriest] Fort buff threshold fix

TJMoats:
The current Power Word: Fort thresholds are a bit hard to hit:

https://puu.sh/BFEsd/bbf3bd6e47.png

This pr should fix that.


emallson:
while i understand allowing a bit lower threshold (say 95%) to account for deaths+brezzes, 50% for average seems way too low.

PR #2394, opened 2018-10-06
[Frost Mage] Fix Winters Reach module

Dambroda:
https://wowanalyzer.com/report/hRNnK37cCfpgBzD2/47-Heroic+G'huun+-+Kill+(6:24)/15-Remyhadley

He reported an issue (winters reach 0% utilization), claiming he used all of his procs. Looking at the events it looks like he was right. This PR should greatly simplify the logic behind this module and resolve the issue of it not counting procs. I also made some convention changes such as using this.debug and capitalization changes to fit with other modules.

### DIFF
--- a/src/parser/mage/frost/modules/traits/WintersReach.js
+++ b/src/parser/mage/frost/modules/traits/WintersReach.js
@@ -5,21 +5,14 @@ import SpellLink from 'common/SpellLink';
 import SpellIcon from 'common/SpellIcon';
 import Analyzer from 'parser/core/Analyzer';
 import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
-import { formatMilliseconds, formatNumber, formatPercentage } from 'common/format';
+import { formatNumber, formatPercentage } from 'common/format';
 
-const CAST_BUFFER = 250;
 const FIGHT_END_BUFFER = 7500;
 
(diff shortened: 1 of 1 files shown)


Sharrq:
I don't know the answer to this, but just to make sure ...

Is it possible to complete a cast of frostbolt and start hard casting flurry before brain freeze procs and therefore have brain freeze at the end of the hard cast?

Dambroda:
I'll go and double check, but I believe it shouldn't be possible. From my recent examinations, the only time brain freeze actually applies _after_ the frostbolt timestamp (i.e., you'd be able to start a hard-cast flurry) is if it was a buffrefresh event. Even in that case, because it's a refresh that means the player had a brain freeze anyway so you wouldn't have been able to start the hardcast.

Sharrq:
I would assume it's not possible cause otherwise casting EB into Flurry would result in the start of a hard cast.

But that's the only issue I see with the simplification so I thought I'd double check

Dambroda:
Yeah just confirmed on a dummy that if you're queuing up flurry (spamming the cast button) after frostbolt and the frostbolt gives you an unexpected proc, it will immediately use the proc and will not start a hard cast. Unless there's some strange stuff going on with how the event timestamps play out, it shouldn't be an issue

PR #2400, opened 2018-10-07
[Core][Azerite] Fixed scaling for split stat traits

emallson:
**Current**
screenshot from 2018-10-06 22-43-49
**New**
screenshot from 2018-10-06 22-47-52

From: https://www.warcraftlogs.com/reports/xR1wHKP3BgNaYFmL/#fight=3&source=19

Some traits (like Gemhide, my go-to example) give multiple different secondaries. These secondaries can scale differently in *exactly one way.* If one of the effects has type 189 (Modify Rating), it is scaled as a secondary regardless of the trait's scaling type. This PR adjusts the azerite scaling code to correctly do this.

This PR also updates all the scaling coefficients, which breaks one of the Disc Priest tests. I have commented out the offending test cases until the maintainer can address this.


emallson:
Looks like @rubensayshi wrote the disc tests that broke due to new scaling coefficients. Just an FYI

PR #2420, opened 2018-10-10
[Ret] Judgment module fix and some other minor things

Juko8:
- Added custom GCD for Wake of Ashes
- Updated use of position prop instead of deprecated statisticOrder
- Fixed judgment consumptions being counted for every tick when a divine storm consumed a judgment

old result: 
billede
new result:
billede

Planning to turn the judgment statistic into a donut so less info is hidden in the tooltip.

### DIFF
--- a/src/parser/paladin/retribution/modules/core/GlobalCooldown.js
+++ b/src/parser/paladin/retribution/modules/core/GlobalCooldown.js
@@ -1,0 +1,30 @@
+import CoreGlobalCooldown from 'parser/shared/modules/GlobalCooldown';
+import SPELLS from 'common/SPELLS';
+import Haste from './Haste';
+import Abilities from '../Abilities';
+
+const MIN_GCD= 750;
+
+/**
+ * The Global Cooldown from Wake of Ashes is currently double dipping haste
+ */ 
+
+class GlobalCooldown extends CoreGlobalCooldown {
+  static dependencies = {
+    abilities: Abilities,
+    haste: Haste,
+  };
+
+  getGlobalCooldownDuration(spellId) {
+    const gcd = super.getGlobalCooldownDuration(spellId);
+    if (!gcd) {
+      return 0;
+    }
+    if (spellId && spellId === SPELLS.WAKE_OF_ASHES_TALENT.id) {
+      return Math.max(gcd / (1 + this.haste.current));
+    }
+    return Math.max(MIN_GCD, gcd);
+  }
+}
+
+export default GlobalCooldown;
--- a/src/parser/paladin/retribution/modules/core/Judgment.js
+++ b/src/parser/paladin/retribution/modules/core/Judgment.js
@@ -84,6 +84,7 @@ class Judgment extends Analyzer {
     const justicarsVengeanceText = this.selectedCombatant.hasTalent(SPELLS.JUSTICARS_VENGEANCE_TALENT.id) ? `<br>Justicars Vengeance consumptions: ${this.justicarsVengeanceConsumptions}` : ``; 
     return (
       <StatisticBox
+        position={STATISTIC_ORDER.CORE(6)}
         icon={<SpellIcon id={SPELLS.JUDGMENT_DEBUFF.id} />}
         value={`${formatPercentage(this.percentageJudgmentsConsumed)}%`}
         label="Judgments Consumed"
@@ -95,7 +96,6 @@ class Judgment extends Analyzer {
       />
     );
   }
-  statisticOrder = STATISTIC_ORDER.CORE(6);
 }
 
 export default Judgment;
--- a/src/parser/paladin/retribution/modules/core/Judgment.js
+++ b/src/parser/paladin/retribution/modules/core/Judgment.js
@@ -9,6 +9,10 @@ import SpellLink from 'common/SpellLink';
 import { formatPercentage } from 'common/format';
 import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
 
+const JUDGMENT_DURATION = 15000;
+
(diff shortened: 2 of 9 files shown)


src/parser/paladin/retribution/modules/core/GlobalCooldown.js
```diff
@@ -0,0 +1,30 @@
...
+    haste: Haste,
+  };
+
+  getGlobalCooldownDuration(spellId) {
+    const gcd = super.getGlobalCooldownDuration(spellId);
+    if (!gcd) {
+      return 0;
+    }
+    if (spellId && spellId === SPELLS.WAKE_OF_ASHES_TALENT.id) {
+      return Math.max(gcd / (1 + this.haste.current));
```
MartijnHols: This Math.max does nothing?

MartijnHols: Please leave a comment to explain this change

src/parser/paladin/retribution/modules/core/Judgment.js
```diff
@@ -17,38 +21,73 @@ class Judgment extends Analyzer {
...
       return;
     }
+    // The event doesn't specify instance on the first target of that type
+    let targetInstance = event.targetInstance;
+    if (targetInstance === undefined) {
+      targetInstance = 1;
+    }
     this.judgmentsApplied += 1;
+    const judgmentTarget = { targetID: event.targetID, targetInstance: targetInstance, timestamp: event.timestamp };
+    this.judgmentTargets.push(judgmentTarget);
```
MartijnHols: Why are you tracking this manually now? Is getBuff tracking bugged?

src/parser/paladin/retribution/modules/core/Judgment.js
```diff
@@ -17,38 +21,73 @@ class Judgment extends Analyzer {
...
-        break;
-      default:
-        break;
+    let targetInstance = event.targetInstance;
+    if (targetInstance === undefined) {
+      targetInstance = 1;
     }
+    this.judgmentTargets.forEach((judgmentTarget, index) => {
+      // remove expired judgments
+      const expirationTimestamp = judgmentTarget.timestamp + JUDGMENT_DURATION;
```
MartijnHols: This should trigger a removebuff, relying on that should be more reliable. I think this could break with lag.

src/parser/paladin/retribution/modules/core/Judgment.js
```diff
@@ -17,38 +21,73 @@ class Judgment extends Analyzer {
...
+      if (expirationTimestamp < event.timestamp) {
+        this.judgmentTargets.splice(index, 1);
```
MartijnHols: Shouldn’t this return then?

src/parser/paladin/retribution/modules/core/GlobalCooldown.js
Juko8: I'm assuming you mean the Math.max in the final return statement and not the nested Math.max that is shown here?

src/parser/paladin/retribution/modules/core/Judgment.js
Juko8: getBuff was returning true for all enemies (or at least those with same targetID) when only one had the debuff

src/parser/paladin/retribution/modules/core/Judgment.js
Juko8: Just a blank `return;` or the spliced judgmentTarget?
It's important that the array is looked fully through as it is possible to consume multiple judgments with 1 cast.

src/parser/paladin/retribution/modules/core/Judgment.js
Juko8: removebuff is triggered correctly already. Whether the judgmentTarget object is removed here or in a removebuff event trigger is entirely irrelevant, it just has to be done before it's counted as consumed.

src/parser/paladin/retribution/modules/core/GlobalCooldown.js
MartijnHols: no

MartijnHols: This one has only 1 parameter, so it does nothing?

src/parser/paladin/retribution/modules/core/Judgment.js
MartijnHols: It is relevant though as `removebuff` will always match exactly how it happened in-game, while this is just a guess at when that might be.

src/parser/paladin/retribution/modules/core/Judgment.js
MartijnHols: blank return

src/parser/paladin/retribution/modules/core/GlobalCooldown.js
Juko8: ah, my bad, it needs MIN_GCD as second parameter

src/parser/paladin/retribution/modules/core/Judgment.js
MartijnHols: In that case please leave a comment in the code explaining why you didn't use `removebuff`

src/parser/paladin/retribution/modules/core/Judgment.js
MartijnHols: But if you hit someone after it was removed, the spell wouldn't have been buffed

sMteX:
This has been opened for a month already, how does it look now @Juko8 @MartijnHols ?

Juko8:
I apologize for all the fuzz with the workaround when i should have seen the proper solution was so much easier with EnemyInstances. Not much reviewing is required now, just check the 2 lines changed in the latest commit.

PR #2422, opened 2018-10-10
[Feral] Wild Fleshrending azerite trait

Anatta336:
Adds tracking for the Wild Fleshrending azerite trait for Feral druids.

"Shred deals X additional damage, and Swipe (or Brutal Slash) deals Y additional damage, to enemies suffering from your Thrash."

The bonus damage is only active if the player has their Thrash debuff on the target, so that's checked for and suggestions generated if it's missing too often.

### Calculating damage
This uses a (I think) new way to calculate bonus damage from azerite traits that add their damage directly into an existing ability. Through in-game testing I've found the total damage of Shred and Swipe follows this formula:
`damageDone = (attackPower * ABILITY_COEFFICIENT + traitBonus) * modifier`

`attackPower` is attack power after weapon DPS is accounted for, which is the value given by the `attackPower` property available on combat log cast events (at least it is for Feral druids.)
The `modifier` term includes a *lot* of stuff. Damage multipliers from player buffs, versatility, Mystic Touch, enemy armor, special boss mechanics like MOTHER's 3rd chamber, and even the critical hit multiplier. The key thing is we don't have to know any of that to find an accurate value for it.

The log gives us `damageDone` and `attackPower`, `ABILITY_COEFFICIENT` is a known constant, and the `calculateAzeriteEffects` function tells us the `traitBonus`. With that information we can calculate a value of `modifier` for each attack. Multiply `traitBonus` by that to find how much damage the trait contributed to that attack.

For this to work the trait's bonus damage must get multiplied by all the same modifiers that the ability's base damage is. This assumption holds up in my testing of Wild Fleshrending, and the numbers it ends up producing are where I'd expect them to be.

### Example logs
3x Wild Fleshrending on MOTHER, including 3rd chamber's damage bonus. This build was designed to maximise Wild Fleshrending's benefit during that burst phase, so the high damage % is to be expected:
`report/xCr6V8fXTMkRq1Bm/7-Normal+MOTHER+-+Kill+(3:13)/5-Anatta/timeline`

1x Wild Fleshrending in a M+ showing it working with Brutal Slash:
`report/mKPqfdtT81RzJHkB/5-+Atal'Dazar+-+Kill+(40:15)/6-Hoofhartd/abilities`

fleshrending01
fleshrending02
fleshrending03
fleshrending04


src/parser/druid/feral/modules/azeritetraits/WildFleshrending.js
```diff
@@ -0,0 +1,215 @@
...
+     * The modifier includes things like shred's 20% buff against bleeding targets, Tiger's Fury, Savage Roar, the Feral Druid aura that's used to adjust DPS balance, and even the damage increase from a critical strike. It also includes things that are difficult to see from our logs such as Mystic Touch or encounter-specific effects like MOTHER taking extra damage in the 3rd chamber. The effect of enemy armor is also included within the modifier.
+     * We know the damageDone, attackPower, ABILITY_COEFFICIENT, and traitBonus which means we can calculate the modifier's value for this attack. With the modifier known we can multiply that by traitBonus to find how much damage it actually contributed to this attack.
+     * 
+     * Assumptions: There are no effects in play that adjust damage by a set number rather than by multiplying, apart from those that appear in the log as absorbed damage. Everything that modifies the ability's base damage also modifies the trait's bonus damage in the same way. I've yet to find anything that violates these assumptions.
+     */
+    
+    let coefficient = 1.0;
+    switch (event.ability.guid) {
```
emallson: would this be better by attaching coefficients to the spellbook and then doing something like `this.abilities.getAbility(event.ability.guid).coefficient`?

Dambroda: I don't know if this works a lot differently than the mage trait Whiteout, but we avoided having to use any spell coefficients by doing something similar, but slightly different logic:

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/mage/frost/modules/traits/Whiteout.js#L79

It makes one key assumption; the traits damage (from calculateAzeriteEffects) is only changed by four things: versatility, spec aura, critting, and external modifiers (like boss phases, chaos brand, etc).

Dambroda: One other note on the PR - should there really be a suggestion for this trait? It feels like Thrash efficiency is a problem on its own, so shouldn't that be its own module (if it isnt one already)? I feel like you'd just have
"You didnt keep thrash up on the target"
"You lost fleshrending damage because you didn't keep thrash up on the target"

Which isn't really two suggestions

Anatta336: **Coefficients**
Putting the coefficients into the spellbook instead of the spec's constants is a good idea. As you show, that'll allow them to be accessed more directly and it'll keep spell-specific information in one place. (DoT duration information is currently stored in Feral's `constants.js` which could benefit from being moved too.)

**Whiteout-style Calculation**
I think that method would work here too. Wild Fleshrending's damage is modified by quite a lot more personal effects than Whiteout's though, so it would be a more complex implementation. 

As an experiment I've implemented that "mage-style" approach too, and they mostly give the exact same result. I'll write more comparing them in the main discussion thread of this PR.

**Thrash Uptime Suggestion**
Wild Fleshrending does somewhat change the Feral rotation. Thrash is designed as an AoE ability and on single target the resources used to keep the bleed up could be used to do almost exactly the same damage directly (varies slightly depending on the player's secondary stats). In some conditions it's a damage loss to keep Thrash active as the direct damage alternatives become temporarily more efficient, so a simple "aim for full uptime" check wouldn't suit Thrash.

Even with Wild Fleshrending traits there can be times in the rotation where it's beneficial to let Thrash's DoT drop off then reapply it when you're ready to spam Shred/Swipe. So I think it's best to check for how well the Wild Fleshrending effect is being applied rather than looking directly at uptime.

Dambroda: Meh, with so many modifiers (now that I look at yours, I'm pretty sure I missed one or two as well) it's hard to say if that's even easier to maintain than the coefficients. I'll let you and others decide which is better when it comes to actually maintaining it The real issue here is not having this data mined from the ability database. Hence the bounty on that... So much could be made simpler if we grab all of this data and update it automatically.

PR #2430, opened 2018-10-12
[Arcane Mage] Bug Fixes & Updates

### LINKED ISSUE #2425: [Arcane Mage] Rethink/Redo AP Pre Req Checks
Got some feedback from the Mage Discord, the Arcane Power Module checks to make sure you meet the following pre-reqs before you cast Arcane Power.

- You had 4 Arcane Charges
- You had more than 40% Mana (unless you have the Overpowered Talent)
- You cast Rune of Power immediately before AP (if you have RoP talented)

The problem, is it only tells you that you cast AP without meeting the pre reqs x times ... it doesnt say WHICH pre-reqs you failed on. 

Additionally, got some feedback that a 50% utilization score for missing one pre req out of 2 casts is a bit harsh. Might count pre reqs separately so its like 5/6 for failing one prereq as opposed to 1/2 for failing one cast ... though would need to adjust the % based on the number of pre-reqs actually checked. i.e. if you have OP and not RoP then you are only checking 1 thing, if you have OP and RoP you are checking 2 things, and if you dont have OP and have RoP you are checking 3.

Finally, should take another look at the acceptable delay on RoP. It is set at 200ms because the thought is that AP lasts 10 seconds, RoP lasts 10 seconds ... they should be cast at the same time always. But one player did RoP > Potion > AP and it dinged him for it cause there was an 800ms delay. Ill have to talk to the Mage Discord to find out if a strict delay is warranted or if it should be loosened up a bit.

Sharrq:
- Fixed Arcane Charge Normalizer so that it will put the energize events after the cast that generated them, but before the Arcane Barrage cast if they were cast on the same time stamp.
- Fixed bug in Rule of Threes where it would count Arcane Barrage as being cast with the Rule of Threes active if Rule of Threes was removed on the same timestamp as the Arcane Barrage cast (even if the Remove Buff event happened first)
- Modified Arcane Power module to count each pre-req separately and use only failed checks for determining utilization instead of failed casts.
- Updated Arcane Power tooltip and suggestion to show which checks were failed

For the AP Change, the difference here is that before these changes, if you cast Arcane Power twice and you missed one check on one of the casts, your utilization is 50%. After the changes, if it checks 3 things per cast and you miss one check on one cast, your utilization is 83% (5/6). Additionally, the players in mage discord were complaining that there is no way to tell which checks you failed aside from pouring over events and timelines. Im not completely sold on this way of displaying it, but it works until i can come up with a better solution.

Fixes #2425

### DIFF
--- a/src/parser/mage/arcane/modules/features/ArcanePower.js
+++ b/src/parser/mage/arcane/modules/features/ArcanePower.js
@@ -171,15 +171,17 @@ class ArcanePower extends Analyzer {
   }
 
   get requiredChecks() {
-    if (this.hasOverpowered && !this.hasRuneOfPower) {
-      return 1;
-    } else if (!this.hasOverpowered && !this.hasRuneOfPower) {
-      return 2;
-    } else if (this.hasOverpowered && this.hasRuneOfPower) {
-      return 3;
-    } else {
-      return 4;
+    let checks = 1;
+    if (!this.hasOverpowered) {
+      //Also checks mana level if you dont have Overpowered talented
+      checks += 1;
     }
+    if (this.hasRuneOfPower) {
+      //Also checks to see if RoP was delayed or missing if RoP is talented
+      checks += 1;
+    }
+    return checks;
+
   }
 
   get failedChecks() {
--- a/src/parser/mage/arcane/modules/features/ArcanePower.js
+++ b/src/parser/mage/arcane/modules/features/ArcanePower.js
@@ -69,6 +69,10 @@ class ArcanePower extends Analyzer {
       const currentManaPercent = event.classResources[0].amount / event.classResources[0].max;
       this.arcanePowerCasted = true;
 
+      if (this.arcaneChargeTracker.charges < 4 || (this.hasRuneOfPower && event.timestamp - this.runeTimestamp > 200) || (!this.hasOverpowered && currentManaPercent < MANA_THRESHOLD)) {
+        this.badUses += 1;
+      }
+
       if (this.selectedCombatant.hasBuff(SPELLS.ARCANE_POWER.id)) {
         this.buffEndTimestamp = event.timestamp + 10000;
         debug && this.log('Arcane Power Cast During Arcane Power Proc');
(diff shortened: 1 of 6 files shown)


src/parser/mage/arcane/modules/features/ArcanePower.js
```diff
@@ -212,7 +245,7 @@ class ArcanePower extends Analyzer {
...
-        return suggest(<>You cast <SpellLink id={SPELLS.ARCANE_POWER.id} /> {this.badUses} times without meeting the pre-requisites. Before casting Arcane Power you should ensure that you have 4 Arcane Charges{this.hasOverpowered ? '' : ' and at least 40% mana to avoid going OOM during Arcane Power'}. {this.hasRuneOfPower ? 'Additionally, you should ensure that you have at least 1 charge of Rune of Power, and you should cast it immediately before casting Arcane Power.' : ''}</>)
+        return suggest(<>You cast <SpellLink id={SPELLS.ARCANE_POWER.id} /> with less than 4 Arcane Charges {this.lowChargesCast} times and with low mana {this.lowManaCast} times. {this.hasRuneOfPower ? 'Additionally, you cast Arcane Power without Rune of Power ' + this.noRuneCast + ' times and delayed Arcane Power after Rune of Power ' + this.delayedRuneCast + ' times.' : ''} It is very important that you do not cast Arcane Power unless you have 4 Arcane Charges, have over 40% Mana (Unless talented into <SpellLink id={SPELLS.OVERPOWERED_TALENT.id} />), and <SpellLink id={SPELLS.RUNE_OF_POWER_TALENT.id} /> was cast immediately before Arcane Power (if talented into Rune of Power).</>)
```
Dambroda: Maybe something similar to the statistic box should be put into the suggestion box? Like:

You cast Arcane Power sub-optimally X of Y times.
- You should have 4 Arcane Charges when you cast Arcane Power. X casts without 4 charges.
- You should cast Rune of Power directly before Arcane Power. X casts without an optimal Rune of Power.
- You should have enough mana (~X% without Overpowered) to not run OOM during the Arcane Power. X casts with low mana.

src/parser/mage/arcane/modules/features/ArcanePower.js
```diff
@@ -153,8 +170,24 @@ class ArcanePower extends Analyzer {
     return 0;
   }
 
+  get requiredChecks() {
```
Dambroda: Could you add some comments here as to how you are getting the numbers? Or, change the conditional logic to make more sense as you read the code. Something like:
```javascript
let requiredChecks = 4;
if(this.hasOverpowered) {
  requiredChecks -= 1; // Don't check for low mana if overpowered is talented
}
if(!this.hasRuneOfPower) {
  requiredChecks -= 2; // Don't check for RoP delayed or missing if not talented
}
// etc...
```

src/parser/mage/arcane/modules/features/RuleOfThrees.js
```diff
@@ -24,7 +24,7 @@ class RuleOfThrees extends Analyzer {
 		if (spellId !== SPELLS.ARCANE_BARRAGE.id) {
 			return;
 		}
-		if (this.selectedCombatant.hasBuff(SPELLS.RULE_OF_THREES_BUFF.id)) {
+		if (this.selectedCombatant.hasBuff(SPELLS.RULE_OF_THREES_BUFF.id,event.timestamp + 1)) {
```
Dambroda: This works for now, but just as a reminder we can fix this with a normalizer as well I think

Sharrq: the order of the events isnt the problem. If it was, I would have just made a normalizer to rearrange them.

As you can see here, the Rule of Threes buff was removed BEFORE the Arcane Barrage cast, and yet it still said he had Rule of Threes when Arcane Barrage was cast
https://media.discordapp.net/attachments/499995593465135104/500054178962604053/unknown.png

src/parser/mage/arcane/modules/features/ArcanePower.js
Sharrq: i wanted to do something like that, but couldnt figure out how to get an ordered list in there .... ill have to play with it some more.

src/parser/mage/arcane/modules/features/RuleOfThrees.js
Dambroda: Wouldnt the normalizer work if you didnt use the timestamp?

Sharrq:
Changed the suggestion and fixed the other items

Sharrq:
Changed the suggest

Sharrq:
Changed the suggestion again. I would like to keep all the info that is in there currently as i believe it is all relevant 

image

Sharrq:
changed the suggestion

image

PR #2431, opened 2018-10-12
[Restoration Shaman] Wellspring efficiency

niseko:
- Added Wellspring efficiency statistics / suggestions / checklist entries as that talent is used a lot now. Wellspring requires hitting a minimum amount of injured people for it to do its full healing, but everything above that threshold doesn't matter.
- Moved Ancestral Vigor into the talents section
- Adjusted Healing Surge mana cost to fit the recent hotfixes

image
image
image


Gebuz:
If you want to avoid having such a long list with all the casts you could instead (or do both) mark the bad casts on the time line with this:
```
if (isBadCast) {
  event.meta = event.meta || {};
  event.meta.isInefficientCast = true;
  event.meta.inefficientCastReason = 'This Wellspring cast hit less than 6 targets.';
}
```

niseko:
Changed the statistic to use efficiency as well for consistency, but someone wanted a list so they're getting a list 📃 

image

Also includes timeline highlighting 
image

MartijnHols:
How about adding labels next to the numbers in the list? And does this really need the microsecond prceision? E.g.

17th 07:31 5 hits

niseko:
Yeah i was a bit too lazy on the formatting, how about this change: 
image

PR #2434, opened 2018-10-12
[Arms Warrior] A Little Bit of Cleanup

Sharrq:
- Fixed some spell IDs that were not correct (according to the console alerts about missing spells from the spellbook
- Updated Config/Contributors
- Fixed some Ability information

### DIFF
--- a/src/parser/warrior/arms/modules/Abilities.js
+++ b/src/parser/warrior/arms/modules/Abilities.js
@@ -82,7 +82,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.CLEAVE_TALENT,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL_AOE,
-        cooldown: haste => 6 / (1 + haste),
+        cooldown: haste => 9 / (1 + haste),
         gcd: {
           base: 1500,
         },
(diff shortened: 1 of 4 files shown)


src/parser/warrior/arms/modules/Abilities.js
```diff
@@ -300,6 +299,9 @@ class Abilities extends CoreAbilities {
         category: Abilities.SPELL_CATEGORIES.UTILITY,
         cooldown: 8,
         gcd: null,
+        castEfficiency: {
+          disabled: true,
```
MartijnHols: What is this supposed to achieve? No such prop exists as far as I am aware.

Sharrq:
I'm pretty sure that's the correct prop. I'll double check.

It was something I added a long time ago, it essentially turns off the x/y casts, the suggestion, and the can be improved. I added it at some point as a better solution instead of omitting the cool down.

src/parser/warrior/arms/modules/Abilities.js
Sharrq: it used to exist. I remember creating it .. but it must have been removed at some point.

It used to disable the efficiency %, the max casts, and the efficiency bar in the abilities section so that it would only say the number of casts. Cause the efficiency of a spell that will likely never get used in a fight or will get used maybe once or twice out of a max 50-60 isnt really helpful imo.

I used to hide that by just getting rid of the cooldown off it, but then made that prop so that you can keep the cooldown info there and still hide the efficiency info. Must have been removed at some point.

MartijnHols: So long as suggestion isn't true, it will not show the suggestion and thus the efficiency percentage. The `disabled` prop hid a lot of data that could still be useful, so it was removed and part of it was implemented in the suggestion boolean.

Sharrq:
removed the prop

PR #2437, opened 2018-10-13
[Enhancement Shaman] Allow early refreshes of Flametongue with Searing Assault talent

mtblanton:
The Searing Assault talent makes Flametongue apply a DoT on hit. This means that refreshing Flametongue early is desirable.

This prevents the suggestion to not refresh Flametongue early when the character has the Searing Assault talent.

Sample parse: https://www.warcraftlogs.com/reports/LFdnzNtybGkDRwrT#fight=23&source=12

### DIFF
--- a/src/parser/shaman/enhancement/modules/core/FlametongueRefresh.js
+++ b/src/parser/shaman/enhancement/modules/core/FlametongueRefresh.js
@@ -10,6 +10,7 @@ class FlametongueRefresh extends Analyzer {
   flametongueCasts = 0;
   earlyRefresh = 0;
   refreshPercentageCast = 0;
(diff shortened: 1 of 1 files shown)


HawkCorrigan:
Bit late to the party.

From what I understand you don't refresh Flame tongue early even when playing with the talent except for the case where you have literally nothing else to do, which in my experience didn't really happen all too often.

Is in your experience not having anything else to do happening so often that you refresh flametongue early more then once per fight on average? @mtblanton

HawkCorrigan:
Yup looks right.
Would you be okay with disabling the module outright when talenting into Searing Assault then?

PR #2441, opened 2018-10-13
[Restoration Shaman] Healing Rain tick bug fix & misc changes

niseko:
Some smaller changes bundled together.
Notable:
- Healing Rain hit count bug fix, with the correct amount of haste, partials would happen at the same time, giving you up to 12 hits for the last tick. The solution isn't perfect as 3 hits + 3 partials will count as 1 tick with 6 hits instead, but that doesn't happen often enough to significantly distort the statistic, and won't get the count >6 (maximum) like it does currently.
- Also filters out full overheals on pets as that indicates that not enough players are standing in it.

Example of partials interfering: (tick + partial within ~20ms)
image
https://wowanalyzer.com/report/ZdGCwjQmXD469Wcx/2-Mythic+Argus+the+Unmaker+-+Kill+(10:06)/12-Spiritshockz

- Healing Rain location for Deluge now scales with the overflowing shores trait, increasing its size.
- Inefficient casts for Healing Wave don't trigger anymore if a Flash Flood buff was used instead, as its essentially the same thing as Tidal Waves.
- Adjusted some scaling Spellinfos
- Added the Rocket Jump racial to the abilities


src/interface/others/RestorationShamanSpreadsheet.js
```diff
@@ -19,7 +19,7 @@ class RestorationShamanSpreadsheet extends React.Component {
 
     const getAbility = spellId => parser._modules.abilityTracker.getAbility(spellId);
     const casts = spellId => getAbility(spellId).casts;
-    const cpm = spellId => casts(spellId) / (parser.fightDuration / 1000 / 60) >= 0 ? (casts(spellId) / (parser.fightDuration / 1000 / 60)).toFixed(2) : '0';
+    const cpm = (spellId, tw = false) => (tw ? getAbility(spellId).healingTwHits : casts(spellId)) / (parser.fightDuration / 1000 / 60) >= 0 ? ((tw ? getAbility(spellId).healingTwHits : casts(spellId)) / (parser.fightDuration / 1000 / 60)).toFixed(2) : '0';
```
emallson: this would be much more readable as a standalone function. nested ternaries are no me gusta

src/parser/shaman/restoration/modules/core/HealingRainLocation.js
```diff
@@ -24,9 +22,10 @@ class HealingRainLocation extends Analyzer {
 
   constructor(...args) {
     super(...args);
-    const hasDeluge = this.selectedCombatant.hasTalent(SPELLS.DELUGE_TALENT.id);
-    const hasRebalancers = this.selectedCombatant.hasFeet(ITEMS.ELEMENTAL_REBALANCERS.id);
-    this.active = hasDeluge || hasRebalancers;
+    this.active = this.selectedCombatant.hasTalent(SPELLS.DELUGE_TALENT.id);
+    if (this.active && this.selectedCombatant.hasTrait(SPELLS.OVERFLOWING_SHORES_TRAIT.id)) {
+      this.healingRainDiameter *= Math.pow(1.1,this.selectedCombatant.traitRanks(SPELLS.OVERFLOWING_SHORES_TRAIT.id).length);
```
emallson: I don't know of any trait effects that don't scale with ilvl that also stack. This one doesn't scale with ilvl, are you sure it stacks?

niseko: yeah, i tested it extensively. each stack has a very noticeable increase in size. here it is 0 -> 3 stacks which is far more than 10%
image

emallson:
wow, good to know!

I should probably go retest a couple of traits to triple-check that they don't stack, sigh...

src/parser/shaman/restoration/modules/core/HealingRainLocation.js
emallson: wow, good to know!

I should probably go retest a couple of traits to triple-check that they don't stack, sigh...

PR #2447, opened 2018-10-15
Ascendence update

HawkCorrigan:
add a check if flame shock is on the target when casting lava burst

also rewrite some of the ascendence stuff

not entirely sure if the gcd check could be done easier but it should work for me
If stuff like that is needed for more modules then making the prop more accessible might be a nice thing to do.

Pretty much just stole the resolveField function form the GlobalCooldown.js

Checked with https://www.warcraftlogs.com/reports/gxWQvyP9dbDtkHpw/#fight=3&source=24

https://cdn.discordapp.com/attachments/271021613938376704/501494928724197376/unknown.png


src/parser/shaman/elemental/modules/core/FlameShock.js
```diff
@@ -15,12 +15,24 @@ class FlameShock extends Analyzer {
...
+      return;
+    }
+
+    const target = this.enemies.getEntity(event);
+    if(target && !target.hasBuff(SPELLS.FLAME_SHOCK.id)){
+      this.badLavaBursts++;
+    }
+  }
 
+  suggestions(when) {
```
Gebuz: I'd suggest using the new suggestion format instead of the old one. Example here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/a3718bd306252403a5990aaa86d2036f848a7929/src/parser/druid/balance/modules/features/MoonfireUptime.js#L30

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -36,6 +55,30 @@ class Ascendance extends Analyzer {
   }
 
   on_byPlayer_cast(event) {
+    const spellId = event.ability.guid;
+    const target = this.enemies.getEntity(event);
+
+
+    if(spellId === SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id) {
+      if(target && target.getBuffUptime(SPELLS.FLAME_SHOCK.id) < ASCENDENCE_DURATION){
```
Gebuz: Ascendance cast doesn't have a target, so this will never be true. How about checking on the first cast with a target after ascendance instead?

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -36,6 +55,30 @@ class Ascendance extends Analyzer {
...
+    if(!this.selectedCombatant.hasBuff(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id, event.timestamp) || spellId === SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id){
+      return;
+    }
+
+    const gcd = this._resolveAbilityGcdField(ability.gcd);
+    if(!gcd){
+      return;
+    }
+
     if (this.selectedCombatant.hasBuff(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id, event.timestamp)) {
```
Gebuz: This if statement will always be true since it's the opposite of the first part of the one in line 73

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -2,23 +2,42 @@ import React from 'react';
...
+    if (typeof value === 'function') {
+      return value.call(this.owner, this.selectedCombatant);
+    } else {
+      return value;
+    }
+  }
+
   numLavaBurstsCast = 0;
   numLightningBoltsCast = 0;
   numOtherCasts = 0;
```
Gebuz: these 3 vars are never used

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -47,24 +90,42 @@ class Ascendance extends Analyzer {
...
           <li>Other Spells: ${this.numCasts.others}</li>
         </ul>
-        `}
+        `;
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id} />}
+        value={`${this.numCasts[SPELLS.LAVA_BURST.id]}`}
+        label="Lava Burst casts"
```
Gebuz: ... during Ascendance? Also average per Ascendance might be a more useful number.

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -47,24 +90,42 @@ class Ascendance extends Analyzer {
...
+        tooltip={tooltip}
       />
     );
   }
+
+  suggestions(when) {
+    const abilities = `Lava Burst${this.selectedCombatant.hasTalent(SPELLS.ELEMENTAL_BLAST_TALENT.id) ? `, Elemental Blast ` : ``} and Earth Shock`;
+    when(this.numCasts.others).isGreaterThan(0)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<span> Maximize your damage during ascendence by not casting weak spells.</span>)
```
Gebuz: I'd suggest adding the abilities that you're supposed to cast to this text aswell, otherwise "other spells" is mentioned before the good spells.

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -35,36 +52,94 @@ class Ascendance extends Analyzer {
     return this.rawUpdate;
```
sMteX: These two getters are essentially the same and `rawUpdate` isn't used anywhere else
```suggestion
    return this.selectedCombatant.getBuffUptime(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id) / this.owner.fightDuration;
```

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -35,36 +52,94 @@ class Ascendance extends Analyzer {
...
   }
 
+  get averageLavaBurstCasts() {
+    return this.numCasts[SPELLS.LAVA_BURST.id]/this.numCasts[SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id];
```
sMteX: Please check for division by zero in case player didn't cast Ascendance
```suggestion
    return (this.numCasts[SPELLS.LAVA_BURST.id] / this.numCasts[SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id]) || 0;
```

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -2,23 +2,40 @@ import React from 'react';
...
+
+  _resolveAbilityGcdField(value) {
+    if (typeof value === 'function') {
+      return value.call(this.owner, this.selectedCombatant);
+    } else {
+      return value;
+    }
+  }
+
+  justEnteredAcendance = false;
```
sMteX: Typo here (and therefore on related lines 65, 69, 70)
```suggestion
  justEnteredAscendance = false;
```

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -35,36 +52,94 @@ class Ascendance extends Analyzer {
...
-        this.numCasts[spellId] += 1;
-      } else {
-        this.numCasts.others += 1;
+    const spellId = event.ability.guid;
+    const target = this.enemies.getEntity(event);
+
+
+    if(spellId === SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id) {
+      this.justEnteredAcendance = true;
+      this.numCasts[SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id]++;
```
sMteX: We don't use increments, please use += 1
```suggestion
      this.numCasts[SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id] += 1;
```

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -35,36 +52,94 @@ class Ascendance extends Analyzer {
...
+    }
+
+    if(this.justEnteredAcendance){
```
sMteX: This logic is somewhat weird. Is the point of this that if the Flame Shock would need to be refreshed during Ascendance (its duration is less than the Ascendance duration), then it's bad cast? If so, you might want to keep track of the Flame Shock duration manually, because as far as I know, `getBuffUptime()` aggregates all Flame Shocks on the target (basically sums up all durations before it as well). Please clarify a little bit.

src/parser/shaman/elemental/modules/talents/Ascendance.js
```diff
@@ -35,36 +52,94 @@ class Ascendance extends Analyzer {
...
+
+    const ability = this.abilities.getAbility(spellId);
+    if(!ability){
+      return;
+    }
+    if(!this.selectedCombatant.hasBuff(SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id, event.timestamp) || spellId === SPELLS.ASCENDANCE_TALENT_ELEMENTAL.id){
+      return;
+    }
+
+    const gcd = this._resolveAbilityGcdField(ability.gcd);
```
sMteX: Why is this here? To ignore spells not on GCD? Why aren't they supposed to be counted as `others`?

src/parser/shaman/elemental/modules/talents/Ascendance.js
HawkCorrigan: Wait so it sums up all flame shocks from different players?

src/parser/shaman/elemental/modules/talents/Ascendance.js
HawkCorrigan: I basically just want to check if you wasted any gcds on spells other then lvb+eb+es.

If you cast anything that is not on the gcd, like trinkets or racials or whatnot, it shouldn't yell at you.

src/parser/shaman/elemental/modules/talents/Ascendance.js
niseko: Should, unless you give the function a sourceID to work with `getBuffUptime(spellId, sourceID)`

sMteX: Even worse, I think it uses past Flame Shocks as well:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/e73129905e5bb4832849d439decb6ed796dd4624/src/parser/core/Entity.js#L81
But from the look of it, try `getBuff`, see what it returns - this line suggests it might have `start` property
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/e73129905e5bb4832849d439decb6ed796dd4624/src/parser/core/Entity.js#L20

HawkCorrigan:
should be fixed now btw

HawkCorrigan:
I rewrote the flame shock check logic a bit so it should work now even if you do multiple casts without a target during ascendence (which shouldn't happen but ye).

PR #2448, opened 2018-10-15
Extend the precast FireElemental checker for StormElemental

HawkCorrigan:
title.

Depends on the fact that Fire and Storm Elementals count as the only pets so it should works as long as there are no trinkets, azerite traits or w/e that spawn "pets" in the sense of the "byPlayerPet_damage" check.

AFAIK and what I checked the only thing I can think of is Echo Of The Elementals Trait whit which the logic still works because the casts are interlinked.

TBF Storm Elemental should not be cast precombat anyway, because it is a .05% damage loss, so it shouldn't be relevant for anyone who plays a proper rotation ^^

Checked with **https://www.warcraftlogs.com/reports/gxWQvyP9dbDtkHpw/#fight=3&source=24**

### DIFF
--- a/src/parser/shaman/elemental/modules/features/StormFireElemental.js
+++ b/src/parser/shaman/elemental/modules/features/StormFireElemental.js
@@ -1,8 +1,9 @@
 import Analyzer from 'parser/core/Analyzer';
 import SPELLS from "common/SPELLS/shaman";
+import TALENTS from "common/SPELLS/talents/shaman";
 import SpellUsable from 'parser/shared/modules/SpellUsable';
 
-class FireElemental extends Analyzer {
+class StormFireElemental extends Analyzer {
 
   static dependencies = {
     spellUsable:SpellUsable,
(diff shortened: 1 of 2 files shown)


HawkCorrigan:
Gonna need to fix a thing with ghuun trinket first before merging

HawkCorrigan:
fixed.

src/parser/shaman/elemental/modules/features/StormFireElemental.js
```diff
@@ -0,0 +1,55 @@
...
+
+  relevantData = this.selectedCombatant.hasTalent(TALENTS.STORM_ELEMENTAL_TALENT.id) ? this.data.StormElemental : this.data.FireElemental;
+
+
+  on_byPlayerPet_damage(event){
+      if(this.last_pet_summon_timeStamp!==null) {
+        return;
+      }
+      if(!this.relevant.damageSpells.includes(event.ability.guid)) {
+        return;
```
Pewtro: Shouldn't this be called `!this.relevantData.damageSpells.includes(...)`

HawkCorrigan: done.
IntelliJ must have missed that for some reason

src/parser/shaman/elemental/modules/features/StormFireElemental.js
```diff
@@ -0,0 +1,55 @@
...
+
+  last_pet_summon_timeStamp=null;
+
+  relevantData = this.selectedCombatant.hasTalent(TALENTS.STORM_ELEMENTAL_TALENT.id) ? this.data.StormElemental : this.data.FireElemental;
+
+
+  on_byPlayerPet_damage(event){
+      if(this.last_pet_summon_timeStamp!==null) {
+        return;
+      }
```
Pewtro: lack of spaces here annoys me 😄

HawkCorrigan: do you mean free lines before and after the ifs?

Pewtro: In this ``      if(this.last_pet_summon_timeStamp!==null) {``

HawkCorrigan: ah aight ^^
I change that at a later point in time :P

PR #2452, opened 2018-10-16
[Brewmaster Monk] [Bug] Checklist fix

joshinat0r:
Checklist failed when a player wasn't using Rushing Jade Wind. This will disable the requirement check when the player wasn't using RJW

### DIFF
--- a/src/parser/monk/brewmaster/modules/features/checklist/Component.js
+++ b/src/parser/monk/brewmaster/modules/features/checklist/Component.js
@@ -21,7 +21,7 @@ class BrewmasterMonkChecklist extends React.PureComponent {
 
   render() {
     const { combatant, castEfficiency, thresholds } = this.props;
-    
+
     const AbilityRequirement = props => (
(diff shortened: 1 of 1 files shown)


joshinat0r:
You're sure about that? RJW has a custom threshold based on its module (essentially checking the uptime), its not a simple "kept it on cooldown"-suggestion

sMteX:
I know, it's just a matter of syntax - what I mean is you don't need ternary operator but just `&&`, and the `React.Fragment <>` is redundant if the only thing inside is another React component. The example was just to show what I mean, not that you should be using `AbilityRequirement`.

joshinat0r:
Oh, yeah you're right. Just blindly coppied emallsons code. Updated the other ones aswell

joshinat0r:
The fragment isn't redundant, the `uptime` makes the parent-element mandatory.
And the parentheses are necessary for multi-line JSX (at least eslint is complaining about it locally)

PR #2454, opened 2018-10-16
[MW] Mana Tea and Lifecycles Suggestion Updates

anom0ly:
Update the Mana Tea and Lifecycles talent suggestions.

Also, add in a cast ratio statistic for EvM and Vivify.

image


yajinni:
Shouldn't src/parser/monk/mistweaver/modules/features/CastRatio.js  have a better name? Feels to generic?

anom0ly:
Yes, I can see that.

Naming convention updated.

PR #2458, opened 2018-10-16
 Allow castEfficiency to be null in ArcaneTorrent 

Salmonhands:
This appears to be the intended functionality based on the fact that this.castEfficiency is later compared to null. 

Might be cleaner to just add support for a 0 casting efficiency if you want me to do that instead.


MartijnHols:
Also please consider adding a changelog entry

Pewtro (changes requested):
With a review already existing, marking this as changes requested. 
- also has a merge conflict to fix.

Pewtro:
Over a month with no updates on this. Closing it for now, if you wish to make the requested updates, you can always just reopen this or a new PR with the updates.

MartijnHols:
Actually it looks like the original file was already changed and does what this PR suggested, so closing since it was already solved.

Thanks for the PR.

PR #2475, opened 2018-10-18
Made regrowth suggestion calculation consider abundance stacks

Qbz23:
The suggestion to cast fewer regrowths without clearcasting now no longer considers regrowths casts with sufficient abundance stacks against you


Qbz23:
@buimichael

buimichael:
@Qbz23 Thanks, please add an entry in CHANGELOG.js with a small info on these changes.

Qbz23:
Added

PR #2479, opened 2018-10-19
[Restoration Shaman] Azerite implementation & Swelling Stream

niseko:
Adds a base class for my azerite as i didn't want to copy & paste the code a dozen times, along with one of the simple traits.
https://www.wowhead.com/spell=275488/swelling-stream

image
image
https://www.warcraftlogs.com/reports/gaj8m7cRHDnvt2zZ/#fight=6&source=17&type=healing
https://www.warcraftlogs.com/reports/3Ca6TBgWLnbR8j2K/#fight=3&source=12&type=healing&translate=true

### DIFF
--- a/src/parser/shaman/restoration/modules/azerite/BaseHealerAzerite.js
+++ b/src/parser/shaman/restoration/modules/azerite/BaseHealerAzerite.js
@@ -124,7 +124,7 @@ class BaseHealerAzerite extends Analyzer {
                     </tr>
                   </thead>
                   <tbody>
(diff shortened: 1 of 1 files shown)


Gebuz:
Just looking at the image, shouldn't overhealing % on trait 1 always be lower than trait 2? Assuming you want to assign the healing done to the first trait first, then second trait etc.

niseko:
I guess that's true and i'm ordering them the wrong way around, since its based on the math. I'll switch them around before displaying.
image

PR #2480, opened 2018-10-19
[Enhancement Shaman] Add ABC to Enhancement checklist

mtblanton:
Adding an Always Be Casting entry to the Enhancement checklist.

Downtime values are stolen from Ret for now. Is there a resource people pull this from or do I need to dig through logs?

`jsconfig.json` fixes an issue I had with VSCode not being able to do the built-in Typescript analysis on modules imported with an absolute path. 

Checked with fights from this raid on Harzwaz: https://www.warcraftlogs.com/reports/PtKqzLw4D62Ynpgk#fight=16&type=damage-done&source=33

image


mtblanton:
Ah, I thought I got all the warnings and didn't rerun lint locally. I'll get that fixed up tomorrow.

src/parser/shaman/enhancement/modules/features/Checklist/Component.js
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  render() {
+    const { thresholds } = this.props;
+
+    return (
+      <Checklist>
+        <Rule
+          name="Always be casting"
+          description={<DowntimeDescription />}
```
sMteX: While not incorrect, this is redundant, you can put the text directly here: 
```suggestion
          description={<>You should try to avoid doing nothing during the fight. If you have to move, try casting something instant with range like <SpellLink id={SPELLS.FLAMETONGUE.id} /> or <SpellLink id={SPELLS.ROCKBITER.id} /></>} 
```

jsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "compilerOptions": {
```
sMteX: I've read `jsconfig.json` is a config file for VSCode, would you mind putting it in .gitignore, so it doesn't interfere with other people's IDEs? That way it should still work for you.

src/parser/shaman/enhancement/modules/features/Checklist/Component.js
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+      <Checklist>
+        <Rule
+          name="Always be casting"
+          description={<DowntimeDescription />} 
+        >
+          <Requirement name="Downtime" thresholds={thresholds.alwaysBeCasting} />
+        </Rule>
+      </Checklist>
```
sMteX: Seeing as you added `preparationRuleAnalyzer.thresholds` into `thresholds`, can you also add PreparationRule here? It can be found in `parser/shared/modules/features/Checklist2/PreparationRule` if I'm not mistaken.

src/parser/shaman/enhancement/modules/features/Checklist/Component.js
mtblanton: I wasn't sure if we'd want that pulled out since it was so long! I'll apply your suggestion!

src/parser/shaman/enhancement/modules/features/Checklist/Component.js
mtblanton: Added! Didn't realize it'd be that easy!

src/parser/shaman/enhancement/modules/features/Checklist/Module.js
```diff
@@ -0,0 +1,38 @@
...
+    combatants: Combatants,
+    castEfficiency: CastEfficiency,
+    preparationRuleAnalyzer: PreparationRuleAnalyzer,
+
+    alwaysBeCasting: AlwaysBeCasting,
+  }
+
+  render() {
+    return (
+      <>
```
sMteX: One last thing that should be removed is this fragment (because <Component> is the only thing inside)

sMteX:
Wait, one last thing - could you add a changelog entry for yourself? Like "Implemented basic Checklist for Enhancement" or something.

PR #2486, opened 2018-10-20
Resto Druid - Clearcasting statistic text abundance fix

Qbz23:
Fixed an issue where the statistic for clearcasting was displaying incorrect regrowth information as a result of recent abundance change


buimichael:
Thanks

PR #2495, opened 2018-10-22
Unify ExpandableStatisticBox with StatisticBox

emallson:
This PR removes the `ExpandableStatisticBox` module by merging its feature-set into `StatisticBox`. The pulldown button only appears when the `StatisticBox` is given children. After testing on every player in several logs I can't find any places where this breaks either non-expandable boxes or expandable boxes.

The motivation for this is that it gives us effectively a `LazyLoadExpandableStatisticBox` for (basically) free. I wanted this functionality for the Brewmaster mastery module, which it turns out is causing some load-time issues on lower end machines (like my laptop, rip). In adding it (in an unpublished version), I realized that much of the code is shared between the two and that merging them would be cleaner.

I will note that lazy loading an expandable box reflows the stat boxes, which may be undesirable.

screenshot from 2018-10-22 17-48-21
screenshot from 2018-10-22 17-48-29
screenshot from 2018-10-22 17-48-38


src/interface/others/LazyLoadStatisticBox.js
```diff
@@ -42,16 +42,19 @@ class LazyLoadStatisticBox extends React.PureComponent {
   }
 
   render() {
-    const { value, ...others } = this.props;
+    const { value, children, ...others } = this.props;
     delete others.loader;
+    delete others.children;
```
MartijnHols: This seems unnecessary.

PR #2498, opened 2018-10-23
Fixed typo in new spec documentation

FROSTB1T:
Fixed typo in new spec documentation


Gebuz:
Hey, thanks for the PR, but it looks like you it's been almost a year since you last pulled from master. We moved the documentation to the wiki more than half a year ago, and "Adding a new spec" has also since been removed (since all specs now have the basics implemented).

PR #2501, opened 2018-10-23
Add Changelog Tracking Per User

### LINKED ISSUE #795: Track which changelog entries are seen so the indicator can show a counter for unseen changes
This should be per-spec based so that looking at the changes for Holy Paladin does not also mark Brewmaster Monk changes as seen. I was thinking of tracking an array of the date that changes were last seen; one entry in such an array for core and 1 per spec.

For example I look at Holy Paladin now it would store 2017-11-12 on core change entries and on holy paladin. Other specs would not include core changes in the changelog but would include that spec's changes.

Or maybe a simpler approach is better; just do it per spec and don't mind that core changes will be shown as new changes for each spec. The benefit is that core changes do affect multiple specs so it's not like they're not relevant for the other specs.

In a cookie probably.

scottymcraig:
Fixes #795 and introduces functionality requested therein.  Data is saved to localstorage based on page view.

_Edit_: Also I should note - the system I'm using for generating a key for each update is SPECyyyyMMdd.Iteration


scottymcraig:
I looked at this first - I tried concatenating a combination of the date, spec and title (namely the first characters of each word), but it still repeated on a few instances.

We could track it by date but that negates them ever seeing any other spec's changes.  While it may not be pertinent (as if they are switching specs later in the game and then come look) because they don't need to see the prior ones, it would technically be inaccurate.

It's up to you, honestly, on how you want to do it.  I can make it work however you want.  I'm all about data integrity and that's why having a more permanent ID was more enticing.

MartijnHols:
Would tracking date per spec and core not achieve the desired effect for the most part? Switching specs would then still remember the spec specific last seen changelog entry date. To solve the seen count for multiple entries per day, you could also store the amount of entries seen on the last day (i.e. If you saw 3 but there are 4 entries, then you missed the last one).

sMteX:
This hasn't been touched in almost 2 months, closing for now. You're free to re-open the PR if you'd like to address the changes.

PR #2502, opened 2018-10-23
[Hunter] Steady Shot GCD, Crows Reset and Aspect of the Wild double dipping

Pewtro:
Updates some cast efficiencies
Updates Aspect of the Wild and double dipping of GCD reduction -- They fixed double dipping of haste and introduced a new double dipping. yay.
Updates A Murder of Crows resetting.

### DIFF
--- a/src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
+++ b/src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
@@ -28,6 +28,7 @@ class AMurderOfCrows extends Analyzer {
   casts = 0;
   applicationTimestamp = null;
   lastDamageTick = null;
+  crowsEndingTimestamp = null;
 
   constructor(...args) {
     super(...args);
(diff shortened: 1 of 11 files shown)


src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
```diff
@@ -43,29 +40,32 @@ class AMurderOfCrows extends Analyzer {
...
 
   on_byPlayer_damage(event) {
     const spellId = event.ability.guid;
-    if (this.lastDamageTick && event.timestamp + MS_BUFFER < this.applicationTimestamp + CROWS_DURATION && event.timestamp > this.lastDamageTick + CROWS_TICK_RATE + MS_BUFFER && this.spellUsable.isOnCooldown(SPELLS.A_MURDER_OF_CROWS_TALENT.id)) {
+    /**
+     * Checks if we've had atleast 1 damage tick of the currently applied crows, and checks that crows is in fact on cooldown.
+     * Then it checks whether the current damage event is less than the full duration of crows, followed by a check to see if more than 1 second has passed since last tick
+     * If more than 1 second has passed, we can assume that crows has been reset, and thus we the CD.
+     */
+    if (this.lastDamageTick && this.spellUsable.isOnCooldown(SPELLS.A_MURDER_OF_CROWS_TALENT.id) && event.timestamp + MS_BUFFER < this.applicationTimestamp + CROWS_DURATION + MS_BUFFER && event.timestamp > this.lastDamageTick + CROWS_TICK_RATE + MS_BUFFER) {
```
Gebuz: Would be easier to read if this was split into multiple lines

Gebuz: `event.timestamp + MS_BUFFER < this.applicationTimestamp + CROWS_DURATION + MS_BUFFER`
`MS_BUFFER` is added on both sides of the less than here.

src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
```diff
@@ -43,29 +40,32 @@ class AMurderOfCrows extends Analyzer {
...
     }
-    if (!this.enemyID) {
-      this.enemyID = encodeTargetString(event.targetID, event.targetInstance);
-    }
     if (this.casts === 0) {
       this.casts++;
       this.spellUsable.beginCooldown(SPELLS.A_MURDER_OF_CROWS_TALENT.id, this.owner.fight.start_time);
       this.applicationTimestamp = this.owner.fight.start_time;
     }
+    //This accounts for the travel time of crows, since the first damage marks the time where the crows debuff is applied
```
Gebuz: Why not use applydebuff instead of damage event?

Pewtro: There is no applydebuff event anymore, which is what broke the original implementation in the first place

Gebuz: oh lul, that's a weird one

src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
Pewtro: Multilined the if statement and split the comments up to make it more readable (imo) and removed the 2x MS_BUFFER

PR #2503, opened 2018-10-24
[Frost Death Knight] Added breath of sindragosa module

jjs451:
Tracks how long the user sustains breath and gives a suggestion if needed

suggestion

image

Example log: https://www.warcraftlogs.com/reports/D4wCaZgKjAXLnMPV/#fight=16&source=13&type=auras&ability=152279

### DIFF
--- a/src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
+++ b/src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
@@ -1,0 +1,99 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import SpellLink from 'common/SpellLink';
+
+import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
+
+import Analyzer from 'parser/core/Analyzer';
+
+class BreathOfSindragosa extends Analyzer{
+  static dependancies = {
+
+  }
+
+  beginTimestamp = 0;
+  casts = 0;
+  badCasts = 0;
+  totalDuration = 0;
+  breathActive = false;
+
+  on_byPlayer_applybuff(event){
+    const spellId = event.ability.guid;
+    if(spellId !== SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id){
+      return;
+    }
+    this.casts += 1;
+    this.beginTimestamp = event.timestamp;
+    this.breathActive = true;
+  }
+
+  on_byPlayer_removebuff(event){
+    const spellId = event.ability.guid;
+    if(spellId !== SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id){
+      return;
+    }
+    this.breathActive = false;
+    const duration = event.timestamp - this.beginTimestamp;
+    if(duration < 15000){
+      this.badCasts +=1;
+    }
+    this.totalDuration += duration;
+  }
+
+  on_finished(event){
+    if(this.breathActive){
+      this.endTimestamp = event.timestamp;
+      const duration = event.timestamp - this.beginTimestamp;
+      if(duration < 20000){
+        this.badCasts +=1;
+      }
+      this.totalDuration += duration;
+    }
+
+  }
+
+  suggestions(when){
+    when(this.suggestionThresholds)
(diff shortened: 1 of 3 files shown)


src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  on_byPlayer_removebuff(event){
+    const spellId = event.ability.guid;
+    if(spellId !== SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id){
+      return;
+    }
+    this.breathActive = false;
+    const duration = event.timestamp - this.beginTimestamp;
+    if(duration < 15000){
```
Gebuz: I would suggest making this a const `GOOD_CAST_THRESHOLD_MS` or something like that.

src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
```diff
@@ -0,0 +1,99 @@
...
+      this.badCasts +=1;
+    }
+    this.totalDuration += duration;
+  }
+
+  on_finished(event){
+    if(this.breathActive){
+      this.endTimestamp = event.timestamp;
+      const duration = event.timestamp - this.beginTimestamp;
+      if(duration < 20000){
```
Gebuz: Shouldn't this be the same as above? (15000 ms).

src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+      this.totalDuration += duration;
+    }
+
+  }
+
+  suggestions(when){
+    when(this.suggestionThresholds)
+      .addSuggestion((suggest, actual, recommended) => {
+        return suggest(<> You are not getting good uptime from your <SpellLink id={SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id} /> casts. Your cast should last <b>at least</b> 15 seconds to take full advantage of the <SpellLink id={SPELLS.PILLAR_OF_FROST.id} /> buff.  A good cast is one that 20 seconds or more.  To ensure a good duration, make you sure have 3 Runes ready and 70 Runic Power pooled before you start the cast.  Also make sure to use <SpellLink id={SPELLS.EMPOWER_RUNE_WEAPON.id} /> before you cast Breath of Sindragosa.</>)
```
Gebuz:
```suggestion
        return suggest(<> You are not getting good uptime from your <SpellLink id={SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id} /> casts. Your cast should last <b>at least</b> 15 seconds to take full advantage of the <SpellLink id={SPELLS.PILLAR_OF_FROST.id} /> buff.  A good cast is one that lasts 20 seconds or more.  To ensure a good duration, make sure you have 3 Runes ready and 70 Runic Power pooled before you start the cast.  Also make sure to use <SpellLink id={SPELLS.EMPOWER_RUNE_WEAPON.id} /> before you cast Breath of Sindragosa.</>)
```

src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+  }
+
+  statistic(){
+      return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.BREATH_OF_SINDRAGOSA_TALENT.id} />}
+        value={`${(this.averageDuration).toFixed(1)} seconds`}
+        label="Average Breath of Sindragosa Duration"
+        tooltip={`You cast Breath of Sindragosa ${this.casts} times for a combined total of ${this.totalDuration / 1000} seconds.  ${this.badCasts} casts were under 20 seconds`}
```
Gebuz: Might want to round the average duration, doesn't really need 3 decimals :P

src/parser/deathknight/frost/modules/features/BreathOfSindragosa.js
jjs451: Actually it probably shouldn't even there.  A cast still going on fight end shouldn't even be counted in this module since you can't trust the duration

PR #2504, opened 2018-10-24
[Enhancement Shaman] Add offensive CD usage rule to checklist

mtblanton:
Also adding missing changelog entry from #2480

Includes checks for Feral Spirit, Earthen Spike, Totem Mastery, and Ascendance

image


mtblanton:
Is there a way to show actual uses / possible uses (if a fight lasts 7 minutes and i use a 3-min CD once, it'd show 1/2) instead of the % value? I feel like I've seen it before but couldn't find it when I poked around.

Pewtro:
> Is there a way to show actual uses / possible uses (if a fight lasts 7 minutes and i use a 3-min CD once, it'd show 1/2) instead of the % value? I feel like I've seen it before but couldn't find it when I poked around.

It should show 1/3 for that fight length. That information is available in the abilities tab though, which makes it somewhat redundant to show in checklist, maybe? But if determined you might be able to steal it off there, else load in a log from each spec and look there?

sMteX:
Eh, there's been several conversations about the idea of cast efficiency and there are 2 "views" on it. For me, I think for spells that have longer cooldown, "casts / max_casts" is more useful metric than "x% on cooldown", because usually they're planned for the encounter and not meant to be spammed on cooldown. Whereas for shorter abilities that are part of the rotation (something like Judgment or Deathbolt) it kinda makes sense to show time on cooldown (which is meant to be maximal), rather than e.g. "45/56 casts". It's hard to make both but if you desire the "casts / max_casts" you should be able to get it via 
```javascript
this.castEfficiency.getCastEfficiencyForSpellId(spellId);
```
This should return an object with `casts` and `maxCasts` among other properties (of course while importing CastEfficiency and setting that as dependency).

mtblanton:
Alright, I added a new `GenericMaxCastEfficiencyRequirement` component that will take a Spell ID and set up the casts/max casts like `GenericCastEfficiencyRequirement`. 

Here's a green parse:
image

A yellow:
image

And a red:
image

src/parser/shared/modules/features/Checklist2/GenericMaxCastEfficiencyRequirement.js
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    const { spell, name, ...others } = this.props;
+
+    const thresholds = this.thresholds;
+    if (!thresholds) {
+      return null;
+    }
+
+    console.log(thresholds);
```
emallson: aside from this stray `console.log`, everything looks good

mtblanton: Are you thinking of something like this?
```
if (max) {
return maxCastsThresholds
} else {
return castsThresholds
}```

I think that makes sense and would get rid of the duplicate code between the two. I was thinking they were different enough to be two components, but thinking about it now this seems better.

mtblanton: I went ahead and added this logic to `GenericCastEfficiencyRequirement` under the isMaxCasts flag!

emallson:
I would like a more prop-like name (maybe `maxCasts`?) but I'm not gonna bikeshed this PR.

dentarin:
so. what do we have to do to update these to get the enhancement analyzer working again? I'm not great at coding, but I can read code just fine and know how to use enhancement shamans. let's get this going guys

Issue #2242, opened 2018-09-20
[General] Azerite Trait Checklist


TJMoats:
The purpose of this issue is to track which non-class specific Azerite traits have been added to WoWAnalyzer.

# General:

- [ ] Earthlink
- [ ] Blood Siphon
- [ ] Lifespeed
- [ ] On My Way
- [ ] Unstable Flames
- [x] Heed My Call

# Tank:

- [ ] Ablative Shielding
- [x] Crystalline Carapace
- [ ] Winds of War
- [ ] Azerite Veins
- [ ] Strength in Numbers
- [ ] Shimmering Haven

# Healer:

- [x] Synergistic Growth
- [x] Ephemeral Recovery
- [x] Blessed Portents
- [x] Bracing Chill
- [x] Concentrated Mending
- [x] Savior
- [x] Woundbinder

# Non-Healer:

- [ ] Azerite Globules
- [x] Gutripper
- [x] Elemental Whirl
- [x] Overwhelming Power

# Defensive:

- [x] Gemhide
- [ ] Impassive Visage
- [ ] Vampiric Speed
- [ ] Longstrider
- [ ] Resounding Protection
- [ ] Self Reliance
- [ ] Azerite Fortification
- [ ] Bulwark of the Masses

# Location:

## Nazmir
- [x] Blood Rite 
- [ ] Filthy Transfusion

## Zuldazar
- [ ] Incite the Pack
- [ ] Rezan's Fury

## Drustvar
- [x] Blightborne Infusion
- [ ] Ruinous Bolt

## Vol'dun
- [x] Swirling Sands
- [ ] Thunderous Blast

## Stormsong Valley
- [x] Tidal Surge 
- [ ] Secrets of the Deep

## Tiragarde Sound
- [x] Meticulous Scheming 
- [ ] Dagger in the Back 
- [x] Tradewinds

## The MOTHERODE!!
- [x] Unstable Catalyst

## Champions of Azeroth
- [ ] Champion of Azeroth

# Engineering: 
- [ ] Barrage Of Many Bombs
- [ ] Ricocheting Inflatable Pyrosaw
- [ ] Relational Normalization Gizmo
- [ ] Synaptic Spark Capacitor
- [ ] Auto-Self-Cauterizer
- [ ] Personal Absorb-o-Tron

# PVP

## Alliance
- [ ] Anduin's Dedication
- [ ] Battlefield Precision
- [ ] Stand As One
[...]

PR #2510, opened 2018-10-25
[General Azerite] Tradewinds trait module

avilene:
Issue #2242
screen shot 2018-10-28 at 10 37 46


fyruna:
Will fix the auto formatting that Prettier did for me.

Gebuz:
I would prefer if it showed the average stat gained as the primary statistic and had the proc count/uptime in the tooltip instead. For instance here is what Overwhelming Power looks like: 
image
You can see in it's implementation how to calculate the stats gained from the trait. Would be nice if it showed the mastery given to friendly targets aswell in the tooltip ("gave X average mastery to other allies" or something like that)

niseko:
Average of X is how most procs of this type are presented, as it gives you a nice readable number. If refreshes trigger it isn't easy to find out, but it also doesn't really matter - you can directly track how often you apply the buff to other players (281844).
https://www.warcraftlogs.com/reports/6xfHtgjh8y7DqTrR#fight=3&type=auras&sourceclass=Any&pins=0%24Separate%24%23244F4B%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24false%24281843&ability=281844&target=1

fyruna:
@niseko I suppose so but it just feels weird to have an average value when it is an actual flat out mastery value (single stack) as opposed to Overwhelming Power that works with stacks.

Dambroda:
Needs to have conflicts resolved; also just for consistency I'll put this in here as well from discord:

The average values arent anything to do with stacks -- it's meant to represent how much of the stat it gave you on average throughout the entire encounter

So if it's procced for 10% of the fight and gives you 2000 mastery when it procs, it gave you an average of 200 mastery

Whether or not it stacks or is just on/off is irrelevant

Dambroda:
Could you post another screenshot of what it looks like now?

src/parser/shared/modules/spells/bfa/azeritetraits/Tradewinds.js
```diff
@@ -0,0 +1,83 @@
...
+  get averageMastery() {
+    return (this.mastery * this.uptime).toFixed(0);
+  }
+
+  handleBuff(event) {
+    if (event.ability.guid !== SPELLS.TRADEWINDS.id) {
+      return;
+    }
+
+    if (event.type === "applybuff" || event.type === "refreshbuff") {
```
Dambroda: Is this a needed check, if you're only calling handleBuff from apply and refresh? I think the reason the other modules have this is that they may check for the buff falling off

fyruna: The buff falling off is not useful to know for this case. This is mainly based off of BloodRite module where it only calls apply/refresh.

Pewtro: Right, but you only ever access handleBuff in any of those events. 
You can safely remove the entire if and simply do procs += 1

Dambroda: I think the blood rite module copied it from somewhere else as well, and that somewhere *did* need the check. If you want to remove it from BloodRite in this PR as well that would be fine

fyruna:
@Dambroda can I have one last eye on this? Made it like ElementalWhirl :)

Dambroda (approved):

PR #2512, opened 2018-10-26
[Azerite] Adding Woundbinder

TJMoats:
As far as I can tell, this is correct. There are no published numbers for this azerite trait, so I built this module based on the theory crafters info in the priest discord. 

screen shot 2018-10-25 at 7 22 12 pm

### DIFF
--- a/src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
+++ b/src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
@@ -52,10 +52,12 @@ class WoundBinder extends Analyzer {
   }
 
   calculateHasteAmount(targetHealthPercent = 1) {
+    // The formula is HasteGained = .34 + percentMissingHealth * .66
     return this.fullHasteValue * (BASE_HASTE_AMOUNT + ((1 - BASE_HASTE_AMOUNT) * (1 - targetHealthPercent)));
   }
 
   on_byPlayer_heal(event) {
+    // We track the last heal from the player to use for the haste calculation
     this.lastTargetHealedPercent = (event.hitPoints / event.maxHitPoints);
   }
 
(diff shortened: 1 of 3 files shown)


Gebuz:
Would be nice if you could also added the haste gained to stat tracker, I can't remember if it's possible to add dynamic amounts though.

TJMoats:
It doesn't look like there is a way to work the variability of WoundBinder into the stat tracker. I have a formula that calculates the average value of Woundbinder that I can maybe use?

src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
```diff
@@ -0,0 +1,83 @@
...
+}, {
+  haste: 0,
+});
+
+const BASE_HASTE_AMOUNT = .34;
+
+/**
+ Your healing effects have a chance to increase your Haste by up to 435 for 6 sec. Healing lower health targets will grant you more Haste.
+
+ This trait scales linearly starting at with 34% of the haste at 0% health and 100% at 100% health.
```
HawkCorrigan: So this is pure speculation at this point?

src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
```diff
@@ -0,0 +1,87 @@
...
+    if (this.active){
+      const { haste } = woundBinderStats(this.selectedCombatant.traitsBySpellId[SPELLS.WOUNDBINDER.id]);
+      this.fullHasteValue = haste;
+    }
+  }
+
+  calculateHasteAmount(targetHealthPercent = 1) {
+    // The formula is HasteGained = .34 + percentMissingHealth * .66
+    return this.fullHasteValue * (BASE_HASTE_AMOUNT + ((1 - BASE_HASTE_AMOUNT) * (1 - targetHealthPercent)));
+  }
```
HawkCorrigan: I would flip either the comment or the term around so they actually match

src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
HawkCorrigan: If it is I would at least add something along the lines of:
"Values are base on tc in the priest discord because no game data is available".

TJMoats: Done

src/parser/shared/modules/spells/bfa/azeritetraits/WoundBinder.js
TJMoats: Done

PR #2521, opened 2018-10-27
Water Elemental statistics

### LINKED ISSUE #2380: [Frost Mage] Forgetting to Summon Water Elemental when specced into anything other than Lonely Winter is not suggested in results.
Example report:
https://www.warcraftlogs.com/reports/hDnPwY8pzxbAcMmH/
Sheepishly reporting, but I didn't summon my Elemental thinking I had Lonely Winter, right up to pull 13. I think it should be suggested as "Major" in player suggestions, if damage from Water Elemental simply returns nill.

jos3p:
This will add a statistics block for the Water Elemental to make dps and actual castinguptime more visible.
waterelestats

fixes #2380


src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,156 @@
...
+			});
+	}
+
+	statistic() {
+    return (
+			<StatisticBox
+  position={STATISTIC_ORDER.CORE(15)}
+  icon={<SpellIcon id={SPELLS.SUMMON_WATER_ELEMENTAL.id} />}
+  value={(
+    <span>
```
Dambroda: Fragments in react are represented by an empty set of brackets: <>   (instead of span)

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,156 @@
...
+      <Icon
+        icon="spell_mage_altertime"
+        style={{
+          height: '1.2em',
+          marginBottom: '.15em',
+        }}
+       />
+      {' '}{formatPercentage(this.activeTimePercentage)}{' %'}
```
Dambroda: You don't need to spread out the brackets like this, use the same concat you use in L143

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,156 @@
...
+      const petId = event.sourceID;
+      this._byPet[petId] = this.byPet(petId).add(event.amount, event.absorbed, event.blocked, event.overkill);
+      if(event.ability.guid === SPELLS.WATERBOLT.id) {
+        this._waterboltHits += 1;
+      }
+    }
+  }
+
+  on_byPlayerPet_cast(event) {
+    if (event.ability.guid === SPELLS.WATERBOLT.id) {
```
Dambroda: We prefer to exit-early in these event listeners rather than use multlevel if statements:
if(guid != WATERBOLT) {
  return;
}

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  get totalByPets() {
+    return Object.keys(this._byPet)
+      .map(petId => this._byPet[petId])
+      .reduce((total, damageValue) => total.add(damageValue.regular, damageValue.absorbed, damageValue.blocked, damageValue.overkill), new DamageValue());
+  }
+
+  on_byPlayerPet_damage(event) {
+    if (!event.targetIsFriendly) {
```
Dambroda: We prefer to exit-early in these event listeners rather than use multlevel if statements

src/parser/mage/frost/modules/checklist/Component.js
```diff
@@ -68,6 +68,7 @@ class FrostMageChecklist extends React.PureComponent {
           {combatant.hasTalent(SPELLS.GLACIAL_SPIKE_TALENT.id) && <Requirement name="Glacial Spike Utilization" thresholds={thresholds.glacialSpikeUtilization} />}
           {combatant.hasTalent(SPELLS.THERMAL_VOID_TALENT.id) && <Requirement name="Thermal Void Avg Duration" thresholds={thresholds.thermalVoidDuration} tooltip="Thermal Void adds time to your Icy Veins every time you cast an Ice Lance that benefits from Shatter (this includes Fingers of Frost proc usage and Ice Lance's that land in Winter's Chill). Maximizing this uptime will increase your damage." />}
           {combatant.hasTalent(SPELLS.RUNE_OF_POWER_TALENT.id) && <Requirement name="Rune of Power Uptime" thresholds={thresholds.runeOfPowerBuffUptime} tooltip="Using Rune of Power effectively means being able to stay within the range of it for it's entire duration. If you are unable to do so or if you frequently have to move out of the range of the buff, consider taking a different talent instead." />}
+          {!combatant.hasTalent(SPELLS.LONELY_WINTER_TALENT.id) && <Requirement name="Water Elemental Utilization" thresholds={thresholds.waterElementalUptime} />}
```
Dambroda: Something we need to fix in the mage modules is the inconsistent capitalization with the rest of WoWAnalyzer; may as well start now -- only spell names and the first words of sentences should be capitalized, even in these titles. (Water Elemental utilization)

jos3p:
I was thinking of tuning the ranges of the suggestion to like

- Major: 0.20 (passive-mode or not/late summoned)
- Avg: 0.50
- Minor issue: 0.75

But while playing with the values i found they can be dynamic, so why not use 
Bonechill vs. Water Elemental compare to check if we have a minor or major=avg issue.

Actually the only important point is when Bone Chilling adds more DPS than the Elemental did.

0. WaterEleDPS
1. [(overall dps - elementaldps) -5%], because no more BoneChilling
2. DPSbc = fb/lance/flurrydps +25%

WaterEleDPS / DPSbc >= 1 -> no issue
WaterEleDPS / DPSbc < 1 -> bad, major, BC would be better

But is it even worth calculating all this if we "know" ~75% uptime is ok?

Dambroda:
It's an interesting idea but what if the player doesn't take bone chilling and takes ray of frost instead? Also I think you mean Lonely Winter when you're comparing the 25%?

The other way (that doesn't rely on a talent) would be to compare the water elemental uptime to the players uptime. In Uldir I think quite a few fights average only around 80% uptime so if we then see the ele has 80% uptime when the player does as well we can assume there was just nothing to hit.

jos3p:
Yes Lonely Winter, sorry good point with ray of frost.
I will look into comparing the uptimes and see what i can do.

src/parser/mage/frost/modules/checklist/Component.js
Sharrq: Yea its a bad habit that I do a lot, ill get around to fixing them at some point.

Sharrq:
Just to add on to your comments Dambroda, I think what you said is mostly accurate, though i wouldnt count casting Summon Water Ele is a mistake midfight if the Ele died during the fight and had to be recast.

There was also fights like Kilrogg in HFC where if you got put into the other room with your Water Ele and came back, your water ele would still be in that room and you would have to recast him or else he would just be sitting over there doing nothing. Granted there arent any fights that do that in the current tier, but that doesnt mean there never will be.

So I think its important to check all of the following
- Was the Water Elemental summoned on pull
- How much of the fight was the Water Elemental not summoned (if it was summoned for part of the fight but not all of it .... i.e. they realized mid fight that it was missing or it died and they didnt notice)
- The pet's downtime (the time it wasnt casting)

Additionally there are situations where like if your Water Elemental is up when you go through the Cysts hallway to Ghuun from Mythrax your Ele will get stuck on the cysts and just cast on them. If the boss gets pulled while he is stuck back there, then it would likely have 0 downtime and be summoned the whole fight but didnt contribute to the fight at all ..... Though now that i think about it, i wonder if the Ele would run into the room once they started casting on the boss/p1 adds ...

jos3p:
so here is what changed compared to first version:

- restructured for better overview

- copy/paste from ABC to get player uptime -- is there a better way to do this?

- petuptime suggestion threshhold now depend on player uptime: -10%, -25%, -30% (eg. player has 83% so the pet should have 73%+) *

- added another suggestion for prePullSummon check: if petrelated actions (begin cast/cast complete/damage Waterbolt) happen between [for mythraxlike pulls?] 100ms-7.5sec* after pull we asume it was casted prepull

- fixed requested changes

update


(* stricten these numbers?
Pet uptime = Time spent casting)

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,241 @@
...
+    }
+  }
+  
+  on_byPlayerPet_damage(event) {
+    if (event.ability.guid !== SPELLS.WATERBOLT.id) {
+      return;
+    }
+    else if (!event.targetIsFriendly) {
+      const petId = event.sourceID;
+      this._byPet[petId] = this.byPet(petId).add(event.amount, event.absorbed, event.blocked, event.overkill);
```
Dambroda: I didn't notice this the first time, but I'm not really seeing the reason to complicate the damage calculations here. What module did you take this from? Because we only have one pet, it's really probably as simple as storing a module variable called _waterboltDamage and on this line instead just doing:
this._waterboltDamage += event.amount + event.absorbed;

And then you can remove lines 94-98, 101-105, etc.

jos3p: I took it from parser/shared/modules/DamageDone.js

Thanks for taking time to review, I will add all these changes shortly

Dambroda: Thanks for working on this! Also I made a mistake in my comment there, its supposed to be event.absorbed || 0 (the mirror image example has it correct)

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,199 @@
...
+      this.wasCastStarted = true;
+      this._timestampLastCast = event.timestamp;
+    }
+  }
+  
+  on_byPlayerPet_cast(event) {
+    if (event.ability.guid !== SPELLS.WATERBOLT.id) {
+      return;
+    }
+    else {
```
Dambroda: Because the last if statement exits-early (with return), you can remove this _else {_  (just make sure to remove the closing bracket on L77 as well!)

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+      this.wasCastStarted = false;
+    }
+  }
+  
+  on_byPlayerPet_damage(event) {
+    if (event.ability.guid !== SPELLS.WATERBOLT.id) {
+      return;
+    }
+    else if (!event.targetIsFriendly && event.ability.guid === SPELLS.WATERBOLT.id) {
```
Dambroda: You've already exited early if the spell guid does not equal waterbolt, so you don't need to check it again here. In fact it would be better to say:

```
if(event.ability.guid !== SPELLS.WATERBOLT.id || event.targetIsFriendly) {
  return;
}
```

And then you can just remove line 84 and the closing bracket on 90.

src/parser/mage/frost/modules/features/WaterElemental.js
```diff
@@ -0,0 +1,199 @@
...
+import Abilities from 'parser/shared/modules/Abilities';
+import Channeling from 'parser/shared/modules/Channeling';
+import Haste from 'parser/shared/modules/Haste';
+import AlwaysBeCasting from './AlwaysBeCasting';
+
+class WaterElemental extends Analyzer {
+  static dependencies = {
+    haste: Haste,
+    abilities: Abilities,
+    globalCooldown: GlobalCooldown, // triggers the globalcooldown event
```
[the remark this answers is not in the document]
jos3p: Ah ok, initially I thought they were needed for some *magic* calculations during Bloodlust/TW maybe because it was included in the source I took them from and they are not used there too (parser/shared/modules/AlwaysBeCasting.js), now tested both and the numbers do stay the same

jos3p:
Almost the same I did with getting the player damage, just with parser/shared/modules/CancelledCasts.js, so I guess I should just import that class too, just like with ABC?

Dambroda:
Ah, ok then. We can leave that part as-is now and maybe figure out later if there's a better way to do it

jos3p:
I did not find a result which could help
`grep -Rl "on_byPlayerPet_cast(event)" .`
`./parser/mage/frost/modules/features/WaterElemental.js`

PR #2522, opened 2018-10-28
[BrM] Changed Purifying Brew cast checklist

emallson:
screenshot from 2018-10-27 19-30-52
screenshot from 2018-10-27 20-21-25

This replaces the cast efficiency checklist item (which is shared with Ironskin Brew due to the shared cooldown/charges). This makes it clearer what a reasonable number of purifies to hit is. If you're casting more than this, either there is downtime where you can safely regenerate brews or you are failing the ISB checklist item (being hit with ISB down).

Three other changes in here:

- (minor) added suggestions for existing PB checklist items
- (minor) deleted old legendary code that'd *somehow* stuck around
- Added [harmonic mean](https://en.wikipedia.org/wiki/Harmonic_mean) as an option for rule performance. This more heavily penalizes underperforming on an individual requirement, without going all the way to `lowest`. Without this, the pictured checklist would pass when it really ought to fail (the first two requirements can technically be passed with 0 casts, but i also don't want them ignored).


src/parser/monk/brewmaster/modules/core/BrewCDR.js
```diff
@@ -43,11 +42,13 @@ class BrewCDR extends Analyzer {
...
     totalCDR += this.bob.cdr;
-    // ...and wrists
-    totalCDR += this.wrists.cdr;
     return totalCDR;
   }
 
+  // ok emallson, for the last time *WRITE DOWN YOUR DERIVATIONS*
+  //
+  // not the derivation, but close enough i guess:
+  // https://github.com/WoWAnalyzer/WoWAnalyzer/pull/1238#discussion_r163734298
```
MartijnHols: Please consider putting the actual comment content here instead

PR #2526, opened 2018-10-28
[RFC] Manual event listening (comment, but do not merge)

MartijnHols:
# Please comment but do not merge

Discussed this shortly on disco; this is my idea for a new way to listen to (combat log) events.

We'd add event listeners by calling addEventListener in the constructor instead of magically through reflection. This solves the problems: 

The `addEventListener` would have the following interface:

```jsx
addEventListener(string|EventFilter eventFilter, func handler): void
```

You'd call it like so:

```jsx
this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.LIGHT_OF_DAWN_CAST), this._onCast);
```

`Events` is an object we make with all available events, except fabricated events. Fabricated events would have to be imported from the module that fabricates them, so the core `Events` object doesn't get messy with spec-specific stuff. Bonus is that it better reveals what modules rely on what fabricated events since these modules would have to be imported.

The `by` filter supports a bitwise operator, e.g. `by(SELECTED_PLAYER | SELECTED_PLAYER_PET)` for all events involved the player and his pet. `to` would also exist. `SELECT_PLAYER` is an import from the analyzer module (probably).

We probably should autobind handler for comfort. Since it's only called in the constructor it shouldn't hurt performance.

----------------------

**This PR does not include a complete implementation. Look at the Light of Dawn module for a diff from the legacy event listening to the proposed method.**

----------------------

**Pros:**
- **Less magic.** `addEventListener` is transparent in what it does as most people are aware of the [observer pattern](https://www.datchley.name/es6-eventemitter/). The actual implementation regardless of how complicated is an implementation detail. This reduces learning curve slightly (although the legacy method also is easy to learn)
- **Allows for easy filtering.** Default filters would be easier to use.
- **Allows adding multiple listeners per event.** This allows better separation of concerns and makes inheritance easier.
- **Allows adding default listeners for common functions.** Stuff like counters (amount of times an event occurred), reducers (such as sum healing/damage done), and uptimes could be provided through easier to use standard solutions.
- **Less boilerplate code.** Checking for things such as spell id is common in many listeners right now, this takes attention away from the logic that matters.
- **The Events class makes it easier to find out about the available events.** In the legacy event listening it's not viable to make this autocomplete.

**Cons:**
- **Not as easy to use as the legacy method.** Existing method is really quick to add a listener to a simple module to do something. New method requires adding a constructor and manually listening to events.
- **Filters are disconnected from the listeners.** Currently the listeners include all filters so they're complete, with the proposed solution they'd be partially in the constructor.

There are no concrete plans to deprecate the legacy method yet. We could end up supporting both, although the option also exists the legacy event listening will be deprecated once we're convinced manual event listening works well enough.

### DIFF
--- a/src/parser/core/EventFilter.js
+++ b/src/parser/core/EventFilter.js
@@ -1,16 +1,30 @@
 class EventFilter {
-  event;
-  constructor(event) {
-    this.event = event;
+  eventType;
+  constructor(eventType) {
+    this.eventType = eventType;
   }
-  by;
-  by(by) {
-    this.by = by;
+  _by;
+  by(value) {
+    if (value === undefined) {
+      return this._by;
+    }
+    this._by = value;
     return this;
   }
-  spell;
-  spell(spell) {
-    this.spell = spell;
+  _to;
+  to(value) {
+    if (value === undefined) {
+      return this._to;
+    }
+    this._to = value;
+    return this;
+  }
+  _spell;
+  spell(value) {
+    if (value === undefined) {
+      return this._spell;
+    }
+    this._spell = value;
     return this;
   }
 }
--- a/src/parser/core/EventFilter.js
+++ b/src/parser/core/EventFilter.js
@@ -1,3 +1,7 @@
+export const SELECTED_PLAYER = 1;
+export const SELECTED_PLAYER_PET = 2;
+const VALID_BY_FLAGS = SELECTED_PLAYER | SELECTED_PLAYER_PET;
+
 class EventFilter {
   eventType;
   constructor(eventType) {
@@ -8,8 +12,14 @@ class EventFilter {
     if (value === undefined) {
       return this._by;
     }
+    if (!this.constructor.validateBy(value)) {
+      throw new Error(`by filter not recognized: ${value}`);
+    }
     this._by = value;
     return this;
+  }
+  static validateBy(value) {
+    return (value & VALID_BY_FLAGS) === value;
   }
   _to;
   to(value) {
--- a/src/parser/core/EventFilter.js
+++ b/src/parser/core/EventFilter.js
@@ -1,0 +1,18 @@
+class EventFilter {
+  event;
+  constructor(event) {
+    this.event = event;
+  }
+  by;
+  by(by) {
+    this.by = by;
+    return this;
+  }
+  spell;
+  spell(spell) {
+    this.spell = spell;
+    return this;
+  }
+}
+
+export default EventFilter;
--- a/src/parser/core/Events.js
+++ b/src/parser/core/Events.js
@@ -1,11 +1,178 @@
 import EventFilter from './EventFilter';
 
 const Events = {
+  /**
+   * BEWARE: These events/properties are NOT COMPLETE. See the Event log for a full list of available props and events.
+   *
+   * Generic props:
+   * - timestamp: the timestamp of the event, relative to log start
+   * - type: the event type, you should generally not use this and properly separate event listeners.
+   * - sourceID: who initiated the event
+   * - sourceIsFriendly: whether the source was friendly to the selected player
+   * - targetID: who was affected by the event
+   * - targetIsFriendly: whether the target was friendly to the selected player. BEWARE: @Any dps classes: make sure if you do a damage statistic that you do NOT include friendly fire to other players (such as from Aura of Sacrifice). This does not gain any damage bonuses.
+   * - ability: object of the ability/spell involved. Shape: { name, guid, type: I believe this is the magic school type, abilityIcon }
+   * - resourceActor:
+   * - classResources: array of resources (mana, energy, etc)
+   * - hitPoints:
+   * - maxHitPoints:
+   * - attackPower:
+   * - spellPower:
+   * - x: x location on the map. See paladin/holy/modules/features/MasteryEffectiveness for an example module that uses this data.
+   * - y: y location on the map
+   * - facing: the direction the player is facing
+   * - sourceMarker:
+   * - targetMarker:
+   * - mapID:
+   * - itemLevel:
+   */
+
+  /**
+   * This event is called for events where the player, a player pet or a target of the player/player pet dealt or took damage.
+   * Event specific props:
+   * - amount: effective damage
+   * - absorbed: damage absorbed by a buff on the target (e.g. https://www.wowhead.com/spell=269279/resounding-protection). This should generally be considered effective damage.
+   * - overkill: if the target died, this is the amount of damage that exceeded their remaining health
+   * - hitType:
+   * - mitigated:
+   * - unmitigatedAmount:
+   * - tick:
+   *
+   * NOTE: Do not use this event to track absorb-healing (e.g. by a spell such as Resounding Protection). Use the `absorbed` event instead.
+   * @returns {EventFilter}
+   */
+  get damage() {
+    return new EventFilter('damage');
+  },
+  /**
+   * This event is called for events where the player, a player pet or a target of the player/player pet was healed.
+   * Event specific props:
+   * - amount: effective healing
+   * - absorbed: healing absorbed by a debuff on the target (e.g. https://www.wowhead.com/spell=233263/embrace-of-the-eclipse). This should generally be considered effective healing.
+   * - overheal: overhealing
+   * @returns {EventFilter}
+   */
+  get heal() {
+    return new EventFilter('heal');
+  },
+  /**
+   * Triggered in addition to the regular heal event whenever a heal is absorbed. Can be used to determine what buff or debuff was absorbing the healing.
+   * NOTE: This should only be used if you need to know **which ability soaked the healing**. If you want to track the amount of absorbed healing by a spell, use the `absorb` prop of the `heal` event.
+   * @returns {EventFilter}
+   */
+  get healabsorbed() {
+    return new EventFilter('healabsorbed');
+  },
+  /**
+   * This event is called for events where the player, a player pet or a target of the player/player pet was healed.
+   * Event specific props:
+   * - ability: The ability responsible for absorbed damage (i.e. the shield)
+   * - amount: effective damage absorbed
+   * - attackerID:
+   * - attackerIsFriendly:
+   * - extraAbility: The damage ability that was absorbed
+   * @returns {EventFilter}
+   */
+  get absorbed() {
+    return new EventFilter('heal');
+  },
+  /**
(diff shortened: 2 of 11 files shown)


Pewtro:
Please don't deprecate old method. I'd be really bummed as I'm quite anti this as explained on discord. 
I disagree on less magic and the learning curve being smaller with this.

MartijnHols:
@Pewtro Since you're such an important contributor, your feedback weighs heavily for this RFC. I did consider not deprecating the old method at all as your feedback on disco which was quite clear, but I think allowing both might lead to an inconsistent and messy codebase. As I said in the original post, in the end we could end up supporting both despite that. But I'm hoping to win you over before then. 😄 

We'll have to cross that bridge when we get to that.

Dambroda:
I can't see how this is more magic than the current reflection event listeners. I personally would very much appreciate a new way to bind to events, both from a development standpoint and a design standpoint. IMO, it does not get more 'magic' than reflection, there is literally no way to find out how to bind to the current events besides looking at the development wiki or copying existing analyzers. Even then, you have no way to figure out what all the event options are that you can bind to unless you already know the names of all the events. Having those options in code so that IDE's can perform those lookups (as they should be able to) would be fantastic.

MartijnHols:
Indeed, this is a concern with the proposal. I considered a few alternatives but I don't think there's anything worthwhile. We'll probably just have to name the listeners properly.

jjs451:
I like the design itself but I'm not sure how I feel about adopting it.  We should pick one design and use it consistently throughout, to echo what others have said.  The sheer work involved in converting everything is something to weigh.  Something else to weigh is this does feel more unfriendly to newcomers.  If we were a team of professional coders I'd say its fine but for an open source project where we already don't have enough support for all the specs I'm not sure I agree with raising the bar for entry.  There's my 2 cents

MartijnHols:
Should be able to make a codemod to convert all existing analyzers, should we get to that point.

I think this is more friendly to newcomers because the observer pattern is so common and thus there's less magic in event handlers, which makes it so you don't have to understand how the magic event listeners work. The analyzers will all look like they're complete classes without any code hidden away needed to work (just an event emitter). The Events object makes it easy to discover the available events too, and the fluent syntax shows the possible filters. All things reducing the learning curve and I reckon is friendlier to newcomers. I do think it might be slightly harder for people new to programming to pick up, but they'll gain in working with a pattern that they'll see and possibly implement later.

MartijnHols:
Thanks a lot for all the feedback! The changes required for implementing this seems to be finalized (a review of the code in this PR would be welcome), and as such I'd like to write out my plans.

My plan is to merge this in soonish and update guides to match. I plan on recommending this approach over the legacy event listening and while I would like to deprecate the legacy method right away, based on your feedback I'll delay the deprecation. I'm pretty convinced this proposal would be an improvement, but I'm afraid I might be prejudiced so I'll slow down.

This means once this is merged, the manual event listening will be available in an experimental (but recommended) state. It could still be scrapped if we learn it's not as great as expected, and I'll take responsibility for reverting any modules that adopted this event listening approach.

Beyond this PR I am thinking of implementing Flow (or maybe TS after all) with a focus on making types for events so the event props are decently documented, auto-completed and possibly even automatically checked.

Sharrq:
I dont know that i fully comprehend all the benefits of this, but based on my understanding of this, i like it. Ill probably move some of the mage modules over to this once its merged since some of them have gotten rather complicated and annoying to maintain (especially stuff like Combustion and Hot Streak for Fire)

Dambroda:
One thing I think we should do immediately is disallow mixing the two types of listeners in one module. Whether or not parents should enforce children to match is up for debate (or whatever the limitation would be), but I think at the very least within the same file it would get way too messy to allow the old and new types in the same place.

src/parser/core/Events.js
```diff
@@ -0,0 +1,179 @@
...
+   * This event is called for events where the player, a player pet or a target of the player/player pet was healed.
+   * Event specific props:
+   * - ability: The ability responsible for absorbed damage (i.e. the shield)
+   * - amount: effective damage absorbed
+   * - attackerID:
+   * - attackerIsFriendly:
+   * - extraAbility: The damage ability that was absorbed
+   * @returns {EventFilter}
+   */
+  get absorbed() {
```
Dambroda: This has the same eventfilter as heal(), but the jsdocs are different? I don't deal with heal events so I am not sure if this is intended or not

src/parser/core/Events.js
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Event specific props:
+   * - stack
+   * @returns {EventFilter}
+   */
+  get removedebuffstack() {
+    return new EventFilter('removedebuffstack');
+  },
+  get refreshbuff() {
+    return new EventFilter('removedebuffstack');
```
Dambroda: Same as previous comment, are these eventfilters correct?

sMteX:
Also, thought occurred - is it planned to make the system support triggered/fabricated events? So far they're not listed.

src/parser/core/EventFilter.js
```diff
@@ -0,0 +1,32 @@
...
+  eventType;
+  constructor(eventType) {
+    this.eventType = eventType;
+  }
+  _by;
+  by(value) {
+    if (value === undefined) {
+      return this._by;
+    }
+    this._by = value;
```
Dambroda: I feel like one or both of these should be in these setters:
1. Clear functionality definition when calling these setters more than once: is it supposed to overwrite the existing filter? That's not really standard builder convention -- maybe it should throw a state error if _by is already set and you call this again? Or should it instead set `this._by = this._by | value;` ?
2. Parameter checking: if one of these values expects one of the enum values, check to make sure it exists?

Also, did you consider having the EventFilter object actually have a filter() function in it that applies its filters to the given event, rather than doing the compiling in the emitter to prepend the checks? I feel like you would eliminate like 10-20 lines of code and it would make more design sense to just have it be inside the filter object.

Dambroda:
> Also, thought occurred - is it planned to make the system support triggered/fabricated events? So far they're not listed.

Good idea

Events.custom('mycustomevent').to(PLAYER).. etc?

sMteX:
> Good catch
> 
> Events.custom('mycustomevent').to(SELECTED_PLAYER).. etc?
Could be. Another option would be either to add another listeners to the Events.js, or have another class with custom events (that could maybe inherit from Events), but that seems messy.

src/parser/core/Events.js
MartijnHols: copy paste error

MartijnHols:
My plan is to have modules that fabricate events return their own (static?) prop which returns the EventFilter. The Events file will only include events provided by WCL.

sMteX:
> My plan is to have modules that fabricate events return their own (static?) prop which returns the EventFilter. The Events file will only include events provided by WCL.

How would that work then? We'd have to import the module and use the prop? Something like 

(ResourceTracker.js)
```javascript
static get spendResource() {
    return new EventFilter('spendresource');
}
```

(e.g. SpellUsable.js, listening to the event and resetting CD)
```javascript
import ResourceTracker from '......';
...
constructor(...args) {
    super(...args);
    this.addEventListener(ResourceTracker.spendResouce.by(SELECTED_PLAYER), this._onSpendResource);
}
```

src/parser/core/EventFilter.js
MartijnHols: In my experience overwriting the existing filter is quite common, as well as making all these methods functional but that's not as performant and I can't imagine this will ever go wrong.

They're not enum values, they're flags. Hard to check for all possible combinations and again there's little to gain I think.

I hadn't considered it, having the filters inside EventFilter does sound appealing, but a big advantage of the compilation is that it's the most performant. With all this logic in the EventEmitter - which I don't think is a bad place for the filter checking either, making EventFilter a dumb struct - it's already well isolated so should be no issue.

MartijnHols:
Pushed fixes for the feedback, still considering a solution for disallowing a combination of the legacy event listeners and the manual event listening.

src/parser/core/EventFilter.js
Dambroda: All of my flags are enums in C#, that's why I used that name 😄  (C# has a [Flags] attribute you can apply to enums) -- but what I mainly meant was that you'd just validate like this:

```javascript
const VALID_FLAGS = SELECTED_PLAYER | SELECTED_PLAYER_PET | OTHER_PLAYER /*etc*/;
validate(flags) {
  return flags & VALID_FLAGS == flags;
}
```
It is a small gain, but it's also pretty small cost IMO

Wouldn't it be more performant to not have to compile at all? Having the filtering done inside the event filter should perform exactly the same outside of just skipping the compiling steps. (instead of prepending after you've built the event, you add in steps to the filter() function as the .spell, .by, etc. functions are called; meaning if .to is never called you are never even performing the if(filter.to) check)

Dambroda: I could be completely visualizing it wrong though. Could you push to test maybe?

Dambroda: I havent tested this or anything and I don't know if this works like I think it would (really not a javascript guru), but this is my general idea below. 

```javascript
  by(value) {
    // Could validate here :]
    const requiresSelectedPlayer = value & SELECTED_PLAYER;
    const requiresSelectedPlayerPet = value & SELECTED_PLAYER_PET;

    let check;
    if (requiresSelectedPlayer && requiresSelectedPlayerPet) {
      check = event => this.owner.byPlayer(event) || this.owner.byPlayerPet(event);
    } else if (requiresSelectedPlayer) {
      check = event => this.owner.byPlayer(event);
    } else if (requiresSelectedPlayerPet) {
      check = event => this.owner.byPlayerPet(event);
    }

    // if we validated we could move this if up and exit early instead
    if(check) {
      const old = this.filter;
      this.filter = function(event) {
        if(!check(event)) {
          return false;
        }
        return old();
      };
    }
    
    return this;
  }

filter(event) {
  return true;
}

```

This would functionally add filters with later calls instead of overwrite them though -- to do it the other way you'd just save _by, _spell, and _to as functions and then filter(event) probably need to call all three (even if they just return true;), which is probably what you were taking into account when you said it would be a performance hit

MartijnHols: As I said, I kinda like that EventFilter acts like a simple struct, I'm not really convinced moving this logic would be an improvement. I did implement everything else though.

MartijnHols:
Thanks all! Let's goooo

Issue #1145, opened 2018-01-07
[Beast Mastery] add a suggestion if people use Multishot on ST bosses


Pewtro:

Pewtro:
Fixed in #2528

PR #2528, opened 2018-10-29
[Beast Mastery] add a suggestion if people use Multishot on ST bosses

Streammz:
Implementation for the following issue: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/1145

This is how it looks like: https://i.imgur.com/t3p8ngF.png

Currently, it checks for cases where multishot has been used (and thus beast cleave has been applied), and no beast cleave damage has been done within that 4 second Beast Cleave window. 

It adds an _avarage_ suggestion for 1-3 multi-shot casts, and a _major_ suggestion for any more than that. These are just numbers I came up with, so if you feel these numbers should be altered, feel free to do so/recommend better ones!

This is my first contribution to WoWAnalyzer. If there's anything important I should know or something I may have forgotten, please let me know 🙂

### DIFF
--- a/src/CONTRIBUTORS.js
+++ b/src/CONTRIBUTORS.js
@@ -507,3 +507,8 @@ export const jos3p = {
   github: 'jos3p',
   discord: 'jos3p#9746',
 };
+export const Streammz = {
+  nickname: 'Streammz',
+  github: 'Streammz',
+  discord: 'Streammz#9999',
+};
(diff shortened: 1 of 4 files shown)


Streammz:
another decent example log: https://www.warcraftlogs.com/reports/RrLzYfna3yMhV7Gd#translate=true&fight=1&type=damage-done&source=28 (5 casts, where only 1 was actually cleaving)

Issue #2535, opened 2018-10-29
[Hunter] Implement damage contribution of all hunter azerite traits


Pewtro:
All hunter azerite traits with a damage component should get updated to show damage contribution, much like was done for Feral Druids here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/druid/feral/modules/azeritetraits/WildFleshrending.js

Traits that need this implementation: 
## 8.0 traits
- [x] Feeding Frenzy - done in #2540
- [ ] Focused Fire - Rapid Fire contribution
- [ ] Steady Aim - Aimed Shot contribution
- [ ] Latent Poison - Latent Poison damage
- [ ] Venomous Fangs - Pet basic attack contribution
- [ ] Wilderness Survival - Mongoose/Raptor contribution
- [x] Wildfire Cluster - Wildfire Bomb contribution
- [ ] Serrated Jaws - Kill command contribution
- [x] Rapid Reload - Multi Shot contribution

## 8.1 traits
- [ ] Dire Consequences - Kill Command contribution
- [ ] Surging Shots - Rapid Fire contribution

Pewtro:
I don't think the damage implementation for these is that important and seeing as the expansion is coming to a close, I think this issue also deserves to be closed...

PR #2540, opened 2018-10-29
[Beast Mastery] Modify the Feeding Frenzy module to additionally track the damage contribution of the trait

Streammz:
As was requested as one of the traits via #2535, I have modified the Feeding Frenzy module to include the damage contribution of the azerite trait. It does this by calculating how much of the damage portion of a Barbed Shot tick was actually contributed via the Feeding Frenzy increased damage. I've verified the accuracy of this contribution using an estimated damage number, which ended up being accurate in nearly all cases I've checked, only being off by 0 or 1 damage in most cases.

Screenshot of how it looks like: https://i.imgur.com/7f4LRx5.png

It works similar to the Wild Fleshrending trait that Feral Druids have; https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/druid/feral/modules/azeritetraits/WildFleshrending.js

It comes with a minor flaw that the attack power of the player can only be retrieved when casting spells, and thus there may be cases where for example an agility potion runs out between casting a spell and the Barbed Shot debuff ticking. For that reason I have included a tilde ( ~ ) just to ensure that the users are aware this may not be 100% accurate. On the multiple logs that I've checked this, this hardly ends up affecting the accuracy of the final damage numbers. In most examples, it was inaccurate by less than 0.1%.

### DIFF
--- a/src/parser/hunter/beastmastery/modules/spells/azeritetraits/FeedingFrenzy.js
+++ b/src/parser/hunter/beastmastery/modules/spells/azeritetraits/FeedingFrenzy.js
@@ -1,12 +1,19 @@
 import React from 'react';
 import Analyzer from 'parser/core/Analyzer';
-import { formatNumber } from 'common/format';
+import StatTracker from 'parser/shared/modules/StatTracker';
+import { formatNumber, formatPercentage } from 'common/format';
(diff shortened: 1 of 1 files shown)


Pewtro:
Kicking travis.

PR #2543, opened 2018-10-29
[Mage] Fix some capitalization and add missing ability buffids

Dambroda:
Fixes a small number of capitalization issues and adds some missing buffids so that they appear in the timeline


Dambroda:
That's what I thought originally, but I think Zero said that it's only spell names and the first word of the requirement; looking at holy pally that's how he does it as well

PR #2550, opened 2018-10-30
[Restoration Shaman] Unleash Life

niseko:
Reworks Unleash Life a bit. Motivation comes from needing to find out the healing distribution between the spells for a future module, and also to improve accuracy, as that wasn't quite there (riptide mostly).

The player_heal function is definitely a bit overloaded, but with how UnL works it does make other solutions hard (unless I'm missing something).
Casts & Heals essentially go through the same logic, as the combat log order for that is unreliable. Either standalone work fine.

I ran into issues using applybuff events for Riptide, with situations like this where Riptide isn't affected by Unleash Life (already consumed by Surge), but the `hasBuff(UnleashLife)` function will always return true here: 
image

Added a tooltip for more detail on the breakdown: 
image
https://www.warcraftlogs.com/reports/pCTvHnmtLR1wjDQa/#fight=50&source=12

### DIFF
--- a/src/parser/shaman/restoration/modules/talents/UnleashLife.js
+++ b/src/parser/shaman/restoration/modules/talents/UnleashLife.js
@@ -17,7 +17,7 @@ import CooldownThroughputTracker from '../features/CooldownThroughputTracker';
 const UNLEASH_LIFE_HEALING_INCREASE = 0.45;
 const BUFFER_MS = 200;
 const UNLEASH_LIFE_DURATION = 10000;
-const debug = true;
+const debug = false;
 
 const CHART_SIZE = 75;
 
@@ -52,7 +52,7 @@ class UnleashLife extends Analyzer {
   };
 
   unleashLifeCasts = 0;
-  unleashLifeRemaining = 0;
+  unleashLifeRemaining = false;
   unleashLifeHealRemaining = 0;
 
   buffedChainHealTimestamp = null;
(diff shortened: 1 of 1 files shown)


src/parser/shaman/restoration/modules/talents/UnleashLife.js
```diff
@@ -55,128 +66,104 @@ class UnleashLife extends Analyzer {
...
-    const hasUnleashLife = this.selectedCombatant.hasBuff(SPELLS.UNLEASH_LIFE_TALENT.id, event.timestamp, BUFFER_MS, BUFFER_MS);
+    if (this.unleashLifeRemaining && (this.lastUnleashLifeTimestamp + UNLEASH_LIFE_DURATION) <= event.timestamp) {
+      this.unleashLifeRemaining = false;
+      debug && console.log("Cast Timed out", event.timestamp);
+      return;
+    }
 
-    if(!hasUnleashLife) {
+    const hasUnleashLife = this.selectedCombatant.hasBuff(SPELLS.UNLEASH_LIFE_TALENT.id, event.timestamp, BUFFER_MS, BUFFER_MS);
+    if (!hasUnleashLife) {
```
jjs451: I'm confused about why another variable is needed.  You're tracking if the buff is up with unleashLifeRemaining, right?  What's requiring you to use a second check for if the buff is up?

src/parser/shaman/restoration/modules/talents/UnleashLife.js
```diff
@@ -55,128 +66,104 @@ class UnleashLife extends Analyzer {
...
+      if (this.healingBuff[spellId]) {
+        this.healingBuff[spellId].castAmount += 1;
+        this.unleashLifeRemaining = false;
+        debug && console.log("Cast:", spellId);
       }
     }
   }
 
-  on_byPlayer_applybuff(event) {
+  on_byPlayer_removebuff(event) {
```
jjs451: Mostly unrelated question, is there no event for removing the unleash life buff?

src/parser/shaman/restoration/modules/talents/UnleashLife.js
niseko: Must be leftover from the old solution. It essentially did nothing and I removed it.

PR #2551, opened 2018-10-30
[Frost Mage] Glacial Spike module rewrite

Dambroda:
This PR is a code-logic and presentational rewrite of the Frost Mage Glacial Spike module. It aims to resolve one known issue with the module:

Utilization percentages above 100%
image

It also aims to present the information in an easier to understand way for viewers.

### Before
image
image

### After
image
image
image
image


Sharrq:
I'm looking at this from my phone so I can't properly review it right now, but the pet freeze should be in the shatter debuffs, right?

Sharrq:
Looks good to me.

I didnt know about that meta thing for highlighting stuff on the timeline, must be newish ..... I remember asking for something like that shortly after the timeline was created. Probably after blizzcon ill go back and add that to some of the other modules as well as fire/arcane.

PR #2553, opened 2018-10-30
[Beast Mastery] Make the Global Cooldown during AOTW use proper haste values.

Streammz:
Currently BM is showing an error on the Timeline for users that the Global Cooldown has not been configured properly, and after investigating, I found that the haste values used during our cooldown AOTW use the wrong numbers with certain buffs up, such as time-warp and bloodlust.

The reason this happens, is because currently it is using the StatTracker to calculate the current haste, but buffs like Time Warp is a percentage based buff, and thus not representable using a stat tracker. Arguably this may even be a different issue itself.

To ensure the `GlobalCooldown.js` module for Beast Mastery is consistent with the shared `CoreGlobalCooldown` module, I've changed it to instead use `this.haste.current` instead, which the shared module also uses.


Streammz:
@MartijnHols Thanks for pointing that out, I've missed that completely. It has been added!

PR #2559, opened 2018-10-31
[Frost Mage] Sort spells on timeline

Dambroda:
The diff gets really messed up when you edit values like these, sorry! Sorted by estimated cast count / cooldown length, rather than the default sorting

Before:
image

After:
image

### DIFF
--- a/src/parser/mage/frost/modules/features/Abilities.js
+++ b/src/parser/mage/frost/modules/features/Abilities.js
@@ -13,24 +13,7 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
-      },
-      {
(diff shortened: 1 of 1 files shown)


Sharrq:
I'm at the airport currently so I won't get a chance to look over this very closely, but from what I can tell the spell order is fine. As far as I can tell that's the only change here but the way git marked it up isn't helping

Sharrq:
I approved it but I'd rather someone else merge it after confirming I didn't miss any changes other than the timeline sort setting changes. Since I'm on mobile in an airport

Pewtro:
I looked earlier and it was fine, just didn't want to merge without a mage looking at it

Issue #2529, opened 2018-10-29
[Beast Mastery] Remaining talent implementation


Pewtro:
While some of these talents have some minor implementation, if they're unticked it means I think there can be done more for them in general - such as statisticboxes, suggestions or more. 

## Level 15
- [x] Killer Instinct
- [x] Animal Companion
- [x] Dire Beast
## Level 30
- [x] Scent of Blood
- [x] One with the Pack
- [x] Chimaera Shot
## Level 45
- [x] Trailblazer
- [x] Natural Mending
- [x] Camouflage
## Level 60
- [x] Venomous Bite
- [x] Thrill of the Hunt
- [x] A Murder of Crows
## Level 75
- [x] Born To Be Wild
- [x] Posthaste
- [x] Binding Shot
## Level 90
- [x] Stomp
- [x] Barrage
- [x] Stampede
## Level 100
- [x] Aspect of the Beast
- [x] Killer Cobra
- [x] Spitting Cobra

PR #2566, opened 2018-11-01
[Beast Mastery] Implement talent statistics and suggestions for the Stampede module

Streammz:
In regards to #2529, this adds more statistics and suggestions for the _Stampede_ level 90 talent.

First it adds an extra expandable statistics sheet in the _Talents_ section:
image
image

Second, it adds a suggestion regarding inefficient Stampede usage, where not all potential stampede hits actually hit the target:
image

The already existing damage contribution is untouched.

### DIFF
--- a/src/parser/hunter/beastmastery/modules/talents/Stampede.js
+++ b/src/parser/hunter/beastmastery/modules/talents/Stampede.js
@@ -2,22 +2,55 @@ import React from 'react';
 
 import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
+import STATISTIC_ORDER from 'interface/others/STATISTIC_ORDER';
 import Analyzer from 'parser/core/Analyzer';
 import ItemDamageDone from 'interface/others/ItemDamageDone';
 import StatisticListBoxItem from 'interface/others/StatisticListBoxItem';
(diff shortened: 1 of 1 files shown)


src/parser/hunter/beastmastery/modules/talents/Stampede.js
```diff
@@ -2,30 +2,143 @@ import React from 'react';
...
+    this.damage += damage;
+
+    if (this.currentCast !== null) {
+      this.currentCast.hits++;
+      this.currentCast.damage += damage;
+    }
+  }
+
+  on_finished() {
+    this.averageHits = this.hits / this.casts.length / STAMPEDE_POTENTIAL_HITS;
```
emallson: this should *probably* be a getter

src/parser/hunter/beastmastery/modules/talents/Stampede.js
```diff
@@ -2,30 +2,143 @@ import React from 'react';
...
+    this.casts.forEach((cast) => {
+      cast.averageHits = cast.hits / STAMPEDE_POTENTIAL_HITS;
```
emallson: this *could* be a getter too, but idk how people feel about getters on object literals

PR #2571, opened 2018-11-01
[Core] Pre-pull events normalizer

Dambroda:
**_Please do not merge until 2-3 approving reviews!_**

As mentioned in #2368, I have tried to adapt this to be available to all specializations.

### The Original Problem
The BuffApply normalizer fabricates applybuff events for buffs that existed on the player before combat. It does this by searching for buffremove events that never had a buffapply event. It works great, but it doesn't actually give us a cast event (which means that cooldowns dont get started, it isnt counted by cast efficiency, etc). This leads to problems like:
image
image

You can see the buff was applied prepull, but the spell cast wasn't counted. 

### The Solution
This normalizer searches for all of the fabricated buffapply events created by the other normalizer, and fabricates cast events for them. It also follows the same logic as that normalizer and searches for damage events that didn't have a corresponding cast event, and fabricates a cast event for it. For specs that use any prepull cast events .classResources, important to note that the fabricated events use the first real cast events class resources and applies it to all of the fabricated events. If there's a chance that's wrong, you can check for .prepull/.fabricated

Importantly, this normalizer also guesstimates prepull timestamps to avoid as many cooldown errors possible. 

### Maintainability
In order to have one of your cooldowns checked, you need to set the _buffSpellId: int_ or _damageSpellIds: [int..]_ property of the Ability. _TODO: Find a way to filter out spells we don't want tracked pre-pull_

### Spec Changes
Some specializations were already solving parts of this issue themselves by checking for applybuff events with .prepull set, and then starting cooldowns manually. I've removed those and labeled their maintainers to double check that this wont break anything. Just to confirm some other things, I looked at some hunter logs:
image

### Moving Forward
There are a couple of things that I'd like to do to make this normalizer more consistent and fix more bugs. They are ordered by feasibility:
1. Find a better way to filter what events we care about fabricating prepull events for. I was thinking check for ability.castEfficiency.suggestion and only put in those, but I saw a few counterexamples straight away when looking at other specs. Looking for other ideas.
2. Handle buffs that apply to other players (for example, Prayer of Mending and other heals)
3. Merge this module with BuffApply, we basically use the same logic for damage spells here that it uses for buffs (damage without cast vs buffremove without buffapply), which also means we could match the buffapply timestamps to the cast timestamps.
4. More accurately fabricate cast events (spells with a cast time get a begincast event instead of only a cast event); the problem here is I don't believe I actually have any way to accurately determine cast times unless we go through and add it to Abilities.


src/parser/hunter/marksmanship/modules/spells/Trueshot.js
```diff
@@ -28,30 +28,8 @@ class Trueshot extends Analyzer {
...
-    //checks if we had a prepull trueshot, in which case the firstCast done symbolises our starting focus of that one trueshot
-    if (this.prepullTrueshots > 0 && this.startFocusForCombatant === 0) {
-      event.classResources.forEach(classResource => {
-        if (classResource.type === RESOURCE_TYPES.FOCUS.id && classResource.amount > this.startFocusForCombatant) {
-          this.startFocusForCombatant += classResource.amount;
-          this.accumulatedFocusAtTSCast += this.startFocusForCombatant;
-        }
-      });
-    }
     const spellId = event.ability.guid;
```
Pewtro: Trueshot precast focus amount needs to stay in the modules.

Dambroda: It uses the same logic as your module (copies the first cast event to guess the prepull focus amount). The amount should be unchanged (see screenshot above)

src/parser/shared/normalizers/PrePullCooldowns.js
```diff
@@ -0,0 +1,172 @@
...
+     * to deduce the cast time. This would require ability info about durations.
+     */
+    // Working backwards, assume worst-case GCD timestamps
+    let totalGCD = 0;
+    for (let i = prepullCasts.length - 1; i >= 0; i -= 1) {
+      const event = prepullCasts[i];
+      const ability = this.abilities.getAbility(event.ability.guid);
+      const gcd = (ability && ability.gcd && ability.gcd.base) || 1500;
+      totalGCD += gcd;
+      event.timestamp = firstTimestamp - totalGCD;
```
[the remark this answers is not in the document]
Dambroda: Because this is a normalizer I don't have access to a lot of those properties that rely on analyzers (haste, gcd, etc.). That was one of the hurdles that led to a delay in this PR, deciding what to do about those issues. For now I decided it was best to just assume worst-case GCD and use base *if* it was there. But now that you've pointed it out I've forgotten that base can be a func as well. So I guess I should just check for both and see if one isn't a func, and use that?

Dambroda:
Thanks for reviewing! Yes adding begincast events was actually in my 'moving forward' section, haha. That would be really nice. And then also I want to adjust the timeline to show these fabricated events (but formatting it clearly). And yeah, only problem again is getting the info into Abilities. We need to find a better way to do this...

PR #2573, opened 2018-11-01
[Restoration Shaman] Ancestral Vigor breakdown

niseko:
Adds more detail to the AV module, partially to see if some of those "lives saved" were shortly after another on the same player (where he'd already be dead). I might filter that in the future if it happens too often in logs.

image
image
https://www.warcraftlogs.com/reports/vfktbD3mCaNR7Q8M/#fight=69&source=28

Thanks to whoever made the lazy-load statistic box expandable


emallson:
> Thanks to whoever made the lazy-load statistic box expandable

You're welcome :)

PR #2578, opened 2018-11-02
Mistweaver Mana Efficiency

awlego:
I implemented the mana efficiency tab for Mistweaver. I need to update the change log for these changes, but I can do that as part of any feedback commits I make.

Log 1 example:
screen shot 2018-11-02 at 1 37 17 am

Log 2 example:
screen shot 2018-11-02 at 1 40 05 am
 You can see in the above screenshot that I also tweaked the Env:Viv cast ratio images to match the sizes of the rest of statistics boxes. If preferred I can split those changes out of this PR.


src/parser/core/healingEfficiency/HealingEfficiencyTracker.js
```diff
@@ -44,6 +44,10 @@ class HealingEfficiencyTracker extends Analyzer {
     spellInfo.damageDone = ability.damageEffective || 0;
     spellInfo.damageAbsorbed = ability.damageAbsorbed || 0;
 
+    spellInfo.manaSpent = this.manaTracker.spendersObj[spellId] ? this.manaTracker.spendersObj[spellId].spent : 0;
+    spellInfo.manaPercentSpent = spellInfo.manaSpent / this.manaTracker.spent;
```
niseko: this line needs to stay where it is, after getCustomSpellStats is called, otherwise mana percentages will be wrong for spells where you changed their expenditure

awlego: Good catch.

src/parser/monk/mistweaver/modules/spells/Vivify.js
```diff
@@ -31,6 +36,12 @@ class Vivify extends Analyzer {
 
   on_byPlayer_heal(event) {
     const spellId = event.ability.guid;
+
+    // todo .hasBuff should use the sourceID so that things work with multiple mistweavers in the group
```
emallson: i don't see `hasBuff` being called here. is this todo out of date? or does this only work correctly with a single MW in the group?

awlego: Todo is out of date. It has been removed. I have also tested with a log with two MWs just to make sure.

awlego:
Any more changes requested? I think I have addressed all of the issues niseko and emallson brought up.

emallson:
@awlego oshit i meant to approve this. sorry! merging momentarily

PR #2590, opened 2018-11-05
[Hunter] A Murder of Crows fixing

Pewtro:
Old (without associated bugfix in this PR which would push it to 26): 
image

New: 
image

Log: https://wowanalyzer.com/report/GrzhBC6MPVZ8cbQp/28-Mythic+Fetid+Devourer+-+Kill+(4:41)/19-Raicha/abilities

Fixes a bug where a player wouldn't be doing any damage in the timespan where crows could reset, so it was never registered as a reset.

Moves Crows into its own module and ensure the maxCasts prop is accurate.

### DIFF
--- a/src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
+++ b/src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
@@ -7,6 +7,7 @@ import SpellUsable from 'parser/shared/modules/SpellUsable';
 import SpellLink from 'common/SpellLink';
 import ItemDamageDone from 'interface/others/ItemDamageDone';
 import StatisticListBoxItem from 'interface/others/StatisticListBoxItem';
+import Abilities from 'parser/shared/modules/Abilities';
 
 /**
  * Summons a flock of crows to attack your target over the next 15 sec. If the target dies while under attack, A Murder of Crows' cooldown is reset.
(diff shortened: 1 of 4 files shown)


src/parser/hunter/shared/modules/talents/AMurderOfCrows.js
```diff
@@ -17,26 +18,61 @@ import StatisticListBoxItem from 'interface/others/StatisticListBoxItem';
...
+    });
+  }
+
+  checkForReset(event) {
+    // Checks if we've had atleast 1 damage tick of the currently applied crows, and checks that crows is in fact on cooldown.
+    if (this.lastDamageTick && this.spellUsable.isOnCooldown(SPELLS.A_MURDER_OF_CROWS_TALENT.id)
+      // Checks whether the current damage event happened while the time passed since crows application is less than the crows duration
+      && this.applicationTimestamp && event.timestamp < this.crowsEndingTimestamp
+      // Checks to see if more than 1 second has passed since last tick
+      && event.timestamp > this.lastDamageTick + CROWS_TICK_RATE + MS_BUFFER) {
```
joshinat0r: Not quite how reliable this is if the ticks are scaling with haste (esp when thinking about negative haste like we had in NH).
But it'll fix more than it breaks

PR #2593, opened 2018-11-05
[Hpriest] Adding renew cast suggestion

TJMoats:
Renew is a pretty bad spell for priests. One of the most common improvements you can make as a player is to limit the number of renews that you cast. Renew should only be cast when moving, and when you have no other instant cast spells. This update tries to validate if you are casting renew well. This is a bit hard to do as it's hard to tell if a player is moving.

screen shot 2018-11-05 at 10 33 02 am

### DIFF
--- a/src/parser/priest/holy/modules/spells/Renew.js
+++ b/src/parser/priest/holy/modules/spells/Renew.js
@@ -198,7 +198,7 @@ class Renew extends Analyzer {
             .actual(<>
               You used Renew {this.badRenews} times when another spell would have been more productive.
               Renew is one of the least efficient spells Holy Priests have, and should only be cast when moving with no other instants available.</>)
-            .recommended(`Two per minute is recommended, except for movement heavy fights.`);
+            .recommended(`Two or less per minute is recommended, except for movement heavy fights.`);
         }
       );
   }
--- a/src/parser/priest/holy/modules/spells/Renew.js
+++ b/src/parser/priest/holy/modules/spells/Renew.js
@@ -1,17 +1,30 @@
+import React from 'react';
+
 import SPELLS from 'common/SPELLS';
 import Analyzer from 'parser/core/Analyzer';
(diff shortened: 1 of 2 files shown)


src/parser/priest/holy/modules/spells/Renew.js
```diff
@@ -110,6 +147,61 @@ class Renew extends Analyzer {
...
+      const sanctifyOnCooldown = this.spellUsable.isOnCooldown(SPELLS.HOLY_WORD_SANCTIFY.id);
+      const serenityOnCooldown = this.spellUsable.isOnCooldown(SPELLS.HOLY_WORD_SERENITY.id);
+      let cohOnCooldown = true;
+      if (this.selectedCombatant.hasTalent(SPELLS.CIRCLE_OF_HEALING_TALENT.id)) {
+        cohOnCooldown = this.spellUsable.isOnCooldown(SPELLS.CIRCLE_OF_HEALING_TALENT.id);
+      }
+      if (sanctifyOnCooldown && serenityOnCooldown && cohOnCooldown) {
+        this.goodRenews += 1;
+      } else {
+        this.badRenewReason.betterspell = (this.badRenewReason.betterspell | 0) + 1;
```
emallson: This should be `||` right? unless there is a reason to use bitwise or that I'm unaware of?

src/parser/priest/holy/modules/spells/Renew.js
```diff
@@ -110,6 +147,61 @@ class Renew extends Analyzer {
...
+      }
+    }
+    this.badRenewReason.stationary = (this.badRenewReason.stationary | 0) + 1;
```
emallson: same as above

TJMoats: Fixed

src/parser/priest/holy/modules/spells/Renew.js
TJMoats: Fixed

PR #2595, opened 2018-11-06
[Blood Death Knight][All tanks] Deaths Caress update + softmit-check

joshinat0r:
**Blood Death Knight**
Updates the Death's Caress-module to account for distance between the player and the target and checks if other ranged spells where available (DnD or BD, since those to more than DC).

The 12 yrd limit right now is kind of arbitrary. I had some distances that were bigger than 30yrds (up to 37yrd), which should be due to travel-time of the spell and the target's hitbox. But I think 12yrd should be a good place to start since it's still in the 8yrd range (+- boss hitbox) for BB.
Will adjust that number according to feedback. But it's for sure more accurate than the current implementation that just flags every single cast as bad.

**Soft Mitigation Check**
Now flags checks that were parried, dodged, missed or immuned as passed.

### DIFF
--- a/src/parser/deathknight/blood/modules/features/Checklist.js
+++ b/src/parser/deathknight/blood/modules/features/Checklist.js
@@ -96,15 +96,16 @@ class Checklist extends CoreChecklist {
             check: () => this.runeTracker.suggestionThresholdsEfficiency,
           }),
           new Requirement({
-            name: <><SpellLink id={SPELLS.MARROWREND.id} /> efficiency</>,
+            name: <><SpellLink id={SPELLS.MARROWREND.id} /> Efficiency</>,
             check: () => this.marrowrendUsage.suggestionThresholdsEfficiency,
           }),
           new Requirement({
-            name: <><SpellLink id={SPELLS.RUNE_STRIKE_TALENT.id} /> efficiency</>,
+            name: <><SpellLink id={SPELLS.RUNE_STRIKE_TALENT.id} /> Efficiency</>,
             check: () => this.runeStrike.cooldownReductionThresholds,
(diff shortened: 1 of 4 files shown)


src/parser/deathknight/blood/modules/features/Checklist.js
```diff
@@ -96,15 +96,16 @@ class Checklist extends CoreChecklist {
             check: () => this.runeTracker.suggestionThresholdsEfficiency,
           }),
           new Requirement({
-            name: <><SpellLink id={SPELLS.MARROWREND.id} /> efficiency</>,
+            name: <><SpellLink id={SPELLS.MARROWREND.id} /> Efficiency</>,
```
Gebuz: imo no capital letter is correct here, since it isn't a title or name.

joshinat0r: I capitalized it because it looked weird when half of the `Efficiency`'s is lowercase and the others are uppercase.
And it appears like blood isn't the only one that's using it both ways.

image
image

joshinat0r:
But isn't the check supposed to check if you failed a specific check? And if you dodged it you passed it, sure not because you played it well but because of rng. But you passed the check nonetheless.

At least my opinion, saying the player failed it even if he accidentally avoided the damage is wrong because he didn't get hit. Not even sure how many tank-mechanics are actually parry/dodge-able

Gebuz:
My understanding is that the purpose of the softmitigation check is whether you played well or not, not whether you died or survived. Getting lucky and dodging when you should've failed because of no AM doesn't make sense to me to mark as a job well done when it isn't.

PR #2602, opened 2018-11-06
Resto Druid - Fixed ordering of WG/Flourish events

buimichael:
A bug was found by a user where a Flourish and WG cast happening at the exact same timestamp did not register in the flourish module.
Log: https://www.warcraftlogs.com/reports/YF9QbHPdWB8V4rKn#fight=last&type=summary&start=11672525&end=11690572&source=5&view=events

This is the events from WCL:
img

The issue here is that the applybuff events from WG comes after the cast and applybuff of flourish, making it seem as the player did not extend Wild Growh with Flourish. Below are two images, the image above shows current live version, the log and image above is referencing the 4th cast, this shows 4 HoTs available, but looking at the event logs it should be at least 6 (cause WG was extended).

The image below is the result of these changes.
[Screenshot from Gyazo](https://gyazo.com/b45b038f05da1b716e7dc359bbb90424)

Live version suggests that 4/6 WGs were extended by Flourish, looking closely in the WCL cast events we can see clearly that all 6 Flourishes were extending WG.

I've extended the WildGrowth.js normalizer to move flourish event forwards and tests that should cover these scenarios.

### DIFF
--- a/src/parser/druid/restoration/CHANGELOG.js
+++ b/src/parser/druid/restoration/CHANGELOG.js
@@ -4,7 +4,12 @@ import { Yajinni, blazyb, fel1ne, Qbz } from 'CONTRIBUTORS';
 
 export default [
   {
-    date: new Date('2018-10-19'), 
+    date: new Date('2018-11-06'),
+    changes: 'Fixed a minor bug with Flourish module and ordering of event being skewed.',
+    contributors: [blazyb],
+  },
+  {
+    date: new Date('2018-10-19'),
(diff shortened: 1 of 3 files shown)


buimichael:
Edit: Noticed additional use cases where event ordering was fucked up, added a MS_BUFFER variable on 100ms to cover for this with additonal tests to cover these use cases.
[Screenshot from Gyazo](https://gyazo.com/c56935a68bd6d87c0cc9306dccd8d665)
[Screenshot from Gyazo](https://gyazo.com/572fa84b3a00b14ae66ca1201f511190)
[Screenshot from Gyazo](https://gyazo.com/a1c62ca9ff8a99637b81cf8270a2e26d)

PR #2610, opened 2018-11-07
[Arms] Modified Mortal Strike analysis for Execution phase

matardarix:
Fixed Mortal Strike castEfficiency with execution phase (as it shouldn't be cast during this, except if you have Executioner's Precision Trait).

Before: 
image

Now: 
image


Pewtro:
I don't think this is the right way of doing this. 

Immediate issues I could think of:
Ghuun is ~25% of fightduration sub 20%. 
Taloc starts at 60%
Fetid takes increased damage sub 50% making that section of the fight significantly shorter.

Not to mention anyone using hero at various spots will change this significantly.

matardarix:
Right way would be to have the total durations of execute phases during a fight to reduce this i think, i'll open an issue on this

Pewtro:
I thought more of this - what you could is keep the castefficiency low (or nonexisting) then make a statistic module showing x good casts / y possible casts or simply z amount of bad casts

Gebuz:
I think one way to go about it would be to find the amount of time spent hitting a target outside of execute range, and the percent of that spent with Mortal Strike off cooldown. I'm not sure exactly how best to do this, since only the dmg events have the HP of the target. 

Might be good enough to just start an execute interval as soon as there is a dmg event to a target in execute range, and then stop it on the first damage event to a target not in execute range after a minimum of X seconds has passed since the last hit on a target in execute range.

I don't know how you'd then get time spent with MS off cooldown for those intervals, but if you just want something simple I guess you could start with just looking at average cooldown and amount of casts during them.

Then you can throw that in a statistic box with "Mortal Strike efficiency outside execute" or something like that.

matardarix:
Currently the main issue is that the checklist display the cast eficiency of mortal strike for the entire fight and coud mislead the player 
image

So i was thinking of something that could look like this instead
image

It uses the maxCasts but outside the Abilities file to avoid the problems reported by Gebuz. But to be really accurate it should be using a duration instead for the maxCasts (I just re-used the bad way used in this PR).

matardarix:
So here's how it looks now: 
When you don't have Executioner's Precision azerite traits: 
image

And with the trait: 
image

The maxCasts used in the first image are calculated based on the execution phases duration of the fight to be more accurate.

Gebuz:
Looks good, just one thing I'd like to see added is a buffer window for how often you can switch in/out of execute to prevent it from constantly switching when you are hitting multiple targets (for instance hitting an add in execute range with sweeping strikes hitting the boss not in execute range). Something like save timestamp on every dmg event on a target in execute range, and do not leave execute range unless it's a few seconds since that timestamp.

Gebuz:
yep!

Gebuz:
Oh sorry, I meant since last dmg event in execute range, so would have to save a timestamp on every event in execute range

```
if(this.isTargetInExecuteRange(event)) {
  this.lastExecuteTimestamp = event.timestamp;
}
if(!this.isTargetInExecuteRange(event) && this.isExecPhase && event.timestamp > this.lastExecuteTimestamp + EXECUTE_BUFFER) {
  ...
}
```

matardarix:
Oh fair enough :sweat_smile: should be good

PR #2611, opened 2018-11-07
[Spriest] Chorus of Insanity azerite trait

TJMoats:
screen shot 2018-11-07 at 9 07 23 am

### DIFF
--- a/src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
+++ b/src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
@@ -1,0 +1,123 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS/index';
+import { formatPercentage } from 'common/format';
+import { calculateAzeriteEffects } from 'common/stats';
+import Analyzer from 'parser/core/Analyzer';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'interface/others/TraitStatisticBox';
+
+const chorusOfInsanityStats = traits => Object.values(traits).reduce((obj, rank) => {
+  const [crit] = calculateAzeriteEffects(SPELLS.CHORUS_OF_INSANITY.id, rank);
+  obj.crit += crit;
+  return obj;
+}, {
+  crit: 0,
+});
+
+export const STAT_TRACKER = {
+  crit: combatant => {
+    return chorusOfInsanityStats(combatant.traitsBySpellId[SPELLS.CHORUS_OF_INSANITY.id]).crit;
+  },
+};
+
+/**
+ * Chorus of Insanity
+ * When Voidform ends, gain 32 Critical Strike for each stack of Voidform. This effect decays every 1 sec.
+ *
+ * Example log: /report/VAZhnPRdxL12w3fG/1-Mythic+G'huun+-+Kill+(8:43)/6-Rxc
+ */
+class ChorusOfInsanity extends Analyzer {
+  crit = 0;
+  totalCrit = 0;
+  lastTimestamp = 0;
+
+  chorusProcs = 0;
+  currentStacks = 0;
+
+  pendingStack = false;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.CHORUS_OF_INSANITY.id);
+    if (!this.active) {
+      return;
+    }
+
+    const { crit } = chorusOfInsanityStats(this.selectedCombatant.traitsBySpellId[SPELLS.CHORUS_OF_INSANITY.id]);
+    this.crit = crit;
+  }
+
+  on_byPlayer_applybuff(event) {
+    this.handleStacks(event);
+  }
+
+  on_byPlayer_applybuffstack(event) {
+    this.handleStacks(event);
+  }
+
+  on_byPlayer_removebuff(event) {
+    this.handleStacks(event);
+  }
+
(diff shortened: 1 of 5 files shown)


src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
```diff
@@ -0,0 +1,123 @@
...
+      // This check adds the first tick.
+      if (this.pendingStack) {
+        this.totalCrit += (this.currentStacks + 1) * this.crit * uptimeOnStack;
+        this.pendingStack = false;
+      }
+
+      this.totalCrit += this.currentStacks * this.crit * uptimeOnStack;
+    }
+
+    if (event.type === "applybuff") {
```
Dambroda: Small one but just for style convention should be single quotes (and removebuff as well)

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
```diff
@@ -0,0 +1,123 @@
...
+
+    this.lastTimestamp = event.timestamp;
+  }
+
+  get uptime() {
+    return this.selectedCombatant.getBuffUptime(SPELLS.CHORUS_OF_INSANITY_BUFF.id) / this.owner.fightDuration;
+  }
+
+  get averageCrit() {
+    return (this.totalCrit / this.owner.fightDuration).toFixed(0);
```
Dambroda: Couldn't you use _getStackWeightedBuffUptime_ to pretty much completely avoid having to do any of these calculations yourself?

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
```diff
@@ -0,0 +1,123 @@
...
+
+  statistic() {
+    return (
+      <TraitStatisticBox
+        position={STATISTIC_ORDER.OPTIONAL()}
+        trait={SPELLS.CHORUS_OF_INSANITY.id}
+        value={`${this.averageCrit} average Crit`}
+        tooltip={`
+          ${SPELLS.CHORUS_OF_INSANITY.name} grants <b>${this.crit} crit per stack</b><br/>
+          You procced <b>${SPELLS.CHORUS_OF_INSANITY.name} ${this.chorusProcs} times</b> with an uptime of ${formatPercentage(this.uptime)}%.
```
Dambroda: It's only stylistic, but arent the numbers usually bolded and the names not (only in second line here):

Currently:
You procced **Corus of Insanity X times** with an uptime of X%.

Standard (imo):
You procced Corus of Insanity **X times** with an uptime of **X%**.

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
TJMoats: getStackWeightedBuffUptime doesn't calculate the initial stack application correctly. There is no initial applybuffstack event, so I need the custom logic that I have here.

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
TJMoats: Fixed

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
TJMoats: Fixed

src/parser/priest/shadow/modules/spells/azeritetraits/ChorusOfInsanity.js
Dambroda: Wouldn't it be better to fabricate the buff update event in that case?

TJMoats: Do you have an example of how to do that? and I don't have the number of stacks until the first removebuffstack. Would I be able to fabricate an event a second before the removebuffstack event?

Dambroda: Oh, I think in this case you could call something like:
```javascript
    this.owner.fabricateEvent({
      ...event, // copies everything from the removebuffstack event except below
      type: 'applybuffstack',
      stack: event.stack + 1,      
      timestamp: /*timestamp copied from the applybuff event; maybe add 1ms?*/,
    }, event);
```

And I'd keep your existing math in there as well to test and ensure you get the same results from both methods as well

TJMoats: I think I may have fabricated the event properly. The numbers are within margin of error.

Dambroda: It looks like you may be missing the source/targets ids which may lead to the non-matching numbers

TJMoats: It was accurate to within .5 crit of what I had for the final number. Good enough for me

Dambroda: Could you add the ids? They are needed for pretty much all all event filtering I think

TJMoats: I just added the event

PR #2615, opened 2018-11-07
[Core / Hunter] AverageTargetsHit and One With The Pack + Stomp 

Pewtro:
Fixes a bug with Born to be wild affecting CDR. 
Fixes formatting with a Survival Azerite trait

Adds a core average targets hit module (unique targets hit is WIP and will be in next PR) 
Implements an actual statistic for Stomp together with the newly added core average targets hit module.
Implements One With The Pack talent. 

/report/A89tf3NJMqCXhKFV/1-Normal+Taloc+-+Kill+(5:32)/15-Shadowmikey
image
image

/report/A89tf3NJMqCXhKFV/5-Normal+Vectis+-+Kill+(5:29)/15-Shadowmikey
image

### DIFF
--- a/src/interface/others/AverageTargetsHit.js
+++ b/src/interface/others/AverageTargetsHit.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class AverageTargetsHit extends React.PureComponent {
+  static propTypes = {
+    casts: PropTypes.string.isRequired,
+    hits: PropTypes.string.isRequired,
+    approximate: PropTypes.bool,
+  };
+
+  render() {
+    const { casts, hits, approximate } = this.props;
+    const averageHits = (hits / casts).toFixed(1);
+    return (
+      <>
+        {approximate && '≈'}{averageHits}{' average '}{averageHits > 1 ? 'targets hit' : 'target hit'}
+      </>
+    );
+  }
+}
+
+export default AverageTargetsHit;
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,11 @@ import Contributor from 'interface/contributor/Button';
 
 export default [
   {
+    date: new Date('2018-11-09'),
+    changes: 'Added an AverageTargetsHit module for general usage.',
+    contributors: [Putro],
+  },
+  {
     date: new Date('2018-11-02'),
     changes: <>Added <SpellLink id={SPELLS.TRADEWINDS.id} /> module.</>,
     contributors: [Fyruna],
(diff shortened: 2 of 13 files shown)


src/interface/others/AverageTargetsHit.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class AverageTargetsHit extends React.PureComponent {
+  static propTypes = {
+    casts: PropTypes.string.isRequired,
+    hits: PropTypes.string.isRequired,
+    approximate: PropTypes.bool,
```
Dambroda: Thoughts on removing this prop and instead using the math in render() to determine if you're showing a rounded number or an accurate number?

Pewtro: The prop is used in all similiar versions of this (ItemDamageDone etc)

Dambroda: That's true, but I also can't find any instances of the prop actually being passed for any of those. So maybe it's worth also removing it from those and applying the same logic to them?

sMteX: If you still wanna keep the prop (I personally don't mind), you might wanna set `defaultProps` to make it clear if it's by default true or false - adding this to the end (outside class)
```javascript
AverageTargetsHit.defaultProps = {
  approximate: true,
};
```

src/interface/others/AverageTargetsHit.js
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  render() {
+    const { casts, hits, approximate } = this.props;
+    const averageHits = (hits / casts).toFixed(1);
```
sMteX: One last thing I forgot - please check for 0 casts here
```suggestion
    const averageHits = ((hits / casts) || 0).toFixed(1);
```

src/interface/others/AverageTargetsHit.js
Dambroda: I should have made it clear too that my comments weren't a blocking review either, that's why I didn't set to Requested Changes. I was hoping for some discussion about the whole idea of that prop -- as we can see that even in the existing modules it is never used. In this very specific case I feel like because the math is actually being done inside the module itself it probably is an opportune time to not include the prop and use some logic to determine whether or not to include the approximation symbol. 

What would approximate even mean when passed to this module? Does someone using it have to check whether or not hits and casts are factors and then decide whether or not to set approximate to true? If they don't it could show ~1.0 casts, even if it's an exact number. I feel like the ~ should only appear if the number shown was actually rounded after dividing.

PR #2616, opened 2018-11-08
[Mage] Update coldsnap to new event listener

Dambroda:
Thoughts on this anonymous function? What should the general cutoff be for anonymous vs fully defined function?


Gebuz:
Personally I'd avoid using anonymous functions because it's more clear what a function does if it has a name. That's just me though, might be fine like this, especially when it's only a few lines and it's the only event listener in the module.

Dambroda:
Removed anonymous function and replaced with a descriptively named function. Did the same for the similar FrozenOrb cooldown module.

PR #2642, opened 2018-11-11
[Prot Paladin] Light/Hand of the Protector checklist

emallson:
screenshot from 2018-11-11 17-03-04

This checklist item switches to Hand of the Protector when that talent is taken.

### DIFF
--- a/src/parser/paladin/protection/modules/spells/LightOfTheProtector.js
+++ b/src/parser/paladin/protection/modules/spells/LightOfTheProtector.js
@@ -14,6 +14,17 @@ import { REDUCTION_TIME as RP_REDUCTION_TIME } from '../talents/RighteousProtect
 
 const HEAL_DELAY_THRESHOLD = 2000;
 
+/**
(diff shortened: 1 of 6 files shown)


src/parser/paladin/protection/modules/spells/LightOfTheProtector.js
```diff
@@ -0,0 +1,158 @@
...
+  get avgDelay() {
+    if(this._delays.length === 0) {
+      return 0;
+    }
+
+    return this._delays.reduce((sum, delay) => sum + delay, 0) / this._delays.length;
+  }
+
+  get overhealRatio() {
+    return this._overhealing / (this._overhealing + this._actualHealing);
```
sMteX: I know it's unrealistic in real logs, but just in the inifitely small chance that there were no heals with the spell, can you check for division by zero?
```suggestion
    return (this._overhealing / (this._overhealing + this._actualHealing)) || 0;
```

PR #2644, opened 2018-11-11
[Demonology] Pet tracking + pet timeline

sMteX:
May the Force be with you while reviewing this.

I apologize profusely in advance. This is ugly and there's no way around it. Demonology pets are ugly, Wild Imp interactions are a nightmare and Power Siphon particularly is rotten with inconsistencies with my code but I don't know what else to do with it.

I **highly** recommend you to read the comments I added throughout the code. **Let's start with nice things**:
- shows pet names in Events tab in grey (like enemy adds), to not get confused with other classes' colors

image
- added a shiny new Pet Timeline tab, that visualizes the timeline I try my best to recover in `DemoPets.js`
    - this shows all **sources** of pets (either spells or talents, active or passive) and pets summoned with it
    - also shows casts we consider as important - [Power Siphon](https://www.wowhead.com/spell=264130/power-siphon), [Nether Portal](https://www.wowhead.com/spell=267217/nether-portal), [Implosion](https://www.wowhead.com/spell=196277/implosion) and [Demonic Tyrant](https://www.wowhead.com/spell=265187/summon-demonic-tyrant)
    - in particular, Nether Portal "window" is highlighted in purple, because the rotation changes in it, and it's very important to see casts during that window, to judge if the player played in it well or not
    - highlights the key casts (if they overlap with another spell, it's shown in tooltip)
    - highlights pets in the timeline
        - green, if they ever get empowered by Demonic Tyrant
        - red, if they ever get destroyed with Power Siphon, Implosion or talent Demonic Consumption

image

### I'll gladly answer any questions you might have about this. Thank you in advance for your patience with this PR.
-------------
**The ugly truth:** I wish for nothing more than for Blizzard passing pet death events into combat log file. If we had the events when pets die, DemoPets would have maybe 100 LOC instead of over 400. The diabolical mumbo jumbo I had to do to guess when pets should die is awful at times. 

First, added PrepullPetNormalizer, which looks at casts or damage events of pets, and adds them to the front of events, to be picked up by DemoPets and not error as nonexistent. 

Second, Demo pets usually have consistent durations - 12s for Dreadstalkers, 15s for everything else, Demonic Tyrant extends certain *existing* pets by 15s. Then there are Wild Imps. These are based on energy. They despawn after running out of energy or after 15s, whichever comes first. But when they're extended, they're not exactly extended, they just cast for free. Luckily, at least in their `cast` events, there's their energy.

Wild Imps also can get destroyed by various means, and this is where the nightmare begins. They can be destroyed with:
1. Demonic Consumption talent (which destroys all imps)
2. Implosion - which pulls all Imps to the target, exploding them, but there's no "connection" between damage event and the Imp, which caused it (source is player). I estimated it by keeping track of their coordinates and assuming they get pulled at the same speed (which btw. I've been told is 37yd/s), I order them by their distance to target and kill them off in this order
3. Power Siphon 

Now Power Siphon has been troubling me for past 2 days and it's still not perfect - it should sacrifice up to 2 Wild Imps, just like that. From Warlock Discord, I've been told it takes the lowest energy imps. So on PS cast, I took current Imps, ordered them ascending by energy and then spawn (so if they have same energy, the oldest are first) and took up to 2 of them. Haha, **nope**, the sacrificed idiots were happily casting after they should've been killed.

So I made a normalizer which tracks Imps active **after** each PS casts, to know that they **can't** be sacrificed. This sorta works, except in a lot of cases it seems that from log's perspective, I sacrificed 2 imps, but from code, I could have sacrificed only one etc. These two just doesn't agree all the time, but I've ran out of ideas so here it is. My plan is that if there's anything that uses Imps or current pets or something, **and** the player has Power Siphon, I'll put a warning somewhere stating that it can be slightly inaccurate.


yajinni:
Bless your soul for the effort and the soul of the person reviewing this lol

src/parser/warlock/demonology/modules/pets/Timeline.js
```diff
@@ -0,0 +1,43 @@
+import SPELLS from 'common/SPELLS';
+
+import { DESPAWN_REASONS } from './TimelinePet';
+import { isPermanentPet } from './helpers';
+
+const debug = true;
+
```
Pewtro: debug =true

PR #2646, opened 2018-11-12
[Core] Add Azerokk's Resonating Heart module

Aelexe:
Added a statistic box for Azerokk's Resonating Heart.

image

### DIFF
--- a/src/parser/shared/modules/items/bfa/dungeons/AzerokksResonatingHeart.js
+++ b/src/parser/shared/modules/items/bfa/dungeons/AzerokksResonatingHeart.js
@@ -1,0 +1,75 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS/index';
+import ITEMS from 'common/ITEMS/index';
+import { formatPercentage } from 'common/format';
+import { calculatePrimaryStat } from 'common/stats';
+
+import Analyzer from 'parser/core/Analyzer';
+import Abilities from 'parser/shared/modules/Abilities';
+
+const BASE_ITEM_LEVEL = 300;
+const BASE_AGILITY_BUFF = 593;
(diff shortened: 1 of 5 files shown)


src/parser/shared/modules/items/bfa/dungeons/AzerokksResonatingHeart.js
```diff
@@ -0,0 +1,75 @@
...
+class AzerokksResonatingHeart extends Analyzer {
+  static dependencies = {
+    abilities: Abilities,
+  };
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrinket(ITEMS.AZEROKKS_RESONATING_HEART.id);
+
+    if(!this.active) {
```
yajinni: I dont think this is needed. 
if(!this.active)

sMteX: I'm not sure, but I think line 33 and 34 would cause errors if it got there without player having the item. My guess is that getItem for it would return undefined.

Aelexe: It also prevents unnecessary code from being executed. The same approach is used across many other modules so I assumed it was standard.

sMteX: Sometimes it doesn't really matter if it runs few other lines even if not active, but here it would probably crash without the return

yajinni: Ok then the rest of it looks good. Gonna approve.

PR #2651, opened 2018-11-12
[TEST] React-vis instead of Victory (instead of Chart.js)

emallson:
See #2489 for motivation on trying a new charting library.

I'm not happy with the responsiveness/feel of victory. I rebuilt the stagger plot (again) in [react-vis](https://uber.github.io/react-vis/), and it seems to be much more performant even on my crappy laptop.

Wanting to merge this into test to try it out on a few more devices and solicit feedback from people in discord.

[Example](https://cdn.discordapp.com/attachments/316864385052049420/510937173738913795/react-vis-plot.webm)


Pewtro:
Seeing as this is merely going into test, I think it should be fine.

PR #2645, opened 2018-11-12
[Hunter] Barrage and Chimaera Shot updates

Pewtro:
Relies on #2615 
Towards #2529 

image

Same issue with package and package-lock as #2644

Issue #2037, opened 2018-07-31
[Hunter] Implement Azerite traits


Pewtro:
I'm already swamped with getting the 3 specs up-to-date, so any help on any of the following traits would be appreciated! 

https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/%5BBFA%5D-Implementing-Azerite-Traits

## Beast Mastery

- [x] Dance of Death: https://www.wowhead.com/spell=274441/dance-of-death
- [x] Feeding Frenzy: https://www.wowhead.com/spell=278529/feeding-frenzy
- [x] Haze of Rage: https://www.wowhead.com/spell=273262/haze-of-rage
- [x] Primal Instinct: https://www.wowhead.com/spell=279806/primal-instincts

## Marksmanship
- [x] Focused Fire: https://www.wowhead.com/spell=278531/focused-fire
- [x] In The Rhythm: https://www.wowhead.com/spell=264198/in-the-rhythm
- [x] Steady Aim: https://www.wowhead.com/spell=277651/steady-aim
- [x] Unerring Vision: https://www.wowhead.com/spell=274444/unerring-vision

## Survival
- [x] Blur of Talons: https://www.wowhead.com/spell=277653/blur-of-talons
- [x] Latent Poison: https://www.wowhead.com/spell=273283/latent-poison
- [ ] Venomous Fangs: https://www.wowhead.com/spell=274590/venomous-fangs
- [x] Wilderness Survival: https://www.wowhead.com/spell=278532/wilderness-survival
- [ ] Wildfire Cluster: https://www.wowhead.com/spell=272742/wildfire-cluster

## Cross spec
- [ ] Serrated Jaws: https://www.wowhead.com/spell=272717/serrated-jaws -- works for BM and SV
- [x] Rapid Reload: https://www.wowhead.com/spell=278530/rapid-reload -- works for MM and BM

Pewtro:
Wildfire Cluster is handled in #3493 . 
Serrated Jaws and Venomous Fang should be handled as part of #2535

PR #2652, opened 2018-11-12
[Hunter] Animal Companion & Pack Alpha

Pewtro:
Second to last round of Beast Mastery talents. 

Relies on the two previous ones as those are the most important talents, so spend your limited review time there. 

Number 1: #2615 
Number 2: #2645 

Towards #2529 & #2037 

Extremely simple Pack Alpha module for now: 
image

Work for future module can be seen here: https://github.com/Pewtro/WoWAnalyzer/blob/08872054615a54058fba06c590c9054aab1e9c00/src/parser/hunter/beastmastery/modules/spells/azeritetraits/PackAlpha.js

Animal Companion module that looks like this: 
image

### DIFF
--- a/src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
+++ b/src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import Analyzer from 'parser/core/Analyzer';
+import SPELLS from 'common/SPELLS';
+import StatisticBox from 'interface/others/StatisticBox';
+import SpellIcon from 'common/SpellIcon';
+import STATISTIC_CATEGORY from 'interface/others/STATISTIC_CATEGORY';
+import { formatNumber } from 'common/format';
+
+/**
+ * Your Call Pet additionally summons the first pet from your stable.
+ * This pet will obey your Kill Command, but cannot use pet family abilities.
+ *
+ * Additionally this talent baseline reduces all pet damage by 40%.
+ *
+ * Example log: https://www.warcraftlogs.com/reports/aNRJKvknX1FYqQ6G#fight=4&type=damage-done&source=7
+ */
+
+class AnimalCompanion extends Analyzer {
+
+  damage = 0;
+  pets = [];
+  mainPetName = null;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.ANIMAL_COMPANION_TALENT.id);
+  }
+
+  on_byPlayerPet_damage(event) {
+    const sourcePet = this.owner.playerPets.find(pet => pet.id === event.sourceID);
+    const foundPet = this.pets.find(pet => pet.sourceID === event.sourceID);
+    const damage = event.amount + (event.absorbed || 0);
+    if (!foundPet) {
+      this.pets.push({ petName: sourcePet.name, sourceID: event.sourceID, damage: damage });
+    } else {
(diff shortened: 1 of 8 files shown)


src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  on_byPlayerPet_damage(event) {
+    const sourcePet = this.owner.playerPets.find(pet => pet.id === event.sourceID);
+    const foundPet = this.pets.find(pet => pet.sourceID === event.sourceID);
+    const damage = event.amount + (event.absorbed || 0);
+    if (!foundPet) {
+      this.pets.push({ petName: sourcePet.name, sourceID: event.sourceID, damage: damage });
+    } else {
+      const indexOfPet = this.pets.findIndex(index => index.sourceID === event.sourceID);
```
sMteX: Since `find()` on line 34 should return the object *itself* if it finds one, you should be able to add the damage directly to foundPet:
```javascript
} else {
  foundPet.damage += damage;
}
```

src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
```diff
@@ -0,0 +1,90 @@
...
+      this.pets[indexOfPet].damage += damage;
+    }
+  }
+
+  on_finished() {
+    let max = -Infinity;
+    let x;
+    for (x in this.pets) {
```
sMteX: I think this could also be simplified with (haven't tried)
```javascript
this.pets.forEach(pet => {
  if (pet.damage > max) {
    max = pet.damage;
    this.mainPetName = pet.petName;
  }
});
```

src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
```diff
@@ -0,0 +1,90 @@
...
+      if (this.pets[x].damage > max) {
+        max = this.pets[x].damage;
+        this.mainPetName = this.pets[x].petName;
+      }
+    }
+  }
+
+  statistic() {
+    const totalDamage = this.pets.reduce((totalDamage, petDamage) => ({ damage: totalDamage.damage + petDamage.damage }));
```
sMteX: Or (just an alternative, this is fine as well, maybe just unnecessarily creates objects, doubt it would make any difference though)
```javascript
this.pets.map(pet => pet.damage).reduce((total, current) => total + current, 0);
```

src/parser/hunter/beastmastery/modules/talents/AnimalCompanion.js
```diff
@@ -0,0 +1,90 @@
...
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.ANIMAL_COMPANION_TALENT.id} />}
+        value={<>{formatNumber(totalDamage.damage)} / {formatNumber(totalDamage.damage / (this.owner.fightDuration / 1000))} DPS</>}
+        label="Animal Companion"
+        tooltip={`Without this talent your main pet would have done 66% more damage than it did as Animal Companion nerfs all pet damage by 40% - however if you're using the azerite trait Pack Alpha, your main pet would have done less damage that way.`}
```
sMteX: Since you're not using any variable in it
```suggestion
        tooltip="Without this talent your main pet would have done 66% more damage than it did as Animal Companion nerfs all pet damage by 40% - however if you're using the azerite trait Pack Alpha, your main pet would have done less damage that way."
```

PR #2655, opened 2018-11-12
[Subtlety Rogue] Add Inevitability module

Aelexe:
Adds a stat block for Inevitability, displaying the damage and duration contributed.

The damage is calculated specifically for the extended duration instead of as a percentage of the whole duration, as damage is weighted towards the start of the duration, meaning not doing so would inflate the value of this trait.

The module does not currently take into account the flat damage added to Backstab and Shadowstrike, but that contributes a fairly minor part to the overall value of this trait.

image

### DIFF
--- a/src/parser/rogue/subtlety/modules/azeritetraits/Inevitability.js
+++ b/src/parser/rogue/subtlety/modules/azeritetraits/Inevitability.js
@@ -1,0 +1,119 @@
+import React from 'react';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'interface/others/TraitStatisticBox';
+
+import { formatNumber, formatPercentage } from 'common/format';
+import SPELLS from 'common/SPELLS/index';
+
+import Analyzer from 'parser/core/Analyzer';
+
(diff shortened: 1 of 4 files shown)


src/parser/rogue/subtlety/modules/azeritetraits/Inevitability.js
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	on_byPlayer_damage(event) {
+		// Total damage during Symbols of Death, after the base duration has passed.
+		if (!this.selectedCombatant.hasBuff(SPELLS.SYMBOLS_OF_DEATH.id) || event.timestamp <= this.lastSymbolsCastTime + SYMBOLS_BASE_DURATION) {
+			return;
+		}
+
+		this.symbolsExtendedDamage += event.amount + event.absorbed;
```
Gebuz: Needs a `calculateEffectiveDamage(event, BONUS_DAMAGE);`

MartijnHols:
Closing because stale for a long time

PR #2662, opened 2018-11-14
[Fury] Cleaning up

matardarix:
Just cleaning up & merged some file + removed legion stuff


src/parser/warrior/fury/modules/core/Rampage.js
```diff
@@ -0,0 +1,96 @@
...
+  on_byPlayer_damage(event) {
+    if (!this.rampage.includes(event.ability.guid)) {
+      return;
+    }
+    this.counter[event.ability.guid] += 1;
+  }
+
+  get cancelledsuggestionThresholds() {
+    const max = Object.values(this.counter).reduce((max, current) => current > max ? current : max, 0);
+    const wasted = Object.keys(this.counter).reduce((acc, current) => acc + max - this.counter[current], 0);
```
sMteX: You could iterate through values again directly
```javascript
const wasted = Object.keys(this.counter).reduce((total, current) => total + (max - current), 0);
```

src/parser/warrior/fury/modules/talents/AngerManagement.js
```diff
@@ -18,29 +22,35 @@ class AngerManagement extends Analyzer {
     spellUsable: SpellUsable,
   };
 
+  cooldownsAffected = [
```
sMteX: Looking at the talent I understand why this is an array, but for Fury it only affects 1 ability - Recklessness. Rewriting it to only take into account this fact (and **not** be generalized for array of spells) would simplify the module.

src/parser/warrior/fury/modules/talents/AngerManagement.js
```diff
@@ -18,29 +22,35 @@ class AngerManagement extends Analyzer {
...
+      return;
+    }
+    const rage = event.classResources.find(e => e.type === RESOURCE_TYPES.RAGE.id);
+    if (!rage || !rage.cost) {
       return;
     }
 
-    const rageSpend = event.classResources.find(e => e.type === RESOURCE_TYPES.RAGE.id).cost / RAGE_NEEDED_FOR_A_PROC;
+    const rageSpend = rage.cost / 10;
     const reduction = rageSpend / RAGE_NEEDED_FOR_A_PROC * CDR_PER_PROC;
```
sMteX: Does this work only in multiples of `RAGE_NEEDED_FOR_A_PROC`? Meaning that 
- 20 rage = 1 sec
- 30 rage = 1 sec
- 40 rage = 2 sec
Or also in "partial CDR", such as
- 20 rage = 1 sec
- 30 rage = 1,5 sec
- 40 rage = 2 sec
? If it's the first one, you need to round down after dividing:
```javascript
const reduction = Math.floor(rageSpend / RAGE_NEEDED_FOR_A_PROC) * CDR_PER_PROC;
```

matardarix: Works like the second way, in "partial CDR" -> 30 rage = 1,5s

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,108 @@
...
+/**
+ Frothing Berserker users should maximize casts with 100 rage to proc Frothing Berserker the post
+ Carnage users should cast as soon as possible, i.e. at 70 rage
+ Massacre users (not from Soul of the Battlelord) should cast as soon as possible, i.e. at 85 rage
+ Whenever Battle Cry is active, user should cast Rampage whenever possible - regardless of talents
+ */
+
+class FrothingBerserker extends Analyzer {
+
+  last_timestamp = null;
```
Dambroda: Variables  should follow naming conventions:
lastTimestamp
castsCounter
etc.

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,108 @@
...
+  premature_counter = 0;
+  reckless_abandon = null;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.FROTHING_BERSERKER_TALENT.id);
+    this.reckless_abandon = this.selectedCombatant.hasTalent(SPELLS.RECKLESS_ABANDON_TALENT.id);
+  }
+
+  on_byPlayer_cast(event) {
```
Dambroda: While not required (as in you don't have to resolve this, just a suggestion), the logic I see in this event makes me believe it would benefit greatly from the new event listeners. Take a look at #2705 (L34-38) for an example of how you can split up event listeners for specific spells and help lower the number of nested conditionals.

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      this.casts_counter++;
+
+      if (rage < 100 && !this.selectedCombatant.hasBuff(SPELLS.RECKLESSNESS.id)) {
+        this.premature_counter++;
+        this.last_timestamp = event.timestamp;
+      }
+    } else if (event.ability.guid === SPELLS.RECKLESSNESS.id && this.last_timestamp && this.reckless_abandon) {
+      if (event.timestamp - this.last_timestamp < 2000) {
```
Dambroda: Avoid using magic numbers like 2000, if you have a cast buffer or something like it, it's good to create a const above the class with a name like PREMATURE_RECKLESSNESS_WINDOW or the like.

src/parser/warrior/fury/modules/talents/FuriousSlash.js
```diff
@@ -0,0 +1,121 @@
...
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.FURIOUS_SLASH_TALENT.id);
+    this.furiousSlashStacks = Array.from({ length: MAX_FURIOUS_SLASH_STACKS + 1 }, x => []);
+  }
+
+  on_byPlayer_damage(event) {
+    const spellId = event.ability.guid;
+    if (spellId !== SPELLS.FURIOUS_SLASH_TALENT.id || event.hit_type === HIT_TYPES.DODGE || event.hit_type === HIT_TYPES.DODGE) {
```
Dambroda: Checking for dodge twice?

src/parser/warrior/fury/modules/talents/FuriousSlash.js
```diff
@@ -0,0 +1,121 @@
...
+import SpellLink from 'common/SpellLink';
+import { formatDuration, formatPercentage } from 'common/format';
+import React from 'react';
+import StatisticBox from 'interface/others/StatisticBox';
+import SpellIcon from 'common/SpellIcon';
+import STATISTIC_ORDER from 'interface/others/STATISTIC_ORDER';
+
+const MAX_FURIOUS_SLASH_STACKS = 3;
+
+class FuriousSlash extends Analyzer {
```
Dambroda: I could be misunderstanding the logic behind this module, but as it's tracking a buff is there are reason it's using onCast events instead of listening to buff update events?

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
matardarix: I'll have to rewrite this module anyway, so i've just removed it for now

src/parser/warrior/fury/modules/talents/FuriousSlash.js
matardarix: It's tracking the buff uptime per stacks, and those stacks are updated by casting Furious Slash

matardarix:
Closing this since Weep is taking up on Fury

PR #2665, opened 2018-11-14
[Arms] Slam suggestion

matardarix:
Added a suggestion on using Slam while Mortal Strike was available 
image

### DIFF
--- a/src/parser/warrior/arms/modules/core/Slam.js
+++ b/src/parser/warrior/arms/modules/core/Slam.js
@@ -41,7 +41,7 @@ class Slam extends Analyzer {
 
   suggestions(when) {
     when(this.badCastSuggestionThresholds).addSuggestion((suggest, actual, recommended) => {
-      return suggest(<>You shouldn't cast <SpellLink id={SPELLS.SLAM.id} /> while <SpellLink id={SPELLS.MORTAL_STRIKE.id} /> is available as Mortal Strike is more rage efficient.</>)
+      return suggest(<>Try to avoid using <SpellLink id={SPELLS.SLAM.id} /> when <SpellLink id={SPELLS.MORTAL_STRIKE.id} /> is available as it is more rage efficient.</>)
         .icon(SPELLS.SLAM.icon)
         .actual(`${formatPercentage(actual)}% of your Slam were used while Mortal Strike or Overpower were available.`)
         .recommended(`${recommended}% is recommended`);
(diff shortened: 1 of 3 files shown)


src/parser/warrior/arms/modules/core/Slam.js
```diff
@@ -0,0 +1,53 @@
...
+  static dependencies = {
+    spellUsable: SpellUsable,
+  };
+
+  badCast = 0;
+  totalCast = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SLAM), this._badCast);
```
sMteX: This has been a hot topic recently on Discord (naming the event listeners) and there's not a definitive answer, but I would suggest a more descriptive name, perhaps `onBadSlamCast` or `onSlamCast`, `handleSlam` etc., `_badCast` doesn't really tell much at first glance.

matardarix: Yeah i've seen the discussion, i have to admit, didn't knew how i should name this to be consistent :p

src/parser/warrior/arms/modules/core/Slam.js
```diff
@@ -0,0 +1,53 @@
...
+
+  badCast = 0;
+  totalCast = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SLAM), this._onSlamCast);
+  }
+
+  _onSlamCast() {
```
sMteX: This also receives `event` as a parameter (see LOD module), but because it's not used here, it's probably fine this way.
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/d5a23314cfa621d63f9df8d6cce9cc5b262125e1/src/parser/paladin/holy/modules/core/LightOfDawn.js#L9

src/parser/warrior/arms/modules/core/Slam.js
```diff
@@ -0,0 +1,53 @@
...
+    this.totalCast += 1;
+    if (this.spellUsable.isAvailable(SPELLS.MORTAL_STRIKE.id)) {
+      this.badCast += 1;
+    }
+  }
+
+  get badCastSuggestionThresholds() {
+    return {
+      actual: this.badCast / this.totalCast ,
```
sMteX: You should probably check for division by zero here
```suggestion
      actual: (this.badCast / this.totalCast) || 0,
```

src/parser/warrior/arms/modules/core/Slam.js
matardarix: Oh I didn't know that, I guess I could change the old way of listening to events on other files at some point

sMteX: Of course, it's *event* listener after all 😄 I guess it will become the preferred way of handling events, but both ways are supported now (not combined though).

matardarix: Yeah that's make sense 😅

PR #2671, opened 2018-11-15
[Fire Mage] Hot Streak Redo

Sharrq:
I redid Hot Streak to be easier to work with, updated it to the new event listeners, and fixed some bugs.

- Split Hot Streak into 3 separate modules (HotStreak, HotStreakWastedCrits, and HotStreakPreCasts)
- Split the double Stat Box into two separate boxes (which should be easier to understand)
- Rewrote the PreCasts suggestion to not include precasting Scorch cause that leads to munched procs.
- Changed the analysis on PreCasts to ignore anything during Combustion, the Firestarter execute, or the Searing Touch execute
- removed some variables that werent used
- Updated the Checklist to look at the new files
- Added comments throughout the code
- Added a position to Heating Up's stat box (Cause it was ordered weird)
- Added timeline highlights for wasted crits
- Changelog stuff

Stat Boxes Before
image

Stat Boxes After
image
image
image

PreCasts suggestion before
image

PreCasts suggestion after
image

Wasted Crits Timeline Highlight
image

### DIFF
--- a/src/parser/mage/fire/modules/features/HotStreak.js
+++ b/src/parser/mage/fire/modules/features/HotStreak.js
@@ -19,15 +19,14 @@ class HotStreak extends Analyzer {
 
   constructor(...args) {
     super(...args);
-    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.PYROBLAST), this.onPyroblastCast);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell([SPELLS.PYROBLAST,SPELLS.FLAMESTRIKE]), this.onHotStreakSpenderCast);
     this.addEventListener(Events.applybuff.to(SELECTED_PLAYER).spell(SPELLS.HOT_STREAK), this.onHotStreakApplied);
-    this.addEventListener(Events.removebuff.to(SELECTED_PLAYER).spell(SPELLS.HOT_STREAK), this.onHotStreakRemoved);
     this.addEventListener(Events.removebuff.to(SELECTED_PLAYER).spell(SPELLS.HOT_STREAK), this.checkForExpiredProcs);
 
   }
 
   //When Pyroblast is cast, get the timestamp. This is used for determining if pyroblast was cast immediately before Hot Streak was removed.
-  onPyroblastCast(event) {
+  onHotStreakSpenderCast(event) {
     this.castTimestamp = event.timestamp;
   }
 
@@ -36,10 +35,7 @@ class HotStreak extends Analyzer {
     this.totalHotStreakProcs += 1;
   }
 
-  onHotStreakRemoved(event) {
-    this.hotStreakRemoved = event.timestamp;
-  }
-
+  //Checks to see if there was a Hot Streak spender cast immediately before Hot Streak was removed. If there was not, then it must have expired.
   checkForExpiredProcs(event) {
     if (!this.castTimestamp || this.castTimestamp + PROC_WINDOW_MS < event.timestamp) {
       debug && this.log("Hot Streak proc expired");
--- a/src/parser/mage/fire/modules/features/HotStreak.js
+++ b/src/parser/mage/fire/modules/features/HotStreak.js
@@ -1,139 +1,62 @@
 import React from 'react';
-import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 import SpellIcon from 'common/SpellIcon';
-import { formatPercentage, formatNumber } from 'common/format';
+import { formatPercentage } from 'common/format';
 import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
-import Analyzer from 'parser/core/Analyzer';
-import HIT_TYPES from 'game/HIT_TYPES';
-import EnemyInstances, { encodeTargetString } from 'parser/shared/modules/EnemyInstances';
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
 
 const debug = false;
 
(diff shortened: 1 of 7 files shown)


src/parser/mage/fire/modules/features/HotStreak.js
```diff
@@ -1,139 +1,62 @@
...
-  get castBeforeHotStreakUtil() {
-    return 1 - (this.noCastBeforeHotStreak / (this.castedBeforeHotStreak + this.noCastBeforeHotStreak));
-  }
-
-  get wastedCritsPerMinute() {
-    return this.wastedCrits / (this.owner.fightDuration / 60000);
+    return 1 - (this.expiredProcs / this.totalHotStreakProcs) || 0;
   }
 
   get hotStreakUtilizationThresholds() {
```
Dambroda: I feel like these should be meaner tbh -- only time they should expire is when you are going into intermission. 95-90-85?

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
```diff
@@ -0,0 +1,104 @@
...
+    this.hasPyromaniac = this.selectedCombatant.hasTalent(SPELLS.PYROMANIAC_TALENT.id);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER), this._onCast);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER), this._onDamage);
+    this.addEventListener(Events.applybuff.to(SELECTED_PLAYER).spell(SPELLS.HOT_STREAK), this.checkForPyromaniacProc);
+
+  }
+
+  //When a spell that contributes towards Hot Streak is cast, get the event info to use for excluding the cleaves from Phoenix Flames on the damage event
+  _onCast(event) {
+    if (!HOT_STREAK_CONTRIBUTORS.includes(event.ability.guid) && event.ability.guid !== SPELLS.FLAMESTRIKE.id) {
```
Dambroda: Should probably only not return early on flamestrike if this.hasPyromaniac -- otherwise it's not considered a contributor; unless I misunderstand the logic here

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
```diff
@@ -0,0 +1,104 @@
...
+    const spellId = event.ability.guid;
+    const castTarget = encodeTargetString(this.lastCastEvent.targetID, event.targetInstance);
+    const damageTarget = encodeTargetString(event.targetID, event.targetInstance);
+    if (!HOT_STREAK_CONTRIBUTORS.includes(spellId) || event.hitType !== HIT_TYPES.CRIT || !this.selectedCombatant.hasBuff(SPELLS.HOT_STREAK.id) || (spellId === SPELLS.PHOENIX_FLAMES_TALENT.id && castTarget !== damageTarget)) {
+      return;
+    }
+
+    if (this.hasPyromaniacProc) {
+      debug && this.log("Wasted Crit Ignored");
+    } else if (this.selectedCombatant.hasBuff(SPELLS.HOT_STREAK.id,null,-50)) {
```
Dambroda: You check for .hasBuff(HOT_STREAK) just a few lines up -- maybe remove this check and move the -50ms buffer to the previous check on L55?

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
```diff
@@ -0,0 +1,104 @@
...
+      this.wastedCrits += 1;
+      this.lastCastEvent.meta = this.lastCastEvent.meta || {};
+      this.lastCastEvent.meta.isInefficientCast = true;
+      this.lastCastEvent.meta.inefficientCastReason = "This cast crit while you already had Hot Streak and could have contributed towards your next Heating Up or Hot Streak.";
```
Dambroda: Maybe something about what they can change as well, like "Use the proc immediately with pyro/fs.." etc?

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
Sharrq: i dont really remember why flamestrike is in this ... it was in the old one and i carried it over anyway cause im sure there is some reason i added it here ... but i dont remember why.

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
Sharrq: this was another thing i remember adding but i dont remember why (and was in the old module so i carried it over) ill dig through my prs maybe i can find it.

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
Sharrq: oh ok, flamestrike was here before because of the expired procs check, i need to remove that from this and add it to the expired procs event listener instead.

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
Sharrq: nvm i just re-read your comment. i thought you were saying to remove the -50 buffer, but you were saying to move it to the other check. yea ill do that.

src/parser/mage/fire/modules/features/HotStreakWastedCrits.js
Dambroda: I thought flamestrike was in here because of pyromaniac? FS has a chance to proc it?

Sharrq: Umm, I don't think so, cause I'm pretty sure the extent of my Pyromaniac check is just whether the player instantly got a new hot streak and has nothing to do with what was cast.

Sharrq:
made the requested changes and also pulled in master to allow arrays in the event filters.

PR #2673, opened 2018-11-15
[Mage] Add a suggestion to talent into Rune of Power when playing Arcane or Fire

Dambroda:
According to sims and mage TCS, there are really no situations where Incanter's Flow or Mirror Image are good options. On average, looking at WCL, a well played RoP increases throughput by about 10% on average. Because it's a playstyle choice, I think it's better to not make it part of the checklist and to keep it as an _Average_ suggestion. I also added an easy way (the constant at the top of the module) to enable/disable the suggestion as talents are rebalanced.

Also, changed module to new eventlistener 😉 

image

### DIFF
--- a/src/parser/mage/shared/modules/features/RuneOfPower.js
+++ b/src/parser/mage/shared/modules/features/RuneOfPower.js
@@ -15,7 +15,7 @@ import SUGGESTION_IMPORTANCE from 'parser/core/ISSUE_IMPORTANCE';
 /*
  * If Rune of Power is substantially better than the rest of the row, enable
  * ROP talent suggestion. At time of writing, it's a substantial increase over
- * incanters flow for fire and arcane in almost all situations.
+ * incanters flow for fire and arcane in all situations.
  */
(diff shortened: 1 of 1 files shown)


src/parser/mage/shared/modules/features/RuneOfPower.js
```diff
@@ -59,31 +75,46 @@ class RuneOfPower extends Analyzer {
...
+            .staticImportance(SUGGESTION_IMPORTANCE.REGULAR);
         });
+      return;
+    }
+
+    if(!this.showSuggestion) {
+      return;
     }
+
+    when(this.damageSuggestionThresholds)
```
Sharrq: just to make sure im reading this right, the return after the SuggestRoP suggestion is preventing this suggestion from loading if the player doesnt have Rune of Power, correct? Since the module now loads when you dont have RoP.

Dambroda: Correct

PR #2674, opened 2018-11-15
[Core] Add Dread Gladiator's Insignia module

Aelexe:
Added Dread Gladiator's Insignia stat block.

image

### DIFF
--- a/src/parser/shared/modules/items/bfa/pvp/DreadGladiatorsInsignia.js
+++ b/src/parser/shared/modules/items/bfa/pvp/DreadGladiatorsInsignia.js
@@ -8,8 +8,8 @@ import { calculatePrimaryStat } from 'common/stats';
 import Analyzer from 'parser/core/Analyzer';
 import Abilities from 'parser/shared/modules/Abilities';
 
-const BASE_ITEM_LEVEL = 280;
-const BASE_PRIMARY_STAT_BUFF = 63;
+const BASE_ITEM_LEVEL = 335;
+const BASE_PRIMARY_STAT_BUFF = 462;
 
 /**
  * Dread Gladiator's Insignia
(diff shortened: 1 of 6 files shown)


src/parser/shared/modules/items/bfa/pvp/DreadGladiatorsInsignia.js
```diff
@@ -0,0 +1,75 @@
...
+import SPELLS from 'common/SPELLS/index';
+import ITEMS from 'common/ITEMS/index';
+import { formatPercentage } from 'common/format';
+import { calculatePrimaryStat } from 'common/stats';
+
+import Analyzer from 'parser/core/Analyzer';
+import Abilities from 'parser/shared/modules/Abilities';
+
+const BASE_ITEM_LEVEL = 280;
+const BASE_PRIMARY_STAT_BUFF = 63;
```
Gebuz: Is this value correct? It sounds very low.
https://www.wowhead.com/item=161676/dread-gladiators-insignia
https://www.wowdb.com/items/161676-dread-gladiators-insignia

Gebuz: The armory value looks way too low aswell, at a quick glance it is likely same value as wowhead (100 ilvl -> 3xprimary stat sounds like the correct ballpark). IDK if the wowdb value is correct either, but it's atleast in the correct range (277/2 = 138.5). Where as both armory and wowhead are both way off.

niseko: Both wowhead and wowdb agree on the number if you give it a realistic itemlevel
https://www.wowhead.com/item=161676/dread-gladiators-insignia&bonus=4982:1532
https://www.wowdb.com/spells/277182-taste-of-victory?itemLevel=340&itemSlot=12&playerLevel=120

Gebuz:
Forgot to mention this, but please add it to StatTracker aswell. Example:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/9c4bb03c2713bdfea93325e7930f18957710d8ef/src/parser/shared/modules/StatTracker.js#L353

Aelexe:
I took the opportunity to add Azerokk's Resonating Heart to the stat tracker as well, since it was missed in the original pull request.

PR #2675, opened 2018-11-15
[Arms] Reworked Executioner's Precision

matardarix:
Changed the suggestions and the Checklist for Executioner's Precision trait
image
image
image

### DIFF
--- a/src/parser/warrior/arms/modules/spells/azeritetraits/ExecutionersPrecision.js
+++ b/src/parser/warrior/arms/modules/spells/azeritetraits/ExecutionersPrecision.js
@@ -7,7 +7,6 @@ import SpellLink from 'common/SpellLink';
 import SPELLS from 'common/SPELLS';
 import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Enemies from 'parser/shared/modules/Enemies';
-
 import SpellUsable from '../../features/SpellUsable';
 
 /**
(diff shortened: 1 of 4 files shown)


src/parser/warrior/arms/modules/spells/azeritetraits/ExecutionersPrecision.js
```diff
@@ -1,56 +1,64 @@
...
+    const ep = enemy.getBuff(SPELLS.EXECUTIONERS_PRECISION_DEBUFF.id);
+    const op = this.selectedCombatant.getBuff(SPELLS.OVERPOWER.id);
+    if ((ep === undefined || op === undefined) || (!ep && !op && !enemy)) {
       return;
     }
 
-    const executionersPrecision = enemy.getBuff(SPELLS.EXECUTIONERS_PRECISION_DEBUFF.id);
-    if (executionersPrecision && executionersPrecision.stacks === 2 && this.spellUsable.isAvailable(SPELLS.MORTAL_STRIKE.id)) {
-      this.wastedProcs += 1;
+    if ((ep.stacks !== 2 && op.stacks !== 2) || (ep.stacks !== 2 && op.stacks === 2) || (ep.stacks === 2 && op.stacks !== 2)) {
```
Sharrq: couldnt this be simplified by just doing if (ep.stacks !== 2 || op.stacks !== 2) instead of writing out every scenario? That way if EP or OP are not 2 then it is bad?

PR #2676, opened 2018-11-15
[Core] Add array support for buffSpellId

### LINKED ISSUE #2305: Abilities: add support for multiple buffSpellIds (an array like `spell`)
Example usage: CombatPotion

niseko:
As i've seen a lot of requests for that

image
prepot:
image
actual pot:
image
different pot:
image

fixes #2305 #2306

### DIFF
--- a/src/parser/shared/modules/features/TimelineTab/TabComponent/SpellTimeline.js
+++ b/src/parser/shared/modules/features/TimelineTab/TabComponent/SpellTimeline.js
@@ -193,18 +193,26 @@ class SpellTimeline extends React.PureComponent {
               })}
             </div>
           )}
-          {this.spells.map(spellId => (
-            <SpellRow
-              key={spellId}
-              className="lane"
-              events={historyBySpellId[spellId] || []}
-              buffEvents={buffEvents[abilities.getBuffSpellId(spellId)]}
-              start={start}
-              totalWidth={totalWidth}
-              secondWidth={secondWidth}
-              showCooldowns={showCooldowns}
-            />
-          ))}
+          {this.spells.map(spellId => {
+            let combinedBuffEvents = [];
+            if (abilities.getBuffSpellId(spellId) && abilities.getBuffSpellId(spellId) instanceof Array) {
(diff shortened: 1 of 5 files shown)


niseko:
Pretty sure that the label being wrong is a separate issue, caused by it just taking the first spell in the array of spells. This PR enables buffSpellId to take arrays, for spells like where you either have 1 spell with multiple buffs it can trigger, or have multiple spells combined as one, which have separate buffs (Hex, the combat potion class).

Gebuz:
Looks good to me.

The Pot Icon fix would likely require something like matching the first icon on it's row instead of using the first spell. I'm not sure if that's even possible without writting some ugly code.

src/parser/shared/modules/features/TimelineTab/TabComponent/SpellTimeline.js
```diff
@@ -193,18 +193,26 @@ class SpellTimeline extends React.PureComponent {
...
-              buffEvents={buffEvents[abilities.getBuffSpellId(spellId)]}
-              start={start}
-              totalWidth={totalWidth}
-              secondWidth={secondWidth}
-              showCooldowns={showCooldowns}
-            />
-          ))}
+          {this.spells.map(spellId => {
+            let combinedBuffEvents = [];
+            if (abilities.getBuffSpellId(spellId) && abilities.getBuffSpellId(spellId) instanceof Array) {
```
sMteX: In order to not call the thing all the time (since afaik, it iterates through Abilities every single time), I'd suggest something like this:
```javascript
let combinedBuffEvents = [];
const buffSpellId = abilities.getBuffSpellId(spellId);
if (buffSpellId && buffSpellId instanceof Array) {
  buffSpellId.forEach(spell => {
    combinedBuffEvents.push(buffEvents[spell]);
  });
} else if (buffSpellId) {
  combinedBuffEvents = buffEvents[buffSpellId];
}
```

Dambroda:
I'll fix #2571 if this gets merged, noted

PR #2680, opened 2018-11-15
[Core] Allow array in new event spell filter

Dambroda:
Not sure if it was intentionally left out. Also was considering adding a way to send either numbers/numbers array instead of only spell objects, but I don't want to mess with the design too much.


Sharrq:
zero approved it so im gonna merge it

PR #2684, opened 2018-11-15
buffSpellId array bugfix

niseko:


Dambroda:
can just use .push instead of prototype, but this needs fix ASAP

PR #2698, opened 2018-11-16
Show cancelled casts on the timeline

niseko:
Seeing cancelled casts can explain a lot of gaps in the timeline without having to dig deeper. I'm offsetting it slightly because a cast finish into a failed cast would fully cover up the icon otherwise.

image
image
image
https://www.warcraftlogs.com/reports/RXWCm7BZMjPb4K6c/#fight=7

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,11 @@ import Contributor from 'interface/contributor/Button';
 
 export default [
   {
+    date: new Date('2018-11-16'),
+    changes: 'Added cancelled casts to the timeline.',
+    contributors: [niseko],
+  },
+  {
     date: new Date('2018-11-14'),
     changes: <>Added <SpellLink id={SPELLS.GIFT_OF_THE_NAARU_MAGE.id} /> to ability list.</>,
     contributors: [Dambroda],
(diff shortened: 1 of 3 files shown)


Dambroda:
I really like this idea, does the offset scale with zoom level? My worry would be that at 1x the offset makes the cast look even further off. Also one other thing I wonder (not blocking) is if we could also show the 'cancelled cast' in the cast bar, maybe also filtered gray.

niseko:
2x, no offset at all
image

2x, with offset
image

1x
image

3x
image

5x
image

10x
image


It didn't have a glow yet when I initially added the offset, maybe that is enough to indicate that there is a cancelled cast hidden?
Issue about showing those casts in the cast-bar is that you don't know *when* they got cancelled as there is 0 information, best I'd be able to do is GCD or start of next cast depending on which is sooner.

niseko:
Disappearing behind other icons only happens in the case that you cast the same spell twice in a row without any pause, and cancel the 2nd one, which seems pretty rare. 

No idea how I'd make something that isn't the spells icon look decent, here's a try with wider ticks (+ trying to add it to the channeling bar, not knowing how long the cast was still bothers me though)
image

MartijnHols:
How about the icon in grayscale at 50% size (or smaller) with a cross over it?

I don't deny it's a tough one

niseko:
Closing this for now as there's work being done on a new timeline in <https://github.com/WoWAnalyzer/WoWAnalyzer/tree/interfeces-3.0>.

PR #2705, opened 2018-11-17
[Fire Mage] Heating Up Redo

Sharrq:
Rewrote Heating Up with the new event listeners.

Also fixed an issue that was causing the module to not properly register that the player was in the Searing Touch execute window if they had the talent.

- Updated to new event listeners
- Fixed Searing Touch execute window
- Increased the thresholds now that the values are accurate
- Changelog


Dambroda:
I didn't look too much at how the logic works, so this time I am intentionally just approving and letting you merge once you're happy with the amount of testing/debugging you've done 👍

Issue #2688, opened 2018-11-15
Test all combat log parsers 


Pewtro:
Ensure every single spec _atleast_ actually compiles

Dambroda:
Great idea. The only problem with this I can forsee would be people using incorrect values (based on bad math or existing code that doesn't work right) as their test criteria, which then fails when the code is fixed. Could lead to some confusion. We'd just need to be very anal about how people determine their test criteria as well

emallson:
thats something i ran into with the old [brm test code.](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/legion/src/Parser/Monk/Brewmaster/Modules/Core/BrewCDR.test.js) It broke several times in rapid succession because i set the test criteria based on bad values (which, tbf, the only values I had access to were the outputs from the analyzer because of the nature of that module).

so its something that will be a learning experience, but i think will be a valuable tool. especially since we have a lot of stable code now for handling CDR / buffs / whatever and for many things can cross-reference with WCL to double-check expected values

MartijnHols:
This is here in the form of integration tests as used in specs such as Holy Paladin.

PR #2708, opened 2018-11-18
Integration Tests for BrM/Prot Pally

emallson:
Adds an integration test for brewmaster and prot paladin that just runs the CombatLogParser through the log used as an example and checks that it completes without errors.

While both these modules worked without issues. To check if the integration test works, i introduced some small `null`/`undefined` kind of errors that typically cause crashes, which then made the tests fail as expected.

There *is* a performance impact for running these of about 6s per integration test. When working locally, the auto-runner should only be re-running your own and thus the slowdown w.r.t. local work should be manageable at about 6s. On Travis, this will add 3-5 minutes to the wait for CI after each update, which is annoying but not unreasonable.

This should *hopefully* help catch things like that thursday adventure before they get pushed live in the future.

This starts to address #2688.

UPDATE: This now also adds support for consistency tests at a per-module granularity. By default, calling `snapshotTest` will test the `statistic()` method and catch changes. It also accepts a `propFn` argument that maps an analyzer to a serialized value to be snapshotted if another property snapshot is desired (for example, on analyzers that do not have statistics you may want to check the suggestion).


emallson:
I may update this PR to [this branch.](https://github.com/WoWAnalyzer/WoWAnalyzer/compare/master...emallson:feature/consistency-test-tools) Reviewer beware.

MartijnHols:
mfw I never hit submit on my review

emallson:
lol well I'll take feedback if you have it and take what is relevant to the later branch.

src/parser/core/tests/CombatLogParserIntegrationTest.js
```diff
@@ -0,0 +1,82 @@
...
+        report.fights.find(({id}) => id === fightId),
+        combatantInfoEvents
+      );
+      events.forEach(event => parser.getModule(EventEmitter).triggerEvent(event));
+      parser.finish();
+      const results = parser.generateResults({
+        i18n,
+        adjustForDowntime: false,
+      });
+      expect(results).toBeTruthy();
```
MartijnHols: I strongly believe you should make the results available for testing. This test isn't nearly as interesting as a test would be that validates if "statistic Y" has the value "1337%", which ensures the validation results (which should be fully consistent regardless of base changes) don't change.

src/parser/monk/brewmaster/CombatLogParser.test.js
```diff
@@ -0,0 +1,14 @@
...
+import CombatLogParser from './CombatLogParser';
+
+const FIGHT_ID = 12;
+const PLAYER_ID = 3;
+
+describe('The Brewmaster Analyzer', integrationTest(
+  CombatLogParser,
+  __dirname + '/test-fixtures/example/meta.json',
+  __dirname + '/test-fixtures/example/combatant-info.json',
+  __dirname + '/test-fixtures/example/events.json.gz',
```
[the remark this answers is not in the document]
emallson: this is something that i have in the works :+1:

src/parser/core/tests/CombatLogParserIntegrationTest.js
emallson: i agree. right now i haven't quite figured out a good way of generating the expected output to be stored in the repository. I've been considering generating & saving the results if they don't exist (and failing the test so that forgetting to commit the expected output fails travis), then checking against them if they do. so then adding consistency tests for a new module is as simple as adding a test like:

```js
describe("MyNewModule", () => {
  // other tests
  consistencyTest(CombatLogParser, MyNewModule, "log-identifier");
});
```
and then `git add <expected output file>`.

I would like to strike a balance between automation and brittle testing, because doing strict equality will likely produce tests that fail often for irrelevant changes (e.g. an invisible change to the html produced by the `StatisticBox` shouldn't fail the test). I haven't done much testing yet but I'm thinking of using a list of props to check for consistency so that module authors can pick out the things they believe should be tested.

MartijnHols: just do a snapshot test

emallson: I didn't realize those existed. Looks good :+1:

emallson:
(Leaving the above Travis comment to show that the snapshot tests do NOT pass on CI if the snapshot files are not committed)

src/parser/core/tests/CombatLogParserIntegrationTest.js
MartijnHols: How about making this a snapshot test?

emallson: So my reason for leaving this as a simple crash test is twofold:

1. Blindly testing against the total serialized output means that the answer to the question "should this test have broken?" is almost always yes. Virtually every core or spec-local change will change *some* output by definition, so the having one big snapshot test only really checks if another spec's changes changed this one.

2. Testing against serialized output of each individual module in an automated fashion is much more complex. I think it is probably possible to do, and it'd likely be value in having a loop over the (spec-local) modules that effectively does `expect(module.statistic()).toMatchSnapshot()` on each instead of requiring it be done individually. This covers 99% of our problem cases in a manner that avoids the "one big snapshot" issue.

As a matter of fact I have talked myself into the latter solution 🤔 im not implementing it tonight, but i think i know how. i'll follow up with you tomorrow or thursday about it.

MartijnHols: I would like to be able to see the impact of core changes.

MartijnHols: Maybe we need to set up optional tests with full snapshots, or a bot doing it in addition to the travis tests

emallson:
screenshot from 2018-11-21 12-37-58

src/parser/core/tests/CombatLogParserIntegrationTest.js
```diff
@@ -0,0 +1,85 @@
...
+
+      Object.values(parserClass.specModules).forEach(moduleClass => {
+        if(moduleClass instanceof Array) {
+          // cannot call parser._getModuleClass at this point in
+          // execution, so we handle the case manually
+          moduleClass = moduleClass[0];
+        }
+        describe(moduleClass.name, () => {
+          suppressLogging(suppressLog, suppressWarn, false);
+          it('matches the statistic snapshot', () => expectSnapshot(parser, moduleClass, integrationSnapshot));
```
MartijnHols: How about suggestions, items and tabs? Also does this verify checklist consistency?

src/parser/core/tests/snapshotTest.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  getChildContext() {
+    return {
+      parser: this.props.parser,
+    };
+  }
+
+  render() {
+    console.log(this.props.children);
```
MartijnHols: left a log

src/parser/core/tests/CombatLogParserIntegrationTest.js
emallson: I'm going to hold off on tabs because Chart.js errors out when rendered this way, so I can't easily do it automatically like I can with stats.

Next push will have checkllist & suggestions though.

scripts/test-tools/download-log.js
```diff
@@ -0,0 +1,72 @@
...
+  }).on('error', console.error);
+}
+
+function write_log(key, fight_id, player_id, cb, meta, combatants, events) {
+  const path = `test-logs/${key}.json.gz`;
+  const out = fs.createWriteStream(path);
+  const compress = zlib.createGzip();
+  compress.pipe(out);
+
+  compress.write(JSON.stringify({
```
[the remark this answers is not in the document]
emallson: it doesn't look like you can manipulate .zip files with the built-in `zlib` module. if you really want that feature, I can find & pull in a dependency for it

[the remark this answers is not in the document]
MartijnHols: Would be nice. Being able to open the zip and open only what you need should make it easier to use.

PR #2727, opened 2018-11-20
[Demonology] Pet tracking refactor and cleanup

sMteX:
The diff on this is ugly but the point is here:
- DemoPets.js was an unmaintainable mess, 440+ lines of glued up spaghetti code that somehow worked
- I broke it down into 6 smaller pieces, each of which takes care of **single** thing - Pet summoning, Wild Imp despawns due to energy, Demonic Tyrant extensions etc.
- these modules have been also updated to the new event listening system (apart from PetSummonHandler, which wasn't updated due to it listening to "spendresource" event)
- old `/pets/DemoPets.js` was moved to `/pets/DemoPets/index.js` so the imports are same
- new DemoPets.js now contains only `PetDamage` and `PetTimeline` objects, API methods and some common helper methods that I didn't want to duplicate across modules
- I also removed some helper arrays in `/pets/CONSTANTS.js`, instead I added props to the pets in `pets/PETS.js` and used those

### DIFF
--- a/src/parser/warlock/demonology/CombatLogParser.js
+++ b/src/parser/warlock/demonology/CombatLogParser.js
@@ -18,6 +18,7 @@ import PetSummonHandler from './modules/pets/DemoPets/PetSummonHandler';
 import WildImpEnergyHandler from './modules/pets/DemoPets/WildImpEnergyHandler';
 import PowerSiphonHandler from './modules/pets/DemoPets/PowerSiphonHandler';
 import DemonicTyrantHandler from './modules/pets/DemoPets/DemonicTyrantHandler';
+import ImplosionHandler from './modules/pets/DemoPets/ImplosionHandler';
 import PetTimelineTab from './modules/pets/PetTimelineTab';
 import PrepullPetNormalizer from './modules/pets/normalizers/PrepullPetNormalizer';
 
@@ -61,6 +62,7 @@ class CombatLogParser extends CoreCombatLogParser {
     wildImpEnergyHandler: WildImpEnergyHandler,
     powerSiphonHandler: PowerSiphonHandler,
     demonicTyrantHandler: DemonicTyrantHandler,
+    implosionHandler: ImplosionHandler,
     petTimelineTab: PetTimelineTab,
     prepullPetNormalizer: PrepullPetNormalizer,
 
--- a/src/parser/warlock/demonology/CombatLogParser.js
+++ b/src/parser/warlock/demonology/CombatLogParser.js
@@ -15,6 +15,7 @@ import SoulShardDetails from './modules/soulshards/SoulShardDetails';
 import DemoPets from './modules/pets/DemoPets';
 import PetDamageHandler from './modules/pets/DemoPets/PetDamageHandler';
 import PetSummonHandler from './modules/pets/DemoPets/PetSummonHandler';
+import WildImpEnergyHandler from './modules/pets/DemoPets/WildImpEnergyHandler';
 import PetTimelineTab from './modules/pets/PetTimelineTab';
 import PrepullPetNormalizer from './modules/pets/normalizers/PrepullPetNormalizer';
 
@@ -55,6 +56,7 @@ class CombatLogParser extends CoreCombatLogParser {
     demoPets: DemoPets,
     petDamageHandler: PetDamageHandler,
     petSummonHandler: PetSummonHandler,
+    wildImpEnergyHandler: WildImpEnergyHandler,
     petTimelineTab: PetTimelineTab,
     prepullPetNormalizer: PrepullPetNormalizer,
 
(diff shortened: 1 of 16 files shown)


sMteX:
Since nobody touched it for over a week I thought I could explain what's going on here, maybe that would help getting this at least reviewed. Starting from the top as it goes in the PR:

1) `CombatLogParser` - I don't think I need to explain anything here
2) `pets/CONSTANTS.JS` - bunch of removed helper arrays:
    - `PET_GUID_TO_SUMMON_ABILITY_MAP` was replaced with the `summonAbility` prop in `PETS.js`
    - `WILD_IMP_GUIDS` was replaced with helper method `isWildImp(guid)` in `pets/helpers.js`
    - `PETS_AFFECTED_BY_DEMONIC_TYRANT_GUIDS` was replaced with inlined array `_petsAffectedByDemonicTyrant` in `demonicTyrantHandler.js`
    - `RANDOM_PET_GUIDS` was replaced by `isRandom` prop in `PETS.js` and utilized also in `isRandomPet(guid)` in `pets/helpers.js`
3) `DemoPets.js` was:
    1) moved into subfolder and renamed to `DemoPets/index.js`
    2) butchered into 6 different submodules to make it clearer and easier to understand (each handler has a clear responsibility instead of one class doing *everything*)
4) `DemoPets/DemonicTyrantHandler.js`
    - takes care of extending pets with Demonic Tyrant and adjusting Wild Imp despawns when its buff is removed
    - corresponding lines in old DemoPets - 191 - 208, 278 - 295
    - `_initializeDemonicTyrantPets()` populates the helper array with guids of used pets that are extended with Demonic Tyrant
    - `onDemonicTyrantCast` - extends current affected pets by 15 seconds (defined in `TimelinePet.extend()`)
    - also, if the player has Demonic Consumption (T100 talent), it sacrifices all Wild Imps
    - `onDemonicPowerRemove` - when Demonic Tyrant buff expires, adjusts expected duration of current Wild Imps (their lifetime is "paused" when DT is active)
5) `DemoPets/ImplosionHandler.js`
    - handles sacrificing Wild Imps via Implosion
    - corresponding lines in old DemoPets - 102 - 153, 262 - 276
    - `onImplosionCast` - gets current Wild Imps and marks them as "imploded"
    - `onImplosionDamage` - tries to pair Wild Imps with the Implosion target - see comments for explanations
6) `DemoPets/PetDamageHandler.js`
    - takes care of tracking pet damage by individual pet guids and instances
    - corresponding lines in old DemoPets - 44 - 52
7) `DemoPets/PetSummonHandler.js`
    - takes care of tracking pet summons and adding them to the Timeline
    - corresponding lines in old DemoPets - 56 - 83, 85 - 87, 214 - 228, 280, 353 - 375, 413 - 425
    - `on_byPlayer_summon` - gathers information about the pet (name, guid, expected duration, summon *spell* and ability, if it's a Wild Imp then also few other properties), creates a new `TimelinePet` and pushes it into the Timeline
    - `on_byPlayer_cast` (for Demonic Tyrant), `on_byPlayer_spendresource` are needed for deciding the summon sources, while `damage, energize, heal and absorbed` events just update the player's current position
    - this is important, because Wild Imps need to track their position for Implosion casts
        - they have their position in their Fel Firebolt casts, but if they aren't allowed any casts, then we have no default value, so they default to player's position (they're summoned around him anyway)
    - `_getPetDuration` - takes the information from `PETS.js` object, but if it's a Wild Imp, it also takes into consideration Demonic Tyrant buff, whose remaining duration is added to the duration of the pet
    - `_getSummonSpell` - this is to distinguish "from where" the pet came - if it's a spell or a talent like Nether Portal or Inner Demons
8) `DemoPets/PowerSiphonHandler.js`
    - takes care of sacrificing Wild Imps with Power Siphon
    - corresponding lines in old DemoPets - 297 - 323
    - Power Siphon sacrifices up to 2 Wild Imps to generate stacks of Demonic Core (which is like a resource for us)
    - the problem is that while PS *should* sacrifice the 2 Wild Imps with lowest energy, usually it happened that the ones we picked by this criteria, were happily casting while supposedly dead
    - I added a normalizer that tracks which Imps were active *after* the cast, so I know which I *can't* sacrifice
    - so in `onPowerSiphonCast` I get current imps that aren't supposed to be imploded, sort them, filter those that shouldn't die, and sacrifice up to 2 Wild Imps
9) `DemoPets/WildImpEnergyHandler.js`
    - takes care of "natural" despawning of Wild Imps
    - corresponding lines in old DemoPets - 155 - 189, 377 - 386
    - Wild Imps despawn if they're killed in other ways, or they live up to 15 seconds, or their energy runs out (which is normally 5 casts, 100 energy, 20 per cast)
    - this module tracks their Fel Firebolt casts and if their current energy reaches 0, despawns them
10) `DemoPets/index.js`
    - remnant of the old monster class
    - currently contains only Timeline object and PetDamage object, API methods and some helper methods that are shared with the handlers
11) `PETS.js`
    - stayed the same, except for `summonAbility` and `isRandom` properties
12) `helpers.js`
    - added `isWildImp(guid)` and `isRandomPet(guid)` helper methods
13 - 17) Various files
    - replaced usages of removed helper arrays with added helper methods or properties in `PETS.js`

PR #2729, opened 2018-11-20
[Assassination] Update to new event listener

Gebuz:
Also changed energy cap to show percent instead of per minute, and added it to the checklist.


src/parser/rogue/assassination/modules/talents/ElaboratePlanning.js
```diff
@@ -20,11 +21,16 @@ class ElaboratePlanning extends Analyzer {
   constructor(...args) {
     super(...args);
     this.active = this.selectedCombatant.hasTalent(SPELLS.ELABORATE_PLANNING_TALENT.id);
+    if (!this.active) {
+      return;
+    }
+    ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.forEach(e => {
+      this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(e), this.addBonusDamageIfBuffed);
```
sMteX: Since `ABILITIES_AFFECTED_BY_DAMAGE_INCREASES` is array of spell objects itself, you should be able to do just 
```javascript
this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(ABILITIES_AFFECTED_BY_DAMAGE_INCREASES), this.addBonusDamageIfBuffed);
```
without the loop.

src/parser/rogue/assassination/modules/talents/MasterAssassin.js
```diff
@@ -23,16 +24,18 @@ class MasterAssassin extends Analyzer {
   constructor(...args) {
     super(...args);
     this.active = this.selectedCombatant.hasTalent(SPELLS.MASTER_ASSASSIN_TALENT.id);
+    if (!this.active) {
+      return;
+    }
+    ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.forEach(e => {
+      this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(e), this.addBonusDamageIfBuffed);
```
sMteX: Same as before

src/parser/rogue/assassination/modules/talents/MasterPoisoner.js
```diff
@@ -19,13 +20,15 @@ class MasterPoisoner extends Analyzer {
...
-  }
-
-  on_byPlayer_damage(event) {
-    const spellId = event.ability.guid;
-    if (!ABILITIES_AFFECTED_BY_POISON_DAMAGE_INCREASES.includes(spellId)) {
+    if (!this.active) {
       return;
     }
+    ABILITIES_AFFECTED_BY_POISON_DAMAGE_INCREASES.forEach(e => {
+      this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(e), this.addBonusDamage);
```
sMteX: Same as before

src/parser/rogue/shared/resources/EnergyCapTracker.js
```diff
@@ -54,15 +54,19 @@ class EnergyCapTracker extends RegenResourceCapTracker {
     return Math.floor(max);
   }
 
+  get wastedPercent() {
+    return this.missedRegen / this.naturalRegen || 0;
```
sMteX: I'm not sure by the operator precedency, I'd put parentheses to be sure
```suggestion
    return (this.missedRegen / this.naturalRegen) || 0;
```

src/parser/rogue/assassination/modules/talents/Nightstalker.js
```diff
@@ -25,28 +26,26 @@ class Nightstalker extends Analyzer {
...
-    if (!this.selectedCombatant.hasBuff(SPELLS.STEALTH.id)) {
+    if (!this.active) {
       return;
     }
-    const spellId = event.ability.guid;
-    if (!ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.includes(spellId)) {
-      return;
-    }
-    if (ABILITIES_AFFECTED_BY_SNAPSHOTTING.includes(spellId)) {
+    ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.forEach(e => {
```
sMteX: I'm not sure if comparing objects directly would work. Might be more elegant to do this:
```javascript
const allowedAbilities = ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.filter(spell => !NIGHTSTALKER_BLACKLIST.some(blacklisted => blacklisted.id === spell.id));
this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(allowedAbilities), this.addBonusDamageIfBuffed);
```
`some()` returns true if it finds any item satisfying the condition. So if it doesn't find the spell in the BLACKLIST with the same ID, it's allowed.

Gebuz: It works, they point to the same object. Anyway your solution is prettier, so I'll use that instead.

sMteX: I mean, if it works, you might as well simplify it even more with include
```javascript
const allowedAbilities = ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.filter(spell => !NIGHTSTALKER_BLACKLIST.includes(spell));
```

PR #2730, opened 2018-11-20
[gitignore] add webstorm 2018.3 encodings.xml to gitignore

Pewtro:
With the new 2018.3 update to webstorm a new filetype was added to the .idea/ folder. 
This makes sure it doesn't get pushed to mess with anyone elses configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ temp
 /.idea/vcs.xml
 /.idea/workspace.xml
 /.idea/codeStyles/*
+/.idea/encodings.xml
 /.vscode
 
 .vs/ProjectSettings.json


MartijnHols:
What does it do?

Pewtro:
This is all I know about it: 
https://www.jetbrains.com/help/webstorm/encoding.html

and this is my current file
```<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="Encoding" addBOMForNewFiles="with NO BOM" />
</project>``` 


But since we ignore all the other .idea files from webstorm, might aswell ignore this aswell

MartijnHols:
We only ignore some files as per the recommendation:

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

MartijnHols:
I want as much preconfigured as possible, so anyone can enjoy proper config out of the box.

PR #2741, opened 2018-11-22
[Frost Mage] Thermal Void rewrite

Dambroda:
Thermal Void is currently in the checklist using flat extension values to determine if the player used it properly. As it is entirely RNG and fight-dependent how long you can extend TV, I think it's better to not have a checklist/suggestion item and instead only include a statistic. Suggestions would probably be tied into FoF/brain freeze efficiency upgrades instead.

### Before:
image
image
image
(worth noting that in the sample provided, frozen orb was cast twice during Icy Veins; just goes to show that we really cannot use extension time as suggestion threshholds here)

### After:
**_No suggestion/checklist items_**

image

### DIFF
--- a/src/parser/mage/frost/CHANGELOG.js
+++ b/src/parser/mage/frost/CHANGELOG.js
@@ -11,7 +11,7 @@ export default [
     contributors: [Dambroda],
   },
   {
-    date: new Date('2018-11-27'),
+    date: new Date('2018-11-24'),
     changes: <>Added a statistics module for <SpellLink id={SPELLS.LONELY_WINTER_TALENT.id} />.</>,
     contributors: [Dambroda],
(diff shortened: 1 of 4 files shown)


Sharrq:
I'll look further into this tomorrow when I get home. I agree about removing the static extension threshold... I never really liked it and tried to make it a reasonable target, but rng can be dumb sometimes.

I think we could eventually look at the brain freeze and fog procs during iv to make sure they are all being used and make sure every iv has at least one frozen orb. But that's for later I think

Sharrq:
I dont know enough about JS to understand what all you are doing in here, but the screenshots and the stat box and everything look fine. So if someone else can look at your code and make sure its good then i approve

PR #2738, opened 2018-11-22
[Affliction] Inevitable Demise trait

sMteX:
image

I **intentionally** didn't add changelog entry as I plan to do more traits (probably today) and I'd rather have 1 entry like "added traits X, Y, Z" than 3 entries for same date (feels less spammy this way and easier to resolve merge conflicts). If you insist on having 3 entries though, I'll make it happen

Issue #2724, opened 2018-11-19
[Warlock] Implement Azerite traits


sMteX:
- [x] Affliction:
    - [x] ~~[Inevitable Demise](https://www.wowhead.com/spell=273521/inevitable-demise) (is changed in 8.1) - #2738~~ 
    - [x] ~~[Sudden Onset](https://www.wowhead.com/spell=278721/sudden-onset) - #2747~~ 
    - [x] ~~[Wracking Brilliance](https://www.wowhead.com/spell=272891/wracking-brilliance) - #2739~~
    - [x] ~~[Cascading Calamity](https://www.wowhead.com/spell=275372/cascading-calamity) - #2740~~
    - [x] ~~[Dreadful Calling](https://www.wowhead.com/spell=278727/dreadful-calling) (damage tuned in 8.1) - #2744~~
    - [ ] ~~[Deathbloom](https://www.wowhead.com/spell=275974/deathbloom) (lol, also is removed in 8.1)~~
- [x] Demonology:
    - [x] ~~[Demonic Meteor](https://www.wowhead.com/spell=278737/demonic-meteor) - #2749~~ 
    - [x] ~~[Supreme Commander](https://www.wowhead.com/spell=279878/supreme-commander) - #2750~~ 
    - [x] ~~[Explosive Potential](https://www.wowhead.com/spell=275395/explosive-potential) - #2751~~ 
    - [x] ~~[Shadow's Bite](https://www.wowhead.com/spell=272944/shadows-bite) - #2754~~ 
    - [x] ~~[Umbral Blaze](https://www.wowhead.com/spell=273523/umbral-blaze) - #2752~~ 
    - [x] ~~[Forbidden Knowledge](https://www.wowhead.com/spell=278738/forbidden-knowledge) (is removed in 8.1) - #2754~~ 
- [x] Destruction:
    - [x] ~~[Accelerant](https://www.wowhead.com/spell=272955/accelerant) (is removed in 8.1) - #2758~~ 
    - [x] ~~[Bursting Flare](https://www.wowhead.com/spell=279909/bursting-flare) - #2760~~ 
    - [x] ~~[Chaotic Inferno](https://www.wowhead.com/spell=278748/chaotic-inferno) - #2765~~ 
    - [x] ~~[Crashing Chaos](https://www.wowhead.com/spell=277644/crashing-chaos) - #2765~~ 
[...]

Issue #2088, opened 2018-08-11
Azerite Traits with fixed damage/healing values


joshinat0r:
Some azerite traits deal a fixed amount of damage (like [Zealotry](https://www.wowhead.com/spell=278615/zealotry) for ret pallys), but only using the value from our azerite-calculation might lead to wrong numbers. 
All % damage modifiers on targets and player would cause numbers that are off, ~~which means we need a module that will keep track of those buffs and debuffs.~~

Things that the module should account for:
- multipliers on player
- multipliers on bosses
- armor if the damage type of the trait is physical (would also need to account for shifting K-values of bosses)
- versatility

Tracking other buffs might be more difficult, such as the raid wide target debuffs applied by monks and demon hunters as we dont have them in the events. The same thing applies to healers and target specific % healing modifiers (like Vampiric Blood).

**EDIT**
As Gebuz mentioned, a proper solution to account for untrackable buffs (untrackable as in we don't want to make new API calls) is to calculate the expected damage of the spell. Check [this](https://github.com/WoWAnalyzer/WoWAnalyzer/issues/2088#issuecomment-414596614) comment on how we would calculate the expected hit size.

Gebuz:
What we could do is calculate the expected hit size and then multiply that based on the actual hit. For instance, let's say I have a trait that adds X dmg to Solar Wrath. I cast Solar Wrath, and calculate that it should hit for Y dmg. If the hit matches that and also hits for Y I can simply contribute the damage the trait added to that. If it hits for more or less, let's call it Z, I just get the multiplier (Z/Y) and multiply the trait damage by that. 
```this.totalDamage += traitBonusDamage * actualDamage / expectedDamage```

HawkCorrigan:
I started doing such a thing (damageModifyTracker) a while back and the general complaint was that it would generate more api calls.
IF I understood it correctly you would need an additional api call per enemy to get debuffs on them.

joshinat0r:
I think getting the expected hit size right is a lot of work. It would be important to have all stats right, as in: keep track of mastery for tanks (to increase AP), all primary stat proc's from trinkets and azerite traits and spec specific stuff (like Shaman's mastery if affected). That's quite some work.

And yeah, making new API calls would give us the events we need to track them, but I think we might get issues with the API-limit, especially with BfA launch.

Gebuz:
[...]

PR #2744, opened 2018-11-23
[Affliction] Dreadful Calling trait

sMteX:
image

Part of #2724. Added a helper function that should *estimate* the bonus part of Azerite traits that affect damage of spells (related issue: #2088). 

Assuming following:
- Azerite traits of this type affect *base* damage of spells
- For casters, this damage is determined by Intellect (and thus Spell Power) and the spell's SP coefficient
- so if `damage = (base + trait bonus) * coefficients`, that means both base and the trait bonus is scaled by same coefficients (like mastery, versatility etc.)
- and even if both damage parts gets scaled by the coefficients, their ratio stays the same regardless of it

We should be able to at least estimate the damage bonus with the following:
```
trait bonus = value that affects the damage
base damage = SP coefficient * current intellect
portion = trait bonus / (base damage + trait bonus)
bonus damage = real damage * portion
```
Example for Unstable Affliction and Dreadful Calling trait (assuming 100 bonus damage from trait, 6000 intellect, 0.16 SP coefficient and 4000 damage from event):
```
traitBonus = 100
intellect = 6000
SPcoeff = 0.16
actualDamage = 4000

baseDamage = 0.16 * 6000 = 960
portion = 100 / (960 + 100) = 0.094339
bonusDamage = 0.094339 * 4000 = 377.356
```

My source from which I got the idea for this, thinks this should be applicable to all Azerite traits (at least the Warlock ones), so I'd like to merge this before updating #2738 in order to not duplicate code in it as well. If others think this should apply to other specs/classes, it might be good idea to move it somewhere else into `common` probably.

### DIFF
--- a/src/common/SPELLS/bfa/azeritetraits/warlock.js
+++ b/src/common/SPELLS/bfa/azeritetraits/warlock.js
@@ -15,4 +15,9 @@ export default {
     name: 'Wracking Brilliance',
     icon: 'spell_shadow_felmending',
   },
+  DREADFUL_CALLING:{
+    id: 278727,
+    name: 'Dreadful Calling',
+    icon: 'inv_beholderwarlock',
+  },
 };
--- a/src/parser/core/calculateBonusAzeriteDamage.js
+++ b/src/parser/core/calculateBonusAzeriteDamage.js
@@ -6,11 +6,12 @@
  * @returns {number} Bonus damage contributed by the trait
  */
 export default function calculateBonusAzeriteDamage(event, traitBonus, ...coefficients) {
-  // In order to be general and not weird to pass attack power into "intellect" field, it's in the coefficients array (which all get multiplied together to form the base damage)
-  const baseDamage = coefficients.reduce((total, current) => total * current, 1);
-  // gets what percentage trait bonus is in the modified base damage
-  const traitPortion = traitBonus / (baseDamage + traitBonus);
   const actualDamage = event.amount + (event.absorbed || 0);
-  // scales down the actual damage according to the percentage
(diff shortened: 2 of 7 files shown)


Gebuz:
Yep, that's how all azerite traits work, so please do add the helper function in to /parser/core. 

Racials like Blood Fury still give AP/SP instead of primary stat iirc? Might need to use that instead of int.

For this trait, I think it should be made more obvious that the CDR is not included in the DPS estimate, and maybe rewrite the line about the int to put emphasis on it only being as accurate as our stat tracker (I still think we are missing a bunch of trinkets?)

sMteX:
I actually used the formula that was in Feeding Frenzy (Beast Mastery Hunter) and Wild Fleshrending (Feral Druid), it seemed more elegant. Function was moved to `parser/core/calculateBonusAzeriteDamage.js` and documented. I also refactored the said 2 modules to use this new function as well (tested on https://www.warcraftlogs.com/reports/TbA6KcXYCLtZd4D1/#fight=2&source=2 and https://www.warcraftlogs.com/reports/bCPHf81hrnyFZp2K/#fight=3&source=4, nothing changed in the output).

Proof that the 2 (my original and the new formula) are the same:
Assuming `total = (base + bonus) * modifiers`
Original:
 ```
ratio = (bonus * modifiers) / ((base + bonus) * modifiers) = bonus / (base + bonus)
Result = total * ratio = (total * bonus) / (base + bonus)
```
New: 
```
modifiers = total / (base + bonus)
Result = bonus * modifiers = (bonus * total) / (base + bonus)
```
Also, updated tooltip for Dreadful Calling
image

PR #2747, opened 2018-11-23
[Affliction] Sudden Onset trait

sMteX:
image

The damage is a little high because it was from Mythic Zul, so I assume heavy multi-dotting, usually it's a lot lower. Part of #2724

### DIFF
--- a/src/parser/warlock/affliction/modules/azerite/SuddenOnset.js
+++ b/src/parser/warlock/affliction/modules/azerite/SuddenOnset.js
@@ -1,0 +1,60 @@
+import React from 'react';
+
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
+import StatTracker from 'parser/shared/modules/StatTracker';
+import calculateBonusAzeriteDamage from 'parser/core/calculateBonusAzeriteDamage';
+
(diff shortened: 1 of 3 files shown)


src/parser/warlock/affliction/modules/azerite/SuddenOnset.js
```diff
@@ -0,0 +1,60 @@
...
+  traitBonus = 0;
+  damage = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.SUDDEN_ONSET.id);
+    if (!this.active) {
+      return;
+    }
+    this.traitBonus = this.selectedCombatant.traitsBySpellId[SPELLS.SUDDEN_ONSET.id].reduce((total, rank) => {
```
Gebuz: is the value of the trait the bonus dmg at max stacks? If that is the case you might need to multiply the sp coef by max stack count (10?).

sMteX: No, I've been told this is the base damage bonus, and the stacking mechanic of Agony is just another multiplier on top of Mastery, Versatility, spec aura etc. So additional stacks will scale the base damage bonus further same as normal Agony damage.

PR #2749, opened 2018-11-23
[Demonology] Demonic Meteor trait

sMteX:
image
image

Part of #2724. Although Hand of Gul'dan damage scales with shards used, it's also as a multiplicative bonus, same as versatility for example, so it's not included.


src/parser/shared/modules/StatTracker.js
```diff
@@ -291,7 +291,6 @@ class StatTracker extends Analyzer {
     [SPELLS.BLUR_OF_TALONS_BUFF.id]: BLUR_OF_TALON_STATS,
     // endregion
     // region Warlock
-    [SPELLS.EXPLOSIVE_POTENTIAL.id]: { haste: 841 },
```
Dambroda: Is this intentional?

sMteX: Yes, it's yet another trait so fixed haste on its own is wrong, and it's added correctly in #2751

PR #2754, opened 2018-11-24
[Demonology] Shadow's Bite and Forbidden Knowledge traits

sMteX:
image
image

Part of #2724. This one was a little complicated, because both of these traits can affect Demonbolt at the same time (there's a very small window, around 3 seconds where this can happen, but I've tested it and it works) + they snapshot the effects on cast, so spell queue was also necessary. 

Because of this, I also had to change the signature of `calculateBonusAzeriteDamage` (and change all existing occurences):
- `coefficients` has to be an explicitly stated array
- there's an optional array parameter `knownOtherEffects` that contains ADDITIVE other trait bonuses
- resulting base damage is then multiplication of all coefficients (like Spell Power, SP coefficients etc.) + sum of all known effects

I'm aware that some of my other opened PRs use the old signature - I'll change them as soon as this PR is merged.

Tested thoroughly on this log: `/report/pbHAc1R3tN94VCQ7/22-LFR+Taloc+-+Kill+(4:44)/249-Velerius`

### DIFF
--- a/src/common/SPELLS/bfa/azeritetraits/__UNUSED.js
+++ b/src/common/SPELLS/bfa/azeritetraits/__UNUSED.js
@@ -789,11 +789,6 @@ export default {
     name: 'Serrated Jaws',
     icon: 'ability_hunter_killcommand',
   },
-  SHADOWS_BITE: {
-    id: 272944,
-    name: 'Shadow\'s Bite',
-    icon: 'spell_shadow_shadowbolt',
-  },
   SHELLSHOCK: {
     id: 274357,
     name: 'Shellshock',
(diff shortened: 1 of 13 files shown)


sMteX:
Signature changed again to more pleasant one:
- `traitBonus` can now be an array of the trait effects
- function now always returns an array of bonus damages, allowing for 2 types of usages:
    - `const [ bonusForSingleTrait ] = calculateBonusAzeriteDamage(damageEvent, traitBonus, coefficient1, coefficient2....)`
    - `const [ bonusForTrait1, bonusForTrait2, ... ] = calculateBonusAzeriteDamage(damageEvent, [traitBonus1, traitBonus2, ...], coefficient1, coefficient2...)`
- in both cases, `traitBonus` means **additive** bonuses to base damage:
    - resulting base damage = **multiplication** of `coefficients` + **sum** of `traitBonuses`

PR #2757, opened 2018-11-25
[Prot Pally] Changed Hammer of the Righteous to track non-filler casts

emallson:
Hammer of the Righteous is a filler spell, doing almost no damage relative to Judgment and Avenger's Shield. It is *always* correct to let it sit at two charges when you have either of the others to cast, including when you switch to the Blessed Hammer talent (which replaces HotR).

This changes the HotR/BH checklist item and suggestion from tracking cast efficiency to tracking bad (non-filler) casts. As far as I know, there isn't a standard set of tools for doing this so I implemented it manually.

I chose to use `isOnCooldown` rather than `isAvailable` deliberately, because it can be correct to use HotR over your second charge of Judgment to put HotR on CD when running the Crusader's Judgment talent (which gives Judgment 2 charges).

screenshot from 2018-11-25 15-50-06
screenshot from 2018-11-25 15-49-54

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -2118,12 +2118,6 @@
       "resolved": "https://registry.npmjs.org/dottie/-/dottie-2.0.0.tgz",
       "integrity": "sha1-2hkZgci41xPKARXViYzzl8Lw3dA="
     },
-    "duplexer": {
-      "version": "0.1.1",
-      "resolved": "https://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
-      "integrity": "sha1-rOb/gIwc5mtX0ev5eXessCM0z8E=",
-      "dev": true
-    },
     "duplexer3": {
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/duplexer3/-/duplexer3-0.1.4.tgz",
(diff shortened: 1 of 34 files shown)


src/parser/paladin/protection/modules/features/Checklist/Module.js
```diff
@@ -30,13 +32,15 @@ class Checklist extends BaseChecklist {
         castEfficiency={this.castEfficiency}
         extras={{
           lotpAbility: this.lotp._activeSpell,
+          hotrAbility: this.hotr._activeSpell,
```
Dambroda: My understanding was that the underscore variable prefix basically meant that it shouldn't be accessed from outside the class -- is this intentional?

src/parser/paladin/protection/modules/features/Checklist/Component.js
```diff
@@ -39,8 +39,13 @@ class ProtectionPaladinChecklist extends React.PureComponent{
           <AbilityRequirement spell={SPELLS.SHIELD_OF_THE_RIGHTEOUS.id} />
           <AbilityRequirement spell={SPELLS.AVENGERS_SHIELD.id} />
           <AbilityRequirement spell={SPELLS.JUDGMENT_CAST_PROTECTION.id} />
-          {combatant.hasTalent(SPELLS.BLESSED_HAMMER_TALENT.id) && <AbilityRequirement spell={SPELLS.BLESSED_HAMMER_TALENT.id} />}
-          {!combatant.hasTalent(SPELLS.BLESSED_HAMMER_TALENT.id) && <AbilityRequirement spell={SPELLS.HAMMER_OF_THE_RIGHTEOUS.id} />}
+          <Requirement
+            name={(
```
Dambroda: Parenthesis could be removed and fragment on one line; but I am not sure which style is preferred

src/parser/paladin/protection/modules/features/Checklist/Module.js
emallson: you're right, i'll fix these.

emallson:
@Dambroda any more comments?

PR #2761, opened 2018-11-26
[Destruction] Rolling Havoc trait

sMteX:
image

Part of #2724. Suffers from same problems as #2760

### DIFF
--- a/src/parser/warlock/destruction/modules/azerite/RollingHavoc.js
+++ b/src/parser/warlock/destruction/modules/azerite/RollingHavoc.js
@@ -1,7 +1,6 @@
 import React from 'react';
 
-import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
-import EventFilter from 'parser/core/EventFilter';
+import Analyzer from 'parser/core/Analyzer';
 
 import SPELLS from 'common/SPELLS';
 import { calculateAzeriteEffects } from 'common/stats';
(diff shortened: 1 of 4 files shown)


src/parser/warlock/destruction/modules/azerite/RollingHavoc.js
```diff
@@ -0,0 +1,83 @@
...
+  onFinished(event) {
+    if (this._currentStacks !== 0) {
+      const uptimeOnStack = event.timestamp - this._lastChangeTimestamp;
+      debug && this.log(`Fight ended, adding rest of RH uptime on ${this._currentStacks} stacks: ${uptimeOnStack}`);
+      this.totalIntellect += this._currentStacks * this.intellect * uptimeOnStack;
+    }
+  }
+
+  get averageIntellect() {
+    return (this.totalIntellect / this.owner.fightDuration).toFixed(0);
```
Dambroda: Rather than calculating up the total intellect yourself, you should use `getStackWeightedBuffUptime / fightDuration` (which gives you the average stacks across the encounter), and then multiply by `this.intellect` to get the average intellect; this removes the need for the event listeners entirely :)

Dambroda: Example: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/shared/modules/items/bfa/raids/uldir/SyringeOfBloodborneInfirmity.js#L45

PR #2767, opened 2018-11-27
[Server] Remove bad dependency from event-stream and update version

Dambroda:
Just testing if this resolves Travis errors.  Travis is bugging out now that flatmap has been removed from npm (see below). New version of event-stream has flatmap removed. I just manually updated the version number and removed the dep to match the npm description, but I assume there's a command I should have run instead that also updates the crc? 

Cause: https://github.com/dominictarr/event-stream/issues/115

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -2329,8 +2329,8 @@
       }
     },
     "event-stream": {
-      "version": "3.3.6",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.6.tgz",
+      "version": "4.0.1",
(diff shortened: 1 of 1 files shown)


MartijnHols:
cd server
npm install
npm audit fix

MartijnHols:
fixed by 0ec23b772abeadd57fd56e46ede439ed7f442de5

I wasn't sure if your PR had really fixed the issue

Dambroda:
fully expected it, I dont really know how to resolve any npm issues that may come up 👍

MartijnHols:
Might have been that it fixed things, just missing the guarantees. Especially since v4 of the package was published by the same person who published the compromised version. And a major version bump generally breaks something.

PR #2769, opened 2018-11-27
[Fury] Frothing Berserker rework + Juggernaut removal + Enrage

wisnoi:
Juggernaut no longer exists, and Frothing Berserker has been properly reworked and statistics implemented

Because of bad commit practice, Enrage has to be lumped here as well. Yikes.


src/parser/warrior/fury/modules/buffdebuff/Enrage.js
```diff
@@ -0,0 +1,96 @@
...
+        minor: 0.7,
+        average: 0.65,
+        major: 0.6,
+      },
+      style: 'percentage',
+    };
+  }
+
+  suggestions(when) {
+    const {
```
Sharrq: I dont think you need this part here, all that gets done automatically if you do the below

suggestions(when) {
    when(this.suggestionThresholds)

src/parser/warrior/fury/modules/buffdebuff/Enrage.js
```diff
@@ -0,0 +1,96 @@
...
+      this.damage += calculateEffectiveDamage(event, this.statTracker.currentMasteryPercentage);
+      this.totalDamage += event.amount;
+    }
+  }
+
+  get uptime() {
+    return this.selectedCombatant.getBuffUptime(SPELLS.ENRAGE.id) / this.owner.fightDuration;
+  }
+
+  get damageTotal () {
```
Sharrq: This is a little nit picky, but you are calling this damageTotal but you arent actually outputing a total amount of damage, you are outputting the Damage Per Second increase. Might want to rename this to be more descriptive of what you are actually getting here.

src/parser/warrior/fury/modules/spells/MissedRampage.js
```diff
@@ -2,57 +2,88 @@ import React from 'react';
...
+      };
+    }
   }
 
   suggestions(when) {
     when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) => {
       return suggest(
         <>
-          There were {actual} times when you cast another ability when you should have cast <SpellLink id={SPELLS.RAMPAGE.id} />.
+          There were {actual} times you casted a rage generating ability when you should have cast <SpellLink id={SPELLS.RAMPAGE.id} />.
```
Sharrq: I would also include information on why they should have cast Rampage instead. This will help players that are new to the spec to understand why they are doing certain things or why it is a mistake to not do so.

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    damage = 0;
+
+    constructor(...args) {
+        super(...args);
+
+        this.active = false;
+
+        if (this.selectedCombatant.hasTalent(SPELLS.FROTHING_BERSERKER_TALENT.id)) {
```
Sharrq: You can shorten this to the following

this.active = this.selectedCombatant.hasTalent(SPELLS.FROTHING_BERSERKER_TALENT.id);
this.addEventListener(Events.damage.by(SELECTED_PLAYER, this.onPlayerDamage);

dont need the if statement for the event listener cause if its not active it wont run anyway.

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,112 @@
...
+            minor: 0.65,
+            average: 0.6,
+            major: 0.55,
+          },
+          style: 'percentage',
+        };
+      }
+
+    suggestions(when) {
+        const {
```
Sharrq: Again, you dont need to do this, the info will get pulled from the suggestion thresholds if you do "when(this.suggestionThresholds)"

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
```diff
@@ -0,0 +1,112 @@
...
+            isLessThan: {
+              minor,
+              average,
+              major,
+            },
+          } = this.suggestionThresholds;
+
+        when(true).isTrue().addSuggestion((suggest) => {
```
Sharrq: Not sure about other people, but personally im not a huge fan of telling the player its a mistake to use the talent they picked unless one talent is vastly greater than the other. Im not going to tell you to change it cause i have no idea how close these talents actually are .... and its just my personal opinion as someone who believes players should play the way they want to play (if the talents are close). Feel free to leave it in though, I know something similar was added for Rune of Power for mages so its fine to have it. I would say just make sure you are only doing this when it is important and not for every talent that isnt the "Cookie cutter best build"

src/parser/warrior/fury/modules/spells/MissedRampage.js
```diff
@@ -2,57 +2,88 @@ import React from 'react';
...
 
-const GENERATORS = [
-  SPELLS.RAGING_BLOW.id,
-  SPELLS.BLOODTHIRST.id,
-  SPELLS.EXECUTE.id,
+/*  Example log:
+ *  https://www.warcraftlogs.com/reports/KhynM7v96cZkTBdg#fight=6&type=damage-done&source=78
+ */
+
+const rageGenerators = [
```
Sharrq: i believe the standard for constants declared up top is to use all caps, so RAGE_GENERATORS instead of rageGenerators. Not sure if that is a hard rule or not, just what ive seen in all the modules ive looked at.

src/parser/warrior/fury/modules/buffdebuff/Enrage.js
matardarix: works only if you add `actual: this.enrageUptime, ` inside the suggestion thresholds tho

Sharrq: yea i missed that. Here is an example from one of the mage modules

  get utilSuggestionThresholds() {
    return {
      actual: this.utilPercent,
      isLessThan: {
        minor: 0.95,
        average: 0.90,
        major: 0.80,
      },
      style: 'percentage',
    };
  }

src/parser/warrior/fury/modules/talents/FrothingBerserker.js
wisnoi: There is absolutely unequivocally no situation in which this talent is an increase in DPS, ever. It's a non-choice. It's not so much cookie cutter as it has "gimping yourself for less DPS and clunkier gameplay". I'd love to see frothing actually be good, but it really isn't. :/

src/parser/warrior/fury/modules/buffdebuff/Enrage.js
Sharrq: basically change this

```
suggestions(when) {
    const {
      isLessThan: {
        minor,
        average,
        major,
      },
    } = this.suggestionThresholds;
     when(this.enrageUptime).isLessThan(minor)
``

to this
```
suggestions(when) {
    when(this.suggestionThresholds)
```

and add `actual: this.enrageUptime,` above `isLessThan: {` in the suggestionThresholds

File: src/parser/warrior/fury/modules/talents/FrothingBerserker.js (hunk 6 above)
Comment (Sharrq) [reply src/parser/warrior/fury/modules/talents/FrothingBerserker.js]: Thats fine. Then absolutely leave it in.

File: src/parser/warrior/fury/modules/talents/FrothingBerserker.js [hunk 8]
```diff
@@ -0,0 +1,99 @@
...
+            return suggest(
+              <>
+              You are currently using <SpellLink id={SPELLS.FROTHING_BERSERKER_TALENT.id} />. It is highly recommended to talent into <SpellLink id={SPELLS.CARNAGE_TALENT.id} />, as there are currently no situations in which  <SpellLink id={SPELLS.FROTHING_BERSERKER_TALENT.id} /> outperforms <SpellLink id={SPELLS.CARNAGE_TALENT.id} />.
+              </>
+            )
+            .icon(SPELLS.CARNAGE_TALENT.icon)
+            .staticImportance(SUGGESTION_IMPORTANCE.REGULAR);
+        });
+    
+        when(this.uptimePercent).isLessThan(0.65)
```
Sharrq: same as the Enrage Module, this can be `when(this.suggestionThresholds)` instead

PR #2770, opened 2018-11-27
[Fury Warrior] Endless Rage talent + Simmering Rage trait 

wisnoi:
Depends on #2769 

Implements rage gen counter for Endless Rage and dmg/rage counter for Simmering Rage.
image 
image


Sharrq:
looks like you have some conflicts to fix here now.

src/parser/warrior/fury/modules/talents/EndlessRage.js
```diff
@@ -0,0 +1,36 @@
...
+
+class EndlessRage extends Analyzer {
+    rageGen = 0;
+    
+    constructor(...args) {
+        super(...args);
+
+        this.active = this.selectedCombatant.hasTalent(SPELLS.ENDLESS_RAGE_TALENT.id);
+
+        this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).to(SELECTED_PLAYER), this.onPlayerBuff);
```
Dambroda: You only actually need one of the to/by filters :)

src/parser/warrior/fury/modules/azerite/SimmeringRage.js
```diff
@@ -0,0 +1,71 @@
...
+
+        const damage = simmeringRageDamage(this.selectedCombatant.traitsBySpellId[SPELLS.SIMMERING_RAGE.id]);
+        this.damage = damage;
+
+        this.addEventListener(Events.energize.by(SELECTED_PLAYER).to(SELECTED_PLAYER), this.onPlayerEnergize);
+        this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onPlayerDamage);
+    }
+
+    onPlayerEnergize(event) {
+        if (event.ability.guid === 278841) {
```
Dambroda: All spell ids should be in the spec modules, if it isn't there already you should create one

src/parser/warrior/fury/modules/azerite/SimmeringRage.js
```diff
@@ -0,0 +1,71 @@
...
+            this.rageGen += 1;
+        }
+    }
+
+    onPlayerDamage(event) {
+        if (!this.rampage.includes(event.ability.guid)) {
```
Dambroda: If you're using the new event listeners, instead of doing .includes you should add the .spell filter to the event listener.

e.g `this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(this.rampage), this.onPlayerDamage);`

However, also see other comment about the rampage array declaration

src/parser/warrior/fury/modules/azerite/SimmeringRage.js
```diff
@@ -0,0 +1,71 @@
...
+const simmeringRageDamage = traits => Object.values(traits).reduce((obj, rank) => {
+    const [ damage ] = calculateAzeriteEffects(SPELLS.SIMMERING_RAGE.id, rank);
+    obj.damage += damage;
+    return damage;
+  }, {
+    damage: 0,
+  });
+
+class SimmeringRage extends Analyzer {
+    rampage = [SPELLS.RAMPAGE_1.id, SPELLS.RAMPAGE_2.id, SPELLS.RAMPAGE_3.id, SPELLS.RAMPAGE_4.id];
```
Dambroda: Things like this should be moved to constants above the class and renamed to fit that convention:
`const RAMPAGE = [...]`

src/parser/warrior/fury/modules/azerite/SimmeringRage.js
```diff
@@ -0,0 +1,71 @@
...
+    simmeringDamage = 0;
+    damage = 0;
+    rageGen = 0;
+
+    constructor(...args) {
+        super(...args);
+
+        this.active = this.selectedCombatant.hasTrait(SPELLS.SIMMERING_RAGE.id);
+
+        const damage = simmeringRageDamage(this.selectedCombatant.traitsBySpellId[SPELLS.SIMMERING_RAGE.id]);
```
Dambroda: Slightly strange to declare this const, it actually ends up taking more linespace than just using `this.damage = simmeringRageDamage...`

If it's for some specific stylistic choice it's not a requirement to change it.

src/parser/warrior/fury/modules/talents/EndlessRage.js
```diff
@@ -0,0 +1,36 @@
...
+
+class EndlessRage extends Analyzer {
+  rageGen = 0;
+    
+  constructor(...args) {
+    super(...args);
+
+    this.active = this.selectedCombatant.hasTalent(SPELLS.ENDLESS_RAGE_TALENT.id);
+
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).to(SELECTED_PLAYER), this.onPlayerBuff);
```
Dambroda: Similarly, `.spell(SPELLS.ENRAGE)` performs the same functionality as L20-22

src/parser/warrior/fury/modules/azerite/SimmeringRage.js
```diff
@@ -0,0 +1,71 @@
...
+        this.rageGen += 1;
+      }
+    }
+
+    onPlayerDamage(event) {
+      if (!this.rampage.includes(event.ability.guid)) {
+        return;
+      }
+
+      this.simmeringDamage += this.damage;
```
Dambroda: Unfortunately, it's slightly more complicated than this when calculating azerite trait damage
See: #2088 
See: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/2749/files#diff-bd998464da96f154c459eeee0fd788c6R48
Unfortunately with rampage because it has 4 different abilities, it slightly complicates the calculations. The coefficients are here:
https://www.wowhead.com/spell=184367/rampage

Because this is a rather big change that will take some research I won't have this block the PR if you think it will take a while to implement

sMteX: Actually the correct syntax for this is here 
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/warlock/demonology/modules/azerite/DemonicMeteor.js#L48

PR #2774, opened 2018-11-27
[Bug] [Restoration Druid] Wild Growth fix

joshinat0r:
`this.wgTracker.wgBuffs.push(event.targetID);` in `on_byPlayer_applybuff` would fail if an applyBuff event would appear before the first cast-event.
This is just initializing the object with the necessary props to avoid crashes, not sure why applyBuffs apear before the castEvent, but this fix shouldn't break the numbers except the `startTimestamp` of the first WG.

@buimichael should give this a closer look to see if everything is still working as intended

### DIFF
--- a/src/parser/druid/restoration/modules/features/WildGrowth.js
+++ b/src/parser/druid/restoration/modules/features/WildGrowth.js
@@ -35,9 +35,6 @@ class WildGrowth extends Analyzer {
       return;
     }
 
-    console.info(event);
-    console.info(this.wgTracker);
-
     if(Object.getOwnPropertyNames(this.wgTracker).length > 0) {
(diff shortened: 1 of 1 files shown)


buimichael:
Oh and also, `startTimestamp` should probably be initialized with the start timestamp of the encounter. Otherwise the if-statement `if(event.timestamp - this.wgTracker.startTimestamp < PRECAST_PERIOD) ` will always fail for this particular use case.

joshinat0r:
How could `&& Object.getOwnPropertyNames(this.wgTracker.wgBuffs).length > 0` in `on_cast` fix the issue if the error is caused because `applyBuff` is applied before the first cast event? 😄 

But I get your point with the first cast, this fix is just ignoring the weird precast.
Not sure how accurate 0.35 avg hits per WG are for this log: report/qDapJ4C7H8TBdm2h/30-Mythic+Mythrax+-+Wipe+27+(5:58)/18-Hotsfourtots

But I'll leave this up to you, is at least going to prevent the crash

PR #2777, opened 2018-11-27
[Retribution] Bad Inquisition casts and other minor changes

Juko8:
Inquisition casts are marked as inefficient if used during Avenging Wrath
billede
Fixed a few things in Abilities
Updated a few things to use new event listeners


src/parser/paladin/retribution/modules/talents/Inquisition.js
```diff
@@ -18,20 +19,29 @@ class Inquisition extends Analyzer {
...
 
-  on_byPlayer_damage(event) {
-    const spellId = event.ability.guid;
-    if (!ABILITIES_AFFECTED_BY_DAMAGE_INCREASES.includes(spellId)) {
-      return;
+  onInquisitionCast(event) {
+    if (this.selectedCombatant.hasBuff(SPELLS.AVENGING_WRATH.id)) {
+      event.meta = event.meta || {};
+      event.meta.isInefficientCast = true;
+      event.meta.inefficientCastReason = 'You refreshed Inquisition during Avenging Wrath';
```
Sharrq: You might want to include some info about why you shouldnt do this, especially for those new to the spec.

PR #2786, opened 2018-11-28
[Arms] Crushing Assault

matardarix:
Added Crushing assault statistic module 
image

### DIFF
--- a/src/parser/warrior/arms/modules/spells/azeritetraits/CrushingAssault.js
+++ b/src/parser/warrior/arms/modules/spells/azeritetraits/CrushingAssault.js
@@ -4,7 +4,7 @@ import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
 import TraitStatisticBox, { STATISTIC_ORDER } from 'interface/others/TraitStatisticBox';
 import React from 'react';
 import Events from 'parser/core/Events';
-import { formatNumber, formatPercentage } from 'common/format';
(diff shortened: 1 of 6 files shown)


src/parser/warrior/arms/modules/spells/azeritetraits/CrushingAssault.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  statistic() {
+    const damageThroughputPercent = this.owner.getPercentageOfTotalDamageDone(this.crushingAssaultDamage);
+    const dps = this.crushingAssaultDamage / this.owner.fightDuration * 1000;
+    return (
+      <TraitStatisticBox
+        position={STATISTIC_ORDER.OPTIONAL()}
+        trait={SPELLS.CRUSHING_ASSAULT_TRAIT.id}
+        value={`${formatPercentage(damageThroughputPercent)} % / ${formatNumber(dps)} DPS`}
```
sMteX: This line, together with `damageThroughputPercent` and `dps` could be simplified to just `this.owner.formatItemDamageDone(this.crushingAssaultDamage)`, it should output exactly the same thing.
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/core/CombatLogParser.js#L527

matardarix: Oh nice, i also changed it in the Seismic Wave file

Issue #2678, opened 2018-11-15
Show a proper Icon & Name for Combat Potions in the timeline


Gebuz:
Currently it always displays the INT pot, which can be confusing if you used another pot. The ideal solution would be to show the first potion used, but if that isn't possible an alternative solution might be to just show a "fake" spell with a generic potion icon and a name like "Combat Potion".

Alastair-Scott:
I've taken a look at this, and it's being caused by the ability.getPrimarySpell() returning the first spell item in the CombatPotion ability. We could add a fake potion as the first item in this list as suggested above, or separate the potions out, which while working would cause two rows on the timeline if two different potions were used. 

Is there a preference for this?

PR #2787, opened 2018-11-28
Updated timeline to show 'Combat Potion' instead of Int Pot

Alastair-Scott:
Fix for #2678

Tried to use Spell Id of 0 for the dummy potion, but it made it appear at the top of the Timeframe.

### DIFF
--- a/src/common/SPELLS/bfa/potions.js
+++ b/src/common/SPELLS/bfa/potions.js
@@ -54,4 +54,9 @@ export default {
     name: 'Steelskin Potion',
     icon: 'inv_alchemy_80_elixir01green',
   },
+  DUMMY_COMBAT_POTION: {
+    id: Number.MAX_VALUE,
+    name: 'Combat Potion',
+    icon: 'trade_alchemy',
+  },
 };
--- a/src/common/SpellLink.js
+++ b/src/common/SpellLink.js
@@ -40,6 +40,15 @@ class SpellLink extends React.PureComponent {
       tooltipDetails.ilvl = ilvl;
     }
 
+    if(id === Number.MAX_VALUE) {
+      return (
+        <span className="pseudolink">
+          {icon && <SpellIcon id={id} noLink style={iconStyle} alt="" />}{' '}
+          {children || (SPELLS[id] ? SPELLS[id].name : `Unknown spell: ${id}`)}
+        </span>
+      );
+    }
+
     return (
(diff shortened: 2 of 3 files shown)


Alastair-Scott:
That was one of the ideas I had when first looking at this and I commented on the issue. Having the chat in discord earlier though, Zero agreed that as it was going with the new timeline, just doing this fix would be good for the next few months. I'm happy to rework all of this though if you think that's best.

Doing it with the split potions would cause two different lines on the Timeline if a healer for example pre potted with an int potion and then used a potion of replenishment later in the fight. I'm easy either way.

Dambroda:
I don't think it's a negative to show two different potions if two different potions were used, and splitting them up should work on the new timeline as well (where it wont be two lanes for different potions anyway) -- plus I think it ends up being actually less code to split it up then it is to program in this edge-case manually

Alastair-Scott:
Ok cool, I'll get to it!

Cheers for the info and feedback!

Dambroda:
Hmm good point, I hadn't thought about the shared cds and cast efficiency. I'd be slightly worried about changing the list at runtime because ability.primarySpell is used as an object key in few places, so if the ordering changes later you may run into issues (but I can't say for sure). 

My only other thought would be to change the primarySpellId on the first occurence of the ability being found, similar to re-sorting but instead making sure that all things that use it get the same primarySpellId.

For example, in ability and abilities.js => getAbility
```javascript
// Ability.js
  primaryOverride = null;
  get primarySpell() {
    if (this.spell instanceof Array) {
      return this.spell[this.primaryOverride || 0];
    } else {
      return this.spell;
    }
  }
```

```javascript
// Abilities.js
  getAbility(spellId) {
    const match = this.activeAbilities.find(ability => {
      if (ability.spell instanceof Array) {
        return ability.spell.some(spell => spell.id === spellId);
      } else {
        return ability.spell.id === spellId;
      }
    });
// NEW CODE 
    if(match && match.spell instanceof Array && match.primaryOverride === null) {
      match.primaryOverride = match.spell.findIndex((spell) => { return spell.id === spellId; });
      console.log("Set primary override of ", match, "to", match.primaryOverride);
    }
    return match;
  }
```

I can't see how it looks with multiple different pots used (like with mana and int propot), but so far on my end it's looking OK. Want to test it and see if it works like you'd expect?

Issue #2779, opened 2018-11-27
Add a normalizer to fabricate a `fightend` event


MartijnHols:
To replace the `finished` event fabrication, which allows us to remove special handling for it in the `fabricateEvent` method (`_isHandlingEvent`).

PR #2788, opened 2018-11-29
[Core] Change finished to fightend and move to a normalizer

Dambroda:
#2779


Dambroda:
_isHandlingEvent was added because finished was fabricated outside of a triggering event. Because that is no longer the case (meaning all fabricated events *are* triggered from another event, even if they dont set event.trigger), the logic is no longer needed to prevent the crash 

And yes I wasn't sure about it either (in terms of putting it into Events). I kinda took a middle ground, but I think that because it's a "core" event and not one from another module it kinda makes sense for it to be in events.

src/parser/core/CombatLogParser.test.js
```diff
@@ -240,13 +239,4 @@ describe('Core/CombatLogParser', () => {
       expect(Object.keys(parser._modules).length).toBeGreaterThan(0);
     });
   });
-  describe('emits static events', () => {
```
MartijnHols: Please maintain this test, it's important that the fightend event being fired is guarded

src/parser/shared/normalizers/FightEnd.js
```diff
@@ -0,0 +1,19 @@
+import EventsNormalizer from 'parser/core/EventsNormalizer';
+
+export const END_EVENT_TYPE = 'fightend';
+
+class FightEnd extends EventsNormalizer {
```
MartijnHols: Please add a comment above the class shortly explaining what it does and why, and optionally possible extensions.

src/parser/shared/normalizers/FightEnd.js
```diff
@@ -0,0 +1,19 @@
...
+
```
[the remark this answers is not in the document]
Dambroda: no

PR #2790, opened 2018-11-29
[Guardian Druid] Add Layered Mane azerite trait

FaideWW:
Implements stat tracking and statistics for Layered Mane.


sMteX:
Note that if #2791 gets merged before this, it's advised you use the same method of adding the buff to StatTracker (importing it as dependency and calling its `add()` method).

FaideWW:
Sure, that's fine. I have a few other traits in progress as well; I'll open PRs for them and keep the same thing in mind.

Pewtro:
#2791 got implemented so this just needs an update, but it looks fine otherwise - so once it's updated I'll come back and approve asap.

PR #2792, opened 2018-11-30
Show the channel of Potion of Replishment

### LINKED ISSUE #2742: Show the channel of Potion of Replishment on the timeline
https://www.wowhead.com/spell=252753/potion-of-replenishment

It has a 10 sec channel, we should show this on the timeline.

Alastair-Scott:
Fixes #2742

While this works, I'm not sure if this is the best place for the new Channelling class to be imported.


src/parser/shared/modules/Channeling.js
```diff
@@ -1,6 +1,7 @@
 import Analyzer from 'parser/core/Analyzer';
 import EventEmitter from 'parser/core/modules/EventEmitter';
 import CASTS_THAT_ARENT_CASTS from 'parser/core/CASTS_THAT_ARENT_CASTS';
+import SPELLS from 'common/SPELLS/index';
```
MartijnHols: This import should not include the index part, this is implicit

Alastair-Scott: Ok I'll remember that for next time, thanks!

PR #2791, opened 2018-11-29
[Core] Issue 2783 - StatTracker update

sMteX:
Fixes #2783 

Added a `add()` method to StatTracker, allowing to add the buffs to it in the opposite direction (from module to StatTracker), so we can delete the hideous wall of imports. I personally checked all affected traits, to see if `changestats` event was triggered when the buff was applied. Some traits also had wrong buff IDs specified so I fixed that too.

Example logs with affected traits:
- Shared
    - Archive of the Titans (0:02.145) - `/report/hW86TknYw4pHyLbd/13-LFR+Taloc+-+Kill+(3:47)/193-Balrouge`
    - Blightborne Infusion (0:05.448 - `/report/rVRPMLZQwXgdkj8W/76-LFR+Taloc+-+Kill+(3:58)/8-Danwo`
    - Blood Rite (1:45.125) - `/report/Y4JfmytrwhqLPgCT/18-LFR+Taloc+-+Kill+(3:45)/134-Arcyy`
    - Elemental Whirl - `/report/Y4JfmytrwhqLPgCT/18-LFR+Taloc+-+Kill+(3:45)/138-Dragonbow`
        - haste proc - 0:27.728
        - versatility proc - 0:29.481
        - mastery proc - 0:36.519
        - crit proc - 0:45.723
    - Gemhide (0:37.845) - `/report/hW86TknYw4pHyLbd/13-LFR+Taloc+-+Kill+(3:47)/193-Balrouge`
    - Meticulous Scheming (2:18.442) - `/report/Gw71yd9QMV8qFpJv/2-LFR+Taloc+-+Kill+(3:59)/37-Kuzbi`
    - Overwhelming Power (0:05.930) - `/report/hW86TknYw4pHyLbd/13-LFR+Taloc+-+Kill+(3:47)/193-Balrouge`
    - Swirling Sands (0:17.712) - `/report/Gw71yd9QMV8qFpJv/2-LFR+Taloc+-+Kill+(3:59)/48-Mageblast`
    - Tradewinds (1:33.891) - `/report/mptX96wzLyVB3ad7/2-LFR+Taloc+-+Kill+(3:30)/10-Noxington`
    - Unstable Catalyst (0:01.104) - `/report/4Gj1TCqR6DJmfnxc/1-LFR+Taloc+-+Kill+(4:30)/8-Kelirah`
- Death Knight
[...]

PR #2794, opened 2018-11-30
[Guardian Druid] Add Twisted Claws azerite trait

FaideWW:
Adds stat tracking and statistics for Twisted Claws.

As discussed, this uses the traditional method of adding tracked stats which is being deprecated in #2791. If that PR is merged first, I will update this one accordingly.


Pewtro:
#2791 got implemented so this just needs an update, but it looks fine otherwise - so once it's updated I'll come back and approve asap.

src/parser/druid/guardian/modules/azeritetraits/TwistedClaws.js
```diff
@@ -0,0 +1,53 @@
...
+  agility: twistedClawsStats,
+};
+
+class TwistedClaws extends Analyzer {
+  agility = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.TWISTED_CLAWS.id);
+    this.agility = twistedClawsStats(this.selectedCombatant);
```
Pewtro: you could return here if you wanted with  a 
```if(!active) {
return
}
```

and then skip the returns in the twistedClaws

PR #2796, opened 2018-12-01
Zip test logs

emallson:
See comments #2708 for details.

This can't be done (without a hell of a lot of work) with the builtin `zlib` module, so we end up with two added deps in dev mode (one for creation of the archive, one for decompression).

This also incidentally makes the tests asynchronous via promises, so thats a small upside.


emallson:
This has been fixed @MartijnHols, which may be important as currently tests for every other PR are going to break until this is merged. If you want the fixes split out into a separate PR from the zip stuff let me know.

also fwiw i did minimal vetting of the modules used for compression/decompression because they should never make it into the final bundle.

src/parser/core/tests/snapshotTest.js
```diff
@@ -55,9 +55,10 @@ export function tab(analyzer, parser=null) {
  */
 export default function snapshotTest(parserClass, moduleClass, key, propFn = statistic) {
```
MartijnHols: Where is the function used? I can't find it. Unused functions should be removed.

src/parser/core/tests/snapshotTest.js
```diff
@@ -55,9 +55,10 @@ export function tab(analyzer, parser=null) {
...
   return () => {
-    const log = loadLogSync(key);
-    const parser = parseLog(parserClass, log);
-    expectSnapshot(parser, moduleClass, propFn);
+    return loadLog(key).then(log => {
+      const parser = parseLog(parserClass, log);
```
MartijnHols: Does this parse the log for every single module now?

emallson:
trying to give travis a kick

src/parser/core/tests/snapshotTest.js
emallson: no, `integrationTest` does not use `snapshotTest` internally (it uses `expectSnapshot`).

However, I'm not sure if `snapshotTest` as-is should live on. If people start using it a lot then it will slow down test times by constantly re-parsing logs. Might be better to just provide a helper to do `beforeAll` parsing and then tell people to use `expectSnapshot`.

src/parser/core/tests/CombatLogParserIntegrationTest.js
```diff
@@ -62,26 +62,31 @@ function checklist(parser) {
 export default function integrationTest(parserClass, key, suppressWarn=true, suppressLog=true) {
```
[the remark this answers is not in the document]
MartijnHols: key

src/parser/core/tests/CombatLogParserIntegrationTest.js
```diff
@@ -0,0 +1,117 @@
...
+  return () => {
+    let log;
+    beforeAll(() => {
+      return loadLog(key).then(res => { log = res; });
+    });
+
+    suppressLogging(suppressLog, suppressWarn, false);
+
+    it('should parse the example report without crashing', () => {
+      const parser = parseLog(parserClass, log);
```
MartijnHols: Why isn’t this in the beforeAll?

src/parser/core/tests/log-tools.js
```diff
@@ -1,14 +1,23 @@
...
+import decompress from 'decompress';
 
 const _CACHE = {};
 
-export function loadLogSync(key, searchPath='test-logs/') {
-  if(_CACHE[key] === undefined) {
-    _CACHE[key] = JSON.parse(zlib.unzipSync(fs.readFileSync(`${searchPath}${key}.json.gz`)).toString());
+// asynchronously load and parse a log. returns a promise that resolves
+// to the log object
+export function loadLog(key, searchPath='test-logs/') {
```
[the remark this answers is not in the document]
MartijnHols: key

src/parser/core/tests/log-tools.js
```diff
@@ -0,0 +1,65 @@
...
+export function parseLog(parserClass, log) {
+  const friendlies = log.report.friendlies.find(({id}) => id === log.meta.player.id);
+  const fight = log.report.fights.find(({id}) => id === log.meta.fight.id);
+  const parser = new parserClass(
+    log.report, 
+    friendlies,
+    fight,
+    log.combatants
+  );
+  parser.normalize(JSON.parse(JSON.stringify(log.events))).forEach(event => parser.getModule(EventEmitter).triggerEvent(event));
```
MartijnHols: This json encoding/parsing seems unnecessary.

emallson: That's the simplest way to deep copy without libs. Normalizers sometimes mutate the event list

src/parser/core/tests/CombatLogParserIntegrationTest.js
emallson: honestly, this should just go away.

initially it was because we want the error to be associated with the test and not the `beforeAll`, but since tests happen asynchronously this may not actually run before other tests that use the `beforeAll` parser.

PR #2801, opened 2018-12-03
[Fury Warrior] Sudden Death talent implementation

Alastair-Scott:
Added in the Sudden Death talent for Fury

Log for testing:
[http://localhost:3000/report/ThbqR7cKj8mZ46pt/3-Mythic+Zek'voz+-+Kill+(6:15)/1-Lobos](http://localhost:3000/report/ThbqR7cKj8mZ46pt/3-Mythic+Zek'voz+-+Kill+(6:15)/1-Lobos)


src/parser/warrior/fury/modules/talents/SuddenDeath.js
```diff
@@ -0,0 +1,71 @@
...
+  onExecuteCast(event) {
+    if (this.selectedCombatant.hasBuff(SPELLS.SUDDEN_DEATH_TALENT_FURY_BUFF.id)) {
+      this.lastSuddenDeathExecuteCast = event.timestamp;
+      this.suddenDeathProcsUsed += 1;
+    }
+    this.lastExecuteCast = event.timestamp;
+  }
+
+  onExecuteDamage(event) {
+    if (this.lastExecuteCast === this.lastSuddenDeathExecuteCast) {
```
Dambroda: You actually only need to check on the damage event to see if the player had the buff a few milliseconds ago, rather than comparing it to the cast event. Technically dont need to listen to the cast event at all :)

src/parser/warrior/fury/modules/talents/SuddenDeath.js
```diff
@@ -0,0 +1,71 @@
...
+      this.suddenDeathDamage += event.amount + (event.absorbed || 0);
+    }
+  }
+
+  onSuddenDeathProc() {
+    this.suddenDeathProcs += 1;
```
Dambroda: Is SpellUsable updated anywhere? If this talent resets the cooldown, it should update SpellUsable to avoid cooldown errors

src/parser/warrior/fury/modules/talents/SuddenDeath.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  get damagePercent() {
+    return this.owner.getPercentageOfTotalDamageDone(this.suddenDeathDamage);
```
Dambroda: Just in general, this metric is not very accurate. Calculating the DPS gain from this trait is quite complicated, as you need to not only count the extra procs on targets above the normal health threshold, you also need to take into account any cooldown reduction it gives you on those below the threshold. Instead of the complicated DPS, maybe two different metrics:
`Extra 'high HP' executes (and you could give the damage values for those) on targets _above_ the standard health threshold`
+
`Cooldown reduction (for targets that were below the health threshold anyway but when you reduced the CD; again requires looking at SpellUsable)`

(In reality, this metric currently overvalues the trait, because it counts all executes, even those on targets below 20% hp, as fully related to the trait even if all the trait did was reduce the CD by a second)

src/parser/warrior/fury/modules/talents/SuddenDeath.js
Alastair-Scott: There's already some code in the Fury SpellUsable file that looks to handle this. Is there anything I also need to add to the analyzer?

src/parser/warrior/fury/modules/talents/SuddenDeath.js
Alastair-Scott: So I found out about the buffer duration on the hasBuff function, and that could be a good way to go with this, but I need to use the cast listener to check how many of the procs were used as the damage can be cleaved onto 4 targets, and we only want to count the one proc being used. 

Also, to check the number of uses against high hp targets, we need to do that on cast, as you could just use execute on a low hp target next to a high hp one and have that damage cleave.

PR #2813, opened 2018-12-08
Fixes #2365: Move battle net API calls to use the new blizzard.com API's

### LINKED ISSUE #2365: Migrate Blizzard API usage to new developer portal
https://develop.battle.net/documentation/guides/migration-guide

hartra344:
Production environment needs 2 new environment variables to hold the ClientId and Client Secret. Also moved client side calls to the server as a proxy with caching on the SQL server to reduce calls


Dambroda:
Some stylistic differences, but I didn't see any issues in my initial pass. But considering the scope of this change I think Zero should have a look. Thanks for this! 👍

server/helpers/wowCommunityApi.js
```diff
@@ -38,33 +39,145 @@ const get = (url, metricLabels) => {
...
+    onSuccess: () => {
+      end({
+        statusCode: 200,
+      });
+    },
+  });
+
+  const tokenData = JSON.parse(tokenRequst);
+  let expireDate = new Date();
+  expireDate = expireDate.setSeconds(expireDate.getSeconds() + tokenData.expires_in);
```
MartijnHols: Could be a one line:

```suggestion
  const expireDate = (new Date()).setSeconds(expireDate.getSeconds() + tokenData.expires_in);
```

server/helpers/wowCommunityApi.js
```diff
@@ -38,33 +39,145 @@ const get = (url, metricLabels) => {
     },
     onFailure: err => {
```
MartijnHols: Please add a check for the error that is thrown when the access token is expired, and automatically refresh the access token then.

scripts/localization/extract-spells.js
```diff
@@ -70,7 +70,7 @@ async function fetchSpellInfo(spellId, language) {
     try {
       // eslint-disable-next-line no-await-in-loop
       spellInfo = await request.get({
-        url: `https://${language.region}.api.battle.net/wow/spell/${spellId}?locale=${language.locale}&apikey=${process.env.REACT_APP_BATTLE_NET_API_KEY}`,
+        url: `https://${language.region}.api.blizzard.com/wow/spell/${spellId}?locale=${language.locale}&access_token=${accessToken}`,
```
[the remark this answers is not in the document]
hartra344: the access token is passed in through the parameters which is fetched before this is called

scripts/localization/extract-spells.js
```diff
@@ -182,16 +182,36 @@ function updateFromSpellList(languageCode, nameProp) {
...
 async function main() {
   // We could decide to split this up at some point; https://us.battle.net/forums/en/bnet/topic/20765127617#2
-  console.log(`Using Battle.net API key: ${process.env.REACT_APP_BATTLE_NET_API_KEY}`);
+  console.log(`Using Battle.net Client ID: ${process.env.REACT_APP_BATTLE_NET_API_CLIENT_ID} and Secret: ${process.env.REACT_APP_BATTLE_NET_API_CLIENT_SECRET}`);
   const languageCodes = Object.keys(languages);
   for (let i = 0; i < languageCodes.length; i += 1) {
+    
     const languageCode = languageCodes[i];
     const language = languages[languageCode];
+    const accessToken = await getAccessToken;
```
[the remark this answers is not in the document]
hartra344: oh nice haha, I just tried to catch everywhere I saw the API being called

server/controllers/api/item.js
```diff
@@ -0,0 +1,69 @@
...
+    console.log('Battle.net response time:', responseTime, 'ms');
+    const json = JSON.parse(response);
+    sendJson(res, json);
+    return json;
+  } catch (error) {
+    const { statusCode, message, response } = error;
+    console.log('Error fetching item', statusCode, message);
+    const body = response ? response.body : null;
+    // Ignore 404 - Item not found errors. We check for the text so this doesn't silently break when the API endpoint changes.
+    const isCharacterNotFoundError = statusCode === 404 && body && body.includes('Unable to get item information.');
```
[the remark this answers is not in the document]
hartra344: will correct var name but that is the error it gives on an item 404, same with spells

server/migrations/20180602142104-spell.js
```diff
@@ -0,0 +1,35 @@
...
+          defaultValue: DataTypes.NOW,
+          allowNull: false,
+        },
+        lastSeenAt: {
+          type: DataTypes.DATE, // this is actually DATETIME
+          defaultValue: DataTypes.NOW,
+          allowNull: false,
+        },
+      }),
+      queryInterface.addIndex('Spell', ['id']),
```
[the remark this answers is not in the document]
hartra344: it would but I'm new to sequelize and wasn't sure if it was needed, seems like I can remove it

server/migrations/20180602142104-item.js
```diff
@@ -0,0 +1,35 @@
...
+          defaultValue: DataTypes.NOW,
+          allowNull: false,
+        },
+        lastSeenAt: {
+          type: DataTypes.DATE, // this is actually DATETIME
+          defaultValue: DataTypes.NOW,
+          allowNull: false,
+        },
+      }),
+      queryInterface.addIndex('Item', ['id']),
```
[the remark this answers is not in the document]
hartra344: it would but I'm new to sequelize and wasn't sure if it was needed, seems like I can remove it

src/interface/character/Parses.js
```diff
@@ -200,7 +200,7 @@ class Parses extends React.Component {
 
     Object.values(updatedTrinkets).map(trinket => {
```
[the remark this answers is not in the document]
hartra344: that shouldn't have been removed

server/helpers/wowCommunityApi.js
hartra344: Had to make the expireDate in the second line a date.now() instead but yeah that works 👍

server/helpers/wowCommunityApi.js
hartra344: I'm trying to catch expired tokens by checking expired time when getting it, granted that could still allow some to slip through. Just made the change to renew it again on a 401 failure

hartra344:
@MartijnHols I think I went through and was able to address all of your comments :)

scripts/localization/extract-spells.js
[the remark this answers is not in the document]
MartijnHols: I'm not seeing it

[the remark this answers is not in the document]
Dambroda: Looks like the method signature doesn't describe it but it's being passed on L137

server/helpers/wowCommunityApi.js
```diff
@@ -36,35 +37,163 @@ const get = (url, metricLabels) => {
     onBeforeAttempt: () => {
       end = blizzardApiResponseLatencyHistogram.startTimer(metricLabels);
     },
+    onFailure: (err, attempt) => {
+      if (err.statusCode === 401 && attempt === 1) {
+        // if the request in unauthorized, try renewing
+        // the battle net creds, only worth trying once though
+        return getAccessToken(region);
```
MartijnHols: This doesn't try renewing if it already has an access token, I think you need to reset it first.

I don't think there's any need to only do it for the first attempt, in fact if the first attempt fails due to a connection error or something like that, subsequent requests might run into this.

server/helpers/wowCommunityApi.js
```diff
@@ -36,35 +37,163 @@ const get = (url, metricLabels) => {
...
+        statusCode: 200,
+      });
+    },
+  });
+};
+const makeBaseUrl = region => `https://${region}.api.blizzard.com`;
+
+const getAccessToken = async (region, corr) => {
+  let end;
+  if (clientToken[region] && !clientToken[region].accessToken && clientToken[region].expires > new Date()) {
```
MartijnHols:
```suggestion
  if (clientToken[region] && clientToken[region].accessToken && clientToken[region].expires > new Date()) {
```

server/helpers/wowCommunityApi.js
hartra344: you're right and I noticed on some testing, I'm allowing it to pass back changes options, you can tell me if you see anything that can go wrong there?

src/interface/character/Parses.js
```diff
@@ -225,11 +228,14 @@ class Parses extends React.Component {
 
     // Skip CN-API due to blizzard restrictions (aka there is no API for CN)
     if (region === 'CN') {
-      this.setState({
-        image: FALLBACK_PICTURE,
-      }, () => {
-        this.load();
-      });
+      this.setState(
```
MartijnHols: Please revert this change, the old style is preferred.

src/interface/character/Parses.js
```diff
@@ -270,12 +276,15 @@ class Parses extends React.Component {
     const imageUrl = `https://render-${this.props.region}.worldofwarcraft.com/character/${image}-main.jpg`;
     const role = data.talents.find(e => e.selected).spec.role;
     const metric = role === 'HEALING' ? 'hps' : 'dps';
-    this.setState({
-      image: imageUrl,
-      metric: metric,
-    }, () => {
-      this.load();
-    });
+    this.setState(
```
MartijnHols: same

src/interface/character/Parses.js
```diff
@@ -406,38 +416,75 @@ class Parses extends React.Component {
     if (this.state.error === ERRORS.CHARACTER_NOT_FOUND) {
       errorMessage = (
         <div style={{ padding: 20 }}>
-          Please check your input and make sure that you've selected the correct region and realm.<br />
-          If your input was correct, then make sure that someone in your raid logged the fight for you or check <a href="https://www.warcraftlogs.com/help/start/" target="_blank" rel="noopener noreferrer">Warcraft Logs guide</a> to get started with logging on your own.<br /><br />
-          When you know for sure that you have logs on Warcraft Logs and you still get this error, please message us on <a href="https://discord.gg/AxphPxU" target="_blank" rel="noopener noreferrer">Discord</a> or create an issue on <a href="https://github.com/WoWAnalyzer/WoWAnalyzer" target="_blank" rel="noopener noreferrer">Github</a>.
+          Please check your input and make sure that you've selected the correct region and realm.
```
MartijnHols: Please revert this change, the old style is preferred.

src/interface/character/Parses.js
```diff
@@ -6,7 +6,7 @@ import { withRouter } from 'react-router-dom';
 import { compose } from 'redux';
```
MartijnHols: This file has a lot of stylistic changes that deviate from the style in the project. Please revert these style changes.

server/helpers/wowCommunityApi.js
```diff
@@ -26,45 +27,178 @@ const get = (url, metricLabels) => {
...
       return shouldRetry;
     },
     onBeforeAttempt: () => {
       end = blizzardApiResponseLatencyHistogram.startTimer(metricLabels);
     },
+    onFailure: async err => {
+      if (err.statusCode === 401) {
+        // if the request in unauthorized, try renewing
+        // the battle net creds
+        const accessToken = await getAccessToken(region);
```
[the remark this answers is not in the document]
hartra344: ah I see what you mean

server/helpers/wowCommunityApi.js
```diff
@@ -26,45 +27,178 @@ const get = (url, metricLabels) => {
...
     onSuccess: () => {
-      end({ statusCode: 200 });
+      end({
+        statusCode: 200,
+      });
     },
   });
+
+  const tokenData = JSON.parse(tokenRequst);
+  const expireDate = new Date().setSeconds(new Date().getSeconds() + tokenData.expires_in);
```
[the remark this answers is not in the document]
hartra344: I think it might be better to actually keep the new Date in its own variable, doing the new in the middle of the statement is a bit odd feeling.

scripts/localization/extract-spells.js
```diff
@@ -172,7 +170,7 @@ function updateFromSpellList(languageCode, nameProp) {
       const spellInfo = spellList.find(item => item.id === spellIdString);
 
       // Some spells aren't localized, use the original name instead.
-      const name = (spellInfo && spellInfo[nameProp]) ? spellInfo[nameProp] : spell.name;
+      const name = spellInfo && spellInfo[nameProp] ? spellInfo[nameProp] : spell.name;
```
MartijnHols: Braces around ternary conditions with more than one condition is preferred because of weird JS operator prioritization.

```suggestion
      const name = (spellInfo && spellInfo[nameProp]) ? spellInfo[nameProp] : spell.name;
```

server/helpers/wowCommunityApi.js
```diff
@@ -26,45 +27,166 @@ const get = (url, metricLabels) => {
...
-      // Previously `shouldRetry` checked for just `err instanceof RequestError || statusCode === 503` - but the Blizzard API is so buggy with random errors, we're better off just retrying for anything unexpected. 503 errors happen regularly which probably means the character wasn't cached, try again once. The API also sometimes randomly throws 404 errors (without "Character not found." in the body) for no reason, we should retry those too.
-      const shouldRetry = !isCharacterNotFoundError;
-      return shouldRetry;
+      const { statusCode } = error;
+      //generally 404 should probably just not be retried
+      return statusCode !== 404;
+    },
+    getUrlWithAccessToken: async () => {
+      const accessToken = await getAccessToken(region);
+      console.log('access token:'+ accessToken);
```
MartijnHols:
```suggestion
```

hartra344:
@MartijnHols applied your minor suggestions

PR #2815, opened 2018-12-08
[Hunter] Careful aim for 8.1

### LINKED ISSUE #2534: [Marksmanship] Careful Aim module for 8.1
Careful Aim is being reworked so a module will be easier to make for this talent.

Pewtro:
image

https://www.warcraftlogs.com/reports/TrfWp1jHdRtQDqkx#fight=2&type=damage-done&source=3

This module shouldn't get implemented before 8.1.

Fixes #2534


src/interface/others/TalentStatisticBox.js
```diff
@@ -21,6 +21,7 @@ const TalentStatisticBox = ({ talent, icon, label, ...others }) => {
 TalentStatisticBox.propTypes = {
   ...StatisticBox.propTypes,
   talent: PropTypes.number.isRequired,
+  value: PropTypes.node.isRequired, //Ensure this is required
```
Dambroda: This is in StatisticBox.propTypes which is expanded here; does the override not work for lint?

Pewtro: Nope, only talent is passed through for me

Dambroda: As discussed in disc, works as intended for react but IDE's don't interpret the expanded proptypes properly for inspections & autocomplete.... 

Can either leave as-is and rely on standard react proptype behavior (console messages) sacrificing IDE autocomplete and inspector warnings _or_ redo the proptypes completely for talent/trait statistics box to not use any expansion, so that IDE's interpret it all properly. But probably shouldn't do a mix of the two (just adding the one prop)

PR #2818, opened 2018-12-09
[Tank] Split mitigation checks into physical and magical damage

hpabst:
This pull request changes the existing mitigation check module to differentiate between physical and magical damage type checks.


src/parser/paladin/protection/modules/features/ShieldOfTheRighteous.js
```diff
@@ -49,7 +49,7 @@ class ShieldOfTheRighteous extends Analyzer {
   constructor(...args) {
     super(...args);
     const boss = findByBossId(this.owner.boss.id);
-    this._tankbusters = boss.fight.softMitigationChecks;
+    this._tankbusters = [...boss.fight.softMitigationChecks.physical, ...boss.fight.softMitigationChecks.magical];
```
emallson: this should only have physical. SotR does nothing against magic damage as of 8.0

src/parser/shared/modules/MitigationCheck.js
```diff
@@ -19,47 +20,77 @@ class MitigationCheck extends Analyzer {
...
+      debuffCheck = [...this.debuffCheckPhysical, ...this.debuffCheckPhysAndMag];
+    } else {
+      checks = this.checksMagical;
+      buffCheck = [...this.buffCheckMagical, ...this.buffCheckPhysAndMag];
+      debuffCheck = [...this.debuffCheckMagical, ...this.debuffCheckPhysAndMag];
+    }
+    if (checks.includes(spell) && !event.tick) {
+      debug && console.log(buffCheck);
+      debug && console.log(debuffCheck);
+      if (buffCheck.some((e) => this.selectedCombatant.hasBuff(e)) || event.hitType === HIT_TYPES.IMMUNE) {
```
emallson: this function got a bit messy in the transition. idk if there is a great way to simplify it, though

src/parser/paladin/protection/modules/features/ShieldOfTheRighteous.js
hpabst: Fixed.

src/parser/shared/modules/MitigationCheck.js
hpabst: My own feeling is that it's fine for the time being. It's a little bit ugly, but like you said there's no obvious way to simplify things.

PR #2821, opened 2018-12-10
[Core] Update Timeline to allow specifying which spell you want shown

Pewtro:
image
image

This allows you to set which spell you want linked in the timeline on the left, if nothing is defined as shownSpell, it'll just default to what it used to. 

Also deleted the redundant 
```JS
                {abilities.getAbility(spellId).name}
```

but if it was used somewhere let me know, I just couldn't find anywhere that would do anything different than regular spelllink. 

With the prepot update (I think), my prior hack to this broke - this fixes it permanently.

### DIFF
--- a/src/parser/hunter/survival/modules/Abilities.js
+++ b/src/parser/hunter/survival/modules/Abilities.js
@@ -216,9 +216,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.HARPOON,
         category: Abilities.SPELL_CATEGORIES.UTILITY,
-        gcd: {
-          base: 500,
-        },
+        gcd: null,
         cooldown: 20,
       },
       {
@@ -293,6 +291,12 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
+      },
+      {
+        spell: SPELLS.MISDIRECTION,
+        category: Abilities.SPELL_CATEGORIES.UTILITY,
+        cooldown: 30,
+        gcd: null,
       },
 
       /**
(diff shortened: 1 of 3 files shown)


Dambroda:
This should probably be somehow worked into the primaryOverride variable that was created for the combat potion changes rather than having two separate variables in the Ability to specify what the shown spell is; maybe convert primaryOverride to a prop or rename it to the shownSpell variable and change the logic in Abilities that sets the override?

Pewtro:
I initially tried with primaryOverride, but a good way is eluding me.

Pewtro:
I think it's fine as is, primaryOverride is just doing the work behind the scenes if you don't set shownSpell - however anyone can now simply do shownSpell if they desire to have a specific spell shown. 

I also can't figure out a good way to replace one or the other reasonably.

src/parser/hunter/survival/modules/Abilities.js
```diff
@@ -113,6 +113,7 @@ class Abilities extends CoreAbilities {
       },
       { //WFI talent here first so that's the icon shown in the timeline - it has no other effect.
         spell: [SPELLS.WILDFIRE_INFUSION_TALENT, SPELLS.VOLATILE_BOMB_WFI, SPELLS.PHEROMONE_BOMB_WFI, SPELLS.SHRAPNEL_BOMB_WFI],
+        shownSpell: SPELLS.WILDFIRE_INFUSION_TALENT,
```
MartijnHols: The first spell in the spell array is shown in the timeline, so then this doesn't do anything?

PR #2832, opened 2018-12-15
[Fury Warrior] Update for 8.1

Alastair-Scott:
Implemented the two new Azerite traits and removed the old two, along with removing the gcd of charge.

Logs to test the new traits:
Unbridled Ferocity: [http://localhost:3000/report/VQwFXKkZHMbGNtf6/2-Mythic+Taloc+-+Kill+(4:10)/36-Lunarion](http://localhost:3000/report/VQwFXKkZHMbGNtf6/2-Mythic+Taloc+-+Kill+(4:10)/36-Lunarion)
Cold Steel, Hot Blood: [http://localhost:3000/report/hzpFMQtHamwXf4xG/1-Mythic+Taloc+-+Kill+(4:54)/10-Sebkay](http://localhost:3000/report/hzpFMQtHamwXf4xG/1-Mythic+Taloc+-+Kill+(4:54)/10-Sebkay)


src/parser/warrior/fury/modules/azerite/UnbridledFerocity.js
```diff
@@ -0,0 +1,93 @@
...
+
+  onBuffApply(event) {
+    this.recklessnessEvents.push({
+      start: event.timestamp,
+      end: null,
+    });
+  }
+
+  onBuffRemoved(event) {
+    const latestRecklessness = this.recklessnessEvents.pop();
```
sMteX: This (and the same thing in `onFightEnd`) popping and pushing back is unnecessary. If the point is to assign `end` property into the last item in `recklessnessEvents`, then you can do just 
```javascript
if (this.recklessnessEvents.length === 0) {
    // shouldn't be necessary but just to be sure 
    return;
}
const lastIndex = this.recklessnessEvents.length - 1;
this.recklessnessEvents[lastIndex].end = event.timestamp;
```

src/parser/warrior/fury/modules/azerite/UnbridledFerocity.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  get damagePercentage() {
+    return this.owner.getPercentageOfTotalDamageDone(this.totalDamage);
+  }
+
+  get increasedRecklessnessDuration() {
+    return this.recklessnessEvents.reduce((total, event) => {
+      const duration = event.end - event.start;
+      return total + (duration > RECKLESSNESS_DURATION ? duration - RECKLESSNESS_DURATION : duration);
```
sMteX: I don't quite understand the logic here - if you had over 10s of the buff duration, it only counts the bonus duration, but otherwise the whole duration? Also, it doesn't seem you distinguish the normal Recklessness (https://www.wowhead.com/spell=1719/recklessness) from the one gained via this trait? 

Some clarification please 😄

sMteX: Wait, probably nevermind - the Recklessness buff has 10 sec duration, so if it's more than that, the trait procced sometimes during that.. And if it's less than 10 sec, then it's the trait again and you count the duration as a whole, correct? 

In which case, a comment explaining this would be nice. Also, you might want to add some kind of buffer, I doubt the actual 10 second buff would last exactly 10000ms, but in reality probably few ms more or less (which means a "trait" proc that should last for 4 seconds could last for let's say 9.85 seconds)

PR #2838, opened 2018-12-18
Add character data caching to character API

### LINKED ISSUE #2471: Cache character data
The Blizzard API is super slow and unstable, so we should cache character info in the character table. 

Consider making separate columns for extra fields like talents.

hartra344:
fixes #2471
Notes:

Current implementation I haven't yet taken the time to update comments and such until it's finalized. Cached data covers all base character data and the minimum useful date from the talents API. This does change the response shape a bit to simplify the data sent back. Once I have confirmation that this is a good approach I'll go through and update all of the comments and documentation.

One more note: Currently I'm caching for an hour, if over an hour how I fetch from the API and fallback to the cached data if that fails. This is the more seamless option, the other option if preferable might be to add a refresh button similar to WCL caching and cache indefinitely until that gets hit. Wanting to hear thoughts on that.


MartijnHols:
A likely issue with using live data is that it may lead to slow responses. How about responding with the cache right away but refreshing async on the server (if older than an hour)? That way it's a new thing next time they reload.

hartra344:
I thought too about the cache first approach, the only concern I had was if someone comes a week later, they're going to get a week old cache unless they refresh.  But now that I think more about it with the data we currently use, that's probably ok since it doesn't change often(stuff like race), if in the future though more data(maybe more talent data for a profile or something) is used it might make sense to think about it.

MartijnHols:
A big problem is that the Blizz API is often very slow, and I think API response is very important, so I'd prioritize that. And as you said the data doesn't change regularly. And if a user notices something that is outdated, refreshing the app is a natural way to get new data. 

You could also make it force refresh if it's older than a month or so, but even that is probably both unnecessary and slow.

server/controllers/api/character.js
```diff
@@ -54,55 +68,107 @@ async function proxyCharacterApi(res, region, realm, name, fields) {
...
-  // Everything after this happens after the data was sent
-  if (!characterInfo) {
+
+  const storedCharacter = await getStoredCharacter(null, realm, region, name);
+  if (storedCharacter) {
+    sendJson(res, storedCharacter);
+  }
+
+  const characterFromApi = await getCharacterFromBlizzardApi(region, realm, name);
+  // If Character Data can't be retrived from API then we fallback to cached data in the db
```
MartijnHols: wrong comment?

hartra344: yep, somehow missed removing that one

PR #2841, opened 2018-12-18
Make token call use already defined get call to clean up wowCommunity…

hartra344:
…Api helper

### DIFF
--- a/server/helpers/wowCommunityApi.js
+++ b/server/helpers/wowCommunityApi.js
@@ -15,7 +15,7 @@ const availableRegions = {
 const USER_AGENT = process.env.USER_AGENT;
 const clientToken = {};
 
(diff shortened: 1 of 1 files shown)


server/helpers/wowCommunityApi.js
```diff
@@ -32,8 +32,13 @@ const get = (url, metricLabels, region) => {
       return statusCode !== 404;
     },
     getUrlWithAccessToken: async () => {
-      const accessToken = await getAccessToken(region);
-      return `${url}&access_token=${accessToken}`;
+      console.log(`skippingAccessToken: ${skipAccessToken}`);
```
MartijnHols: Please clean up console.logs in production code

```suggestion
```

PR #2849, opened 2018-12-20
[Bug] default to 'us' region for our spell/item endpoint

joshinat0r:
Adds some checks to the spell/item endpoints to default to the US-region and to check if a spellId was given.
The character-parses page right now is doing calls like `https://wowanalyzer.com/api/item/undefined/null?`
Undefined because the region is not set for that call and null because of logs without advanced combatlogging.


src/common/makeApiUrl.js
```diff
@@ -19,12 +19,18 @@ export function makeCharacterApiUrl(characterId = null, region = null, realm = n
   });
 }
 
-export function makeItemApiUrl(itemId, region) {
+export function makeItemApiUrl(itemId, region = 'us') {
+  if (!parseInt(itemId)){
```
MartijnHols:
```suggestion
   if (!Number.isInteger(itemId)){
```

src/common/makeApiUrl.js
```diff
@@ -19,12 +19,18 @@ export function makeCharacterApiUrl(characterId = null, region = null, realm = n
...
+    return false;
+  }
   const parts = ['item', region, itemId];
   return makeApiUrl(parts.map(part => encodeURIComponent(part)).join('/'));
 }
 
-export function makeSpellApiUrl(spellId, region) {
+export function makeSpellApiUrl(spellId, region = 'us') {
+  if (!parseInt(spellId)){
```
MartijnHols:
```suggestion
   if (!Number.isInteger(spellId)){
```

MartijnHols:
@joshinat0r thanks for this PR! I had reworked the blizzard API a bit and inspired by our PR, the region param has now been removed so that the default is no longer necessary. b5632ba44415af7fc110e2d13fef3edbb790d9cc

Issue #2086, opened 2018-08-11
StatTracker: Primary Stat raid buffs


niseko:
Arcane Intellect, Battle Shout and their scrolls give you a 10% (7%) multiplayer on the given mainstat. This is included in the stats on pull if you had that buff already, but your stats also need to be updated if you lose or gain them (deaths for example).
Temporary mainstat buff applications / removals also need to be increased by 10% (7%).

PR #2854, opened 2018-12-21
[Demonology] Shadow's Bite cleanup

sMteX:
Pre-8.1 we had Forbidden Knowledge trait, which increased damage of Demonbolt. Shadow's Bite increased that too, that's why I made a common core module that took care of attributing bonus damage to both of the traits. Now that FK is gone, that core no longer needs to be separate so I just merged the functionality (copied the code, renamed variables..) into the Shadow's Bite module.


src/parser/warlock/demonology/modules/azerite/ShadowsBite.js
```diff
@@ -1,31 +1,100 @@
...
+  }
+
+  onDemonboltCast(event) {
+    // Shadow's Bite snapshots the bonus on cast
+    this._queue.push({
+      timestamp: event.timestamp,
+      targetID: event.targetID,
+      targetInstance: event.targetInstance,
+      applySB: this.selectedCombatant.hasBuff(SPELLS.SHADOWS_BITE_BUFF.id),
+      intellect: this.statTracker.currentIntellectRating,
```
Anatta336: Rather than relying on `statTracker` to be correct you can access the player's spell power directly from `event.spellPower` for a cast event such as this.

sMteX: This might reveal that we have a more severe problem in StatTracker as a whole, that we might be double-dipping the on-pull buffs all the time. The log in the link shows the player's intellect as 9198, which is indeed passed as `_pullStats`:
`StatTracker: 00:00.000 - STR=557 AGI=1130 INT=9198 STM=9562 CRT=1868 HST=1055 MST=795 VRS=184 AVD=0 LCH=0 SPD=0 ARMOR=1142`

But then, the buff normalizer fabricates applybuffs for Intel flask, Intel food, which gets picked up again
```
StatTracker: (0 -> 1) Flask of Endless Fathoms @ 00:00.000 - Change: STR=0 AGI=0 INT=238 STM=0 CRT=0 HST=0 MST=0 VRS=0 AVD=0 LCH=0 SPD=0 ARMOR=1142 
StatTracker: 00:00.000 - STR=557 AGI=1130 INT=9436 STM=9562 CRT=1868 HST=1055 MST=795 VRS=184 AVD=0 LCH=0 SPD=0 ARMOR=1142 
StatTracker: (0 -> 1) Bountiful Captain's Feast @ 00:00.000 - Change: STR=0 AGI=0 INT=100 STM=0 CRT=0 HST=0 MST=0 VRS=0 AVD=0 LCH=0 SPD=0 ARMOR=1142 
StatTracker: 00:00.000 - STR=557 AGI=1130 INT=9536 STM=9562 CRT=1868 HST=1055 MST=795 VRS=184 AVD=0 LCH=0 SPD=0 ARMOR=1142
```
resulting in 338 more intellect than it should have (*maybe*).

At this point, I'm conflicted as what to believe - do the `_pullStats` contain all the buffs already? Is `event.spellPower` accurate?

Anatta336:
I believe I've found how to avoid the double counting issue for prepull buffs.
Specifically for the log we've been looking at (report/V2Zpv9NzdqBJgXky/11-Heroic+Fetid+Devourer+-+Kill+(1:15)/9-Zalaberto/events) it's an intellect feast and intellect flask.

As part of the `Combatants` constructor, `this.applyBuff` is called for everything found in `combatantInfo.auras`. But they don't get marked as being prepull auras so `StatTracker` processes them and applies the stat changes. So far as I'm aware `combatantInfo.auras` will only ever contain auras that were active before the pull, so we can safely mark them all as being prepull.

We can add that property to line 45 of `parser\shared\modules\Combatants.js`
```
combatantInfo.auras.forEach((aura) => {
  this.applyBuff({
    ability: {
      abilityIcon: aura.icon,
      guid: aura.ability,
    },
    sourceID: aura.source,
    targetID: combatantInfo.sourceID,
    timestamp: combatantInfo.timestamp,
    prepull: true,
  });
});
```

With this change the `event.spellpower` and `this.statTracker.currentIntellectRating` are much closer aligned. They diverge over the fight due to `StatTracker` not yet dealing with percentage effects (see issue #2086). In this case it's [Arcane Intellect](https://www.wowhead.com/spell=1459/arcane-intellect) and [Armor Skills](https://www.wowhead.com/spell=76277/armor-skills). When the battle potion wears off `StatTracker` reduces intellect by 900, but the combatant will have actually lost 900*1.10*1.05 = 1039.5 intellect.

In my opinion this pull request can fix the prepull double-counting issue and leave dealing with percentage effects to another PR. I'll have a go at it myself tomorrow now that I'm a bit more familiar with it.

sMteX:
Another solution could be to drop the line altogether, since ApplyBuff normalizer (`parser/shared/normalizers/ApplyBuff`) does the same thing anyway (loop through combatant events and for each aura found, emit an `applybuff` *with* `prepull` and `__fabricated` flags).

sMteX:
Log: http://localhost:3000/report/V2Zpv9NzdqBJgXky/11-Heroic+Fetid+Devourer+-+Kill+(1:15)/9-Zalaberto/events

### Now:

Notice the "changebuffstack" events with no "applybuff" on 0:00.00 (these are from Combatants constructor, calling `Entities.applyBuff()` directly) - these get picked up by StatTracker and re-applied even though they're already in the `_pullStats`, resulting in 338 additional Intellect the player shouldn't have.
image
image

### Change:

If we remove the loop in constructor, as well as the check in `Combatants.on_applybuff()`, we get this:

The ApplyBuff normalizer takes care of applying the buff the regular way, "changebuffstack" gets called every time, so it should recognize the buffs and add them to the "_buffs" array, yet StatTracker correctly doesn't add them, because ApplyBuff normalizer marks them with "prepull".
image
image

Anatta336:
Yours sounds like the better solution. Currently we have some prepull buffs events fabricated by `Combatants` (fabricated with incorrect properties) and others by `ApplyBuff`, far better to have them all coming from the normalizer that exists specifically to do that.

Looking through the code history, the loop that applies the buffs in `Combatants` originates in 9da26e8812ec213e112398f277ab46d03e52b6c9 (2 years ago) when tracking of enemy (de)buffs was first added. I'm fairly sure that application of enemy buffs is also handled by the `ApplyBuff` normalizer, so it should no longer be needed at all.

MartijnHols:
I'm not familiar with the behavior of that code, but it looks good?

PR #2858, opened 2018-12-21
[BrM] Gift of the Ox

emallson:
screenshot from 2018-12-21 11-44-46

Also includes code to estimate the value of agility / mastery to the heal, which is not used yet but is being used in an in-progress stat weights PR.


src/parser/monk/brewmaster/normalizers/ExpelHarm.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      // quick check that EH is after ALL orbs
+      if(events[idx+1].timestamp === event.timestamp && isGotOxHeal(events[idx+1])) {
+        console.error('gotox orb after eh');
+      }
+      // find the start of run of orbs consumed
+      // safe --- game won't let you cast EH with 0 orbs
+      let i = idx;
+      while(events[i-1].timestamp === event.timestamp) {
```
sMteX: I think some kind of buffer would be helpful here. In this log for example https://wowanalyzer.com/report/t2ZK6pQvgLGqj7by/1-Mythic+MOTHER+-+Kill+(4:19)/14-Mantyke/events, I don't think the heal events from EH happen exactly at the same timestamp, e.g. here I think at least the last 2 of the 3 heals were triggered from EH, yet they're not on the same timestamp
image

src/parser/monk/brewmaster/modules/spells/GiftOfTheOx.js
```diff
@@ -0,0 +1,107 @@
...
+    // Heal = 1.5 * (6 * WDPS + BonusAgi + BaseAgi) * Mastery * Vers
+    //
+    // With BaseAgi known, we get:
+    //
+    // BonusHeal = 1.5 * (6 * WDPS + BonusAgi + BaseAgi) * Mastery * Vers - 1.5 * (6 * WDPS + BaseAgi) * Mastery * Vers
+    //           = Heal * (1 - (6 WDPS + BaseAgi) / (6 WDPS + BonusAgi + BaseAgi))
+    //           = Heal * (BonusAgi / (6 WDPS + BonusAgi + BaseAgi))
+    //
+    // and similar for bonus WDPS healing and base agi healing
+    const denom = (6 * this._wdps + this.stats.currentAgilityRating);
```
sMteX: I don't really understand the math here, why and how, so I'm just gonna trust you on this 😅  If it's wrong, it's wrong just for BMs 🤷‍♀️

src/parser/monk/brewmaster/normalizers/ExpelHarm.js
emallson: good catch, i hadn't seen any so i didn't implement it but it is handled now :+1:

emallson:
@sMteX the bare `EventFilter` has been removed and replaced with an event imported from the gotox normalizer (which was also changed `tick -> orb-generated` so that I can just filter on that).

PR #2872, opened 2018-12-24
[Affliction] Deathbolt talent update

sMteX:
Deathbolt deals damage based on remaining duration of DoTs on the target. This PR adds the possibility to see average remaining DoT duration on Deathbolt casts (both total and per-dot basis). Since this uses the same code for getting default dot durations as in Darkglare, I refactored that out into constants

Example log: `/report/W7JnA1zd9qFc8ypD/1-Heroic+Taloc+-+Kill+(3:47)/13-Sindarin`

image


src/parser/warlock/affliction/modules/features/Darkglare.js
```diff
@@ -74,18 +63,11 @@ class Darkglare extends Analyzer {
   constructor(...args) {
     super(...args);
     // if player has Absolute Corruption, disregard the Corruption duration (it's permanent debuff then)
+    this._dotDurations = getDotDurations(this.selectedCombatant.hasTalent(SPELLS.CREEPING_DEATH_TALENT.id));
     this._hasAC = this.selectedCombatant.hasTalent(SPELLS.ABSOLUTE_CORRUPTION_TALENT.id);
```
Pewtro: Might want to move the comment down here since _dotDurations doesn't have anything to do with AC?

Issue #2824, opened 2018-12-11
Trinket: World Boss: Ancient Knot of Wisdom


niseko:
https://ptr.wowhead.com/item=166793/ancient-knot-of-wisdom&bonus=1502:5479

![image](https://user-images.githubusercontent.com/2842471/49835829-b64e9c80-fda0-11e8-9e98-5637e7c5338a.png)

5chmidt:
I can start working on this.  Here is the log a found to test with.
https://www.warcraftlogs.com/reports/t2ZK6pQvgLGqj7by#fight=1&source=13

PR #2873, opened 2018-12-24
Dev/ancient knot of wisdom

5chmidt:
Added Ancient Knot of Wisdom trinket stat tracker.  
Reports using the item will now show use count, percent up time and average intellect gained.

When reviewing this PR please double check:
`      this.statTracker.add(SPELLS.WISDOM_OF_THE_FOREST_LORD.id, {
        intellect: this.intellect * 0.6,
      });`

Intellect * 0.6 is the correct average intellect value, but I'm not sure if it should be added that way to the stat tracker, because it is actually ticking down.
Thanks,

tooltip
#2824


Pewtro:
The **actual current** amount is what needs adding to the statTracker, not the average. 
There are modules relying on accurate number representation in the statTracker so average amount won't cut it sadly

src/parser/shared/modules/items/bfa/AncientKnotOfWisdom.js
```diff
@@ -0,0 +1,84 @@
...
+  casts = 0;
+  intellect = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrinket(
+      ITEMS.ANCIENT_KNOT_OF_WISDOM.id
+    );
+    if (this.active) {
+      this.intellect = calculateSecondaryStatDefault(385, 1060, this.selectedCombatant.getItem(ITEMS.ANCIENT_KNOT_OF_WISDOM.id).itemLevel);
```
sMteX: After lengthy discussion on Discord we figured out the most correct solution is to use per-stack intellect values, and scaling from the base ilvl of 355 (as is on Wowhead), which is 800 intellect (40 intellect per stack).
```suggestion
      this.intellectPerStack = calculatePrimaryStat(355, 40, this.selectedCombatant.getItem(ITEMS.ANCIENT_KNOT_OF_WISDOM.id).itemLevel);
```
This will return intellect per stack for the given item level, which you can directly plug in the statTracker line and also use that for the average intellect.

Note: scaling this off 385 version (53 intellect per stack) isn't necessarily wrong, but because of rounding errors, this ends up being 1 intellect per stack off, resulting in 20 intellect difference, that's why I suggested base ilvl

Gebuz:
imo the average int value should be the primary value shown, and # of uses + uptime should be in the tooltip instead. That's also how most other trinkets values are shown iirc.

5chmidt:
Created get method for averageIntellect() and verified that it is returning the same value as was previously calculated.  Will now account for boss dying before all stacks have ticked off.

sMteX:
I think the current version of it is correct - personally I'm confused by `getStackWeightedBuffUptime` too, but looking at https://github.com/WoWAnalyzer/WoWAnalyzer/blob/223b489ccdf4417cdfbcf57152c0e477dbd4312c/src/parser/core/Entity.js#L109 it looks like the end of the fight is already accounted for. 

Also you can take a look at this commit when I was making the Bursting Flare module - https://github.com/WoWAnalyzer/WoWAnalyzer/pull/2760/commits/3db23e2fb400cd2ed0c9b15f98c59006a64b72f7

I was calculating the mastery manually (time on each stack * stacks * mastery per stack and averaging it in the end), and replacing it with weighted buff uptime produced the same results

sMteX:
Also, the issue with buffs that count down stacks instead of up is a problem, but since https://github.com/WoWAnalyzer/WoWAnalyzer/commit/066a7b6644d0d4b4af03e6f69fb9989afcb5ddcc the player's Intellect (and Intellect only) is update quite frequently on every player's cast and incoming heal.

So even if the trinket was triggered on 0:04.872 and had *technically* incorrect value for 4 seconds till 0:08.970 (instead of 1060 it was 53 I think), due to those changes, the StatTracker corrected the Intellect only 464ms after, because Fury of Elune was *casted* at 0:05.336.

It's a bandaid, sure, but until we figure out a solution for that type of buffs, this should be fine.

PR #2882, opened 2018-12-26
[Outlaw] Alacrity, BladeRush, QuickDraw and WeaponMaster talents added

HawkCorrigan:
https://www.warcraftlogs.com/reports/NytaDgKG8QqW4hBP#fight=3&type=damage-done&source=17
https://www.warcraftlogs.com/reports/rD1a87qGnYcMpy4F#fight=2&type=damage-done

### DIFF
--- a/src/common/SPELLS/rogue.js
+++ b/src/common/SPELLS/rogue.js
@@ -228,7 +228,7 @@ export default {
     id: 248210,
     name: 'The First of the Dead',
     icon: 'inv_glove_cloth_raidwarlockmythic_q_01',
-  }, 
+  },
   FIND_WEAKNESS_BUFF: {
     id: 91021,
     name: 'Find Weakness',
@@ -410,7 +410,7 @@ export default {
     name: 'Toxic Blade',
     icon: 'inv_weapon_shortblade_62',
   },
-  
+
 
   //Outlaw
 
@@ -476,8 +476,22 @@ export default {
     name: 'Blade Flurry',
     icon: 'ability_warrior_punishingblow',
   },
-
+  BLADE_RUSH_DAMAGE:{
+    id:271881,
+    name: 'Blade Rush Damage',
+    icon: 'ability_arakkoa_spinning_blade',
+  },
   //Buffs
+  ALACRITY_BUFF: {
+    id: 193538,
+    name: "Alacrity",
+    icon: "ability_paladin_speedoflight",
+  },
+  OPPORTUNITY: {
+    id: 195627,
+    name: "Opportunity",
+    icon: "ability_rogue_pistolshot",
+  },
   SCULL_AND_CROSSBONES: {
     id: 199603,
     name: 'Skull and Crossbones',
--- a/src/parser/rogue/outlaw/modules/talents/Alacrity.js
+++ b/src/parser/rogue/outlaw/modules/talents/Alacrity.js
@@ -1,0 +1,69 @@
+import React from 'react';
+import Analyzer from 'parser/core/Analyzer';
+import SPELLS from 'common/SPELLS';
+import SpellIcon from 'common/SpellIcon';
+import { formatDuration, formatPercentage } from 'common/format';
+import StatisticBox from 'interface/others/StatisticBox';
+import AlacrityTimesByStacks from 'parser/rogue/outlaw/modules/talents/AlacrityTimesByStacks';
+
+const HASTE_PER_STACK = 0.02;
+
+class Alacrity extends Analyzer {
+
+  static dependencies = {
+    alacrityTimesByStacks: AlacrityTimesByStacks,
+  };
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.ALACRITY_TALENT.id);
+  }
+
+
+  get alacrityTimesByStack() {
+    return this.alacrityTimesByStacks.alacrityTimesByStacks;
+  }
+
+
+  get uptime() {
+    return this.selectedCombatant.getBuffUptime(SPELLS.ALACRITY_BUFF.id) / this.owner.fightDuration;
+  }
+
+  get averageHaste() {
+    return this.alacrityTimesByStacks.averageAlacrityStacks*HASTE_PER_STACK;
+  }
+
+  statistic() {
+
+    return (
+      <StatisticBox
+        icon={<SpellIcon id={SPELLS.ALACRITY_TALENT.id} />}
+        value={`${formatPercentage(this.averageHaste)} %`}
+        label={`Alacrity Average Haste Gain (Uptime ${formatPercentage(this.uptime)}%) `}
+      >
+        <table className="table table-condensed">
+          <thead>
+            <tr>
+              <th>Stacks</th>
+              <th>Time (s)</th>
+              <th>Time (%)</th>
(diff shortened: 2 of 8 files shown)


src/parser/rogue/outlaw/modules/talents/QuickDraw.js
```diff
@@ -0,0 +1,46 @@
...
+
+  damageGained = 0;
+  comboPointsGained = 0;
+
+  _onPistolShot(event) {
+    if (!this.selectedCombatant.hasBuff(SPELLS.OPPORTUNITY.id)) {
+      return;
+    }
+    this.comboPointsGained +=
+      (event.resourceChange > 0 ? event.resourceChange - BASE_CP_GENERATED : 0) || 0;
```
TsaboTavok: This does not take in to account Broadside buff. 
Also this will count CPs that are overcapped as generated.

src/parser/rogue/outlaw/modules/talents/AlacrityTimesByStacks.js
```diff
@@ -0,0 +1,84 @@
...
+  lastAlaUpdate = this.owner.fight.start_time;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.ALACRITY_TALENT.id);
+
+    this.alacrityStacks = Array.from({length: MAX_ALA_STACKS + 1}, x => []);
+  }
+
+  handleStacks(event, stack = null) {
```
TsaboTavok: There is a fabricated event `changebuffstack` . It already unifies all the different buff events in to 1 that you can handle.

src/parser/rogue/outlaw/modules/talents/AlacrityTimesByStacks.js
```diff
@@ -0,0 +1,84 @@
...
+
+  on_byPlayer_removebuffstack(event) {
+    if (event.ability.guid !== SPELLS.ALACRITY_BUFF.id) {
+      return;
+    }
+    this.handleStacks(event);
+  }
+
+  on_fightend(event) {
+    this.handleStacks(event, this.averageAlacrityStacks);
```
TsaboTavok: Passing  `averageAlacrityStacks` here is wrong and not needed here, as you use it to set `lastAlaStack` , that is no longer needed as its the end of the fight.

src/common/SPELLS/rogue.js
```diff
@@ -476,8 +476,22 @@ export default {
     name: 'Blade Flurry',
     icon: 'ability_warrior_punishingblow',
   },
-
+  BLADE_RUSH_DAMAGE:{
```
TsaboTavok: I noticed that BladeRush Energy spell is missing and causing an error in the energy tab. Cam you also add that please.

src/parser/rogue/outlaw/modules/talents/QuickDraw.js
```diff
@@ -0,0 +1,46 @@
...
+    this.damageGained += calculateEffectiveDamage(event, DAMAGE_INCREASE) || 0;
+  }
+
+  statistic() {
+    return (
+      <TalentStatisticBox
+        talent={SPELLS.QUICK_DRAW_TALENT.id}
+        value={<ItemDamageDone amount={this.damageGained} />} />
```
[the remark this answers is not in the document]
HawkCorrigan: totally agree with that but I have no earthly clue on what else to present for that talent.
My other idea was just displaying both metrics (damage gained and cp gained) but I didnt know about any StatisticBox that supported that kind of thing.

I could just get rid of that module completely but I personally would rather have some information which is lacking something extra than no information about it at all.

src/parser/rogue/outlaw/modules/talents/Alacrity.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  get averageHaste() {
+    return this.alacrityTimesByStacks.averageAlacrityStacks*HASTE_PER_STACK;
+  }
+
+  statistic() {
+
+    return (
+      <StatisticBox
```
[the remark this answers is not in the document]
HawkCorrigan: I honestly don't know how often dropping of would happen in actual play.
From what I've seen you usually don't drop it off when having proper uptime on the boss if there are no errors in rotation.

My guess was that displaying number of times buff was dropped would mostly just turn up false positives but I might be completely wrong with that.

I would completely follow your call with that.

src/parser/rogue/outlaw/modules/talents/QuickDraw.js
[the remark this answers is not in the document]
TsaboTavok: Maybe just keep number or procs/cps gained.

src/parser/rogue/outlaw/modules/talents/Alacrity.js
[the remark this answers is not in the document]
TsaboTavok: You are right. It will always drop on fights like Vectis, but we cant suggest to swap the talent since alternatives are not viable.

In that case I will just suggest making it look like the weapon enchants and reuse at least some of the code.

src/parser/rogue/outlaw/modules/talents/QuickDraw.js
```diff
@@ -18,27 +18,40 @@ class QuickDraw extends Analyzer {
...
-    this.comboPointsGained +=
-      (event.resourceChange > 0 ? event.resourceChange - BASE_CP_GENERATED : 0) || 0;
+    const cpGeneratedByBroadside = this.selectedCombatant.hasBuff(SPELLS.BROADSIDE.id) ? 1 : 0;
+    const cpForEvent = (event.resourceChange || 0) + (event.waste || 0) - cpGeneratedByBroadside - BASE_CP_GENERATED;
+    this.procs += cpForEvent > 0 ? 1 : 0;
+    if(cpForEvent>0 && event.waste>0){
+        this.comboPointsWasted+=1;
+    }
+    if(cpForEvent>0){
+      console.log(event);
```
TsaboTavok: I think this needs to be removed or set to debug mode, but I'm not sure what the policy is

src/parser/rogue/outlaw/modules/talents/QuickDraw.js
```diff
@@ -18,27 +18,40 @@ class QuickDraw extends Analyzer {
...
     if (!this.selectedCombatant.hasBuff(SPELLS.OPPORTUNITY.id)) {
       return;
     }
-    this.comboPointsGained +=
-      (event.resourceChange > 0 ? event.resourceChange - BASE_CP_GENERATED : 0) || 0;
+    const cpGeneratedByBroadside = this.selectedCombatant.hasBuff(SPELLS.BROADSIDE.id) ? 1 : 0;
+    const cpForEvent = (event.resourceChange || 0) + (event.waste || 0) - cpGeneratedByBroadside - BASE_CP_GENERATED;
+    this.procs += cpForEvent > 0 ? 1 : 0;
+    if(cpForEvent>0 && event.waste>0){
+        this.comboPointsWasted+=1;
```
TsaboTavok: `comboPointsWasted` seems to be calculated but not used. I suggest removing it to keep it simple

HawkCorrigan:
Gonna close it for now and put in PRs in smaller batches over the next few days.

Issue #2883, opened 2018-12-26
Replace tooltips with react-tooltip-lite


MartijnHols:
I want to move to https://github.com/bsidelinger912/react-tooltip-lite. This allows us to finally have React components and regular elements in the tooltips without using the unsafe HTML attribute.

There's not a lot of work needed to implement this (basically just add the library to the `package.json` and start using it), but all of the existing tooltips would have to be migrated to prevent the debt and size consumption of two tooltip libraries.

This can be done in either `master` or `next`. `master` might be more troublesome since you might have to update whenever a PR is merged, while `next` will be more troublesome for *me* since I'll have to do that before releasing it :')

tl;dr:
1. add the new library. 
2. replace all data-tips with the new library. Convert all the strings with HTML in them to JSX.
3. Remove the `react-tooltip` dependency

sMteX:
I'lll try to look at it in upcoming days

MartijnHols:
Thanks @sMteX! Please don't forget to claim the bounty (it's prepaid).

PR #2884, opened 2018-12-27
[NEXT] Tooltip library update

sMteX:
My current progress on #2883, for Zero's reviews and feedback.


src/interface/report/Results/Timeline/Lane.js
```diff
@@ -58,28 +59,32 @@ class Lane extends React.PureComponent {
...
     return (
-      <div
+      <Tooltip
         key={`recharge-${left}`}
+        content="Charge Restored"
         className="recharge"
-        style={{
+        wrapperStyles={{
           left,
+          position: 'absolute',
```
MartijnHols: Is this necessary? It should have been set by the CSS

sMteX:
As I mentioned in DM's on Discord, the issue with manually setting `display` and `position` properties is that although CSS takes care of that (like in the Timeline for example), the problem is in the library itself. Instead of using CSS files, they render the style props directly onto the wrapper component 
https://github.com/WoWAnalyzer/react-tooltip-lite/blob/322867d6273f4a8cb4afcba496be66844ad97eb6/src/index.jsx#L147

So even if the classname of the wrapper component and tag name is correct and the styles actually *are* applied, these inline styles override it, so I have to override it there too. 

Solution would be to do a little rehaul of the CSS in the library, extracting as much as possible into a css file, so the least amount of CSS is inlined, then we can leverage the CSS classes a little more and simplify the code.

PR #2890, opened 2018-12-29
[Next] Updated Player Select

Alastair-Scott:
Added a player overview when selecting a player.
Double clicking the player will skip this and go straight to analysis


src/interface/playerSelection/PlayerInfo.js
```diff
@@ -0,0 +1,107 @@
...
+        </div>
+      </>
+    );
+  }
+
+  get averageIlvl() {
+    const gear = this.state.gear;
+
+    // eslint-disable-next-line no-restricted-syntax
+    const filteredGear = gear.filter(g => g.itemLevel !== 0 && !EXCLUDED_ITEM_SLOTS.find(x => x === gear.indexOf(g)));
```
MartijnHols: Why the check for item level 0? Does this ignore empty slots?

MartijnHols:
```suggestion
    const filteredGear = gear.filter((item, slotId) => item.itemLevel !== 0 && !EXCLUDED_ITEM_SLOTS.includes(slotId));
```

several things:
- please give variables a proper name
- looping inside a loop is bad practice
- you can get the index as a second param, no need to do indexOf

src/interface/playerSelection/PlayerInfo.js
```diff
@@ -0,0 +1,107 @@
...
+    return filteredGear.reduce( ( total, item ) => total + item.itemLevel, 0 ) / filteredGear.length;
+  }
+}
+
+function _parseTalents(talents) {
+  const _talentsByRow = {};
```
MartijnHols:
```suggestion
  const talentsByRow = {};
```

Please don't prefix local variables

MartijnHols: I'd prefer it if you used a `reduce` instead.

src/interface/playerSelection/PlayerSelection.scss
```diff
@@ -1,87 +1,284 @@
...
+      .player-background {
+        background-position: center;
+        background-size: cover;
+      }
+
+      .player-gear {
+        display: grid;
+        grid-template-columns: 45px 15px auto 30px 15px 45px 45px 15px 30px auto 15px 45px;
+        grid-template-rows: 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px 20px;
+        grid-template-areas:
```
MartijnHols: This looks very fragile and I would prefer any other solution. You could hardcode the locations, or you could make three elements (two for the columns and one for the row) and sort and group the items to let CSS auto-position them. I don't think the CSS is the best place to decide where items are rendered.

MartijnHols: I checked the armory, and they actually use the three elements approach.

src/interface/playerSelection/PlayerTile.js
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  timeout = null;
+
+  render() {
+    const { player, selectedPlayer, analysisUrl, handleClick, history } = this.props;
+    return (
+      <span
+        key={player.guid}
+        onClick={() => { this.timeout = setTimeout(function(){ handleClick(); }, 200); }}
```
MartijnHols: Do browsers not have this behavior by default? I remember something like that being the case.

src/interface/playerSelection/PlayerTile.js
```diff
@@ -0,0 +1,43 @@
...
+        onDoubleClick={() => { clearTimeout(this.timeout); history.push(analysisUrl); }}
+        className={`player${selectedPlayer && selectedPlayer.guid === player.guid ? ' selected' : ''}`}
```
MartijnHols: Please use curly brackets around ternaries with multiple conditions as operator priority in JS is weird and omitting curly brackets might lead to bugs.

src/interface/playerSelection/PlayerTile.js
```diff
@@ -0,0 +1,43 @@
...
+      >
+        <div className="card">
+            <div className="avatar" style={{ backgroundImage: `url(${player.avatar})` }}>
+              <RoleIcon id={player.spec.role} className="role-icon" />
+            </div>
+          <div className="about">
+            <h1 className={player.spec.className.replace(' ', '')}>{player.name}</h1>
+            <small>{`${player.spec.specName} ${player.spec.className}`}</small>
```
MartijnHols:
```suggestion
            <small>{player.spec.specName} {player.spec.className}</small>
```

src/interface/playerSelection/index.js
```diff
@@ -57,39 +58,47 @@ class PlayerSelection extends React.PureComponent {
...
-                  <div className={`background ${spec.className.replace(' ', '')}-bg`} />
-                </div>
-              </div>
-            </Link>
-          );
-        })}
+      <div className={`player-selection-container${this.state.selectedPlayer ? ' show-info' : ''}`}>
+        <div className="player-selection">
+          {players.sort(sortPlayers).map(player => {
+            return (
```
MartijnHols:
```suggestion
          {players.sort(sortPlayers).map(player => (
```

src/interface/playerSelection/playerInfo/Azerite.js
```diff
@@ -0,0 +1,115 @@
...
+import Paladin from 'common/SPELLS/bfa/azeritetraits/paladin';
+import Shaman from 'common/SPELLS/bfa/azeritetraits/shaman';
+import Warlock from 'common/SPELLS/bfa/azeritetraits/warlock';
+import Monk from 'common/SPELLS/bfa/azeritetraits/monk';
+import DeathKnight from 'common/SPELLS/bfa/azeritetraits/deathknight';
+import Priest from 'common/SPELLS/bfa/azeritetraits/priest';
+import Druid from 'common/SPELLS/bfa/azeritetraits/druid';
+import Unused from 'common/SPELLS/bfa/azeritetraits/__UNUSED';
+
+const AZERITE_SPELLS = Object.values({...General, ...Hunter, ...Paladin, ...Shaman, ...Warlock, ...Monk, ...DeathKnight, ...Priest, ...Druid, ...Unused});
```
MartijnHols: I think you can just use this instead:

```jsx
import AZERITE_SPELLS from 'common/SPELLS/bfa/azeritetraits';
```

src/interface/playerSelection/playerInfo/Azerite.js
```diff
@@ -0,0 +1,115 @@
...
+    const azerite = this.props.azerite;
+
+    return (
+      <>
+        <h3>
+          Azerite Powers
+        </h3>
+        <div className="azerite-traits">
+          {Object.keys(azerite).map(slotId => {
+            return (
```
MartijnHols:
```suggestion
          {Object.keys(azerite).map(slotId => (
```

src/interface/playerSelection/playerInfo/PlayerGearHeader.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PlayerGearHeader = ({ player, averageIlvl }) => {
+
+
+  return (
```
MartijnHols:
```suggestion
const PlayerGearHeader = ({ player, averageIlvl }) => (
```

src/interface/playerSelection/playerInfo/Talents.js
```diff
@@ -0,0 +1,60 @@
...
+    const rows = [15, 30, 45, 60, 75, 90, 100];
+
+    return (
+      <>
+        <h3>
+          Talents
+        </h3>
+        <div className="talent-info">
+          {talents.map((spellId, index) => (
+            <div className="talent-info-row" key={index} style={{ marginBottom: '0.8em', fontSize: '1.3em' }}>
```
MartijnHols: Super nit pick, but I generally prefer `key` to be the first property always rather than anarchy.

src/interface/playerSelection/playerInfo/Gear.js
```diff
@@ -0,0 +1,51 @@
...
+  render() {
+    const { gear } = this.props;
+
+    return (
+      <>
+        {
+          gear.filter(item => item.id !== 0)
+          .map(item => {
+            // Items seem to turn epic from 340 item level, but WCL doesn't show this properly
+            let quality = item.itemLevel >= 340 ? ITEM_QUALITIES.EPIC : item.quality;
```
MartijnHols: Please move 340 to a constant above the class.

src/common/enchantIdMap.js
```diff
@@ -0,0 +1,44 @@
+/*
+  Professions Included:
+    - Enchanting
+    - Engineering
+    - Tailoring
+*/
+
+export default {
```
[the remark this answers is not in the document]
Alastair-Scott: I got the bases for this by setting up a web scraper from wowhead. If needed, I could look into creating a script that can generate this in python or something in the future

src/interface/playerSelection/PlayerInfo.js
Alastair-Scott: I think that's why I had it in there. It's a while ago since I last looked into this :)

Alastair-Scott: Much cleaner way of doing it - changed as suggested

src/interface/playerSelection/PlayerSelection.scss
Alastair-Scott: Fair enough :*(

src/interface/playerSelection/playerInfo/Talents.js
Alastair-Scott: I've put key first everywhere else, not sure why I didn't for this one!

src/common/enchantIdMap.js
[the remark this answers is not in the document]
Alastair-Scott: Aye that'd be a good idea

src/interface/playerSelection/playerInfo/Gear.js
Alastair-Scott: Done

src/interface/playerSelection/playerInfo/PlayerGearHeader.js
Alastair-Scott: Hadn't realised you could do this. All done

src/interface/playerSelection/playerInfo/Azerite.js
Alastair-Scott: Done

src/interface/playerSelection/playerInfo/Azerite.js
Alastair-Scott: Sorted

src/interface/playerSelection/PlayerInfo.js
Alastair-Scott: I used a reducer for both gear and talents now, but left traits as is because it's a little more involved.

src/interface/playerSelection/PlayerTile.js
Alastair-Scott: It wasn't working for me at any rate. When I was double clicking without the timeout, I'd find that the player would first be clicked and show the right panel, and then the position of the cards would change, and the second click would be on a different person and analyze that person.

Adding the timeout fixed the issue

src/interface/playerSelection/PlayerInfo.js
MartijnHols: Unequipped item slots do still count toward the average though.

Also is there a way to detect if gear was broken?

Alastair-Scott: It was actually for something else I was trying and forgot to take out. Removed it now.

Sadly that's not something we have data for

yajinni:
Just wondering, how is someone supposed to know double clicking will analyze that character? I assume most will select and hit the analyze button.

MartijnHols:
thanks!

PR #2898, opened 2018-12-30
[BrM] Added "Straight, No Chaser" trait

emallson:
This trait is annoying as hell. It adds a Grand Crusader-like cooldown reset that is untrackable, but also tied to our most important defensive ability.

This uses the same method as GC to estimate the lower bound on the number of charges reset, and ties that into the general brew CDR code.
screenshot from 2018-12-30 00-20-09
screenshot from 2018-12-30 00-17-16


src/parser/monk/brewmaster/modules/spells/azeritetraits/StraightNoChaser.js
```diff
@@ -0,0 +1,161 @@
...
+        return 1;
+      }
+
+      return n / k * binom(n-1, k-1);
+    }
+
+    // pmf of the binomial distribution
+    const reset_prob = (i) => binom(this.isbCasts, i) * Math.pow(SNC_PROC_CHANCE, i) * Math.pow(1 - SNC_PROC_CHANCE, this.isbCasts - i);
+
+    const reset_probs = Array.from({length: this.isbCasts}, (_x, i) => {
```
sMteX: I guess I understand the point here, but wouldn't the graph look a little nicer if it was from like.. first half only? It seems to reach the 0 asymptote very quickly and looks a little harder to read this way? Just an idea, if it's not viable or doesn't make sense, keep it this way.

src/parser/monk/brewmaster/modules/spells/azeritetraits/StraightNoChaser.js
```diff
@@ -0,0 +1,161 @@
...
+    return (
+      <TraitStatisticBox
+        trait={SPELLS.STRAIGHT_NO_CHASER.id}
+        value={(
+          <>
+          ≥{this.resets} Charges Gained<br />
+          {formatNumber(this.avgArmor)} Armor Gained
+          </>
+        )}
+        tooltip={`There are no logged events for SNC's generated brew charges, so this is an <em>estimate</em> based on casts that occurred while your brews are on cooldown. <b>If you have low cast efficiency, this will be <em>underestimated!</em></b><br/><br/>Straight, No Chaser gave <b>${this.armor}</b> armor, and was up for ${formatPercentage(this.selectedCombatant.getBuffUptime(SPELLS.STRAIGHT_NO_CHASER_BUFF.id) / this.owner.fightDuration)}% of the fight.`}
```
sMteX: The buff uptime would be nicer as a getter than inlined here.

src/parser/monk/brewmaster/modules/spells/azeritetraits/StraightNoChaser.js
```diff
@@ -0,0 +1,171 @@
...
+    // pmf of the binomial distribution
+    const reset_prob = (i) => binom(this.isbCasts, i) * Math.pow(SNC_PROC_CHANCE, i) * Math.pow(1 - SNC_PROC_CHANCE, this.isbCasts - i);
+
+    const reset_probs = Array.from({length: this.isbCasts}, (_x, i) => {
+      return { x: i, y: reset_prob(i) };
+    });
+
+    const RANGE_THRESHOLD = 0.001;
+    const rangeMin = reset_probs.findIndex(({y}) => y >= RANGE_THRESHOLD);
+    const rangeMax = reset_probs.slice(rangeMin).findIndex(({y}) => y < RANGE_THRESHOLD);
```
sMteX: This is most likely wrong - `slice()` returns a shallow copy of the original array, so `findIndex` doesn't preserve the original indexes (tried in console):
```
const a = [0, 0, 1, 2, 1, 0, 0];
console.log(a.findIndex(x => x > 0.001));
> 2
console.log(a.slice(2).findIndex(x => x < 0.001));
> 3
```
Adding the `rangeMax` to `rangeMin` to get the real `rangeMax` might work though - even in the example, 2 + 3 gives 5 which is the index of the first 0 on the right side.

-----------
Other solution might be to completely filter the low values anyway and pass those into the chart directly?
```javascript
// (make reset_probs let) or add the filter after Array.from()
reset_probs = reset_probs.filter(({ y }) => y > RANGE_THRESHOLD);
```

emallson: good catch! the log i had pulled up happened to look ok so i didn't see it. should be fixed in a moment.

emallson: screenshot from 2019-01-04 16-42-42
looks like this now

PR #2899, opened 2018-12-30
add browser search via opensearch

### LINKED ISSUE #1940: Implement browser search
Add "search" so you can type wowanalyzer in the omnibar, hit TAB and then paste url to parse it

ForsakenInvocations:
added an [opensearch](https://github.com/dewitt/opensearch) description, so you can search by report code directly in the omnibar / search bar
should fix #1940


MartijnHols:
Hey thanks for the PR. 

This is good to go, but it doesn’t yet close #1940 as a requirement of that is that it accepts urls as well as report codes, just like the report bar on the homepage. If that’s not possible, let me know and in that case this will suffice. 

Merging this since it is a useful and complete contribution, even if it doesn’t yet close the related issue. Thanks!

ForsakenInvocations:
Hey.
As far as I know Open Search does not support multiple search schemes. The only possibility would be to add two search engines to the website. One to search by report code and another one to search by URL.
So if you really need just one engine for both then I will take another look at it but that would demand a lot of effort, about another three hours, so a higher bounty would be justified. If you are fine with having two engines then I will create a pull request for you to take a look at it.

Best regards,
Chris

PR #2900, opened 2018-12-30
Mark Prot Paladin Supported

emallson:
screenshot from 2018-12-30 12-06-43

(also updated the description for brm because it needed doing badly)

**DO NOT MERGE BEFORE #2893**


emallson:
Probably, but the checklist is in a much better state now than it was.

I'm so hyped for this to get merged though \o/

Issue #2880, opened 2018-12-26
[BrM] Present Mitigation Value a la Teachings of the Red Crane sheet


emallson:
The [TotRC spreadsheet](https://www.peakofserenity.com/bfa/mistweaver/spreadsheet/) evaluates the performance of different traits and trinkets based on an actual log. I want to replicate that functionality for brewmasters in the analyzer. The goal here is to tell two things:

1. How much "effective healing" was provided by each stat, and how does that compare to the amount of stats from gear?
2. How much EHP was provided by each stat, and how does that compare to the amount of stats from gear?

where "effective healing" is either damage mitigated or healing done.

See [`feature/brm-sheet`](https://github.com/emallson/WoWAnalyzer/compare/master...emallson:feature/brm-sheet) for the current codebase.

Stats to Evaluate:

*EHRPS:*

- [x] Armor
- [x] Agility (Dodge)
- [x] Agility (Stagger)
- [x] Agility (GotOx)
- [x] Mastery (Dodge)
- [x] Mastery (GotOx)
- [x] Versatility (DR)
- [x] Versatility (Healing)
- [x] Crit (Celestial Fortune)
- [x] Crit (Critical Heals)
- [ ] Haste (extra Purifies)
- [ ] Validate splitting of heals (compare to holy paladin; lower priority because our heals are pretty shit)

*EHP:*

- [ ] HP
- [ ] Armor
- [ ] Agility
- [ ] Versatility
- [ ] Avoidance (?)

abelito75:
I know TotRC is just a reference but since it has been sacked for about a year you might want to change the name and remove the link :P also is this still not fully done or naw

abelito75:
Is this still something looking to be completed?

PR #2901, opened 2018-12-30
[BrM] Stat Values tab

emallson:
Show stat values for brewmasters in a "Mitigation Value" tab.

See #2880 for more context.

~~Includes changes from #2881, so that should be reviewed/merged first.~~

Scroll down for updated screenshots.
screenshot from 2018-12-30 13-46-57
screenshot from 2018-12-30 13-46-05

See the markdown file for details on calculations (it is also linked from the tab, I just haven't updated the screenshot).


emallson:
My handling of crit here is actually slightly wrong because I forgot to include the 10% base crit chance that brewmasters get, which inflates the value of crit rating. x.x

emallson:
would like to note that i spoke with one of the MW TCs on how to handle crit valuation yesterday, and they recommended a different method than what I used here. I'm not confident that what they proposed will work for us because we have so few crits that *don't* overheal, but at some point I'm going to try it.

which is to say I think it'd be fine to accept the PR with the current crit code, but it may change soon:tm:

emallson:
temporarily closing this after more thought. i have a couple of changes to make before i'm ready for it to go live (and i don't really want it going live til dazaralor anyway)

emallson:
screenshot from 2019-01-12 14-42-29

Ok, I've dealt with the main issues I had, which were:

1. It was difficult to tell if a stat's value was coming from *skew* (i.e. someone with vers RA, or stacking vers enchants+gear) or from a stat's inherent value. By showing the amount of rating and the value per-rating, we can avoid this.
2. Changed from using the amount of stats from the *next* 5 ilvls to the *previous* 5 ilvls. This is mostly a semantic change, but basically we can't evaluate adding stats the same way as evaluating stats you already have. Since I'm evaluating stats you already have, the new way is internally consistent.

src/parser/monk/brewmaster/modules/features/MitigationSheet.js
```diff
@@ -0,0 +1,460 @@
...
+    const gear = this.selectedCombatant._combatantInfo.gear.filter(({id, itemLevel}) => id !== 0 && itemLevel > 1);
+    return gear.reduce((sum, {itemLevel}) => sum + itemLevel, 0) / gear.length;
+  }
+
+  constructor(...args) {
+    super(...args);
+
+    const fight = this.owner.fight;
+    if(fight.size === 5) {
+      this.K = MPLUS_K;
```
MartijnHols: I think this should be a getter. Also the name should be longer, how about `kFactor`? Just K is going to be hard to work with.

emallson:
I can work on splitting the mitigationsheet file into per-stat files, that seems like a totally reasonable request (and probably would've made it much easier to review, sorry).

The reason I deviated from the existing stat values layout is twofold:

**First:** I have (in-progress) plans to extend this to include azerite traits and trinkets, several of which actually have names long enough that they break the existing statbox-based layout (this may no longer be a problem in `next`).

**Second:** The existing layout isn't wide enough to have multiple columns. The issue with only having the normalized column basically boils down to mastery in combination with the low scaling of secondaries on gear. Our mastery is *insane* on most fights if you look at it point-for-point. It is consistently 15-20x as powerful as agility per point, and 2-6x as good as armor.

If you only had normalized values, I am very concerned that the *perception* would be that you should always take gear with mastery on it, even at large ilvl drops. Additionally showing the rating scaled by the number of points you got relative to ilvl addresses this perception by showing that **per ilvl** mastery is similar in value to both agility and armor, so you should still be taking ilvl over mastery as their combined value outweighs the value of using lower-ilvl mastery gear.

That said: if that isn't clear then it obviously isn't accomplishing the stated goal. I opted for both because we still have things like enchants, gems, and stat food where the amount of stat gained is much closer, which gives the normalized column value.

I like the idea of a toggle hiding the breakdowns by default, that should help the table feel less overwhelming.

MartijnHols:
Splitting the file up is not required to have this merged btw. This iteration just needs a changelog entry.

MartijnHols:
kValue?

src/parser/monk/brewmaster/modules/constants/Mitigation.js
```diff
@@ -9,8 +9,25 @@ export const ULDIR_K = [
...
+   9311.4, // Normal
+  10275.3, // Heroic
+  11390.4, // Mythic
+];
 
 export function diminish(stat, K) {
   return stat / (stat + K);
 }
+
+export function lookupK(fight) {
```
MartijnHols: We're probably going to want multi-raid support soon, especially with the raid release around 8.1.5

emallson: agreed, but the only way to do that based on available info is to map `boss id -> instance` and i didn't want to tackle that yet.

MartijnHols: We should have the data in the raids folder. Maybe make it a part of the raid config file?

emallson: i'll check into it :+1:

PR #2906, opened 2019-01-03
[#1668] Message when parsing log from different patch

Alastair-Scott:
Wrapped the fight selector with a new check.

Due to the Prepatch and Legion sites being removed, I've just provided text warnings for the moment, but this can be changed to a redirect as needed.

Also, very happy if someone has some better ideas for the text.


emallson:
can we get a screenshot? i skimmed it during raid break tonight and didn't see any glaring issues; i'll take a better look this weekend.

Alastair-Scott:
Previous Expansion: 
image

Previous patch:
image

PR #2919, opened 2019-01-05
[#1668] Updated PatchChecker to provide a link to older versions of the site as we handle them

Alastair-Scott:
Due to the positioning of this element, I swapped out the makeAnalyzerUrl function for window.location in a couple of places in this file.

If you use a link that goes directly to a fight or a particular player, this element does not have access to that, and if you click continue, it'd just go back to fight selection.


src/common/PATCHES.js
```diff
@@ -0,0 +1,16 @@
+const PATCHES = [
+  {
+    name: '8.0',
+    timestamp: 1534197600000,
```
Dambroda: Could you leave a comment somewhere with which region/tmz you are using for these timestamps? For maintainability on following patches.

src/interface/report/PatchChecker.js
```diff
@@ -52,7 +53,13 @@ class PatchChecker extends React.PureComponent {
     const reportTimestamp = report.start;
     const reportDate = new Date(report.start).toString();
 
-    if (reportTimestamp >= LATEST_PATCH || this.continue) {
+    const expansionStartPatch = PATCHES[0];
+    // report's patch is the last patch where the reports timestamp is higer than the patch timestamp
+    // find works on first item so we reverse the array to start from the end
+    // use slice to copy the array first so we don't mutate PATCHES
+    const reportPatch = PATCHES.slice().reverse().find(patch => reportTimestamp > patch.timestamp);
```
Dambroda: Rather than doing a reverse (which relies on the the array being updated in a specific order), maybe a sort instead?

src/interface/report/PatchChecker.js
```diff
@@ -61,20 +68,31 @@ class PatchChecker extends React.PureComponent {
...
                   {
-                    reportTimestamp >= EXPANSION_START ? (
+                    reportTimestamp >= expansionStartPatch.timestamp ? (
                       <Trans>
                         WoWAnalyzer is constantly being updated to support the latest changes. This can cause some functionality to be modified for the latest talents/traits/trinkets or be removed.
                         <br /><br />
                         This could mean that some parts of your report will no longer be analysed accurately.
                         <br /><br />
-                        If you would still like to view the analysis, you can click 'Continue anyway' below. 
+                        If you would like to view the analysis on an onlder version of WoWAnalyzer, please
```
Dambroda:
```suggestion
                        If you would like to view the analysis on an older version of WoWAnalyzer, please 
```

src/interface/report/PatchChecker.js
```diff
@@ -61,20 +68,31 @@ class PatchChecker extends React.PureComponent {
 
           <div className="panel">
             <div className="panel-heading">
-              <h2><Trans>{reportTimestamp >= EXPANSION_START ? "This report is for an earlier patch" : "Sorry, this report is for a previous expansion" }</Trans></h2>
+              <h2><Trans>{reportTimestamp >= expansionStartPatch.timestamp ? "This report is for an earlier patch" : "Sorry, this report is for a previous expansion" }</Trans></h2>
```
MartijnHols:
```suggestion
              <h2><Trans>{reportTimestamp >= expansionStartPatch.timestamp ? 'This report is for an earlier patch' : 'Sorry, this report is for a previous expansion'}</Trans></h2>
```

PR #2920, opened 2019-01-06
[Warlock] Probability playground

sMteX:
I got inspired by emallson and Putro and tried to make usage of some probability distributions in my Warlocks. Specifically, I looked at classic [binomial distribution](https://en.wikipedia.org/wiki/Binomial_distribution) and its generalized neighbor, [Poisson binomial distribution](https://en.wikipedia.org/wiki/Poisson_binomial_distribution) (the difference is merely in the fact that binomial has fixed probability chance for all events whereas Poisson can have different probabilities for each event).

Originally I wanted to implement it in more places but I'd rather not inflate this PR any more than needed. Added both distributions, their [PMF](https://en.wikipedia.org/wiki/Probability_mass_function) and [CDF](https://en.wikipedia.org/wiki/Cumulative_distribution_function) functions and an example of implementing it.

With these functions you can show:
- what is the % chance of exactly K successful out of N events
- what is the % chance of getting <= K successful out of N events
- draw a distribution graph similar to emallson's Expected mitigation by Mastery
- get the number of events (k) you were *most likely* to get given the probabilities

I would personally show almost all of these but I've talked about it with the Warlock theorycrafter Motoko and we agreed that for majority of the people, it could be just confusing, so in the end I only used the last thing and shown the % of actual procs vs expected procs, to see if they were (un)lucky or if it's just normal.

(I might add an information what was the expected amount of procs in the tooltip, awaiting response from him)

image

### DIFF
--- a/src/parser/warlock/affliction/modules/talents/SoulConduit.js
+++ b/src/parser/warlock/affliction/modules/talents/SoulConduit.js
@@ -9,7 +9,7 @@ import SpellLink from 'common/SpellLink';
 
 import StatisticListBoxItem from 'interface/others/StatisticListBoxItem';
 
-import { binomialCDF } from 'parser/warlock/shared/probability';
+import { binomialPMF } from 'parser/warlock/shared/probability';
 import { UNSTABLE_AFFLICTION_DEBUFFS } from '../../constants';
 import SoulShardTracker from '../soulshards/SoulShardTracker';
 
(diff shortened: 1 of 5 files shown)


src/parser/warlock/affliction/modules/talents/SoulConduit.js
```diff
@@ -39,11 +41,24 @@ class SoulConduit extends Analyzer {
     const avgDamage = this._totalUAdamage / (this._totalTicks > 0 ? this._totalTicks : 1);
     const shardsGained = this.soulShardTracker.getGeneratedBySpell(SPELLS.SOUL_CONDUIT_SHARD_GEN.id);
     const estimatedUAdamage = shardsGained * TICKS_PER_UA * avgDamage;
+    const totalSpent = this.soulShardTracker.spent;
+    // since we want to get the amount of shards we would MOST LIKELY get, we intentionally need to try higher values for k than usually possible
+    const { partial: probabilities } = binomialCDF(Math.round(totalSpent / 2), totalSpent, SC_PROC_CHANCE);
```
emallson: so if i understand this right, you're calling it with a reasonably high value just to get it to compute the first `Math.round(totalSpent / 2)` entries in `partial`?

src/parser/warlock/affliction/modules/talents/SoulConduit.js
```diff
@@ -39,11 +41,24 @@ class SoulConduit extends Analyzer {
...
+    // with 15% chance per shard, it's very unlikely that we would get half of the shards refunded, so the maximum must be somewhere lower than that, and "partial" array of the return object contains those values
+    let maxP = 0;
+    let max = -1;
+    probabilities.forEach((p, k) => {
+      if (p > maxP) {
+        maxP = p;
+        max = k;
+      }
+    });
```
emallson: this could be "simplified" to a `reduce`:

```js
// argmax(probabilities) -- if only Math.argmax existed
const max = probabilities.reduce((max, p, k, probs) => (p > probs[max]) ? k : max, 0);
```

now, whether this is desirable or not is up for debate. i'll leave which version to use up to you

src/parser/warlock/affliction/modules/talents/SoulConduit.js
sMteX: Not sure if I understood you right, I'm calling CDF (even though I don't need it), because in `partial` I get PMF values for `k` values lower than `Math.round(totalSpent / 2)`. Since it's a binomial distribution, the pattern looks roughly like this 

image

and I just want to know, at which `k` the probability is the highest, the peak of the function. Same goes for Poisson binomial distribution, for fun I charted it in Excel and it looks roughly the same.

src/parser/warlock/affliction/modules/talents/SoulConduit.js
sMteX: While it's an elegant one-liner, I think for the sake of readability, I'd prefer my solution (I had to look up the third and fourth argument to `reduce`)

sMteX:
After some testing:
- Poisson Binomial Distribution really does look similar to Binomial Distribution (here's a real log example of getting 0 - 44 procs from 44 casts)
image
- calculating Poisson PMF even repeatedly for worst case scenarios is blazing fast with such low N values we use here. Because of that, I dropped the returned lookup table and I could simplify the code by just using PMF all the time. Left the CDF functions for possible future uses.

src/parser/warlock/affliction/modules/talents/SoulConduit.js
emallson: sorry, i wasn't clear, but i figured it out by talking with you on discord (the question was specifically why `Math.round(totalSpent / 2)` and not just `totalSpent`).

Issue #2706, opened 2018-11-18
Add a way to precisely determine phases


MartijnHols:
We need a way to determine boss phase start/end. Preferably down to the level of knowing when ad waves spawn within a phase (micro phases).

WCL fights endpoint does list phases, but not timings of them. Kihra mentioned possibly exposing this in the API but I wouldn't expect that any time soon without constant nagging. Alternatively we could try the Wipefest API but that adds latency, another point of failure and no guarantees. Long story short it would be nice if we can figure out a way to determine this ourselves. But then how to make this well maintainable?

We could do 1 extra API call to determine this but that would be troublesome for high traffic moments. We could gracefully downgrade to non-phase support then, but that might make usage of this much harder. Maybe the API call is worth it given the gain. Plus Kihra mentioned possibly adding support for "additional queries" support in the events API, which could help us reduce this back to 1 call when that arrives.

kfinch:
For the extra API call issue, could add a "show phases" button and not make the call until user presses it.

MartijnHols:
1 extra API call is acceptable. I do like that idea as phase selection will probably be behind a toggle anyway, but micro phases (which lets us know when adds are up) will be required before parsing.

MartijnHols:
Fixed in #2923

PR #2923, opened 2019-01-07
[#2706] Phase Support

Alastair-Scott:
Implemented boss fight phase support

This works for phases triggered by the following event types:
- Cast
- Begincast
- Removebuff
- Applybuff
- Removedebuff

We also have functionality to support add waves as micro phases. This is currently set up for the Nerubian Voidweavers on Zek'vos


src/common/SPELLS/encounter.js
```diff
@@ -1374,6 +1374,11 @@ export default {
     name: "Titan Spark",
     icon: "item_holyspark",
   },
+  ZEKVOS_ENCOUNTER_EVENT: {
```
Dambroda:
```suggestion
  ZEKVOZ_ENCOUNTER_EVENT: {
```

src/raids/uldir/Zekvoz.js
```diff
@@ -20,5 +23,86 @@ export default {
...
+        },
+      },
+      {
+        id: '2',
+        name: 'Stage Two: Deception',
+        difficulties: [FIGHT_DIFFICULTIES.NORMAL, FIGHT_DIFFICULTIES.HEROIC],
+        filter: {
+          type: 'cast',
+          ability: {
+            id: SPELLS.ZEKVOS_ENCOUNTER_EVENT.id,
```
Dambroda:
```suggestion
            id: SPELLS.ZEKVOZ_ENCOUNTER_EVENT.id,
```

src/raids/uldir/Zekvoz.js
```diff
@@ -20,5 +23,86 @@ export default {
...
+        },
+      },
+      {
+        id: '2',
+        name: 'Stage Two: Corruption',
+        difficulties: [FIGHT_DIFFICULTIES.MYTHIC],
+        filter: {
+          type: 'cast',
+          ability: {
+            id: SPELLS.ZEKVOS_ENCOUNTER_EVENT.id,
```
Dambroda:
```suggestion
            id: SPELLS.ZEKVOZ_ENCOUNTER_EVENT.id,
```

src/raids/uldir/Zekvoz.js
```diff
@@ -20,5 +23,86 @@ export default {
...
+        },
+      },
+      {
+        id: '3',
+        name: 'Stage Three: Corruption',
+        difficulties: [FIGHT_DIFFICULTIES.NORMAL, FIGHT_DIFFICULTIES.HEROIC],
+        filter: {
+          type: 'cast',
+          ability: {
+            id: SPELLS.ZEKVOS_ENCOUNTER_EVENT.id,
```
Dambroda:
```suggestion
            id: SPELLS.ZEKVOZ_ENCOUNTER_EVENT.id,
```

src/common/SPELLS/encounter.js
```diff
@@ -1374,6 +1374,11 @@ export default {
     name: "Titan Spark",
     icon: "item_holyspark",
   },
+  ZEKVOZ_ENCOUNTER_EVENT: {
```
MartijnHols: Please add this as a constant to the top of the `Zekvoz.js` file. One of the long term goals is to drop the SPELLS files and only put the spells close to relevant places that also use it so that they can be code split properly. Looking at your usage, I reckon you can put it in the Zekvoz file just file since that should always be loaded when this spell object is relevant.

src/common/SPELLS/encounter.js
```diff
@@ -1721,6 +1726,11 @@ export default {
     name: "Reorigination Blast",
     icon: "ability_druid_cresentburn",
   },
+  GHUUN_REORIGINATION_BLAST_3: {
```
MartijnHols: Same but Ghuun.js.

src/interface/report/BossEventLoader.js
```diff
@@ -0,0 +1,169 @@
...
+
+export class BossEventsLoadError extends ExtendableError {
+  reason = null;
+  constructor(reason) {
+    super();
+    this.reason = reason;
+  }
+}
+
+class BossEventsLoader extends React.PureComponent {
```
MartijnHols:
```suggestion
class BossEventLoader extends React.PureComponent {
```

Please try to keep the class name the same as its file

src/interface/report/BossEventLoader.js
```diff
@@ -0,0 +1,169 @@
...
+import { fetchEvents, LogNotFoundError } from 'common/fetchWclApi';
+import { captureException } from 'common/errorLogger';
+import makeAnalyzerUrl from 'interface/common/makeAnalyzerUrl';
+import { findByBossId } from 'raids';
+
+import Ghuun from './images/Ghuun.gif';
+import handleApiError from './handleApiError';
+import './EventParser.css';
+
+export class BossEventsLoadError extends ExtendableError {
```
MartijnHols:
```suggestion
export class BossEventLoadError extends ExtendableError {
```

src/interface/report/index.js
```diff
@@ -23,42 +24,50 @@ const Report = props => (
           refreshReport={refreshReport}
         >
           {fight => (
-            <PlayerSelection
+            <BossEventLoader
```
MartijnHols: This will currently hold back player selection and all the rest until this is done. This slows down the initial load when opening a direct link to a player analysis page. Are there any objections for adding this to the EventParser next to the parser, character profile and events loaders so that it happens async with those calls?

src/interface/report/BossEventLoader.js
```diff
@@ -0,0 +1,169 @@
...
+  loadEvents() {
+    const { report, fight } = this.props;
+
+    const bossConfig = findByBossId(fight.boss);
+    if (bossConfig.fight.phases && bossConfig.fight.phases.length !== 0) {      
+      const filters = [];
+
+      const phasesForDifficulty = bossConfig.fight.phases.filter(phase => phase.difficulties.includes(fight.difficulty));
+
+      const abilities = phasesForDifficulty.reduce((abilityIds, phase) => {
```
MartijnHols: Why do abilities get special treatment? I don't understand the way you're building the filter; shouldn't the abilities be a part of the filter they're listed with instead of being a global filter? The global filter seems like it might restrict what we can make for filters.

src/raids/uldir/Ghuun.js
```diff
@@ -18,5 +21,35 @@ export default {
       ],
       magical: [],
     },
+    phases: [
+      {
+        id: '1',
```
MartijnHols: can we make this an int?

src/interface/report/EventParser.js
```diff
@@ -127,7 +128,7 @@ class EventParser extends React.PureComponent {
         this.stopFakeNetworkProgress();
         timeAvailable && console.time('full parse');
         const parser = new parserClass(report, player, fight, combatants, characterProfile);
-        return this.parseEvents(parser, report, player, fight, events);
+        return this.parseEvents(parser, report, player, fight, [...bossEvents, ...events]);
```
MartijnHols: We're going to need this before the parse events method is called for phase selection in the future, would it be possible to make a list of all the phasestart/phaseend events before this and then pass that as-is to the events array? i.e. Instead of determining the events in the normalizer, how about doing it on the line above this and just letting the normalizer insert the events at the correct index?

src/common/SPELLS/encounter.js
sMteX: Not related to the review, but while it makes sense somewhere, wouldn't that lead to a LOT of code duplication when spells are used in more places? especially for things like SpellLink etc.?

MartijnHols: It can be exported if it is used elsewhere. As for SpellLink I have no solution yet. This is a big reason why we're not there yet, but these spells don't appear to be spells we'll reuse.

Alastair-Scott: This ability has actually been used by a few bosses to trigger a phase change. I don't think it's going to be used in many places so I can certainly move it.

src/interface/report/index.js
Alastair-Scott: I thought of adding it here so that the data is loaded once per fight selection so it shouldn't be called again when changing player. I could have that logic wrong though

src/interface/report/BossEventLoader.js
Alastair-Scott: What do you mean here? I'm trying to group all of the filters as best as possible to simplify the query logic for WCL.

src/raids/uldir/Ghuun.js
Alastair-Scott: I did originally, but when I came to implementing the micro phases, though having them as a string would allow for more customisation. This also works better for things like the chambers on Mother if we wanted to do those with id's of 'C1', 'C2' and 'C3'

At least that was my idea...

src/interface/report/BossEventLoader.js
MartijnHols: Doesn't it do `(type='removebuff' and ability.id=263504)`?

Alastair-Scott: Nope. I get all the different ability ids from the config that trigger phase changes, and get all events based with that id. I then filter those events during my normalizer at the moment. All the abilities that have been used to trigger events so far have very few events associated with them so I'm not having loads of useless events passed back

src/interface/report/index.js
MartijnHols: Most common use case is a player reviewing multiple fights. Could also cache the result in the redux store, but even then since it's cached server side I am not sure if there's any need.

src/common/SPELLS/encounter.js
MartijnHols: The way chunking will work is each relevant part will be separate, so for boss phasing it will probably end up being so each boss is its own chunk. If it's a boss specific thing, it helps to be put with the boss specific stuff. If it's shared, it might belong in its own file or something that's shared. If it's going to be used with SpellLink then spells might make sense and we'll have to figure out a solution later.

Boss specific abilities seems like something that can grow big and will be mostly exclusive to that boss, so splitting it from the start would be very nice.

src/raids/uldir/Ghuun.js
MartijnHols: Maybe then it should be an int. What is the id used for?

src/interface/report/BossEventLoader.js
MartijnHols: Is there any reason not to combine it? You could make a filter builder as simple as:

```jsx
export default function makeWclFilter(obj) {
  const filter = [];
  Object.keys(obj).forEach(key => {
    const value = obj[key];
    if (typeof value === 'number') {
      filter.push(`${key}=${value}`);
    } else {
      filter.push(`${key}='${value}'`);
    }
  });
  return filter.join(' AND ');
}

```

Alastair-Scott: I could certainly extract my function out, but I feel that my version does produce a better filter.

(ability.id IN (1, 2, 3, 4, 5, x)) 
OR 
(Custom_query_1) // used for add wave
OR
(Custom_query_2) // used for different add wave

src/interface/report/index.js
Alastair-Scott: Ok cool, I'll migrate the code to the EventParser

src/raids/uldir/Ghuun.js
Alastair-Scott: I only use it in one place at the moment, and that's for matching the end event of add waves to the start. It could probably be used in the future for filtering the data down too if you only want to parse events for a particular phase

src/interface/report/BossEventLoader.js
MartijnHols: Why is it more optimised? This would get several unnecessary events if the spell were used for other things.

src/raids/uldir/Ghuun.js
MartijnHols: Then it should probably be the key of an object instead

src/interface/report/BossEventLoader.js
Alastair-Scott: Instead of having ability.id = 1 or ability.id = 2 or ability.id = 3

PR #2924, opened 2019-01-07
[NEXT] Styling idea for the scrollbar on the timeline

Alastair-Scott:
Was just an idea I had to make the scroll bar look a little more fancy. 

Current:
image

Idea:
image

### DIFF
--- a/src/interface/report/Results/Timeline/Timeline.scss
+++ b/src/interface/report/Results/Timeline/Timeline.scss
@@ -1,19 +1,20 @@
 .spell-timeline-container {
   width: 100%;
   overflow: auto;
+  padding-bottom: 6px;
 
   &::-webkit-scrollbar {
-    width: 12px;
+    height: 12px;
   }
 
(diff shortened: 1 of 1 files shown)


Dambroda:
I like the idea of reskinning it for sure, but am not really a fan of how this looks. I think keeping everything solid is the way to go, rather than this glow/whatever you want to call it 

That's my opinion ofc, so looking to see what others say :)

Alastair-Scott:
Just having the borders coloured does make it match the icons fairly well and so on the full page it looks better. I do think I should remove the box shadows and just have it as solid colours though

Dambroda:
Yeah its just the glow I'm personally not a fan of

Alastair-Scott:
Scrollbar with no box shadow
image

sMteX:
How about filling it with the color (like the bars for GCD in the timeline)? I think that would fit more for the overall theme we're on.

Alastair-Scott:
With solid colours:
image

PR #2937, opened 2019-01-14
[NEXT] Rewriting Line charts to react-vis [PART 1]

sMteX:
It's a mess.. Point is to make it look the same as before (or as closest as it gets). 

### In this part
- Brewmaster Monk
    - Mastery Value chart
image

    - Stagger chart (stolen from `emallson` in `test` branch, customized)

image

- Elemental/Enhancement Shaman
    - Maelstrom chart

image

- Protection Paladin/Blood DK 
    - Self Heal chart (obviously pieced together from multiple screenshots)

image

### Notable changes
- reworked charts had CSS files made into SCSS (for mixins)
- Tooltip.css was also converted to SCSS
- `react-tooltip-lite` "styles" were added as mixins to Tooltip.scss
    - these are used in `react-vis` chart components like (custom) VerticalLine, Hint, Crosshair etc.
    - they're intended to reduce duplicity and get the same or almost same look as the rest of the tooltips
- in majority of the cases, chart legend is above the chart, but `react-vis` renders it defaultly below the chart
    - this is solved with `align-legend` mixin + inline positioning it on top and aligning to the middle (or sorta middle) while also giving the XYPlot larger top margin
    - it would be nice if it was somehow made simpler/more automatic
- EDIT: some charts are further extracted into separate files in #2943, if that's your concern during review 

### Missing (will be in next parts):
- Arcane Mage 
    - Mana Level chart
- Vengeance DH 
    - Pain chart
- Protection Paladin
    - Grand Crusader chart
- Shadow Priest
    - Voidform chart
- Hunters
    - Focus chart
- shared
    - raid health chart
    - Mana Level chart
    - Mana Usage chart
- Death Knights
    - Rune Details chart
- Brewmaster Monk
    - Straight, No Chaser chart

PR #2941, opened 2019-01-14
[NEXT] Rewriting Line charts to react-vis [PART 2]

sMteX:
## Depends on #2937, it should be merged first.

This also includes commits from #2937, that's why the diff is so large.

### In this part:
- Protection Paladin
    - Grand Crusader chart
image
- Brewmaster Monk
    - Straight, No Chaser chart
image
- Death Knights
    - Rune Details chart

image

- Vengeance DH 
    - Pain chart (the weird Pain wasted at the end is a matter of missing data (null values), I filled them in with last known value (same with boss health and Pain)

image

### Notable changes:
- I refactored Grand Crusader, Mastery Value and Straight, No Chaser charts into one single chart (since they're all nearly identical) to reduce code duplicity
- Improved Legend positioning - should now always be in the middle (XYPlot still needs `margin={{ top: 30 }}` because it's not aplied as a  `style` so it can't be made into CSS rule
- EDIT: some charts are further extracted into separate files in #2943, if that's your concern during review 

### Missing (will be in next parts):
- Arcane Mage 
    - Mana Level chart
- Shadow Priest
    - Voidform chart
- Hunters
    - Focus chart
- shared
    - raid health chart
    - Mana Level chart
    - Mana Usage chart


MartijnHols:
thanks

PR #2942, opened 2019-01-15
replaced item() by statistic()

Jackalx:
replaced item() by statistic()

### DIFF
--- a/src/parser/core/Analyzer.js
+++ b/src/parser/core/Analyzer.js
@@ -79,7 +79,7 @@ class Analyzer extends EventSubscriber {
   }
 
   // Override these with functions that return info about their rendering in the specific slots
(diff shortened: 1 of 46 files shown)


MartijnHols:
This was not what was intended at all. Did you test the change? I'm pretty sure this doesn't even render.

`statistic()` is a method that already exists and does not support the item result, nor should it. The goal is to convert all of the item objects to the format that is already required by statistic. [Here is an example commit.](https://github.com/WoWAnalyzer/WoWAnalyzer/commit/89473cee6b0f65f66bad696ce8ec56862e29f0b2#diff-a0070669e44d826e4825a77af12b7320)

I'm closing this PR since it is nowhere close to mergeable or what we need.

PR #2950, opened 2019-01-23
[Azerite] Adding Treacherous Covenant

TJMoats:
I wasn't sure what this trait should look like, so feel free to give me some feedback about styling and which metrics should be displayed.

ItemDamageTaken Might be to niche, but there may be more things that use it moving forward. I don't have a problem removing it if ya'll don't think it is needed.

screen shot 2019-01-23 at 10 17 28 am

The new icon is from https://thenounproject.com/search/?q=skull&i=1174134


src/interface/others/ItemDamageTaken.js
```diff
@@ -0,0 +1,30 @@
...
+  };
+
+  render() {
+    const { amount, approximate } = this.props;
+    const { parser } = this.context;
+
+    return (
+      <>
+        <img
+          src="/img/damage-taken.png"
```
MartijnHols: Please use the shield icon for damage taken as I reckon it has been used in that context previously.

src/parser/shared/modules/spells/bfa/azeritetraits/TreacherousCovenant.js
```diff
@@ -0,0 +1,97 @@
...
+  debuffActive = false;
+  extraDamageTaken = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.TREACHEROUS_COVENANT.id);
+    if (!this.active) {
+      return;
+    }
+    const { stat } = treacherousCovenantStat(this.selectedCombatant.traitsBySpellId[SPELLS.TREACHEROUS_COVENANT.id]);
```
MartijnHols: Please add this to the statTracker too. Search for `this.statTracker.add` in existing files for an example.

MartijnHols:
Please add a changelog entry too

PR #2956, opened 2019-01-27
[Sub Rogue] Improvements in Nightblade analysis

TsaboTavok:
Hi,

This PR contains several improvements to the Nightblade analysis, IMO the biggest gap in Subtlety right now.

1.  Check for early refreshes, account for the cases when it should be refreshed early
2. Shift focus from up time to effective damage increase. This is much more accurate in target switch and downtime situations.


src/parser/rogue/subtlety/modules/core/NightbladeEarlyRefresh.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+  on_byPlayer_spendresource(event) {
+    const comboPointsSpent = event.resourceChange;
+    if (event.resourceChangeType !== RESOURCE_TYPES.COMBO_POINTS.id) {
+      return;
+    }
+
+    //Update duration.
+    DOTS[0].duration = (comboPointsSpent * 2 + 6) * 1000;
```
MartijnHols: Please do not modify constants. You should probably work with a class property instead.

src/parser/rogue/subtlety/modules/features/checklist/Module.js
```diff
@@ -14,6 +14,7 @@ import AlwaysBeCasting from '../AlwaysBeCasting';
 import CastsInShadowDance from '../../core/CastsInShadowDance';
 import CastsInStealth from '../../core/CastsInStealth';
 import NightbladeDuringSymbols from '../../core/NightbladeDuringSymbols';
+import NightbladeEarlyRefresh from '../..//core/NightbladeEarlyRefresh';
```
MartijnHols:
```suggestion
import NightbladeEarlyRefresh from '../../core/NightbladeEarlyRefresh';
```

src/parser/rogue/subtlety/modules/core/NightbladeEarlyRefresh.js
MartijnHols: How about adding a getDot method to the core EalyDotRefreshes module that returns `this.constructor.dots.find(element => {
      return element.debuffId === event.ability.guid;
    });` so you can override that to provide custom values instead?

PR #2968, opened 2019-01-31
More mote functionality

HawkCorrigan:
Depends on #2965 cause I am bad at the gits.

Adds a table that displays what abilities were buffed by MotE-talent.

If anyone know a better way to go about the whole array stuff, I'd appreciate it.


HawkCorrigan:
image

PR #2974, opened 2019-02-02
Maelstrom tracker

HawkCorrigan:
So appearently there was no chart for general resources.

I think this should work for pretty much every class and every resource if done properly.
I am not entirely sure if the graph and the breakdown should be bundled tho.

Idk if this will cause any issues with the updated version to react-viz on next but I think that should be solvable.


HawkCorrigan:
image
image
Pictures for good measure

PR #2980, opened 2019-02-06
[Havoc DH] Level 100 talents

yajinni:
Created a log with the 3rd talent, but unfortunately due to its nature i have no way of calculating its fury generation.  

Talent in question:
**Insatiable Hunger:**  Demon's Bite generates up to 10 additional Fury.
As for **Demons Bite:** Quickly attack for (45.279% of Attack power) Physical damage. Generates 20 to 30 Fury.

Since they both generate variable amount of fury, and logs dont show the extra fury Insatiable hunger provides, i have no way of separating the 2 and calculating the extra it provides. A log if anyone wants to look into it:
https://www.warcraftlogs.com/reports/AZMDnzrG48KJLgP6/#fight=last&type=resources&view=events&spell=117

image
image

image
image

### DIFF
--- a/src/parser/demonhunter/havoc/modules/talents/DemonBlades.js
+++ b/src/parser/demonhunter/havoc/modules/talents/DemonBlades.js
@@ -67,8 +67,10 @@ class DemonBlades extends Analyzer{
       <TalentStatisticBox
         talent={SPELLS.DEMON_BLADES_TALENT.id}
         position={STATISTIC_ORDER.OPTIONAL(6)}
-        value={(<>{this.furyPerMin} fury per min <br />
-                {this.owner.formatItemDamageDone(this.damage)}</>)}
+        value={(<>
+                {this.furyPerMin} fury per min <br />
+                {this.owner.formatItemDamageDone(this.damage)}
+              </>)}
         tooltip={`
           ${formatThousands(this.damage)} Total damage<br />
           ${this.effectiveFuryGain} Effective fury gained<br />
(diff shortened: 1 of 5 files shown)


src/parser/demonhunter/havoc/modules/talents/ImmolationAura.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  statistic(){
+    this.effectiveFuryGain = this.furyGain - this.furyWaste;
+    return (
+      <TalentStatisticBox
+        talent={SPELLS.IMMOLATION_AURA_TALENT.id}
+        position={STATISTIC_ORDER.OPTIONAL(6)}
+        value={(<>{this.furyPerMin} fury per min <br />
+                {this.owner.formatItemDamageDone(this.damage)}</>)}
```
TJMoats: This looks a bit cluttered, I think it would read nicer like: 

```
value={(<>
    {this.furyPerMin} fury per min <br />
    {this.owner.formatItemDamageDone(this.damage)}
</>)}
```

src/parser/demonhunter/havoc/modules/talents/DemonBlades.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  statistic(){
+    this.effectiveFuryGain = this.furyGain - this.furyWaste;
+    return (
+      <TalentStatisticBox
+        talent={SPELLS.DEMON_BLADES_TALENT.id}
+        position={STATISTIC_ORDER.OPTIONAL(6)}
+        value={(<>{this.furyPerMin} fury per min <br />
+                {this.owner.formatItemDamageDone(this.damage)}</>)}
```
TJMoats: Same here, I think moving it so that you have breaks in code makes it far easier on the eyes.

PR #3003, opened 2019-02-20
Only apply own prepull buffs

HawkCorrigan:
```

BlazybYesterday at 11:49 AM
@Restoration Druid 
Lifebloom uptime shows 100% more than actual.
/report/naqbT14ywgHpWh7D/25-Mythic+Conclave+of+the+Chosen+-+Wipe+11+(4:51)/94-Syqu
/report/Ywx3Ky6h8bFHgAvX/17-+Freehold+-+Kill+(33:39)/1-Baconsaber
@Zerotorescue Could it be related to some changes you've done recently? I ran an older branch (~2 weeks) and it works as expected.
```


HawkCorrigan:
Prepull buffs from druid 2(Duskleaf) get applied to both Entities in the normalizers, so we get an entry in the buffHistory that cannot be matched to any removeBuff.

This issue gets masked by the fact that the changeBuffStack method doesnt check for sourceId which means that the 100% additional uptime gets split between all single segments of buffUptime.

The bug should only happen when calling player.getBuffUptime, so only for Livebloom, Earthshield, Efflo and Photo and only when two players are using it.

PR #3004, opened 2019-02-20
Fixing Lifebloom Fix?

HawkCorrigan:
This is still not a nice fix but it at least fixes the changebuffstack events.

I would like to completely get rid of the one rogue buff event in the buff history but for that I need more time to understand why it even exists.

Live:
grafik

Fixed:
grafik

Report:
http://localhost:3000/report/naqbT14ywgHpWh7D/25-Mythic+Conclave+of+the+Chosen+-+Wipe+11+(4:51)/94-Syqu/events

### DIFF
--- a/src/parser/shared/modules/Entities.js
+++ b/src/parser/shared/modules/Entities.js
@@ -97,7 +97,7 @@ class Entities extends Analyzer {
 
     debug && this.log(`Apply buff stack ${event.ability.name} to ${entity.name}`);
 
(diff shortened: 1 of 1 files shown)


MartijnHols:
Since this is such an edge case could you add a test that verifies the fix so it can't be reintroduced?

HawkCorrigan:
ye I'll add something over the weekend regarding tests

PR #3006, opened 2019-02-20
[WIP] Rework prepull buff normalizers

MartijnHols:
This adds a `CombatantInfoAuras` normalizer to add all prepull buffs to the `combatantinfo.auras` prop and rework `ApplyBuff` to just use that instead, greatly simplifying it.

This also fixes one of the rare bugs where if the same buff was applied to a target by two different sources, it would only fabricate 1 applybuff.


MartijnHols:
I think I'm going to move this to `next` instead. I want to rework normalizers so they're called before the other modules are initialized and fear it might be a risky move on master.

MartijnHols:
This still needs some more work, leaving it open for the time being

MartijnHols:
Closing this as this is on hold indefinitely

PR #3011, opened 2019-02-21
#2963 Fetch Heart of Azeroth item level from Blizzard API and display the value on Raid Composition page

### LINKED ISSUE #2963: NEXT: Fetch heart level from Blizz API and add it to raid composition page
image

The data is not yet available in the API data available. You should be able to get it with the `items` field from the Blizzard character API. We do not fetch that yet, you'll have to implement that. We don't care about most of the rest of the items field data, although maybe the character page uses it?

4Ply:
Resolves #2963 
The result looks like this:

Please note that this is my first contribution to this project, and so please forgive me if I have done something in the wrong place. If so - I'll happily fix it!

### DIFF
--- a/server/constants/index.js
+++ b/server/constants/index.js
@@ -1,0 +1,3 @@
+export const HEART_OF_AZEROTH = {
+  id: 158075,
+};
--- a/server/controllers/api/character.js
+++ b/server/controllers/api/character.js
@@ -2,6 +2,7 @@ import Express from 'express';
 import Sequelize from 'sequelize';
 import Raven from 'raven';
 import { StatusCodeError } from 'request-promise-native/errors';
+import { HEART_OF_AZEROTH } from '../../constants'
 
 import BlizzardCommunityApi from 'helpers/BlizzardCommunityApi';
 import RegionNotSupportedError from 'helpers/RegionNotSupportedError';
@@ -46,14 +47,14 @@ function getCharacterId(thumbnail) {
 }
 
 async function getCharacterFromBlizzardApi(region, realm, name) {
-  const response = await BlizzardCommunityApi.fetchCharacter(region, realm, name, 'talents');
+  const response = await BlizzardCommunityApi.fetchCharacter(region, realm, name, 'talents,items');
   const data = JSON.parse(response);
   if (!data || !data.thumbnail) {
     throw new Error('Corrupt response received');
   }
   // This is the only field that we need and isn't always otherwise obtainable (e.g. when this is fetched by character id)
   // eslint-disable-next-line prefer-const
-  const { talents, thumbnail, ...other } = data;
+  const {talents, thumbnail, items, ...other} = data;
   delete other.calcClass;
   delete other.totalHonorableKills;
   delete other.level;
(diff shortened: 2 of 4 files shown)


HawkCorrigan:
Would you mind posting the wcl log link as well? Code looks good to me but I am curious why the MW doesn't have a neck displayed. (there aren't really any people playing without the neck so I am curious if he is... weird or if it's a bug)

HawkCorrigan:
Oh and I just reread the issue.

I am pretty sure that Zer0 is referencing the neck level so the... azerite power up level and not the ilvl of the item itself.

4Ply:
@HawkCorrigan Thanks, I shall update this PR tonight.

/report/2Yjwczk8KRnyVvrt/6-Heroic+Champion+of+the+Light+-+Kill+(3:30)

Sorry, I believe "Vastus" does not have the neck item due to the caching of the character object, I shall upload an updated screenshot after flushing the cache anyway as I need to use the power level instead of the item level.

4Ply:
Updated to use `azeriteItem.azeriteLevel` instead of `tooltipParams.timewalkerLevel`:
@HawkCorrigan the item display issue was indeed a caching problem, well caught!

image

server/constants/index.js
```diff
@@ -0,0 +1,3 @@
+export const HEART_OF_AZEROTH = {
```
HawkCorrigan: IMO it would better fit here: https://github.com/WoWAnalyzer/WoWAnalyzer/tree/master/src/common/ITEMS/bfa

server/controllers/api/character.js
```diff
@@ -46,14 +47,14 @@ function getCharacterId(thumbnail) {
...
-  const response = await BlizzardCommunityApi.fetchCharacter(region, realm, name, 'talents');
+  const response = await BlizzardCommunityApi.fetchCharacter(region, realm, name, 'talents,items');
   const data = JSON.parse(response);
   if (!data || !data.thumbnail) {
     throw new Error('Corrupt response received');
   }
   // This is the only field that we need and isn't always otherwise obtainable (e.g. when this is fetched by character id)
   // eslint-disable-next-line prefer-const
-  const { talents, thumbnail, ...other } = data;
+  const {talents, thumbnail, items, ...other} = data;
```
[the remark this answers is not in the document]
4Ply: Auto-formatting, I will update my config to match this repo's style

server/constants/index.js
4Ply: Ah, I wasn't sure if /server was meant to remain localized to its own directory. 
Do you have a filename suggestion? I feel like the item is unique enough to not fall under any general category :/

4Ply: This will cause an import out of /server such as `import heartOfAzeroth from '../../../src/common/ITEMS/bfa/heartOfAzeroth';`, which will prevent server from being extracted out of this repository, are we sure this is something we want to do? 

I don't see any other instances of this happening in /server

HawkCorrigan: ye thinking about it, it does make sense to keep it here.
If Zer0 has anything against it, it should be easy enough to change anyway.

I'd say readd the whitespaces please and I'll merge it.

4Ply: Great, done

HawkCorrigan:
Thanks for the contribution. 👍

MartijnHols:
This doesn't appear to work: https://next.wowanalyzer.com/report/YcbxZv1hKX4GVr82/33-Mythic+Grong+-+Wipe+23+(4:26)/

4Ply:
@MartijnHols are those characters still cached? If so then the update won't display

PR #3012, opened 2019-02-22
Stormbringer

HawkCorrigan:
Dunno what to add here really.
Purely a spell logic change that adds the resets from Stormbringer.

### DIFF
--- a/src/parser/shaman/enhancement/modules/core/Stormbringer.js
+++ b/src/parser/shaman/enhancement/modules/core/Stormbringer.js
@@ -10,18 +10,25 @@ class Stormbringer extends Analyzer {
     spellUsable: SpellUsable,
   };
 
+
   constructor(...args){
     super(...args);
-    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.STORMBRINGER), this.onReset());
-    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.STORMSTRIKE), this.onNoCooldown());
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.STORMBRINGER_BUFF), this._onReset);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.STORMSTRIKE), this._onNoCooldown);
   }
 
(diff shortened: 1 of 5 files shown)


src/parser/shaman/enhancement/modules/core/Stormbringer.js
```diff
@@ -0,0 +1,32 @@
...
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.STORMBRINGER), this.onReset());
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.STORMSTRIKE), this.onNoCooldown());
+  }
+
+  onReset(event) {
+    this.spellUsable.endCooldown(SPELLS.STORMSTRIKE.id);
+  }
+
+  onNoCooldown() {
+    if(!this.selectedCombatant.hasBuff(SPELLS.STORMBRINGER.id)){
```
4Ply: You could simply this if statement by inverting it.
Also, I am not sure how strict the formatting standards are here, but this statement is missing some whitespaces :)

```
if (this.selectedCombatant.hasBuff(SPELLS.STORMBRINGER.id)) {
    this.spellUsable.endCooldown(SPELLS.STORMSTRIKE.id);
}

HawkCorrigan: Usually we prefer early returns over nested statements.
I mean here it is pretty trivial anyway but if the code starts getting more complex, you will run into issues with having it nested that way.

grafik
(Holy paladin is maintained by the boss so I usually refer to his stuff regarding code style.
Let's take that as reference. If you would nest everything here it would start getting kinda messy.

Regarding the whitespaces, what I have seen most people do is use whitespaces for comparing, objects but not for functions and such. Refer to the above screenshot for the same thing with the functions for the first if clause. 

And here you can see the whitespaces for the curly braces I am refering to.
grafik

HawkCorrigan:
Anyway looks like I broke something, gonna have to rewrite something real quick.

HawkCorrigan:
Okay I fixed it.
This should solve all cooldown issues with Stormbringer but the ones caused by Windstrike still prevail.
Gonna have to see how to solve this, together with the Metamorphosis stuff.

PR #3016, opened 2019-02-23
Character cache storage - persist heartOfAzeroth field in Sequelize cache

### LINKED ISSUE #2963: NEXT: Fetch heart level from Blizz API and add it to raid composition page

4Ply:
This resolves the issue where the updated Character data fetched the Blizzard API is not defined in the Character model, and thus does not display when the character is fetched from the cache.

Definition:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/next/server/controllers/api/character.js#L77

Resolves #2963

### DIFF
--- a/server/migrations/20190223014101-character.js
+++ b/server/migrations/20190223014101-character.js
@@ -1,0 +1,15 @@
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return Promise.all([
+      queryInterface.addColumn('Character', 'heartOfAzeroth', Sequelize.JSON, {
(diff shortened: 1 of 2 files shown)


4Ply:
Related to https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3011

PR #3023, opened 2019-02-26
[MW] JSS Uptime Suggestion / Checklist / Stats

anom0ly:
Added new information on JSS soothing mist uptime and corresponding statistics, checklist item, and suggestion..

image


yajinni:
if this is a talent shouldnt it use the talent box and not statistic box?

anom0ly:
Interesting - I honestly didn't know we had added that. Let me update not only this new talent statistic, but also the others, which I still using just the plain statisticbox component.

anom0ly:
Updated screenshot:
image

src/parser/monk/mistweaver/modules/talents/SpiritOfTheCrane.js
```diff
@@ -132,30 +132,25 @@ class SpiritOfTheCrane extends Analyzer {
...
-          >
-            Mana Returned
-          </dfn>
-        )}
+        label="Mana Returned"
+        tooltip={`
+          You gained a raw total of ${((this.manaReturnSotc + this.sotcWasted) / 1000).toFixed(0)}k mana from SotC with ${(this.sotcWasted / 1000).toFixed(0)}k wasted.<br>
+          You lost ${(this.totmOverCap + this.totmBuffWasted)} Teachings of the Monestery stacks
+          <ul>
+          ${this.totmOverCap > 0 ?
```
MartijnHols: For future reference the indentation here gets a bit weird, but I'll fix it when I merge this into `next`.

PR #3039, opened 2019-03-03
[Core] Hide empty rules in the checklist

Sharrq:
The old checklist would hide a rule if all of the requirements underneath it were hidden, but it doesnt look like the new checklist does that. This should fix that.

Before (Use GS and Use talents are empty)
image

After
image

### DIFF
--- a/src/parser/shared/modules/features/Checklist/Rule.js
+++ b/src/parser/shared/modules/features/Checklist/Rule.js
@@ -102,8 +102,11 @@ class Rule extends React.PureComponent {
 
     const performance = this.performance;
     const passed = this.passed;
+    const requirements = this.props.children;
+    console.log(this.props);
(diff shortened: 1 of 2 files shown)


src/parser/shared/modules/features/Checklist/Rule.js
```diff
@@ -94,6 +102,13 @@ class Rule extends React.PureComponent {
 
     const performance = this.performance;
     const passed = this.passed;
+    const requirements = this.props.children;
+    console.log(this.props);
+    console.log(requirements);
```
emallson: looks like you left some stray `console.log`s in. other than that it looks good

Sharrq: woops, should be removed now.

PR #3041, opened 2019-03-03
[Retribution Paladin] Wake of Ashes Module

LordSkeletor:
Statistic Box
-Track Average Targets Hit
-Track "Bad" casts that hit 0 targets

Suggestions
-Recommend 0 "Bad" casts with advice on how to avoid.


LordSkeletor:
Thanks. Wasn't aware of that.

PR #3045, opened 2019-03-08
[Feral] Gushing Lacerations azerite trait

Anatta336:
Adds tracking for the [Gushing Lacerations](https://www.wowhead.com/spell=278509/gushing-lacerations) azerite trait. Provides a simple statistic with damage increase and combo points generated (and wasted) from the trait.

gushinglacerations01

### DIFF
--- a/src/parser/druid/feral/modules/azeritetraits/GushingLacerations.js
+++ b/src/parser/druid/feral/modules/azeritetraits/GushingLacerations.js
@@ -1,0 +1,94 @@
+import React from 'react';
+
+import { formatNumber } from 'common/format';
+import SPELLS from 'common/SPELLS';
+import { calculateAzeriteEffects } from 'common/stats';
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
+import calculateBonusAzeriteDamage from 'parser/core/calculateBonusAzeriteDamage';
+import StatTracker from 'parser/shared/modules/StatTracker';
+import TraitStatisticBox, { STATISTIC_ORDER } from 'interface/others/TraitStatisticBox';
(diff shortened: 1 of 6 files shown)


src/parser/druid/feral/modules/azeritetraits/GushingLacerations.js
```diff
@@ -0,0 +1,94 @@
...
+  traitBonus = 0;
+  ripTickCoefficient = 0;
+  attackPower = 0;
+  totalDamage = 0;
+  comboProcs = 0;
+  wastedCombo = 0;
+
+  constructor(...args) {
+    super(...args);
+    if (!this.selectedCombatant.hasTrait(SPELLS.GUSHING_LACERATIONS_TRAIT.id)) {
```
sMteX: Unusual setting of the `active` flag but last time I checked, the default was indeed `true` so this is ok. Just caught my attention because usually it's 
```javascript
this.active = ...;
if (!this.active) {
   return;
}
```

src/parser/druid/feral/modules/azeritetraits/GushingLacerations.js
```diff
@@ -0,0 +1,94 @@
...
+
+  statistic() {
+    return (
+      <TraitStatisticBox
+        position={STATISTIC_ORDER.OPTIONAL()}
+        trait={SPELLS.GUSHING_LACERATIONS_TRAIT.id}
+        value={(
+          <>
+            <ItemDamageDone amount={this.totalDamage} /><br />
+            {((this.comboProcs - this.wastedCombo) / (this.owner.fightDuration / 1000 / 60)).toFixed(1)} combo points per minute
```
sMteX: This would be nicer to have as a getter instead, something like `effectiveComboPointsPerMinute`, your choice.

Anatta336:
Thank you! Made the suggested changes.

Issue #2199, opened 2018-09-15
Holy Paladin: Azerite Trait: Grace of the Justicar


MartijnHols:
Since it beacon transfers it's not as simple as WCL's indicator

5chmidt:
I can start working on this item.
Here is an example log with the trait equipped:
https://www.warcraftlogs.com/reports/9Gh6XMafg7TCFdpj#fight=4&type=healing&source=7&ability=278785&view=events

PR #3050, opened 2019-03-10
Holy Paladin: Azerite Trait: Grace of the Justicar

5chmidt:
Added azurite power statistic for Grace of the Justicar which also tracks the healing transfer from Grace of the Justicar to beacon targets.  The tooltip reports back how many target were hit per use out of 10.  #2199
image


src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,69 @@
...
+import TraitStatisticBox, { STATISTIC_ORDER } from 'interface/others/TraitStatisticBox';
+import ItemHealingDone from 'interface/others/ItemHealingDone';
+
+/**
+ * Grace of the Justicar
+ * Judging a foe heals up to 10 allies within 8 yards of that enemy for 899. (@iLvl 400)
+ * Example Log: https://www.warcraftlogs.com/reports/kMbVanmJwCg7WrAz#fight=last&type=summary&source=2
+ */
+class GraceOfTheJusticar extends Analyzer {
+  graceHealing = 0;
```
MartijnHols:
```suggestion
  healing = 0;
```

The module is already called GraceOfTheJusticar

src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,69 @@
...
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    if (spellId === SPELLS.GRACE_OF_THE_JUSTICAR.id) {
+      this.graceHealing += event.amount + (event.absorbed || 0);
+      this.targetsHit += 1;
+      return;
+    }
+    if (spellId === SPELLS.BEACON_OF_LIGHT_HEAL.id && this.justJudged) {
+      this.beaconTransfer += event.amount + (event.absorbed || 0);
```
MartijnHols: This is incorrect as events are not exactly sequential. See the `beacontransfer` event for the correct way, there are several modules using it.

src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,69 @@
...
+  beaconTransfer = 0;
+  casts = 0;
+  justJudged = false;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.GRACE_OF_THE_JUSTICAR_TRAIT.id);
+  }
+
+  on_byPlayer_cast(event) {
```
MartijnHols: Have you considered using the manual event listening through `addEventListener`? This is sort of the preferred way (though the legacy method isn't deprecated yet).

src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,69 @@
...
+
+  statistic() {
+    return (
+      <TraitStatisticBox
+        position={STATISTIC_ORDER.OPTIONAL()}
+        trait={SPELLS.GRACE_OF_THE_JUSTICAR.id}
+        value={<ItemHealingDone amount={this.totalHealing} />}
+        tooltip={`
+          Total healing done: ${formatNumber(this.totalHealing)}<br />
+          Players hit per cast: ${this.playersHitPerCast.toFixed(2)}
```
MartijnHols: This seems interesting enough, how about adding it as a part of the value?

src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
5chmidt: I have converted everything but the on_beacontransfer event.  I am attempting to add the event like this:
this.addEventListener(Events.beaconTransfer.by(SELECTED_PLAYER).source(SPELLS.GRACE_OF_THE_JUSTICAR), this.onBeaconTransfer);
But that is not correct.  I did a search and I don't see a beacontransfer listener being used.
Can you point me in the right direction?

HawkCorrigan: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/106a68c71a4f1c0460b931348e270eed175060a8/src/parser/core/Events.js

5chmidt: @HawkCorrigan thanks.  I found that also.  I guess I would have to add a filter for beacontransfer events?

5chmidt: I tried to add a new Event like this:

`  get beacontransfer() {
    return new EventFilter('beacontransfer');
  },
`

but was unsuccessful.

HawkCorrigan: do you have a log with the trait at hand?

5chmidt: https://www.warcraftlogs.com/reports/kMbVanmJwCg7WrAz#fight=last&type=summary&source=2

HawkCorrigan: Just tested it real quick

Technically adding the EventFilter works.
You propably went wrong when adding the eventListener
The beacontransfer still has Beacon as it's spell instead of Grace which means you can't filter it how you wrote it above.
Additionally it would need to be .spell instead of .source

grafik

HawkCorrigan: so adding this:
```
  get beacontransfer() {
    return new EventFilter('beacontransfer');
  },
```
enables you to do 
```
this.addEventListener(Events.beacontransfer.by(SELECTED_PLAYER), this.onBeaconTransfer);
```

HawkCorrigan:
Why is this whole thing in the shared section tho?

src/parser/shared/modules/spells/bfa/azeritetraits/GraceOfTheJusticar.js
5chmidt: @HawkCorrigan Thanks, that did the trick.

5chmidt:
> Why is this whole thing in the shared section tho?

I put it there because that is where I found the other azurite traits.
I can move it to a class specific folder.

5chmidt:
I have moved it into a paladin folder.
It should be okay to merge now.

5chmidt:
Hmmmm, did not pass the build tests after I moved everything to the paladin folder.
Is there something else I need to do to get this to integrate?

src/parser/core/Events.js
```diff
@@ -90,6 +90,13 @@ const Events = {
    * BEWARE: Blizzard also sometimes uses this event type for mechanics and spell ticks or bolts. This can even occur in between a begincast and cast finish!
    * @returns {EventFilter}
    */
+  get beacontransfer() {
+    return new EventFilter('beacontransfer');
+  },
+  /**
```
sMteX: This jsdoc should be *before* the `get beacontransfer() {`, and the one above `beacontransfer` should be here

src/parser/paladin/shared/spells/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,77 @@
...
+      <TraitStatisticBox
+        position={STATISTIC_ORDER.OPTIONAL()}
+        trait={SPELLS.GRACE_OF_THE_JUSTICAR.id}
+        value={(
+          <>
+            <ItemHealingDone amount={this.totalHealing} /><br />
+            {this.playersHitPerCast.toFixed(1)} players hit per judgement.
+          </>
+        )}
+        tooltip={`
```
sMteX: Since today we merged `next` to `master` branch, the tooltip library changed and it no longer allows for HTML tags in tooltips (in interpolated strings, that is). In order to preserve the style, the `tooltip` property must be changed to:
```javascript
tooltip={(
  <>
    Total healing done: <b>{formatNumber(this.totalHealing)}</b><br />
    Beacon healing transfered: <b>{formatNumber(this.beaconTransfer)}</b>
  </>
)}
```

src/parser/core/Events.js
```diff
@@ -78,11 +78,18 @@ const Events = {
   get absorbed() {
     return new EventFilter('absorbed');
   },
-  /**
+   /**
+   * This event is called when a beacon transfer occurs.
+   * @returns {EventFilter}}
+   */
+  get beacontransfer() {
```
MartijnHols: Please don’t add spec specific event types to the Events object. Instead make it a static prop on the class fabricating the event. I don’t remember what other analyzer has that already, but there are examples in the codebase

MartijnHols: I finally had a moment to look for the example, see 

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/9100029083007fb7cfaa1b5375f58fb777f1714b/src/parser/paladin/holy/modules/beacons/BeaconHealSource.js#L32-L34

MartijnHols: Took me a moment to realize that actually the event you want, so it's already there, you just got to use it. heh.

5chmidt: I am importing BeHealSource
`import BeaconHealSource from '../../../holy/modules/beacons/BeaconHealingDone.js';`
Adding the dependencies:
```
  static dependencies = {
    beaconHealSource: BeaconHealSource,
  };
```
and adding a listener for the event:
`this.addEventListener(this.beaconHealSource.beacontransfer.by(SELECTED_PLAYER), this.onBeaconTransfer);`
But I am getting a type error when I compile attempt to add the listener.
Unhandled Rejection (TypeError): Cannot read property 'by' of undefined

I think I am using it is the same was as the example @MartijnHols  gave.
Do I need to add another new EventFilter for beacontransfer?

sMteX: AFAIK, `BeaconHealSource` doesn't need to be added as a dependency (if you only use it for the event filter). Import the class, don't add as a dependency, and then use this instead:

```javascript
this.addEventListener(BeaconHealSource.beacontransfer.by(SELECTED_PLAYER), this.onBeaconTransfer);
```

`beacontransfer` is a static getter, bound to the class, not an instance of it.

5chmidt: I got rid of the dependency and changed to listen for: `BeaconHealSource.beacontransfer.by(SELECTED_PLAYER)`
But I am still getting the same error where it does not like the `.by(SELECTED_PLAYER)` method.

src/parser/paladin/shared/spells/azeritetraits/GraceOfTheJusticar.js
```diff
@@ -0,0 +1,85 @@
...
+
+
+/**
+ * Grace of the Justicar
+ * Judging a foe heals up to 10 allies within 8 yards of that enemy for 899. (@iLvl 400)
+ * Example Log: https://www.warcraftlogs.com/reports/kMbVanmJwCg7WrAz#fight=last&type=summary&source=2
+ */
+class GraceOfTheJusticar extends Analyzer {
+  static dependencies = {
+    beaconHealSource: BeaconHealSource,
```
MartijnHols: Seeing as this depends on a Holy module, unless you want to make this specific to the Holy implementation and want to support this trait for Ret/Prot, this should be in the `holy` spec folder instead of shared. You can make an `azeritetraits` folder in Holy if you want.

src/parser/core/Events.js
MartijnHols: `this.beaconHealSource.beacontransfer.by(SELECTED_PLAYER)` should work since it's a non-static class property. It is used by BeaconHealingDone too.

src/parser/paladin/shared/spells/azeritetraits/GraceOfTheJusticar.js
5chmidt: I tested that and  trait still provides healing when you swap out of holy, so I put it in shared.  My thought being, the non-beacon specific events will still fire and let users know how much healing the trait is doing.

MartijnHols: BeaconHealSource does not exist on the other specs (and should not exist), so if you want to do that you'd have to make that logic specific to the Holy module. You can extend the module if that's what you want, same way as we do for several core modules.

5chmidt: Okay, that makes sense, I'll change it to the holy folder.
The crazy relative path needed to import the beacon event was my first clue that things were not as they should be.

src/parser/core/Events.js
5chmidt: Thanks, I'll give it a go.  Hopefully I can get this PR finished tonight!

5chmidt:
I have moved the module to holy and the event listener is working correctly on my local machine.  I ran a few tests and double checked my math, everything looks to good.  But I am still getting a build failure.
When I look at the details it is telling me that the snapshot did not match for paladin/holy/CombatLogParser.test.js.  
The previous version and snapshot did not contain the Grace of the Justicar trait so they shouldn't match.  I did a bit of reading and I believe I need to update the snapshot to contain the new trait and add it to my PR.  Is this the correct way to proceed?

MartijnHols:
Indeed you need to update the snapshot. That way the diff shows everything that was changed and we can verify it wasn't anything unintended.

Run: `npm test holy/CombatLogParser` and then press `u` in the command window when it says the snapshots need updating (`npm test holy/CombatLogParser -- -u` might work, not sure).

src/CHANGELOG.js
```diff
@@ -7,6 +7,11 @@ import SpellLink from 'common/SpellLink';
 import Contributor from 'interface/contributor/Button';
 
 export default [
+  {
```
MartijnHols: This should be in the Holy Paladin changelog rather than in core

PR #3068, opened 2019-03-15
[MW] Added RJW Efficiency to Spreadsheet Tab

anom0ly:
Simple change to add RJW to spreadsheet tab for Mistweaver.

Also, added spell data for Bonded Souls and Mirror.


src/parser/monk/mistweaver/modules/talents/RefreshingJadeWind.js
```diff
@@ -44,6 +44,10 @@ class RefreshingJadeWind extends Analyzer {
     return (this.healsRJW / this.castRJW) / TARGETSPERCAST || 0;
   }
 
+  get rjwEffectiveness() {
+    return (this.healsRJW / (this.castRJW * TARGETSPERCAST)).toFixed(4) || 0;
```
sMteX: Not sure if the `|| 0` would evaluate correctly. I suggest splitting it up:
```javascript
const ret = (this.healsRJW / (this.castRJW * TARGETSPERCAST)) || 0; // for the lack of better variable name
return ret.toFixed(4);
```

anom0ly: Sound good. Just pushed the change.

With a better variable name. ;)

PR #3069, opened 2019-03-15
[Windwalker] Updating Cast Efficiency and some other minor fixes

Juko8:
- Added a bunch of utility spells to Cast Efficiency
- Reorganized a lot of spells in Cast Efficiency, the git overview looks really weird because of it
- Fixed a few typos
- Updated Mastery: Combo Breaker module suggestion thresholds and tooltip
- Fixed SCK incorrectly being marked as a bad cast when the player had a "Dance of Chi-Ji" proc


src/parser/monk/windwalker/modules/resources/EnergyCapTracker.js
```diff
@@ -76,7 +76,7 @@ class EnergyCapTracker extends RegenResourceCapTracker {
             </Tooltip>
           </div>
         )}
-        footerStyle={{ overflow: 'hidden' }}
+        footerstyle={{ overflow: 'hidden' }}
```
MartijnHols: This prop doesn't exist anymore and should be removed

PR #3071, opened 2019-03-17
[Windwalker] Updated Energizing Elixir handling

Juko8:
- Removed Energizing Elixir module. The energy is now added as regen and it'd be very tricky to accurately determine how much of the energy from talent was a gain or waste, which was all the module showed before
- Updated EnergyCapTracker to add the EE regen. There are a few prediction errors happening, but it seems to be because of the order of events and there should still be an accurate estimate of the time spent at cap.


src/parser/monk/windwalker/CHANGELOG.js
```diff
@@ -5,6 +5,11 @@ import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 
 export default [
+  {
+    date: new Date('2019-01-17'),
```
Sharrq: Its March not January :)

PR #3072, opened 2019-03-18
[Affliction] UI upgrade

sMteX:
Diff is enormous but to be fair, it's like 80% snapshot 😄 

Revamped the Affliction stat boxes to new format. In the process, I've tried adding a debuff variant of `getBuffHistory` to Enemies (also a combined version, for where multiple debuff IDs are used for single "effect", like Unstable Affliction has actually 5 different debuffs). It's used in UptimeBar

## Before: 
image
image

## After:
image
image


src/parser/shared/modules/Enemies.js
```diff
@@ -29,6 +29,51 @@ class Enemies extends Entities {
     return enemy;
   }
 
+  getCombinedDebuffHistory(...spellIds) {
```
emallson: So, I haven't really used the `getXHistory` family of methods for the stuff I maintain. What is this supposed to do?

sMteX: The original `getBuffHistory` in I think `Entity.js` returns an array of objects corresponding to certain buffId which contains useful info about it - start, end of the buff (most importantly), max stacks of it etc. Mostly `start, end, stacks` is used throughout the code though.

Since when talking about debuffs, we usually want debuff uptimes *across enemies*, this is why I had to make `getDebuffHistory`, it works pretty much identical to the part used in `getBuffUptime` - for each enemy, gets buff history for the given debuff and pushes its start and end event into some kind of array. Then it sorts the array by time and "combines" overlapping buffs into one single array of objects `{ start, end }`.

`getCombinedDebuffHistory` does the exact same, it just gets debuff history of multiple spells and treats them as "single spell". Perfect example of it is Unstable Affliction - it's 1 spell, but it actually applies up to 5 different debuffs with different IDs.

emallson: 👍

sMteX:
Finally added the docs comment and fixed relative import

PR #3079, opened 2019-03-20
MotE

HawkCorrigan:
Fixing the logic by removing everything that depends on Blizz and just tracking everything manually cause the events are basically just in a random order at this point.

### DIFF
--- a/src/parser/shaman/elemental/modules/talents/MasterOfTheElements.js
+++ b/src/parser/shaman/elemental/modules/talents/MasterOfTheElements.js
@@ -7,31 +7,34 @@ import SpellIcon from 'common/SpellIcon';
 import SPELLS from 'common/SPELLS';
 import calculateEffectiveDamage from 'parser/core/calculateEffectiveDamage';
 import SpellLink from 'common/SpellLink';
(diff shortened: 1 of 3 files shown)


src/parser/shaman/elemental/modules/talents/MasterOfTheElements.js
```diff
@@ -102,16 +97,45 @@ class MasterOfTheElements extends Analyzer {
...
+          .icon(SPELLS.MASTER_OF_THE_ELEMENTS_TALENT.icon)
+          .staticImportance(SUGGESTION_IMPORTANCE.MAJOR);
+      });
+  }
 
+  statistic() {
+    let value = `${formatPercentage(this.damagePercent)} %`;
+    let label = `Contributed ${formatNumber(this.damagePerSecond)} DPS (${formatNumber(this.damageGained)} total damage, excluding EQ).`;
+    if (this.isBugged) {
+      value = `BUGGED`;
```
MartijnHols: You should elaborate this state with a tooltip

HawkCorrigan:
I think that should be enough for a basic explanation now.

PR #3080, opened 2019-03-20
[Windwalker] Chi and Azerite statistics update

Juko8:
- Added a `<BoringResourceValue>` tag
- Updated statistics for Chi waste, Fury of Xuen and Glory of the Rising dawn

Before: 
billede
billede
billede

After: 
billede
billede

More updates to statistics will come in a later PR to keep this one small


src/parser/monk/windwalker/modules/resources/ChiDetails.js
```diff
@@ -18,6 +20,10 @@ class ChiDetails extends Analyzer {
     return this.chiTracker.wasted;
   }
 
+  get chiWastedPercent() {
+    return this.chiWasted / (this.chiWasted + this.chiTracker.generated);
```
HawkCorrigan: very unlikely but it might crash
please add a default value

```suggestion
return (this.chiWasted / (this.chiWasted + this.chiTracker.generated)) || 0
```

src/interface/statistics/components/BoringResourceValue/BoringResourceValue.scss
```diff
@@ -0,0 +1,26 @@
...
+    line-height: 1;
+
+    > div:first-of-type {
+      margin: 12px 0 4px 0;
+    }
+    > small {
+      display: block;
+    }
+  }
+}
```
HawkCorrigan: Not entirely sure how to handle this tbh.
To me it is completely redundant with the scss for boring-spell-value.

I would suggest to rename either file to BoringValue.scss and use that for both.
This wouldn't be anything to block merging imo.

PR #3082, opened 2019-03-21
[MW] Doughnut Update

abelito75:
Updated the other files to track their ef double splash healing and modified the donut chart to count it. From the logs I have tested we are nearly perfect. 

There appears to be some bad logic in EssenceFontMastery as it is counting the first and second gusts that procs off of it as its own healing. As well it seems we have set the suggestion thresh holds a bit too low depending if they take upwelling or not since upwelling doubles the amount of time they have and getting 2 casts off in 8 seconds seems like a given.

### DIFF
--- a/src/parser/monk/mistweaver/CHANGELOG.js
+++ b/src/parser/monk/mistweaver/CHANGELOG.js
@@ -7,7 +7,7 @@ import SpellLink from 'common/SpellLink';
 export default [
   {
     date: new Date('2019-03-21'),
-    changes: 'Made mastery track more accurately and updated the donut chart to reflect that.',
+    changes: 'Updated mastery tracking to more accurately reflect the spell that triggered the <SpellLink id={SPELLS.GUST_OF_MIST.id} />. This includes updates to the statistic and healing efficiency sections.',
     contributors: [Abelito75],
   },
   {
--- a/src/parser/monk/mistweaver/CHANGELOG.js
+++ b/src/parser/monk/mistweaver/CHANGELOG.js
@@ -6,6 +6,11 @@ import SpellLink from 'common/SpellLink';
 
 export default [
   {
+    date: new Date('2019-03-21'),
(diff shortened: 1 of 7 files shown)


anom0ly:
So thinking about this more broadly across the Mistweaver module, where does Gust of Mists get attributed?

I think in the context, the initial heal should get the Gust attributed to that spell. Ex. Casting Vivify and procing Gust attributes the healing to Vivify. If however the target has an Essence Font buff on them, then the first Gust proc should go to the spell that procc'ed it, like above. However, the second gust proc should be attributed to Essence Font. 

The `EssenceFontMastery` module already tracks the difference between the two Gust heal. Could the logic there be changed to something like the below:
```js
on_byPlayer_heal(event) {

    const targetId = event.targetID;
    if (spellId === SPELLS.GUSTS_OF_MISTS.id) {
      if (!this.combatants.players[targetId]) {
        return;
      }
      if (this.combatants.players[targetId].hasBuff(SPELLS.ESSENCE_FONT_BUFF.id, event.timestamp, 0, 0) === true && !this.gustHeal) {
        debug && console.log(`First Gust Heal: Player ID: ${event.targetID}  Timestamp: ${event.timestamp}`);
        this.healEF += 1;
        this.gustHeal = true;
      } else if (this.combatants.players[targetId].hasBuff(SPELLS.ESSENCE_FONT_BUFF.id, event.timestamp, 0, 0) === true && this.gustHeal) {
        this.healEF += 1;
        this.healing += (event.amount || 0) + (event.absorbed || 0);
        this.gustHeal = false;
      }
    }
  }```

This would actually update the chart, and also the Healing Efficiency table. Let me know what you think.

abelito75:
Thats a good call. I just didn't know what it all touched and after looking it seemed to be in some important places so I didn't want to touch something that could have a cascading effect. I felt like something like this could happened but since I'm still fairly new I just didn't want to break other things without realizing it. I'll change it and see if anything breaks. Nothing should

PR #3083, opened 2019-03-21
Fix prepull normalizer for demonology warlock

gwelican:
Some pet summons instantly trigger a cast(demonic consumption for example), these events are pushed before the actual summon, so the normalizer created fabricated summons, as they were already present when the boss was pulled.
My change will remove that extra summons.

Here is how it looks currently: https://www.wowanalyzer.com/report/XnMK8R3wq1NTQd7Y/22-Heroic+Mekkatorque+-+Kill+(5:23)/Gwelican/pet-timeline

### DIFF
--- a/src/parser/warlock/demonology/modules/pets/normalizers/PrepullPetNormalizer.js
+++ b/src/parser/warlock/demonology/modules/pets/normalizers/PrepullPetNormalizer.js
@@ -34,7 +34,7 @@ class PrepullPetNormalizer extends EventsNormalizer {
       }
       debug && console.log(`(${this.owner.formatTimestamp(event.timestamp, 3)}) Event`, event);
       if (event.type === 'summon' && event.ability && PET_SUMMON_ABILITY_IDS.includes(event.ability.guid)) {
-        this.checkForPetsWhichSummonWasLaterThanItsFirstCast(event, fabricatedEvents)
+        this.checkForPetsWhichSummonWasLaterThanItsFirstCast(event, fabricatedEvents);
         summonedPets.push(encodeTargetString(event.targetID, event.targetInstance));
         debug && console.log(`(${this.owner.formatTimestamp(event.timestamp, 3)}) Pet summon, added to array. Current array: `, JSON.parse(JSON.stringify(summonedPets)));
       }
--- a/src/parser/warlock/demonology/modules/pets/normalizers/PrepullPetNormalizer.js
+++ b/src/parser/warlock/demonology/modules/pets/normalizers/PrepullPetNormalizer.js
@@ -18,6 +18,9 @@ class PrepullPetNormalizer extends EventsNormalizer {
   // This normalizer looks at first 30 seconds (because that's hypothetically the longest any temporary pet can live, given a 15 second duration and 15 second extension via Demonic Tyrant, realistically lower because of GCD and cast time of DT)
   // And if it finds begincast, cast or damage events from a pet that isn't summoned yet, fabricates a summon event for them
 
+  // Sometimes pet's first cast happens before it's summon event(for example demonic consumption), both the event and the summon happens at the same time
+  // checkForPetsWhichSummonWasLaterThanItsFirstCast will check if they both have the same time, then it deletes the fabricated event
+
   normalize(events) {
     debug && console.log('playerPets', this.owner.playerPets.sort((pet1, pet2) => pet1.id - pet2.id));
     const maxTimestamp = this.owner.fight.start_time + MAX_TEMPORARY_PET_DURATION;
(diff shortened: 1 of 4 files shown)


src/parser/warlock/demonology/modules/pets/normalizers/PrepullPetNormalizer.js
```diff
@@ -86,6 +91,15 @@ class PrepullPetNormalizer extends EventsNormalizer {
     events.unshift(...fabricatedEvents);
     return events;
   }
+  checkForPetsWhichSummonWasLaterThanItsFirstCast(event, fabricatedEvents) {
+    // delete fabricated event, the pet immediately started casting something and events were pushed before the summon
+    for (let k = 0; k < fabricatedEvents.length; k++) {
+      if (fabricatedEvents[k].__originalTimeStamp === event.timestamp) {
```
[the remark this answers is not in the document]
gwelican: so what if we completely ignore the timestamp, and just work with pet guid?
My understanding is that normalizer will get the first 30 seconds of the fight.

I suppose petguid is unique, so if a pet does dmg with guid:1 then you summon a pet, which has guid:1, then we can safely remove thet pet, right?

gwelican:
@sMteX can you please review again? I took your advice and wrote a normalizer, that will pull summon events before demon spellcasting. 
So there will be no fabricated event in the first place.

This solution also helps my other PR about calculating demonic consumption power on tyrant.

src/parser/warlock/demonology/modules/pets/normalizers/SummonOrderNormalizer.js
```diff
@@ -0,0 +1,68 @@
...
+
+  // helper
+  isDogSummon(event) {
+    return event.type === 'summon' && (event.ability.guid === SPELLS.DREADSTALKER_SUMMON_1.id || event.ability.guid === SPELLS.DREADSTALKER_SUMMON_2.id);
+  }
+
+  swapEvents(_events, previousEventIndex, previousEvent) {
+    _events.splice(previousEventIndex, 1);
+    _events.push(previousEvent);
+    _events.__modified = true;
```
sMteX: You're changing the property on the array. This should probably be 
```suggestion
    previousEvent.__modified = true;
```

src/parser/warlock/demonology/modules/pets/normalizers/SummonOrderNormalizer.js
```diff
@@ -0,0 +1,68 @@
...
+const MS_BUFFER = 100;
+
+class SummonOrderNormalizer extends EventsNormalizer {
+
+  normalize(events) {
+    const _events = [];
+
+    events.forEach((event, idx) => {
+      _events.push(event);
+
```
sMteX: Seeing as both conditionals do the same (check for some condition, then iterate backwards from the event up to MS_BUFFER ms, then if some other condition is true, swap the events), you might be able to combine them together. I think this could work:

```javascript
// make a new helper
if (this.isDogSummon(event) || this.isTyrantSummon(event)) {
   const castTimestamp = event.timestamp;

   for (let previousEventIndex = idx; previousEventIndex >= 0; previousEventIndex -= 1) {
      const previousEvent = _events[previousEventIndex];
      if ((castTimestamp - previousEvent.timestamp) > MS_BUFFER) {
         break;
      }
      if (this.isSharpenedDreadfangs(previousEvent, event) || 
          this.isDemonicConsumption(previousEvent, event) || 
          this.isDemonFire(previousEvent, event)) {
        this.swapEvents(_events, previousEventIndex, previousEvent);
        break;
      }
   }
}
```

Now whether that is a *good idea* to join it together, is a different question. I think it's fine to leave it like you already have, I think it might be more readable.

src/parser/warlock/demonology/modules/pets/normalizers/SummonOrderNormalizer.js
gwelican: you are right. updated.

src/parser/warlock/demonology/modules/pets/normalizers/SummonOrderNormalizer.js
gwelican: I agree it is more readable this way, I found a sample code in resto normalizer, maybe there should be a more general solution which could be reused?

anyway, updated the PR

gwelican: About other pets, I did not see an issue with other pets so far, to limit the scope of this PR, I would keep it this way, until there is a bug with other pets as well.

sMteX:
Thanks for putting up with my comments! ^^

PR #3087, opened 2019-03-26
[Hunter] add Focused Fire Azerite Trait to Marksmanship

Pewtro:
image

Log: https://www.warcraftlogs.com/reports/47LJvZ9BgdhR8TXf#fight=43&type=summary&source=16


yajinni:
is there a better way to show the effective focused gained/total focus gained part?

Pewtro:
I don’t know, can it? 
I think it fits the rest of the hunter modules that show similiar things like that which is 
Gain/PossibleGain - but if you have something in mind I’d be happy to look at it

PR #3101, opened 2019-04-05
April fools reverts for bdk-vdh-havoc

yajinni:


HawkCorrigan:
I'm gonna take it that it works

PR #3104, opened 2019-04-12
Questionably Epic Tool Link - Healer Stat Weights

anom0ly:
This PR is to create a Stat weight link out to the Questionably Epic tool for gear, azerite trait, and other comparisons. [Link](https://questionablyepic.com/live/)

The tool offers healers the ability to compare azerite items, trinkets, and in the future overall gear. This link would provide that tool the stat weights computed by our StatWeight module. It is initially defaulted to being hidden, but any healing module can quickly enable it with a simple toggle.

Thanks.

Forgot to attach a screenshot of the change.

image

Link it generates: https://www.questionablyepic.com/live?name=Anom&realm=Mal%27Ganis&region=us&CriticalStrike=1.25&Haste(HPCT)=0.59&Haste(HPM)=0.47&Mastery=0.87&Versatility=1.16&Leech=1.80


anom0ly:
Just a note, Voulk (QE Live Dev) is working on the updates to that tool to leverage the import, but its not ready just yet. I wanted to get our side ready though, so turning on the integration is a quick PR process. Thanks.

src/parser/shared/modules/features/BaseHealerStatValues.js
```diff
@@ -416,6 +419,7 @@ class BaseHealerStatValues extends Analyzer {
   moreInformationLink = null;
   statistic() {
     const results = this._prepareResults();
+    this.url = '';
```
MartijnHols: I think this should be a local variable instead of a class variable, since it's only used here.

anom0ly: So I wasn't too sure on the different between local and class variables. I think its just the naming convention, but if I am wrong please let me know. I just pushed the update that I believe covers this point.

Thanks.

MartijnHols: I meant that it should be a variable scoped within the method rather than as a class property, so it's not used anywhere else.

anom0ly: Gotcha. Hopefully good to go now.

PR #3112, opened 2019-04-22
Added Trident of Deep Ocean to Item Statistics

Zeboot:
Added the Tank polearm "Trident of Deep Ocean" (chance to proc an absorption shield that, upon expiration, grants a stack of primary stat, up to 3 stacks)

Example Log: https://www.warcraftlogs.com/reports/6xHLtAFW4yC73mRD/#fight=28&source=2

Basic Box:
image

Tooltip:
image


src/parser/shared/modules/items/bfa/raids/crucibleofstorms/TridentOfDeepOcean.js
```diff
@@ -0,0 +1,192 @@
...
+    }, {});
+
+    Object.values(buffStacks).forEach((stackChain) => {
+      stackChain.forEach((stack) => {
+        const stackSize = stack.stacks;
+        const stackStart = stack.timestamp;
+        if (cleanStacks[lastHandledStack]) {
+          const previousStack = cleanStacks[lastHandledStack];
+          const lastOccurrence = previousStack[previousStack.length - 1];
+          if (this.isBuggedStackChange(lastOccurrence.start, stackStart)){
```
emallson: what exactly does "bugged" mean here? Looking at the definition of `isBuggedStackChange` it appears you're looking for stack changes that within a 100ms window? A comment explaining what this condition is would be useful

src/parser/shared/modules/items/bfa/raids/crucibleofstorms/TridentOfDeepOcean.js
```diff
@@ -0,0 +1,192 @@
...
+    const buffStacks = this.selectedCombatant.getBuffHistory(SPELLS.CUSTODY_OF_THE_DEEP_BUFF.id).reduce((obj, buff) => {
+      obj[buff.start] = buff.stackHistory;
+      return obj;
+    }, {});
+
+    Object.values(buffStacks).forEach((stackChain) => {
+      stackChain.forEach((stack) => {
+        const stackSize = stack.stacks;
+        const stackStart = stack.timestamp;
+        if (cleanStacks[lastHandledStack]) {
```
emallson: I believe the preferred form of key-check here is `if(cleanStacks[lastHandledStack] !== undefined)` to avoid the implicit "falsey" check that JS does.

src/parser/shared/modules/items/bfa/raids/crucibleofstorms/TridentOfDeepOcean.js
Zeboot: I actually ended up using the stack tracking code from the Navigation enchant analyzer, since I looked there to see how multiple stack average uptime had been done previously. In my opinion this could use either being moved into a separate module so that it can be reused more freely or manually rewritten using the event listeners (buffapply, refresh, etc). Let me know if you want me to do either of these things.

Other than that, I believe it is just checking that the same buff didnt register as changing in too short of a delay, as this wouldn't be realistic / bugged.

emallson: 👍 i will add an issue for it and approve the PR

PR #3115, opened 2019-04-24
[Outlaw] Abilities, Buffs, and Spells updates

Coywolf:
This is a few general changes to the spec. I added a couple of missing spells that were causing the Combo Point Usage and Energy Usage tabs to error. Did a pass on the abilities module to tweak/add some basic recommendations, and added a buffs module, basing it on other specs. Added a contributors entry for myself.

(Split 1 of 5 from my original: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3113)

### DIFF
--- a/src/parser/rogue/outlaw/modules/Abilities.js
+++ b/src/parser/rogue/outlaw/modules/Abilities.js
@@ -68,6 +68,21 @@ class Abilities extends CoreAbilities {
         },
         enabled: combatant.hasTalent(SPELLS.GHOSTLY_STRIKE_TALENT.id),
       },
+      {
+        spell: SPELLS.BETWEEN_THE_EYES,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        cooldown: 30,
+        gcd: {
+          static: standardGcd,
+        },
+      },
+      {
+        spell: SPELLS.PISTOL_SHOT,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        gcd: {
+          static: standardGcd,
+        },
+      },
       // Rotational (AOE)
       {
         spell: SPELLS.BLADE_FLURRY,
@@ -88,15 +103,6 @@ class Abilities extends CoreAbilities {
         },
         castEfficiency: {
           suggestion: true,
-        },
-      },
-      {
-        spell: SPELLS.VANISH,
-        category: Abilities.SPELL_CATEGORIES.UTILITY,
-        cooldown: 120,
-        gcd: null,
-        castEfficiency: {
-          suggestion: false,
         },
       },
       {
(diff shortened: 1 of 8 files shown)


src/parser/rogue/outlaw/modules/Abilities.js
```diff
@@ -120,6 +135,7 @@ class Abilities extends CoreAbilities {
         },
         castEfficiency: {
           suggestion: true,
+          extraSuggestion: <>In most cases this should be considered part of your rotation, only below <SpellLink id={SPELLS.ROLL_THE_BONES.id} icon /> on priority. However you should delay using it to line it up with <SpellLink id={SPELLS.BLADE_FLURRY.id} icon /> when there is more than one target.</>,
```
TsaboTavok: In my opinion, the first part is over the top. Its enough to write `Only delay XXX for AoE scenarios`. 
Not insisting on it tho

src/parser/rogue/outlaw/modules/core/ComboPoints.js
```diff
@@ -39,10 +39,10 @@ class ComboPoints extends Analyzer {
...
     resourceSuggest(when, this.comboPointTracker, {
-      spell: SPELLS.PISTOL_SHOT, // 1 CP
+      spell: SPELLS.PISTOL_SHOT, // 1 CP, 2 with proc and Quick Draw
       minor: 0,
-      avg: 0.1,
-      major: 0.2,
+      avg: 0.05,
+      major: 0.1,
       // TODO: Beta logs incorrectly add bonus CP from an extra Sinister Strike to Pistol shot. 
       // Want to see if its fixed in PrePatch or can be fixed on WCL side before creating workaround.
```
TsaboTavok: I am amusing you tighten the suggestions because this "bug" is resolved? Would be nice to remove the comment if its no longer relevant.

src/parser/rogue/outlaw/modules/Abilities.js
```diff
@@ -65,9 +69,26 @@ class Abilities extends CoreAbilities {
         },
         castEfficiency: {
           suggestion: true,
+          recommendedEfficiency: 0.95,
+          extraSuggestion: 'The only time you should delay casting this is to avoid over-capping Combo Points.',
```
TsaboTavok: Not technically correct, sometimes needs to be saved for burst windows.
This is why I prefer to not be super prescriptive with suggestions, setting ` recommendedEfficiency: 0.95,` already implies it should never be delayed without a good reason.

src/parser/rogue/outlaw/modules/Abilities.js
Coywolf: That's a good idea, that's a much more succinct way of putting it

src/parser/rogue/outlaw/modules/core/ComboPoints.js
Coywolf: Yep fair, missed the comment. And yea, my logic was just to align it to other rotational abilities, comparing to SS here and the other specs

emallson:
@TsaboTavok have the recent changes addressed your comments sufficiently to merge?

TsaboTavok:
All good from me

PR #3116, opened 2019-04-24
[Outlaw] Energy and ComboPoint tracker updates

Coywolf:
Changes to account for energy max and regen from Adrenaline Rush and Buried Treasure.
Added a way to account for the missing CPs generated from the bonus hit of Sinister Strike

(Split 2 of 5 from my original: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3113/files)

### DIFF
--- a/src/parser/rogue/shared/resources/EnergyCapTracker.js
+++ b/src/parser/rogue/shared/resources/EnergyCapTracker.js
@@ -38,13 +38,8 @@ class EnergyCapTracker extends RegenResourceCapTracker {
   static exemptFromRefund = [
   ];
 
-  increasedBaseRegen(){
-    return 0;
-  }
-
   naturalRegenRate() {
     let regen = super.naturalRegenRate();
-    regen += this.increasedBaseRegen();
     if (this.selectedCombatant.hasTalent(SPELLS.VIGOR_TALENT.id)) {
       regen *= VIGOR_REGEN_MULTIPLIER;
     }
--- a/src/parser/rogue/shared/resources/EnergyCapTracker.js
+++ b/src/parser/rogue/shared/resources/EnergyCapTracker.js
@@ -11,10 +11,13 @@ import RegenResourceCapTracker from 'parser/shared/modules/RegenResourceCapTrack
 import SpellEnergyCost from './SpellEnergyCost';
 
 const BASE_ENERGY_REGEN = 10;
+const BURIED_TREASURE_REGEN = 4;
 const VIGOR_REGEN_MULTIPLIER = 1.1;
+const ADRENALINE_RUSH_REGEN_MULTIPLIER = 1.6;
 
 const BASE_ENERGY_MAX = 100;
 const VIGOR_MAX_ADDITION = 50;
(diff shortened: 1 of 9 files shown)


src/parser/rogue/shared/resources/EnergyCapTracker.js
```diff
@@ -38,8 +38,13 @@ class EnergyCapTracker extends RegenResourceCapTracker {
   static exemptFromRefund = [
   ];
 
+  increasedBaseRegen(){
+    return 0;
+  }
+
   naturalRegenRate() {
     let regen = super.naturalRegenRate();
+    regen += this.increasedBaseRegen();
```
Coywolf: Is there a better way to accomplish this? Buried Treasure is a modifier to base regen, applied before the Vigor multiplier. It didn't look like I could just modify the static base regen property as it needs to be dynamic

TsaboTavok: I think you want to update `static baseRegenRate = BASE_ENERGY_REGEN;` based on this buff, so maybe change it to be a function. 

You can also try overriding  `naturalRegenRate()` by applying Buried Treasure or (Buried Treasure*Vigor).
But  not sure if its better then your solution.

Coywolf: I gave it a shot to move the adjustment logic to be just within the subclass, but it still feels a bit odd. In my override of naturalRegenRate(), I just call an update to set baseRegenRate as the first step, before going up to super.naturalRegenRate(). 

Having baseRegenRate be a function seems like it'd be ideal, but I think that would require the RegenResourceCapTracker to be implementing it as a function, right?

TsaboTavok: Not a JS expert, so cant tell you the optimal way to do this, but you can chanage the RegenResourceCapTracker. Clearly you have a valid use case for it. There are only 2 usages of it right now so it wont be hard to do

Coywolf: Gave that a shot in the last commit. I just added a new method in the RegenResourceCapTracker which by default returns the baseRegenRate. This way it doesn't have any impact to the other specs using it and can just be overridden as required

src/parser/shared/modules/RegenResourceCapTracker.js
```diff
@@ -195,6 +195,16 @@ class RegenResourceCapTracker extends Analyzer {
...
   }
 
+  /**
+   * IMPLEMENTME
+   * Some specs have modifiers to the base regen rate that apply before haste, which should
+   * be implemented here.
+   * @returns {number} Base regen rate of the resource.
+   */
+  getBaseRegenRate(){
+    return this.constructor.baseRegenRate;
```
TsaboTavok: I don't like having 2 ways to do something, so I would have just this method that everyone must override, but better ask maintainers what the policies are =)

TsaboTavok:
Everything apart from the minor comment above looks good

PR #3117, opened 2019-04-27
Fixed issue where rdruid clearcasting module was using regrowths cast during innervate

Qbz23:


Qbz23:
@buimichael

Qbz23:
Doesnt increment totalRegrowths but it's not actually used anywhere not sure if it should be removed.

PR #3126, opened 2019-05-04
Removed unneeded tab

abelito75:
Old feature removed
The spreadsheet that used this data is out of date and has not been maintained thus this data can only hinder people and should be removed. 

Talked to Anomoly and he agrees.


emallson:
@abelito75 can you remove the `MonkSpreadsheet.js` file as well?

abelito75:
Yeah. I didn't even think to look there. Thanks for the catch. Time to make mw great again (as if its not great already :P)

PR #3128, opened 2019-05-06
[Shaman] Added Igneous Potential Azerite Trait

niseko:
For both Elemental
image
and resto
image

Also moved Synapse Shock so that resto can implement it.


niseko:
As is stated in the tooltip

HawkCorrigan:
Ye I saw that
I am still not sure if having this box is not just leading to more problems.

Impo the information provided is not helpful in trying to either improve gameplay or judge how good the trait actually is.

The only point I am seeing for having it is striving for completion, which I am not too convinced by.

I can already see people showing up and asking why we are suggesting IP because the tool tells them that it is "bad".

If you want to go ahead with it, don't let me stop you obviously but ye my 2ct.

niseko:
I don't mind removing it from elemental as it was intended for restoration either way; as there is no simc implementation for it (yet, my PR is pending). The trait should always show as best regardless of the lava surge component as long as the player is pressing lava burst on cd (single target anyway), given how strong it is. Screenshot is from one of my Uu'nat tries and its far out ahead of any other trait I have, closest is lava shock @ 3.5% / trait.

Its just one of those hidden things where you don't even know how much it contributed to lava burst unless you either manually calculate it (where people will forget the class aura and probably even vers) or trust simcraft which doesn't give you fight specific values.

Issue #2962, opened 2019-01-29
Add average item level and average heart level to raid composition page


MartijnHols:
![image](https://user-images.githubusercontent.com/4565223/51935953-6a66be00-2407-11e9-9797-60385a055858.png)

| Item level | Heart level | ...roles |
| --|--|-- |
| 373 | 23 | #2961 |

florianschut:
I'll have a look at this assuming no more changes are needed to #3020 😃

florianschut:
iLvL is already in, I want to try to add HoA level to it this weekend but want to do that simultaneously with moving the calculations to the RaidCompositionDetails.js to keep PlayerLoader.js clean and the details self-contained... I do need to wrap my head around the react stuff as that is what ground me to a halt with this last week... If this is inconvenient regarding the release of next poke either here or on Discord.

Scotsoo:
Done! Added in my latest commits!

MartijnHols:
Thanks!

PR #3130, opened 2019-05-10
[Core] Adding average HOA lvl in RaidCompositionDetails

Scotsoo:
Implements #2962


Scotsoo:
First time using react so any feedback is appreciated

Sharrq:
Can you also include a screenshot so whoever is reviewing it can see what it looks like while they are looking at the code?

Scotsoo:
Here you are!
image

src/CHANGELOG.js
```diff
@@ -1,12 +1,18 @@
...
+import { Yajinni, Mamtooth, Zerotorescue, Putro, joshinator, Gebuz, ackwell, emallson, blazyb, Dambroda, Nalhan, Satyric, niseko, Khadaj, Fyruna, Matardarix, jos3p, Aelexe, Chizu, Hartra344, Hordehobbs, Dorixius, Sharrq, Scotsoo } from 'CONTRIBUTORS';
 import ItemLink from 'common/ItemLink';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 import SpellLink from 'common/SpellLink';
 import Contributor from 'interface/contributor/Button';
 
 export default [
+  {
+
```
emallson: avoid blank lines like this in the future

src/interface/report/RaidCompositionDetails.js
```diff
@@ -10,23 +10,31 @@ class RaidCompositionDetails extends React.PureComponent {
...
-    const { tanks, healers, dps, ranged, ilvl } = this.props;
+    const { tanks, healers, dps, ranged, ilvl, heartLvl } = this.props;
 
     return (
       <div className="raid-composition">
         <div className="bar">
           <div className="panel-body" style={{ padding: 0 }}>
             <div className="flex">
-              <div className="flex-sub icon">
+            <div className="flex-sub icon">
```
emallson: this should be indented as it was

src/CHANGELOG.js
Scotsoo: Fixed!

src/interface/report/RaidCompositionDetails.js
Scotsoo: Fixed also!

Scotsoo:
@emallson yeah, I was concerned too with the requests - but I too saw they were cached cause they were originally called in the specific player's screen. I couldn't figure out anywhere else to get the HOA level from - shame it doesn't come back in the player's items on the combatants object where I think we have access to all other bits of gear.

PR #3133, opened 2019-05-14
[Core] Add basic dispel tracker

### LINKED ISSUE #3064: Add a statistic: dispelled debuffs
A list of all debuffs the player dispelled with their dispell.

The `dispell` event has an `extraAbility` showing what was dispelled, so it's part of the data we receive.

avilene:
Resolves #3064 

image

> Yes, dispel is with one l, dispells is with 2. #dictionary

Used this log for testing: https://www.warcraftlogs.com/reports/YTmBqXd694N3gDJF


MartijnHols (approved):
thanks

PR #3141, opened 2019-05-17
Added Void Stone

Zeboot:
**REQUIRES https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3139 TO BE MERGED FIRST**

Adds the trinket "Void Stone" from crucible of storms to the item analysis page

Example log: https://www.warcraftlogs.com/reports/vZQfwra1xVK76M4L/#fight=13&source=6

Box:
image

Tooltip:
image


Zeboot:
I actually have a question about the display this item in particular: Ideally, I would like to display the absorbs per usage not in the tooltip, but in a dropdown, similar to how it's done on the Navigation enchants or for example Thermal Void:
image
Looking at the code for these, however, they're using statistics boxes that have been marked as deprecated. Is there any other way to do this already implemented or should I just either use the deprecated box or edit the existing Statistic/create a new class to allow dropdowns? (or just leave it as is if you'd rather have the tooltip)

Zeboot:
I've adjusted the box by adding a dropdown functionality identical to the one used in StatisticBox to Statistic via the dropdown prop, as StatisticBox is marked as deprecated and Statistic didn't contain this functionality yet.
New look:
image
image
image

PR #3142, opened 2019-05-19
Dev/avenging crusader

5chmidt:
image
Created a statistic box for Avenging Crusader which will track the combined healing because hit and crits have separate ids.  The statistic box will also track how much healing is transferred to beacon target(s).   Judgment and Crusader hits while Avenging Crusader is active will add them to the tally of healing spells.


src/parser/paladin/holy/modules/AlwaysBeCasting.js
```diff
@@ -8,23 +8,28 @@ class AlwaysBeCasting extends CoreAlwaysBeCastingHealing {
...
     SPELLS.HOLY_LIGHT.id,
     SPELLS.HOLY_SHOCK_CAST.id,
-    // ABILITIES.JUDGMENT_CAST.id, // Only with either JoL
     SPELLS.LIGHT_OF_DAWN_CAST.id,
     SPELLS.LIGHT_OF_THE_MARTYR.id,
     SPELLS.BESTOW_FAITH_TALENT.id,
     SPELLS.HOLY_PRISM_TALENT.id,
     SPELLS.LIGHTS_HAMMER_TALENT.id,
-    // ABILITIES.CRUSADER_STRIKE.id, // Only with Crusader's Might, is added in constructor if applicable
+    SPELLS.LAY_ON_HANDS.id,
```
MartijnHols: LoH is off the GCD actually; https://www.wowhead.com/spell=633/lay-on-hands

src/parser/paladin/holy/modules/talents/AvengingCrusader.js
```diff
@@ -0,0 +1,90 @@
...
+          <>
+            <ItemHealingDone amount={this.totalHealing} /><br />
+            {formatPercentage(this.critRate)}% Crit Rate
+          </>
+        )}
+        tooltip={(
+          <>
+            Hits: <b>{this.hits}</b> Crits: <b>{this.crits}</b><br />
+            Overhealed: <b>{formatPercentage(this.overHealing / (this.healing + this.overHealing))}%</b><br />
+            Beacon healing transfered: <b>{formatNumber(this.healingTransfered)}</b><br />
```
MartijnHols: How about showing beacon healing overhealing as well? (not sure if possible)

5chmidt: @MartijnHols, that would be possible, but I didn't know if it would be helpful.  
If the beacon target(s) are being over-healed, the corrective action would be to keep moving your beacon as people get topped off.  I'm not sure how many people would actually do that.

MartijnHols: I wasn't really thinking about how people would be using it, just that it's interesting info they can use for various reasons. Using your case as an example, maybe seeing it would also help to relatavize the throughput because the fight is less tank damage heavy.

PR #3150, opened 2019-05-23
[Outlaw] Fixing a bug with the low value RTB suggestion

Coywolf:
Previous way was counting all casts made instead of rerolling, when it should have been counting how many of the rolls were delayed. The fix is to mark the cast objects with a flag that can be counted, instead of a separate counter variable.


TsaboTavok:
Looks reasonable to me

Issue #3149, opened 2019-05-22
[Prot Pal] Self-Heal Timing chart shows heals occuring at the HP *after* the heal


emallson:
For example, a heal for 209k (w/ 50k overhealing) is shown happening at 100% HP. It should appear at the previous HP value.

abelito75:
@emallson close this please

emallson:
Addressed by #3151

PR #3151, opened 2019-05-23
Updated

abelito75:
#3149 Ticket related
Fix dot location
Fix tool tip
this was before the fix
pre
this was after the fix
post

### DIFF
--- a/src/parser/shared/modules/features/SelfHealTimingGraph.js
+++ b/src/parser/shared/modules/features/SelfHealTimingGraph.js
@@ -80,7 +80,7 @@ class SelfHealTimingGraph extends Analyzer {
       });
 
     const _casts = this._selfhealTimestamps.map(event => {
-      const p = (event.hitPoints / event.maxHitPoints) || 0;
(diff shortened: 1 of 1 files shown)


emallson:
such a descriptive PR name lol

src/parser/shared/modules/features/SelfHealTimingGraph.js
```diff
@@ -80,15 +80,15 @@ class SelfHealTimingGraph extends Analyzer {
       });
 
     const _casts = this._selfhealTimestamps.map(event => {
-      const p = (event.hitPoints / event.maxHitPoints) || 0;
+      const p = ((event.hitPoints - (event.amount || 0) + (event.absorb || 0)) / event.maxHitPoints) || 0;
```
emallson: since you re-use this formula, should probably do like

```js
const startingHP = event.hitPoints - (event.amount || 0) + (event.absorb || 0);
const p = startingHP / event.maxHitPoints;
```
and then re-use `startingHP` on line 91

abelito75:
Its a very accurate pr name

PR #3152, opened 2019-05-23
fix crash in player loader when a friendly is missing

emallson:
specifically: if a player combatant is in the log, but not in the
report.friendlies list, the player selection screen would crash. this
fixes it

see log warcraftlogs.com/reports/HF93kqPQMDhx6Bbg#fight=1 for an example of the issue


emallson:
tagging @yajinni for review

PR #3154, opened 2019-05-24
Search

Zeboot:
Should implement #1940 as requested.

I'm not entirely sure if this is how you wanted it done (and I haven't been able to test it 100% due to the opensearch scheme not allowing localhost:3000 as target and I didn't want to bother setting up a subdomain to test with for something relatively simply, I've successfully tested it by searching on the live version of wowanalyzer and then just replacing the start of the URL with localhost:3000/search/), but I've been able to successfully input the same report schemes usable in the homepage search box as a route to redirect to the report, if one is found. If the scheme is invalid, it shows the supported schemes.

Let me know if you had something different in mind for how you wanted it to work.

Here's a gif of how I manually tested:
https://puu.sh/DwveK/711a223c3f.gif


src/interface/home/Page.js
```diff
@@ -28,6 +28,7 @@ const Premium = lazyLoadComponent(() => retryingPromise(() => import(/* webpackC
 const About = lazyLoadComponent(() => retryingPromise(() => import(/* webpackChunkName: 'About' */ 'interface/about').then(exports => exports.default)));
 const HelpWanted = lazyLoadComponent(() => retryingPromise(() => import(/* webpackChunkName: 'HelpWanted' */ 'interface/helpwanted').then(exports => exports.default)));
 const ContributorPage = lazyLoadComponent(() => retryingPromise(() => import(/* webpackChunkName: 'ContributorPage' */ 'interface/contributor/Page').then(exports => exports.default)));
+const Search = lazyLoadComponent(() => retryingPromise(() => import(/* webpackChunkName: 'ContributorPage' */ 'interface/search').then(exports => exports.default)));
```
MartijnHols: The chunk name should differ from contributorpage.

src/interface/search/index.js
```diff
@@ -0,0 +1,64 @@
...
+        {this.state.valid ? <>Searching for </> : <>Invalid search parameter: </>}<b>{this.props.query}</b><br />
+        <br />
+        <Trans>
+          Supported search terms:<br />
+          <ul>
+            <li>&lt;report code&gt;</li>
+            <li>https://www.warcraftlogs.com/reports/&lt;report code&gt;</li>
+            <li>https://www.warcraftlogs.com/character/&lt;region&gt;/&lt;realm&gt;/&lt;name&gt;</li>
+            <li>https://worldofwarcraft.com/&lt;language-code&gt;/character/&lt;realm&gt;/&lt;name&gt;</li>
+            <li>https://www.wowchina.com/&lt;language-code&gt;/character/&lt;realm&gt;/&lt;name&gt;</li>
```
MartijnHols: I don’t think we support these links since the Chinese API doesn’t work

src/interface/home/Page.js
Zeboot: oh whoops I thought I changed that when I copied the previous line, guess I forgot. I'll fix that real quick

src/interface/search/index.js
Zeboot: I just copied the tooltip from the main page report search box, I can remove that line if it's not supported after all

PR #3155, opened 2019-05-25
[Outlaw] BTE, Dispatch, Opportunity suggestions

Coywolf:
Adding modules for tracking inefficient casts related to Between the Eyes and Opportunity procs. Suggestions include percentage of sinister strike casts while an opportunity proc is up, percentage of BTE casts that were done at an inefficient time, and percentage of dispatch casts that could have been BTE.

I became a big fan of the FilteredDamageTracker per tsabo's suggestion and was able to use it for all three modules. My original PR was attempting to count how many BTE casts were delayed, and I was struggling when trying to make the suggestion scale with fight length. I realized though that the suggestion wasn't really about between the eyes anyway, it was more about the dispatch casts, so I split that into it's own module. The BetweenTheEyes module is now only for tracking BTE casts done at incorrect times.

(Split 3 of 5 from my original: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3113)


src/parser/rogue/outlaw/modules/spells/BetweenTheEyes.js
```diff
@@ -0,0 +1,51 @@
...
+        average: 0.05,
+        major: 0.1,
+      },
+      style: 'number',
+    };
+  }
+
+  suggestions(when) {
+    when(this.thresholds).addSuggestion((suggest, actual, recommended) => {
+      return suggest(<>You casted <SpellLink id={SPELLS.BETWEEN_THE_EYES.id} /> without having <SpellLink id={SPELLS.RUTHLESS_PRECISION.id} /> active. When you don't have either the <SpellLink id={SPELLS.ACE_UP_YOUR_SLEEVE.id} /> or <SpellLink id={SPELLS.DEADSHOT.id} /> traits, the only time you should use <SpellLink id={SPELLS.BETWEEN_THE_EYES.id} /> as your damaging finisher is during <SpellLink id={SPELLS.RUTHLESS_PRECISION.id} />.</>)
```
Coywolf: Something I couldn't quite decide on, is the second sentence here needed? Is it preferable to give all of the information about why something should/shouldn't be done, or stick to only what's relevant to that player in that fight?

TsaboTavok: In suggestions it's good to give more information

TsaboTavok:
I think this is good, only slightly concerned about how the thresholds are calculated.

Rather then doing `[Bad sinister strike]/[All sinister strikes]` it may be better to use `[Bad sinister strike]/[Good Pistol Shots]`.

I don't think this blocks a merge.

PR #3165, opened 2019-06-02
Dev/ignition mages fuse

5chmidt:
Created an item statistic for Ignition Mage's Fuse, showing average haste gained with tooltip to display utilization and buff up-time.  A suggestion will be shown if the utilization is less than %80 of potential uses.

image

Example Log:
https://www.warcraftlogs.com/reports/2aC7VhbnRdJ8H4TX#fight=last&source=13

### DIFF
--- a/src/parser/shared/modules/items/bfa/dungeons/IgnitionMagesFuse.js
+++ b/src/parser/shared/modules/items/bfa/dungeons/IgnitionMagesFuse.js
@@ -1,0 +1,56 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import ITEMS from 'common/ITEMS';
+import UptimeIcon from 'interface/icons/Uptime';
+import HasteIcon from 'interface/icons/Haste';
+import ItemStatistic from 'interface/statistics/ItemStatistic';
+import BoringItemValueText from 'interface/statistics/components/BoringItemValueText';
+import Analyzer from 'parser/core/Analyzer';
+import { formatPercentage, formatNumber } from 'common/format';
+import { calculateSecondaryStatDefault } from 'common/stats';
+
+/**
+ * Ignition Mage's Fuse -
+ * Use: Ignite the fuse, gaining 312 Haste every 4 sec. Haste is removed after 20 sec.
+ */
+class IgnitionMagesFuse extends Analyzer {
+  statBuff = 0;
+  statBudget = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrinket(ITEMS.IGNITION_MAGES_FUSE.id);
+
+    if(this.active) {
+      this.statBuff = calculateSecondaryStatDefault(300, 84, this.selectedCombatant.getItem(ITEMS.IgnitionMagesFuse.id).itemLevel);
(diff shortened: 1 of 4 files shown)


src/parser/shared/modules/items/bfa/dungeons/IgnitionMagesFuse.js
```diff
@@ -0,0 +1,110 @@
...
+      actual: this.utilization,
+      isLessThan: {
+        minor: .8,
+        average: .7,
+        major: .6,
+      },
+      style: 'number',
+    };
+  }
+  suggestions(when) {
```
Zeboot: Might be worth using the inbuilt "cast efficiency" suggestion here instead of creating it for yourself again
example: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/94848149ec7e9b63e8f97f648282006c04ebf392/src/parser/shared/modules/items/bfa/raids/crucibleofstorms/LurkersInsidiousGift.js#L51-L60

5chmidt: Thanks for the suggestion, I didn't know that existed.  I'll add it to the PR.

5chmidt: I changed to this:
   ```
 this.abilities.add({
      spell: SPELLS.IGNITION_MAGES_FUSE_BUFF,
      buffSpellId: SPELLS.IGNITION_MAGES_FUSE_BUFF.id,
      name: ITEMS.IGNITION_MAGES_FUSE.name,
      category: Abilities.SPELL_CATEGORIES.ITEMS,
      cooldown: ACTIVATION_COOLDOWN,
      gcd: null,
      castEfficiency: {
        suggestion: true,
      },
    });
```
But it's no longer picking up the number of casts correctly, this might be because its a stacking buff, I'm not sure, I double checked the spellBuffId and it is correct.  But on the suggestion it is always showing zero casts.

Is there a way to specify the number of casts?  I have `this.users` returning the correct value, if there is a way to pass it through.

Comment (Zeboot) [reply src/parser/shared/modules/items/bfa/dungeons/IgnitionMagesFuse.js]: it looks like for some reason that specific trinket doesn't have a cast event associated with it, which is why the tracker isn't tracking it by default (might be something worth adding into a normalizer at some point to fix missing "CAST" events)

adding the, albeit deprecated, casts and maxCasts should be enough to make it work
```javascript
this.abilities.add({
      spell: SPELLS.IGNITION_MAGES_FUSE_CAST,
      name: ITEMS.IGNITION_MAGES_FUSE.name,
      category: Abilities.SPELL_CATEGORIES.ITEMS,
      cooldown: ACTIVATION_COOLDOWN,
      castEfficiency: {
        suggestion: true,
        casts: () => this.uses,
        maxCasts: () => this.possibleUseCount,
      },
    });
```

5chmidt: I have added `maxCasts: () => this.possibleUseCount` to the PR.  I looked around for an example that shows the fabricating of a cast event.  If someone could point me in the right direction that would be great.
Thanks,

PR #3176, opened 2019-06-10
Add normalizer for buffs that are missing cast events

Zeboot:
Some items that provide buffs do not produce associated cast events for using the item (see Ignition Mage's Fuse for example), leading to them having to manually track uses for e.g. the usage statistic.

This PR adds a normalizer that fabricates cast events when the buffs given are applied.
Currently only contains the Ignition Mage's fuse (as this is where the cast being missing was noticed), but obviously provides an easy way to add further items to it.

### DIFF
--- a/src/parser/shared/normalizers/MissingCasts.js
+++ b/src/parser/shared/normalizers/MissingCasts.js
@@ -13,10 +13,12 @@ class MissingCasts extends EventsNormalizer {
   ];
 
   normalize(events) {
-    const missingCastEvents = events.filter(event => event.type === 'applybuff' && this.constructor.missingCastBuffs.includes(event.ability.guid));
+    const missingCastEvents = events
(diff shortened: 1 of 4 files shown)


src/parser/shared/normalizers/MissingCasts.js
```diff
@@ -0,0 +1,42 @@
+import EventsNormalizer from 'parser/core/EventsNormalizer';
+import SPELLS from 'common/SPELLS';
+
+/**
+ * During analysis there's no way to know at a `begincast` event if it will end up being canceled. This marks all `begincast` events by the player with an `isCancelled` property whether it was cancelled.
```
MartijnHols: this description seems invalid

Zeboot: dammit I keep forgetting to change comments when I copy code from another module, fixing that now

PR #3180, opened 2019-06-15
fix shared CDs in timeline (e.g. potions)

Zeboot:
This should fix the wrong potion being shown in the timeline (or rather, int potion being shown for everyone) by grouping timeline spells by "canonical ID" (first ID if multiple IDs share cooldown) and sending the applybuff event associated with the cast instead of its shared ID.

Should also fix a number of other issues on the timeline with spells that are considered shared cooldown.

Examples of timeline on live:
image
image

Examples of timeline with this fix:
image
image

I tried to keep this fix as "low impact" as possible as I do not fully know what effects larger changes would have globally (especially in relation to abilities that share a cooldown), hence why the fix isn't as "clean" as I'd like it to be. Let me know if you had sth else in mind for this


src/interface/report/Results/Timeline/Component.js
```diff
@@ -109,6 +109,18 @@ class Timeline extends React.PureComponent {
     return eventsBySpellId;
   }
 
+  _getCanonicalId(spellId){
```
MartijnHols: This seems sensible, though not 100% sure it's without problems.

src/parser/shared/modules/items/Potion.js
```diff
@@ -53,9 +53,12 @@ class Potion extends Analyzer {
       ...this.constructor.extraAbilityInfo,
     });
     if (this.constructor.extraAbilityInfo.buffSpellId) {
-      this.buffs.add({
-        spellId: this.constructor.extraAbilityInfo.buffSpellId,
-        triggeredBySpellId: this.constructor.spells.map(spell => spell.id),
+      //assign each buff its corresponding spell ID
```
MartijnHols: makes sense

src/parser/shared/normalizers/PrePullCooldowns.js
```diff
@@ -109,7 +108,11 @@ class PrePullCooldowns extends EventsNormalizer {
         for (let i = 0; i < buffSpells.length; i += 1) {
           if (buffSpells[i].buffId === event.ability.guid) {
             debug && console.debug(`Detected a precast buff cooldown: ${event.ability.name}`);
-            prepullCasts.push(this.constructor._fabricateCastEvent(event, buffSpells[i].castId));
+            if(buffSpells[i].castId instanceof Array && buffSpells[i].castId.includes(event.ability.guid)){ //try to find corresponding cast, otherwise pass list of casts
+              prepullCasts.push(this.constructor._fabricateCastEvent(event));
```
MartijnHols: shouldn't this pass `event.ability.guid` as second param?

Zeboot: without a param it defaults to event.ability as the new event.ability so the guid will be copied over too

src/interface/report/Results/Timeline/Component.js
Zeboot: it's the method used in SpellUsable.js to track shared cooldowns so I figured it'd be safe to use here too since the timeline should group abilities by shared cooldowns too I think

Issue #3169, opened 2019-06-09
Even More Info on Cooldown page for Balance druid produces error


Mrnikbobjeff:
Error occurred here: https://www.wowanalyzer.com/report/mwH9yDnMVhjkpRFf/18-Mythic+Lady+Jaina+Proudmoore+-+Wipe+17+(9:12)/Nusigma/cooldowns
Stack Trace:
TypeError: Cannot read property 'timestamp' of undefined
    at https://www.wowanalyzer.com/static/js/2.9ab68ef6.chunk.js:1:13904
    at Array.map (<anonymous>)
    at t.value (https://www.wowanalyzer.com/static/js/2.9ab68ef6.chunk.js:1:13741)
    at li (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:167501)
    at si (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:167294)
    at hi (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:171451)
    at Ki (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:186442)
    at Gi (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:186819)
    at Ra (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:193311)
    at Na (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:192691)
    at Fa (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:198906)
    at Cn (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:138005)
    at HTMLDocument.o (https://www.wowanalyzer.com/static/js/81.2b9901ac.chunk.js:1:214616)

The above error occurred in the component:
    in t
    in li
    in ul
    in O
    in div
    in div
    in o
    in a
    in div
    in div
    in a
    in Connect(a)
    in a
    in a
    in Connect(a)
    in a
    in a
    in Connect(a)
    in t
    in withRouter(Connect(a))
    in a
    in Connect(a)
    in a
    in Connect(a)
    in a
    in Connect(a)
    in t
    in withRouter(Connect(a))
    in a
    in qn
    in t
    in t
    in div
    in a
    in Connect(a)
    in t
    in withRouter(Connect(a))
    in Unknown
    in t
[...]

PR #3184, opened 2019-06-16
Even More Info on Cooldown Dirty Fix

abelito75:
This pr is for #3169 
This is an EXTREMELY dirty fix this. Logic for this fix was just check if we have object we are looking for and if we don't then its a layer deeper. The issue lies on how this class is getting events

So what was done originally was we'd pass something like 
{event: {…}, amount: 2399, absorbed: 0, overheal: 0, count: 1}
absorbed: 0
amount: 2399
count: 1
event: {timestamp: 326345, type: "heal", sourceID: 24, sourceIsFriendly: true, targetID: 3, …}
overheal: 0

Meaning to get the data we need we'd have to dive down into event

But for some reason now we are getting a mix of events like that and events like 
{timestamp: 325804, type: "cast", sourceID: 24, sourceIsFriendly: true, targetID: 24, …}
ability: {name: "Innervate", guid: 29166, type: 8, abilityIcon: "spell_nature_lightning.jpg"}
armor: 2301
attackPower: 12659
classResources: [{…}]
facing: 581
globalCooldown: {timestamp: 325804, trigger: {…}, type: "globalcooldown", ability: {…}, sourceID: 24, …}
hitPoints: 249180
itemLevel: 409
mapID: 1354
maxHitPoints: 249180
rawResourceCost: {0: 0}
resourceActor: 1
resourceCost: {0: 0}
sourceID: 24
sourceIsFriendly: true
spellPower: 12172
targetID: 24
targetIsFriendly: true
timestamp: 325804
type: "cast"
x: -78402
y: -111286

and there isn't an event object in there so we were trying to get a value from undefined 

TLDR We are pass in wrapped events and unwrapped events and this causes us to hit issues with attempting to get data from undefined objects

NOTE: Due to getting events in two ways we can get some very strange results, no nulls, NA, or crashes but illogical numbers.

I thought it would be better to provide a fix so the service isn't just crashing then go into detail


abelito75:
I'd imagine this would work and be cleaner. I wasnt really thinking too critically. Had some other irl stuff on my mind while idly doing this

yajinni:
Closing due to this PR taking it over and solving the issue in a better way
https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3194

Issue #2940, opened 2019-01-14
Implement time filtering (phase selection)


MartijnHols:
# Requires considerable knowledge of the project. It is recommended to _not_ attempt this as your first issue but to get familiar with the project first through easier things.

Now that #2706 has been completed, giving us info about when the fight's phases are, the next step is to implement time filtering. i.e. to be able to parse only the events within the select time frame.

I'll write the full specifications later. The gist of it is:

- Be able to filter the results by the selected time period.
- Modules have to properly show the right info for the time period. No leakage from outside the time period unless it would make the resulting stat more accurate.
- Normalizers should obviously continue to work, as best as possible.
- Provide a phase selection dropdown in the results page that applies a filter by the fight start and end timestamps.
- it should have cooldown and buff info from outside the phases

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=68697545)](https://www.bountysource.com/issues/68697545-implement-phase-selection-filtering)

MartijnHols:
Increased:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=68697545)](https://www.bountysource.com/issues/68697545-implement-phase-selection-filtering)

mul1sh:
@MartijnHols this bounty looks interesting 🙂  let me try it out

MartijnHols:
Increased:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=68697545)](https://www.bountysource.com/issues/68697545-implement-phase-selection-filtering)

mul1sh:
@MartijnHols I've been getting familiar with the project and I think I fully understand it now, will do my PR tomorrow 🙂

PR #3190, opened 2019-06-19
Time Filtering & Phase selection

Zeboot:
Figured I'd give #2940 a shot.

Please let me know of issues you find / comments / further todo / comments / etc.
Main things missing is testing to see what modules don't behave as expected. I expect a lot of modules that utilize timestamps in their display to need updating to contain the "offset_time", but small things like that are something that can be fixed as they're discovered "live", as I'll be fixing any issues that come up with the phase selection in the future even past the PR. Any modules I've tested for functionality (aka showing the correct actual data for the phase) work as expected.

**Currently Implemented**
- Event filtering by phase
- Event filtering by time
- Phase selection dropdown
- Only phases that were reached available to select
- Phase information gets passed in the fight object so other modules can be made to behave differently depending on if a phase is selected or not
- buff information from previous phases
- Death recap timestamps adjusted
- Timeline timestamps adjusted
- Warning about potentially inaccurate results if phase filtering is active
- Cooldown information from outside the phase

When phase information about a boss is found, a phase dropdown is displayed. Otherwise, only the time selection is displayed.

Selecting a phase sets the time filtering to the start and end timestamps of the phase. Selecting a time period resets phase to "all phases" and limits the events by the filter provided.

Example new fight object:
```javascript
fight: {
 start_time: 9198573, //required, timestamp of the start of the selected time period
 end_time: 9214374, //required, timestamp of the end of the selected time period
 offset_time: 161450, //required, amount of ms between start of fight and start of time period
 original_end_time: 9214374, //timestamp of the end of the fight
 boss: 2273, //required, boss ID
 filtered: true, //required, true if the fight is in any way filtered to a time period
 phase: "I1", //optional, phase key if a phase is selected
}
```
fight.phase corresponds to the key of the phase as defined in the boss object
Example boss object (src\raids\\{raid}\\{boss}.js):
```javascript
{
 id: 2273,
 ...
 fight: {
  ...
  phases: {
   P1: {...},
   I1: {
    name: "Intermission One",
    difficulties: [...],
    filter: {...},
    ...
   },
  },
 },
}
```
Additional information about the phase can be provided in this object and referenced via phase key

**Known Issues**
- Stacked buff uptimes incorrect (also incorrect on WCL) **fixed**
- Mitigation Details incorrect **fixed**
- Intermission Timeline container too wide **fixed**
- Page "freezes" when a large new phase is parsed (maybe show loading screen?) **fixed**
- Stagger information at start of phase missing **fixed**
- over 100% cast efficiency on some cooldowns **fixed**
- restorecharge events displayed before timeline **fixed**

examples of current state:
Phase switching gif: https://puu.sh/DK02K/545d3ba931.gif
image
image
image
image
image

on shorter pulls it only provides a dropdown for the phases that were actually reached:
image

image

fight object now passes offset_time in addition to start_time and end_time, offset_time is the time between the start of the fight and the start of the phase (which is encoded in start_time), 0 if the selected phase started at the beginning of the fight. fight.phase contains the phase key (as defined in the boss definition) if a phase is selected, otherwise remains undefined, providing a way for modules to check whether the report is filtered by phase or not

Example: Timeline gets shifted accordingly using offset_time, other modules need to be adjusted similarly, cooldowns from before the phase get displayed too

image


Zeboot:
should finally be ready for review, I think I've implemented what's requested in #2940, all that's left is feedback / further testing (which would be easier done on "live" since it should be restricted to individual modules displaying slightly incorrect information which is easily fixed)

niseko:
Over all it already seems to be working impressively well, just trying things out and listing things I notice.

using `/report/ZCYNL6r1DvKcMJhq/55-Mythic+Lady+Jaina+Proudmoore+-+Kill+(8:38)/Niseko`
Screenshots are filtering 07:00.000 to end

image
- Trying to adjust the selected field in the above screenshot in either time selection often sets the time I'm trying to adjust to the maximum amount (I want to set the milliseconds to 0, but am getting 8:38:876, overwriting the first 2 fields as well) or similar unexpected behavior. Doesn't happen if the first 2 fields are still on 0.
- A "reset filter" button or similar would be ideal as w/o a phase selector you manually have to set the start time back to 0
Similarly, selecting "All Phases" after manually inputting a time does not reset the time filter; maybe using a "Custom" entry in the combobox for manual time filtering makes sense for that
- Cooldowns, Triage, Raid health tabs and the 2 mana charts in statistics aren't adjusted and start at 0 seconds
- The "Be Well Prepared" checklist will set pre-pot usage to false if filtering more than ~20 second away from combat start (selected another player for this)
- Mana pool graph starts the first tick always at 100% mana 
image

- Issues with cast efficiency
image
image

- Not sure if this is feasible or worthy of implementation; or even relevant to this PR but it would be interesting to have the filter in the URL so you can share a filtered report.

Zeboot:
> Over all it already seems to be working impressively well, just trying things out and listing things I notice.
> 
> using `/report/ZCYNL6r1DvKcMJhq/55-Mythic+Lady+Jaina+Proudmoore+-+Kill+(8:38)/Niseko`
> Screenshots are filtering 07:00.000 to end
> 
> image
> 
> * Trying to adjust the selected field in the above screenshot in either time selection often sets the time I'm trying to adjust to the maximum amount (I want to set the milliseconds to 0, but am getting 8:38:876, overwriting the first 2 fields as well) or similar unexpected behavior. Doesn't happen if the first 2 fields are still on 0.

Should be fixed and general time selection behaviour improved in https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/7c01baf14ff0ea9f88a75b5f9ecaae4bb3e21113

> * A "reset filter" button or similar would be ideal as w/o a phase selector you manually have to set the start time back to 0
>   Similarly, selecting "All Phases" after manually inputting a time does not reset the time filter; maybe using a "Custom" entry in the combobox for manual time filtering makes sense as a solution for both of those things

https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/f95164adf3d420236d80bc13ce0641c109de63fc and https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/36599d5448a33b23cfde4bcd2853a4546fc065e9 : added a "custom" selector if manual filtering is selected, also added a reset button

> * Cooldowns, Triage, Raid health tabs and the 2 mana charts in statistics aren't adjusted and start at 0 seconds

fixed in https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/51bfbc040f8df2324e81a2f8bc33c070080a44c3 https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/e6e8c889e2f0eb9751d9ba32dc573da905941fd7 https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/edb3131eb3bb2eadc61ac347446915cda186bfb0 https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/aed0d3e24beb3e647223655bd4054230094e1a7c

> * The "Be Well Prepared" checklist will set pre-pot usage to false if filtering more than ~20 second away from combat start (selected another player for this)

looking into this next, not sure how easy this'll be

> * Mana pool graph starts the first tick always at 100% mana

fixed in https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3190/commits/bf5f4ccc8b4b9c5cbd7056bbf3619a9e656abbb4 : copying the first mana value found as initial value

>   image
> * Issues with cast efficiency

looking into this

>   image
>   image
> * Not sure if this is feasible or worthy of implementation; or even relevant to this PR but it would be interesting to have the filter in the URL so you can share a filtered report.

honestly I'm not even sure where I'd begin with this, I might look into it if I end up having time but this seems like something that'd be easier added later down the line as a separate PR

src/interface/report/EventParser.js
```diff
@@ -91,12 +91,11 @@ class EventParser extends React.PureComponent {
...
   makeEvents(parser) {
-    const { bossPhaseEvents, events } = this.props;
-    let combinedEvents = bossPhaseEvents ? [...bossPhaseEvents, ...events] : events;
+    let { events } = this.props;
     // The events we fetched will be all events related to the selected player. This includes the `combatantinfo` for the selected player. However we have already parsed this event when we loaded the combatants in the `initializeAnalyzers` of the CombatLogParser. Loading the selected player again could lead to bugs since it would reinitialize and overwrite the existing entity (the selected player) in the Combatants module.
-    combinedEvents = combinedEvents.filter(event => event.type !== 'combatantinfo');
-    combinedEvents = parser.normalize(combinedEvents);
-    return combinedEvents;
+    events = events.filter(event => event.type !== 'combatantinfo');
+    events = parser.normalize(events).sort((a,b) => a.timestamp - b.timestamp); //sort now normalized events
```
MartijnHols: please leave a comment to explain why (I can see you're sorting it now)

src/parser/shared/modules/features/RaidHealthTab/TabComponent/RaidHealthChart.js
```diff
@@ -57,8 +58,8 @@ class RaidHealthChart extends React.Component {
   }
 
   render() {
-    const { players, deaths, startTime, endTime } = this.props;
-
+    const { players, deaths, startTime, endTime, offsetTime } = this.props;
+    console.log(startTime, endTime, offsetTime);
```
MartijnHols:
```suggestion
```

Zeboot:
> A quick initial review. This looks like really amazing work!
> 
> One thing that we'll have to improve at a later time is the time selection. It's definitely good enough for initial release, but not very convenient in the long run.
> 
> The phase selection/time filter form does need to be integrated/hidden into the layout more. I'll think about this some more in the next few days to give some suggestions or contribute it myself.
> 
> I'll do another more extensive review in a few days, though I don't expect much more to come out of this.

Thanks!
Yea I wasn't 100% sure on how to implement the phase selection / time filtering interface. I'm not the greatest at frontend design so I just opted for having something functional over something pretty and figured it's easy enough to move later on once a better "spot" is set (I kinda like the way WCL does it in that compact display, something similar would definitely take a lot less space too). If you end up having something specific in mind I'm sure I can implement that.

The one thing I'm not too happy with myself right now is the responsiveness of the actual time filter input. The entire PR was done with only phase filtering in mind, after that was done and functional (yesterday I believe), I realized that I'm effectively filtering by timestamps anyways so that time input was more or less "slapped on" to make it useful outside of fights that have predefined phases we can detect, I'm gonna be trying to make that more responsive too

src/interface/report/EventParser.js
Zeboot: added a comment to the code, basically the reason is that after normalizing, the new events (like the fabricated prepull events etc) would get "mixed in" with the cast events "kept" from before the filter (to allow for cooldown information to remain accurate), additionally since buff events from before the filter get moved to the beginning of the filter, whereas cast events do not, this led to events being "out of order" even just before the phase. Granted the only events impacted by this were events before the filter / counted as prepull, so it didn't make a big impact, but I wanted to fix it anyways even if just for the events tab. Figured the easiest way to fix that is to just sort the full events list one last time before actually handing it off to the parser

Zeboot:
Heya just wanted to check if there's any update on this as I'd preferably like to fix/change/implement features before raid hits and takes up most of my free time :^)

Sharrq:
Out of curiosity (not required, im just curious), is there stuff in place to be able to use the time filtering in other modules? Like if i wanted to see the downtime during the 10 second window that a cooldown was used? Like some function or something that i can throw timestamps at?

Zeboot:
> Out of curiosity (not required, im just curious), is there stuff in place to be able to use the time filtering in other modules? Like if i wanted to see the downtime during the 10 second window that a cooldown was used? Like some function or something that i can throw timestamps at?

there's no separate time filtering function currently, as the filtering happens before the parser is initialized, though you can obviously easily just filter the events by a given timestamp (though I can implement this as an exported function of the time filter so you could in theory pass timestamps to filter by, not sure how much use that would see though)

Sharrq:
As a caster the biggest reason i was looking at it is because for some specs (like Arcane Mage and Fire Mage), your cooldown window is super critical, so if you have downtime during those buffs thats pretty bad, so if i was able to do something alongside the downtime stat box that was like "Your downtime was 15%", "Your downtime during Combustion was 2%" with an associated suggestion or whatever ... i think that would help players to understand why their damage is low.

idk how much use it would get either, but it might be useful if it is easy to add. It isnt really part of the scope of the issue or the feature, i was just curious.

Zeboot:
> As a caster the biggest reason i was looking at it is because for some specs (like Arcane Mage and Fire Mage), your cooldown window is super critical, so if you have downtime during those buffs thats pretty bad, so if i was able to do something alongside the downtime stat box that was like "Your downtime was 15%", "Your downtime during Combustion was 2%" with an associated suggestion or whatever ... i think that would help players to understand why their damage is low.
> 
> idk how much use it would get either, but it might be useful if it is easy to add. It isnt really part of the scope of the issue or the feature, i was just curious.

I'll quickly throw in an exported function that filters a set of events by given timestamps with the same function (to maintain cooldown information etc), but what you're describing would be a larger undertaking as modules such as downtime usually hook directly into their owner's event list, so those would have to be manually adapted to also be able to be called with a list of events instead (that you can then prefilter)

Sharrq:
Yea, i know the Downtime stuff would have to be modified to allow it to look at a subset of events. But it is a step towards it, and i might look into the downtime side of it myself.

Zeboot:
The filter function (that keeps "relevant" buff events etc) now gets exported to allow usage from external modules so other modules can be adjusted to allow for more precisely filtered information (aka multiple subsets of the fight)

Zeboot:
Fixed the conflicts so I'll merge it and close the issue now, I'll get started on writing up that news article though I don't know how much time I'll have to finish it before the weekend due to raid

PR #3192, opened 2019-06-23
[Outlaw] Finishers Module

Coywolf:
Adding a module for tracking efficient use of combo point spenders. Per tsabo's suggestion, I attempted to create a shared rogue module which could be reused by the other specs later on. The shared module has two main methods to implement in a spec specific module which define the CP threshold and the extra suggestion verbiage.

(Split 5 of 5 from my original: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3113)


emallson:
@TsaboTavok would you mind reviewing this?

PR #3195, opened 2019-06-25
Essence parsing & The Ever Rising Tide

niseko:
general support for essences
to check if someone has a specific essence you have to use the spellIds kihra is using in his talent pane
image

including the ever rising tide essence as an example
image
image
image

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,7 +9,7 @@ import Contributor from 'interface/contributor/Button';
 export default [
   {
     date: new Date('2019-06-25'),
-    changes: <>Added <SpellLink id={SPELLS.EVER_RISING_TIDE_MAJOR.id} /> and <SpellLink id={SPELLS.EVER_RISING_TIDE_MINOR.id} /> essences.</>,
+    changes: <>Added <SpellLink id={SPELLS.EVER_RISING_TIDE_MAJOR.id} /> and <SpellLink id={SPELLS.EVER_RISING_TIDE.id} /> essences.</>,
     contributors: [niseko],
   },
   {
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -8,6 +8,11 @@ import Contributor from 'interface/contributor/Button';
 
 export default [
   {
+    date: new Date('2019-06-25'),
+    changes: <>Added <SpellLink id={SPELLS.EVER_RISING_TIDE_MAJOR.id} /> and <SpellLink id={SPELLS.EVER_RISING_TIDE_MINOR.id} /> essences.</>,
+    contributors: [niseko],
+  },
+  {
     date: new Date('2019-06-19'),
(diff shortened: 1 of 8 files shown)


buimichael:
Looks really nice, just a few suggestions:
1. It would be nice to have a breakdown on each cast, to see how many stacks you managed to get for each cast and how much healing that contributed. Similar to how the Flourish module for resto druid or evangelism.

2. Think you missed the haste part of the essence. "... , and you gain 10% haste". I guess this is a little harder to translate to throughput, put it should be pretty straight forward to give the user avg haste gained over the fight. Best would obviously be to see how much throughput the haste contributed, but that's probably too class specific (not all spells scales with haste etc). HPM that is...

3. The mana regeneration stop, wouldn't this be dynamic in case you have external buffs that increases your MP5? Such as blessing of wisdom, maybe even symbol of hope? Most likely major Lucid dreams, however you can't equip 2 majors so I guess that's not an issue at least. Could be that these are all counted as external mana and is not affected by the drawback.

niseko:
>1. It would be nice to have a breakdown on each cast, to see how many stacks you managed to get for each cast and how much healing that contributed. Similar to how the Flourish module for resto druid or evangelism.

So like a table with: Cast#; avg.stacks, max stacks; healing contribution

>2. The 10% haste bonus during the buff is multiplicative with your existing Haste.

I'll see if i can manage something; multiplicative means one can't just do a simple calculation but look at the haste amount and changes during the buff uptime

> 3. The mana regeneration stop, wouldn't this be dynamic in case you have external buffs that increases your MP5? Such as blessing of wisdom, holy priest talent Enlightment possibly even Symbol of Hope? Most likely major Lucid dreams, however you can't equip 2 majors so I guess that's not an issue at least. Could be that these are all counted as external mana and is not affected by the drawback.

Mana regen is rarely adjusted directly; but Enlightenment actually does affect it which is decently easy to account for. Most things will still provide mana during ERT (wisdom, symbol, resurgence, mana pots etc)

niseko:
image
Split it into 2 because it felt like it got too much for 1 statistic box, and had to swap them around because dropdowns behave weirdly with statistic groups.

Ended up not adding average stacks for space reasons.

> Best would obviously be to see how much throughput the haste contributed, but that's probably too class specific (not all spells scales with haste etc).

Another issue is that the 10% haste during that duration is worth a lot more than haste usually would be, as it allows quicker stacking of the healing buff thus if that even gives you 1 more stack, its another 4% throughput for 8 seconds.

niseko:
An issue that I just found while trying out logs from the live game; the given spellids by warcraftlogs aren't set per trait like I was expecting but vary per rank, so there actually are 8-12 spellids per trait which WCL could give us. 
I've swapped it over to use trait IDs as they are distinctive regardless of rank or version. There probably is a better solution for this though

emallson:
@buimichael since you basically already reviewed this, wanna go over it again?

buimichael:
Looks good.

PR #3197, opened 2019-06-28
[BrM] Hot Trub Damage Estimator

emallson:
Shows an estimate of the damage the Hot Trub PvP talent would do if you'd used the Conflict & Strife essence. Hot Trub deals damage equal to 20% of purified damage, up to a maximum of 20% of your max HP per cast.

Hot Trub is not usable in PvE content *yet* so this lets us get an idea for how much damage it does in current content.

(screenshot looks mega large for some reason, not sure why)

Screenshot from 2019-06-28 12-00-25

### DIFF
--- a/src/parser/monk/brewmaster/modules/features/HotTrubTimeMachine.js
+++ b/src/parser/monk/brewmaster/modules/features/HotTrubTimeMachine.js
@@ -11,7 +11,7 @@ import { EVENT_STAGGER_POOL_ADDED, EVENT_STAGGER_POOL_REMOVED } from '../core/St
 import AgilityValue, { staggerPct } from './AgilityValue';
 import { lookupK } from '../constants/Mitigation';
 
-const debug = true;
+const debug = false;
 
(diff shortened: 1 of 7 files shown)


src/parser/monk/brewmaster/modules/features/HotTrubTimeMachine.js
```diff
@@ -0,0 +1,161 @@
...
+import StatTracker from 'parser/shared/modules/StatTracker';
+import StatisticBox from 'interface/others/StatisticBox';
+import SpellIcon from 'common/SpellIcon';
+import ItemDamageDone from 'interface/others/ItemDamageDone';
+import { formatNumber, formatPercentage } from 'common/format';
+import { EVENT_STAGGER_POOL_ADDED, EVENT_STAGGER_POOL_REMOVED } from '../core/StaggerFabricator';
+import AgilityValue, { staggerPct } from './AgilityValue';
+import { lookupK } from '../constants/Mitigation';
+
+const debug = true;
```
abelito75: Switch to false

Issue #3202, opened 2019-07-07
Update the "Be well prepared" checklist with new 8.2 food/flasks/enchants


buimichael:

emallson:
got a report on discord that machinist's brilliance is counting as a low-quality enchant

buimichael:
Think this issue has been resolved, closing.

PR #3206, opened 2019-07-11
Issue #3202 - 8.2 flasks, potions and ring enchants

leozhekov:
I am not sure if I've done this correctly as I don't really have much knowledge of react, but following my logic I believe I've successfully replaced the potions and flasks. I tested it out with the ring enchants and it no longer showed me suggestions while wearing the 60 stat ones.


leozhekov:
I have no idea how to fix these, sorry. I haven't touched the warlock combatlogparser..

PR #3212, opened 2019-07-15
Adjust FoodChecker tiers

niseko:
Secondary stat food is better for most; if not all specs than the feast, so it shouldn't be regarded as mid-tier food.


leozhekov:
I wholeheartedly agree with you but decided against including them in the high tier array, because I am not knowledgeable enough for the other classes.

niseko:
I missed the PR or I'd already have said something in there

PR #3213, opened 2019-07-18
[MW] WotC infographic

abelito75:
WotC infographic that displays HPCT and HPS for each spell in wotc

wotcinfographic


emallson:
image

@abelito75 ur PR is shit

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+  inWotc = false;
+  wotcTime = 0;
+
+  constructor(...args) {
+    super(...args);
+    const hasEssence = this.selectedCombatant.hasEssence(SPELLS.CONFLICT.traitId);
+    if(!hasEssence){
+        return;
+    }
+    this.active = this.selectedCombatant.hasMajor(SPELLS.CONFLICT.traitId);
```
emallson: are both of these checks necessary? (legit not sure)

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+    this.damageSpell = event.ability.name;
+  }
+
+  on_byPlayer_heal(event) {
+    const spellId = event.ability.guid;
+    if(spellId === SPELLS.WAY_OF_THE_CRANE_HEAL.id || spellId === SPELLS.WAY_OF_THE_CRANE_HEAL_HONOR.id){
+      if(this.customMap.get(this.damageSpell)===undefined){
+        this.customMap.set(this.damageSpell, [1, event.amount || 0, event.overheal || 0]);
+      }else{
+        let array = this.customMap.get(this.damageSpell);
```
emallson: instead of using an array for this, you should use an object. It is much easier to read:

```js
this._wotcHealing[this.damageSpell].heal += event.amount;
this._wotcHealing[this.damageSpell].overheal += event.overheal || 0;
```

than what you currently have

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+ * Simple info graphic that shows what spell did what healing during wotc window, the hps of each spell, and hptc
+ */
+
+class WayOfTheCrane extends Analyzer {
+
+  customMap = null;
+  damageSpell = "";
+  lastTimeStamp = 0;
+  inWotc = false;
+  wotcTime = 0;
```
emallson: several of these look like internal values that shouldn't be read or used by other modules. you should prefix those with a `_` so other people know.

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+            arrayOfDamageSpells.map(spell => (
+              <tr>
+              <td>{spell}</td>
+              <td>{formatNumber(this.customMap.get(spell)[1]/this.wotcTime)}</td>
+              <td>{formatNumber(this.customMap.get(spell)[1]/(this.customMap.get(spell)[0]/3*1.5))}</td>
+              </tr>
+            ))
+          }
+          </tbody>
+      </table>
```
emallson: This could use a brief explanation of the HPS/HPCT fields. For example:

```The healing done by your damaging abilities <b>during Way of the Crane</b>. That is: the HPS column doesn't count time outside of WotC.```

File: src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js [hunk 5]
```diff
@@ -0,0 +1,132 @@
...
+          arrayOfDamageSpells.map(spell => (
+            <div>{spell} did {this.customMap.get(spell)[1]} healing, {this.customMap.get(spell)[2]} overhealing in {this.customMap.get(spell)[0]/3} casts.</div>
+          ))
+        )}
+      >
+      <table className="table table-condensed">
+        <thead>
+          <tr>
+            <th>Spell</th>
+            <th>During Wotc HPS</th>
```
emallson: adjust the column name so this doesn't line-wrap. you can use a tooltip or description below the table to give details about the calculation of the column

emallson:
@abelito75 also consider setting the HPCT of `Melee` to 0 or N/A because melees don't have a cast time.

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
abelito75: Pretty sure. The essence object will bug out and die if you try to check if an essence is major if it doesn't exist. And wotc is the conflict (major) part of conflict and strife. So 99% sure

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
abelito75: Only if js was a strongly typed language I could just slap a private on it :( but good idea will do 👍

emallson: yea :(

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
emallson: 👍

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+      this.wotcTime += (this.selectedCombatant.fightDuration - this.lastTimeStamp)/1000;
+    }
+  }
+
+  statistic() {
+    let totalHeal = 0;
+    this.customMap.forEach(function(value, key) {
+      totalHeal += value[1]; 
+    });
+    console.log(totalHeal);
```
Zeboot: you forgot a console.log enabled in here

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    this.active = this.selectedCombatant.hasMajor(SPELLS.CONFLICT.traitId);
+    if (!this.active) {
+      return;
+    }
+    this.customMap = new Map();
+  }
+
+  on_byPlayer_damage(event) {
+    if(!this.selectedCombatant.hasBuff(SPELLS.WAY_OF_THE_CRANE.id)){
```
Zeboot: if you use an eventfilter instead of the old event methods you can avoid having to check for IDs in here which to me seems cleaner

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
+
+import SPELLS from 'common/SPELLS';
+import Analyzer from 'parser/core/Analyzer';
+import StatisticBox from 'interface/others/StatisticBox';
```
Zeboot: probably worth using a different box for this seeing as StatisticBox is deprecated, I don't know how important that deprecation warning is though, maybe using an ItemStatistic? (e.g. voidstone module for the dropdown)

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
abelito75: I'm not sold on eventfilters. Not my cup O'tea

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
abelito75: So happy I wasn't the only know that didn't realize that happened

Zeboot: idk when it happened but I remember seeing it in the StatisticBox source file when I implemented my voidstone module so I added the dropdown functionality to interface/statistic/Statistic (which ItemStatistic uses)

abelito75: Hmm... I wonder which box is correct here. Since AzeritePowerStatistic isn't fully accurate but is closer than item imo 🤔

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -31,33 +33,34 @@ class WayOfTheCrane extends Analyzer {
     if (!this.active) {
       return;
     }
+    this._gcd = (1500 / (1 + this.statTracker.hastePercentage(this.statTracker.currentHasteRating)))/1000;
+    this._gcd = Math.max(this._gcd, .75);
```
emallson: i'm not sure if something to track the gcd exists or not, but this should definitely be a getter because your gcd can change over the course of the fight. for example:

```js
get _gcd() {
    const gcd = 1.5 / (1 + this.statTracker.currentHastePercentage);
    return Math.max(gcd, 0.75);
}
```

Then it will be accurate each time you access it

src/parser/monk/mistweaver/modules/spells/essences/WayOfTheCrane.js
```diff
@@ -31,33 +33,34 @@ class WayOfTheCrane extends Analyzer {
...
-        if(event.overheal){
-          overheal += event.overheal;
-        }
-        array = [array[0] + 1, heal, overheal];
-        this.customMap.set(this.damageSpell, array);
+        const healingEvents = this.customMap.get(this._damageSpell);
+        healingEvents.casts +=1;
+        healingEvents.healing += event.amount;
+        healingEvents.overheal += event.overheal||0;
+        this.customMap.set(this._damageSpell, healingEvents);
```
emallson: much improved :+1:

abelito75:
DO NO MERGE YET. I Just found a bug that breaks wowA for mws if they don't have wotc.

PR #3215, opened 2019-07-19
[MW] Mana tea Rewrite

abelito75:
Made mana tea to be dynamic and not need maintenance ever again :feelsgood: 
Please look decently close to make sure there isn't anything stupid in here.... I made this a bit late at night and don't know if I'm thinking correctly


emallson:
code looks OK, just not sure if there is some in-game reason to not accept this. a lot of effects like this use a whitelist, but i'm not even sure where you'd find an effect that costs mana but wouldn't be a class ability?

abelito75:
I don't think there is but I'd rather not have to come back to a file if I can make it in a generic way that works. The big reason for rewriting it was because wotc was added to our pve abilities and with the old code it was being read as a non-mana costing spell which is SUPER wrong. The way it's build will also allow us to see what spells are all casted during the mana tea window so we might be able to add checks to make sure you are casting mana costing spells during in the window

PR #3226, opened 2019-07-20
[BrM] Updated hit-tracking blacklist and mitigation constants for EP

emallson:
So many fucking dots and raid AoEs this tier >.<


emallson:
abusing executive privilege

PR #3231, opened 2019-07-21
Multi phase instances

Zeboot:
Allows for a phase to appear "multiple times" without having to define multiple instances in the boss info (while still allowing the dropdown to work properly).

Passing the keyword "multiple" in a phase info allows the selector to display a number behind the phase name and makes the filter ignore "eventInstance", therefore allowing it to occur multiple times.

The Phase Parser assigns each phase an array of start timestamps and an array of end timestamps for each instance to preserve each phase instance having the same "key" and other phase information.

The Phase Selector then builds an internal dictionary mapping a string of key + instance to the actual phase information (actual key, instance, name, etc) that is then passed to the phase filter. This allows the phase key passed in the fight object / elsewhere to remain e.g. "P1" instead of "P1_1", therefore allowing easy access to information about the phase (name, etc) from other modules without having to find for example a "separator token" like an underscore to find the actual phase name.

The fight object is appened an "instance" attribute alongside the "phase" attribute if an instance is selected to give the modules information about this.

Examples:

image
image


Sharrq:
Awesome! That was one of the annoyances i ran into when i was adding the phases to the configs.

Zeboot:
Yeah I was setting up the palace information on a local branch earlier this week too and ran into that exact issue so I never pushed that and instead figured I'd wait for the weekend and fix it then 👍

Sharrq:
The biggest thing i ran into was there are phase transitions that dont involve a buff or a cast and i had no good way of identifying the correct time, so i had to pick the closest thing, which means some phase start times are 10-15 seconds off depending on the phase.

For example ....

Blackwater Behemoth i had to mark the phase start when he starts casting Cavatation cause i cant see when it is interrupted.
Zaqul i had to use the first time he casts Dark Passage

Azshara was probably the most difficult one. I can see the start of the first Intemission because Azshara casts "Queen's Decree", but she doesnt cast it again in the second Intermission. And there is nothing else that she does during that intermission that i can use since i cant get her health %. Phase 3 was hard also cause i wanted to use when Lady Ashtongue was spawned, but i tried using what was in the Zekvoz config from Uldir and i could not get it to work. After a couple hours i gave up and used the first time they cast Drain Ancient Ward, which i think is like 15-20 seconds into the phase.

Zeboot:
Yeah there's a couple types of transitions we don't have proper support for yet, it's something that's def on my radar since phase information is obviously more important now, I'll be looking into it to try and make our phase transitions more accurate

Sharrq:
interrupt would be huge.

I dont know much about WCL queries, but looking at their guide on it there seemed to be one for boss health, which would be amazing ... but idk how that would actually work or if its even possible.

PR #3235, opened 2019-07-23
SUPER IMPORT Bug Fix

abelito75:
If you didn't have mana tea as a talent this map would never get made so while going through suggestion threshholds it would go into "RenewingMistDurationManaTea" which depends on mana tea's cast variable to not be null

This is strange because both modules are disabled but w/e I fixed it


emallson:
this seems like it might need to be addressed in the core instead of here, but the fix seems ok so i'll merge it so the mw module actually works

PR #3238, opened 2019-07-25
Quick Fix for TheEverRisingTideCrash

HawkCorrigan:


HawkCorrigan:
@niseko 
The offending log https://wowanalyzer.com/report/vNQyRq6Zbpmgjh4A/14-Mythic+Lady+Ashvane+-+Kill+(7:57)/Jovepriest

Looped permanently when the array was not empty after taking out the EverRisingTide Events and everything between them.

I think this fix won't break anything but better you check it properly.

TJMoats:
@HawkCorrigan looks like the haste portion of this module is now just showing "0"

Issue #2174, opened 2018-09-11
Show a "Degraded experience" toaster when a module is disabled due to an error


MartijnHols:

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=63291570)](https://www.bountysource.com/issues/63291570-show-a-degraded-experience-toaster-when-a-module-is-disabled-due-to-an-error)

MartijnHols:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=63291570)](https://www.bountysource.com/issues/63291570-show-a-degraded-experience-toaster-when-a-module-is-disabled-due-to-an-error)

MartijnHols:
Would be nice to have this in the `next` branch.

MartijnHols:
Updated to $100: [![Bountysource](https://api.bountysource.com/badge/issue?issue_id=63291570)](https://www.bountysource.com/issues/63291570-show-a-degraded-experience-toaster-when-a-module-is-disabled-due-to-an-error)

buimichael:
@MartijnHols Could you explain in detail how this should be done? What should be the trigger of the toaster? Should it be generally per spec or module? 

What do you have in mind?

MartijnHols:
The parser has error handling so that if a module has an error it gets caught silently and the module and all modules that depend on it get disabled. If this happens for anything I’d like there to be a single small info box on the page that explains the degraded experience and perhaps shows a list of disabled modules (since this could be lengthy this should probably be hidden behind a click).

Zeboot:
@buimichael are you working on this one? If not I'll throw something together but if you're already working on it / gonna do it no worries

buimichael:
@zeboot nope, go ahead :)

Remixt:
I'm working on this right now, but I'm new to the project.

Sharrq:
There is already an open PR for this, so i wouldnt start anything. Likely to be closed shortly.

MartijnHols:
This has been implemented

PR #3248, opened 2019-07-31
Degraded experience toaster

Zeboot:
#2174 

Adds a warning about a degraded experience if a module is disabled during loading. Also disables any modules depending on this module. Catches errors during initialization of modules (during constructor), event handling, as well as result generation (and new types can be added easily). Outside of production these errors don't get suppressed / modules stay enabled and throw their errors to allow debugging during development.

image

image

### DIFF
--- a/src/interface/report/Results/DegradedExperience.js
+++ b/src/interface/report/Results/DegradedExperience.js
@@ -27,6 +27,7 @@ class DegradedExperience extends React.PureComponent {
         <Danger style={{ marginBottom: 30 }}>
         <h2> Degraded Experience </h2>
           One or more modules have encountered an error and have been disabled along with modules depending on them. Results may be inaccurate and / or incomplete. <br />
+          Please report this issue to us on <a href="https://wowanalyzer.com/discord">Discord</a> or <a href="https://github.com/WoWAnalyzer/WoWAnalyzer">GitHub</a> so we can fix it! <br />
           {this.state.expanded && (
             <><br />The following modules have been disabled due to errors: <br />
               <div style={{color: "white"}}>
--- a/src/interface/report/Results/DegradedExperience.js
+++ b/src/interface/report/Results/DegradedExperience.js
@@ -38,7 +38,7 @@ class DegradedExperience extends React.PureComponent {
               <div key={state}>
                 <br />The following modules have been disabled due to errors during {state}: <br />
                 <div style={{color: "white"}}>
-                  {disabledModules[state].sort((a,b) => a.name.localeCompare(b.name)).map((m, i) => <span key={i}>{m.name}<br /></span>)}
+                  {disabledModules[state].sort((a,b) => a.name.localeCompare(b.name)).map((m, i) => <React.Fragment key={i}>{m.name}<br /></React.Fragment>)}
                 </div>
               </div>
             );
--- a/src/interface/report/Results/DegradedExperience.js
+++ b/src/interface/report/Results/DegradedExperience.js
@@ -6,7 +6,7 @@ import { DISABLED_STATE } from 'parser/core/CombatLogParser';
 
 class DegradedExperience extends React.PureComponent {
   static propTypes = {
-    disabledModules: PropTypes.array.isRequired,
+    disabledModules: PropTypes.object.isRequired,
   };
 
   constructor(...args){
@@ -15,6 +15,10 @@ class DegradedExperience extends React.PureComponent {
       expanded: false,
     };
     this.toggleDetails = this.toggleDetails.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps){
+    this.forceUpdate();
   }
 
   toggleDetails() {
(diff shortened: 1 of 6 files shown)


Sharrq:
Might be worth adding a line to mention that they can report the issue on Discord so it can get fixed.

Zeboot:
oh yeah completely forgot about that, I'll add a discord / github link

Zeboot:
found an issue in this while adding more functionality, I'll be fixing it after raid, don't merge this yet

Zeboot:
should be fixed

Zeboot:
I can definitely add that (already have a function for counting disabled modules across types), issue is just finding which module to list in the actual message / as "primary" error. Could maybe just pick the first entry of the first error type that isn't empty? (since by that point the list isn't sorted so the first entry should be the first error that occured, not a dependency error)

src/interface/report/Results/DegradedExperience.js
```diff
@@ -0,0 +1,54 @@
...
+  constructor(...args){
+    super(...args);
+    this.state = {
+      expanded: false,
+    };
+    this.toggleDetails = this.toggleDetails.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps){
+    this.forceUpdate();
```
MartijnHols: This is both considered bad practice in React and unnecessary, since it receiving props will already trigger a rerender except if shouldComponentUpdate is implemented. As it happens, it is because at the top of this class your extending React.PureComponent rather than React.Component, which automatically does that for you. If you change that to React.Component, shouldComponentUpdate will be cleared and this will rerender every time it gets new props.

Zeboot: oh nice I was struggling with getting it to update like that and couldn't think of an easy fix for that (relatively new to react), I'll implement that real quick!

MartijnHols:
I figured it would be the original module that had an error, which I suppose is probably the first in the list.

Zeboot:
should both be fixed, this is how the box looks like now: 
image

MartijnHols:
Did 10 modules error or is that just wording? 😄

Zeboot:
For testing I just added 
´´´
const a = undefined;
a.test();
´´´

to constructors / event methods / statistics methods, in this case 10 modules errored total (with 2 of them actually throwing errors, the other 8 being dependency errors)

MartijnHols:
Ah ok. I think it's important to distinguish modules erroring and modules just disabled because they depend on it. Wouldn't want people to think things were poorly built.

I also had another idea, would it be possible to include the error message in the "Show more" box? Getting more information could maybe get someone to look into the issue and fix it.

I know I keep expanding the scope. This is already good to go as-is, so feel free to merge it if you don't want to or don't have time to update it.

src/parser/core/CombatLogParser.js
```diff
@@ -562,13 +586,16 @@ class CombatLogParser {
     this.getModule(EventEmitter).addEventListener(...args);
   }
 
-  deepDisable(module) {
+  deepDisable(module, state = MODULE_ERROR.RESULTS) {
     console.error('Disabling', module.constructor.name);
+    if(module.active){ //prevent duplicate entries
```
MartijnHols: Could just early return the function even before the console.error since running deepDisable should be pointless when a module is already disabled.

src/parser/core/CombatLogParser.js
```diff
@@ -562,13 +586,16 @@ class CombatLogParser {
     this.getModule(EventEmitter).addEventListener(...args);
   }
 
-  deepDisable(module) {
+  deepDisable(module, state = MODULE_ERROR.RESULTS) {
```
MartijnHols: Any particular reason why results is the default value? Could this be a required argument instead?

Zeboot:
Hmm yeah I see what you mean. I just changed the wording to show that a lot of the disabled modules are within dependencies and not actual errors, I can update the error tracking real quick to distinguish between module errors and dependency errors.

Regarding the error in the show more box, I can definitely add that but I'm not sure how bloated that could get if multiple modules actually errored, I'll give it a shot tho

src/parser/core/CombatLogParser.js
```diff
@@ -627,67 +652,88 @@ class CombatLogParser {
...
+              }
+            }
+            if (module.suggestions) {
+              module.suggestions(results.suggestions.when, { i18n });
+            }
+          }catch(e){ //error occured during results generation of module, disable module and all modules depending on it
+            if (process.env.NODE_ENV !== 'production') {
+              throw e;
+            }
+            this.deepDisable(module);
```
MartijnHols: MODULE_ERROR.RESULTS?

src/parser/core/CombatLogParser.js
Zeboot: oh yeah this is a remnant from when I didn't catch event handling yet so the deepDisable method was only used for results errors, I'll move it to a required argument

Zeboot:
This is how it looks now:
image
image
image

let me know if the error display is too much, might just limit it to the first line of the error instead of the full stacktrace, that should provide all the necessary information

MartijnHols:
thanks!

PR #3257, opened 2019-08-04
updated news header to stay within the "container" on mobile

GrantJoshua1995:
Previously the header for news and the twitter link was on the right (which looks good on desktop) on mobile (making it go off screen, and as such causing a horizontal scrollbar).

I have now updated this header and the twitter link to stay within the container, so that it doesn't cause scrolling to the right on smaller devices. Which should improve responsiveness on the news section when viewing in mobile.

### DIFF
--- a/src/articles/2018-08-03-DevotionAura/index.js
+++ b/src/articles/2018-08-03-DevotionAura/index.js
@@ -97,7 +97,7 @@ export default (
 
     The methods used to get you a DRPS value for Devotion Aura on WoWAnalyzer gives us very accurate results. The methods are described in this article so that anyone can review and verify them. If you have any feedback let me know on Discord; @Zerotorescue.<br /><br />
 
-    More info on Devotion Aura: <a class="overflow-link" href="https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md">https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md</a><br />
-    More info on all auras: <a class="overflow-link" href="https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/">https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/</a>
+    More info on Devotion Aura: <a className="overflow-link" href="https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md">https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md</a><br />
+    More info on all auras: <a className="overflow-link" href="https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/">https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/</a>
   </RegularArticle>
 );
--- a/src/articles/2018-08-03-DevotionAura/index.js
+++ b/src/articles/2018-08-03-DevotionAura/index.js
@@ -97,7 +97,7 @@ export default (
 
     The methods used to get you a DRPS value for Devotion Aura on WoWAnalyzer gives us very accurate results. The methods are described in this article so that anyone can review and verify them. If you have any feedback let me know on Discord; @Zerotorescue.<br /><br />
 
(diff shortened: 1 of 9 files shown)


GrantJoshua1995:
Hey @MartijnHols  , my mistake didn't mean to commit that! Should hopefully be reverted now to what was in master!

GrantJoshua1995:
Got some further code to add to this before it gets pulled into master @MartijnHols . Found another thing which is causing horizontal overflowing on the news section. 
Shall I close this PR and start again, or just push another change to the branch?

(Sorry new to the WoWAnalyzer repo, and want to make sure I'm doing it properly!)

MartijnHols:
You never have to close a PR; if you fix something just push new commits. 

In the case where you’re working on a separate feature please consider using a different branch and making a separate new PR for it though. This makes reviewing a lot easier.

GrantJoshua1995:
Ah that's what I thought, thanks for the heads up. Figured I'd ask just in case it was different. 

I'll get my changes on here later

GrantJoshua1995:
@MartijnHols Fixed all the problems I encountered earlier. Had to reduce some font-sizes etc. However at least from a front page perspective, it looks a lot better now I think!

src/articles/2018-08-03-DevotionAura/index.js
```diff
@@ -97,7 +97,7 @@ export default (
 
     The methods used to get you a DRPS value for Devotion Aura on WoWAnalyzer gives us very accurate results. The methods are described in this article so that anyone can review and verify them. If you have any feedback let me know on Discord; @Zerotorescue.<br /><br />
 
-    More info on Devotion Aura: <a href="https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md">https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md</a><br />
-    More info on all auras: <a href="https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/">https://sacredshielding.wordpress.com/2018/07/23/reading-the-color-of-your-aura-in-battle-for-azeroth/</a>
+    More info on Devotion Aura: <a class="overflow-link" href="https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md">https://github.com/MartijnHols/HolyPaladin/blob/master/Spells/Talents/60/DevotionAura.md</a><br />
```
MartijnHols: Because `class` is a reserved word in JavaScript, React complains about using the `class` attribute. In React please use the `className` attribute instead.

src/interface/news/index.js
```diff
@@ -10,14 +10,14 @@ const News = () => (
...
+    <div className="flex flex-news">
       <div className="flex-main">
         <h1 id="news-top"><Trans>New stuff</Trans></h1>
       </div>
-      <div className="flex-sub" style={{ paddingTop: 30 }}>
+      <div className="flex-sub flex-sub-news">
         <small>More news?</small>
         <span style={{ fontSize: 18, marginLeft: 10 }}>
-          <TwitterIcon colored /> <a href="https://twitter.com/WoWAnalyzer">Follow @WoWAnalyzer</a>
+          <TwitterIcon colored /> <a href="https://twitter.com/WoWAnalyzer" style={{fontSize: '16px'}}>Follow @WoWAnalyzer</a>
```
MartijnHols:
```suggestion
          <TwitterIcon colored /> <a href="https://twitter.com/WoWAnalyzer" style={{ fontSize: 16 }}>Follow @WoWAnalyzer</a>
```

In React it's best practice to use an integer for such properties. I don't remember if React will actually throwing a warning for this though.

Will this change the size of the text?

src/articles/2018-08-03-DevotionAura/index.js
GrantJoshua1995: Ah of course, silly error! I primarily work with React, you'd think I wouldn't make these kind of errors...  Fixed it now.

src/interface/news/index.js
GrantJoshua1995: > In React it's best practice to use an integer for such properties. I don't remember if React will actually throwing a warning for this though.
> 
> Will this change the size of the text?

Yes it will. Currently I I saw it sitting at 18px, which as you get to smaller devices, the link starts to wrap. 
I thought the better solution to this would be having it just being 16px all the time as even on desktop that still looks good to me! 
Happy to put this in a media query instead if you'd rather 18px at larger resolutions.

MartijnHols: 16px sounds good

PR #3267, opened 2019-08-09
Add bad Furious Gaze statistics

viridis:
- Fix calculation for Furious Gaze count
- Add calculations for bad Furious Gaze count
- Add tooltip to explain bad Furious Gaze

image

image


src/parser/demonhunter/havoc/modules/spells/azeritetraits/FuriousGaze.js
```diff
@@ -47,11 +52,21 @@ class FuriousGaze extends Analyzer{
...
-    this.furiousGazeProcsCounter += 1;
+  onFuriousGazeApply(event) {
+    if (this.furiousGazeStartTime === undefined) {
+      this.furiousGazeStartTime = event.timestamp;
+    }
+  }
+
+  onFuriousGazeRemove(event) {
+    if (this.furiousGazeStartTime !== undefined) {
+      this.furiousGazeProcsCounter += Math.round((event.timestamp - this.furiousGazeStartTime) / FURIOUS_GAZE_DURATION);
```
MartijnHols: I don't really understand what is going on. Doesn't the buff not get applied at all if you didn't finish the cast? In that case why does the player get partial credit for the cast?

viridis: If the player currently has the buff and manages to get it a second time within the first buff's duration (Going Eyebeam, Meta reset through Chaotic Transformation, Eyebeam) there is no second applybuff event. I have added more decomposition and explanation in a new commit to clarify.

viridis: And the rounding ist just because some events are 11995 or 12012 instead of 12000 (But im assuming you weren't pointing at that?)

PR #3270, opened 2019-08-11
More phase filters

Zeboot:
adding more filters / triggers for phase transitions to make it easier for us to detect phase changes.

New filters so far:
- applydebuff
- interrupt (see blackwater behemoth for an example)
- dispel
- time (creates phase transition at given fight duration in milliseconds, can be used to create repeating phase starts from this point onwards)
- health (creates trigger at first event to the target mob that marks it as past the given health percentage)

Full trigger documentation can be found in common/fabricateBossPhaseEvents.js

Additionally, each trigger can now be passed an optional filter.offset parameter that shifts the phase start to the time of the trigger event plus the offset in milliseconds (can be negative).

If anyone knows of a better / easier way to track health that I just completely missed or has other trigger types they'd like supported too, please let me know.


Zeboot:
found a cleaner way to find the health percentage by looking at only the first matched event in the WCL filter so we only get one event per phase filter, not multiple hundred depending on accuracy

Sharrq:
oooo. ill update some of the phases once this goes through. Will make it a lot easier.

PR #3276, opened 2019-08-15
lava burst check's for flame shock on damage instead of on cast

MikeCook9994:
8.1 introduced an undocumented change so that lava burst checks if the target has flame shock on damage instead of cast to guarantee the crit.

@HawkCorrigan

### DIFF
--- a/src/CONTRIBUTORS.js
+++ b/src/CONTRIBUTORS.js
@@ -717,11 +717,11 @@ export const Draenal = {
   mains: [{
     name: 'Draenal',
     spec: SPECS.ELEMENTAL_SHAMAN,
-    link: 'https://worldofwarcraft.com/en-us/character/us/malganis/draenal'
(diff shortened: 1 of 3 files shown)


src/CONTRIBUTORS.js
```diff
@@ -711,3 +711,17 @@ export const Wing5wong = {
...
+  mains: [{
+    name: 'Draenal',
+    spec: SPECS.ELEMENTAL_SHAMAN,
+    link: 'https://worldofwarcraft.com/en-us/character/us/malganis/draenal'
+  },
+  {
+    name: 'MagicEraser',
+    spec: SPECS.FROST_MAGE,
+    link: 'https://worldofwarcraft.com/en-us/character/us/malganis/magiceraser'
+  }]
```
abelito75: add a comma after the ] and add a ; after the last } in the file

PR #3279, opened 2019-08-15
Fix damage calculation for Streaking Stars.

viridis:
Damage calculations were naively done by multiplying occurences by amount of
damage the trait gives. This does not take into consideration absorb
shields, crits, damage boosts, ... etc. Now using the ability tracker to
get the correct values (which now also perfectly match the WCL logs
unlike before)

@Gebuz

### DIFF
--- a/src/common/SPELLS/bfa/azeritetraits/druid.js
+++ b/src/common/SPELLS/bfa/azeritetraits/druid.js
@@ -156,6 +156,11 @@ export default {
     name: 'Streaking Stars',
     icon: 'spell_nature_natureguardian',
   },
+  STREAKING_STAR: {
+    id: 272873,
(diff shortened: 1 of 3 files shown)


Gebuz:
Thanks! I would suggest also using the new value in line 108 (X damage per proc of streaking stars) to using the new value/number of hits, and adding "average" to the sentence. Something like this:
`{formatNumber(totalDamage / (this.totalCastsDuringCelestialAlignment - this.badCasts))} average damage per proc of <SpellLink id={SPELLS.STREAKING_STARS.id} /><br />`

viridis:
@Gebuz Sorry about that, completely forgot about that.

Gebuz:
Thanks!

PR #3280, opened 2019-08-16
Add timefilter shortcut to Death Recap

Zeboot:
idea by @abelito75 

Adds a button to the death recap that lets you filter the events to only include events prior to the death  to make suggestions / analysis more accurate using the existing time filtering functionality. In the case of multiple deaths it filters events to the time between the deaths.

Also attaches the filtering functionality (both time and phase based) to the combatlogparser object so other modules can make use of this functionality

Examples:
image

image
image

image

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -8,6 +8,7 @@ import { change, date } from 'common/changelog';
 import Contributor from 'interface/contributor/Button';
 
 export default [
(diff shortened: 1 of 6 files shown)


Gebuz:
Would be sweet to have this for the cooldown tab aswell!

Zeboot:
That's definitely something I can easily add in another PR after this gets merged (using the new hooks added in this PR), I'd like to avoid adding it to this one directly as to not have multiple features in one PR

PR #3282, opened 2019-08-16
Add Heart of Azeroth to Character Info panel

viridis:
- List large active essence
- List smaller passive essences
- Show rank through appropriate border color

image


viridis:
Sidenote, I used the code from Azerite Traits to 'load missing traits' and changed it to what I THINK is what it needs to do. I'm not entirely sure on how it works or whether it even works in Development tbh. Initially i was getting all icons, and then after a while some started disappearing and became question marks. If this is indeed something that should also work perfectly on Development, please tell me how i fix it (if it is actually broken xD)

Issue #2612, opened 2018-11-07
[SPriest][BFA] Azerite Traits for Shadow


TJMoats:
- [x] Chorus of Insanity
- [x] Death Throes 
- [ ] Depth of the Shadows
- [ ] Searing Dialogue
- [ ] Spiteful Apparitions 
- [ ] Thought Harvester 
- [ ] Whispers of the Damned

Pewtro:
Goodbye BFA, Goodbye Azerite.

PR #3285, opened 2019-08-17
[Shadow Priest] Add Whispers of the Damned azerite trait

DylanDirlam:
- Tracks additional damage gained
- Tracks additional insanity generated from critical strikes
On the checklist of #2612

### DIFF
--- a/src/parser/priest/shadow/modules/spells/azeritetraits/WhispersOfTheDamned.js
+++ b/src/parser/priest/shadow/modules/spells/azeritetraits/WhispersOfTheDamned.js
@@ -36,10 +36,10 @@ class WhispersOfTheDamned extends Analyzer {
   constructor(...args) {
     super(...args);
     this.active = this.selectedCombatant.hasTrait(SPELLS.WHISPERS_OF_THE_DAMNED.id);
-    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(MIND_BLAST_SPELLS), this.onDamageEvent);
     if (!this.active) {
       return;
     }
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(MIND_BLAST_SPELLS), this.onDamageEvent);
 
     const { damage } = whispersOfTheDamnedStats(this.selectedCombatant.traitsBySpellId[SPELLS.WHISPERS_OF_THE_DAMNED.id]);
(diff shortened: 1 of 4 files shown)


src/parser/priest/shadow/modules/spells/azeritetraits/WhispersOfTheDamned.js
```diff
@@ -0,0 +1,73 @@
...
+
+    const { damage } = whispersOfTheDamnedStats(this.selectedCombatant.traitsBySpellId[SPELLS.WHISPERS_OF_THE_DAMNED.id]);
+    this.damageValue = damage;
+  }
+
+  onDamageEvent(event) {
+    this.damageDone += this.damageValue;
+
+    if (event.hitType === HIT_TYPES.CRIT) {
+      this.insanityGained += 20;
```
HawkCorrigan: This will be misleading if the spell overcaps.

src/parser/priest/shadow/modules/spells/azeritetraits/WhispersOfTheDamned.js
```diff
@@ -0,0 +1,73 @@
...
+class WhispersOfTheDamned extends Analyzer {
+  damageValue = 0;
+  damageDone = 0;
+  insanityGained = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTrait(SPELLS.WHISPERS_OF_THE_DAMNED.id);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(MIND_BLAST_SPELLS), this.onDamageEvent);
+    if (!this.active) {
```
HawkCorrigan: iirc there is something that would trigger the early returns anyway if active is set to false, so the eventListener would be created but not executed but for the sake of clarity I would switch the early return with the addEventListener.

src/parser/priest/shadow/modules/spells/azeritetraits/WhispersOfTheDamned.js
DylanDirlam: Updated to use an energize event and `resourceChange` to get exactly how much insanity was gained from the trait procing

HawkCorrigan:
Thanks

PR #3288, opened 2019-08-18
[Shadow Priest] Add Spiteful Apparitions azerite trait

DylanDirlam:
- Tracks additional damage from Spiteful Apparitions done by Shadowy Apparitions to targets with Vampiric Touch
Completes part of #2612

### DIFF
--- a/src/parser/priest/shadow/modules/spells/azeritetraits/SpitefulApparitions.js
+++ b/src/parser/priest/shadow/modules/spells/azeritetraits/SpitefulApparitions.js
@@ -30,7 +30,6 @@ class SpitefulApparitions extends Analyzer {
 
   damageValue = 0;
   damageDone = 0;
-  wastedApparitions = 0;
 
   constructor(...args) {
     super(...args);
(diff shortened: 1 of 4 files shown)


src/parser/priest/shadow/modules/spells/azeritetraits/SpitefulApparitions.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(SPELLS.SHADOWY_APPARITION_DAMAGE), this.onDamageEvent);
+
+    const { damage } = spitefulApparitionsStats(this.selectedCombatant.traitsBySpellId[SPELLS.SPITEFUL_APPARITIONS.id]);
+    this.damageValue = damage;
+  }
+
+  onDamageEvent(event) {
+    const enemy = this.enemies.getEntity(event);
+    if (!enemy) {
```
Sharrq: its minor, but i think these can be combined

if (!enemy || !enemy.hasBuff(SPELLS.VAMPIRIC_TOUCH.id, event.timestamp)) {
    return;
}

DylanDirlam:
I have updated the module to use `calculateBonusAzeriteDamage` instead of just adding up the damage.

PR #3297, opened 2019-08-21
[Fire Mage] Lucid Dreams Support & CLF Haste Buff

Sharrq:
Adds Fire Mage support for Lucid Dreams Major and Minor
Adds Spell IDs for Condensed Life Force
Adds the stacking Haste buff from Condensed Life Force
Adjusted Hyperthread Wristwraps to give double the reduction during Lucid Dreams to compensate for the doubled cool down regen from Lucid.

### DIFF
--- a/src/parser/shared/modules/Haste.js
+++ b/src/parser/shared/modules/Haste.js
@@ -35,7 +35,9 @@ class Haste extends Analyzer {
     [SPELLS.QUICK_THINKER_BUFF.id]: 0.15,
     [SPELLS.EMPOWER_RUNE_WEAPON.id]: 0.15, // Frost DK
     [SPELLS.EVER_RISING_TIDE_CHARGING_BUFF.id]: 0.1, // Essence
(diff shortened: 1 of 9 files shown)


src/parser/shared/modules/Haste.js
```diff
@@ -35,6 +35,7 @@ class Haste extends Analyzer {
     [SPELLS.QUICK_THINKER_BUFF.id]: 0.15,
     [SPELLS.EMPOWER_RUNE_WEAPON.id]: 0.15, // Frost DK
     [SPELLS.EVER_RISING_TIDE_CHARGING_BUFF.id]: 0.1, // Essence
+    [SPELLS.GUARDIAN_OF_AZEROTH_HASTE_BUFF.id]: 0.02, // Essence
```
Gebuz: This is per stack right? So it should be 
```
[SPELLS.GUARDIAN_OF_AZEROTH_HASTE_BUFF]: {
  hastePerStack: 0.02
}
```

Sharrq: yep, thanks. I didnt know how to add a stacking haste buff so i added it like that temporarily.

PR #3311, opened 2019-09-05
[Mana Efficiency] Added time spent column

niseko:
grafik


src/parser/core/healingEfficiency/HealingEfficiencyBreakdown.js
```diff
@@ -170,6 +171,7 @@ class HealingEfficiencyBreakdown extends React.Component {
           {formatNumber(spellDetail.manaSpent)}
           {' (' + formatPercentage(spellDetail.manaPercentSpent) + '%)'}
         </td>
+        <td>{spellDetail.timeSpentCasting !== 0 ? (formatNumber(spellDetail.timeSpentCasting / 1000) + 's (' + formatPercentage(spellDetail.percentTimeSpentCasting) + '%)') : '-'}</td>
```
abelito75: Is there a reason you don't use formatDuration or formatMilliseconds?

niseko: both return a format akin to m:ss which i'm not a fan of when its almost always going to be < 60 second, but i changed it around for consistency sake
image

abelito75: I could see this easily go above 60 for mw on a lot of fights but 👍 Looks good to me now

PR #3318, opened 2019-09-09
[Prot warrior] Last update till release maybe

abelito75:
Threshold changes as the old ones were outdated.
Moved file to a new location as it is more logical.
Brace for Impact infographic changed to m:ss from m:ss.sss
Brace for impact bug fixed

### DIFF
--- a/src/parser/warrior/protection/modules/features/BlockCheck.js
+++ b/src/parser/warrior/protection/modules/features/BlockCheck.js
@@ -9,16 +9,53 @@ import SPELLS from 'common/SPELLS';
 const debug = false;
 
 //Moved from another file as it was easier to keep track of with this name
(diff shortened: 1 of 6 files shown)


src/parser/warrior/protection/modules/features/BlockCheck.js
```diff
@@ -85,4 +118,4 @@ class ShieldBlockMitigation extends Analyzer {
   statisticOrder = STATISTIC_ORDER.CORE(10);
 }
 
-export default ShieldBlockMitigation;
\ No newline at end of file
+export default BlockCheck;
```
emallson: fun fact: you can write `export default class BlockCheck extends Analyzer { .... }` and not have to duplicate the class name

PR #3320, opened 2019-09-12
[Windwalker] Update Spinning Crane Kick

Juko8:
image

Suggestions regarding bad casts were inaccurate and misleading, and have therefore been removed


abelito75:
Is average hits the average damage tics ticks or average number of mobs hit?

Juko8:
It counts all ticks @abelito75

abelito75:
Sounds good to be

emallson:
@abelito75 if you're gonna review actually make a review :anger:

abelito75:
It was a question >:(

PR #3324, opened 2019-09-17
[DPriest] Add additional metrics to Purge the Wicked

TJMoats:
Old:
Screen+Shot+2019-09-17+at+8 50 09+AM

New:
Screen+Shot+2019-09-17+at+8 44 14+AM


src/parser/priest/discipline/modules/features/PurgeTheWicked.js
```diff
@@ -25,12 +28,41 @@ class PurgeTheWicked extends Analyzer {
...
   get uptime() {
     return this.enemies.getBuffUptime(this.dotSpell.id) / this.owner.fightDuration;
   }
 
+  get extraPTWs() {
+    return this.ptwApplications - this.ptwCasts;
+  }
+
+  get ptwDamage() {
+    return this.abilityTracker.getAbility(SPELLS.PURGE_THE_WICKED_BUFF.id).damageEffective;
```
karlpralow: I think you're not accounting for dying ennemies wich happens pretty often for PtW spreads. You're implying that all PtW spreads will always have their full duration wich is often not the case when it spreads to adds that are cleaved down pretty fast (Radiance, Orgozoa, Zacool). I think it would be safer to collect debuff applications that are after a penance cast and sum that damage of the events to that target before debuff fades.

karlpralow:
This is a great addition!
I would maybe but the DPS line as a tooltip to the extra ticks because it might be confusing to some players that you're saying that PtW talent has provided x damage or that the extra ticks provided x damage. If it's 0, we don't want people thinking PtW vs SW:P provided 0 damage benefit, only that the spread mechanic provided 0 damage.

PR #3325, opened 2019-09-18
[Prot Warrior] Block Checker Update

abelito75:
Now only checks blockable hits (retroactive)
Before:
before

After:
after

This will drop the stat a little bit but make it more accurate because before it only checked if it was physical and you had a block spell up so things that were aoe that were not blockable but happened while you had Shield block up counted as a good thing even thought it didn't matter

### DIFF
--- a/src/parser/warrior/protection/modules/features/BlockCheck.js
+++ b/src/parser/warrior/protection/modules/features/BlockCheck.js
@@ -6,14 +6,22 @@ import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
 import Analyzer from 'parser/core/Analyzer';
 import SPELLS from 'common/SPELLS';
 
+import ShieldBlock from '../spells/ShieldBlock';
+
 const debug = false;
 
 //Moved from another file as it was easier to keep track of with this name
 class BlockCheck extends Analyzer {
-  physicalHitsWithShieldBlock = 0;
-  physicalDamageWithShieldBlock = 0;
-  physicalHitsWithoutShieldBlock = 0;
-  physicalDamageWithoutShieldBlock = 0;
+  static dependencies = {
+    shieldBlock: ShieldBlock,
+  };
+
+  physicalHitsWithBlock = 0;
+  physicalDamageWithBlock = 0;
+  physicalHitsWithoutBlock = 0;
+  physicalDamageWithoutBlock = 0;
+
+  listOfEvents;
 
   bolster = this.selectedCombatant.hasTalent(SPELLS.BOLSTER_TALENT.id);
   heavyRepercussions = this.selectedCombatant.hasTalent(SPELLS.HEAVY_REPERCUSSIONS_TALENT.id);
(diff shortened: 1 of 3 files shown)


src/parser/warrior/protection/modules/features/BlockCheck.js
```diff
@@ -55,29 +64,48 @@ class BlockCheck extends Analyzer {
...
+      };
+      this.listOfEvents.push(event);
     }
   }
 
   on_fightend() {
+    const blockableSet = new Set();//this is master list of all BLOCKED events in the fight
+    blockableSet.add(1);//make it so if they never hit sb we still get data from the melees they take
+    this.shieldBlock.shieldBlocksDefensive.forEach(function(block){
+      block.eventName.forEach(function(blockedAbility){
```
emallson: shouldn't this be `block.eventSpellId`?

src/parser/warrior/protection/modules/features/BlockCheck.js
```diff
@@ -55,29 +64,48 @@ class BlockCheck extends Analyzer {
...
+      console.log(`Hits without block spell up ${this.physicalHitsWithoutBlock}`);
+      console.log(`Damage without block spell up ${this.physicalDamageWithoutBlock}`);
+      console.log(`Total physical ${this.physicalDamageWithoutBlock}${this.physicalDamageWithBlock}`);
     }
   }
 
   get suggestionThresholds() {//was in here before but is/was never used and appears to be very high requirements that are unreasonable maybe lower and add laster?
     return {
-      actual: this.physicalDamageWithShieldBlock / (this.physicalDamageWithShieldBlock + this.physicalDamageWithoutShieldBlock),
+      actual: this.physicalDamageWithBlock / (this.physicalDamageWithBlock + this.physicalDamageWithoutBlock),
```
emallson: this should probably just be % of hits blocked, with special handling for blockable tankbusters. i get the idea (threshold on % of damage where you had block up), but a) this is currently incorrect (to do that correctly it should be `sum of raw damage from blocked hits / sum of raw damage`), and b) its possible to do well on that purely by luck even with bad SB uptime.

src/parser/warrior/protection/modules/features/BlockCheck.js
```diff
@@ -94,8 +122,8 @@ class BlockCheck extends Analyzer {
   }
 
   statistic() {
-    const physicalHitsMitigatedPercent = this.physicalHitsWithShieldBlock / (this.physicalHitsWithShieldBlock + this.physicalHitsWithoutShieldBlock);
-    const physicalDamageMitigatedPercent = this.physicalDamageWithShieldBlock / (this.physicalDamageWithShieldBlock + this.physicalDamageWithoutShieldBlock);
+    const physicalHitsMitigatedPercent = this.physicalHitsWithBlock / (this.physicalHitsWithBlock + this.physicalHitsWithoutBlock);
+    const physicalDamageMitigatedPercent = this.physicalDamageWithBlock / (this.physicalDamageWithBlock + this.physicalDamageWithoutBlock);
```
emallson: this should be `sum of damage blocked / sum of post-mit damage`

src/parser/warrior/protection/modules/features/BlockCheck.js
abelito75: Yes. I added eventSpellId this pr and forgot to update

PR #3328, opened 2019-09-21
Issue-2533 - added dire consequences module.

leozhekov:


Juko8:
I felt a bit sceptical about this %haste being additive, so i looked into it. Seems like it is indeed multiplicative from what is currently implemented in other modules. Beastmastery also already has a module with the Dire Beast talent, https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/hunter/beastmastery/modules/talents/DireBeast.js, consider expanding that or make this more similar to it (or both).

Juko8:
and i almost forgot to mention it, but `statTracker.hasteRatingPerPercent` should indeed be used to make any stat conversions between ratings and percentages. Having duplicate constants in different places will be a strain on maintainability.

PR #3329, opened 2019-09-24
Channeling trinkets update

yajinni:
Noticed some times that were usable by mutiple classes werent having thier channeling taking into account when calculating down time and being shown on the timeline.

Attempted to fix this problem. Not sure if this was the best solution.
Originally i attempted to add the code to the trinket item itself and couldnt figure out a way to make it work. So decided to mimic how each class has a channeling for its own spec and made a shared one figuring there wouldnt be many items like this.


Sharrq:
if you update the snapshots then i think its fine

PR #3340, opened 2019-10-01
[Core] Cast Efficiency Fix

Sharrq:
Cast Efficiency was showing > 100% for the Time on Cooldown for a lot of spells.
This was because when calculating the Time on CD, it was essentially doing (SpellCooldown * SpellCasts / FightDuration).

So the remaining cooldown of the spell beyond the end of the fight was counting towards the time on cooldown.

image

In the above example, Combustion was cast 3 times (120 second cooldown each) and the fight was 299 seconds long. 
So it was doing 360/299= 120%

Instead it should have been 2 casts for 120s each and a third for 33s to total 273/299

image

I am not sure if this is the best way to fix it or not, but its the only way i could make it work.


Sharrq:
Dont merge this. I just noticed @emallson has a PR in with a Cast Efficiency fix in it and i want to wait until that is live to see if this is still needed or not.

emallson:
@Sharrq my fix is related to abilities with multiple charges and cooldown resets. it *shouldn't* change any other spells (though it might conflict with your PR)

Sharrq:
Yea, i just want to make sure mine still works right after yours.

Sharrq:
Still looks the same after your changes, so i think its ok.

PR #3342, opened 2019-10-02
[MW] Tier 45 Comparison tool

abelito75:
Takes their play style and allows them to compare talents on this tier. This can be kind of skewed due to the fact if you don't take a certain talent then you won't try to min max it but /shrug

new


abelito75:
oops added prot warrior changes here let me remove those

PR #3353, opened 2019-10-12
We missed some pots

abelito75:
Including Empowered Proximity for potion checker


src/parser/shared/modules/items/CombatPotion.js
```diff
@@ -25,6 +25,7 @@ class CombatPotion extends Potion {
     SPELLS.POTION_OF_REPLENISHMENT,
     SPELLS.POTION_OF_WILD_MENDING,
     SPELLS.POTION_OF_FOCUSED_RESOLVE,
+    SPELLS.POTION_OF_EMPOWERED_PROXIMITY.id,
```
Juko8: .id is not needed here

abelito75: 👍

PR #3363, opened 2019-10-16
Add Icefury Efficiency checklist item and suggestion

MikeCook9994:
Icefury is currently one of the most popular ele shaman talents at the moment and a lot of players will struggle to actually fully utilize it by using all of the charges before it expires because of how tight it can be to use them all especially when it sync with Stormkeeper (which is often). Many players will try to over optimize with Master of the Elements and ultimately cost themselves DPS.

af2c518361ed7219f9877af705373be5


MikeCook9994:
@HawkCorrigan chance you could take a look at this?

HawkCorrigan:
Looks good
For consistencies sake I would prefer the new style  of event listener but I would be okay with merging it. So your call

PR #3364, opened 2019-10-16
Include precast Stormkeeper in total number of Stormkeeper casts.

myran2:
Currently, the Stormkeeper cast ~5 seconds before pull is not included in the cast efficiency calculation. This results in suggestions that say you missed a cast of Stormkeeper when you really did not.

Here is an example WA report that demonstrates the issue: 
https://wowanalyzer.com/report/rJXM1TmkpKPgN2n7/4-Mythic+Abyssal+Commander+Sivara+-+Kill+(2:50)/Thejigglr/overview
The report says that I only used 2/3 possible Stormkeeper casts.

If you check the actual log (https://www.warcraftlogs.com/reports/rJXM1TmkpKPgN2n7/#fight=4&source=7&type=auras&ability=191634), you can see that I had the Stormkeeper buff 3 times during the fight, so a cast was not recorded.

This change manually tracks Stormkeeper casts, which results in the correct total number of casts being recorded.

### DIFF
--- a/src/parser/shaman/elemental/modules/talents/Stormkeeper.js
+++ b/src/parser/shaman/elemental/modules/talents/Stormkeeper.js
@@ -58,7 +58,6 @@ class Stormkeeper extends Analyzer {
     const stormKeeperAbility = this.abilities.abilities.find(function(element) {
       return element.spell === SPELLS.STORMKEEPER_TALENT;
     });
-    console.log(stormKeeperAbility);
     stormKeeperAbility.castEfficiency.casts = ()=>{
       return this.stormkeeperCasts;
     };
--- a/src/parser/shaman/elemental/modules/talents/Stormkeeper.js
+++ b/src/parser/shaman/elemental/modules/talents/Stormkeeper.js
@@ -6,8 +6,6 @@ import { formatNumber, formatPercentage } from 'common/format';
 
 import Analyzer from 'parser/core/Analyzer';
 
-import Abilities from 'parser/core/modules/Abilities';
-
 import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
 
 const AFFECTED_ABILITIES = [SPELLS.LIGHTNING_BOLT_OVERLOAD.id,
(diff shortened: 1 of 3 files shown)


src/parser/shaman/elemental/modules/talents/Stormkeeper.js
```diff
@@ -40,6 +47,23 @@ class Stormkeeper extends Analyzer {
...
+    if (spellId === SPELLS.STORMKEEPER_TALENT.id) {
+      this.stormkeeperCasts += 1;
+    }
+  }
+
+  on_fightend(event){
+    const stormKeeperAbility = this.abilities.abilities.find(function(element) {
+      return element.spell === SPELLS.STORMKEEPER_TALENT;
+    });
+    console.log(stormKeeperAbility);
```
MikeCook9994: Should remove

myran2: woops

src/parser/shaman/elemental/modules/talents/Stormkeeper.js
```diff
@@ -40,6 +47,22 @@ class Stormkeeper extends Analyzer {
...
   }
 
+  on_byPlayer_applybuff(event){
+    const spellId = event.ability.guid;
+    if (spellId === SPELLS.STORMKEEPER_TALENT.id) {
+      this.stormkeeperCasts += 1;
+    }
+  }
+
+  on_fightend(event){
```
MikeCook9994: Hawk recommended on my PR a different, newer style for event subscription. This is quite nice as it can handle some of the filtering for you.

```javascript
import Events from 'parser/core/Events';
import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';

constructor(...args) {
  this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.STORMKEEPER_TALENT), (event) => { ... });
  this.addEventListener(Events.fightend, (event) => { ... });
}
```

Dambroda:
As I mentioned in discord, you can handle this automatically by adding the proper field (triggeredBySpellId) to the Buffs spec file, rather than handing this yourself :)

myran2:
Thanks for the heads up! I followed your suggestions and everything still seems to be working correctly (and with much less new code 👍).

PR #3373, opened 2019-10-22
[Hpriest]Stat Weights

TJMoats:


src/parser/priest/holy/modules/core/EchoOfLightMastery.js
```diff
@@ -97,6 +99,10 @@ class EchoOfLightMastery extends Analyzer {
       this.targetMasteryPool[targetId].pendingHealingBySpell[spellId] = 0;
     }
     this.targetMasteryPool[targetId].pendingHealingBySpell[spellId] += rawHealing;
+    if (event.hitType === HIT_TYPES.CRIT) {
+      // This is just to track how much of a EoL tick can be contributed to a crit.
```
rp4rk: Definitely great you've left a comment here, I'd make a suggestion to put it in a less passive voice, 
```suggestion
      // Track how much of a EoL tick can be contributed to a crit.
```

Google have an excellent guide to code comments and PRs in general: https://developers.google.com/style/voice

src/parser/priest/holy/modules/core/EchoOfLightMastery.js
```diff
@@ -121,6 +127,13 @@ class EchoOfLightMastery extends Analyzer {
     // The total amount that should be drained from the pool
     const poolDrainTotal = this.targetMasteryPool[targetId].pendingHealingTotal * poolDrainPercent;
```
rp4rk: In this circumstance, is the pool the amount of EoL healing derived from mastery alone? Understand this is not a core concern of the PR, but would be great to have context here.

PR #3379, opened 2019-10-25
Build / Rotation Selection

Zeboot:
Adds a framework to allow build / rotation selection (e.g. no-icelance) on a per spec basis.
Builds get defined in each spec's CONFIG file in the following example format:
```js
builds: {
  NO_IL: {
    url: "noil",
    name: "No Ice lance",
    icon: <img src={NoIceLance} alt="No Ice Lance" className="icon" />,
    supported: true
  }
}
```
Setting supported to false will stop the build from appearing in the selection (and hide the selection of no supported builds are set), while still allowing the build to be used in analyzers (to develop a build's support without having to immediately mark it as supported just to have its URL reference defined)

the url parameter gets passed into the URL so you can share links with rotations selected.
Within individual analyzers, you can access the URL of the current build under 
```js
this.owner.build
```
This value is undefined if no rotation is selected (aka standard).

Additionally, this.owner.builds will contain the object of builds defined in the spec's config file, with the active build (if one is selected) marked with an "active" flag.

Examples for testing for build selection in modules:
```js
if(this.owner.build === undefined) { ... } //only if no build is selected

if(this.owner.build === this.owner.builds.NO_IL.url) { ... } //only if NO_IL is active, solution below is preferred!
if(this.owner.builds.NO_IL.active) { ... } //only if build NO_IL is active 
```

Any feedback would be greatly appreciated as I haven't worked with URL manipulation before and UI isn't exactly my strength.

Screenshots (spec image for No-IL placeholder):

image
image

If no rotations are defined, the buttons do not appear:
image


src/interface/common/makeAnalyzerUrl.js
```diff
@@ -35,6 +35,9 @@ export default function makeReportUrl(report = undefined, fightId = undefined, p
             parts.push(duplicatePlayerNames.length > 1 ? `${playerId}-${prettyEncodeURI(playerName)}` : prettyEncodeURI(playerName));
             if (tab) {
               parts.push(tab);
+              if(build){
+                parts.push(build);
```
Dambroda: just a thought, maybe it should be build/tab rather than tab/build ? I think 'build' is of bigger scope

Zeboot: just pushed a change for that but I ran into slight issues when there's no build / no selected one, gonna try to fix those tomorrow probably

Sharrq:
I cant comment on the code cause i have no idea what any of it means, but i like the way it looks and such.

Also, just make sure you remove the stuff from the Frost Mage config before you merge it cause I am travelling later this week/this weekend and wont be able to add the analysis for it before then. Dont want people to click it and then wonder why nothing changed :)

Zeboot:
> I cant comment on the code cause i have no idea what any of it means, but i like the way it looks and such.
> 
> Also, just make sure you remove the stuff from the Frost Mage config before you merge it cause I am travelling later this week/this weekend and wont be able to add the analysis for it before then. Dont want people to click it and then wonder why nothing changed :)

I added a "supported" flag to builds and set the NoIL build to it, to hide builds that aren't supported yet while still allowing code (e.g in analyzers) to reference it.

Sharrq:
Locally i am trying to build on this and it keeps telling me this.owner.build is undefined even though i clicked the NoIL button. Not sure if there is something else i need to do to make this work?

Sharrq:
what is odd is that if i output this.owner, it shows the build
image

but if i output this.owner.build, it just says undefined

Zeboot:
> what is odd is that if i output this.owner, it shows the build
> image
> 
> but if i output this.owner.build, it just says undefined

should be fixed, you should now be able to test for selected build within analyzers like the following examples: 

```js
if(this.owner.build === undefined) { ... } //only if no build is selected
if(this.owner.builds.NO_IL.active) { ... } //only if build NO_IL is active 
```

Sharrq:
it seems to work now. Thanks!

Sharrq:
Ok, so its working everywhere except the checklist. getting Cannot read property 'build' of undefined

Zeboot:
> Ok, so its working everywhere except the checklist. getting Cannot read property 'build' of undefined

I don't think the checklist by default gets passed the fight object, you'd probably have to pass it through manually

Issue #3372, opened 2019-10-22
[Brewmaster] Celestial Fortune Tab


abelito75:
If the heal comes from an item or azerite it shows up as the spellId and no icon
log:
https://wowanalyzer.com/report/3m7QKDyVWpLFJ8zA/28-Mythic+Orgozoa+-+Wipe+21+(3:09)/Serenely/celestial-fortune

look at Serenely or Barreleye and you can see it

abelito75:
I fixed this.

PR #3381, opened 2019-10-28
[Brew] Items for Celestial Fortunes

abelito75:
Shows item names and icons as well to having a link to the item on wowhead. #3372
Also snuck in adding UT in general so we could include that :)
before:
before
after:
after


abelito75:
Heck. I didn't mean to add those other changes to this branch :( Its just a typo so its not the end of the world but still :(

PR #3382, opened 2019-10-28
[PROT WARRIOR] Added a highlight for bad Shield Block casts

abelito75:
As said in the title
before:
before

After:
after


abelito75:
Updated wording for tooltip but too zzz to take another picture
new wording: "This Shield Block didn't block enough damage nor did you cast enough Shield Slams."

PR #3387, opened 2019-10-31
[Restoration Shaman] Track & Show CH casts without bounces

niseko:
image
screw hunters

moved my commonly used nth function to remove redundancy


src/parser/shaman/restoration/modules/spells/ChainHeal.js
```diff
@@ -4,24 +4,76 @@ import SpellIcon from 'common/SpellIcon';
...
+    });
+  }
+
+  processBuffer() {
+    if (this.buffer.length === 0) {
+      return;
+    }
+    this.cast++;
+    this.chainHeals[this.cast] = {};
+    const cast = this.buffer.find(event => event.type === "cast");
```
Juko8: Looks really great, but variable names are a bit confusing. Using `cast`, `this.cast` and `this.casts` is probably not a good idea.

PR #3393, opened 2019-11-09
[MW] Upwelling

abelito75:
Added a statistic for upwelling the talent. The upwelling talent provides a stacking buff that is invisible so basically a steal of the essence font modules had to be done.
new
newtooltip

### DIFF
--- a/src/parser/monk/mistweaver/modules/talents/Upwelling.js
+++ b/src/parser/monk/mistweaver/modules/talents/Upwelling.js
@@ -168,11 +168,11 @@ class Upwelling extends Analyzer {
           </div>
             <ul>
               <li>Extra Bolts: {formatNumber(this.extraBolts)}</li>
-              <li>Extra Bolts Healing: {formatNumber(this.totalHealing)} ({formatPercentage(this.overhealingBolt)}% overheal)</li>
+              <li>Extra Bolts Healing: {formatNumber(this.totalHealing)} ({formatPercentage(this.overhealingBolt)}% overhealing)</li>
               <li>Extra Hots: {formatNumber(this.extraHots)}</li>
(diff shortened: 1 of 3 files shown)


src/parser/monk/mistweaver/modules/talents/Upwelling.js
```diff
@@ -0,0 +1,184 @@
...
+  efHeal(event) {
+    if(event.tick){
+      this.hotHeal(event);
+    }else{
+      this.boltHeal(event);
+    }
+  }
+
+  hotHeal(event){
+    const targetID = event.targetID;//short hand
```
rp4rk: Not sure this comment is adding value, if you truly wanted shorthand here you could try,
```js
const { targetID } = event;

// or

hotHeal({ targetID}) {...}
```

src/parser/monk/mistweaver/modules/talents/Upwelling.js
```diff
@@ -0,0 +1,184 @@
...
+    if(this.hotMap.has(targetID)){//check if hot heals before bolt hits (should never happen but logs)
+      if(this.hotMap.get(targetID).fullCount || event.timestamp - BASE_HOT_TIME > this.hotMap.get(targetID).applicationTime){//check if its an extra bolt from ef or was part of the core 18
+        this.efHotHeal += (event.amount || 0) + (event.absorbed || 0);
+        this.efHotOverheal += event.overheal || 0;
+      }
+    }
+  }
+
+  boltHeal(event){
+    const targetID = event.targetID;//short hand
```
rp4rk: See above

src/parser/monk/mistweaver/modules/talents/Upwelling.js
abelito75: Didn't know you could do this but short hand was more or less just meaning so I don't have to keep typing `event.targetID`. :P Will do it next time though :)

PR #3394, opened 2019-11-09
Protwarrior Bug fix

abelito75:
If shield block was casted before the fight begin the modules would die. (warcraft doesn't have shield block as a prepullbuff)


abelito75:
👍 Im planning on refactoring it in a bit to use the new event listeners but wanted to give a fix so the page wouldn't die on precasts

PR #3397, opened 2019-11-12
[MW] RM updates + hot tracking

abelito75:
Updated RM to include the healing from extra vivify cleave, extra evn hits, and extra ef mastery procs. Also made it so hot tracker can actually track rem bounces.

**### BIG NOTE HERE** These changes will show very little this patch but next patch it will be more useful. They are adding these features in its just they are extending the duration you extend hots so this is just getting work done ahead of time

new

### DIFF
--- a/src/parser/monk/mistweaver/modules/core/HotTracker.js
+++ b/src/parser/monk/mistweaver/modules/core/HotTracker.js
@@ -7,6 +7,8 @@ import calculateEffectiveHealing from 'parser/core/calculateEffectiveHealing';
 const PANDEMIC_FACTOR = 1.3;
 const PANDEMIC_EXTRA = 0.3;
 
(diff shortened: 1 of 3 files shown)


src/parser/monk/mistweaver/modules/core/HotTracker.js
```diff
@@ -7,6 +7,8 @@ import calculateEffectiveHealing from 'parser/core/calculateEffectiveHealing';
 const PANDEMIC_FACTOR = 1.3;
 const PANDEMIC_EXTRA = 0.3;
 
+const TFT_REM = 10000;//10 seconds for rem
```
MikeCook9994: team fight tactics?

abelito75: Thunder focus tea

Issue #3296, opened 2019-08-21
Holy Paladin: Lights Decree


5chmidt:
Ret Paladins have a tracker for the lights decree azurite trait which is now being used by glimmer paladins to extend our major cool down. We want to make a similar statistic tracker for holy that tracks HPS and additional glimmer output. 
Here is a link to a log that uses the trait:

https://www.wowhead.com/spell=286229/lights-decree
https://www.warcraftlogs.com/reports/jhTGQNDA2CXpdKZL/

5chmidt:
There are still some issues with #3402 I'll leave this here until we get everything cleaned up.

5chmidt:
Started reworking logic for Light's Decree here:
https://github.com/WoWAnalyzer/WoWAnalyzer/compare/master...5chmidt:bug/lights-decree

PR #3402, opened 2019-11-16
[Holy Paladin] Add Light's Decree statistics

bramdevries:
### Added

- Added Light's decree statistics about healing done, beacon healing and added uptime for Holy Paladin's

---

https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3296


5chmidt:
@bramdevries thanks for contributing!
I didn't see this before the PR was merged but there seems to be an issue with the calculations.  

Here is an extreme example of the calculation errors:
https://wowanalyzer.com/report/QcPYC2TWGBXvKtrh/30-Heroic+Za'qul+-+Kill+(5:54)/Daddypala/standard/character

image
The fight is 5 minutes 54 seconds long and Avenging Wrath is cast 4 times throughout the flight.  At 0:01, 1:46, 3:26 and 5:04.  The Light's Decree trait adds a 5 second extension to each Avenging Wrath that is cast, it does not apply on Vision of Perfection procs.  Therefore, it should be showing 20 seconds added.  

The additional Avenging Wrath time is caused by Vision of Perfection procs and should not be counted towards bonus healing from the Light's Decree trait.  Vision of Perfection procs for 35% of Avenging Wrath's base duration (20 seconds), which is accounts for the 7 second Avenging Wrath procs at: 0:54, 1:17, 4:14, 4:55 and 5:49.  

Vision of Perfection can proc during an Avenging Wrath, which causes the 7 seconds to be added to the duration of the current Avenging Wrath.  The base ability lasts 20 second, plus a 5 second extension from Light's Decree, plus a 25% extension from gives us a calculated duration of 31.25 seconds.  You can see the actual duration is being extend at 0:01(38.26 sec), 1:47(52.26 sec) and 3:27 (38.32 sec).  This means that at 0:01 there was one Vision of Perfection proc, three at 1:47 and one at 3:27.

It would probably be helpful to have a separate statistic tracker for the Vision of Perfection procs, but for now I think we at least need to change the extra duration tracker to only count times Avenging Wrath is cast.  It should be a pretty easy change: duration = casts * 5.

Finally, the tooltip needs some work.  Your string formatting is not correctly applying.
image

Since this has already been merged into master another PR is required to correct it.

bramdevries:
> Finally, the tooltip needs some work. Your string formatting is not correctly applying.

Not really sure what's up with that? When I look at it locally it does show up correctly:

image

Could there be something wrong with the `Trans` component? I noticed other modules (such as GlimmerOfLight) don't use `Trans` in their tooltip.

PR #3405, opened 2019-11-17
[Frost Death Knight] Fix crash in FDK rune tab

jjs451:
Stops dev instances from crashing and gets rid of the unknown spell text on live


src/common/SPELLS/deathknight.js
```diff
@@ -318,6 +318,12 @@ export default {
     icon: 'boss_odunrunes_blue',
   },
 
+  BREATH_OF_SINDRAGOSA_TALENT_RUNE_GAIN:{
+    id: 303753,
+    name: 'Breath of Sindragosa',
+    icon: 'inv_misc_rune_10'
```
jjs451:
```suggestion
    icon: 'inv_misc_rune_10',
```

PR #3406, opened 2019-11-18
[Frost Death Knight] Update statistic order to use new way

jjs451:
Feel like this one is self explanatory.  Just aiming to clear the deprecation warnings


jjs451:
This has been open for a week and since its a fairly minor change I'm gonna go ahead and hit merge.

PR #3410, opened 2019-11-24
[Holy Paladin] Correctly calculate extended duration provided by Light's Decree

bramdevries:
### Fixed

- Light's Decree statistic now only shows extended duration from actual Avenging Wrath casts, since procs from Vision Of Perfection are not extended.

### DIFF
--- a/src/parser/paladin/holy/modules/azeritetraits/LightsDecree.js
+++ b/src/parser/paladin/holy/modules/azeritetraits/LightsDecree.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Trans } from '@lingui/macro';
-import Analyzer from 'parser/core/Analyzer';
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
 import AbilityTracker from 'parser/shared/modules/AbilityTracker';
 import SPELLS from 'common/SPELLS';
 import AzeritePowerStatistic from 'interface/statistics/AzeritePowerStatistic';
@@ -10,6 +10,7 @@ import SpellIcon from 'common/SpellIcon';
 import BoringSpellValueText from 'interface/statistics/components/BoringSpellValueText/index';
 
 import BeaconHealSource from '../beacons/BeaconHealSource';
+import Events from 'parser/core/Events';
 
 const AW_BASE_DURATION = 20;
 const LIGHTS_DECREE_DURATION = 5;
@@ -37,6 +38,9 @@ class LightsDecree extends Analyzer {
     if (!this.active) {
       return;
     }
+
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.onHeal);
+    this.addEventListener(this.beaconHealSource.beacontransfer.by(SELECTED_PLAYER), this.onBeaconTransfer);
   }
 
   hasAvengingWrathBuffThroughLightsDecree(event) {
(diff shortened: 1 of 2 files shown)


5chmidt:
We have two open pull requests up to fix the same issue. We need to merge #3410 and #3411.

bramdevries:
> We have two open pull requests up to fix the same issue. We need to merge #3410 and #3411.

I guess this one can be merged first to fix the Vision Of Perfection incorrectness and then yours can be focussed on adding additional statistics

src/parser/paladin/holy/modules/azeritetraits/LightsDecree.js
```diff
@@ -68,7 +77,7 @@ class LightsDecree extends Analyzer {
     this.healingFromGlimmer += healingDone;
   }
 
-  on_beacontransfer(event) {
+  onBeaconTransfer(event) {
     if (!this.hasAvengingWrathBuffThroughLightsDecree(event)) {
       return;
     }
```
5chmidt: Tracking this.healingFromGlimmer is a little misleading.  If you were to take some other talent instead of Light's Decree you would still be casing Holy Shocks and which would contribute to glimmer healing.  You would be casting half as many Holy Shocks in the 6.25 second interval when wings is not up if you didn't have Light's Decree.  Therefore I don't think its fair to count 100% of glimmer healing during the wings extension toward the traits contribution.  Additionally the effect of the extra glimmers persists for 30 seconds not the 6.25 second extension granted by Light's Decree.  The correct way to track how Light's Decree effects glimmer would be to calculate the number that would have been cast without light's decree and the quantity that was cast with it.  Then track all the glimmer healing done by glimmers that were applied during the Light's Decree window.  Then count a percentage of the total toward the bonus healing provided by the trait as some glimmers would still be applied without it.  Additionally you would want to count additional glimmer triggers for damage and healing.  For example if 1 Holy Shock cast could have been done in the window without Light's Decree but 3 Holy Shocks were cast in the same 6.25 second window with Light's Decree, you would want to count all the damage and healing from all glimmers toward the extra healing provided by the trait.

src/parser/paladin/holy/modules/azeritetraits/LightsDecree.js
```diff
@@ -89,33 +98,21 @@ class LightsDecree extends Analyzer {
   }
```
5chmidt: I believe totalHealing() is double counting beacon healing, because beacon healing will also trigger onHeal, it just won't pass through information about the healing source spell.  You can check this by adding a console.log that prints when on onHeal event is triggered from beacon of light's spell id.

bramdevries:
@5chmidt seeing as you've changed almost everything in your own PR (https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3411/files) it  might be better to simply close this one?

5chmidt:
> @5chmidt seeing as you've changed almost everything in your own PR (https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3411/files) it might be better to simply close this one?

@bramdevries Yea, I think closing this PR is probably the way to go.  Sorry about high-jacking your PR.  I made a branch to test the changes and got a little carried away.  I like where you are headed with tracking the Glimmer of Light procs as part Light's Decree's benefit because that is definitely a major reason to use the trait.  However its difficult to accurately represent in a statistic.

PR #3411, opened 2019-11-26
[Holy Paladin] Lights Decree *BUG*

5chmidt:
The live version of the Light's Decree statistic has several inaccurate and misleading statistics.  An example log which shows them is [here.](https://wowanalyzer.com/report/QcPYC2TWGBXvKtrh/30-Heroic+Za'qul+-+Kill+(5:54)/Daddypala/standard/character)  I have a more in-depth explanation on the calculation issues in the PR for #3402 #3296.

Here is the live version of the statistic:
image
Currently we are counting all healing that is done with Avenging Wrath up as +healing for the statistic.  I think this is misleading, if wings was not up, presumably we would still be healing, but our heals would be less effective without the 30% critical strike and 20% damage and healing buff.  When showing statistics for the trait I am comparing with doing a similar rotation without the damage/ crit buff. This causes the statistic shown to be lower than the previous version that compares healing during wings against standing still and not casting anything for the same period of time.

In the logs I checked the duration of Avenging Wrath with Light's Decree and Sanctified Wrath is 31.25 seconds.  I appears that both the base Avenging Wrath (20 seconds) and Light's Decree(5) seconds are being extended by 25% which gives us the correct duration calculation.

The goal in most Light's Decree builds is to cast as many Holy Shocks as possible so a tool-tip was added to shows how many Holy Shocks were cast and the number of extra granted extending the Avenging Wrath buff.  As Avenging Wrath reduces the CD by 50% to assumption is you would cast half as many without wings up.  This assumes that if you messed up your cast order with wings up you would do the same without it. 

Total healing is calculated from the bonus healing provided by the flat +20% to healing and during the extra duration given by the trait.  Extending Wings also extends the +30% crit buff, I am using the crit contribution formula [here](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/paladin/holy/modules/StatValues.md) to correctly account for both the additional crit chance and not give credit for overhealing.  The total HPS is the net increase from the flat healing and crit% buffs combined.

The same is done with damage, calculating the flat damage contribution along with the crit contribution on all critical hits.  For Mythic+ and raid farming, the damage output of wings can become as important as the healing it provides.

Here is the corrected version of the statistic box:
image

Here is the tooltip which shows the damage, healing, and holy shock casts.
image

I considered accounting for the damage and healing that was done by Glimmer\ Radiant Incandescence as they also contribute to the damage and healing from Holy Shock which make Light's Decree a BIS trait.  I decided that each other trait had its own statistic tracker to handle those and they didn't need to also be tracked on Light's Decree.


5chmidt:
Don’t merge this yet. I just realized that I’m not checking for over healing in my flat healing calculation. I’ll add this back in and push the final PR tomorrow.

abelito75:
wait why no

5chmidt:
@abelito75 this pull request should be ready to go.  I had to fix the calculation for HPS added by the flat 20% healing bonus.  Previously its was giving 20% credit on over healing during wings.  Its has been fixed to correctly decrease the additional healing granted by the trait if the 20% bonus healing is used for over healing.

PR #3413, opened 2019-11-27
Add common Earth Shield module for shamans. Update resto earth shield module to use it

MikeCook9994:
Add common module for earth shield and update existing resto module to use it. 

Spell list based off of wowhead spell data
https://www.wowhead.com/spell=974/earth-shield


src/parser/shaman/shared/constants.js
```diff
@@ -0,0 +1,13 @@
+import SPELLS from 'common/SPELLS';
+
+export const HEALING_ABILITIES_AMPED_BY_EARTH_SHIELD = [
```
niseko: Thanks for catching that, you're indeed correct that many of the spells in my list don't work with earth shield as they're not shaman specific. There are a few spells missing however:

```suggestion
export const HEALING_ABILITIES_AMPED_BY_EARTH_SHIELD = [
	SPELLS.HEALING_WAVE.id,
	SPELLS.ASCENDANCE_INITIAL_HEAL.id,
	SPELLS.WELLSPRING_HEAL.id,
	SPELLS.UNLEASH_LIFE_TALENT.id,
```
ascendance redistribution and cbt technically do as well by proxy; but they're such special cases that i'll look at it at some other time

src/parser/shaman/shared/talents/EarthShield.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  onEarthShieldHeal(event) {
+    this.healing += (event.amount + (event.absorbed || 0));
+  }
+
+  onEarthShieldAmpSpellHeal(event) {
+    const combatant = this.combatants.players[event.targetID];
+    if (combatant && combatant.hasBuff(SPELLS.EARTH_SHIELD_TALENT.id, event.timestamp)) {
+      console.log(event);
```
niseko: leftover console.log
```suggestion
```

src/parser/shaman/shared/talents/EarthShield.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  statistic() {
+    return (
+      <StatisticBox
+        label={<SpellLink id={SPELLS.EARTH_SHIELD_TALENT.id} />}
+        category={STATISTIC_CATEGORY.TALENTS}
+        position={STATISTIC_ORDER.OPTIONAL(45)}
+        tooltip={`${formatPercentage(this.owner.getPercentageOfTotalHealingDone(this.healing))}% from the HoT and ${formatPercentage(this.owner.getPercentageOfTotalHealingDone(this.buffHealing))}% from the healing amp.`}
```
niseko: I think amplification or increase instead of shortening it to `amp` sounds more clear here

src/parser/shaman/shared/constants.js
MikeCook9994: Oh, I totally missed the "see more" button on the spell data in wowhead page I linked. I'll fix that up.

PR #3416, opened 2019-11-30
Judgment of light

abelito75:
Finally doing a Feature request from Zero from 2017


5chmidt:
This PR looks good. Do you have an example log to test with?  I don't know anyone that has been Judging Light for the last two tiers.

abelito75:
Yeah I have a test log... It's an uldir log but still should be valid. It just died bc I didn't update the tests

src/parser/paladin/holy/modules/talents/JudgmentOfLight.js
```diff
@@ -0,0 +1,57 @@
...
+
+  counter = JUDGEMENT_HEALS;
+  wasted = 0;
+  casts = 0;
+
+  constructor(...args) {
+    super(...args);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.JUDGMENT_OF_LIGHT_TALENT.id);
+    if (!this.active) {
+      return;
```
HawkCorrigan: This does nothing

Issue #3409, opened 2019-11-23
Holy Paladin: Glimmer Build Tracker


5chmidt:
Create a toggle for the glimmer build, which should be triggered when using 3 glimmer, Sanctified Wrath, and Holy Avenger.

Suggestions that are not important:
- Mana Utilization (not important in glimmer) #3417
- Light of Dawn Utilization (generally better to use Holy Shock over Light of Dawn) #3418
- Light's Decree #3296
- Wasted Infusion of Lights is not important.
- Add support for Vision of Perfection Trait. #3450

Add suggestion for:
- Casting Crusader Strike when Holy Shock is off CD. #3445

Test log:
https://wowanalyzer.com/report/Ccxz4VWDtdm3kKp9/5-Heroic+Za'qul+-+Kill+(6:13)/1-Holyschmidt

PR #3417, opened 2019-12-01
[Holy Paladin] Glimmer of Light Build

5chmidt:
- Added a build for glimmer paladin, this can be toggled on and off by the user.
image

- The default suggestion to cast Judgement with Judgement of Light as been changed to only appear with judgement of light is active.
image
(Without Judgement of Light)

image
(With Judgement of Light)

- The more info link was changed to point to the questionably epic glimmer paladin build article, or the default location depending on the selected spec.
- The mana usage suggestion was removed for glimmer paladins as the low mana requirements of Crusader Strike, Holy Shock and Light of Dawn allow for optimal performance without using all your mana.

There are still a few items I would like to add to the build from #3409, but I think this is a good start with the other changes to glimmer coming through later PRs to tweak suggestions for specific modules.


joshinat0r:
will look at #3418 in a second

PR #3432, opened 2019-12-17
Blood of the Enemy

jjs451:
image

The damage stat for the major does not include the damage from the crits gained from the +25% crit chance.  Not sure how to accurately model that without guessing if a crit came from baseline crit chance or the bonus crit chance


package.json
```diff
@@ -17,6 +17,7 @@
     "es6-error": "^4.1.1",
     "fuse.js": "^3.4.5",
     "history": "^4.9.0",
+    "open": "6.4.0",
```
emallson: these changes to `package.json` seem unintentional? I don't see `open` used anywhere in your code.

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+
+    const rank = this.selectedCombatant.essenceRank(SPELLS.BLOOD_OF_THE_ENEMY.traitId);
+    this.statTracker.add(SPELLS.BLOOD_SOAKED_HASTE_BUFF.id, {
+      crit: (rank === 1) ? 0 : 8,
+      haste: 548,
+    });
+
+    this.statTracker.add(SPELLS.BLOOD_SOAKED_STACKING_BUFF.id, {
```
emallson: this should probably be:

```js
if(rank > 1) {
   this.statTracker.add(SPELLS.BLOOD_SOAKED_STACKING_BUFF.id, { crit: 8 });
}
```

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,183 @@
...
+          recommendedEfficiency: 0.80,
+        },
+      });
+    }
+    
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(SPELLS.BLOOD_OF_THE_ENEMY), this.onMajorCastDamage);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDamage);
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.BLOOD_SOAKED_STACKING_BUFF), this.handleStacks);
+    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.BLOOD_SOAKED_STACKING_BUFF), this.handleStacks);
+    this.addEventListener(Events.applybuffstack.by(SELECTED_PLAYER).spell(SPELLS.BLOOD_SOAKED_STACKING_BUFF), this.handleStacks);
```
emallson: some indentation issues

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+    if (!enemy.hasBuff(SPELLS.BLOOD_OF_THE_ENEMY.id, event.timestamp, 0, 0, this.selectedCombatant)) {
+      return;
+    }
+    const isCrit = event.hitType === HIT_TYPES.CRIT || event.hitType === HIT_TYPES.BLOCKED_CRIT;
+    if (!isCrit) {
+      return;
+    }
+
+    this.bonusDamage += calculateEffectiveDamage(event, BOTE_DAMAGE_BONUS);
```
emallson: the bonus crit damage is only at R3, but this event handler fires for R1 and R2 as well. should only add the event listener for R3 BotE

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  handleStacks(event) {
+    if (this.rank < 2) {
+      return;
+    }
```
emallson: it is preferable to only add event listeners that will be used in the current fight. if the player has R1 BotE, don't even add the event listeners for `handleStacks`.

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,183 @@
...
+
+    const uptimeOnStack = event.timestamp - this.lastStackTimestamp;
+    this.totalCrit += this.currentStacks * CRIT_PER_STACK * uptimeOnStack;
+    console.log(this.currentStacks * CRIT_PER_STACK * uptimeOnStack);
```
emallson: stray `console.log`

package.json
jjs451: Should be gone now.  I was having issues with WSL and needed to manually set the `open` version to get the site running on my machine.

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
jjs451: I saw and fixed the issue with the gcd block not being indented.  If there are others I'm not seeing them

jjs451:
Updated the tooltip on the stat box for the major to make it in line with most other tooltips
image

jjs451:
Ended up making more changes beyond the first clean up.  Reduced the number of arguments to `enemy.hasBuff` on line 113.  Also reworked the tooltip to be on the stat box itself instead of as its own tooltip element.  I needed a conditional tooltip showing damage from the amplified crits granted by the R3 major so I did that logic on lines 156 and 172.  It works, but I'm not sure if we have a different standard for handling that situation.

src/parser/shared/modules/spells/bfa/essences/BloodOfTheEnemy.js
```diff
@@ -0,0 +1,188 @@
...
+    return (HASTE_PROC_AMOUNT * this.uptime);
+  }
+
+  get averageCrit() {
+    return (this.totalCrit / this.owner.fightDuration).toFixed(0);
+  }
+
+  statistic() {
+    const rank = this.selectedCombatant.essenceRank(SPELLS.BLOOD_OF_THE_ENEMY.traitId);
+    const rankThreeTooltip = (rank > 2) ? <><br />Damage done from increased critical hit damage: {formatNumber(this.bonusDamage)}</> : <> </>;
```
MartijnHols:
```suggestion
    const rankThreeTooltip = (rank > 2) ? <><br />Damage done from increased critical hit damage: {formatNumber(this.bonusDamage)}</> : '';
```

PR #3433, opened 2019-12-17
[Fire Mage] Merged Combustion & Converted to TypeScript

Sharrq:
I merged the Combustion modules back together and tried my hand at converting the module to TypeScript.

If someone can take a look at this and make sure i didnt screw anything up given that this is my first attempt at TypeScript.

Also removed a couple console logs from the last change i made that i forgot to remove


src/parser/mage/fire/modules/features/Combustion.tsx
```diff
@@ -0,0 +1,267 @@
...
+
+class Combustion extends Analyzer {
+  static dependencies = {
+    spellUsable: SpellUsable,
+    abilityTracker: AbilityTracker,
+  };
+  protected spellUsable!: SpellUsable;
+  protected abilityTracker!: AbilityTracker;
+
+  hasFlameOn: boolean; hasPhoenixFlames: boolean; hasFirestarter: boolean;
```
MartijnHols: Each prop on a separate line please

src/parser/mage/fire/__snapshots__/CombatLogParser.test.js.snap
```diff
@@ -3391,17 +3391,29 @@ exports[`The Fire Mage Analyzer analyzers Combatants matches the statistic snaps
...
-exports[`The Fire Mage Analyzer analyzers CombustionCharges matches the suggestions snapshot 1`] = `Array []`;
-
-exports[`The Fire Mage Analyzer analyzers CombustionFirestarter matches the statistic snapshot 1`] = `"module inactive"`;
-
-exports[`The Fire Mage Analyzer analyzers CombustionFirestarter matches the suggestions snapshot 1`] = `"module inactive"`;
-
-exports[`The Fire Mage Analyzer analyzers CombustionSpellUsage matches the statistic snapshot 1`] = `"module has no statistic method"`;
-
-exports[`The Fire Mage Analyzer analyzers CombustionSpellUsage matches the suggestions snapshot 1`] = `Array []`;
+exports[`The Fire Mage Analyzer analyzers Combustion matches the suggestions snapshot 1`] = `
```
MartijnHols: Did you change the rules too or was this unintended?

src/parser/mage/fire/modules/features/Combustion.tsx
```diff
@@ -0,0 +1,267 @@
...
+  castEvent!: {
```
MartijnHols: This should not have the !, but its value should have | undefined appended.

src/parser/mage/fire/modules/features/Combustion.tsx
```diff
@@ -0,0 +1,267 @@
...
+      this.hasPhoenixFlames = this.selectedCombatant.hasTalent(SPELLS.PHOENIX_FLAMES_TALENT.id);
+      this.hasFirestarter = this.selectedCombatant.hasTalent(SPELLS.FIRESTARTER_TALENT.id);
+      this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.COMBUSTION), this.onCombustion);
+      this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(DAMAGE_SPELLS), this.onDamage);
+      this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell([SPELLS.FIREBALL, SPELLS.SCORCH]), this.onCast);
+      this.addEventListener(Events.begincast.by(SELECTED_PLAYER).spell([SPELLS.FIREBALL, SPELLS.SCORCH]), this.onCast);
+  }
+
+  //When Combustion is cast, check to see how many charges of Fire Blast and Phoenix Flames are available. If there is less than (Max Charges - 1) then its a bad Combustion cast. 
+  onCombustion(event: any) {
```
MartijnHols: Should be CastEvent

src/parser/mage/fire/modules/features/Combustion.tsx
MartijnHols: Looking at your usage, it should probably be `CastEvent | undefined`

src/parser/mage/fire/__snapshots__/CombatLogParser.test.js.snap
Sharrq: I didnt change any of the rules, or anything in the checklist, just the module names

src/parser/mage/fire/modules/features/Combustion.tsx
Sharrq: I think i tried that but then flagTimeline() complained that castEvent could be undefined. I told VS Code to auto fix it and that is what it did. I can try and change it back and see if it works though.

Sharrq:
I could separate them again if you think it is better the other way. I separated them in the first place cause it was a nightmare trying to fit everything into a single on_byPlayer_cast. But then Event listeners made it easier to separate things.

The biggest reason i wanted to merge them back together was because there are portions of each of them that overlap with the others. Like some of them have the same event listeners, they just check something different, so it made sense to me to combine them.

But if it is considered cleaner to leave them separated then i can do that.

MartijnHols:
Well an important guideline for writing code is the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). Looking at your code, it seems you have three different features;

- Firestarter (don't cast Combustion during Firestarter)
- Fresh Combustion casts (with low Fire Blast charges)
- Casts during Combustion (Combustion usage)

These "features" seem to only have in common that they interact with Combustion, but I reckon they are all independent rules. Keeping them separate will keep the modules smaller, and fewer code is much easier to reason about and therefore maintain than a lot of code where you have to make a mentally try to keep the 3 different requirements separate. This may be easy now as you're writing it since the entire thing is in your head, but if you want to make a simple change 6 months from now you'll have to get all 3 features in your head at the same time even if you just modify a single thing, just to avoid breaking something else.

I absolutely believe keeping them separate will be beneficial to you.

src/parser/mage/fire/__snapshots__/CombatLogParser.test.js.snap
MartijnHols: So then there's a new bug since the snapshot suggests a new suggestion is being done?

src/parser/mage/fire/modules/features/Combustion.tsx
MartijnHols: How about making the event a part of the parameters of `flagTimeline`? Then TS will see that you always take care to ensure the property has a value and pass.

src/parser/mage/fire/__snapshots__/CombatLogParser.test.js.snap
Sharrq: i guess so. i didnt even look at what changed in the snapshot, woops

src/parser/mage/fire/modules/features/Combustion.tsx
Sharrq: I think i tried that too, but i might have just thought about that and decided to let vs code fix it instead, not sure

Sharrq:
Gonna just close this and fix it via a different PR, easier that way.

PR #3437, opened 2019-12-19
Desperate prayer

HawkCorrigan:
Some check it against edge please.


src/parser/priest/shared/CHANGELOG.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import {
+  HawkCorrigan,
+} from 'CONTRIBUTORS';
+import { change, date } from 'common/changelog';
+
+export default [
+  change(date(2019, 12, 19), <>Fixed some ordering issues in the DP module.</>, [HawkCorrigan]),
```
MartijnHols: This changelog is unused. Please add it to the changelog of all Priest specs (so 3 entries).

In addition please use <SpellLink /> for DP as not everyone may understand it.

abelito75: I'm a bit late but shadow doesn't have access to Desperate prayer so plz no change log there

abelito75:
Is this a dead pr?

HawkCorrigan:
ye propably
I cant test it properly and it doesnt look like anyone cares too much about edge anyway

PR #3445, opened 2020-01-01
Patch 8.3 Glimmer Update

5chmidt:
Patch 8.3 is changing the way Glimmer of Light plays by capping the number of friendly/ enemy targets.  As the change to the glimmer cap will not effect non-glimmer variants I have moved the suggestions to the glimmer build only.  The tracked statistics will still show up for anyone using a glimmer of light trait if the have the build selected or not.  Until 8.3 the statistic will show up with a cap of 99, which looks a little weird on the tool-tip but will report correctly.
image

Here is what the tool-tip will look like after 8.3.
image

There are two suggestions for the glimmer build, to not over-cap and to avoid refreshing glimmer builds early.  Both suggestions are not strict requirements more rules of thumb. I have attempted to set the suggestion thresholds accordingly.  For example with wings up you will probably cast more than 8 Holy Shocks in 30 seconds, which will put you over cap with correct play.  If a member of your party has a ticking DoT the correct play would be to recast Holy Shock to heal them even if it requires refreshing glimmer early.  The current values are probably set a little high and will not show the suggestion often enough, but they can be lowered after the first few weeks of 8.3 when the meta shakes out.

image
Early refreshes will probably be more prevalent with the limit of 8 as you will need to keep up glimmers on priority targets.  Currently the threshold for suggestions is 15, 25 and 35 percent for total up-time lost to early refreshes.

image
The glimmer cap will drop the last cast glimmer and therefore will not cause a large loss in up-time percentage.  The current threshold is set at 10, 20 and 30%.  This can probably be lowered, even with current play style of stacking haste cool downs, only high end raiders are breaking the 10% mark so this could stand to be lowered. 

#3409


MartijnHols:
"Try to avoid overwriting buffs too early." almost suggest "cast Holy Shock less often". I think this might be a misleading suggestion. Holy Shock is still the strongest heal even without Glimmer, is it not?

src/parser/paladin/holy/modules/azeritetraits/GlimmerOfLight.js
```diff
@@ -108,42 +134,75 @@ class GlimmerOfLight extends Analyzer {
...
-      )
+    if (this.owner.builds.GLIMMER.active){
+      when(this.suggestEarlyRefresh).addSuggestion((suggest, actual, recommended) => {
+        return suggest(
+          <Trans>
+            Your usage of <SpellLink id={SPELLS.GLIMMER_OF_LIGHT.id} /> can be improved. Try to avoid overwritting buffs too early.
+          </Trans>,
+        )
+          .icon(SPELLS.GLIMMER_OF_LIGHT.icon)
+          .actual(`Percentage uptime lost to early refresh was ${formatPercentage(this.earlyGlimmersWasted)}%`)
```
MartijnHols: I think this is worded slightly vague. Casting HS while already capped will just apply a new buff, right? So no actual uptime of what was max feasible with the cap will be lost. How about:

```suggestion
        .actual(`${formatPercentage(this.overCapGlimmersWasted)}% buff duration dropped due to buff cap`)
```

5chmidt: > Holy Shock is still the strongest heal even without Glimmer, is it not?  @MartijnHols 

Here is a top [mythic log](https://www.warcraftlogs.com/reports/6MbwHrG1Whkgj3Kp#fight=24&type=healing&source=6) from a kill in August, Glimmer is doing 32.70% while Holy Shock is doing 22.22%.  Even in [lower level play](https://www.warcraftlogs.com/reports/4qWtBdMw2TvNhpJX#fight=29&type=healing&source=15) when glimmer is doing 38.6% overhealing it is still doing as much healing to Holy Shock.  The goal is always to cast as many Holy Shocks as possible, but you also need to be mindful of who already has a glimmer and who does not.  Correct play will result in a number of early refreshes, to get the most benefit from Glimmer of Light you want to stay as close to 8 active glimmers for as much of the fight as possible.

> Casting HS while already capped will just apply a new buff, right? So no actual up-time of what was max feasible with the cap will be lost.

I agree that the wording needs some clarity, however I don't think combining the suggestions is the solution.  The suggestion starting at line 191 counts the glimmers cast over the cap, the first suggestion is for counting early refreshes.  These are two separate items that can be improved upon.   With current haste levels you are only getting above the 8 Glimmer cap by stacking all your haste cooldowns.  Moving into 8.3 you will want to stagger your cooldowns to stay at cap for the majority of the fight, the current play style is to get 15+ glimmers rolling into major damage events then holding CDs for the next big raid damage.  The suggestion is there to remind players of the cap and ensure they change their play accordingly.

It is true that when you already have 8 glimmers up an early refresh actually the same as recasting on a previous target.  I will swap the order of the if statements accordingly.

5chmidt: I have changed the statistic to count all Holy Shocks cast with 8 glimmers up as an over-cap cast.  I have therefore increased the cap suggestion threshold to 15% as it will be counting more glimmers than the previous method.

I made some changes to the suggestion wording in the suggestion.
image

Both the suggestion text and the suggestion threshold prioritize healing low health targets, but if >%30 of your Holy Shocks are going to the same person it is leaving a lot of potential healing on the table and should be pointed out.  (Everyone has that one guildie who is always standing in the fire, but 30% is a little too much).

PR #3446, opened 2020-01-01
Update code integration tests to be more maintainable.

MartijnHols:
- Move the log zips into spec folder. This makes it less suggestive that you should only have 1 integration tests, allows us to more execute these tests separately and allows us to colocate the log files
- Fix bug where all channeling events were triggered twice by disabling the ChannelingShared module that caused the bug
- Fix a bunch of console warnings/errors when running (integration) tests
- Separate test workflow jobs by type
- Add yarn scripts to run a specific test type
- Build the app during testing phase to speed up pipeline

TODO:
- Colocate snapshots: https://github.com/facebook/jest/issues/1650#issuecomment-475912058
- Build a new solution for shared channeling events


emallson:
@MartijnHols not sure if its a github hiccup or problem with the new workflow spec, but its stuck at `Tests Expected — Waiting for status to be reported`

PR #3448, opened 2020-01-02
[WIP] Phase/Time Eventfilter TS conversion

Zeboot:
WIP, looking for feedback/help here as I've never worked with TypeScript before.

Specifically not sure about all the "casting" and ```!``` I'm using right now, especially in TimeEventFilter.ts, if anyone has tips on how to do this better, they'd be much appreciated.
This is still a bit messy as these files touch a lot of other parts of the code and I tried to find a balance between eliminating a lot of "any" declarations while not editing every single file in the codebase.

Also fixes the issue mentioned in discord regarding time input breaking if you delete its contents


src/interface/report/PhaseParser.ts
```diff
@@ -0,0 +1,91 @@
...
+    end_time: number,
+    boss: number,
+  },
+  bossPhaseEvents: PhaseEvent[],
+  children: (isLoading: boolean, phases?: {[key:string]: Phase}) => ReactNode,
+}
+
+interface State {
+  isLoading: boolean,
+  phases?: {[key:string]: Phase},
```
MartijnHols:
```suggestion
  phases: { [key:string]: Phase } | undefined,
```

I believe propName**?** is intended for optional props, while in this case the prop will always exist, it may just have an undefined value. Not sure though, but it may still be worthwhile to make the distinction.

src/interface/report/PhaseParser.ts
```diff
@@ -0,0 +1,91 @@
...
+      this.parse();
+    }
+  }
+
+  makePhases() : {[key:string]: Phase}{
+    const { bossPhaseEvents, fight } = this.props;
+    if(!bossPhaseEvents){
+      return {};
+    }
+    const phaseStarts = Array.from(new Set<string>(bossPhaseEvents.filter((e : PhaseEvent) => e.type === EventType.PhaseStart).map((e: PhaseEvent) => e.phase.key))); //distinct phase starts
```
MartijnHols: this is a lot of code on one line. If you're doing distinct two or more times, maybe make a function for it?

e.g. `const distinct = <T extends any>(items: T[]) => Array.from(new Set<T>(items))`

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -36,13 +47,19 @@ class PhaseSelector extends React.PureComponent {
   //builds a dictionary of phases / phase instances to keep track of in order to be able to attribute a unique "key" to each phase for the dropdown
   //without losing the actual key (and without having to for example replace an "instance token" like an underscore)
   buildPhases() {
-    const phases = [];
+    const phases : {
```
MartijnHols:
```suggestion
    const phases: Phase[]
```

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -36,13 +47,19 @@ class PhaseSelector extends React.PureComponent {
...
-      if (phase.start.length !== phase.end.length) {
-        phases.push({ name: phase.name, key: key, start: phases.start[0], instance: 0 });
+      if (phase!.start!.length !== phase!.end!.length) {
+        phases.push({ name: phase.name, key: key, start: phase!.start![0], instance: 0 });
       } else {
-        phases.push(...phase.start.map((start, index) => (
+        phases.push(...phase!.start!.map((start, index) => (
           {
             name: phase.name,
             key: key,
```
MartijnHols:
```suggestion
            key,
```

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -36,13 +47,19 @@ class PhaseSelector extends React.PureComponent {
...
+      key: string;
+      instance: number;
+      start: number;
+      multiple?: boolean;
+    }[] = [];
     Object.keys(this.props.phases).forEach(key => {
       const phase = this.props.phases[key];
-      if (phase.start.length !== phase.end.length) {
-        phases.push({ name: phase.name, key: key, start: phases.start[0], instance: 0 });
+      if (phase!.start!.length !== phase!.end!.length) {
```
MartijnHols:
```suggestion
      if (phase.start!.length !== phase.end!.length) {
```

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -36,13 +47,19 @@ class PhaseSelector extends React.PureComponent {
...
+        phases.push({ name: phase.name, key: key, start: phase!.start![0], instance: 0 });
```
MartijnHols:
```suggestion
        phases.push({ name: phase.name, key: key, start: phase.start![0], instance: 0 });
```

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -36,13 +47,19 @@ class PhaseSelector extends React.PureComponent {
...
       } else {
-        phases.push(...phase.start.map((start, index) => (
+        phases.push(...phase!.start!.map((start, index) => (
```
MartijnHols:
```suggestion
        phases.push(...phase.start!.map((start, index) => (
```

src/interface/report/Results/PhaseSelector.tsx
```diff
@@ -85,7 +102,7 @@ class PhaseSelector extends React.PureComponent {
         {fight.filtered && !fight.phase && <option key="custom" value={SELECTION_CUSTOM_PHASE}>Custom</option>}
         <option key="all" value={SELECTION_ALL_PHASES}>All Phases</option>
         {Object.keys(phases).map(key =>
-          <option key={key} value={key}>{phases[key].name}{phases[key].multiple ? ' ' + (phases[key].instance + 1) : ''}</option>,
+          <option key={key} value={key}>{phases[key].name}{phases[key].multiple ? ' ' + (phases[key]!.instance! + 1) : ''}</option>,
```
MartijnHols:
```suggestion
          <option key={key} value={key}>{phases[key].name}{phases[key].multiple ? ' ' + (phases[key].instance! + 1) : ''}</option>,
```

src/interface/report/Results/PhaseSelector.tsx
MartijnHols: Needing so many ! is a sign that types can be improved. Is it maybe an idea to have two different types for Phases; e.g. `PhaseConfig` and `Phase`? I'm not sure how it's used so I can't really tell

src/interface/report/Results/TimeFilter.tsx
```diff
@@ -1,18 +1,23 @@
...
 import TimeInput from './TimeInput';
 
-class TimeFilter extends React.PureComponent {
-  static propTypes = {
-    fight: PropTypes.object.isRequired,
-    isLoading: PropTypes.bool.isRequired,
-    applyFilter: PropTypes.func.isRequired,
-  };
+interface Props {
+  fight: any,
```
MartijnHols: You used a fight type earlier with more details, if you make that an interface and export it, you can use it here to have at least a minimal amount of type info.

src/interface/report/Results/TimeInput.tsx
```diff
@@ -64,45 +68,45 @@ class TimeInput extends React.PureComponent {
 
   }
 
-  handleChangeM(e) {
-    const val = parseInt(e.target.value, 10);
+  handleChangeM(e : React.ChangeEvent<HTMLInputElement>) {
+    const val = parseInt(e.target.value, 10) || 0;
```
MartijnHols:
```suggestion
    const val = Number(e.target.value) || 0;
```

src/interface/report/Results/TimeInput.tsx
```diff
@@ -64,45 +68,45 @@ class TimeInput extends React.PureComponent {
...
       this.forceUpdate();
       return;
     }
     this.changeTime(MINUTE * val + SECOND * this.state.seconds + this.state.milliseconds);
   }
 
-  handleChangeS(e) {
-    const val = parseInt(e.target.value, 10);
+  handleChangeS(e : React.ChangeEvent<HTMLInputElement>) {
+    const val = parseInt(e.target.value, 10) || 0;
```
MartijnHols:
```suggestion
    const val = Number(e.target.value) || 0;
```

src/interface/report/Results/TimeInput.tsx
```diff
@@ -64,45 +68,45 @@ class TimeInput extends React.PureComponent {
...
       this.forceUpdate();
       return;
     }
     this.changeTime(MINUTE * this.state.minutes + SECOND * val + this.state.milliseconds);
   }
 
-  handleChangeMs(e) {
-    const val = parseInt(e.target.value, 10);
+  handleChangeMs(e : React.ChangeEvent<HTMLInputElement>) {
+    const val = parseInt(e.target.value, 10) || 0;
```
MartijnHols:
```suggestion
    const val = Number(e.target.value) || 0;
```

src/interface/report/Results/Timeline/Lane.js
```diff
@@ -110,9 +110,9 @@ class Lane extends React.PureComponent {
     const { children, style } = this.props;
 
     const ability = children[0].ability;
-    if(children[0].type === PRE_FILTER_COOLDOWN_EVENT_TYPE || children[0].type === "cast"){ //if first cast happened before phase
+    if(children[0].type === EventType.FilterCooldownInfo || children[0].type === "cast"){ //if first cast happened before phase
```
MartijnHols:
```suggestion
    if(children[0].type === EventType.FilterCooldownInfo || children[0].type === EventType.Cast){ //if first cast happened before phase
```

src/interface/report/Results/Timeline/Lane.js
```diff
@@ -110,9 +110,9 @@ class Lane extends React.PureComponent {
...
       const nextChildren = children.slice(1, children.length); //all children following the first cast
-      const nextCast = nextChildren.findIndex(e => e.type === "cast" || e.type === PRE_FILTER_COOLDOWN_EVENT_TYPE) + 1; //add 1 since we're searching through the events FOLLOWING the initial cast
+      const nextCast = nextChildren.findIndex(e => e.type === "cast" || e.type === EventType.FilterCooldownInfo) + 1; //add 1 since we're searching through the events FOLLOWING the initial cast
```
MartijnHols:
```suggestion
      const nextCast = nextChildren.findIndex(e => e.type === EventType.Cast || e.type === EventType.FilterCooldownInfo) + 1; //add 1 since we're searching through the events FOLLOWING the initial cast
```

src/interface/report/Results/TimeFilter.tsx
Zeboot: I wasn't sure where to put a fight definition like that so I just opted to use any here and update it at a later stage when somebody converts that part to TS. if you have an idea where you want the fight object to be exported from I can try to fit all the attributes it possesses into it

MartijnHols: Just put it in the first place you run into it and move it if/when you find a better location

src/interface/report/Results/PhaseSelector.tsx
Zeboot: wasn't entirely sure how to solve this since the phase type is probably too overloaded for its own good but I think I have a solution for this in the new commits? def need more feedback on that one tho

PR #3452, opened 2020-01-08
[Core] Nya'lotha Bosses

Sharrq:
Adds the boss info, phases, images for Nyalotha


Sharrq:
This is probably good enough to review and merge. The bosses that I need to tweak, i can take another look once the raid goes live and i have some better logs to work with. 

As is, there is just a couple missing phases and one that is a few seconds off.

MartijnHols:
Thanks!

PR #3460, opened 2020-01-18
[Frost Death Knight] Update spec compatibility to 8.3

jjs451:
No patch notes for FDK in 8.3


MartijnHols:
Thanks! FFR you can merge your own PRs when you either don't think they require a review, or you're ready to patch it up should there be a problem.

PR #3462, opened 2020-01-23
Corruption parsing & Ineffable Truth

niseko:
A rough implementation of the corruption system so that you can at least find out which corruptions the player has (per spellId or name) and how many (spellId only).

Because I forgot to split it into 2 branches, here's also a first corruption: 
- https://ptr.wowhead.com/spell=318484/ineffable-truth

Its not 100% confirmed that stacking is additive, but its the one that makes the most sense here and who knows when I'll find somebody who has 2 of them. This implementation at least fixes a ton of the "spell cast before cooldown finished" issues for everyone using it.

image


Pewtro:
I think this looks good and works as a good baseline for further work

PR #3475, opened 2020-01-31
Blood of the Enemy bug fix

jjs451:
Discovered that the bonus crit damage is based on a buff on the player, not the debuff on the enemy as I previously assumed.  This PR fixes that and makes sure the icons show up if the player has rank 4 equipped.  

I'll leave this open until I get home from work tomorrow so I can check it again when I'm less sleep deprived.


Pewtro:
Thanks!

PR #3481, opened 2020-02-05
Corruption: Void Ritual

niseko:
image


Pewtro:
Once the tests clears, this looks good to me!

PR #3488, opened 2020-02-14
Apply hotfixes to corruptions

niseko:
Forgot this was hardcoded so it's a bit late


Pewtro:
Can you fix the corruption amount on avoidant at the same time? 🙏

niseko:
not sure what the values are since all the ones on the auction house appear to be from before the patch, since for whatever reason they're unchanged

niseko:
only took 14 minutes to find a 16% one

PR #3511, opened 2020-03-05
[MW] Average time between RSK casts stat for RM

FraunchToost:
Adds statistics module in the talents section for "Average time between RSK casts" when Rising Mist is talented.  Calculates time from start of encounter until final RSK cast event, then divides by total RSK casts.

image

Not sure if I followed naming conventions well enough, feel free to suggest better names/labels.  😄


src/parser/monk/mistweaver/modules/talents/AverageTimeBetweenRSKs.js
```diff
@@ -0,0 +1,63 @@
...
+import TalentStatisticBox from 'interface/others/TalentStatisticBox';
+import { STATISTIC_ORDER } from 'interface/others/StatisticBox';
+import Events from 'parser/core/Events';
+
+/* Add in Statistic box to show average time between RSK casts when Rising Mist is talented.
+*/
+
+class TimeBetweenRSKs extends Analyzer {
+  static dependencies = {
+    abilityTracker: AbilityTracker,
```
abelito75: You don't use this any where.

src/parser/monk/mistweaver/modules/talents/AverageTimeBetweenRSKs.js
```diff
@@ -0,0 +1,63 @@
...
+    this.active = this.selectedCombatant.hasTalent(SPELLS.RISING_MIST_TALENT.id);
+    if(!this.active){
+      return;
+    }
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.RISING_SUN_KICK), this.onRSK);
+  }
+
+  onRSK(event) {
+    const spellId = event.ability.guid;
+    if (SPELLS.RISING_SUN_KICK.id !== spellId) {
```
abelito75: With event listener this function will only be called even the cast is rising sun kick so this check is redundant

src/parser/monk/mistweaver/modules/talents/AverageTimeBetweenRSKs.js
```diff
@@ -0,0 +1,63 @@
...
+      return;
+    }
+    this.totalRSKCasts += 1;
+    this.rskEvents.push(event.timestamp);
+  }
+
+  get lastRSKTimestamp() {
+    return this.rskEvents[this.totalRSKCasts-1];
```
abelito75: If you only every care about the last rsk you don't need to save all of them. You could just save 1 and keep over riding it till the end of the fight.

src/parser/monk/mistweaver/modules/talents/AverageTimeBetweenRSKs.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  get rskWindow () {
+    return this.lastRSKTimestamp - this.owner.fight.start_time;
```
abelito75: Not a true average time between casts as if they don't cast rsk at the start then the number will be larger than indented. You'd need to do this.lastRSKTimestamp - this.rskEvents[0].timestamp;

PR #3512, opened 2020-03-06
[StatTracker] Stat multipliers

### LINKED ISSUE #2085: StatTracker: Armor Specialization
Buffs that get applied / removed which give a primary stat are increased by 5% by the Armor Specialization, which is included in the pull-stats, but currently not added for temporary buffs.

### LINKED ISSUE #2086: StatTracker: Primary Stat raid buffs

niseko:
Base for further multiplier stuff, split up so its easier to review.
This will allow temporary buffs add the correct stat amounts as they do get buffed by things like armor specialization, arcane intellect and stat% corruptions.

This PR will mostly change secondary stat values as it increases the value of intellect vs secondaries.

resolves #2086, resolves #2085


src/parser/shared/modules/StatTracker.js
```diff
@@ -684,9 +724,24 @@ class StatTracker extends Analyzer {
       debug && this._debugPrintStats(this._currentStats);
       this._triggerChangeStats(event, before, delta, after);
     }
+    if (statMult) {
+      // ignore prepull buff application, as they're already accounted for in combatantinfo
```
emallson: oh, you addressed that here. how does this interact with dying and getting AI/shout after res?

niseko: Dying clears all buffs, meaning this function gets called with `event.newStacks = 0` and 
`event.oldStacks = 1`. ` this.removeStatMultiplier(statMult, true)` will then remove the multiplier as well as the stats provided by that multiplier, and if someone re-applies the buff after reviving, it will be added to the current stats again.

src/parser/shared/modules/StatTracker.js
```diff
@@ -373,6 +373,13 @@ class StatTracker extends Analyzer {
       ...this._pullStats,
     };
 
+    // Really hoping people don't run around with wrong armor types
```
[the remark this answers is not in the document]
niseko: I saw no easy way to check all their items armor types, so I left it at that. might be worth revisiting in general as a checklist item as well

Issue #2246, opened 2018-09-20
[Havoc Demon Hunter] Azerite Traits


joshinat0r:
Havoc Demon Hunter specific azerite traits

- [ ] Burning Soul
- [ ] Eyes of Rage
- [ ] Furious Gaze
- [ ] Infernal Armor
- [ ] Revolving Blades
- [ ] Seething Power
- [ ] Soulmonger
- [x] Thirsting Blades
- [ ] Unbound Chaos

torothin:
After looking into this issue I have found that the following traits are done:
Eyes of Rage
Furious Gaze
Revolving Blades

torothin:
Thirsting Blades is being worked on.

PR #3518, opened 2020-03-09
[Havoc Demon Hunter] Azerite Traits #2246 -Thirsting Blades Trait

torothin:
Added Thirsting Blades Azerite trait for Havoc DH.


.env.local.example
```diff
@@ -1,3 +0,0 @@
-# Copy this file and rename it to `.env.local` to configure your own WCL API key. You can get your own API key from the bottom of the Warcraft Logs settings: https://www.warcraftlogs.com/profile
```
niseko: please don't delete this file as its used as an example

src/parser/demonhunter/havoc/CHANGELOG.js
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
 
-import { Mamtooth, Yajinni, Zerotorescue, Viridis } from 'CONTRIBUTORS';
+import { Mamtooth, Yajinni, Zerotorescue, Viridis, Torothin } from 'CONTRIBUTORS';
```
niseko: add yourself to the CONTRIBUTORS.js file as well

src/parser/demonhunter/havoc/modules/spells/azeritetraits/ThirstingBlades.js
```diff
@@ -0,0 +1,105 @@
...
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER), this.onApplyBuff);
+    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER), this.onRemoveThirstingBlades);
+
+    const { damage } = azeriteTraitStats(this.selectedCombatant.traitsBySpellId[SPELLS.THIRSTING_BLADES.id]);
+    this.damagePerThirstingBlades = damage;
+
+  }
+
+  onApplyBuff(event) {
+    if(event.ability.name === "Thirsting Blades") {
```
niseko:
```suggestion
    if(event.ability.guid === SPELLS.THIRSTING_BLADES.id) {
```
We usually check for abilities by using their ID as that won't run into issues like uniqueness. 

You can alternatively attach to the spell you're looking for to the event listener itself 
(like `this.addEventListener(Events.applybuffstack.by(SELECTED_PLAYER).spell(SPELLS.THIRSTING_BLADES), this.onApplyBuffStack);`
instead of
`this.addEventListener(Events.applybuffstack.by(SELECTED_PLAYER), this.onApplyBuffStack);` in row 38)
and then you don't have to check for the correct spell at all.

torothin: fixed this.

src/parser/demonhunter/havoc/CHANGELOG.js
torothin: added

.env.local.example
torothin: Forgot I didn't make a copy just renamed so added the file back

Issue #2247, opened 2018-09-20
[Vengeance Demon Hunter] Azerite Traits


joshinat0r:
Vengeance Demon Hunter specific azerite traits

- [ ] Burning Soul
- [ ] Cycle of Binding
- [ ] Essence Sever
- [ ] Gaping Maw
- [ ] Infernal Armor
- [ ] Revel in Pain
- [ ] Rigid Carapace
- [ ] Soulmonger

torothin:
Going to work Infernal Armor first then Revel in pain based on the popularity of those two traits.

PR #3523, opened 2020-03-14
Infernal Armor feature added per issue #2247

torothin:
This feature adds infernal armor to vengeance dh.  This fix could work for havok if I could find logs to test.  All vengeance tests produced damage number equal to warcraft logs and armor values as expected compared to wowhead azerite scaling.


src/parser/demonhunter/vengeance/modules/spells/azeritetraits/InfernalArmor.js
```diff
@@ -0,0 +1,111 @@
...
+    const damageThroughputPercent = this.owner.getPercentageOfTotalDamageDone(this.damage);
+    const dps = this.damage / this.owner.fightDuration * 1000;
+    const buffUptimePercent = this.buffUptime / this.owner.fightDuration;
+
+    return (
+      <AzeritePowerStatistic
+        size="flexible"
+      >
+        <BoringSpellValueText spell={SPELLS.INFERNAL_ARMOR}>
+          <img
```
emallson: You should use the `<ItemDamage />` element for this.

torothin: Added <ItemDamageDone /> for the DPS part but didn't see any options for the armor part so left it as is.

Pewtro: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/shared/modules/spells/bfa/azeritetraits/Gemhide.js#L90 
Use this for armor

PR #3547, opened 2020-03-30
Torothin arms warrior overpower

torothin:
Added a statistic to show the wasted procs which doesn't change the logic from the original component.  Also added two logs for testing to show 0 and >0 overpower wasted.

### DIFF
--- a/src/parser/warrior/arms/modules/core/Overpower.js
+++ b/src/parser/warrior/arms/modules/core/Overpower.js
@@ -10,6 +10,14 @@ import StatisticBox, { STATISTIC_ORDER } from 'interface/others/StatisticBox';
 import ExecuteRange from './Execute/ExecuteRange';
 import SpellUsable from '../features/SpellUsable';
 
+/**
+ * Logs used to test:
+ * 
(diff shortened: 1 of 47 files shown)


src/parser/warrior/arms/modules/core/Overpower.js
```diff
@@ -60,6 +71,28 @@ class OverpowerAnalyzer extends Analyzer {
...
+        position={STATISTIC_ORDER.OPTIONAL(6)}
+        value={(
+          <>
+              {this.wastedProc} <small>wasted buffs</small><br />
+              {this.overpowerCasts} <small>total casts</small>
+          </>
+        )}
+        tooltip={(
+          <>
+            If Overpower is casted with two buffs applied
```
emallson, line 89: as a non-warrior it is not obvious to me what buffs you're talking about. would this be obvious to a new arms player?

this is the only issue i see with the PR

torothin: I re-wrote the tooltip to hopefully be more clear.

PR #3550, opened 2020-04-09
Simple adjustments to Enhancement Shaman

djanse:
I've made some changes to enhancement shaman and plan to do more in the future.
The summare is that as following;
* Decreased damage gained for Forceful Winds from 100% to 80%.
* Fixed Lightning Shield not showing up anywhere.


src/parser/shaman/enhancement/modules/talents/LightningShield.js
```diff
@@ -48,29 +47,23 @@ class LightningShield extends Analyzer {
...
   suggestions(when) {
-    when(this.suggestionThresholds).addSuggestion((suggest, actual) => {
-      return suggest(<>You should fully utilize your <SpellLink id={SPELLS.LIGHTNING_SHIELD_TALENT.id} /> by using it before combat.</>)
-        .icon(SPELLS.LIGHTNING_SHIELD_TALENT.icon)
-        .actual(<>You kept up <SpellLink id={SPELLS.LIGHTNING_SHIELD_TALENT.id} /> for ${formatPercentage(actual)}% of the fight.</>)
-        .recommended(<>It is possible to keep up <SpellLink id={SPELLS.LIGHTNING_SHIELD_TALENT.id} /> for 100% of the fight by casting it pre-combat.</>);
-    });
+    const lightningUptime = this.selectedCombatant.getBuffUptime(SPELLS.LIGHTNING_SHIELD_TALENT.id) / this.owner.fightDuration;
+
+    when(lightningUptime).isLessThan(0.95)
```
Pewtro, line 53: You should use suggestionThresholds for suggestions as it will then be consistent with other modules

src/parser/shaman/enhancement/modules/talents/LightningShield.js
```diff
@@ -1,45 +1,44 @@
...
+      case SPELLS.LIGHTNING_SHIELD.id:
+        return true;
+      default:
+        return false;
+    }
+  }
+
   on_byPlayer_damage(event) {
-    if (event.ability.guid!==SPELLS.LIGHTNING_SHIELD_TALENT.id) {
+    if (this.isEventLightningShield(event) === false) {
```
Pewtro, line 29: is there a reason to add more lines of code to achieve the same functionality, and make it non-consistent with other modules?

src/parser/shaman/enhancement/modules/talents/ForcefulWinds.js
```diff
@@ -5,12 +5,12 @@ import { formatNumber, formatPercentage } from 'common/format';
...
 
-import StatisticBox from 'interface/others/StatisticBox';
+import TalentStatisticBox from 'interface/others/TalentStatisticBox';
 import calculateEffectiveDamage from 'parser/core/calculateEffectiveDamage';
 
 
 const FORCEFUL_WINDS = {
-  INCREASE: 1,
+  INCREASE: 0.8,
 };
```
Pewtro, line 14: Shouldn't this just be a variable and not an object? There seems to be nothing else related to this so just naming it ``const FORCEFUL_WINDS_INCREASE = 0.8`` would be more consistent

djanse: I agree, it was already like than once i touched the file and didn't consider changing it.

src/parser/shaman/enhancement/modules/talents/LightningShield.js
djanse: I didn't know that suggestionThresholds was used across the system. I returned the get function

src/parser/shaman/enhancement/modules/talents/ForcefulWinds.js
```diff
@@ -5,13 +5,10 @@ import { formatNumber, formatPercentage } from 'common/format';
...
 
-import StatisticBox from 'interface/others/StatisticBox';
+import TalentStatisticBox from 'interface/others/TalentStatisticBox';
 import calculateEffectiveDamage from 'parser/core/calculateEffectiveDamage';
 
-
-const FORCEFUL_WINDS = {
-  INCREASE: 1,
-};
+const FORCEFUL_WINDS_INCREATE = 0.8;
```
Pewtro, line 11: Spelling :) Last thing before it looks good to go (from my pov)

src/parser/shaman/enhancement/modules/talents/LightningShield.js
djanse: I was kind of getting used to how the event handlers worked. This has been changed. in the current version. Newest verion which i'm working on works with typescript. (Credits to Hawk)

src/parser/shaman/enhancement/modules/talents/ForcefulWinds.js
djanse: Has been adjusted

PR #3553, opened 2020-04-13
Added statistics for Arcanic Pulsar azerite trait

djanse:
Added statistics for Arcanic Pulsar azerite trait.


src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+    );
+
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER)
+        .spell(SPELLS.STARSURGE_MOONKIN),
+      this.onStarsurgeDamage,
+    );
+  }
+
+  onStarsurgeCast() {
```
Sharrq, line 60: add an event type to this

`onStarsurgeCast(event: CastEvent)`

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+      return;
+    }
+
+    const buff = this.selectedCombatant.getBuff(SPELLS.ARCANIC_PULSAR_BUFF.id);
+    if (!buff) {
+      this.celestialAlignmentCount++;
+    }
+  }
+
+  onStarsurgeDamage() {
```
Sharrq, line 72: event type

`onStarsurgeDamage(event: DamageEvent)`

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+        .spell(SPELLS.STARSURGE_MOONKIN),
+      this.onStarsurgeDamage,
+    );
+  }
+
+  onStarsurgeCast() {
+    if (this.firstStarsurgeCast) {
+      this.firstStarsurgeCast = false;
+      return;
+    }
```
Sharrq, line 64: I dont know Boomkin, but what is the reasoning behind needing to exclude the first Starsurge of the fight, and only the first Starsurge of the fight?

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const buff = this.selectedCombatant.getBuff(SPELLS.ARCANIC_PULSAR_BUFF.id);
+    if (!buff) {
+      this.celestialAlignmentCount++;
+    }
```
Sharrq, line 69: are you just checking to see if they have the buff or not? Cause you can use hasBuff instead of getBuff for that.

```
if (!this.selectedCombatant.hasBuff(SPELLS.ARCANIC_PULSAR_BUFF.id) {
    this.celestialAlignmentCount++;
}
```

src/parser/druid/balance/CombatLogParser.js
```diff
@@ -78,6 +79,7 @@ class CombatLogParser extends MainCombatLogParser {
     highNoon: HighNoon,
     powerOfTheMoon: PowerOfTheMoon,
     streakingStars: StreakingStars,
+    ArcanicPulsar: ArcanicPulsar,
```
Sharrq, line 82: `arcanicPulsar: ArcanicPulsar,`

camelcase

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  from 'interface/statistics/components/BoringSpellValueText';
+import ItemDamageDone from 'interface/ItemDamageDone';
+import { calculateAzeriteEffects } from 'common/stats';
+
+class ArcanicPulsar extends Analyzer {
+  /**
+   * Starsurge's damage is increased by 2181. Every 9 Starsurges, gain
+   * Celestial Alignment for 6 sec.
+   *
+   * Example Log:
```
Sharrq, line 19: Did you mean to add an exmaple log here?

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
djanse: This is some left of code from testing the stack count on the buff. changed it to hasBuff.

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
djanse: This is a hotfix counting the amount of times that Celestial Alignment has been cast. For some reason i cannot call the stacks of the buff using getBuff(), it throws me an error.

Pewtro: This is because it's applied on damage and not on cast - you should move the checks around to account for that.

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  onStarsurgeCast(event: CastEvent) {
+    if (this.firstStarsurgeCast) {
+      this.firstStarsurgeCast = false;
+      return;
+    }
+
+    if (!this.selectedCombatant.hasBuff(SPELLS.ARCANIC_PULSAR_BUFF.id)) {
+      this.celestialAlignmentCount++;
+    }
```
Pewtro, line 69: Isn't this wrong? That's not how many times Celestial Alignment was triggered? 
Shouldn't this be tracked through a apply_buff event?

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    this.bonusDamage
+      = this.selectedCombatant.traitsBySpellId[SPELLS.ARCANIC_PULSAR_TRAIT.id]
+      .reduce((total, rank) => {
+        const [damage] = calculateAzeriteEffects(
+          SPELLS.ARCANIC_PULSAR_TRAIT.id,
+          rank,
+        );
+        return total + damage;
+      }, 0);
```
Pewtro, line 46: You can't calculate gained damage like that. It's much more intricate to calculate the actual gained damage, see my implementation of something similiar here: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/hunter/beastmastery/modules/spells/azeritetraits/FeedingFrenzy.tsx#L133

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
djanse: Considering the Arcanic Pulsar buff is applied starting at one stack, and dissapears at 9 stacks. It would indeed be better to use a applybuff or applybuffstack. However the issue is that the buff will never reach 9 stacks, instead it will just disappear when you cast Starsurge at 8 stacks. And Celestial Alignment will be activated.

The best way imo would be to check if the buff has 8 stacks when Starsurge is cast however, [this.selectedCombatan.getBuff(...).stacks](url) will return an error, even though the console.log output shows its part of the object.

djanse: Maybe i should use: floor(STARSURGE_CAST_COUNT / 9) ?

abelito75: Does the applybuffstack not work for this?

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
djanse: Interesting aproach, Took me some time to figure out what your 2nd used coefficient was (FEEDING_FRENZY_DAMAGE_COEFFICIENT)  since that was changed since it was changed to 0.216. in v8.2.

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
Sharrq: getBuff(SPELLS).stacks will error if they dont have the buff on them. 

So you can do
buff = this.selectedCombatant.getBuff(SPELLS)
if (buff && buff.stacks > x)

But what exactly are you trying to get here? The total number of Celestial Alignment Buffs? (Casts and Procs), or arcanic pulsar procs?? Your variable is saying one thing and your if is saying another

djanse: After discussing with abelito on how to approach this way in the best manner, i've determined that storing the stacks is the best option in order to avoid collision with regular CA casts and VOP procs.

Dambroda: To explain how this works: there are two primary times you *wont* have the Arcanic Pulsar buff when you cast starsurge (SS): the first SS of the fight, and the first SS you cast right after getting the celestial alignment (CA) proc from the trait. With the firststarsurgecast boolean, the first case is eliminated, and you are left with only the casts after you get a proc. Thus, you can increment your CA proc counter. There are some other fringe cases that aren't accounted for here that would lead to some inaccuracies.. for example, getting battle res'd (in which case this would count an *extra* proc), or the fight/phase ending before you cast SS again after getting the proc (in which case this would count one too few procs).

All in all it's a pretty backwards way logically to solve the problem, but I appreciate the ingenuity of the workaround :) That said, Sharrq is probably correct in saying the error you were getting originally was just that you weren't checking for the buff actually existing before trying to get the stack count. If that is the case, going back to the original idea will be far far cleaner and clearer:

> The best way imo would be to check if the buff has 8 stacks when Starsurge is cast however, this.selectedCombatan.getBuff(...).stacks has thus far always provided me with an error

djanse: > To explain how this works: there are two primary times you _wont_ have the Arcanic Pulsar buff when you cast starsurge (SS): the first SS of the fight, and the first SS you cast right after getting the celestial alignment (CA) proc from the trait. With the firststarsurgecast boolean, the first case is eliminated, and you are left with only the casts after you get a proc. Thus, you can increment your CA proc counter. There are some other fringe cases that aren't accounted for here that would lead to some inaccuracies.. for example, getting battle res'd (in which case this would count an _extra_ proc), or the fight/phase ending before you cast SS again after getting the proc (in which case this would count one too few procs).

I think i've already worked around all of the possible issues that could occour with having an in accurate reading of the AP stacks and CA procs. Including the inaccuracy of being CRed

> All in all it's a pretty backwards way logically to solve the problem, but I appreciate the ingenuity of the workaround :) That said, Sharrq is probably correct in saying the error you were getting originally was just that you weren't checking for the buff actually existing before trying to get the stack count. If that is the case, going back to the original idea will be far far cleaner and clearer:

Interestingly enough i also tought that the error was happening because the buff = undefined at some times. And had already checked if the buff existed in the same way Sharqq proposed. However the error seemed to presist; http://prntscr.com/segzc5

Dambroda: Do you have a log for this?

Dambroda: Glad this got resolved! We figured out it was just a Typescript error...

djanse: Thanks again for your time! Removed some unneccesary code <3

src/parser/druid/balance/modules/talents/azeritetraits/ArcanicPulsar.tsx
Pewtro: Your solution is still lacking versatility scaling, tauren scaling and anything else that would affect spell damage (such as boss armor etc)

Pewtro: See here for a longer, more commented out version than mine: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/druid/feral/modules/azeritetraits/WildFleshrending.js

djanse: Do you mean the version in the latest commits, on the one when you first reviewed that part of code?

Sharrq:
I will let Putro review/approve it. Im fine with everything I saw previously (except the multiline event listeners and such that i still hate).

But @Pewtro had questions about vers and such, so ill let him approve it or request more changed.

Pewtro:
I don't think that these calculations will match actual damage numbers in-game, so they would need to be updated to match those numbers.

djanse:
> I don't think that these calculations will match actual damage numbers in-game, so they would need to be updated to match those numbers.

Within the latest commit [437518a](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3553/commits/437518a9bf15605d24ca7e42c55809ef7ce3e1a9) the damage seems to be correct with what the blizzard tooltip suggests, including scaling of vers and mastery or any other buff that increases the damage.

My calculations can be found here: [Arcanic Pulsar Calc](https://docs.google.com/spreadsheets/d/1zrLYEXy0Ny0FOPYcNhmWSdAwsGem-1Shu74f-Y3hJCc/edit?usp=sharing)
In the first of calculations i add the base damage on top of the base damage of starsurge which is Spell Power * 230%. I than apply the calculations that WoW applies as well. (Mastery, Vers, buffs etc.)

The second set of calculations is just the starsurge damage without the arcanic pulsar. Which results into a lower amount.

The third set is the calculations for the trait itself.

-------

What the function 'calculateBonusAzeriteDamage' currently does is take the base trait damage, check how much of that is part of the total raw damage in percentage. This means that any scaling of Spell Power is applied, resulting in different values when intellect (SP) is increased. (Which i've debugged locally).

Any other multiplication be it; Chaos Brand, Moonkin Form, Torrent of Elements, A crit hit. should not effect scale which was calculated since it would apply to both the trait damage and starsurge damage.

I hope my explanation was clear enough.

Pewtro:
If you verify them on individual hits on logs, then it's fine. 
Tooltips can't be trusted.

djanse:
> If you verify them on individual hits on logs, then it's fine.
> Tooltips can't be trusted.

It is as accurate as it can get. When I recalculate the total starsruge damage that it should have dealth with the current buffs and sp/mastery/vers (at the moment of cast/damage). and compare it to the damage dealt it matches for 98% with the calculations. I'm not sure what causes the slight difference however.

Dambroda:
The logic was fixed in calculating when the procs happened, and the damage calculation seems to have been updated to be on par with how every other azerite trait is currently being calculated. Not seeing any other issues (except the multiline formatting Sharqq mentioned.. which I think is out of the scope of individual PR's at this point), good work!

PR #3563, opened 2020-04-19
Reduced avatar and article image file sizes

djanse:
I had noticed that loading the homepage of wowanalyzer loaded several large images. In this PR i've looked over the avatar images and several article images. The following actions have been performed;

- Convert from PNG to JPG if no alpha pixels could be found.
  - Only if the image ended up being smaller
- Resize avatars to 256x256 if they were larger than this
  - The current maximum display size for avatars is 200px in height, correct me if i'm wrong.
  - This should make rendering images be easier on the browser.
- Check if avatars larger than 30kb could've been optimized in file size.

When i loaded the homepage on my development server, the total downloaded size was 14.2Mb. After reducing the images sizes this went down to 9.7Mb. Resulting in ~4.5 MB in bandwidth savings. This is a huge deal especially for mobile 3G users or people with unusual slow internet.


djanse:
I'm at a loss as to why two tests failed for this pull request. Any ideas?

emallson:
the tests fail because we use snapshot tests to check for unintended changes. The snapshots include file paths, so the change from .png to .jpg triggers them.

MartijnHols:
If you run the tests using `yarn test -u`, it should update the snapshots and fix the pipelines. I think this should be good to go with that fix.

MartijnHols:
Thanks! :D

PR #3569, opened 2020-05-04
[BrM] Fixed CF Absorbs

### LINKED ISSUE #3562: [Brewmaster] Celestial Fortune value in Mit Values table is incorrect.
This bug has two parts:

1. BrM is missing the 5% crit passive in the stat tracker
2. Absorbs are no longer showing up in CF bonus healing.

The first is a trivial fix, the second may be more involved.

emallson:
Fixes #3562

The main problem was that WCL stopped adding the `absorb` field to `applybuff` events, so no absorbs were getting recorded.

Also added BrM's base +5% crit and removed the scaled rating columns from the mitigation values table.


Pewtro:
Looks good to me at initial glance, but it's out-of-date with base branch and can't do anything to fix it up :o

PR #3576, opened 2020-05-12
Resolved issues where intellij IDEs would wrap long lines of code.

djanse:
I've had it happen on multpile PRs that the reviewer would tell me that I should not be wrapping my long lines of code in most of the cases. This happened because the project settings stored in the .idea project folder were set to wrap at 80 characters.

This PR has removed this enforced wrapping from the code style stored in the ide project folder.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -24,6 +24,7 @@
       <option name="USE_DOUBLE_QUOTES" value="false" />
       <option name="FORCE_QUOTE_STYlE" value="true" />
       <option name="ENFORCE_TRAILING_COMMA" value="WhenMultiline" />
+      <option name="IMPORTS_WRAP" value="0" />
       <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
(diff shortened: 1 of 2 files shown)


Pewtro:
I think it's worth keeping the margin, just to not use the auto-formatting in some files. It makes sense to have the wrapping in the vast majority of files in the codebase.

djanse:
In the last commit I've disabled import wrapping seperatly and reset set the line wrapping as following; Visual guideline at 80 characters and wrap at 999 characters. These are the settings that .js files had.

Abilities wrapping seemed to have been an issue with in typescript files where inline if-statements (and some other things) would wrap. Within the latest commit such a thing should not happen.

I've tried to apply the current set codestyles onto CHANGELOG files but no wrapping seems to be taking place.

Pewtro:
Why are you removing all these formatting rules? A lot of those are helping enforce the style guide and should not be removed... https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Meta:-Style-guide

Pewtro:
Made the changes more sane in #3578 in conversation with Vetyst.

PR #3577, opened 2020-05-12
Disallow ++ and -- outside of for loops

Pewtro:
Enforces the style-guide more accurately: https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Meta:-Style-guide

The only commit I am a bit worried about is the last one, I couldn't get no-plusplus to allow the lines where we had something like the below codesnippet, which only happened in our azerite extract tool which is going to be gone in SL, so I think it's okay to add it to eslintignore for now. 
```
for (i = 0, j = l; i < l; i++, j--) {
    doSomething(i, j);
}
```

It's just weird because eslint says that what I added should allow for it..

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ node_modules
 /src/localization/_build/
 /src/localization/*/*.js
 /scripts/localization/*.js
+/scripts/azerite/extract.js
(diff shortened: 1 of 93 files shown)


djanse:
Went over all changes of ++ and -- conversion. I didn't come across any issues where a ++ turned into a -= 1 or anything of the likes.

PR #3584, opened 2020-05-19
Updated Lightning Shield suggestion and damage display.

djanse:
I noticed that the Lightning Shield JS file was still around. This has been deleted. I've also re-evaluated the suggestion that was being displayed, this should now be less confusing. In addition absorbes were already included TSX version of Lightning Shield.


src/parser/shaman/enhancement/modules/talents/LightningShield.tsx
```diff
@@ -7,106 +7,94 @@ import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
...
   }
 
   get suggestionThresholds() {
     return {
       actual: this.getLightningShieldUptime,
       isLessThan: {
-        major: 1,
+        minor: 1,
+        major: 0.95,
+        recommended: 1.00,
```
Sharrq, line 66: You dont need to specify recommended in the thresholds. Just list the thresholds (Minor, Average, Major), recommended gets populated with minor automagically.

src/parser/shaman/enhancement/modules/talents/LightningShield.tsx
```diff
@@ -7,106 +7,94 @@ import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
...
         category={'TALENTS'}
       >
-        <BoringSpellValueText
-          spell={SPELLS.LIGHTNING_SHIELD_TALENT}
-        >
-          <>
-            <ItemDamageDone amount={this.damage} />
-          </>
+        <BoringSpellValueText spell={SPELLS.LIGHTNING_SHIELD_TALENT}>
+            <ItemDamageDone amount={this.damageGained} />
```
Sharrq, line 97: Personally i would consider including other information like the Uptime considering it is a threshold for a suggestion. 

Also idk if you are using ItemDamageDone because you think you have to or because you like the layout of it, but you can lay it out pretty much however you want, as an example, you could do something like this....

```
<BoringSpellValueText spell={SPELLS.LIGHTNING_SHIELD_TALENT}>
    <>
        <ItemDamageDone amount={this.damageGained} /><br />   
        <UptimeIcon /> {formatPercentage(this.getLightningShieldUptime,0)}% <small>buff uptime</small>
    </>
</BoringSpellValueText>
```

The end result would look like this (if you do this, make sure you change the size of the stat box to "flex" instead of small
image

Or you could display the raw damage done instead of DPS, or you could display both, or you could display the important stuff and put the rest in a tooltip ... pretty much whatever you want. There are plenty of examples around the other specs.

Sharrq: Ignore me, i didnt know that lightning shield was a 1 hour buff. Disgreard

PR #3586, opened 2020-05-24
Generic Hot tracker and Rising Mist update

abelito75:
I made the hottracker we have generic and added a few hooks in there for some of the fancy things that might happen in game such as crit, dynamic length hots with max durations, custom duration based on events,if they bounce or not, and tracking healing from the original duration ends. Rising mist has also been updated to more accurate capture the hot fixes that happened a while ago. There will most likely be another rising mist pr soon.


src/parser/monk/mistweaver/modules/core/HotTrackerMW.js
```diff
@@ -0,0 +1,64 @@
...
+    return 20000 * 2;
+  }
+
+  calculateMaxRemDuration(event){
+    const spellId = event.ability.guid;
+    if(event.selectedCombatant.hasBuff(SPELLS.THUNDER_FOCUS_TEA.id) && spellId === SPELLS.RENEWING_MIST_HEAL.id){
+      return 30000;
+    }
+    return 20000;
+  }
```
Pewtro, line 32: calculateMaxDuration and calculateMaxRemDuration have a lot of hardcoded MS values, maybe they should be moved to consts to follow the rest of the codebase in that regard of not having magic numbers?

src/parser/shared/modules/HotTracker.js
```diff
@@ -87,37 +62,87 @@ class HotTracker extends Analyzer {
...
     }
+
+    hot.attributions.forEach(att => {
+      att.healing += healing;
+    });
+    hot.boosts.forEach(att => {
+      att.healing += calculateEffectiveHealing(event, att.boost); 
+    });
+
+    if(event.hitType === HIT_TYPES.CRIT && hot.effectOnCrit){//if the heal crits and it does something fancy on crits do that
```
abelito75, line 139: review on my own thing for memory. 
should be `this.hotInfo[spellId].effectOnCrit` on `hot.effectOnCrit`

PR #3588, opened 2020-05-25
Converted remaining Enhancement Shaman files to typescript.

djanse:
I've converted the remaining files to typescript.

PR #3589, opened 2020-05-25
Enhancement cleanup

djanse:
Note: This depends on PR #3588 

I've cleaned-up the enhancement shaman modules so that they are better readable. They are now also on par with each other with namings of properties and placement of comments/methods. This PR has (should not atleast) no functional changes. the code was just re-ordered/cleaned.


src/parser/shaman/enhancement/modules/core/FlametongueRefresh.tsx
```diff
@@ -1,33 +1,36 @@
...
+  protected flametongueTimestamp: number = 0;
+  protected flametongueCasts: number = 0;
+  protected earlyRefresh: number = 0;
 
   constructor(options: any) {
     super(options);
-    this.active
-      = !this.selectedCombatant.hasTalent(SPELLS.SEARING_ASSAULT_TALENT.id);
+
+    this.active = !this.selectedCombatant.hasTalent(SPELLS.SEARING_ASSAULT_TALENT.id);
```
Sharrq, line 20: Is it intended to mark this as active if they DONT have Searing Assault?

src/parser/shaman/enhancement/modules/core/Rockbiter.tsx
```diff
@@ -34,9 +35,7 @@ class Rockbiter extends Analyzer {
   }
 
   onEnergize(event: EnergizeEvent) {
-    if (event.classResources &&
-      event.classResources[0].amount >=
-      MAELSTROM_THRESHOLD) {
+    if (event.classResources && event.classResources[0].amount >= MAELSTROM_THRESHOLD) {
```
Sharrq, line 38: Zero doesnt want people using classResources[0]

https://github.com/WoWAnalyzer/WoWAnalyzer/issues/2387

src/parser/shaman/enhancement/CHANGELOG.tsx
```diff
@@ -6,6 +6,7 @@ import SpellLink from 'common/SpellLink';
 import { change, date } from 'common/changelog';
 
 export default [
+  change(date(2020, 4, 25), <>Converted remaining Enhancement Shaman files to typescript and cleaned its modules.</>, [Vetyst]),
```
Sharrq, line 9: date is wrong. (Its May not April)

src/parser/shaman/enhancement/modules/core/FlametongueRefresh.tsx
djanse: For the current support this is correct. Usually what you do with Flametongue is refresh your own buff, which has a 16 second duration. However, this talent will force you to also refresh a debuff on your target which lasts for 6 seconds. As you can see this will conflict with each other.

src/parser/shaman/enhancement/CHANGELOG.tsx
djanse: This is resolved in PR #3588 which the current changes are based on. I think it will be wiser to have that merged before this PR.

src/parser/shaman/enhancement/modules/core/Rockbiter.tsx
djanse: I've changed this into a classResources.find

PR #3590, opened 2020-05-26
Fixed garrote early refresh tracking.

djanse:
Seems the convertion in #3580 was wrong for the GarroteNormalizer. I'll take the time to look through all the changes tomorrow afternoon (EU) in that PR again to make sure no other modules suffer from this.


Sharrq:
Did you test this with the log that that guy posted to confirm it was working properly?

Dambroda:
Merging because it fixes a mistake in a previous PR, even if it doesnt fully solve the new issue (but it probably does anyway)

Issue #1999, opened 2018-07-26
[Discipline Priest][BFA] Azerite Traits for Discipline


rp4rk:
https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/%5BBFA%5D-Implementing-Azerite-Traits

- [ ] http://bfa.wowhead.com/spell=278902/contemptuous-homily
- [x] http://bfa.wowhead.com/spell=275542/depth-of-the-shadows
- [ ] http://bfa.wowhead.com/spell=277680/gift-of-forgiveness
- [ ] http://bfa.wowhead.com/spell=272775/moment-of-repose
- [ ] http://bfa.wowhead.com/spell=269555/torment-of-torments
- [ ] http://bfa.wowhead.com/spell=273308/weal-and-woe
- [ ] http://bfa.wowhead.com/spell=278643/enduring-luminescence

PR #3594, opened 2020-05-27
Enduring Luminescence Azerite trait analyzer

Tiphess:
Hello, this is my first attempt at a contribution. 
I based myself on this issue https://github.com/WoWAnalyzer/WoWAnalyzer/issues/1999 to see that a few azerite traits for Discipline were not implemented. I based a lot of this code with the existing Depth of the Shadows module since these traits are mechanically similar (Both extend atonements duration and have a direct healing increase), so I figured it would be an easy way to start contributing.

I also noticed that the Atonement coefficient did not match the 8.3 nerf to 50%, so I adjusted that.

This is the result:
EnduringLuminescence_stat_box


src/parser/priest/discipline/modules/azeritetraits/EnduringLuminescence.js
```diff
@@ -0,0 +1,159 @@
...
+
+    _lastCastIsPowerWordRadiance = false;
+    _atonementsAppliedByRadiances = [];
+
+    constructor(...args) {
+        super(...args);
+        this._maxTargets = 5;
+        this.active = this.selectedCombatant.hasTrait(SPELLS.ENDURING_LUMINESCENCE.id);
+
+        if (!this.active) {
```
abelito75, line 41: Point less code if you don't use actionlisteners. 
Please update to use actions listeners

src/parser/priest/discipline/modules/azeritetraits/EnduringLuminescence.js
```diff
@@ -0,0 +1,159 @@
...
+
+    on_byPlayer_cast(event){
+        const spellId = event.ability.guid;
+
+        if(spellId !== SPELLS.POWER_WORD_RADIANCE.id && spellId !== SPELLS.EVANGELISM_TALENT.id) {
+            return;
+        }
+
+        if (spellId === SPELLS.POWER_WORD_RADIANCE.id) {
+            console.log('Last cast is radiance');
```
abelito75, line 57: no console.log in production

Tiphess:
I have committed the requested changes. Also cleared up some methods names since the old ones became a bit redundant by adding the actionlisteners.

PR #3601, opened 2020-05-29
Add a workflow that deletes old artifacts using a CRON script

Pewtro:
This cron job should purge all existing artifacts we have lying around. 
After it has run, it should be adjusted to run much more rarely so we don't run cron jobs all the time. 

I hope this fixes the issue plagueing hunter now, as everything works locally for them, but not on production so I have no way of verifying if this is actually what is causing the issue or not. 

Here it is completing its run on my local branch: https://github.com/Pewtro/WoWAnalyzer/actions/runs/119061952

### DIFF
--- a/.github/workflows/deleteOldArtifacts.yml
+++ b/.github/workflows/deleteOldArtifacts.yml
@@ -1,0 +1,14 @@
+name: 'Delete old artifacts'
+
+on:
+  schedule:
+    - cron: '*/5 * * * *' # every 5 minutes - Should be set to something like 0 0 */14 * * for every 14 days, or maybe every 7 days?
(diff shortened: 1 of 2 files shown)


MartijnHols:
I'd rather not put this in as artifacts are necessary for jobs to run quickly. Can this be configured to only be activated manually?

Pewtro:
I saw some options where it would react to specific comments on pull requests/issues - like "Please delete all artifacts" or similiar, would that solve it?

Entirely manual triggering is not possible afaik.

MartijnHols:
Should no longer be necessary as the issue seems to have been fixed

PR #3606, opened 2020-06-01
Flametongue and Frostbrand changes

djanse:
The following has been changed to Flametongue and Frostbrand;

- Statistics now display both uptime and damage.
- Added Early refreshes suggestion for Frostbrand.
- Changed suggestions for Flametongue uptime, Flametongue early refreshes and Frostbrand uptime
- Reworked how early refreshes were determined.  The previous method did not take into account that the buff was extended by de remaining time up to 20.8s
- Reordered Natural Harmony to match blizzards tooltip order.


Dambroda:
Please include screenshots of new tooltips  :)

djanse:
Screenshot_1
Screenshot_2

djanse:
> Please include screenshots of new tooltips :)

Screenshot_3
Screenshot_4
Screenshot_5
Screenshot_6

djanse:
I wasn't sure which tooltip you were revering to so i delivered screenshots of everything i thought you could think of.

Dambroda:
Another question -- did you already look into extending one of these modules:

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/shared/modules/earlydotrefreshes/EarlyDotRefreshes.js

djanse:
> Another question -- did you already look into extending one of these modules:
> 
> https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/parser/shared/modules/earlydotrefreshes/EarlyDotRefreshes.js

No I haven't, I didn't even know it existed 😨. Allow me some time to look and play with it. I'll most likely change my PR to include that rather than my current solution. I'll change this to draft for the meantime. Thanks for the suggestion!

Dambroda:
As mentioned in discord, I read "early refresh" and assumed dot instead of buff. Dont think we have a core buff refresh module. I'll leave it up to you if you want to work on that or move forward with this version

djanse:
@Dambroda would you mind taking another look at this? I don't think i can get it any better without better understanding of certain aspects of wowanalyzer.

src/parser/shaman/enhancement/modules/checklist/Component.tsx
```diff
@@ -51,7 +51,7 @@ const EnhancementShamanChecklist = ({ castEfficiency, combatant, thresholds }: a
 
       <Rule
         name="Maintain your buffs"
-        description={"You should maintain your buffs in order to passivly increase your damage done to targets without refreshing them to early."}
+        description="You should maintain your buffs in order to passivly increase your damage done to targets without refreshing them to early."
```
Dambroda, line 54: I know this may not have been you, but I think it's important to fix the rule description:

"~~to~~ **too** early"

Also, the wording here feels a bit off overall. Maybe something like this?

"Maintaining your buffs passively increases your damage output. However, refreshing them too early leads to time wasted that could be spent on direct damage spells."

Pewtro:
Any update on addressing the PR comments, so we can look into getting this in, and is it even relevant with Shadowlands looming?

PR #3608, opened 2020-06-03
Removed duplicate entry of Guardian Spirit in defensive buffs.

djanse:
This caused a massive amount of errors (local development) to be generated when viewing death recaps (where guardian spirit was active on the player that died).
Aditionally, Guardian Spirit was shown as twice on the character that died eventhough that is not the case.

Before PR:
before-pr
before-pr

After PR:
after-pr
after-pr

[See Report](https://wowanalyzer.com/report/dfcJYaAg8Z7Q6KvW/8-Mythic+Maut+-+Kill+(5:26)/Part%C3%ADcle/standard/death-recap)

### DIFF
--- a/src/CHANGELOG.js
+++ b/src/CHANGELOG.js
@@ -9,6 +9,7 @@ import { change, date } from 'common/changelog';
 import Contributor from 'interface/ContributorButton';
 
 export default [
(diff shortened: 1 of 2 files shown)


Dambroda:
Thanks!

PR #3609, opened 2020-06-03
Disable Healthstone module if no warlock was found.

djanse:
Death recap would incorrectly suggest that a Healthstone was available to the user if a warlock was not present during that fight. 
In this PR I changed it so that the Healthstone module will only load if a warlock was found in the fight.

Before: 
before

After: 
after


src/parser/shared/modules/items/Healthstone.js
```diff
@@ -1,18 +1,37 @@
...
+import AbilityTracker from 'parser/shared/modules/AbilityTracker';
+import { Class } from 'parser/core/Events';
 import Potion from './Potion';
-
 /**
  * Tracks Healthstone cooldown.
  */
 class Healthstone extends Potion {
+  // Potion does not use Combatants, other dependencies are inherit from Potion.
+  static dependencies = {
```
Dambroda, line 15: I believe the preferred way to do this is


``` 
dependencies = {
    ...Potion.dependencies,
    combatants: Combatants
 }

```

Which saves you from having to import unused deps

src/parser/shared/modules/items/Healthstone.js
```diff
@@ -1,18 +1,37 @@
...
     SPELLS.HEALTHSTONE,
   ];
   static recommendedEfficiency = 0;
   static extraAbilityInfo = {
     isDefensive: true,
   };
+
+  get isAvailable() {
+    const players = Object.values(this.combatants.players);
+    return players.find(combatant => (combatant.spec.className === Class.Warlock));
```
Dambroda, line 33: Is `some` more intuitive than `find` here? Non-blocking of the PR, I personally just prefer to see something that doesn't rely on truthyness

src/parser/shared/modules/items/Potion.js
```diff
@@ -35,6 +35,12 @@ class Potion extends Analyzer {
 
   constructor(...args) {
     super(...args);
+
+    if (!this.isAvailable) {
+      this.active = false;
```
Dambroda, line 40: Can you check to see if this change works if you simply move the `active = false` into an extended constructor in `Healthstone`? Rather than creating a new function to extend. `active = false` should prevent all events from being called for the module and statistics from being displayed. But there is some other behind the scenes stuff to test with that as well.

djanse: I tried that before but unfortunately that didn't seem to have much effect on the ability being available. I even setup a constructor as following for the Healthstone;

```js
constructor(...args) {
  super(...args);
  this.active = false;
}
```

src/parser/shared/modules/items/Healthstone.js
djanse: What is the main difference between `find` and `some`?

src/parser/shared/modules/items/Healthstone.js
djanse: Oh right, I completely forgot about that, will change it once I've got time for this!

src/parser/shared/modules/items/Healthstone.js
Dambroda: `find` returns the found value, `some` returns a boolean -- your arrow function wouldn't even have to change I think

djanse: That seems to be better. Since it now returns either `undefined` or `Combatant`. Would also be typescript proof.

PR #3611, opened 2020-06-03
[Survival] Update Vipers Venom and Serpent Sting logic

Pewtro:
https://www.warcraftlogs.com/reports/6x8VpvqdJNcf9bQX/#fight=8&source=15
Old: 
image
New: 
No suggestions

Old:
image

New: 
image

Old without errors:
image

New without errors (now with errors):
image

Log: https://www.warcraftlogs.com/reports/6x8VpvqdJNcf9bQX/#fight=4&source=15
Old with errors:
image

New with errors:
image

PR #3612, opened 2020-06-03
[Hunter] Extend DeathTracker to ignore Feign Death casts as deaths

### LINKED ISSUE #2670: [Hunter] Wrong death statistic
Whenever a hunter uses feign death, it will incorrectly register as time spent dead - however there is no cast event for feign death. 

My idea is to lower timespent dead whenever the hunter is dead for less than 1second.

Pewtro:
Blocked by #3611 
Fixes #2670 

Due to combatlog restrictions, Feign Death is not shown as a cast event to other players (to make it more believable, otherwise it is cheesable with WA/AddOns easier than already). 
This change ignores extremely short deaths for hunters. 

Log: 
Before: 
image
After:
No suggestion

Log: 
Before:
image

After:
image

### DIFF
--- a/src/parser/hunter/shared/modules/core/DeathTracker.tsx
+++ b/src/parser/hunter/shared/modules/core/DeathTracker.tsx
@@ -1,0 +1,31 @@
+import CoreDeathTracker from 'parser/shared/modules/DeathTracker';
+
+
+/**
+ * Due to combatlog restrictions it is not possible to know whether the hunter cast Feign Death or the hunter actually died.
+ * For this module we assume any single death lasting less than 0.2% of an encounter (1.2 seconds on a 10 minute fight) was due to Feign Death, and thus we remove it as it is improbable anyone is that fast to ress anyone.
+ */
+const TIME_SPENT_DEAD_THRESHOLD = 0.002; //0.2%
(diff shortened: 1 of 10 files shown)


src/parser/hunter/shared/modules/core/DeathTracker.tsx
```diff
@@ -0,0 +1,32 @@
+import CoreDeathTracker from 'parser/shared/modules/DeathTracker';
+
+
+/**
+ * Due to combatlog restrictions it is not possible to know whether the hunter cast Feign Death or the hunter actually died.
+ * For this module we assume any single death lasting less than 0.2% of an encounter (1.2 seconds on a 10 minute fight) was due to Feign Death, and thus we remove it as it is improbable anyone is that fast to ress anyone.
+ */
+const TIME_SPENT_DEAD_THRESHOLD = 0.002; //0.2%
+// @ts-ignore duplicate indicator, but it's extended just like other modules
+class DeathTracker extends CoreDeathTracker {
```
Pewtro, line 10: I'm really confused about this ts-ignore, and I don't know why it's necessary - but it kept insisting there was a duplicate identifier..
image

PR #3614, opened 2020-06-04
[Hunter] Add a pet basic attack tracking module

Pewtro:
image
image


Pewtro:
Found this in an old branch of mine, brushed it up to match current situation in-game - and it's interesting to see for people who don't use the macros in their offensive abilities what they lost out on :)

PR #3621, opened 2020-06-05
Jetbrains codestyle update

djanse:
Jetbrains IDEs were configured to remove all blank lines between methods. Which is inconsistent with codestyles mentioned in [airbnb/javascript](https://github.com/airbnb/javascript#whitespace--after-blocks). 

Currently it would format the code as following
before

I believe this is what we wish to see
after


Pewtro:
Thanks!

PR #3623, opened 2020-06-06
[Core] Move some helpers from class/spec modules to shared folders

Pewtro:


Pewtro:
I found several more handleStacks() functions around the core and in other specs/classes which I intend to update in a future PR.

PR #3624, opened 2020-06-08
[Core] RFC: Guild search and guild reports page

Dambroda:
### Changes / Additions Overview
Most of this is copied from the character page and converted to TS.

**Current:**
image

**After:** 
image
image

**Optional "minimalist" change, to keep things as simpler (but less obvious to find)**
image

**Guild page (alliance):**
image

**Guild page (horde):**
image

Shows report for last 3 months. Can show 25, 50, or 100 reports. Can be filtered by zone.

### Optional TODOS
 (at this point, _most_ of these would be future additions rather than in this PR)

- Add an article? I can't even see where to view articles from the home page anymore.
- Change to the minimalist view if the tabs are unliked
- Add guild to 'recently viewed' history. This would mean less characters/reports would appear in the list, but if you use the guild page as your main navigational tool it would be helpful
- Add a way to view the fights for the listed guild reports on the guild page 
  - Click on an expand button and see the fights for the report (one on-demand wcl api call) 
  - **OR**
  - The fights are all shown in the reports list (lots more api calls), similar to the WCL "calendar" view but in a list form
- Add a generated guild crest using the crest information from the blizz api, either instead of or in addition to the alliance/horde image.


Pewtro:
```
Routing / sluggin'. Should the URL if the guild contains a space be /club+camel/ ? Same in the document title
```

Yes, I think /club+camel/ is perfectly fine, AND should make it easy for people to navigate to that URL. 

```
Guild page filtering. Filter by date (3 months) is currently behind the scenes. Keep that or remove? Add a visible date filter component? Filter a raw number of reports instead like character parses (100 items)? Default to current "newest" zone, or "all"?
```

How long are we allowed to keep this with the new ToS, is that part of the keep for 30(?) days maximum? I think we should default to all, and X number of reports per page.. No one goes back 50 reports for example.. and if they do it's such an edge case..

```
Alliance / horde images. The images I am using are currently from a steam workshop. I looked into it, and as long as we (aka zero) are "subscribed" to the page I think we have the rights to use it.. but we'd lose them if the workshop item was removed, and steam has all rights as well. I assume we'd rather just use something whose license was clearer, but I don't know how to find them.
```

Oof... Yeah... Could we use anything blizzard owned?

```
I installed a dep for typescript definitions. --frozen-lockfile? Does this mean I commit my lockfile changes orrr
```

Yes - you'd install the dependency and then ``yarn``, not ``yarn upgrade`` like I did and then you'd add the lockfile changes also (which should be fairly minimal with just one dependency added.

Dambroda:
`Yes, I think /club+camel/ is perfectly fine, AND should make it easy for people to navigate to that URL.`

Yeah the routing was just an issue on my end that confused me. But another thing I want to fix is replacing the document title with the real guild name once returned by the API. We don't even do that with characters at the moment, whatever you search for is what you get.

`How long are we allowed to keep this with the new ToS, is that part of the keep for 30(?) days maximum?`
Reports are from WCL, not Blizzard. Storing the blizz api information (which is just guild name, faction, etc.) is all on the server and just used for caching. So we can get whatever WCL gives us (years and years)

`I think we should default to all, and X number of reports per page.. No one goes back 50 reports for example.. and if they do it's such an edge case..`

I agree with default to all zones. On the other point, not currently planning on paging, so in the latest push I added a selector that defaults to only 25 visible reports, with an option to view 50 or 100 for those edge cases.
image


Another question: keep the report date view as-is, or maybe change to `"X days|hours|minutes"` ago?

Pewtro:
> Another question: keep the report date view as-is, or maybe change to "X days|hours|minutes" ago?
 Current is better than X ago imo. It's clearer when something happened that way

Dambroda:
PR updated and ready for review!

src/interface/home/ReportSelectionHeader.js
```diff
@@ -21,25 +27,54 @@ class ReportSelectionHeader extends React.PureComponent {
...
+    switch (this.state.searchType) {
+      case STATE_SEARCH_CHAR:
+        return (<>
+          <NameSearch type={SearchType.CHARACTER} />
+          <br />
+          <Warning>
+            <Trans>
+              The character page will only show fights that have been ranked by Warcraft Logs. Wipes are not included and during busy periods there might be a delay before new reports appear. You can still analyze these fights by manually finding the report on Warcraft Logs and using the report link.
+            </Trans>
+          </Warning>
```
Pewtro, line 68: Do guilds have the same delay limitation? If so, we might need a warning

Pewtro (commented):
Very minor things, but overall LGTM!

src/interface/GuildReports/GuildReports.tsx
```diff
@@ -0,0 +1,532 @@
...
+      );
+    } else if (
+      this.state.error === ERRORS.WCL_API_ERROR ||
+      this.state.error === ERRORS.UNKNOWN_API_ERROR ||
+      this.state.error === ERRORS.UNEXPECTED
+    ) {
+      errorMessage = (
+        <Trans>
+          {this.state.errorMessage} Please message us on{' '}
+          <a
```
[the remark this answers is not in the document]
Dambroda: I cleaned it up a bit by creating reusable elements for the links

src/interface/home/ReportSelectionHeader.js
Dambroda: No idea... but I personally think the warning for characters probably shouldnt be on the front page as well. All of the relevant information is in the guild/character page error messages...

Pewtro (approved):

PR #3629, opened 2020-06-10
[Core] Shared Binomial Calculations and Graphs

Pewtro:
This is a draft for discussion of the move from having binomial calculations spread around different modules (Prot paladin, Brewmaster monks, Warlocks and hunters) and gathering them in one place. 

At the same time, ideally also make it easier to use the function for future modules, and to easily implement them into a specs module.


src/parser/shared/modules/helpers/Probability.tsx
```diff
@@ -1,3 +1,114 @@
+import OneVariableBinomialChart from 'interface/others/charts/OneVariableBinomialChart';
+import React from 'react';
+import { formatNumber, formatPercentage } from 'common/format';
+
+/**
+ * pn is the mean value of procs
+ * @param procChance
+ * @param procAttempts
+ */
+export function probabilityPN(procChance: number, procAttempts: number) {
```
emallson, line 10: `probabilityPn` -> `expectedProcCount`

src/parser/shared/modules/helpers/Probability.tsx
```diff
@@ -1,3 +1,114 @@
...
+  return procChance * procAttempts;
+}
+
+/**
+ * qn is the mean value of non procs
+ * @param procChance
+ * @param procAttempts
+ */
+export function probabilityQN(procChance: number, procAttempts: number) {
```
emallson, line 19: `probabilityQN` -> `expectedNonProcCount`

Pewtro:
Yeah I called them probabilityQN/PN because they were in the probability file but makes sense! I'll think of nicer names for them and rename to that 👍

PR #3638, opened 2020-06-19
[Arcane Mage] Integration Test Update

Sharrq:
Updates the integration tests and example report for Arcane which hasnt been updated since Battle for Dazaralor


Pewtro:
Looks good!

PR #3645, opened 2020-06-28
chore: improve initial page load performance

vpicone:
## Problem
The current heuristic for which changes to show is all of the changes since five articles ago. Considering the last article was written roughly a year ago, this is a substantial amount and the 5th article (the one we print changelogs through) was roughly 2 years ago. This is particularly problematic because the images in article previews are a hit to performance, but they're no where near above the fold.

## Solution
This change combines the articles and changelog entries so that they are paginated together. In this way, the initial page load isn't dependent on the amount of time since the last article.

## Testing
I've hosted some previews for comparison. Compare the initial page load of [current state](https://wow-analyzer-i9w3sth5y.vercel.app/) vs. [the proposed change](https://wow-analyzer-r2mttvovk.vercel.app/).

## Results
Time to interactive 11s → 6s
Max first input delay 4s → 0.3s
First cpu idle 10.9s → 5.7s

The max first input delay is particularly nice, considering most folks are coming to immediately paste a log.


public/index.html
```diff
@@ -1,102 +1,138 @@
 <!DOCTYPE html>
 <html lang="en" prefix="og: http://ogp.me/ns#">
+  <head>
```
vpicone, line 3: Most of these changes are due to prettier formatting. The only update was combining the two google font links and updating them to use the latest google fonts api.

vpicone:
@rp4rk Do we need to get a second review here? Just want to make sure I'm not missing something.

Pewtro:
Thanks!

PR #3650, opened 2020-07-09
[Core] Cooldown Pre Casts

Sharrq:
Adds the ability to be able to show the casts x seconds or x spells before the cooldown on the cooldown page.

### DIFF
--- a/src/parser/shared/modules/CooldownThroughputTracker.js
+++ b/src/parser/shared/modules/CooldownThroughputTracker.js
@@ -5,6 +5,7 @@ import Panel from 'interface/others/Panel';
 import CooldownIcon from 'interface/icons/Cooldown';
 import CooldownOverview from 'interface/others/CooldownOverview';
 import Analyzer from 'parser/core/Analyzer';
+import { EventType } from 'parser/core/Events';
 import CASTS_THAT_ARENT_CASTS from 'parser/core/CASTS_THAT_ARENT_CASTS';
 
 const debug = false;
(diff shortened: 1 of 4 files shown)


src/parser/shared/modules/CooldownThroughputTracker.js
```diff
@@ -52,10 +53,19 @@ class CooldownThroughputTracker extends Analyzer {
...
+      start: timestamp - (cooldownSpell.startBufferMS || 0),
       end: null,
       events: [],
     };
+    if (cooldown.startBufferMS) {
+      const combatantCasts = this.owner.eventHistory.filter(event => event.type === EventType.Cast && event.sourceID === this.selectedCombatant._combatantInfo.sourceID);
+      const preCasts = combatantCasts.filter(event => event.timestamp >= timestamp - cooldown.startBufferMS && event.timestamp <= timestamp);
+      preCasts.reverse().forEach(event => cooldown.events.unshift(event));
+    } else if (cooldown.startBufferSpells) {
+      const combatantCasts = this.owner.eventHistory.filter(event => event.type === EventType.Cast && event.sourceID === this.selectedCombatant._combatantInfo.sourceID);
```
jjs451, line 65: If you wanted to go all out on reducing duplicate code, you could do something like this

    if (cooldown.startBufferMS || cooldown.startBufferSpells) {
        const combatantCasts = this.owner.eventHistory.filter(event => event.type === EventType.Cast && event.sourceID === this.selectedCombatant._combatantInfo.sourceID);

        if (cooldown.startBufferMS) {
            const preCasts...
        } else if (cooldown.startBufferSpells) {
            const preCasts...
        }
    }

Not really a big deal so I'll set this to approved and unmerged and leave it up to you.

Sharrq: Yea i was looking at reducing it into something a bit shorter ... similar to what you did, maybe ill throw that in

Sharrq:
@Dambroda has this change bundled in with his PR, so closing this.

PR #3652, opened 2020-07-11
[Core] Event History utility function and some TS conversions

Dambroda:
What's in here?

**TS Conversion**
- Converted EventEmitter to TS
- Converted MODULE_ERROR to TS enum
- Converted CooldownThroughputTracker to TS
- Stictified some existing TS

**New functionality**
- New module EventHistory that can be imported as a dependency
  - Adds EventHistory.last(count, maxTime, filterDef)
   ```
  /**
   * @param count the maximum number of events to return, or null for no limit
   * @param maxTime the maximum number of milliseconds to look back, or null for no limit
   * @param filterDef an optional EventFilter to apply to all events
   * @returns the last `count` events that match the given filters, with the oldest events first
   */
- CooldownThroughputTracker (and children) cooldownSpells have new options (thanks Sharrq):
  - `startBufferMS`: If provided, the cooldown tracker will add X ms of events from before the cooldown started to the cooldown window
  - `startBufferEvents`: If provided, limits the number of events added to the cooldown window
  - `startBufferFilter`: An `EventFilter` to filter the events for, defaults to `Events.cast.by(SELECTED_PLAYER)`
  - NOTE: `startBufferEvents` and `startBufferMS` are and'ed, not or'ed. Providing both will filter by both.

EXAMPLES:

```javascript
    // Get the last three cast events by the selected player
    this.eventHistory.last(3, null, Events.cast.by(SELECTED_PLAYER));

    // Get the last three flurry casts by the selected player (will potentially go back to fight start!)
    this.eventHistory.last(3, null, Events.cast.by(SELECTED_PLAYER).spell(SPELLS.FLURRY)));

    // Get the last 5 seconds of cast events for the selected player
    this.eventHistory.last(null, 5000, Events.cast.by(SELECTED_PLAYER));

    // Get the last 10 seconds of cast events for the selected player, but give me a maximum of the latest 5
    this.eventHistory.last(5, 10000, Events.cast.by(SELECTED_PLAYER))
```

```javascript
    {
      spell: SPELLS.COMBUSTION,
      startBufferEvents: 10,
      summary: [
        BUILT_IN_SUMMARY_TYPES.DAMAGE,
      ],
    },
```

Note that the cooldown durations are different (because no timestamp filter was added), and 10 events before Combustion are included:

image
(also this is not what it's actually being changed to Sharrq, just an example! keeping your 4 second windows)

### DIFF
--- a/src/parser/core/modules/EventEmitter.ts
+++ b/src/parser/core/modules/EventEmitter.ts
@@ -292,11 +292,11 @@ class EventEmitter extends Module {
     console.error('Disabling', name, 'and child dependencies because an error occured:', err);
     // Disable this module and all active modules that have this as a dependency
     this.owner.deepDisable(module, ModuleError.EVENTS, err);
-    Sentry && Sentry.withScope(scope => {
+    window.Sentry && window.Sentry.withScope(scope => {
       scope.setTag('type', 'module_error');
       scope.setTag('spec', `${this.selectedCombatant.spec.specName} ${this.selectedCombatant.spec.className}`);
       scope.setExtra('module', name);
-      Sentry.captureException(err);
+      window.Sentry && window.Sentry.captureException(err);
     });
   }
 }
--- a/src/parser/core/modules/EventEmitter.ts
+++ b/src/parser/core/modules/EventEmitter.ts
@@ -1,13 +1,19 @@
-import MODULE_ERROR from 'parser/core/MODULE_ERROR';
-import Events from '../Events';
-
+import ModuleError from 'parser/core/ModuleError';
+import Events, { Event } from '../Events';
 import Module from '../Module';
+import { EventListener } from '../EventSubscriber';
 import { SELECTED_PLAYER, SELECTED_PLAYER_PET } from '../Analyzer';
-import EventFilter from '../EventFilter';
+import EventFilter, { SpellFilter } from '../EventFilter';
 
 const CATCH_ALL_EVENT = 'event';
 
 const PROFILE = false;
+
+type BoundListener = {
+  eventFilter: EventFilter<string> | string;
+  module: Module;
+  listener: EventListener<Event>;
+}
 
 /**
  * This (core) module takes care of:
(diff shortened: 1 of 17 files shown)


src/parser/core/modules/EventEmitter.ts
```diff
@@ -82,59 +89,73 @@ class EventEmitter extends Module {
...
-  _prependSpellCheck(listener, spell) {
+
+  createSpellCheck(spell: SpellFilter): ((event: Event) => boolean) {
     if (spell instanceof Array) {
-      return function (event) {
-        if (event.ability && spell.some(s => s.id === event.ability.guid)) {
-          listener(event);
-        }
-      };
+      // @ts-ignore adding typechecking to <generic> event properties will be a lot of work
```
MartijnHols, line 147: Isn’t the Event type sufficient here?

src/parser/core/modules/EventEmitter.ts
```diff
@@ -257,24 +278,26 @@ class EventEmitter extends Module {
...
+  _handleError(err: Error, module: Module) {
     if (process.env.NODE_ENV !== 'production') {
       throw err;
     }
     const name = module.key;
     console.error('Disabling', name, 'and child dependencies because an error occured:', err);
     // Disable this module and all active modules that have this as a dependency
-    this.owner.deepDisable(module, MODULE_ERROR.EVENTS, err);
+    this.owner.deepDisable(module, ModuleError.EVENTS, err);
+    // @ts-ignore Sentry does not exist on window, need Sentry types
```
MartijnHols, line 295: I got you fam

MartijnHols: **sentry.d.ts**
```ts
import * as SentryType from '@sentry/browser'

declare global {
  // May be undefined when adblockers block Sentry (for privacy concerns)
  const Sentry: typeof SentryType | undefined
}
```

src/parser/core/modules/EventEmitter.ts
Dambroda: No, because `Event` does not define` .ability` (or .to, .by, etc); I had started normalizing the Event types by creating things like `AbilityEvent` `TargetedEvent`, etc, but things like event listeners are currently utilizing some of the lazy typing (by checking for undefined) and not being strict about it. So, we can either add things like `.ability .to .by` to Event and be kinda lazy, or we can get very strict all the way up to typechecking the EventListener you provide to `.addEventListener` to make sure it accepts the correct Event type parameter for the EventFilter you've given it.

Or, maybe there's an easier solution I'm missing!

src/parser/core/modules/EventEmitter.ts
Dambroda: this will define Sentry but not window.Sentry, will it work if I change the code to `Sentry && Sentry.withScope` ? Or do I need to define Sentry onto `window`

MartijnHols: it should work without window, otherwise I have other projects I need to fix

Dambroda: Cool beans!

src/parser/core/modules/EventEmitter.ts
Dambroda: This is being resolved in a future PR; converting CLP and a lot of other core modules

src/parser/core/modules/EventEmitter.ts
MartijnHols: image

bigoof

MartijnHols: you are right
I had it wrong
do yours

PR #3661, opened 2020-07-24
Fix crash when character profile 404s

Dambroda:
I misinterpreted the logic of how this worked before converting to TS -- fixed a crash caused by throwing an error when race was not found even if the characterprofile object was sent (as an error)


Dambroda:
Instead of this bandaid fix I just noticed zero stopped bad characterProfiles from being sent to the parser

PR #3685, opened 2020-08-08
[Shadowlands] Change remaining handleStacks usage to use currentStacks helper

Pewtro:
Should have no influence over any results - simply using a cleaner and more standardized approach to handling buffs. 

This PR does not adjust traits and essences using handleStacks as we'll be removing those come SL.


Pewtro:
@MartijnHols Could you check the change to the github workflow if it's fine? It seems to fit what the purpose of the action is, and allows for multi-branching a bit better.

PR #3703, opened 2020-09-03
[Hunter] First two legendaries and initial conduit implementation

Pewtro:


emallson (approved):
Imagine forgetting a changelog entry KEKW

PR #3706, opened 2020-09-08
[Core] Warning Message for Shadowlands Prepatch

Sharrq:
Added a warning message to notify users that we will be removing support for Azerite, Essences, Corruption, and all BFA items with the launch of Prepatch.


abelito75 (approved):
Looks good. All hail shadowlands, the remover of corruption, essences, and azerite.

PR #3729, opened 2020-09-21
Shadowlands blood

yajinni:
-Moved Blood Tap from spells to talents
-Removed Ossuary from talents
-Since Ossuary is baseline modified some of other modules affected by its calculations


src/parser/deathknight/blood/modules/talents/Heartbreaker.js
```diff
@@ -8,13 +8,10 @@ import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 class Heartbreaker extends Analyzer {
   rpGains = [];
   hsCasts = 0;
-  deathStrikeCost = 45;
+  deathStrikeCost = 40;
```
joshinat0r, line 11: Might be a good idea to make this a const for this module, with deathStrikeCost always being 40RP

joshinat0r (changes requested):
Just this one tiny `const` change, everything else looks good

PR #3737, opened 2020-09-22
[Shadowlands][Core] Castle Nathria Boss configs & default header

Sharrq:
- Added the Castle Nathria boss configs
- Added a placeholder background/headshot for each boss (will update once blizz posts their images)
- Replaced the default header background for Shadowlands (Instead of the anduin vs sylvanas bfa one)

Default header when boss isnt found
image

Placeholder Background for Castle Nathria bosses
image


MartijnHols (approved):
### Thanks a lot!

PR #3745, opened 2020-09-23
[Shadowlands][Enhancement] Removed Earth Elemental suggestion

TheMusicMeister:
Based on comments from Enhancement Shaman discord, Earth Elemental is no longer a valid offensive GCD.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,4 +39,3 @@ temp
 .vs/slnx.sqlite-journal
 .vs/VSWorkspaceState.json
 .vs/WoWAnalyzer/v15/.suo
-/.vs/WoWAnalyzer Repo/v16/.suo
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ temp
 .vs/slnx.sqlite-journal
 .vs/VSWorkspaceState.json
 .vs/WoWAnalyzer/v15/.suo
+/.vs/WoWAnalyzer Repo/v16/.suo
--- a/src/parser/shaman/enhancement/modules/Abilities.tsx
+++ b/src/parser/shaman/enhancement/modules/Abilities.tsx
@@ -1,9 +1,7 @@
-import React from 'react';
 
 import SPELLS from 'common/SPELLS';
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -39,3 +39,4 @@ temp
 .vs/slnx.sqlite-journal
 .vs/VSWorkspaceState.json
 .vs/WoWAnalyzer/v15/.suo
+/.vs/WoWAnalyzer Repo/v16/.suo
```
Sharrq, line 42: Dont edit .gitignore

src/parser/shaman/enhancement/modules/Abilities.tsx
```diff
@@ -47,15 +47,6 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
-        castEfficiency: {
```
Sharrq, line 50: What are you trying to accomplish?

Do you want to remove it from the list of spells at the bottom of the Statistics tab? Or are you just trying to remove the Suggestion that says you arent casting it enough?

Sharrq (changes requested):

src/parser/shaman/enhancement/modules/Abilities.tsx
TheMusicMeister: This particular change is to remove it from the spell suggestions, as it should no longer be considered as an option for an offensive GCD.

Sharrq: then just change suggestion: true to suggestion: false and get rid of recommendedEfficiency and extraSuggestion

TheMusicMeister: Got it, will do.

Sharrq:
You also need to resolve the ESLint errors (React and SpellLink are imported but not used in Abilities), and add a changelog entry.

TheMusicMeister:
> 
> 
> You also need to resolve the ESLint errors (React and SpellLink are imported but not used in Abilities), and add a changelog entry.

Done.

Sharrq (approved):

PR #3746, opened 2020-09-23
[Shadowlands][Enhance] Removed broken line to correct page breaking

TheMusicMeister:
This line was causing a page-breaking error as the requisite talent module was removed, but the ability was not added back in as a baseline ability as it will be in Shadowlands.  I removed the Threshold line to allow the page to load properly.

I will work on adding this ability back in as a baseline spell and will re-add the Threshold at that point.


TJMoats:
I actually noticed this bug when checking on your other PR. Thanks for patching it 👍

Pewtro:
This doesn't remove anything?

TheMusicMeister:
Wait...  Did someone already fix it??  The line is gone...

Pewtro:
No it's still there, I was just blind - since this is on our development branch it's safe for you to add it back in this PR though :)

TheMusicMeister:
Later on (probably once SL actually launches) I plan to come up with a way to track how many Maelstrom Weapon stacks were gained from Lightning Shield, but for now this will simply track uptime.

Pewtro:
Should add average and minor - surely there are levels of severity for uptime?

HawkCorrigan:
> 
> 
> Should add average and minor - surely there are levels of severity for uptime?

Its a 1 hour buff. You should have it perma up.

TheMusicMeister:
> 
> 
> > Should add average and minor - surely there are levels of severity for uptime?
> 
> Its a 1 hour buff. You should have it perma up.

Half hour in SL, but yeah, similar concept.  If you don't have it up, you're uh... you're doing it wrong.

Pewtro:
Should it be considered major then? Like how impactful is it.

HawkCorrigan:
There is only one problem tbh.
Its is exclusive with Earth Shield iirc. So there is some merit to not having it up in those cases,

I would be fine with either igoring the requirement if you had ES up or just keeping the module as you made it now.

TheMusicMeister:
I can add a check to see if the player is actively using ES, and can have this ignored in that case.  As for the major level, it's going to be vital for optimal DPS, so I as an Enhancement player would consider it a major loss if it's not constantly up.

HawkCorrigan:
You will only gain anything from it if you get meleed tho right?

TJMoats:
It's still a free buff. There should never be a case in which a raider would not be better off without one of the shields up.

HawkCorrigan:
Oh ye sure I just mean in regards to it being "impactful".
There is not much reason to ever not have it up.

Pewtro:
Everything is vital for optimal DPS, doesn't mean all our checks are only major thresholds - which is why I'm asking into the levels of impactfulness of it.

TheMusicMeister:
It is true that it would only have an impact if the player were to be hit by a melee attack, which would (hopefully) never happen in most raid cases...  I'll adjust the suggestion to be minor, since the spell doesn't **really** add DPS in and of itself except in extreme edge cases.

HawkCorrigan:
I'd personally consider it minor for enhancement.
@TheMusicMeister Do you want to add some check for Eaerth Shield or not so much?
I dont think it should be useful to intentionally not apply Lightning so you can get ES, but that is just bad design on Blizzards part xD.
Your call imo

TheMusicMeister:
Yeah, I can add a check for Earth Shield.  It's a good point to point out that they're mutually exclusive, so it would be counterintuitive to scold the player for not using LS when they're actually correctly using something more useful for a situational encounter like ES.

HawkCorrigan:
Aight, fair enough.

TheMusicMeister:
Changed from major to minor suggestion.  I also removed the Lightning Shield uptime from the main checklist, as it's really not something vitally important to Enhance optimization.

TheMusicMeister:
Added simple check to add Earth Shield uptime to the overall threshold check.  To match that it now checks both shields, I renamed the main threshold function to elementalShield- instead of lightningShield-.

src/parser/shaman/shared/spells/LightningShield.tsx
```diff
@@ -0,0 +1,42 @@
...
+        minor: 0.95,
+      },
+      style: 'percentage',
+    };
+  }
+
+  suggestions(when: any) {
+    when(this.elementalShieldUptimeThreshold)
+      .addSuggestion((suggest: any, actual: any, recommended: any) => {
+        return suggest(<span>Remember to have <SpellLink id={SPELLS.LIGHTNING_SHIELD.id} /> (or <SpellLink id={SPELLS.EARTH_SHIELD_TALENT.id} /> up as constantly as possible. As a 30 minute buff, one should always be cast before combat as well as just after res, if possible.</span>)
```
Sharrq, line 34: Use fragments <> </> instead of span

src/parser/shaman/enhancement/modules/checklist/Module.tsx
```diff
@@ -29,6 +31,7 @@ class Checklist extends BaseChecklist {
   protected flametongue!: Flametongue;
   protected flametongueRefresh!: FlametongueRefresh;
   protected hailstorm!: Hailstorm;
+  protected lightningShield!: LightningShield;
```
Sharrq, line 34: Looks like you added this to Module, but never added it to the Checklist? So this doesnt actually show in the checklist.

Sharrq (changes requested):

Pewtro:
This seems to be fixed (the page breaking) - if there's a need for a suggestion or checklist item please open it in a new PR.

PR #3749, opened 2020-09-24
[Core] Covenant systems & Combatant Faker

Pewtro:
Also a small change to how I handled the yarn updating in package.json/yarn.lock -- rerun yarn if you want after this merge but not needed.


src/parser/hunter/survival/modules/spells/conduits/StrengthOfThePack.tsx
```diff
@@ -23,12 +23,12 @@ class StrengthOfThePack extends Analyzer {
...
   constructor(options: any) {
     super(options);
-    this.active = false;
+    this.active = this.selectedCombatant.hasConduitBySpellID(SPELLS.STRENGTH_OF_THE_PACK_CONDUIT.id);
     if (!this.active) {
       return;
     }
 
-    this.conduitRank = 1; //TODO: Find out the proper way of parsing conduit ranks
+    this.conduitRank = this.selectedCombatant.conduitRankBySpellID(SPELLS.STRENGTH_OF_THE_PACK_CONDUIT.id);
```
Dambroda, line 31: This feels a little bit repetitive because conduitRankBySpellID also checks if the conduit is active

```typescript
super(options);
let rank = this.selectedCombatant.conduitRankBySpellID(SPELLS.STRENGTH_OF_THE_PACK_CONDUIT.id);
if(!rank) {
  this.active = false;
  return;
}
this.conduitRank = rank;
```

Purely stylistic though, not something id request changes for if you disagree

src/parser/core/Combatant.ts
```diff
@@ -233,13 +236,11 @@ class Combatant extends Entity {
   //region Covenants
   covenantsByCovenantID: { [key: number]: Covenant } = {};
 
-  _parseCovenants(covenants: Array<Covenant>) {
-    if (!covenants) {
+  _parseCovenants(covenant: Covenant) {
```
Dambroda, line 239: _parseCovenant

src/parser/core/Combatant.ts
```diff
@@ -251,13 +252,11 @@ class Combatant extends Entity {
   //region Soulbinds
   soulbindsBySoulbindID: { [key: number]: Soulbind } = {};
 
-  _parseSoulbinds(soulbinds: Array<Soulbind>) {
-    if (!soulbinds) {
+  _parseSoulbinds(soulbind: Soulbind) {
```
Dambroda, line 255: _parseSoulbind

src/interface/report/CombatantInfoFaker.ts
```diff
@@ -1,14 +1,162 @@
...
+  fakedPlayer.conduits = fakeConduitGenerator();
   fakedPlayer.error = null;
   return fakedPlayer;
 }
 
+//region Shadowlands Systems
+
+//region Covenant Generation
+
+function fakeCovenantGenerator(specID: number) {
```
Dambroda, line 27: The design of this is a little weird, I would have thought doing something like this would be easier for people to use:

```typescript

type FakeInfo = 
  {
    covenant: Covenant, // or id?
    soulbind: Soulbind // or id?
    ... conduits, etc
  }

const SPEC_CONFIGS: { [specId: number]: FakeInfo } = 
{
  [SPECS.MARKSMANSHIP_HUNTER.id]: { covenant: COVENANTS.VENTHYR, soulbind: SOULBINDS.GENERAL_DRAVEN, [...] } 
}
```

So that people could change their specs all in one place instead of having to go through each of the arrays. Like every other comment this one is pretty stylistic and not blocking

Dambroda (commented):
Everything looks fine, all comments here are either nitpicky or about general design which is subjective anyway

src/interface/report/CombatantInfoFaker.ts
Pewtro: That is way easier 😊

Dambroda (approved):

src/parser/hunter/survival/modules/spells/conduits/StrengthOfThePack.tsx
Pewtro: I agree on the spec specific ones, but on the covenant ones I can't do it without also allowing conduitRank to be a boolean so those stayed hasCovenant && hasConduit

Issue #3570, opened 2020-05-04
Add support for Anonymous logs


niseko:
Anonymous logs are a new feature of warcraftlogs, where all playerinfo is hidden.
Example: https://www.warcraftlogs.com/reports/a:nGVCbkm7jtLNDwhg

exportedCharacters and server fields are not in the API request for those which likely will crash the page currently, and the front page link parsing will need to be adjusted

PR #3758, opened 2020-09-27
Ability to lookup Anonymous Logs - Issue 3570

adammkelly:
New to committing to this repo, so if I have missed any steps, please let me know.

Fixes: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3570
Example URL:
image

image
image
image


Sharrq:
Nice! Ill let someone else review it, but i did notice that the character tab is freaking out .... presumably because its trying to lookup info from the armory without a player name. Might need to stop it from doing whatever its doing on anonymous logs

image

adammkelly:
Thanks for taking a look @Sharrq , I will check that out.

adammkelly:
@Sharrq I am seeing this on none anonymous report on master also, do you see the same?

Sharrq:
i pulled down your pr and saw it towards the bottom of the page.  Might have just been on my end

src/parser/core/Combatant.ts
```diff
@@ -74,7 +74,12 @@ class Combatant extends Entity {
     if (!this.owner.characterProfile) {
       return null;
     }
-    const raceId = this.owner.characterProfile.race;
+    let raceId = this.owner.characterProfile.race;
+    if (raceId === null) {
+      // When it is an anonymous report we won't have any race.
```
MartijnHols, line 79: Isn't `characterProfile` always falsey already for anonymous reports?

adammkelly: The object exists but many of the fields are null or empty.

src/parser/core/Combatant.ts
```diff
@@ -74,7 +74,12 @@ class Combatant extends Entity {
     if (!this.owner.characterProfile) {
```
MartijnHols, line 74:
```suggestion
    if (!this.owner.characterProfile || ! this.owner.characterProfile.race) {
```

In that case I would recommend doing this instead. We already have to consider the race being unknown as we can't always query the character from the Armory. Things would be simpler if there was only 1 such situation for this in our code.

src/parser/core/Combatant.ts
```diff
@@ -71,10 +71,15 @@ class Combatant extends Entity {
     return SPECS[this.specId];
   }
   get race(): Race | null {
-    if (!this.owner.characterProfile) {
+    if (!this.owner.characterProfile || !this.owner.characterProfile.race) {
       return null;
     }
-    const raceId = this.owner.characterProfile.race;
+    let raceId = this.owner.characterProfile.race;
+    if (raceId === null) {
```
MartijnHols, line 78: Thanks for implementing the previous suggestion. I reckon that should make this code (and the "anonymous" race) obsolete. Can it be removed without affecting the behavior?

adammkelly: This causes issues later on within the code where race ID is expected to be set to something other than null/undefined. That is why I opted for an Anonymous race.

MartijnHols: That is strange. `null` should be a common value for race as we often can't fetch the character profile, so it should work just fine. Do you have an example of something that would break?

adammkelly: Look at the lines just below this, it breaks and from what I experienced it was just a ripple affect.

MartijnHols (approved):
Thanks!

PR #3760, opened 2020-09-29
[Core] ParseResults TS conversion

Dambroda:
This is to actually type `suggestions(when: any) ...` etc

Its really ugly, mainly due to the fact that we allow both boolean and number thresholds with the same function


Dambroda:
Going to merge this into Shadowlands branch instead.

Issue #3712, opened 2020-09-10
Migrate Charts to Vega-Lite


emallson:
It is *so* much nicer to work with than ReactVis.

- [x] Distance Moved
- [x] DPS/HPS/DTPS
- [x] Stagger
- [x] Mana
- [x] Self-Heal Timing
- [x] Raid Health
- [x] SPriest: Void Form
- [x] Hunter: Focus Cap
- [x] VDH: Pain
- [x] Ele: Maelstrom
- [x] Generic Donut Chart
- [x] DK: Rune chart
- [x] Generic Binomial Chart

Current work is on [this branch](https://github.com/WoWAnalyzer/WoWAnalyzer/compare/shadowlands...emallson:vega)

PR #3765, opened 2020-09-30
Migrate charts to Vega (from ReactVis)

emallson:
Tracking issue: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3712

## Charts Removed

- Shadow Priest Void Form (relevant mechanic removed)
- Vengeance DH Pain (VDH doesn't work on shadowlands branch, so I couldn't test a conversion)

## Picture Time!

Screenshot from 2020-09-30 18-36-41

Screenshot from 2020-09-30 18-36-51

Screenshot from 2020-09-30 18-37-04
Screenshot from 2020-09-30 18-37-19
Screenshot from 2020-09-30 18-57-35
Screenshot from 2020-09-30 19-00-17


Sharrq:
Looks like this breaks the version of the Mana Graph that Arcane Mage is using (It is close to the Mana Level Graph but without HPS.
https://media.discordapp.net/attachments/316864385052049420/761005348802658314/unknown.png

src/parser/hunter/beastmastery/modules/core/SpellUsable.tsx
```diff
@@ -10,9 +10,7 @@ class SpellUsable extends CoreSpellUsable {
   lastPotentialTriggerForBarbedShotReset: CastEvent | null = null;
 
   on_byPlayer_cast(event: CastEvent) {
-    if (super.on_byPlayer_cast) {
-      super.on_byPlayer_cast(event);
-    }
+    super.on_byPlayer_cast(event);
     const spellId = event.ability.guid;
```
Pewtro, line 14: I've had issues with doing this before where TS screams at me (and it giving errors in things shown) but if bumping it to 3.8 fixes that then yay. 
I even tried with optional chaining to no avail.

src/parser/hunter/marksmanship/modules/core/SpellUsable.tsx
```diff
@@ -19,9 +19,7 @@ class SpellUsable extends CoreSpellUsable {
         this.lastPotentialTriggerForRapidFireReset = null;
       }
     }
-    if (super.on_byPlayer_cast) {
-      super.on_byPlayer_cast(event);
-    }
+    super.on_byPlayer_cast(event);
   }
```
Pewtro, line 23: same as BM

src/parser/priest/holy/modules/features/HolyPriestHealingEfficiencyBreakdown.tsx
```diff
@@ -83,8 +83,7 @@ class HolyPriestHealingEfficiencyBreakdown extends HealingEfficiencyBreakdown {
                 </tr>
               </thead>
               <tbody>
-                // @ts-ignore
-                <this.HealingEfficiencyTable tracker={tracker} showHealing={this.state.showHealing} />
+                {this.HealingEfficiencyTable({ tracker, showHealing: this.state.showHealing })}
               </tbody>
```
Pewtro, line 87: There are some more ts-ignore in this file, are these removable with this change as well?

src/parser/shaman/shared/maelstromchart/Maelstrom.js
```diff
@@ -32,83 +22,91 @@ const Maelstrom = props => {
...
-    resource.push({x: secIntoFight, y:item.current});
-    waste.push({x: secIntoFight, y:item.waste});
+    rawData.push({kind: 'Maelstrom', x: secIntoFight, y:item.current});
+    rawData.push({kind: 'Wasted', x: secIntoFight, y:item.waste});
   });
 
+  const data = {
+    data: rawData,
+  };
+
```
Pewtro, line 38: Something akin to this is seemingly everywhere - can it be made more generic so that only select changes need to be passed to a chart module?

Pewtro (commented):
Overall looks pretty good! Just the minor comments here and there, nothing blocking (if the MM/BM things work fine).

Remember to delete react-vis from our dependencies if we're done with it.

src/parser/priest/holy/modules/features/HolyPriestHealingEfficiencyBreakdown.tsx
emallson: I have no idea. I did the minimal change to make it compile and no more.

src/parser/hunter/beastmastery/modules/core/SpellUsable.tsx
emallson: with TS 3.8, it actually gave me an error because the condition was always true :eyes:

src/parser/shaman/shared/maelstromchart/Maelstrom.js
emallson: I would say it is probably not a great idea. I did that for the footerchart specifically but we will likely need to undo it in the future. Vega supports multiple datasets as input, a feature that I used repeatedly in converting charts. For example: the Stagger chart has 4 datasets as input: hp, stagger, purifies, and deaths.

While the `BaseChart` *could* infer that array input = single dataset, I think it is more trouble than its worth.

src/parser/priest/holy/modules/features/HolyPriestHealingEfficiencyBreakdown.tsx
Pewtro: All good!

src/parser/shaman/shared/maelstromchart/Maelstrom.js
Pewtro: Fair!

PR #3768, opened 2020-10-01
[Core] Parseresults TS conversion

Dambroda:
If you are considering doing a review, the only files that need review are ParseResults.TS and (see comment)


src/parser/hunter/survival/modules/spells/ButcheryCarve.tsx
```diff
@@ -46,13 +47,13 @@ class ButcheryCarve extends Analyzer {
 
   get avgTargetsHitThreshold() {
     return {
-      actual: (this.targetsHit / this.casts).toFixed(1),
+      actual: Number((this.targetsHit / this.casts).toFixed(1)),
```
Dambroda, line 50: Potentially `+((this.targetsHit / this.casts).toFixed(1))` instead; do we have a preference?

Pewtro: I think Number() is more telling than +() is

Dambroda (commented):
95% of this PR is simply updating some types (`suggestions(when: When)`) with their imports, etc. I've double checked everything in this PR, but the following still could use review:

shared/modules/AlwaysBeCasting - confirm processCast is not used anywhere

ParseResults.tsx - pretty much a complete rewrite that could use review

Pewtro:
> shared/modules/AlwaysBeCasting - confirm processCast is not used anywhere

RollTheBonesTracker has its own variation, I see it used nowhere else.

Pewtro:
I read through the ParseResults change and compared to what it was before, and it seems like a solid 1:1 rewrite. 👍🏼

PR #3773, opened 2020-10-01
[MW] Spell scaling updated

abelito75:
Update to allow stat values to be accurate


src/parser/monk/mistweaver/modules/features/StatValuesSpellInfo.js
```diff
@@ -134,4 +134,108 @@ export default {
...
+  },
+  [SPELLS.FAELINE_STOMP_HEAL.id]: { 
+    int: true,
+    crit: true,
+    hasteHpm: true,
+    hasteHpct: false,
+    mastery: false,
+    vers: true,
+  },
+  [SPELLS.FAELINE_STOMP_ESSENCE_FONT]: {
```
anom0ly, line 209: Missing a `.id` at the end of this.

src/parser/monk/mistweaver/modules/features/StatValuesSpellInfo.js
```diff
@@ -134,4 +134,108 @@ export default {
...
+    hasteHpm: false,
+    hasteHpct: false,
+    mastery: false,
+    vers: true,
+  },
+  [SPELLS.TEAR_OF_MORNING_HEAL.id]: { 
+    int: true,
+    crit: true,
+    hasteHpm: true,
+    hasteHpct: true,
```
anom0ly, line 181: hasteHpct and hasteHpm should be set to false here, both values are already being calculated by the proccing spells. No need to double dip.

src/parser/monk/mistweaver/modules/features/StatValuesSpellInfo.js
```diff
@@ -134,4 +134,108 @@ export default {
...
+    crit: true,
+    hasteHpm: false,
+    hasteHpct: true,
+    mastery: false,
+    vers: true,
+  },
+  [SPELLS.FAELINE_STOMP_HEAL.id]: { 
+    int: true,
+    crit: true,
+    hasteHpm: true,
```
anom0ly, line 204: hasteHpm should be false here, as the initial heal is a direct heal on a set of players.

src/parser/monk/mistweaver/modules/features/StatValuesSpellInfo.js
```diff
@@ -134,4 +134,108 @@ export default {
...
+    hasteHpm: false,
+    hasteHpct: true,
+    mastery: false,
+    vers: true,
+  },
+  [SPELLS.ENVELOPING_BREATH.id]: { 
+    int: true,
+    crit: true,
+    hasteHpm: true,
+    hasteHpct: true,
```
anom0ly, line 165: hasteHpct should be false here based on the fact that EvnB is not cast, but procc'ed from another cast, EvM

anom0ly (commented):
Looks good. A few spell scaling updates needed.

PR #3787, opened 2020-10-02
[DPriest] Typescript Conversion

TJMoats:
This one was a bit of a pain to convert. I fully expect there to be a ton of feedback on some of the typing XD


TJMoats:
@karlpralow I've converted the module that you just made over to tsx. You might want to verify that I didn't bug anything up.

src/parser/priest/discipline/modules/core/Channeling.ts
```diff
@@ -8,21 +9,21 @@ const debug = false;
 
 class Channeling extends CoreChanneling {
   _previousPenanceTimestamp = this.owner.fight.start_time - PENANCE_MINIMUM_RECAST_TIME;
-  _hasCastigation = null;
-  _bolt = 0;
+  _hasCastigation: boolean = false;
```
rp4rk, line 14: Can these (and similar occurrences) not be handled by type inference?

src/parser/priest/discipline/modules/features/PurgeTheWicked.tsx
```diff
@@ -46,47 +50,40 @@ class PurgeTheWicked extends Analyzer {
...
     const uptime = this.uptime || 0;
 
     when(uptime).isLessThan(SuggestionThresholds.PURGE_THE_WICKED_UPTIME.minor)
-      .addSuggestion((suggest, actual, recommended) => {
+      .addSuggestion((suggest: SuggestionFactory, actual: number, recommended: number) => {
         return suggest(<span>Your <SpellLink id={this.dotSpell.id} /> uptime can be improved.</span>)
           .icon(this.dotSpell.icon)
           .actual(`${formatPercentage(uptime)}% uptime`)
-          .recommended(`>${Math.round(formatPercentage(recommended))}% is recommended`)
+          .recommended(`>${formatPercentage(recommended)}% is recommended`)
```
rp4rk: Was this `Math.round` causing problems?

src/parser/priest/discipline/modules/spells/SinsOfTheMany.tsx
```diff
@@ -21,38 +21,42 @@ const SINS_OF_THE_MANY_FLOOR_BONUS = 0.03;
  * from the passive and from 6 onwards it only decreases 0.005.
  * Hence this map with the values for each Atonement count.
  */
-const BONUS_DAMAGE_MAP = {
```
rp4rk, line 25: The object here is more legible for looking at target count to bonus damage mapping

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "noEmit": true,
     "jsx": "preserve",
     "typeRoots": ["./node_modules/@types"],
-    "baseUrl": "src"
+    "baseUrl": "src",
+    "downlevelIteration": true
```
rp4rk: What are the implications of this configuration change?

rp4rk (approved):
Looks good overall, I have a few questions but I don't think they're blockers

src/parser/priest/discipline/modules/features/PurgeTheWicked.tsx
Dambroda: formatPercentage would return a string, so Math.round which takes a number would have been angry. The way to fix it depends on what the end goal is.. but I'm pretty sure in this case the correct answer is to provide the second argument to the format function to trim off the decimal places

src/parser/core/EventGrouper.ts
```diff
@@ -8,7 +10,10 @@
  */
 
 export default class EventGrouper {
-  constructor(threshold) {
+  threshold: number;
+  cache: any;
```
Dambroda: `cache: { [stem: number]: Event<any>[] }`

src/parser/core/Module.ts
```diff
@@ -12,7 +12,7 @@ export interface Options {
 class Module {
   static dependencies: { [desiredPropName: string]: typeof Module } = {};
 
-  protected readonly owner!: CombatLogParser;
+  public readonly owner!: CombatLogParser;
```
Dambroda: What required this change?

src/parser/priest/discipline/modules/Abilities.ts
```diff
@@ -60,7 +60,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.POWER_WORD_SOLACE_TALENT,
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: haste => 12 / (1 + haste),
+        cooldown: (haste:number) => 12 / (1 + haste),
```
Dambroda: At the top of the file,
`spellbook(): SpellbookAbility[] {` 

Will both typecheck the abilities in the file and also automatically infer the `haste` type

src/parser/shared/modules/CooldownThroughputTracker.tsx
```diff
@@ -25,14 +25,14 @@ type TrackedEvent = CastEvent | HealEvent | AbsorbedEvent | DamageEvent | ApplyB
 
 export type CooldownSpell = {
   spell: any,
-  summary: Array<BUILT_IN_SUMMARY_TYPES>,
+  summary: Array<BUILT_IN_SUMMARY_TYPES | any>,
```
Dambroda: This is blocking, please remove this and we can work to fix whatever issue it was added to resolve

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
```diff
@@ -16,8 +16,8 @@ const MAX_TIME_SINCE_CAST = 250; // ms
  events so that the applications are always right after the cast.
  */
 class PowerWordRadianceNormalizer extends EventsNormalizer {
-  normalize(events) {
-    const fixedEvents = [];
+  normalize(events: Array<ApplyBuffEvent | RefreshBuffEvent | ApplyBuffStackEvent | CastEvent>) {
```
Dambroda: This looks at all events, its `Event<any>` (same with fixedEvents declaration)

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
```diff
@@ -37,6 +37,7 @@ class PowerWordRadianceNormalizer extends EventsNormalizer {
         const spellId = event.ability.guid;
         if ((event.timestamp - lastRadianceTimestamp) < MAX_TIME_SINCE_CAST && BUFFS_TO_MOVE.includes(spellId)) {
           event.timestamp = lastRadianceTimestamp;
+          // @ts-ignore
```
Dambroda: Instead, because it's actually added in a lot of normalizers, you should add `__modified?: boolean` to the Event interface right below `__fabricated?: boolean`

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
```diff
@@ -51,4 +51,13 @@ class AtonementSuccessiveDamage extends EventsNormalizer {
     return fixedEvents;
   }
 }
+
+interface ModifiedDamageEvent extends DamageEvent {
+  __modified: boolean
```
Dambroda: __modified should be added to the base Event interface (see previous comment)

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
```diff
@@ -1,17 +1,17 @@
...
-import isAtonement from "../modules/core/isAtonement";
-import { ATONEMENT_DAMAGE_SOURCES } from "../constants";
+import isAtonement from '../modules/core/isAtonement';
+import { ATONEMENT_DAMAGE_SOURCES } from '../constants';
 
 class AtonementSuccessiveDamage extends EventsNormalizer {
-  normalize(events) {
-    const fixedEvents = [];
-    const _damageEventIndexes = [];
+  normalize(events: Array<ModifiedDamageEvent | ModifiedHealEvent>) {
```
Dambroda: `Event<any>`

Dambroda (changes requested):
I got through about half of the files so far, but I figured I'd submit the first part so you can get a head start. Also, I can't leave a comment on it because it there were no line changes, but also please add the typing to CONFIG like Holy  :)

src/parser/priest/discipline/modules/spells/Penance.tsx
```diff
@@ -89,4 +88,12 @@ class Penance extends Analyzer {
   }
 }
 
+export interface PenanceDamageEvent extends DamageEvent {
+  penanceBoltNumber: number;
+}
+
```
Dambroda: (referring to this comment in other comments)

```typescript
export function IsPenanceDamageEvent(event: DamageEvent): event is PenanceDamageEvent {
  return (event as PenanceDamageEvent).penanceBoltNumber !== undefined;
}
```
 and the same for the heal event

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/parser/priest/discipline/modules/spells/Penance.tsx
```diff
@@ -41,10 +39,11 @@ class Penance extends Analyzer {
   }
 
   get currentBoltNumber() {
+    // @ts-ignore
     return [...this.eventGrouper].slice(-1)[0].length - 1; // -1 here for legacy code
   }
 
-  on_byPlayer_damage(event) {
+  on_byPlayer_damage(event: PenanceDamageEvent) {
```
Dambroda: `DamageEvent`

src/parser/priest/discipline/modules/spells/Penance.tsx
```diff
@@ -54,7 +53,7 @@ class Penance extends Analyzer {
     event.penanceBoltNumber = this.currentBoltNumber;
```
Dambroda: `(event as PenanceDamageEvent).penanceBoltNumber`

(same last two comments for the heal listener)

src/parser/priest/discipline/modules/spells/Castigation.tsx
```diff
@@ -20,12 +19,12 @@ class Castigation extends Analyzer {
...
 
-  constructor(...args) {
-    super(...args);
+  constructor(options: any) {
+    super(options);
     this.active = this.selectedCombatant.hasTalent(SPELLS.CASTIGATION_TALENT.id);
   }
 
-  on_byPlayer_damage(event) {
+  on_byPlayer_damage(event: PenanceDamageEvent) {
```
Dambroda: Just a `DamageEvent` (cant assume all events grabbed by this listener are penance events)

The way you can typecheck things like this is with typeguards (refer to previous comment)
```typescript
on_byPlayer_damage(event: DamageEvent) {
  if(!IsPenanceDamageEvent(event)) {
    return;
  }
  // Everything after that check will automatically cast 'event' to PenanceDamageEvent

  // If you trust your other modules to only add the penanceBoltNumber property to penance events you could technically even remove the event.ability.guid check here now
  if (event.ability.guid !== SPELLS.PENANCE.id || event.penanceBoltNumber !== 3) {
      this._isCastigationBolt = false;
      return;
    }
```
Same with the Heal event listener

src/parser/priest/discipline/modules/spells/Evangelism.tsx
```diff
@@ -19,19 +18,28 @@ class Evangelism extends Analyzer {
...
+
+  constructor(options: any) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.EVANGELISM_TALENT.id);
   }
 
   get evangelismStatistics() {
-    return Object.keys(this._evangelismStatistics).map(key => this._evangelismStatistics[key]);
+    // @ts-ignore
+    return Object.keys(this._evangelismStatistics).map((key: number) => this._evangelismStatistics[key]);
```
Dambroda: `return Object.keys(this._evangelismStatistics).map(Number).map((key: number) => this._evangelismStatistics[key]);`

It's annoying, but Object.keys always returns a string array regardless of the TypeScript index type. You can map the keys to number pretty safetly by just adding a `.map(Number)`

src/parser/core/Module.ts
TJMoats: There are non-analyzer modules that have the analyzer passed down to them. The Atonement Healing Breakdown uses this variable, but it pulls it from the parent that creates the module. There are a number of modules that do something similar, and I didn't see too much of a stink from making it public instead of protected. I can modify that module to make it fetch the owner some other way if you think I should.

src/parser/priest/discipline/modules/core/Channeling.ts
TJMoats: I just find that it's cleaner to add the types. They can be inferred if you think I should remove the types. Mind you the disc modules are a bit of a mixed bag for typing modules. There are a ton of nulls that end up being booleans, so I added these in a lot of places to maintain my sanity :D

src/parser/priest/discipline/modules/features/PurgeTheWicked.tsx
TJMoats: What he said ^^ formatPercentage was returning a string.

src/parser/priest/discipline/modules/spells/SinsOfTheMany.tsx
TJMoats: objects being indexed with ints is usually a pretty gangly practice imo. I'd almost rather make it into a function instead of a purely int indexed object. Mind you there are objects in these modules that are indexed by objects so /shrug

I can change it if you really think it should remain an object.

tsconfig.json
TJMoats: One of the modules uses a custom class that someone wrote that allows you to call it's members like an array. It's the EventGrouper class. I didn't 100% get why it needed to be like that, but this change is necessary for typescript to allow iteration in that class.

TJMoats: I'd also 100% be down to axe that class, disc is the only spec that uses it as far as I can tell.

src/parser/shared/modules/CooldownThroughputTracker.tsx
TJMoats: Yeah, I didn't really know how to handle this one. I have no idea what the module I changed this for was trying to accomplish.

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/93f86fead27addf6aec9a2d5860f6e43f246bb45/src/parser/priest/discipline/modules/features/CooldownThroughputTracker.ts#L43

I was trying to fix the typing here ^^ 

I'd LOVE some additional insight into what this is actually doing.

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
TJMoats: The base event doesn't have ability: Ability which is what is being used in this module. That's why I manually added the possible types.

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
TJMoats: Again, this module uses things that exist in ModifiedDamageEvent | ModifiedHealEvent, but not in the base Event<any>. How should I go about typing that?

src/parser/priest/discipline/modules/spells/Penance.tsx
TJMoats: I can't typecheck this at function call time as the field is added 100% in this module. Are you saying that I should add IsPenanceDamageEvent typechecking for the other modules that use the penanceBoltNumber?

src/parser/priest/discipline/modules/spells/Penance.tsx
TJMoats: I can't use the typechecking to guard this function, this function is what adds the field to the event.

src/parser/core/EventGrouper.ts
TJMoats: Fixed!

src/parser/priest/discipline/modules/Abilities.ts
TJMoats: Fixed!

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
TJMoats: Fixed!

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
TJMoats: Fixed!

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
TJMoats: I've added casting to every place that you made this request. Lemme know if I should go down a different route!

src/parser/priest/discipline/modules/spells/Penance.tsx
TJMoats: Ignore that, I get it now :D

src/parser/priest/discipline/modules/spells/Evangelism.tsx
TJMoats: Fixed!

src/parser/shared/modules/CooldownThroughputTracker.tsx
TJMoats: For now I've commented out the code in that module

tsconfig.json
rp4rk: I think I authored it back when it was looking like the only sane way of handling penance and Atonement was to group events within a threshold, but from the looks of things `downLevelIteration` brings typescript closer inline as a superset of ES so I'm not certain here. I'll keep this in mind as I'm working through the Discipline stuff for shadowlands and see how it pans out.

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
Dambroda: I have already added a type guard for this exact sitch! 
```typescript
     if(!HasAbility(event)) {
        return;
      }
```
Can add that right after the first `fixedEvents.push(event)` and all of the other instances of event.ability shouldn't error. Your casting is also fine in this case.

The reason this is important is because by telling the method that the arguments are of a certain type, the scope of the function will take your word as law. The way you had it originally, you could technically write `console.log(event.ability.guid)` on L20, before any of the `event.type` checks, and typescript would say that's fine. However, because `normalize` actually looks at all events, not only ability events, you would always get a crash that says `event.ability` is undefined when you try to access `guid`.

Technically resolved conversation, I'll leave casting vs typeguard up to you

src/parser/shared/modules/CooldownThroughputTracker.tsx
Dambroda: I had to look into this,

image

it looks like its perfectly valid, and used in interface/others/Cooldown.js L 294-297

In the core CooldownThroughputTracker (i.e. the file this comment was originally for):
```typescript
export type SummaryDef = {
  label: string,
  tooltip: string,
  value: number | string,
}
export type CooldownSpell = {
  spell: any,
  summary: Array<BUILT_IN_SUMMARY_TYPES | SummaryDef>,
  startBufferFilter?: EventFilter<any>,
  startBufferMS?: number,
  startBufferEvents?: number,
  petID?: number,
};
```

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
```diff
@@ -1,35 +1,35 @@
...
+        !(event as DamageEvent).targetIsFriendly &&
+        ATONEMENT_DAMAGE_SOURCES[(event as DamageEvent).ability.guid]
       ) {
         _damageEventIndexes.push(eventIndex);
         _encounteredTargetIDs = [];
         return;
       }
 
-      if (event.type === EventType.Heal && isAtonement(event)) {
+      if (event.type === EventType.Heal && isAtonement((event as HealEvent))) {
```
Dambroda: Just as a comment, I am going to PR a fix for this so that you don't have to cast this every time. I'll @ you to see the fix

src/parser/priest/discipline/modules/spells/Schism.tsx
```diff
@@ -88,6 +97,7 @@ class Schism extends Analyzer {
     this._lastSchismCast = event;
 
     // Assume every schism is bad
+    // @ts-ignore
```
Dambroda: This probably needs to be cleaned up; right now the event itself is being used to index the object, could that be changed to index on the events timestamp?

src/parser/priest/discipline/modules/spells/Atonement.tsx
```diff
@@ -73,37 +83,41 @@ class Atonement extends Analyzer {
...
       debug && console.warn('Atonement: was applied prior to combat');
     }
     const timeSinceApplication = event.timestamp - refreshedTarget.lastAtonementAppliedTimestamp;
     if (timeSinceApplication < ((this.atonementDuration) - IMPROPER_REFRESH_TIME)) {
       this.improperAtonementRefreshes.push(refreshedTarget);
+      // @ts-ignore
       debug && console.log(`%c${this.combatants.players[event.targetID].name} refreshed an atonement too early %c${timeSinceApplication}`, 'color:red', this.currentAtonementTargets);
       this.eventEmitter.fabricateEvent({
         type: 'atonement_refresh_improper',
-        timestamp: event.timestamp,
```
Dambroda, line 106: I believe these (the timestamps) should stay, and the TS ignore can be moved up. I didn't realize how badly the shape was for the EventEmitter, but once that's fixed the ignores shouldn't be needed

src/parser/priest/discipline/modules/features/PurgeTheWicked.tsx
Dambroda: Do you know if you want to keep the decimals? If you do `formatPercentage(recommended, 0)` it will look the same as the previous Math.round (I believe)

src/parser/priest/discipline/modules/features/Checklist/Component.tsx
```diff
@@ -14,8 +14,8 @@ import Requirement from 'parser/shared/modules/features/Checklist/Requirement';
 import PreparationRule from 'parser/shared/modules/features/Checklist/PreparationRule';
 import GenericCastEfficiencyRequirement from 'parser/shared/modules/features/Checklist/GenericCastEfficiencyRequirement';
 
-const DisciplinePriestChecklist = ({ combatant, castEfficiency, thresholds }) => {
-  const AbilityRequirement = props => (
+const DisciplinePriestChecklist = ({ combatant, castEfficiency, thresholds }: any) => {
```
Dambroda: `combatant: Combatant, castEfficiency: CastEfficiency, thresholds: any` (im pretty sure I have the thresholds type somewhere added but I think it requires going through *every* one of your thresholds and making sure they are correct.. wont block this PR for that)

tsconfig.json
Dambroda: Was this resolved? I wasn't sure if the final decision was to actually change tsconfig or to add some tsignores where the different iterables were used

Dambroda:
Conflicts, that one question about tsconfig (wary about changing anything there), and then LGTM :)

tsconfig.json
TJMoats: I think the options are

1) Change the config
2) Rewrite the module

I don't think that this is something that will ts-ignore well.

Dambroda: @rp4rk From what you looked into is this config safe to merge in?

TJMoats:
@Dambroda 

I reverted the modules that were being pains back to js. 
I reverted the config change. 
I reverted the public/protected change.

src/parser/priest/discipline/normalizers/PowerWordRadianceNormalizer.ts
```diff
@@ -26,15 +26,15 @@ class PowerWordRadianceNormalizer extends EventsNormalizer {
       fixedEvents.push(event);
 
       if (event.type === EventType.Cast) {
-        const spellId = event.ability.guid;
+        const spellId = (event as CastEvent).ability.guid;
```
Dambroda: With new AnyEvent you can remove these casts now :D

src/parser/priest/discipline/normalizers/AtonementSuccessiveDamage.ts
Dambroda: I've put this fix in, AnyEvent now doesn't require the casts in this file (`as XEvent` should all be able to remove)

src/parser/priest/discipline/modules/spells/Penance.js
```diff
@@ -61,7 +57,7 @@ class Penance extends Analyzer {
 
     this.eventGrouper.processEvent(event);
 
-    event.penanceBoltNumber = this.currentBoltNumber;
+    (event).penanceBoltNumber = this.currentBoltNumber;
```
Dambroda: unneeded parenthesis

src/parser/priest/discipline/modules/spells/Contrition.tsx
```diff
@@ -4,26 +4,29 @@ import SPELLS from 'common/SPELLS';
...
     statTracker: StatTracker,
     penance: Penance, // we need this to add `penanceBoltNumber` to the damage and heal events
   };
+  protected statTracker!: StatTracker;
+  protected penance!: Penance;
 
   healing = 0;
   damagePenalty = 0;
-  penanceBoltEstimation;
+  penanceBoltEstimation: any
```
Dambroda: semicolon

src/parser/priest/discipline/modules/features/AtonementHealingDone.tsx
```diff
@@ -5,17 +5,21 @@ import SpellLink from 'common/SpellLink';
...
 import Analyzer from 'parser/core/Analyzer';
 import HealingValue from 'parser/shared/modules/HealingValue';
+import { AbsorbedEvent, DamageEvent, HealEvent } from 'parser/core/Events';
+import { IsPenanceDamageEvent } from 'parser/priest/discipline/modules/spells/Helper';
 
 import isAtonement from '../core/isAtonement';
 import AtonementDamageSource from './AtonementDamageSource';
 import AtonementHealingBreakdown from './AtonementHealingBreakdown';
-import Penance from '../spells/Penance';
+import Penance  from '../spells/Penance';
```
Dambroda: accidental extra space

Dambroda (approved):
Few small formatting things that got messed up with the merge, but I am approving this overall. Fix any/none, merge, and then in the next few weeks I'll take the second pass through disc and finish the conversions and resolve any of the leftover comments

TJMoats:
Changes made for that feedback!

Issue #3726, opened 2020-09-21
[Shadowlands][Discipline Priest] Shadowlands Update


rp4rk:
# Prepatch
These are the changes that should be prioritised in the event that we would like to support prepatch, this is unlikely but it makes sense to prioritise them just to be safe. 

## Spellbook
* [ ] Add Shadow Covenant module to encompass new mechanics
* [x] Add Mind Blast to the list of spells considered for Atonement
* [x] Add Mind Sear to the list of spells considered for Atonement
* [x] Update cooldown information for Power Word: Solace
* [ ] Add Spirit Shell module

# Shadowlands
These changes will not be needed until the logs for Shadowlands raid content start becoming available, meaning they can be undertaken as part of hacktober safely.

## Legacy
* [x] Delete Azerite Power modules

## Spellbook
* [ ] Power Infusion

### Covenant Abilities
* [x] Mindgames
* [x] Boon of the Ascended
* [ ] Unholy Transfusion (_Low priority, at risk of being changed before launch_)
* [ ] Fae Blessings (_No Discipline specific interactions, reach out to other priest developers_)

### Covenant Signatures
> _Note these may be implemented at a much more generic level, be sure to check before starting work on them._

* [ ] Door of Shadows
* [ ] Phial of Serenity
* [ ] Fleshcraft
* [ ] Soulshape

## Legendaries
- [ ] Clarity of Mind
- [ ] The Penitent
- [ ] Kiss of Death

## Conduits
- [ ] Exaltation
- [ ] Shattered Perceptions
- [ ] Courageous Ascension
- [ ] Clear Mind
- [x] Shining Radiance
- [ ] Fae Fermata (_No Discipline specific interactions, reach out to other priest developers_)
- [ ] Chartiable Soul
- [ ] Light's Inspiration
- [ ] Swift Penitence
- [ ] Pain Transformation
- [ ] Rabid Shadows
- [ ] Festering Transfusion
- [ ] Mind Devourer

## Soulbinds
_There are currently no Discipline specific interactions from Soulbinds that are known._

ThePkell:
Hey guys, I'm really new to this, but is there something that you can task out to me that I can work on?

karlpralow:
hey @ThePkell there's still alot to be done so your help would be great. There's multiple things to do right now, if you want to hit me up on discord we can discuss this (Oratio in Focused Will)

Ogofo:
Hello,
[...]

PR #3792, opened 2020-10-03
[Discipline] Power Word: Solace and Mind Games bugfix

rp4rk:
# Description
Updates the cooldown of Power Word: Solace to be 15 seconds as described in [this issue](https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3726).

Also fixes the casing of Mind Games import to be correct.

### DIFF
--- a/src/parser/priest/discipline/CombatLogParser.js
+++ b/src/parser/priest/discipline/CombatLogParser.js
@@ -40,7 +40,7 @@ import Schism from './modules/spells/Schism';
 
 import SinsOfTheMany from './modules/spells/SinsOfTheMany';
 
-import Mindgames from './modules/shadowlands/covenants/Mindgames';
+import MindGames from './modules/shadowlands/covenants/MindGames';
 
 import { ABILITIES_AFFECTED_BY_HEALING_INCREASES } from './constants';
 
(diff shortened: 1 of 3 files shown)


src/parser/priest/discipline/CombatLogParser.js
```diff
@@ -97,7 +97,7 @@ class CombatLogParser extends CoreCombatLogParser {
     lucidDreams: LucidDreams,
 
     // Covenants
-    mindgames: Mindgames,
+    mindgames: MindGames,
```
karlpralow, line 100: it's Mindgames, not Mind Games actually.
https://shadowlands.wowhead.com/spell=323706/mindgames

src/parser/priest/discipline/CombatLogParser.js
```diff
@@ -40,7 +40,7 @@ import Schism from './modules/spells/Schism';
 
 import SinsOfTheMany from './modules/spells/SinsOfTheMany';
 
-import Mindgames from './modules/shadowlands/covenants/Mindgames';
+import MindGames from './modules/shadowlands/covenants/MindGames';
```
Dambroda, line 43: It's supposed to be Mindgames; file needs to be renamed (class inside and other references were already fixed)

PR #3797, opened 2020-10-04
[Mistweaver] update MistweaverHealingEfficiency.js

pingypong:
fixed a few copy/paste errors affecting the overheal column in detailed view


src/parser/monk/mistweaver/modules/features/MistweaverHealingEfficiencyTracker.js
```diff
@@ -103,23 +104,26 @@ class MistweaverHealingEfficiencyTracker extends HealingEfficiencyTracker {
...
     return spellInfo;
   }
 
   getYulonDetails(spellInfo) {
     // Get all soothing breath and enveloping breath healing since its all bc of yu'lon
     spellInfo.healingDone = this.healingDone.byAbility(SPELLS.SOOTHING_BREATH.id).effective;
-    spellInfo.overhealingDone = this.healingDone.byAbility(SPELLS.SOOTHING_BREATH.id).effective;
-    // Enveloping breath part
+    spellInfo.overhealingDone = this.healingDone.byAbility(SPELLS.SOOTHING_BREATH.id).overheal;
+    // Enveloping breath part -- should this be attributed to yulon? we attribute viv cleave healing to rem but here we don't attribute the envB to envM? will keep for now
```
abelito75, line 115: This makes sense. Put it there

abelito75 (changes requested):
Outside of the envb thing it looks good

pingypong:
care before merging this, looks like there's an issue with mana efficiency caused by another change on shadowland.wowanalyzer.com: 
image

Pewtro:
have you tried to run `rm -rf nodemodules` followed by a `yarn`?

pingypong:
my local is running fine, that SS is from shadowlands.wowanalyzer.com

abelito75:
I should have stated this then leaving it implied but you should go edit MistweaverHealingEfficiencyDetails to state that envb in bundled up with env

pingypong:
I'm not sure what you mean @abelito75 by that, do you a more detailed comment?

abelito75:
MistweaverHealingEfficiencyDetails is where we state that 
`<SpellLink id={SPELLS.GUSTS_OF_MISTS.id} /> healing is added to the appropriate spell that caused the gust. <SpellLink id={SPELLS.ESSENCE_FONT.id} /> is given the healing from duplicated gusts, since without <SpellLink id={SPELLS.ESSENCE_FONT.id} /> the second gust would not have happened. <SpellLink id={SPELLS.RENEWING_MIST.id} /> is given the splash healing of <SpellLink id={SPELLS.VIVIFY.id} />'s heal since without <SpellLink id={SPELLS.RENEWING_MIST.id} />, <SpellLink id={SPELLS.VIVIFY.id} /> wouldn't have splashed`
which means we should also say the same about enveloping mist and enveloping breath

pingypong:
added some breaks in as well since it was becoming quite the wall of text 
image

PR #3798, opened 2020-10-04
Remove channel duration from cast efficiency calculations

Juko8:
Cooldown starts when channels start, and it's therefore not relevant to include channeling time the same way as casted abilities that start their cooldown on finished cast. 

Channeled spells before: 
billede
billede

Fists of Fury after: (couldn't find MM logs that work in both live website and my SL branch)
billede

There might arise some issues if a channeled spell can reset it's cooldown during its channel, but i am not aware of any spell being able to do so, and this will give a more accurate cast efficiency regardless. I looked into some options of handling timeSpentChanneling similarly to how timeSpentOnGCD, but could not figure out a solution, so i will leave that up to someone else if they find it necessary. 

Also includes a small change to WW channeling because there is some weirdness with Fists of Fury channel that i fixed while doing this.


Pewtro (approved):
Thanks for doing this! Should fix quite a few issues that pop up every now and then! 👍🏼

PR #3804, opened 2020-10-04
Update priest.js

jalbane:
Add Boon of the Ascended covenant ability.

### DIFF
--- a/src/common/SPELLS/shadowlands/covenants/priest.js
+++ b/src/common/SPELLS/shadowlands/covenants/priest.js
@@ -1,6 +1,15 @@
 export default {
   //region Kyrian
-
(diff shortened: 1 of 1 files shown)


Pewtro:
This doesn't add Boon of the Ascended anywhere - it needs to be added to the spellbook either through the abilities file either directly or through its own module (probably the best in this case since it can then easily be shared across all 3 specs). 
See Covenant ability implementation in hunter module here for examples: https://github.com/WoWAnalyzer/WoWAnalyzer/tree/shadowlands/src/parser/hunter/shared/modules/spells/covenants

Pewtro:
Closing this since #3851 is fully implemented version of this Covenant Ability.

PR #3806, opened 2020-10-04
Update report header design

MartijnHols:
Old
image

New
image


Dambroda:
Thoughts on the MM:SS:MSMS boxes text being right-aligned and/or shrunk a bit more? I'm still not even sure about the time filters being different boxes

MartijnHols:
I'd have made them a single box if I thought it was easy to make usable

Issue #3799, opened 2020-10-04
Convert project to TypeScript


MartijnHols:
To improve maintainability and ease of development, we would like to convert all files to TypeScript. Helping with this does not require any WoW or class knowledge. Just pick a random JavaScript file, change its extension to `.ts` or `.tsx` (required if the file contains JSX) and start fixing type issues. Generally the easiest are spec-specific modules with no dependencies. Core modules can be hard to add typing to. When converting React component, please also consider #TODO: converting it to a functional component.

Avoid using vague types like `any` or `object`, use `unknown` instead if necessary.

Because there are a lot of files in the project, and most of them were written without consideration for types, this is a lot of work that can not be done in one go. This issue serves to point anyone interested in the right direction.

Spec maintainers can still opt out of TS if they desire.

abelito75:
Is this far enough along where we can close it? The good old github break down says we are over 98% of the way there?

MartijnHols:
Yes! Terrific work everyone. Thanks you so much!

PR #3812, opened 2020-10-05
refactor: convert to typescript

NikoDyring:
This is a contribution to #3799 
Converted url parsing to typescript. Added an interface for the QueryParams. 
Also added an interface for the parseVersionString function


src/common/makeWclApiUrl.ts
```diff
@@ -1,12 +1,13 @@
 import makeApiUrl from './makeApiUrl';
+import { QueryParams } from './QueryParams';
 
-const WCL_API_KEY = process.env.REACT_APP_WCL_API_KEY || undefined;
-if (!WCL_API_KEY && process.env.NODE_ENV === 'development') {
```
MartijnHols, line 4: You are changing the behavior here in such a way that I think this would now break in production. I am pretty sure the development check is required.

src/common/makeWclApiUrl.ts
```diff
@@ -1,12 +1,13 @@
...
+const WCL_API_KEY = process.env.REACT_APP_WCL_API_KEY ?? undefined;
```
MartijnHols, line 4: || May be important here since `REACT_APP_WCL_API_KEY` is likely a string.

src/common/makeWclApiUrl.ts
NikoDyring: If there's no key, then it would be on development. But even if you pass an empty string, it would still be defined as a string and not an undefined. This would still break production, if it was an empty key. Right?

src/common/makeWclApiUrl.ts
NikoDyring: I've always thought it was better to use nullish coalescing operators than logical disjunction. Therefore my change. 👍

src/common/makeWclApiUrl.ts
MartijnHols: The webapp used by the production environment also omits the API key because it is the only way to keep it a secret. This check ensures the app still works in production with that setup.

The env var `REACT_APP_WCL_API_KEY` could be omitted entirely, which would make it `undefined`. It could also be left to an empty value like `REACT_APP_WCL_API_KEY=` (that being the entire line), which would make it an empty string.

src/common/makeWclApiUrl.ts
MartijnHols: Sure, but in this case the desired behavior is to turn an empty string into undefined as well.

src/common/makeWclApiUrl.ts
NikoDyring: Alright! Thanks for teaching me some new stuff. 👍 Guess we will disregard those changes in makeWclApiUrl.js then. Is there any other issues with the other contributions I've done?

src/common/makeWclApiUrl.ts
NikoDyring: Alright! Thanks for teaching me some new stuff. 👍 Guess we will disregard those changes in makeWclApiUrl.js then. Is there any other issues with the other contributions I've done?

NikoDyring:
Gonna start over. I'm way tired to keep working on this. 🙈 Will open a new pull request tomorrow when I've got time to contribute. Sorry for wasting your time @MartijnHols

PR #3813, opened 2020-10-05
Convert Time components to Typescript

### LINKED ISSUE #3800: Convert React class components to functional components
To improve maintainability and ease of development, we would like to convert all React class component to functional components. Helping with this does not require any WoW or spec knowledge. Just pick a random React class component and convert it to a functional component. When doing this, please also consider #3799 as TS Props are much cleaner than propTypes.

Reimplement state and other methods using hooks.

When converting components that use the Redux or react-router HOCs, please consider using hooks instead.

chagriali:
This PR fixes #3800.


MartijnHols (approved):
Sweet, thanks!

PR #3814, opened 2020-10-06
Convert ReportDurationWarning from React class component to functional component

### LINKED ISSUE #3800: Convert React class components to functional components

William-Schutte:
fixes #3800 
Converted a smaller component (no state/hooks). Converted getter to generic function. Nothing else altered.

Let me know if this needs any further modifications!

### DIFF
--- a/src/interface/report/ReportDurationWarning.tsx
+++ b/src/interface/report/ReportDurationWarning.tsx
@@ -18,16 +18,16 @@ const ReportDurationWarning = ({ duration }: Props) => {
   };
 
   return (
-      <div className="container">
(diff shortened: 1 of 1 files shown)


MartijnHols (approved):
Thanks!

PR #3815, opened 2020-10-06
[Core] fix character and guild search submit action

ChristopherKiss:
When trying to use character or guild search it fails to submit.

error

We can see that the realm is undefined even though it has been selected.

error2

This is because the onChange handler argument is a string of realm not an object.

error3


MartijnHols (approved):
Nice. Thanks!

PR #3816, opened 2020-10-06
Shadowlands

alliepet:


abelito75:
Looks good besides the wrong change log. This should be in the mw change log

abelito75:
I got the changelog conflict for you! You should update your branch though to prevent it from happening again!

PR #3818, opened 2020-10-06
[Monk] Fallen Order

abelito75:
the icon is pointing to the correct thing (or so wowHead says) but won't load. I don't think its an issue right now
I don't know how far we want to break things down but I think this grouping is fair
FallenOrderToolTip
FallenOrder


Pewtro:
They only exist on shadowlands.wowhead.com, not on wowhead.com - I assume that'll change once we go into prepatch.

abelito75:
I don't think that is fully true. There are other spells that only exist on Shadowlands that get pulled in just fine.

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    // In my testing this line always returns false but its what putput uses so I have faith
+    // it will work in the future. (might just be a bad log too)
+    this.active = this.selectedCombatant.hasCovenant(COVENANTS.VENTHYR.id);
+
+    if (!this.active) {
+      return;
+    }
+
+    options.abilities.add({
```
Dambroda, line 42: rather than using options.abilities, use `this.abilities` (after adding `protected abilities!: Abilities;` to L21)

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
```diff
@@ -0,0 +1,144 @@
...
+import ItemHealingDone from 'interface/ItemHealingDone';
+import React from 'react';
+import BoringSpellValueText from 'interface/statistics/components/BoringSpellValueText';
+import { formatNumber } from 'common/format';
+import SpellUsable from 'parser/shared/modules/SpellUsable';
+import COVENANTS from 'game/shadowlands/COVENANTS';
+
+class FallenOrder extends Analyzer {
+  static dependencies = {
+    spellUsable: SpellUsable,
```
Dambroda, line 18: Unused

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    //summon events (need to track this to get melees)
+    this.addEventListener(Events.summon.by(SELECTED_PLAYER).spell([SPELLS.FALLEN_ORDER_OX_CLONE, SPELLS.FALLEN_ORDER_TIGER_CLONE, SPELLS.FALLEN_ORDER_CRANE_CLONE]), this.trackSummons);
+
+    //mistweaver spells
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER_PET).spell([SPELLS.FALLEN_ORDER_ENVELOPING_MIST, SPELLS.FALLEN_ORDER_SOOTHING_MIST]), this.mistHealingTracker);
+    //brewmaster spells
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER_PET).spell([SPELLS.FALLEN_ORDER_KEG_SMASH, SPELLS.FALLEN_ORDER_BREATH_OF_FIRE,SPELLS.BREATH_OF_FIRE_DEBUFF]), this.brewDamageTracker);
+    //windwalker spells
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER_PET).spell([SPELLS.FALLEN_ORDER_SPINNING_CRANE_KICK, SPELLS.FISTS_OF_FURY_DAMAGE]), this.windDamageTracker);
```
Dambroda, line 63: Is it FISTS_OF_FURY for both WW and other specs? No special ID for FO?

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  handleMelee(event: DamageEvent) {
+    //if we don't know who its from then we can't add it
+    if(!event.sourceID){
+      return;
+    }
+
+    const id: number = event.sourceID;
+    //if we are not tracking this guy we don't know what it is
+    if(!this.cloneMap.has(id)){
```
Dambroda, line 97: You can remove this first check, you're doing it twice (checking for undefined after `.get`)

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
```diff
@@ -0,0 +1,144 @@
...
+      return;
+    }
+
+    const cloneType: number | undefined = this.cloneMap.get(id);
+    //because typescript won't let me just do this.cloneMap.get(id) i might as well add a null check too
+    if(!cloneType){
```
Dambroda, line 103: `=== undefined` is better here, you dont need a falsey check when dealing with this number

Dambroda (changes requested):

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
abelito75: Fallen Order clones are weird af where sometimes it uses the core spell and sometimes they recreate it. In this instance they use the core spell for FoF but then recreate SCK

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
abelito75: We don't talk about that (will fix)

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
abelito75: Yeah. You're right

src/parser/monk/shared/modules/covenants/FallenOrder.tsx
abelito75: Thanks!

Dambroda:
> I don't think that is fully true. There are other spells that only exist on Shadowlands that get pulled in just fine.
> If you check grounding breath you can see it work just fine
Im assuming it has to do with whether the spell re-uses an existing icon or is a brand new icon for shadowlands

abelito75:
Ahhh. That makes sense now that you put it that way. Silly putput thinking I had a brain.

Dambroda (approved):

PR #3820, opened 2020-10-06
[MW] Updated Renewing Mist to Use Event Listeners

anom0ly:
Simple update to REM module to use event listeners.


abelito75 (approved):

abelito75:
Fix up the change log then we can blast!

PR #3822, opened 2020-10-06
[MW] Update Spinning Crane Kick and Essence Font modules to Event Listeners

anom0ly:
Updating EF and SCK modules to use event listeners.


abelito75:
It won't let me flag it but there is an unused variable in EF (its channel time)

abelito75 (approved):

PR #3825, opened 2020-10-07
Combatant / RShaman Typescript

niseko:
Updated my WCLFetch files using the new types. Converted the Combatant file as well to get this to work.


src/parser/shared/modules/Combatants.ts
```diff
@@ -1,31 +1,34 @@
+import { AnyEvent, HasTarget } from 'parser/core/Events';
 import Entities from './Entities';
 import Combatant from '../../core/Combatant';
 
 class Combatants extends Entities {
-  players = {};
+  players: Array<Combatant> = [];
```
Dambroda: I could be wrong here, but I think this is probably best left as an object with the index as a number. Wouldn't an array fill in a bunch of `undefined`s for the unused playerIDs?

src/parser/shaman/restoration/modules/talents/AncestralProtectionTotem.tsx
```diff
@@ -13,26 +13,31 @@ import STATISTIC_CATEGORY from 'interface/others/STATISTIC_CATEGORY';
...
     combatants: Combatants,
   };
+
+  protected combatants!: Combatants;
+
   loaded = false;
-  aptEvents = [];
-  constructor(...args) {
-    super(...args);
+  aptEvents: Array<Event<EventType.Cast | EventType.ApplyDebuff>> = [];
```
Dambroda: Does `Array<CastEvent | ApplyDebuffEvent>` work here or does it scream at you?

src/parser/shared/modules/Combatants.ts
niseko: The parent has an array defined so I had to decide between either changing this to an array or the parent to an object
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/src/parser/shared/modules/Entities.ts#L35

So, guess I'm changing both to an object then.

src/parser/shaman/restoration/modules/talents/AncestralProtectionTotem.tsx
niseko: Does work, which also means that I can remove one of the ifs further down. 👍

src/parser/shared/modules/Combatants.ts
Dambroda: This is reminding me, during the original conversion I feel like it was weird because one of the Entities subclasses used an array and the others all used an object. If you were able to convert both to object and no tests broke then LGTM

niseko: image
everything seems to be fine

src/parser/shaman/restoration/modules/talents/AncestralProtectionTotem.tsx
```diff
@@ -56,11 +61,12 @@ class AncestralProtectionTotem extends Analyzer {
         OR
         (type='${EventType.Cast}' AND ability.id=${SPELLS.TOTEMIC_REVIVAL_CAST.id})
       )`,
+      timeout: 2000,
```
Dambroda, line 64: It feels wrong that WclOptions requires a timeout when it has a default value. Obviously not in the scope of this

Dambroda (approved):

Dambroda:
new eslint conficts KEKW, enjoy. LGTM tho :)

PR #3828, opened 2020-10-07
[MW] Final Spells Update to Event Listeners

anom0ly:
Final spells, Soothing Mist and Enveloping Mist, updated to use Event Listeners.


src/parser/monk/mistweaver/modules/spells/EnvelopingMists.js
```diff
@@ -23,55 +22,45 @@ class EnvelopingMists extends Analyzer {
...
-    if (debug) {
-      console.log(`EvM Healing Contribution: ${this.healingIncrease}`);
-      console.log(`EnM Boost`, this.evmHealingIncrease);
-      console.log("gusts env healing: ", this.gustsHealing);
-    }
-  }
+ 
 
   statistic() {
     return (
```
abelito75, line 68: I can't flag it directly and its not vital to this pr but we need to go back and update all StatisticBoxes to be Statistics instead.

src/parser/monk/mistweaver/modules/spells/EnvelopingMists.js
```diff
@@ -23,55 +22,45 @@ class EnvelopingMists extends Analyzer {
...
-    if ((spellId === SPELLS.GUSTS_OF_MISTS.id) && (this.lastCastTarget === event.targetID) && this.numberToCount >0) {
+    if ((spellId === SPELLS.GUSTS_OF_MISTS.id) && (this.lastCastTarget === targetId) && this.numberToCount >0) {
       this.gustProc += 1;
       this.gustsHealing += (event.amount || 0) + (event.absorbed || 0);
       this.numberToCount -= 1;
     }
-
+    
     if (this.combatants.players[targetId]) {
       if (this.combatants.players[targetId].hasBuff(SPELLS.ENVELOPING_MIST.id, event.timestamp, 0, 0) === true) {
```
abelito75: Now that I think about it this is wrong as this only checks if they have the buff, not that the buff is YOUR buff. This means our life cocoon modules are wrong too but its outside of the scope of this pr so lets merge it and i'll circle back and fix it up!

src/parser/monk/mistweaver/modules/spells/EnvelopingMists.js
```diff
@@ -23,55 +22,45 @@ class EnvelopingMists extends Analyzer {
...
     const spellId = event.ability.guid;
-
+    
     if (UNAFFECTED_SPELLS.includes(spellId)) {
-      debug && console.log('Exiting');
       return;
     }
 
-    if ((spellId === SPELLS.GUSTS_OF_MISTS.id) && (this.lastCastTarget === event.targetID) && this.numberToCount >0) {
+    if ((spellId === SPELLS.GUSTS_OF_MISTS.id) && (this.lastCastTarget === targetId) && this.numberToCount >0) {
```
abelito75: This should probably be its own block of code in its own method

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,30 +31,35 @@ class SoothingMist extends Analyzer {
...
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.handleSoothingMist);
+    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.SOOTHING_MIST), this.removeBuffSoothingMist);
+    this.addEventListener(Events.fightend, this.end);
   }
 
-  on_byPlayer_heal(event) {
+  handleSoothingMist(event) {
     const spellId = event.ability.guid;
 
     if (spellId === SPELLS.SOOTHING_MIST.id) {
```
abelito75: This should probably be its own block of code in its own method

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,30 +31,35 @@ class SoothingMist extends Analyzer {
...
       this.soomTicks += 1;
       this.lastSoomTickTimestamp = event.timestamp;
     }
 
-    if (spellId === SPELLS.GUSTS_OF_MISTS.id && this.lastSoomTickTimestamp === event.timestamp && this.gustProc < Math.ceil(this.soomTicks/8)) {
+    if (spellId === SPELLS.GUSTS_OF_MISTS.id && this.lastSoomTickTimestamp === event.timestamp && this.gustProc < Math.ceil(this.soomTicks / 8)) {
```
abelito75: This should probably be its own block of code in its own method

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,30 +31,35 @@ class SoothingMist extends Analyzer {
...
 
-    if(this.soomInProgress){
-      if(spellId === SPELLS.VIVIFY.id || spellId === SPELLS.ENVELOPING_MIST.id) {
+    if (this.soomInProgress) {
+      if (spellId === SPELLS.VIVIFY.id || spellId === SPELLS.ENVELOPING_MIST.id) {
         this.castsInSoom += 1;
       }
-      if(spellId === SPELLS.SOOTHING_MIST.id) {//if they refresh soom for some stupid reason
+
+      if (spellId === SPELLS.SOOTHING_MIST.id) {//if they refresh soom for some stupid reason
```
abelito75: This should probably be its own block of code in its own method

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,30 +31,35 @@ class SoothingMist extends Analyzer {
...
   }
 
-  on_byPlayer_cast(event) {
+  castSoothingMist(event) {
     const spellId = event.ability.guid;
 
-    if(this.soomInProgress){
-      if(spellId === SPELLS.VIVIFY.id || spellId === SPELLS.ENVELOPING_MIST.id) {
+    if (this.soomInProgress) {
+      if (spellId === SPELLS.VIVIFY.id || spellId === SPELLS.ENVELOPING_MIST.id) {
```
abelito75: This should probably be its own block of code in its own method

abelito75 (changes requested):
Gotta break it down some more or you're basically using the old magic method but renamed to your liking :P

anom0ly:
I think I should have gotten it all squared away.

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,55 +31,61 @@ class SoothingMist extends Analyzer {
   constructor(...args) {
     super(...args);
     this.assumedGCD = 1500 / (1 + this.statTracker.hastePercentage(this.statTracker.currentHasteRating)) *.95;
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SOOTHING_MIST), this.castSoothingMist);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.VIVIFY), this.castSoothingMistVivify);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.ENVELOPING_MIST), this.castSoothingMistEnvelopingMist);
```
abelito75: You could combine both of these into 1 action listener and 1 function since the functions are the exact same
something like
`this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell([SPELLS.ENVELOPING_MIST, SPELLS.VIVIFY]), this.castDuringSoothingMist);`

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
```diff
@@ -33,55 +31,61 @@ class SoothingMist extends Analyzer {
...
+    if (this.soomInProgress) {
+      this.castsInSoom += 1;
     }
   }
 
-  on_byPlayer_removebuff(event) {
-    const spellId = event.ability.guid;
+  castSoothingMist(event) {
+    //if they refresh soom for some stupid reason
+    this.endStamp = event.timestamp;
```
abelito75: This had a 
if(this.soomInProgress) before hand to check if its a recast while they are already channeling it. It needs to stay in
IE
`if(this.soomInProgress) {
        this.endStamp = event.timestamp;
        this.checkChannelTiming();
        this.castsInSoom = 0;
}`

abelito75 (changes requested):
Almost there two things
One thing to keep thing simple and not repeat code
The other thing is just an if statement went missing

abelito75:
Updated change log for you buddy 👍

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
anom0ly: Thanks. I didn't realize we could track 2 or more spells per listener. Updated in newest push

src/parser/monk/mistweaver/modules/spells/SoothingMist.js
anom0ly: Too many line deletes at once. I added back in the check to this. Thanks!

anom0ly:
> 
> 
> Updated change log for you buddy 👍

Thanks! Should have all updates in, including a logic change I believe makes sense.

Let me know if you see anything in your review.

abelito75 (approved):

abelito75:
Thanks for the pr!

PR #3829, opened 2020-10-07
[MW] Add stack management and gcd usage stats to InvokeChiji module

vohrr:
image


src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -2,23 +2,72 @@ import React from 'react';
...
+  handleChijiStart(event) {
+    this.chijiActive = true;
+    this.chijiStart = this.lastGlobal = event.timestamp;
+    this.chijiGlobals += 1;
+    this.chijiUses +=1;
+  }
+
+  handleGlobal(event){
+    //cant get eventlistener to find chiji's death so we will check against duration here
+    if(event.timestamp - this.chijiStart > SPELLS.INVOKE_CHIJI_THE_RED_CRANE_TALENT.duration) {
```
abelito75: This seems like a very round about way to determine if chi-ji is alive or death. The direct method would be to create the action listener
`this.addEventListener(Events.death.to(SELECTED_PLAYER_PET), this.handleChiJiDeath);`
You will still need a single variable for true/false if he is up or not but its much cleaner and will avoid errors in the future

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -28,6 +77,27 @@ class InvokeChiJi extends Analyzer {
   handleEnvelopingBreath(event) {
     this.envelopHealing += (event.amount || 0) + (event.absorbed || 0);
   }
+  
+  //stackbreakown management
+  handleApplyChijiStack() {
```
abelito75: From looking at logs this appears to never fire when you're at 3 stacks as you don't gain a stack in game.
You're going to have to track damage events from bok, sck, rsk, and totm bok to determin wasted stacks

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -2,23 +2,72 @@ import React from 'react';
...
+
+  handleGlobal(event){
+    //cant get eventlistener to find chiji's death so we will check against duration here
+    if(event.timestamp - this.chijiStart > SPELLS.INVOKE_CHIJI_THE_RED_CRANE_TALENT.duration) {
+      this.chijiActive = false;
+    }
+    if(this.chijiActive) {
+      this.chijiGlobals += 1;
+      //if timebetween globals is longer than the gcd add the difference to the missed gcd tally
+      //we dont care about accounting for channels/casts longer than the gcd since you shouldnt be using them during chiji anyway
```
abelito75: If you're taking kyrian it might actually be worth to channel essence font during the middle of chi-ji so the time between should be accounted for imo
these could be handled with beginchannel and endchannel events

abelito75 (changes requested):
Seems close

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
vohrr: F me I was trying to do death.by not death.to before. readded

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
vohrr: refactored to use damage events for stack management

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
vohrr: added handling for kyrian chiji

vohrr:
@abelito75  updated

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -1,33 +1,136 @@
...
+  }
+
+  handleEssenceFontStart(event) { 
+    if(this.chijiActive && this.selectedCombatant.hasBuff(SPELLS.WEAPONS_OF_ORDER_BUFF_AND_HEAL.id)) { 
+       this.efChannelStart = this.lastGlobal = event.timestamp;
+    }
+  }
+
+  handleEssenceFontEnd(event) {
+    const gcd = 1500 / (1 + this.statTracker.hastePercentage(this.statTracker.currentHasteRating));
```
abelito75: iirc ef's channel time is snapshotted at the haste you have at the start of the channel as well as the gcd starting at the start of the channel
On top of this iirc essence font itself has a GCD event where you can just grab the gcd from
So I might have miss lead you with start and end channel since you can do it with
gcd and end channel

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -1,33 +1,136 @@
...
     if (!this.active) {return;}
     this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell(SPELLS.GUST_OF_MISTS_CHIJI), this.handleGust);
     this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell(SPELLS.ENVELOPING_BREATH), this.handleEnvelopingBreath);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.ENVELOPING_MIST), this.handleEnvelopCast);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.INVOKE_CHIJI_THE_RED_CRANE_TALENT), this.handleChijiStart);
+    this.addEventListener(Events.death.to(SELECTED_PLAYER_PET), this.handleChijiDeath);
+    this.addEventListener(Events.death.to(SELECTED_PLAYER), this.handleChijiDeath);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell([SPELLS.BLACKOUT_KICK, SPELLS.RISING_SUN_KICK_SECOND, SPELLS.BLACKOUT_KICK_TOTM]), this.handleOvercapStacks)
+    //need a different eventlistener beacause chiji currently only applies 1 stack per cast of sck, not on each dmg event
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SPINNING_CRANE_KICK), this.handleOvercapStacks);
```
abelito75: You gotta make sure sck does damage

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
```diff
@@ -1,33 +1,136 @@
...
+import StatTracker from 'parser/shared/modules/StatTracker';
+
+/** 
+ * Blackout Kick, Totm BoKs, Rising Sun Kick and Spinning Crane Kick generate stacks of Invoke Chi-Ji, the Red Crane, which reduce the cast time and mana 
+ * cost of Enveloping Mist by 33% per stack, up to 3 stacks.
+ * These abilities also heal 2 nearby allies for a Gust of Mist heal.
+ * Casting Enveloping Mist while Chiji is active applies Enveloping Breath on up to 6 nearby allies within 10 yards.
+*/
 class InvokeChiJi extends Analyzer {
+  static dependencies = {statTracker: StatTracker};
```
abelito75: plz no one line for static depends

abelito75 (changes requested):
You need to fix the conflict too

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
vohrr: updated

src/parser/monk/mistweaver/modules/talents/InvokeChiJi.js
vohrr: updated to use cast and end channel

vohrr:
@abelito75  I don't appear to have permission to resolve the conflict in git? Unless I'm doing something wrong here
image

abelito75:
You gotta do it in your IDE. You will need to PULL the code down

PR #3831, opened 2020-10-08
[Shadowlands][Core] Delete BFA Modules

Sharrq:
- Removed Azerite Traits from several specs that are not maintained anymore
- Removed the shared BFA Items/Essences/Traits/Enchants/etc
- Removed the BFA Raids (except Nyalotha)
- Updated Snapshots
- Updated the Prepatch Warning to be present tense instead of future tense (For when Prepatch drops)

I opened a bunch of recent logs to look for any issues and fixed everything i found, didnt notice any additional errors. Also I tried not to touch specs that I knew were actively maintained (Aside from removing references to things that were removed).

Also did not touch Spell IDs or Item IDs


Pewtro:
I don't think we should remove gear that people are still using (such as Nyalotha trinkets) -- Azerite being up to the individual maintainer.

PR #3833, opened 2020-10-08
[MW] Deny harm includes both expel harm heals now

alliepet:
Abe missed another one smh


abelito75:
Looking much better. Thanks for updating your branch :D

Issue #3742, opened 2020-09-23
[Core] Update Stat Rating to % calculations


Pewtro:
We need to do multiple changes for this

- [ ] Update stat rating per % at level 50 (X haste, X crit, X mastery spell point, X versatility)
- [ ] Update stat rating per % at level 60 (33 haste, 35 crit, 35 mastery spell point, 40 versatility)
- [ ] Update stat rating per % to account for DR

abelito75:
This was completed

PR #3836, opened 2020-10-08
RFC: [Core] Implement the Diminishing Return system for Stats introduced in Shadowlands

Pewtro:
Towards #3742

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -13,6 +13,7 @@ const prepatch = date(2020, 10, 13);
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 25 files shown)


Pewtro:
Specifically looking for a healer to look over these changes as I replaced some things used in their StatValue modules

Pewtro:
Merging this in 12 hours or so regardless of reviews - even if I'd have liked some eyes on it it cannot wait, we are entirely reliant on this being changed to serve any sort of accurate feedback to the player in the prepatch.

Pewtro:
With prepatch launching today I have decided to merge this - I have verified stat amounts on selected logs, but there could easily be something that need to be looked into. I'll keep my eyes on this using logs from NA raiding tonight.

PR #3839, opened 2020-10-08
Convert Consumable Checkers to TypeScript

Zeboot:


Dambroda (approved):
small things, all non-blocking:

- I think for event listeners its less common to `.bind(this)`
- Spell.ts no NL @ EOF
- Replacing other instances of Spell that were done manually before (if you don't I will go do it right after this merges); eg SpellbookAbility.spell, etc

PR #3841, opened 2020-10-08
[Hunter] Legendaries and tweaks

Pewtro:


src/parser/hunter/beastmastery/modules/items/DireCommand.tsx
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * TODO Find a log with both Dire Command and Dire Beast talent and account for them using the same spell for buff
+ * Kill Command has a 15% chance to also summon a Dire Beast to attack your target.
+ *
+ * Example log:
+ *
+ */
+class DireCommand extends Analyzer {
+
+  damage = 0;
```
abelito75: should probably do 
damage: number = 0;

src/parser/hunter/beastmastery/modules/items/DireCommand.tsx
```diff
@@ -0,0 +1,93 @@
...
+  activeDireBeasts: string[] = [];
+  targetId = '';
```
abelito75: Should give type

src/parser/hunter/beastmastery/modules/items/QaplaEredunWarOrder.tsx
```diff
@@ -0,0 +1,108 @@
...
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.QAPLA_EREDUN_WAR_ORDER_EFFECT.bonusID);
+    if (!this.active) {
+      return;
+    }
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.BARBED_SHOT), this.onBarbedShotCast);
+  }
+
+  get totalPossibleCDR() {
```
abelito75, line 41: Extremely unlikely but you should probably put a check to make sure you don't divide by zero (i know its like almost impossible as a bm to not kill command but its a good idea to avoid it even if edge case)
(i say it here since all division happens off of this)

src/parser/hunter/beastmastery/modules/spells/CobraShot.tsx
```diff
@@ -138,7 +138,8 @@ class CobraShot extends Analyzer {
       >
         <BoringSpellValueText spell={SPELLS.COBRA_SHOT}>
           <>
-            {formatNumber(this.effectiveKCReductionMs / 1000)}s / {this.totalPossibleCDR / 1000}s <br />
+            {formatNumber(this.effectiveKCReductionMs / 1000)}s / {this.totalPossibleCDR / 1000}s
```
abelito75, line 141: Same dealio. Probably should check if bad divides no matter how rare 
As they say at my work users data is dirty

abelito75 (changes requested):
I'm a little anal about possible /0 even if its gonna be a one in a million

abelito75:
Next time please leave a description

PR #3842, opened 2020-10-08
Convert Spells to TypeScript

Zeboot:
I didn't convert BFA spells as they will be gone with slands, I can add those if wanted. Hopefully I caught every spot spells were redefined at


Dambroda (approved):
I hate that your ide is moving import react to the bottom, but everything else looks good

PR #3844, opened 2020-10-09
[Core] Convert Enemy/Enemies to TS; made Entities generic & abstract

Dambroda:
@Pewtro can you confirm the logic in my comment in your careful aim file here? 

I added targetIsFriendly to a lot of events here -- just from the looks of things. for consistency, it seems like targetIsFriendly and sourceIsFriendly should probably exist in any event that has source/sourceID/target/targetID


Pewtro:
I remember it complaining about possible nulls, so I erred on the side of caution when updating it - but it looks fine

Pewtro:
Oof, this needs more manual help than a simple shadowlands merge

Dambroda:
image

Issue #3835, opened 2020-10-08
[Core] Move Stat constants out of StatTracker


Pewtro:
We have a lot of stat constant used in StatTracker.js, that should probably be in its own separate StatConstants.js file or similiar.

anontyro:
hey @Pewtro I can look into this if you like :)

Pewtro:
Specifically we have SPEC_MULTIPLIERS, DEFAULT_BUFFS, statMultiplierBuffs and statMultiplier that could possibly deserve moving out.

anontyro:
sure I can look into moving these :)

Pewtro:
I don't think all need moving tbh 😄

anontyro:
sure I'll look at what makes sense to move and aim to move only those

anontyro:
@Pewtro made a PR for this let me know if there is anything that should be updated/ changed :) and I'll fix it tomorrow

abelito75:
Reading the attempted PR on this it seems its not an issue anymore. Closing.

PR #3846, opened 2020-10-09
[Core] Move Stat constants out of StatTracker

anontyro:
Split out the static objects into there own file along with slight renaming and fixing the statTracker. Let me know if there is anything that needs updating here

#3835

### DIFF
--- a/src/parser/shared/modules/StatTracker.js
+++ b/src/parser/shared/modules/StatTracker.js
@@ -1,6 +1,4 @@
 import SPELLS from 'common/SPELLS/index';
-import ITEMS from 'common/ITEMS/index';
-import { calculateSecondaryStatDefault, calculatePrimaryStat } from 'common/stats';
 import { formatMilliseconds } from 'common/format';
 import SPECS from 'game/SPECS';
 import RACES from 'game/RACES';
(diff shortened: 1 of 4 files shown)


Dambroda:
I'm not sure if this is something we want to do. I'll leave it up for some other opinions. But this also broke some integration tests so that would need to be resolved beforehand.

Pewtro:
So the reason I made the issue was when I was adjusting StatTracker for the Diminishing Returns and got annoyed with the fact we were defining all those default buffs in there - that said we almost never work in that file so maybe it's fine. 

In theory I'm /care either way.

anontyro:
so I've fixed the problems the integration tests had, I think it does work to pull out some of the larger objects, I guess it also depends on how much these objects grow overtime aswell

src/parser/shared/modules/StatTracker.js
```diff
@@ -28,322 +28,14 @@ class StatTracker extends Analyzer {
...
-    [SPELLS.MIGHT_OF_THE_BLACKROCK.id]: { versatility: 411 },
-    // endregion
-  };
-
   statBuffs = {};
 
   _pullStats = {};
   _currentStats = {};
 
-  statMultiplier = {
```
emallson, line 327: `statMultiplier` is an internal object for `StatTracker`, not a buff list. It should stay in StatTracker. the other stuff seems fine.

emallson (changes requested):

src/parser/shared/modules/StatTracker.js
emallson: @anontyro just checking in. are you planning to follow up on this PR?

Pewtro:
Closing this due to inactivity.

PR #3847, opened 2020-10-09
[Core] Change Thresholds to allow use of average/major as the "primary" breakpoint for suggestions

Dambroda:
When dealing with issues where the suggestion is either on or off, and on is a major issue but off is no issue.


Pewtro:
Thanks!

PR #3852, opened 2020-10-10
[Core] Change all spells to follow a more generic type

Pewtro:
I also removed all properties that weren't of a generic type that were also unused. 

There should be no functional changes anywhere. 

The reviewers requested are because I significantly  touched files you've recently edited (You're the maintainers of the modules)


Zeboot (approved):
Looks good to me, the extracting to constants was exactly what I meant in discord the other day, looks / is much cleaner that way (rather than storing all that info on the spell itself for some reason) 👍

MartijnHols:
The nice thing of the `manaCost` is that it clearly signifies the resource type and avoids problems where the resource is used for another reason. A generic resourceCost prop has no indication, which can be especially troublesome for specs with multiple (e.g. Holy Paladin's Mana and Holy Power).

src/parser/priest/discipline/constants.ts
```diff
@@ -32,5 +32,12 @@ export const ATONEMENT_DAMAGE_SOURCES = {
 export const ATONEMENT_COEFFICIENT = 0.50;
 export const POWER_WORD_RADIANCE_COEFFICIENT = 0.625;
 
+export const PENANCE_COEFFICIENCT = 0.4;
```
rp4rk, line 38: :eyes:

Pewtro: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3852/files#diff-0f66a9298cd77a68f793d0800a2b8476R60

niseko (approved):
Resto shaman part is fine.

Pewtro:
> The nice thing of the `manaCost` is that it clearly signifies the resource type and avoids problems where the resource is used for another reason. A generic resourceCost prop has no indication, which can be especially troublesome for specs with multiple (e.g. Holy Paladin's Mana and Holy Power).

While I agree - wouldn't we then be able to add something else for those things? secondaryResource maybe?

Dambroda:
Wouldn't a simple solution to that be indicating the resource in the resourceCost; rather than being a number, make it an object? 

`resourceCost: { type: RESOURCE_TYPES.MANA, cost: 5000 }`, something along those lines?

That still feels a little weird for specs that only have one resource... but I agree that having resourceCost with no indication as to what resource you're referring to is probably not good

MartijnHols:
I reckon accidentally using resourceCost somewhere has a higher chance than accidentally using manaCost in the wrong place, and secondaryResource is not much clearer.

You could do an array for resources like the combat log, but that seems over complicated. A simple optional manaCost etc prop seems a lot easier to me.

Pewtro:
What about simply adding a ``resourceType: string`` optional prop to spells alongside resourceCost? 
That way we don't have to create objects like ``resource: { type: RESOURCE_TYPES.MANA, cost: 5000 }``. 
Unsure which is the better of those options and I don't really mind either way, but I do strive to avoid having 8 different optional props simply covering the different resource types.

Dambroda:
I don't really see the problem with the 8 optional props -- they don't really clutter anything up as you only include the ones you use 😄  

I guess another way to put it, at what point of this paring down of the SPELLS definitions do you remove resourceCost as well and move all of those constant values to 
``` const MANA_COSTS = [ { SPELLS.VIVIVY, 5000}, { SPELLS.ENVELOPING_MIST, 5000 } ... ] ```

If the argument is about clarity, I agree that it's become a bit of a slippery slope with what people are tossing into their spells defintions; I think extracting the other constants from there is 👍. But it definitely cant stay as is (im pretty sure right now SpellManaCosts module would break the spell data for all of the other types of resources?), and the other generic solutions don't really provide any benefit over the optional props (that I can see at least)

abelito75:
As long as you don't mess anything up I think its fine but if my stuff stops working I'm going to complain to you till you fix it

Pewtro:
I don't see a reason it would break anything as is?

Dambroda:
previously if you attempted to use a specific module like SpellManaCost to get your spell info you would only get mana info. If you passed in a spell/event that didn't cost mana, you got back 0. Now, the behavior is not well defined because some of them check resourceCost (previously manaCost, or whatever, which was filtered by field name) and return that if its found, overriding anything in the events classResources. for example, word of glory will return 3 (holy power) even though it doesnt cost mana, despite using `SpellManaCost.getCost`.

Pewtro:
Right but we should already be filtering by the spells we're interested for such a module, and we can also filter through the combatlog and check what resourceType was used there etc? I feel like that is something that will be avoided if a module does the checks I would think were necessary before any functional code executes?

Dambroda:
Not that I'm aware of, these modules are broad in scope and not meant to require extra logic from the module using it. In particular they tend to be used to provide the resourceCost property to events, which is done simply by adding the module as a dependency

Pewtro:
But they don't break? No integration tests failing 🤔

Dambroda:
Because it's not currently covered by integration tests; I dont know of any specs w/ multiple resources that currently have something implemented like that. I pulled your branch and created a Holy Power tracker (using the same code as SpellManaCost, just with resource type set to Holy power), and you can see it populates incorrectly:



It sets both the holy power cost and the mana cost to the same value;

The bug is currently only in SpellManaCost, but afaik there currently isnt a way to fix it with the way its done atm in this pr

abelito75:
It might not be a point to even debate anymore but `resourceCost` could be one of the worst things to do. If you have a spell that is shared between multiple specs and costs different resources for any number of specs then you couldn't generically use this field. IE vivify

emallson:
@Pewtro where is this PR going? i see some big merge conflicts now, and it sounds like the `resourceCost` change is controversial?

Pewtro:
Yeah, I'm killing it for now and I'll just hate the way we do spell typing

PR #3863, opened 2020-10-12
[Core] Convert AbilityTracker to TS

Dambroda:
@joshinat0r Can you take a peek at AoESpellEfficiency, you were the last one to modify the dependency there from Abilities to AbilityTracker, but AbilityTracker doesnt store any cooldown information from what I can see? Or is it in some JS file that isnt finding it

@niseko I added an abilitytracker to resto shaman, basically just to add typing to the field you add there

The removal of a lot of `|| 0` is because I added default values to the TrackedAbility


src/parser/shared/modules/AbilityTracker.ts
```diff
@@ -0,0 +1,130 @@
...
+    cast.casts = cast.casts + 1;
+    if (event.resourceCost && event.resourceCost[RESOURCE_TYPES.MANA.id] !== undefined) {
+      cast.manaUsed = cast.manaUsed + event.resourceCost[RESOURCE_TYPES.MANA.id];
+    }
+  }
+
+  getAbility(spellId: number, abilityInfo: Ability | null = null) {
+    let ability = this.abilities.get(spellId);
+    if (ability === undefined) {
+      ability = {
```
Dambroda, line 55: I believe its pretty safe to default all of these numerical values to zero, and it saves throughout the codebase the potential of them being undefined. Is there any cases people can think of where having one of these values set to undefined is useful to know, compared to it being zero?

EDIT: Comment below about ele shaman below could be an indicator as to keeping the default `undefined` instead of 0; easier error catching? Does that justify having to check for undefined everywhere else? Penny for your thoughts

src/parser/shared/modules/features/AoESpellEfficiency.tsx
```diff
@@ -47,12 +47,16 @@ class AoESpellEfficiency extends Analyzer {
   }
 
   get maxCasts() {
-    const cooldown = this.abilityTracker.getAbility(this.ability.id).cooldown;
+    const cooldown = this.abilities.getAbility(this.ability.id)?.cooldown ?? Infinity;
```
Dambroda, line 58: Basically causes maxCasts to return 0 if the ability info isnt found

Dambroda:
@HawkCorrigan @MikeCook9994 I had to update the ele shaman tests because the changes I made here broke your snapshot; the .damageHits being undefined left a blank spot in your statistic previously, but the default value set here changed that to 0.

Before: 
image

While updating I noticed a spell id or two was incorrect (leading to the weird `0 / _` ), so I fixed that as well. Not sure if it will even be relevant in shadowlands or not...

image

joshinat0r:
On `AoESpellEfficiency`, just guessing but I think that the [ability prop](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ebcc0f4f37e9a9912d2817c1e95e026c2fdf0a1f/src/parser/shared/modules/AbilityTracker.js#L18) in `event` contained the object from [Abilities.js](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ebcc0f4f37e9a9912d2817c1e95e026c2fdf0a1f/src/parser/deathknight/blood/modules/Abilities.js#L51) in which case I'd get the cooldown in seconds from the ability.
Not quite sure why I moved from `Abilities` to `AbilityTracker` but shouldn't really matter as long as you get the cooldown of the ability.

The general gist of the AoESpellEfficiency was to see if someone delayed a cast of an AoE spell on purpose to hit multiple targets later on. As in `pull -> 1 hit -> 1mins later -> 1 hit` if one could go `pull -> no cast -> 50sec into the pull -> 3 hits`.
But both spells that made use of this (Consumption & Dragon Roar) are dead talents in BfA, so wouldn't be surprised if the module worked in Legion and broke at some point without anyone noticing.

MikeCook9994:
@Dambroda Thanks for the fix. This will be relevant in shadowlands. I honestly never looked at why it was empty...

PR #3865, opened 2020-10-12
Holy paladin core updates for shadowlands Also updated HoW for ret

abelito75:
The majority of this is making sure nothing breaks when pre-patch drops. I might have missed a few things and should go clean some more up but this is a large step I want to get out first


src/parser/paladin/holy/modules/talents/DevotionAuraDamageReduction.js
```diff
@@ -59,7 +59,7 @@ class DevotionAuraDamageReduction extends Analyzer {
 
   constructor(...args) {
     super(...args);
-    this.active = this.selectedCombatant.hasTalent(SPELLS.DEVOTION_AURA_TALENT.id);
+    this.active = false;//this.selectedCombatant.hasTalent(SPELLS.DEVOTION_AURA_TALENT.id);
```
abelito75: Just so if anyone looks back. this is a temp turn of so we can get paladin working for tomorrow which is prepatch. There are also other issues with this now as people can swap from devo aura mid fight which could cause issues

PR #3867, opened 2020-10-13
Convert Fury warrior to TypeScript and use Statistic

DylanDirlam:
First time using TypeScript, change recommendations encouraged.

- Converted Fury Warrior talents to TypeScript
- Use Statistic instead of TalentStatisticBox


DylanDirlam:
Have updated the PR with the remaining files for Fury Warrior, this should finish up the conversion for fury to TS.

MartijnHols (approved):

PR #3875, opened 2020-10-13
[MW] Added stacked management breakdown to InvokeChiji module

vohrr:
remaking the pr to fix conflicts/update things for the conversion to  typescript


abelito75:
Not sure why this failed. Might want to commit another PR just to make it start again and fix that if it breaks again

PR #3887, opened 2020-10-14
Update spec configs and compatibilities for unsupported specs

Dambroda:


Khazakdk:
What was the criteria for removing a spec maintainer?  Might be worth writing down on this PR to point back to if anyone comes back and asks why they were removed.

Dambroda:
Mix of general activity, when the spec was last updated, and whether they have indicated they are going to be actively maintaining in SL. I figure if someone is going to be updating they can just re-add themselves when they change patchCompatibility in the coming months for SL 😄 . I was going to @ everyone who was removed, but I figured some might not appreciate it

MartijnHols (approved):
Thanks!

PR #3889, opened 2020-10-14
Update index Article to Typescript

chagriali:


Dambroda (changes requested):
I believe this will not work as-is -- ArticleLoader explicitly looks for index.js

chagriali:
Yes you are totally right, I will change all the article to Ts files, then make the necessary changes to Articleloader.

PR #3894, opened 2020-10-15
Lets see how much i break

abelito75:
<3 zero
I changed the one liner
If this fails i'll be gone for a while so if you want to take a crack at it do it.


MartijnHols (changes requested):
This will break the env vars

abelito75:
Not gonna lie here. I have no clue what im doing with lingui and just throwing things at a wall till something works... I also don't understand what you mean as I have small smooth brain... please explain like Im a 5 year old

abelito75:
Dram daddy fixed it bc bad

PR #3896, opened 2020-10-15
[Fury Warrior] Fix Warpaint not calculating

DylanDirlam:
This PR fixes Warpaint module not calculating properly.

Before: `event.absorbed` was returning undefined and calculating NaN when added together.
image

After: If `event.absorbed` is undefined, default it to 0 to prevent calculation error.
image

### DIFF
--- a/src/parser/warrior/fury/modules/talents/Warpaint.tsx
+++ b/src/parser/warrior/fury/modules/talents/Warpaint.tsx
@@ -28,12 +28,13 @@ class Warpaint extends Analyzer {
   }
 
   onPlayerDamageTaken(event: DamageEvent) {
+    const eventDamageTaken = ((event.amount || 0) + (event.absorbed || 0));
     if (this.selectedCombatant.hasBuff(SPELLS.ENRAGE.id)) {
-      const preMitigatedDamage = (event.amount + (event.absorbed || 0)) / (1 - REDUCTION_BONUS);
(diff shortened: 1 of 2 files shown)


src/parser/warrior/fury/modules/talents/Warpaint.tsx
```diff
@@ -29,11 +29,11 @@ class Warpaint extends Analyzer {
 
   onPlayerDamageTaken(event: DamageEvent) {
     if (this.selectedCombatant.hasBuff(SPELLS.ENRAGE.id)) {
-      const preMitigatedDamage = (event.amount + event.absorbed) / (1 - REDUCTION_BONUS);
+      const preMitigatedDamage = (event.amount + (event.absorbed || 0)) / (1 - REDUCTION_BONUS);
```
abelito75: iirc if the damage is completely absorbed then amount will be undefined so you should check for that... 
Might be worth it to just make a constant for ((event.amount || 0) + (event.absorbed || 0)) since you use it twice

abelito75 (changes requested):
nice

src/parser/warrior/fury/modules/talents/Warpaint.tsx
DylanDirlam: Done

Sharrq (approved):

Sharrq:
Looks like you made the change Abe requested (even though he is wrong and event.amount will be 0 if its completely absorbed). So gonna go ahead and approve it

PR #3903, opened 2020-10-16
Converted interface/common class components to functional components in typescript

giubatt:
Some more converted components to ts and functions


giubatt:
Updated the CyclingVideo snapshot, the component no longer unnecessarily passes the `randomValue` property to the html video element

src/interface/common/ActivityIndicator.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from 'react';
+
+export interface Props {
```
MartijnHols, line 3: As a general rule of thumb I never export Props as I feel `ComponentProps` is a much better way of getting them from a component as it actually touches the Component var that would later be used in the render function. You don't use the import anywhere either, so this could probably be omitted in the future.

MartijnHols (approved):
Nice, thanks!

PR #3905, opened 2020-10-16
Resto Shaman Localization

niseko:
Added localization to everything that I could find that has to do with Restoration Shaman. Also includes an attempt of fixing broken suggestions.

Also added the Language Switcher to the top of the page.
image
image

Intended to be temporary while we have no translations yet, so people who are working on translation like @AmaniZandalari can actually see / test the results.


abelito75 (approved):

abelito75:
litty ty

PR #3912, opened 2020-10-16
Update spec list style once more

Dambroda:
BEFORE

image

AFTER

image


abelito75:
I like it. I love it. I want some more of it

PR #3916, opened 2020-10-17
[Core] Convert DispelTracker to TS

Dambroda:


src/CHANGELOG.tsx
```diff
@@ -15,7 +15,7 @@ export default [
   change(date(2020, 10, 14), <>Adjusted the values of <SpellLink id={SPELLS.ARCANE_INTELLECT.id} /> and <SpellLink id={SPELLS.BATTLE_SHOUT.id} />.</>, niseko),
   change(date(2020, 10, 13), 'Internationalization for each spec', Jafowler),
   change(date(2020, 10, 13), 'Fixed typo in the Timeline buffs timelineHighlight propery', Sharrq),
-  change(date(2020, 10, 13), 'Converted Distanced Moved module to TypeScript', Dambroda),
+  change(date(2020, 10, 13), 'Converted DistancedMoved and DispelTracker modules to TypeScript', Dambroda),
```
abelito75, line 18: Are we just going to throw stuff in random old change logs?

Dambroda: If its the same type of change and not *that* old, then yes I have done it quite a few times

PR #3919, opened 2020-10-17
[Prot Paladin] Added Word of Glory analyzer

emallson:
Screenshot from 2020-10-17 15-35-57
WoG gets used a little differently by Prot than Holy, so it warrants its own analysis code.


src/parser/shared/modules/SpellUsable.test.js
```diff
@@ -231,6 +231,7 @@ describe('core/Modules/SpellUsable', () => {
         },
         targetIsFriendly: true,
         maxCharges: 1,
+        name: undefined,
```
MartijnHols, line 233: what is this?

src/parser/paladin/protection/modules/spells/WordOfGlory.tsx
```diff
@@ -0,0 +1,131 @@
...
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, { CastEvent, HealEvent, ApplyBuffEvent, RemoveBuffEvent } from 'parser/core/Events';
+import { When, ThresholdStyle } from 'parser/core/ParseResults';
+import { Options } from 'parser/core/Module';
+import HIT_TYPES from 'game/HIT_TYPES';
+
+const OVERHEAL_THRESHOLD = 0.75;
+
+export default class WordOfGlory extends Analyzer {
+  _sl = {
```
MartijnHols: In general I find it cleaner to make all these separate props. Mutating objects leads to bugs quicker, and objects are inherently more complex.

src/parser/paladin/protection/modules/spells/WordOfGlory.tsx
```diff
@@ -0,0 +1,131 @@
...
+    super(options);
+
+    this.addEventListener(Events.cast.spell(SPELLS.WORD_OF_GLORY).by(SELECTED_PLAYER), this._cast);
+    this.addEventListener(Events.cast.spell(SPELLS.SHIELD_OF_THE_RIGHTEOUS).by(SELECTED_PLAYER), this._sotrCast);
+    this.addEventListener(Events.applybuff.spell(SPELLS.SHINING_LIGHT), this._slApply);
+    this.addEventListener(Events.removebuff.spell(SPELLS.SHINING_LIGHT), this._slRemove);
+    this.addEventListener(Events.heal.spell(SPELLS.WORD_OF_GLORY).by(SELECTED_PLAYER), this._heal);
+  }
+
+  private _heal(event: HealEvent) {
```
MartijnHols: No need for _ prefix with the private keyword. That was just a thing for JS where there was no such thing

src/parser/paladin/protection/CombatLogParser.ts
```diff
@@ -34,6 +35,7 @@ class CombatLogParser extends CoreCombatLogParser {
     // Spells
     lightOfTheProtector: LightOfTheProtector,
     hotr: HammerOfTheRighteous,
+    wog: WordOfGlory,
```
MartijnHols: I think it's nicer to write it out, much easier to debug by anyone including people less familiar with the spec should you run into issues.

MartijnHols (changes requested):

src/parser/shared/modules/SpellUsable.test.js
emallson: This test was failing because the prop `name` was set to undefined by the calling code, but not present here in the expected value.

However, this test is also not run on CI (pretty sure) so it was only failing locally. Just an annoyance. Would be better to clean up the `SpellUsable` tests so they don't check literally every prop but that would require me sitting down and spending time to understand what they are actually trying to test.

src/parser/paladin/protection/modules/spells/WordOfGlory.tsx
emallson: I still find it useful as an indicator when using the method, but I'm ambivalent on it and could remove the _s.

PR #3922, opened 2020-10-17
[Windwalker] Minor SL updates

Juko8:
Some things that were changed with the spec since i did the big SL update.
Energizing Elixir GCD removed, Expel Harm being rotational and Blackout Kick CDR during Serenity


src/parser/monk/windwalker/modules/Abilities.js
```diff
@@ -165,9 +176,6 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.ENERGIZING_ELIXIR_TALENT,
         category: Abilities.SPELL_CATEGORIES.COOLDOWNS,
         cooldown: 60,
```
Dambroda, line 181:
```suggestion
        cooldown: 60,
        gcd: null,
```

Dambroda (approved):
Approved; I did add one stylistic suggestion based on the wiki recommendation to make off-gcd abilities clear by explicitly setting null

Issue #3802, opened 2020-10-04
Convert legacy event handler methods to manual event listeners


MartijnHols:
The legacy event handlers (e.g. `on_byPlayer_cast`) are deprecated but there are still a lot of older files using them. Please help us convert them to the manual event listener format (e.g. `this.addEventListener(Events.cast.by(SELECTED_PLAYER), this.onCast);`).

MartijnHols:
When this is finished we should enable the `@typescript-eslint/camelcase` eslint rule again.

PR #3923, opened 2020-10-17
Event listeners/core

Zeboot:
Start of #3802 

All other event listeners are converted already, they depend on this PR to be done first. I'll open those after core passes tests so I can merge in. Especially review spell filters and classes extending some of the changed classes.

After all the class ones are reviewed and merged in I'll add a PR that disabled legacy listener support and re-renables the camel case rule


Dambroda (approved):
I mean there's obviously no way anyone is going to look through all of these identical changes -- you're not even having to update integration tests so I'm going to approve this.

My one concern / comment is about modules that extend other modules that are already subscribed to events.  (NOTE: This comment is not specifically about this PR, moreso about a problem I foresee that I'd like to brainstorm some solutions to)

I look at something like this module that extends SpellUsable and see this:
image

Previously, it was clear that a function was an event listener by the name. Now, you know a function is an event listener because it's referenced in addEventListener inside the constructor. You cannot see the subscribing code now. The problem I foresee is developers double-listening to events because they dont realize the event was subscribed to in the parent class, and they should instead be overriding the existing listener. Maybe this is not really a problem? Thoughts?

src/parser/deathknight/blood/modules/talents/Consumption.js
```diff
@@ -3,10 +3,11 @@ import AoESpellEfficiency from 'parser/shared/modules/features/AoESpellEfficienc
 
 class Consumption extends AoESpellEfficiency {
 
+
```
Dambroda: but whyyy

src/parser/demonhunter/havoc/modules/core/Channeling.js
```diff
@@ -1,8 +1,16 @@
 import SPELLS from 'common/SPELLS';
 import CoreChanneling from 'parser/shared/modules/Channeling';
+import Events from 'parser/core/Events';
+import { SELECTED_PLAYER } from 'parser/core/Analyzer';
 
 class Channeling extends CoreChanneling {
 
+	constructor(options){
+		super(options);
```
Dambroda, line 9: whitespace in this file is all sorts of weird now

src/parser/deathknight/blood/modules/talents/Consumption.js
abelito75: Im blocking this PR as long as the extra new line is there

PR #3925, opened 2020-10-18
Event listeners/warlock

Zeboot:
Batch of changes for #3802 

once all batches are reviewed and merged, I'll disable legacy event listener support and enable the camel_case rule

Some formatting might be off due to large amount of changes. main things to look out for are inheritance and spell filters


Dambroda (approved):
Lint then LGTM

PR #3943, opened 2020-10-18
[MW] typescript and translation fixes

abelito75:
Yup


niseko (approved):
Looks good to me, just the translation changes should probably be reverted whenever we figure out what is wrong with the suggestion translations, as seeing and being able to change the position of variables is useful / sometimes needed for translations.

PR #3944, opened 2020-10-18
Hunter catchall event filter removal

Zeboot:
moves catchall listeners to new events.any

### DIFF
--- a/src/parser/hunter/marksmanship/modules/spells/AimedShot.tsx
+++ b/src/parser/hunter/marksmanship/modules/spells/AimedShot.tsx
@@ -3,7 +3,6 @@ import Analyzer, { SELECTED_PLAYER, Options } from 'parser/core/Analyzer';
 import Events, { ApplyBuffEvent, CastEvent, EventType, RefreshBuffEvent, RemoveBuffEvent, Event } from 'parser/core/Events';
 import SpellUsable from 'parser/shared/modules/SpellUsable';
 import Abilities from 'parser/core/modules/Abilities';
-import EventEmitter from 'parser/core/modules/EventEmitter';
 import DeadEye from 'parser/hunter/marksmanship/modules/talents/DeadEye';
 import { DEAD_EYE_AIMED_SHOT_RECHARGE_INCREASE, TRUESHOT_AIMED_SHOT_RECHARGE_INCREASE } from 'parser/hunter/marksmanship/constants';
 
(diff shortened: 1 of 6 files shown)


src/parser/hunter/marksmanship/modules/spells/AimedShot.tsx
```diff
@@ -37,14 +36,14 @@ class AimedShot extends Analyzer {
...
-    this.addEventListener(EventEmitter.catchAll, this.onEvent);
+    this.addEventListener(Events.any, this.onEvent);
     this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.AIMED_SHOT), this.onCast);
     this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell([SPELLS.TRUESHOT, SPELLS.DEAD_EYE_BUFF]), (event: ApplyBuffEvent) => this.onAffectingBuffChange(event));
     this.addEventListener(Events.refreshbuff.by(SELECTED_PLAYER).spell([SPELLS.TRUESHOT, SPELLS.DEAD_EYE_BUFF]), (event: RefreshBuffEvent) => this.onAffectingBuffChange(event));
     this.addEventListener(Events.removebuff.to(SELECTED_PLAYER).spell([SPELLS.TRUESHOT, SPELLS.DEAD_EYE_BUFF]), (event: RemoveBuffEvent) => this.onAffectingBuffChange(event));
   }
 
-  onEvent(event: any) {
+  onEvent(event: Event<any>) {
```
MartijnHols: Didn't @Dambroda add AnyEvent for this?

Zeboot: AnyEvent doesnt work here as the eventtype for the catchall filter is EventType.Event, which maps to its own type, so TS complains about the function having the wrong parameter. as far as I know we can't make EventType.Event map to AnyEvent within the MappedEventTypes type, as it would be self-referencing

Dambroda: If you add `[EventType.Event]: Event<EventType.Event>,` to the mapping does it work as intended?

Zeboot: nah because then you're limited to an event type that has no content (so just timestamp and type, that's it) as the parameter for your function, so you can't do other checks for example to see  what fields exist (that's basically what putro and I discussed in discord the other day)

Zeboot: okay I get what you mean now (as per discord); thought you meant something else. I'll adapt as such

Dambroda (approved):
LGTM, merge it if it passes typechecks :D

PR #3946, opened 2020-10-18
[Core] Language typescript

niseko:
Less sure about these purecomponents, but hopefully its fine.


Dambroda (approved):
Theres a double space in the changelog, for you or someone else to fix in a later PR (I call dibs)

PR #3957, opened 2020-10-19
remove legacy listener support

Zeboot:
now that all event listeners are converted to the new type, removes support of legacy listeners.
All that's missing to reenable the camelcase rule are a bunch of API objects like fight, gonna be looking at those next

### DIFF
--- a/src/parser/core/Analyzer.ts
+++ b/src/parser/core/Analyzer.ts
@@ -11,68 +11,7 @@ import { EventType, MappedEvent } from './Events';
 export { SELECTED_PLAYER, SELECTED_PLAYER_PET };
 export type Options = _Options
 
-const EVENT_LISTENER_REGEX = /on_((by|to)Player(Pet)?_)?(.+)/;
-
-/**
- * Get a list of all methods of all classes in the prototype chain until this
- * class. Orginal source: https://stackoverflow.com/a/40577337/684353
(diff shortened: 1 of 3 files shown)


src/parser/core/Analyzer.ts
```diff
@@ -11,88 +11,22 @@ import { EventType, MappedEvent } from './Events';
...
 class Analyzer extends EventSubscriber {
-  hasLegacyEventListener = false;
 
   /**
    * Called when the parser finished initializing; after all required
    * dependencies are loaded, normalizers have ran and combatants were
    * initialized. Use this method to toggle the module on/off based on having
    * items equipped, talents selected, etc.
    */
+  // eslint-disable-next-line @typescript-eslint/no-useless-constructor
```
Dambroda, line 22: Is there a reason the constructor cant be removed?

Dambroda (approved):
One question (I assume theres a reason for that), otherwise lgtm

src/parser/core/Analyzer.ts
Zeboot: it can be removed; I left it there as a "documentation" as that comment prior is probably a starting point for new contributors extending the analyzer module, and didn't wanna leave that comment floating (see other methods being defined as empty with exceptions for example)

PR #3960, opened 2020-10-19
enable camelcase lint rule

Zeboot:
Most instances are either API objects (which obviously stay that way so gained exceptions), or the fight interface which will be fixed when the files in question get converted to TS, dropping the proptypes for the actual interface


Dambroda:
offset_time is not from any API and could be renamed (EDIT: wait didnt you add that? 🤔 )

Zeboot:
yap :D left that one more cause it fits the current parameters of fight to keep it similar (and cause it already has an exception defined to it by zero) (see "or the fight interface" in the original post ;) )

Dambroda:
Im not blocking, but if anything I think it *shouldn't* match the WCL format as its not a WCL field (for the fight interface as well)

Dambroda (approved):

MartijnHols (approved):
My hero!

abelito75 (approved):
My Hero!

PR #3964, opened 2020-10-20
Convert Boring Components to TypeScript

Khazakdk:
Converted a few additional related files while I was at it.  Tagging rogue specifically for this PR because I had to modify a couple files that were using the resource lookup and want to make sure the changes are sane. @darthelit


tyndivelspaz:
Looks good. Doesn't seem like anything SHOULD be affected by the change and rogue should still work without having `undefined` returned.

tyndivelspaz (approved):

src/parser/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -21,20 +21,23 @@ class BladeRush extends Analyzer {
   }
 
   onFinishMove(event: CastEvent) {
-    if (!getResource(event.classResources, RESOURCE_TYPES.COMBO_POINTS.id)) {
+    if (!event.classResources || !getResource(event.classResources!, RESOURCE_TYPES.COMBO_POINTS.id)) {
```
MartijnHols:
```suggestion
    if (!event.classResources || !getResource(event.classResources, RESOURCE_TYPES.COMBO_POINTS.id)) {
```

The first check should already tell TS that event.classResources now no longer can be undefined.

MartijnHols (changes requested):

src/parser/rogue/outlaw/modules/talents/BladeRush.tsx
tyndivelspaz: Good catch. I need to get out if JS mode and into TS mode to catch things like this when reviewing.

src/parser/rogue/outlaw/modules/spells/BladeFlurry.ts
```diff
@@ -20,14 +20,16 @@ class BladeFlurry extends Analyzer {
   }
 
   onFinishMove(event: CastEvent) {
-    if (!getResource(event.classResources, RESOURCE_TYPES.COMBO_POINTS.id)) {
+    if (!event.classResources || !getResource(event.classResources!, RESOURCE_TYPES.COMBO_POINTS.id)) {
```
Khazakdk:
```suggestion
    if (!event.classResources || !getResource(event.classResources, RESOURCE_TYPES.COMBO_POINTS.id)) {
```

Pewtro:
Looks good to me, thanks!

PR #3966, opened 2020-10-20
[Rogue] Invigorating shadowdust

tyndivelspaz:


src/parser/rogue/shared/shadowlands/legendaries/InvigoratingShadowdust.tsx
```diff
@@ -0,0 +1,56 @@
...
+        cooldowns = [SUBTLETY_ABILITY_COOLDOWNS, SHARED_ABILITY_COOLDOWNS];
+        break;
+      default:
+        cooldowns = [SHARED_ABILITY_COOLDOWNS]
+    }
+
+    cooldowns.map((cooldown, index) => {
+      const { id } = cooldown[index];
+      return this.spellUsable.reduceCooldown(id, this.cooldownReduction, event.timestamp);
+    });
```
Pewtro, line 55: You need to verify the cooldowns are actually on cooldown or it will error out.

src/parser/rogue/shared/shadowlands/legendaries/InvigoratingShadowdust.tsx
```diff
@@ -0,0 +1,56 @@
...
+    super(options);
+    this.spec = this.selectedCombatant.spec().specName;
+    this.active = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.INVIGORATING_SHADOWDUST.bonusID);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.VANISH), this.onCast);
+  }
+
+  onCast(event: CastEvent) {
+    let cooldowns: SpellList[];
+    switch (this.spec) {
+      case 'Assassination':
```
Pewtro, line 31: I would populate the cooldowns list in the constructor so it doesn't run every single onCast event - there's no reason for the switch to run every single time either. 
also you could do: 
```js 
let cooldowns: SpellList[] = [SHARED_ABILITY_COOLDOWNS]
```

Followed by a 
```js
cooldowns.push(SPEC_ABILITY_COOLDOWNS)
```

src/parser/rogue/shared/constants.ts
```diff
@@ -1,4 +1,39 @@
...
+  ...SPELLS.KIDNEY_SHOT,
+  ...SPELLS.EVASION,
+  ...SPELLS.SHIV,
+  ...SPELLS.FEINT,
+  ...SPELLS.BLIND,
+  ...SPELLS.CLOAK_OF_SHADOWS,
+  ...SPELLS.TRICKS_OF_THE_TRADE,
+  ...SPELLS.SHROUD_OF_CONCEALMENT,
+};
 //region Legendaries
```
Pewtro, line 40: isn't it easier to simply spread the SHARED_ABILITY_COOLDOWNS etc, when you're using them in the array?

Pewtro (changes requested):
A couple things.

I think something like this could use a statistic to show some information - but if that's on a todo till you get hands on logs, then that's okay too.

src/parser/rogue/shared/constants.ts
darthelit: If I put them in an array then I'll have an array of an array and I'm not wanting that. I can possibly refactor this so that each spec has the SHARED cooldowns and then grab the proper const based on the spec being used from the module instead of using a switch-case.

tyndivelspaz: Think I came up with a better solution that what I had using only the class constants that include the shared spells in an array in constants.

tyndivelspaz:
This will definitely have a statistic on it but I'm not 100% positive what to display on it. I left it as a TODO until I can find some logs utilizing this legendary.

src/parser/rogue/shared/constants.ts
```diff
@@ -1,38 +1,42 @@
...
+];
+
+export const OUTLAW_ABILITY_COOLDOWNS: SpellList[] = [
+  SPELLS.BETWEEN_THE_EYES,
+  SPELLS.BLADE_FLURRY,
+  SPELLS.ROLL_THE_BONES,
+  SPELLS.ADRENALINE_RUSH,
+  SPELLS.GRAPPLING_HOOK,
+  SHARED_ABILITY_COOLDOWNS,
+];
```
Pewtro, line 38: Shouldn't the SHARED_ABILITY_COOLDOWNS be spread like ``...SHARED_ABILITY_COOLDOWNS`` ?

tyndivelspaz: I believe you are correct. Otherwise I'll have an embedded array.

Pewtro (approved):

Pewtro:
Thanks!

PR #3969, opened 2020-10-20
Update some news components to Typescript

chagriali:


src/interface/news/ImageArticle.tsx
```diff
@@ -1,7 +1,13 @@
+/* eslint react/prop-types: 0 */
```
MartijnHols, line 1: Doesn't this get disabled for TS automatically?

src/interface/news/ImageArticle.tsx
```diff
@@ -1,7 +1,13 @@
...
-import PropTypes from 'prop-types';
 
-const ImageArticle = props => {
+interface Props {
+  title: React.ReactNode,
+  image: React.ReactNode,
+  style: object
+}
+
+const ImageArticle: React.FC<Props> = (props) => {
```
MartijnHols:
```suggestion
const ImageArticle = (props: Props) => {
```

I prefer this style since it doesn't add an incorrect children prop and doesn't rely on type conversion; it just is a component given that Props is of the correct type.

src/interface/news/Preview.tsx
```diff
@@ -1,10 +1,15 @@
...
 
 import makeUrl from './makeUrl';
 
-const Preview = props => {
+interface Props {
+  title: React.ReactNode,
+  image: string
+}
+
+const Preview: React.FC<Props> = props => {
```
MartijnHols:
```suggestion
const Preview = (props: Props) => {
```

same

MartijnHols (commented):
Sweet, thanks! Some minor code style comments

Issue #3675, opened 2020-08-01
[Shadowlands] [Mage] Conduits checklist


Dambroda:
Other related issues:
-  #3657 Core Class Changes
- #3721 Legendaries

Conduits
-----
####  Shared Endurance
*these are not performance based and probably low priority, but still might be worth tracking the stats if we have time*
- [ ] [Cryro-Freeze](https://shadowlands.wowhead.com/spell=337123/cryo-freeze) (While inside Ice Block, you heal for X% of max HP over the duration)
- [x] [Diverted Energy](https://shadowlands.wowhead.com/spell=337136/diverted-energy) (Your Barriers heal you for X% of the damage absorbed)
- [x] [Tempest Barrier](https://shadowlands.wowhead.com/spell=337293/tempest-barrier) (Gain a shield that absorbs X% of your maximum HP for 15% after you Blink)

#### Shared Finesse
- [x] [Flow of Time](https://shadowlands.wowhead.com/spell=336636/flow-of-time) (The cooldown of Blink is reduced by X seconds) - *Needs either spellbook updated to change CD or a module to reduce it manually each cast*
- [ ] [Incantation of Swiftness](https://shadowlands.wowhead.com/spell=337275/incantation-of-swiftness) (Invisibility increases your movespeed by X% for 6 seconds) - *probably not worth tracking*
- [x] [Winter's Protection](https://shadowlands.wowhead.com/spell=336613/winters-protection) (The cooldown of Ice Block is reduced by X sec.) - *same as flow of time, needs CD updated*
- [x] [Grounding Surge](https://shadowlands.wowhead.com/spell=336777/grounding-surge) (Successfully interrupting an enemy with Counterspell reduces its cooldown by X sec) - ***need more info**, do we know if it was successful? probably not worth tracking as a statistic but at least getting the cooldown reduction right will help reduce CD module errors*

#### Arcane Throughput
- [x] [Arcane Prodigy](https://shadowlands.wowhead.com/spell=336873/arcane-prodigy) (Arcane Missiles reduces the cooldown of Arcane Power by X sec.) - *Definitely needs a module to reduce the CD. For statistics, I think it would be great to see the total CD reduction, the CD reduction wasted (by casting AM while AP is not on CD), and then maybe even seeing if the CD reduction gave you an extra cast or not*
[...]

Issue #3657, opened 2020-07-24
[Shadowlands] [Mage] Class updates for Shadowlands


Dambroda:
Other related issues:
-  #3675 Conduits and Soulbinds
- #3721 Legendaries

Logs
-----

Shared Mage Changes
-----
- [x] Mirror Image Baseline
- [x] Focus Magic Talent added to all Mage Spellbooks & Talents
- [x] Shimmer Recharge increased (25 sec)
- [x] Alter Time and Polymorph Mawrat added to all Mage Spellbooks
- [ ] Alter Time module to show health gained?
- [ ] Rune of Power drops on CD use and on cast
- [x] Mirror Image is pseudo Defensive now
- [x] Add Focus Magic module to track the buff uptime
- [ ] Kyrian Covenant Ability
- [ ] Necrolord Covenant Ability
- [ ] Night Fae Covenant Ability
- [ ] Valkyr Covenant Ability

Fire Changes
-----
- [x] Kindling reduction is 1.5 sec instead of 1 sec
- [x] Phoenix's Flames baseline
- [x] Frostbolt, Arcane Explosion added to Spellbook
- [x]  Pyroclasm damage bonus increases (240%)
- [x] Dragon's Breath CD reduced by 2 seconds (18 sec)
- [x] Remove Hyperthread Wristwraps, Lucid Dreams, and Blaster Master
- [ ] Phoenix Flames not guaranteed to crit
- [ ] Phoenix Flames on ST only during Combustion if using Pyroclasm
- [ ] Phoenix Flames 2/3+ targets unless you are pooling for Combustion 
- [x] Phoenix Flames recharge timer reduced to 25 sec
- [x] From the Ashes Talent added to Spellbook & Talents

Frost Changes
-----
- [x] Remove No Ice Lance build
- [x] Fire Blast and Arcane Explosion added to Spellbook
- [x] Remove Glacial Assault and Whiteout
- [ ] Glacial Spike can be used before or after Flurry (Preferably after). So GS > Flurry > IL > IL or FB > Flurry > GS > IL
- [x] It is no longer acceptable to hold Brain Freeze for Glacial Spike
- [x] Without Glacial Spike, Winter's Chill should be used on Ice Lance x2
- [x] Radiant Spark is acceptable to precast before Flurry (Along with Frostbolt and Ebonbolt)
- [x] Munching Brain Freeze should be avoided at all costs
[...]

PR #3970, opened 2020-10-20
[Mage] From the Ashes, Arcane Orb, & General Cleanup

### LINKED ISSUE #3948: [Arcane Mage] Arcane Orb multiple target
The tooltip of the "Use your talents effectively" category says for Arcane Orb Avg. Hits Per Cast: "Arcane Orb is primarily an AoE Spell, so you should only choose it on fights with multiple targets [...]"
Well, the whole talent row doesn't have any single target spells. Reverberate ist useless in pure single target, and Supernova is in all practical purposes less damage than Arcane Orb. So yeah, Arcane Orb *is* the singletarget choice. ^^

Sharrq:
- Added the From the Ashes Talent for Fire (Towards #3657 )
- Added Covenant Abilities to the Spellbook for all 3 specs
- Removed the GCD for Icy Veins and Arcane Power
- Adjusted Arcane Orb since it is acceptable for ST now (Fixes #3948 )
- Adjusted Conduit data for the latest Beta Build (Towards #3675 )
- Removed types from the implicitly defined variables in all specs

I also undid some of what i did with constants and have gone the route of using a constants file for the whole class instead of individual spec constants. Additionally, if a constant is only used in one module, I removed it from the centralized constants file and put it back into the relevant module.


src/parser/mage/arcane/modules/talents/ArcaneOrb.tsx
```diff
@@ -77,11 +77,11 @@ class ArcaneOrb extends Analyzer {
...
+        tooltip={`You averaged ${formatNumber(this.averageHitsPerCast)} hits per cast of Arcane Orb. ${this.badCasts > 0 ? `Additionally, you cast Arcane Orb ${this.badCasts} times without hitting anything.` : '' } Casting Arcane Orb when it will only hit one target is still beneficial and acceptable, but if you can aim it so that it hits multiple enemies then you should.`}
       >
-        <BoringSpellValueText spell={SPELLS.ARCANE_FAMILIAR_TALENT}>
+        <BoringSpellValueText spell={SPELLS.ARCANE_ORB_TALENT}>
           <>
-            {formatNumber(this.averageHitPerCast)} <small>Average hits per cast</small>
+            {formatNumber(this.averageHitsPerCast)} <small>Average hits per cast</small>
           </>
```
Pewtro, line 85: If you wanted to you could use the AverageTargetsHit helper here

src/parser/mage/fire/modules/talents/FromTheAshes.tsx
```diff
@@ -0,0 +1,58 @@
...
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(FIRE_DIRECT_DAMAGE_SPELLS), this.onCritDamage);
+  }
+
+  //Look for crit damage events to reduce the cooldown on Kindling
+  onCritDamage(event: DamageEvent) {
+    if (!this.spellUsable.isOnCooldown(SPELLS.PHOENIX_FLAMES.id) || event.hitType !== HIT_TYPES.CRIT) {
+      return;
+    }
+    const chargesOnCD = PHOENIX_FLAMES_MAX_CHARGES - this.spellUsable.chargesAvailable(SPELLS.PHOENIX_FLAMES.id);
+    this.cooldownReduction += this.spellUsable.reduceCooldown(SPELLS.PHOENIX_FLAMES, MS_REDUCTION_PER_STACK * chargesOnCD);
```
Pewtro: This should be chargesOffCD, right?

Pewtro (commented):
1 possibility for now or for future (re: avg targets)
1 clarifying variable name question

Nothing blocking.

src/parser/mage/fire/modules/talents/FromTheAshes.tsx
Sharrq: No. From the Ashes gives you CDR for every charge that is on CD. so this is the max charges (2) - the number available

Pewtro: image
I just looked it up on wowhead and it says off cooldown?

Sharrq: woops, i misread it

Pewtro: That said - is it really per off CD charge for the CDR, and not only for the mastery? Blizzard tooltips fine at work here 😄

Sharrq: actually, re-reading it again, i suppose its 1 sec CDR regardless of charges, and the mastery is based on charges

emallson:
@Sharrq looks like you handled the requested changes. we good to merge (after conflict resolution)?

Sharrq:
Yea I'm not home right now but my plan is to merge this and then if I change the average hits thing I'll do it for arcane orb and frozen orb in a separate pr

PR #3974, opened 2020-10-21
[MW] Fixed mana tea duration in T45Comparison

vohrr:
Was hardcoded to 12s in the T45 Comparison, added duration property to the spell and renamed a const to make sense


abelito75 (changes requested):
Good catch but if you're gonna fix this you might as well fix lifecycles math while you're in here as its still states (best + 875) / 2715 which isn't accurate
the best way to do this would probably be something like
`(best + (SPELLS.VIVIFY.cost * .25)) / (SPELLS.VIVIFY.cost * .25 + SPELLS.ENVELOPING_MIST.cost * .25)`
So then in the future we only ever have to update the spells file

abelito75 (approved):
nice and thanks for following up on discord!

PR #3975, opened 2020-10-21
Update RegularAcrticle to Typescript

ssabbar:


src/interface/news/RegularArticle.tsx
```diff
@@ -1,12 +1,20 @@
...
 
 import Contributor from 'interface/ContributorButton';
 import makeNewsUrl from 'interface/news/makeUrl';
 import Panel from 'interface/others/Panel';
+import { ContributorType } from 'interface/ContributorButton/ContributorButton';
 
-const RegularArticle = props => {
+interface Props {
+  title: ReactNode;
+  bodyStyle?: object;
```
guyius: This should be `React.CSSProperties`

ssabbar: I used object because the Panel component used object as a type for bodystyle prop

emallson: That is something that can be cleaned up as part of this PR. CSS props should have the correct type.

ssabbar: I've made the changes

guyius:
Can't approve, but LGTM!

ssabbar:
Thanks :)

PR #3978, opened 2020-10-21
Convert ProgressBar to TS, and small cleanup

guyius:
Refactor ProgressBar to TS, clean up some unused props, move global variables into the component.


Pewtro (approved):
Looks good to me, thanks!

PR #3982, opened 2020-10-21
[Hunter] Streamline was fixed by blizzard woo + sorting in resourceUsage

Pewtro:


Pewtro:
Thank you putro, you're amazing and awesome!

PR #3988, opened 2020-10-22
[Enhancement] Lightning/Earth Shield added to checklist 

Pewtro:
This fixes issues in this PR #3746 so it can get merged, the PR was 1000 commits behind shadowlands so it was neigh impossible to rebase back, so did it in my own PR with cherry-picking.

Now: 
image

This also fixes an issue with the performance thresholds that was introduced after we allowed only certain values to be set in the thresholds, do you agree with this solution? @Dambroda 

Before making the threshold change: 
image


Pewtro:
Lightning Shield seems to be fixed now.

PR #4000, opened 2020-10-24
Migrage PlayerSection to TS, Update dps range roles to num 3 to sort …

guyius:
Migrate PlayerSection to TS.

I'm not sure if having both melees and ranged DPS as 2 was intended or not, I've updated range to 3 so the sort function will first return melee and then range. If this is intended let me know and I'll change back.
In addition not sure if I should put the Player interface in a more general place, hard to understand if there is any pattern for interfaces, for now, I'll leave it here, and once we start using it in more places we can move it as well.

### DIFF
--- a/src/interface/report/PlayerSelection/index.tsx
+++ b/src/interface/report/PlayerSelection/index.tsx
@@ -10,9 +10,10 @@ import './PlayerSelection.scss';
 const ROLE_SORT_KEY: {[key: string]: number} = {
   [ROLES.TANK]: 0,
   [ROLES.HEALER]: 1,
+  //Different sort for range/melee was tested and felt intuitive.
+  //Because of this all DPS are treated the same for sorting purposes.
   [ROLES.DPS.MELEE]: 2,
-  //I'm not sure if this is a bug or intended behaviour, But if this is 2 then there is no sort based or range/melle only on dps.
-  [ROLES.DPS.RANGED]: 3,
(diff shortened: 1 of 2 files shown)


src/interface/report/PlayerSelection/index.tsx
```diff
@@ -1,22 +1,23 @@
...
 import './PlayerSelection.scss';
 
-const ROLE_SORT_KEY = {
+const ROLE_SORT_KEY: {[key: string]: number} = {
   [ROLES.TANK]: 0,
   [ROLES.HEALER]: 1,
   [ROLES.DPS.MELEE]: 2,
-  [ROLES.DPS.RANGED]: 2,
+  //I'm not sure if this is a bug or intended behaviour, But if this is 2 then there is no sort based or range/melle only on dps.
+  [ROLES.DPS.RANGED]: 3,
```
MartijnHols: It was intended to be 2. We experimented with grouping melee and ranged, and found it more intuitive to just group all DPS together. Maybe leave a comment to clarify that?

MartijnHols (changes requested):
Thanks, yeah you're right the sort key was unclear and should have had a comment. If you fix that this should be good to go.

src/interface/report/PlayerSelection/index.tsx
guyius: Thanks for your response, updated.

MartijnHols (approved):
Thanks!

PR #4001, opened 2020-10-24
[BrM] Track Bad Tiger Palm casts 

emallson:
Screenshot from 2020-10-24 15-50-28
Screenshot from 2020-10-24 15-50-45

I went with a lower threshold than the Prot Paladin filler spell checklist item because BrM has no procs, so it is much easier to not accidentally use the filler spell when something better is available.

This **also** includes a change to the timeline: when a tooltip is present, we no longer link to wowhead. Wowhead links have a tooltip, and the tooltip shows up on top of our tooltip with no real way to prevent overlap. As far as I can tell, there isn't a way to hide wowhead tooltips on specific links, so I just removed them (but, again, **only** from spells that have other tooltips present).


src/parser/monk/brewmaster/modules/spells/TigerPalm.tsx
```diff
@@ -0,0 +1,192 @@
...
+    this.wastedCDR += TIGER_PALM_REDUCTION - actualReduction;
+
+    if (this.bocApplyToTP) {
+      this.bocHits += 1;
+    } else {
+      this.normalHits += 1;
+    }
+  }
+
+  onDamageTaken(event: DamageEvent) {
```
joshinat0r: What is this doing?

joshinat0r (approved):
Not sure what `onDamageTaken` is supposed to do, I'll approve it anyways in case its something for an upcoming feature.

src/parser/monk/brewmaster/modules/spells/TigerPalm.tsx
emallson: Grabbing attack power. It was for face palm relics, where I used ap to determine if it was a FP proc or not, and I guess I missed removing the last vestige.

PR #4002, opened 2020-10-24
[Prot Paladin] Change SotR tracking from "good casts" to hits mitigated

### LINKED ISSUE #3873: Prot Paladin Checklist: SotR Uptime
With the change to Holy Power, SotR can reach 100% effective uptime vs physical, non-bleed hits. Add a hits mitigated metric to the checklist.

Based on feedback from HoW: 95%+ should be the target.

emallson:
image

This PR also moves the list of abilities to ignore from the Brewmaster section to the shared section.

After some discussion with the Prot TCs, the decision was made to outright remove "good cast" tracking instead of leaving both in together. The frequency with which you can press SotR now means that trying to define good casts doesn't make sense---you have to cast it a LOT to avoid capping HP.

Closes #3873


src/parser/shared/modules/hit-tracking/utilities.ts
```diff
@@ -0,0 +1,17 @@
+import Enemies from 'parser/shared/modules/Enemies';
+import { DamageEvent } from 'parser/core/Events';
+import MAGIC_SCHOOLS from 'game/MAGIC_SCHOOLS';
+
+import IgnoredAbilities from './IgnoredAbilities';
+
+export function notableEnemy(enemies: Enemies, event: DamageEvent): boolean {
+    return !event.sourceIsFriendly || event.sourceID === undefined || !enemies.getEntities()[event.sourceID];
```
Dambroda: logic here seems weird; returns true (is notable) if the sourceID is undefined or the enemy entity doesnt exist?

emallson: yup, i flipped the logic in it on accident

Dambroda (approved):

PR #4003, opened 2020-10-25
[WW] Add Expel Harm to checklist and minor fixes in Emperor's Capacitor

Juko8:


Dambroda (approved):

Dambroda:
Are the i18 reverts just to prevent the suggestions from breaking on prod? I think it will just be caught in other PRs and pushed to prod anyway...

Juko8:
> 
> 
> Are the i18 reverts just to prevent the suggestions from breaking on prod? I think it will just be caught in other PRs and pushed to prod anyway...

That's because it started updating the localization files, which i figured would happen on production anyway and i didn't want to bother handling the merge errors.

PR #4004, opened 2020-10-25
Update several files to TS, Type heart of azeroth

guyius:
Move PlayerTile and characters to TS, add a new heart of Azeroth type file. Also refactored `PlayerTile` to react hooks.

My work on migrating to TS raises a few questions for me, Is there any pattern for typings? I don't see any dedicated type files, for small things it's usually fine to have things in the file consuming it, but long term we will have a lot of shared types. Was there any plan or talk about this?
Also when creating new types (for example here with HOA), I just looked into a few examples of the object on the chrome console and copied that into a type. Is there a safer way?

Regarding refactor to hooks, I didn't touch the redux store for now as it's probably better if someone with more knowledge of the project will add a more generic way to use `useStore` and `useContext` across the app.


src/common/contributor.ts
```diff
@@ -1,22 +1,27 @@
 import { Spec } from 'game/SPECS';
 import * as contributors from 'CONTRIBUTORS';
 
+import { HeartOfAzeroth } from './heartOfAzeroth';
+
 export type Character = {
```
Dambroda, line 4: This is not the `character` you want to convert for use in the store or the api response. This is just the definition for contribs to declare their character links. You'll want to create a new type somewhere else. Changing the commas to semi's should stay though

src/interface/actions/characters.ts
```diff
@@ -1,17 +1,17 @@
 import { makeCharacterApiUrl } from 'common/makeApiUrl';
+import { Character } from "common/contributor";
```
Dambroda, line 2: Make sure you update this import when making the new type

Dambroda (changes requested):

src/common/contributor.ts
guyius: Done

Dambroda (approved):

PR #4006, opened 2020-10-25
Fix spec check

tyndivelspaz:


src/CHANGELOG.tsx
```diff
@@ -10,6 +10,7 @@ import Contributor from 'interface/ContributorButton';
 
 // prettier-ignore
 export default [
+  change(date(2020, 10, 25), 'Fixed Invigorating Shadowdust module', Tyndi),
```
abelito75: this shouldn't go in the master change log but rather a rouge change log

abelito75 (changes requested):

PR #4007, opened 2020-10-25
[Frost Death Knight] RP tracking for Breath of Sindragosa and suggestion tweaks

Khazakdk:
The logs don't contain any information on RP spent so I had to do it all manually.  I'm not entirely sure what I did was sane.  At a high level its simple, every damage event per unique timestamp means 16 (or less with hypothermic presence) rp was spent, so add the rp spent to a list.  When the buff falls off or the fight ends with the buff up, sum the list and fabricate a cast event using the damage ability so resourceTracker picks it up.  I use the damage ability so as to not throw off the overall casts for breath.    

I'm mostly worried about the event fabrication. Its possible there's a more elegant way to do this that I don't know about.  I initially tried fabricating a cast event on every damage tick but that led to the resource tracker counting every tick as a cast and I need it to just show 1 cast event.  

image
image


src/parser/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
```diff
@@ -0,0 +1,175 @@
...
+const HYPOTHERMIC_PRESENCE_COST_REDUCTION = .35;
+
+class BreathOfSindragosa extends Analyzer {
+  static dependencies = {
+    eventEmitter: EventEmitter,
+  }
+
+  protected eventEmitter!: EventEmitter;
+
+  damageAbility!: Ability;
```
Dambroda: This `!` (and the one for triggeringCast) should, for more 'correct' type safety, be removed from here and moved to a place where you know it will actually be set (specifically it looks like in the fabricate function)

src/parser/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
```diff
@@ -0,0 +1,175 @@
...
+  onDamage(event: DamageEvent) {
+    if (event.timestamp === this.mostRecentTickTime) {
+      return;
+    }
+
+    if (!this.damageAbility) {
+      this.damageAbility = event.ability;
+    }
+
+    this.tickAmounts.push(this.selectedCombatant.hasBuff(SPELLS.HYPOTHERMIC_PRESENCE_TALENT.id, event.timestamp) ? BREATH_COST_PER_TICK * (1 - HYPOTHERMIC_PRESENCE_COST_REDUCTION) : BREATH_COST_PER_TICK);
```
Dambroda: Not sure if this needs to be an array, couldnt it just be a running total if all you're doing is `.reduce` at the end?

Dambroda (approved):
This looks good to me. The fabrication of the cast event is a bit hacky (and is actually at risk for self-referencing, luckily the cast and damage events for breath are different ids :P ); but if it works as intended its fine IMO. The "elegant" way to do it would be to modify frost's RunicPowerTracker to edit the tracked spender values using the logic you have here, rather than fabricating a new cast event that doesn't ever technically happen. But I think you have things pretty well contained here if you dont want to make that change.

A few other tiny comments that arent blocking.

Khazakdk:
> This looks good to me. The fabrication of the cast event is a bit hacky (and is actually at risk for self-referencing, luckily the cast and damage events for breath are different ids :P ); but if it works as intended its fine IMO. The "elegant" way to do it would be to modify frost's RunicPowerTracker to edit the tracked spender values using the logic you have here, rather than fabricating a new cast event that doesn't ever technically happen. But I think you have things pretty well contained here if you dont want to make that change.

Unfortunately the only events I get from breath once it starts are damage events.  I did start out attempting to put a listener in Frost's RP tracker and have it update the spender object but I got a point where I was essentially copy pasting large pieces of ResourceTracker into the child class with slight tweaks.  That seemed dangerous as it meant I would have to keep a close eye on ResourceTracker and manually move code updates over, so the lesser evil seemed to be to hack together a fake cast and hand it off to let ResourceTracker handle it.  This whole thing does hinge on the damage event and cast event having different Ids which isn't great but having to manually track runes and runic power because that info isnt in logs isn't great either.  I'll think it over and maybe see if I can shift it over to RunicPowerTracker.

src/parser/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
Khazakdk: Good call, still getting used to TS

src/parser/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
Khazakdk: It may or may not be, my next task is to add a module for Hypothermic Presence, which grants a buff that reduces the cost of your RP spenders.  Because RP spent on BoS isn't logged I'm going to need to do something to pull out specifically which ticks of Breath are at a reduced cost.  I won't be surprised if I end up ripping out half of this when I start on that task.  Maybe I fabricate resourcespent events...

Dambroda:
> > This looks good to me. The fabrication of the cast event is a bit hacky (and is actually at risk for self-referencing, luckily the cast and damage events for breath are different ids :P ); but if it works as intended its fine IMO. The "elegant" way to do it would be to modify frost's RunicPowerTracker to edit the tracked spender values using the logic you have here, rather than fabricating a new cast event that doesn't ever technically happen. But I think you have things pretty well contained here if you dont want to make that change.
> 
> Unfortunately the only events I get from breath once it starts are damage events. I did start out attempting to put a listener in Frost's RP tracker and have it update the spender object but I got a point where I was essentially copy pasting large pieces of ResourceTracker into the child class with slight tweaks. That seemed dangerous as it meant I would have to keep a close eye on ResourceTracker and manually move code updates over, so the lesser evil seemed to be to hack together a fake cast and hand it off to let ResourceTracker handle it. This whole thing does hinge on the damage event and cast event having different Ids which isn't great but having to manually track runes and runic power because that info isnt in logs isn't great either. I'll think it over and maybe see if I can shift it over to RunicPowerTracker.

You can add a damage listener to the tracker which is what I meant -- same sort of logic as here (damage events per tick, etc) just adapted to work properly by actually editing the costs value of the breath cast

Khazakdk:
Now that ResourceTracker is in TypeScript I was able to apply Dambroda's suggestion and moved the manual runic power tracking to RunicPowerTracker

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -180,7 +180,7 @@ class ResourceTracker extends Analyzer {
     return this.getResource(event)?.cost;
   }
 
-  getResource(event: CastEvent | HealEvent | EnergizeEvent ) {
+  getResource(event: CastEvent | HealEvent | EnergizeEvent | DamageEvent) {
```
Dambroda: Was this extra type actually used in the end anywhere? I thought they were both CastEvents that had the actual classResource props

Dambroda (commented):
One last question, then LGTM!

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
Khazakdk: Oh good catch, I did at one point need to call getResource on my damage event but took it out a few iterations later and forgot to change that back

Dambroda (approved):

PR #4019, opened 2020-10-26
[MW] fix for HotTrackerMW not detecting talent choices that affect hot duration

vohrr:


abelito75:
Rejecting since I got a cleaner fix.

PR #4022, opened 2020-10-26
Update several components to TS, small update to player title text

guyius:
Several components migrated to ts, update player title text in `PlayerGearHeader`

Before: 
Screenshot 2020-10-26 at 22 58 38

After:
Screenshot 2020-10-26 at 22 58 44


src/common/ItemIcon.tsx
```diff
@@ -7,16 +7,15 @@ import Icon from './Icon';
 type Props = {
   id: number;
   noLink?: boolean;
-  details?: Record<string, any>;
```
joshinat0r, line 10: It looks like the `details`-prop is used in [ItemLink](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/271b3d751735258c15dce435877cf4c2f1dda6ce/src/common/ItemLink.js#L39) to override the quality of an item (because the fallback is using the quality from `ITEMS`)

Some Nyalotha items use it without prividing `quality`, so we *should* keep it, although `ItemLink` is still using iLvl 370 to determine if blue or purple (so it isn't reliable right now anyways).

src/interface/report/Results/Character/PlayerInfo/PlayerGearHeader.tsx
```diff
@@ -1,10 +1,15 @@
...
-const PlayerGearHeader = ({ player, averageIlvl }) => (
+interface Props {
+  player: Combatant;
+  averageIlvl: number;
+}
+
+const PlayerGearHeader = ({ player, averageIlvl }: Props) => (
   <div className="player-gear-header">
     <div className={`${player.spec.className.replace(' ', '')} player-name`}>
       <Link to={makeCharacterUrl(player)}>{player.name}{player.characterProfile && <> - {player.characterProfile.realm}</>}</Link>
```
joshinat0r: Centering this already made it prettier, but I think moving the complete realm name to the second line would look a bit cleaner compared to `Synida - Laughing \n Skull`.

Talking about something like this `{player.characterProfile && <><br />{player.characterProfile.realm}</>}`, or maybe check the length of character & realm to see if it fits in a single line or should break between char and realm (since `.player-gear-header` is always roughly the same width)

joshinat0r (changes requested):
Looks good, just two minor things

src/common/ItemIcon.tsx
joshinat0r: Actually, no `<ItemIcon />` isn't using `details` right now, but I think we should keep it if someone needs to overwrite it in the future.

src/interface/report/Results/Character/PlayerInfo/PlayerGearHeader.tsx
guyius: Great suggestion, also decided to remove the hyphen, this is the new look:
Screenshot 2020-10-29 at 23 41 22

joshinat0r (approved):

PR #4023, opened 2020-10-27
[Death Knight] Enchants and some minor fix

joshinat0r:
Got a bit out of hand, only wanted to start with some enchant stuff:

- fixed the encounter statistic page in the character pane (WCL omits combatantInfos by default for some time now, not sure when they changed it but that meant that the necessary fields were missing)
- fixed some Unholy crashes due to spellIds in the talents ( @Khazakdk spell id for soulreaper changed & unholy pact were missing, should only fix crashes)
- added all DK enchants (which are unlike other enchants spells, not items)
- fixed some spellbook-errors (missing abilities / AMZ GCD fix)
- Relish in Blood fix (had a case where the wasted RP was negative, is now using the waste-prop of the energize event)
- Relish in Blood suggestion based on wasted RP percentage
- Rune of the Fallen Crusader statistic for all DK specs (average strength gain + heal/overheal) [FC example log](https://www.warcraftlogs.com/reports/TaY8KQ6rJ9qnxHP7#fight=13&type=summary&source=2)
- Rune of Hysteria for all DK specs (RP gained/wasted) [Hysteria example log](https://www.warcraftlogs.com/reports/PAWXRY4Qcq7CtLpr#fight=9&type=summary&source=134)

FC
image

Hysteria
image
image


Khazakdk (approved):

Khazakdk:
Looks good, only nonblocking comment is it seems weird to put runeforges inside the DK folder and merge with the spells in COMMON.  Not sure if any of the alternatives are better though, just an unfortunate reality of runeforges being special snowflake enchantments.

joshinat0r:
Moved the enchants into the deathknight-spell file, that takes care of the weird import from `parser` to `common`.
The only really clean solution would be to rework enchants like reglitch mentioned, not worth imo just because DKs are wierd.
Will merge this version.

Issue #3877, opened 2020-10-13
Prot Paladin: Update Talents


emallson:
Prot actually got talent updates (jealous) in Shadowlands. Need to go through and add support for the relevant ones.

abelito75:
You might want to look at shared as it does handle some of the generic talnets :D 
Edit: src/parser/paladin/shared

hpabst:
# Shadowlands Prot Paladin Talents:
## SL Talent List
### Row 1
- Holy Shield (Unsupported)
- Redoubt (Unsupported)
- Blessed Hammer (Unsupported)

### Row 2
- First Avenger (Unsupported)
- Crusader's Judgement (Unsupported)
- Moment of Glory (Unsupported)

### Row 3
- Utility talents, likely no need to support.

### Row 4
- Utility talents, likely no need to support.

### Row 5
- Divine Purpose (updated by Abe)
- Holy Avenger (updated by Abe)
- Seraphim (needs to be updated for SL)

### Row 6
- Utility talents, likely no need to support.

### Row 7
- Sanctified Wrath (Unsupported)
- Righteous Protector (Unsupported)
- Final Stand (AoE taunt, likely no need to support)

## Priority Support
Wowhead recommends the following raiding talents at the moment: Holy Shield, Crusader's Judgement, Divine Purpose, Sanctified Wrath. I'll start working on these going backwards from row 7.

abelito75:
I haven't checked but you might want to implement the HoW module I made so you can have accurate cast efficiency suggestions for the spell

hpabst:
Addendum: Crusader's Judgment is somewhat supported through the GrandCrusader and SpellUsable modules for prot paladin. We could leave as is, modify to make it more explicit that Judgment charges were added, or refactor to make GrandCrusader proc tracking shared between it's own analyzer and a separate Crusader's Judgment analyzer (which could track something like wasted Grand Crusader procs because Judgment was sitting at 2 charges already).

Thoughts? Regardless I'll work on converting the existing GrandCrusader work to TS.

emallson:
I say leave it as-is. Unfortunately, SpellUsable is required to handle it
so we can never get it into its own module. At the same time, the analysis
makes the most sense as part of GrandCrusader (since the reset is tied to
that) and checklist/suggestions already exist for munching too many procs
by way of cast efficiency.

abelito75:
[...]

PR #4024, opened 2020-10-27
[Prot Paladin] Update Seraphim and Righteous Protector for SL

### LINKED ISSUE #4013: Prot Paladin: Righteous Protector still shows LotP CD reduction in the tooltip
unknown-4

hpabst:
# Notes
- Updates the existing Seraphim and Righteous Protector analyzers for the changes these two talents went through in SL.
- Remove Seraphim buff from StatTracker as it is now a percentage-based buff rather than a flat stat increase.
- [Log used in testing](https://www.warcraftlogs.com/reports/xXW9f4KnmdF2MYV8#fight=41&type=damage-done&source=2)
- Resolves #4013 
- Towards #3877 

# Testing
- `yarn lint`, `yarn build`, `yarn test:integration`

# Screenshots
2020-10-27 09_46_32-Mythic Wrathion - Kill (4_22) by Decix in Ny'alotha - WoWAnalyzer


src/parser/paladin/protection/modules/talents/Seraphim.tsx
```diff
@@ -0,0 +1,99 @@
...
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(SPELLS.SERAPHIM_TALENT.id);
+    if (!this.active) {
+      return;
+    }
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SERAPHIM_TALENT), this.onCast);
+    const localStatTracker = (options.statTracker as StatTracker);
+    const bonusStats = this.getBonusStats(localStatTracker);
+    localStatTracker.add(SPELLS.SERAPHIM_TALENT.id, bonusStats);
```
emallson: sera also multiplies stats gained from other buffs (e.g. trinkets), which means that this value is going to be an underestimate.

importantly: this also means that haste will be wrong when sera + another haste buff is active

src/parser/paladin/protection/modules/talents/Seraphim.tsx
```diff
@@ -0,0 +1,99 @@
...
+  get averageVersatilityGain(): number {
+    return this.bonusStatGains.map((gain) => gain.versatility).reduce((sum, next) => sum + next, 0) / this.bonusStatGains.length;
+  }
+
+  get averageHasteGain(): number {
+    return this.bonusStatGains.map((gain) => gain.haste).reduce((sum, next) => sum + next, 0) / this.bonusStatGains.length;
+  }
+
+  onCast(event: CastEvent): void {
+    this.bonusStatGains.push(this.getBonusStats(this.statTracker));
```
emallson, line 66: what about stat DRs?

(i don't know what code we have in place to handle those at this time---boy does it complicate things tho)

emallson (changes requested):
instead of adding the bonus stats to the stat tracker on init, you should add a function (iirc) that when called will produce the stats based on current stats. will need to double-check the exact forms that you can add, but I know that a functional form is supported.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: The stat DRs only apply to the amount of rating required per percentage increase, yeah? Currently this records the raw rating gain rather than percentage gain, so we shouldn't need to compensate for stat DRs.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: Do we have support for calculating the overall stat gain with other buffs taken into account through `StatTracker`? A quick look through the code indicates that `StatTracker` just applies stat rating changes and stat rating multipliers iteratively as buffs are applied and fall off which indicates that buffs are treated as snapshots rather than dynamic values.

emallson: So i did some testing on Seraphim on live (with no trinkets / azerite /essences so no procs). Here's the before and after:

## Before
Capture
## After

image

Notably, these are all exactly 8% stat gain---not 8% rating gain---with the exception of Haste, which uses the 8% rating + 8% stat gain rule (see [here](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/src/parser/shared/modules/Haste.js#L223-L225)). % Stat gain ignores DRs, which means that this will always be 8% (except haste, because reasons).

That means that we actually need some new machinery in `StatTracker` to handle Vers/Crit/Mastery. However, the `Haste` module already works. Since we *mostly* need that to be correct, we could merge this with Sera haste added to the Haste module and the others just showing on the tooltip as `uptime * 8%`.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
emallson: My thinking had been that we don't want to show raw rating gains without context for how effective those ratings actually were, but per my comment above i think that is a moot point for Sera.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: Regarding the changes needed for `StatTracker` to handle the non-haste stats: It looks like StatTracker currently supports tracking percentage-based buffs, it just lacks a mechanism for adding new buffs to the list of currently tracked buffs. It should be fairly simple for me to add what's needed in this CR.

Otherwise, I've created a prot `Haste` module following the line of the ret `Haste ` module and registered the Seraphim talent with that.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: Updated Seraphim with displaying the percentage gain rather than raw rating gain:

2020-10-29 11_20_29-Mythic Wrathion - Kill (4_22) by Decix in Ny'alotha - WoWAnalyzer

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: Resolved in latest commit.

src/parser/shared/modules/StatTracker.js
```diff
@@ -455,12 +455,29 @@ class StatTracker extends Analyzer {
...
+    if (!buffId || !stats) {
+      throw new Error(`StatTracker.addStatMultiplierBuff() called with invalid buffId ${buffId} or stats`);
+    }
+    if (typeof buffId === 'object') {
+      buffId = buffId.id;
+    }
+    if (this.statMultiplierBuffs[buffId]) {
+      throw new Error(`Stat buff with ID ${buffId} already exists`);
+    }
+    this.statMultiplierBuffs[buffId] = stats;
```
emallson: This is incorrect. Seraphim is not a stat multiplier. For example: Battle Shout increases your agility/strength by 5%---by multiplying by 1.05. 50 mainstat becomes 52.5. Seraphim increases your secondaries by *adding* 8% to each: 50% becomes 58%, not 54%.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
```diff
@@ -37,47 +39,61 @@ class Seraphim extends Analyzer {
...
 
   get averageHasteGain(): number {
-    return this.bonusStatGains.map((gain) => gain.haste).reduce((sum, next) => sum + next, 0) / this.bonusStatGains.length;
+    return this.bonusStatGains.map((gain: BonusStats) => gain.haste).reduce((sum, next) => sum + next, 0) / this.bonusStatGains.length * this.uptime;
   }
 
-  onCast(event: CastEvent): void {
+  get statMultipliers() {
+    return {
+      haste: 1.0 + STAT_MODIFIER,
```
emallson: Omit haste from the interactions with `StatTracker` entirely, since it uses Bloodlust-style addition which is covered by the `Haste` module. The `Haste` module listens for stat changes, so it will double-count haste if it shows up in both modules.

emallson (changes requested):

src/parser/shared/modules/StatTracker.js
emallson: However, once again I'll say: we don't need to support the non-Haste buffs right this second. We can revisit them down the line if you don't want to add support for this kind of buff right now. The haste portion is the most important one and is already covered by the `Haste` module.

hpabst: That makes sense and clears up some of the confusion I was having. I'll adjust the analyzer to support only the haste buff for the time being.

src/parser/paladin/protection/modules/talents/Seraphim.tsx
hpabst: I've removed haste from the stattracker in the latest commit.

src/parser/shared/modules/StatTracker.js
hpabst: For the time being I'm leaving off support for tracking the changes to non-haste secondaries from Seraphim.

emallson (approved):

PR #4031, opened 2020-10-29
Update game directory files to typescript

jos3p:
Updated the src/game subdirectory to TS


src/game/DIFFICULTIES.ts
```diff
@@ -11,7 +12,7 @@ const DIFFICULTIES = {
 };
 export default DIFFICULTIES;
 
-export function getLabel(difficulty) {
+export function getLabel(difficulty: number | undefined) {
```
guyius: It's prefered to use this syntax `difficulty?: number` this way users of the function don't have to pass undefined but can just ignore the param.

src/game/FACTIONS.ts
```diff
@@ -0,0 +1,8 @@
+const FACTIONS: {
+  [key: string]: string;
```
guyius: As this is a list that will never change we can type it with the values and users of the type will get error suggestions on things like lowercase vs uppercase, or wrong options.
`[key: string]: 'ALLIANCE' | 'HORDE';`

src/game/getAverageItemLevel.ts
```diff
@@ -0,0 +1,9 @@
+import GEAR_SLOTS from './GEAR_SLOTS';
+
+// Some Tabards have a higher ilvl than 1 so exclude them from the avg ilvl (not sure about shirts, but including them)
+const EXCLUDED_ITEM_SLOTS = [GEAR_SLOTS.SHIRT, GEAR_SLOTS.TABARD];
+
+export default function getAverageItemLevel(gear: any[]) {
```
guyius: gear can be typed as an `Item[]` coming from here => `import { Item } from 'parser/core/Events';`

jos3p:
Thank you for the review, I have added the suggested changes and also edited `getAverageItemLevel.ts` to use `item: Item`.

guyius:
Nice job, LGTM, but I don't have permission to merge.

joshinat0r:
But I have, looks good. Will merge this once the CI is done 👍

PR #4033, opened 2020-10-30
[Core] ResourceTracker to TypeScript

Khazakdk:
Tagging @Pewtro since I had to change a couple hunter files, the rest of this was straightforward TS conversion


src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -104,13 +127,18 @@ class ResourceTracker extends Analyzer {
...
         return;
     }
     const eventResource = this.getResource(event);
+
+    if(!eventResource){
+      return;
+    }
+
     if (eventResource && eventResource.max) {
       this.maxResource = eventResource.max; // track changes in max resource, which can happen due to procs / casts
```
Pewtro, line 143: no reason to do ``if(eventResource)`` when you've just checked above

Pewtro (approved):
The hunter things seem fine. 

Are you sure you want to do the divide by 10 within ``getReducedCost`` - shouldn't that be reserved for anything that *actually* reduces the cost, and not just weird * 10 / 10 back and forth that blizzard does? Non-blocking, merely design question. 

One other note on an if check also non-blocking.

src/parser/deathknight/frost/modules/runicpower/RunicPowerTracker.ts
```diff
@@ -1,18 +1,19 @@
...
 
-  getReducedCost(event: any) {
-    if (!this.getResource(event).cost) {
-      return 0;
+  getReducedCost(event: CastEvent) {
+    const cost = this.getResource(event)?.cost;
+    if (cost) {
+      return cost / 10;
     }
-    return this.getResource(event).cost / 10;
```
Dambroda, line 15: Missing a return here?

src/parser/deathknight/unholy/modules/runicpower/RunicPowerTracker.ts
```diff
@@ -11,10 +11,10 @@ class RunicPowerTracker extends ResourceTracker {
   }
 
   getReducedCost(event: CastEvent) {
-    if (!this.getResource(event).cost) {
-      return 0;
+    const cost = this.getResource(event)?.cost;
+    if (cost) {
+      return cost / 10;
     }
-    return this.getResource(event).cost / 10;
```
Dambroda, line 17: And here

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -13,20 +34,22 @@ class ResourceTracker extends Analyzer {
...
 
   // stores resource gained/spent/wasted by ability ID
-  buildersObj = {};
-  spendersObj = {};
+  buildersObj: {[index: number]: BuilderObj} = {};
+  spendersObj: {[index: number]: SpenderObj} = {};
 
   // TODO set this to the resource you wish to track constructor.. see the appropriate objects in game/RESOURCE_TYPES
-  resource;
+  resource!: Resource;
```
Dambroda, line 47: you could make the class + resource abstract so that extending classes will get an error if they dont set it, but this is fine for now

src/parser/deathknight/frost/modules/runicpower/RunicPowerTracker.ts
Khazakdk: Yes and no.  It does look a little weird having a potential path of no return but any cast event with runic power as a resource will always have a cost.  If `this.getResource(event)?.cost` is null or undefined then there's nothing to return anyways

Khazakdk:
Regarding the divide by 10, all 3 DK specs do this so we don't have manually divide by 10 on every display of Runic Power.  Since Blizz multiplies all RP values by 10 in the logs, it makes things easier for us to just undo that multiply by 10 here so it displays properly in every bar and graph

PR #4034, opened 2020-10-30
3801 - updated the deprecated statisticbox elements of havoc DH

leozhekov:
Here's a comparison:
image

For Netherwalk, I have decided to not render the table if there are no spells that were immuned.


leozhekov:
Fork, sorry. I forgot that this thing hasn't been merged yet and my push would trigger this PR... Sorry for having many files here!

leozhekov:
Anyways.. 
That's before:

image

And that's after:
image

leozhekov:
I know this module has not been updated in 2 years, but should anyone that wants to maintain it, it would be without deprecated elements and stuff..

src/common/SPELLS/demonhunter.ts
```diff
@@ -154,21 +154,16 @@ const spells: SpellList<DemonHunterSpell> = {
     painCost: 30,
   },
   IMMOLATION_AURA: {
-    id: 178740,
-    firstStrikeSpellId: 187727,
+    id: 258920,
+    firstStrikeSpellId: 187727, // fix it
```
joshinat0r: What is this used for? `firstStrikeSpellId` is used twice in our codebase, once here and a second time in the `DemonHunterSpell` definition.
It looks like [this commit](https://github.com/WoWAnalyzer/WoWAnalyzer/commit/ff740f9c0fb42568e8c6bda5e40e9176a235c823) added them, but it doesn't look like we need anymore

joshinat0r (changes requested):
Just this one comment, everything else looks good.

For the future, you can revert commits in git, git will update the PR automatically.
If you want to continue working on something while a review is already pending just create a new branch from your PR-branch and leave a note in the second PR (along the lines `depends on #4034`).
Larger PRs are work and no one likes work 😄

src/common/SPELLS/demonhunter.ts
leozhekov: I have no idea. I just added the comment with this because, well.. the game shows a completely different ID to the old one.. All I wanted was to update the statistics and Immolation aura wasn't working so I fixed the IDs.

joshinat0r: I'm just saying that we can remove `firstStrikeSpellId` here and from `DemonHunterSpell` altogether since its unused.
There is a merge conflict at this line anyways, so while you're at it please remove this remnant completely.

joshinat0r (approved):

PR #4036, opened 2020-10-31
i18n of PremiumLoginPanel and PremiumPage

jos3p:
Added i18n support for the Premium sing-in page (`wowanalyzer.com/premium`)


src/interface/PremiumLoginPanel/PremiumLoginPanel.js
```diff
@@ -74,21 +75,25 @@ export class PremiumLoginPanel extends React.PureComponent {
           <div className="text">
             <h1>
               <Textfit mode="single">
-                Hello {user.name}.
+                <Trans>Hello</Trans> {user.name}.
```
niseko: I think this should encapsulate the whole sentence so you can change the order in some languages, needs an ID as well

niseko (changes requested):
Thanks, just one thing

src/interface/PremiumLoginPanel/PremiumLoginPanel.js
jos3p: Ah that makes sense, thank you

Dambroda (approved):

PR #4038, opened 2020-10-31
Finish migrating PlayerInfo to TS and react hooks.

guyius:
Finish migrating PlayerInfo to TS and react hooks.


joshinat0r:
Updating Azerite and Essences now had to feel like a waste of time, but looks good 👍

PR #4039, opened 2020-10-31
[Elemental] Add Static Discharge Module

HawkCorrigan:


src/parser/shaman/elemental/modules/talents/StaticDischarge.tsx
```diff
@@ -0,0 +1,74 @@
...
+        major: 0.85,
+      },
+      style: ThresholdStyle.PERCENTAGE,
+    };
+  }
+
+  suggestions(when: When) {
+    when(this.suggestionThresholds)
+      .addSuggestion((suggest, actual, recommended) =>
+        suggest(<span>Maximize your damage with Static Discharge by only using it with Flame Shock active.</span>)
```
MartijnHols: Please see https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Suggestions for how to formulate suggestions.

MartijnHols (changes requested):

src/parser/shaman/elemental/modules/talents/StaticDischarge.tsx
```diff
@@ -0,0 +1,81 @@
...
+      },
+      style: ThresholdStyle.PERCENTAGE,
+    };
+  }
+
+  suggestions(when: When) {
+    when(this.suggestionThresholds)
+      .addSuggestion((suggest, actual, recommended) =>
+        suggest(
+          <span>Maximize your damage with Static Discharge by only using it with Flame Shock active on an enemy in range.
```
MartijnHols: This is still missing the "Explain what was found" aspect. How about

`You missed ${formatPercentage(1 - actual)}% of the ticks of your <SpellLink id={SPELLS.STATIC_DISCHARGE.id} />. Try to maximize the ticks by only using it while Flame Shock is active on an enemy in range.`

Does it not hit anyone if Flame Shock is not active? [The tooltip](https://www.wowhead.com/spell=342243/static-discharge) suggests Flame Shock only indicates a preferred target, while the ticks will land on someone so long as they're in range. If that is the case I would make it:

`You missed ${formatPercentage(1 - actual)}% of the ticks of your <SpellLink id={SPELLS.STATIC_DISCHARGE.id} />. Try to land all ticks by only using it while you are sure there will be at least 1 enemy in range for the duration.`

HawkCorrigan: Ye the tooltip is wrong.
I updated the suggestion.

Pewtro (approved):
Looks good to me with the wording now

Pewtro:
Thanks!

PR #4041, opened 2020-10-31
[Elemental] Echoing shock

HawkCorrigan:
image

### DIFF
--- a/src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
+++ b/src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
@@ -29,10 +29,8 @@ class EchoingShock extends Analyzer {
     this.casts += 1;
 
     if(event.ability.guid !== SPELLS.LAVA_BURST.id){
-      return;
+      this.badCasts += 1;
     }
-
-    this.badCasts += 1;
   }
 
   get efficiency() {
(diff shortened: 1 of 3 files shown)


src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+
+  onCast(event: CastEvent) {
+    if(!this.selectedCombatant.hasBuff(SPELLS.ECHOING_SHOCK_TALENT.id)){
+      return;
+    }
+
+    this.casts += 1;
+
+    if(event.ability.guid !== SPELLS.LAVA_BURST.id){
```
joshinat0r, line 31: I'm a bit confused, this returns when the used spell was not Lava Burst, that means only Lava Bursts are counted towards `badCasts`.
The suggestions tells me to maximize Echoing Shock by using it **with** Lava Burst.

When I look at [this](https://www.warcraftlogs.com/reports/LVt29NWTBPjkYq7F#fight=53&type=casts&source=17&translate=true&view=events) log (just a random ranking from WCL) it seems like the player is following Echoing Shock with Earth Shock and Flame Shock, **not Lava Burst**. Is the suggestion-text wrong or am I missing something?

joshinat0r (changes requested):

src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
HawkCorrigan: Nah I am just dumb.
The logic is wrong.
You are supposed to only use it on Lava Burst on Beta.

joshinat0r (approved):

PR #4053, opened 2020-11-04
[Core] run build & docker actions only on upstream

joshinat0r:
This way the `Build` & `Publish Docker Image` jobs only run on upstream, since they fail on forks.
[Looks like this on forks](https://github.com/joshinat0r/WoWAnalyzer/actions/runs/346249621) compared to [previously](https://github.com/joshinat0r/WoWAnalyzer/actions/runs/328925972)


Pewtro:
Cheers lad!

Issue #3222, opened 2019-07-20
Add support for Enhancement Shaman


MartijnHols:
Since I do not know Enhancement Shaman, I can not specify a definition of done. If you think it's ready to be marked as supported, you'll have to somehow convince us of this as well. That probably means it has to check similar things to what guides tell you to look for, such as in Icy Veins' [how to improve](https://www.icy-veins.com/wow/how-to-improve-enhancement-shaman) section.

MartijnHols:
Added a bounty to equal a 12 month subscription to the game:

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=77338846)](https://www.bountysource.com/issues/77338846-add-support-for-enhancement-shaman)

Yoiro:
Hey, I'm playing Enhancement and I'd like to contribute here, but I'm not sure where to start or what to look for.

abelito75:
[https://github.com/WoWAnalyzer/WoWAnalyzer/wiki](url)
Is a good starting place but its a pretty manual process of going through all the files in the enhancement parser and making sure they are all valid :P

djanse:
I've a friend playing Enhancement Shaman at the moment. In order to help him improve in his raiding situations I'll be making a lot of changes in order too get the analyser up-to-date. Expect a lot to be updated in the current month.

kenrms:
I can help, too. I play enhance now and plan on continuing in SL. Already got a chunk of it started on my fork.

abelito75:
Sound great!

PR #4055, opened 2020-11-05
[Enhancement] Calculate Hailstorm stack statistics

darya-talanina:
Related issue: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3222

Ideas behind this analyzer:

- Hailstorm stacks are gained when player uses Maelstrom Weapon stacks, but not when Maelstrom Weapon stacks expire by themselves
- `onMaelstromWeaponStackApply` method helps track current number of Maelstrom Weapon stacks
- `onSpellWithCastTimeCast` method is used to detect when Maelstrom Weapon stacks are spent and determine the number of Hailstorm stacks gained

Used [this logs](https://www.warcraftlogs.com/reports/WYf4JAB7vPaVKrgD) to test.

Screenshot 2020-11-08 at 15 37 47

Screenshot 2020-11-08 at 15 37 52

### DIFF
--- a/src/parser/shaman/enhancement/CHANGELOG.tsx
+++ b/src/parser/shaman/enhancement/CHANGELOG.tsx
@@ -6,6 +6,7 @@ import SpellLink from 'common/SpellLink';
 import { change, date } from 'common/changelog';
 
 export default [
+  change(date(2020, 11, 5), <>Added stack statictics for <SpellLink id={SPELLS.HAILSTORM_TALENT.id}/></>, [Mae]),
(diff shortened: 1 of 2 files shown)


Pewtro:
The statistic feels very busy. Could some of it be moved together and/or moved to the tooltip? What is the most important thing that people can utilize for improving? 

An idea could be merging the gained/overcapped one to 15/57 stacks gained as an example?

darya-talanina:
@Pewtro You're right. I reworked the statistic to only keep the % of used stacks (as this is what matters most) and moved the breakdown of stacks used / gained / overcapped into the tooltip: you can see it on the new screenshots I uploaded.

Pewtro:
LGTM!

PR #4058, opened 2020-11-06
[Death Knight] Runeforge suggestions

joshinat0r:
Created a shared module for DKs that easily creates suggestions based on the enchant used, currently a BDK can run Razorice and WoWA wont complain.
The BDK one will throw a major suggestion for all enchants except FC and Hysteria since there is no point in running anything other.
Unholy & 2h Frost could adopt the module already, DW Frost would need some more logic to handle offhand enchants.

image

The suggestion-part in `src/parser/deathknight/shared/runeforges/RuneForgeChecker.tsx` looks a bit ugly because typescript kept yelling at me, there isn't going to be a suggestion when `activeSuggestion` is empty so it shouldn't really matter.

Also fixed an issue with Hysteria, the tooltip was showing `potentially gained DS` while it should've displayed `actually gained DS`. Amount of wasted DS by capping RP is now also displayed.


src/parser/deathknight/shared/runeforges/RuneOfHysteria.tsx
```diff
@@ -98,9 +106,9 @@ class RuneOfHysteria extends Analyzer {
         size="flexible"
         tooltip={(
           <>
-            You gained {Math.floor(this.runicPowerGainedByHysteria)} RP by using {SPELLS.RUNE_OF_HYSTERIA.name}, resulting in {this.gainedDeathStrikes} additional Death Strikes.<br />
+            You gained {Math.floor(this.runicPowerGainedByHysteria)} RP by using {SPELLS.RUNE_OF_HYSTERIA.name}, resulting in {this.gainedDeathStrikes} additional {SPELLS.DEATH_STRIKE.name}.<br />
```
Khazakdk: Is it possible to make the mention of DS conditional on if the players spec is blood?  Occurred to me that someone playing hysteria on frost or unholy might take this tooltip the wrong way

joshinat0r: No one cares about the other specs anyways 😄 
But you're right, DS hint is now only shown when the player is blood. Could add other spells for frost/unholy, for now its just showing `Gained x RP and wasted X RP` for them

Pewtro:
LGTM, and the one review comment has been solved - thanks!

PR #4059, opened 2020-11-06
3801 - Replaced the deprecated StatisticBoxes for feral druid

leozhekov:
Small comparison can be seen here:
image


Pewtro (approved):
Looks good to me, thanks!

PR #4062, opened 2020-11-07
[Core] Conduit, Soulbind, Covenant parsing as it works per latest raid testing

Pewtro:
Depends on #4061

### DIFF
--- a/src/parser/core/Combatant.ts
+++ b/src/parser/core/Combatant.ts
@@ -3,20 +3,11 @@ import RACES from 'game/RACES';
 import TALENT_ROWS from 'game/TALENT_ROWS';
 import GEAR_SLOTS from 'game/GEAR_SLOTS';
 import traitIdMap from 'common/TraitIdMap';
-import { Enchant } from "common/ITEMS/Item";
+import { Enchant } from 'common/ITEMS/Item';
 import SPELLS from 'common/SPELLS';
 import { findByBossId } from 'raids';
 import CombatLogParser, { Player } from 'parser/core/CombatLogParser';
-import {
-  Buff,
-  CombatantInfoEvent,
-  EventType,
-  Item,
-  Trait,
-  Soulbind,
-  Conduit,
-  Covenant,
-} from 'parser/core/Events';
+import { Buff, CombatantInfoEvent, Conduit, EventType, Item, SoulbindTrait, Trait } from 'parser/core/Events';
 
 import Entity from './Entity';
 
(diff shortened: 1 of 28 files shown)


src/parser/core/Combatant.ts
```diff
@@ -105,13 +96,16 @@ class Combatant extends Entity {
     };
 
     this._parseTalents(combatantInfo.talents);
-    this._parseTraits(combatantInfo.artifact);
-    this._parseEssences(combatantInfo.heartOfAzeroth);
-    this._parseCovenant(combatantInfo.covenant);
-    this._parseSoulbind(combatantInfo.soulbind);
-    this._parseConduits(combatantInfo.conduits);
+    this._parseTraits(combatantInfo.artifact as Trait[]);
+    this._parseEssences(combatantInfo.heartOfAzeroth as Essence[]);
```
Dambroda, line 100: should these be in an `else` after checking for shadowlands expansion? I havent looked at the actual combatantinfo shape, but I would assume if the artifact/hoa are essences *or* conduits it might be weird to parseEssences using the conduit data?

src/parser/core/Combatant.ts
```diff
@@ -105,13 +96,16 @@ class Combatant extends Entity {
...
     this._parseGear(combatantInfo.gear);
     this._parsePrepullBuffs(combatantInfo.auras);
+    if (combatantInfo.expansion === 'shadowlands') {
+      this._parseCovenant(combatantInfo.covenantID);
+      this._parseSoulbind(combatantInfo.soulbindID);
+      this._parseSoulbindTraits(combatantInfo.artifact as SoulbindTrait[]);
```
Dambroda, line 106: its not a huge deal to resolve these, but the way to do it without casting is typeguards  https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards


alternatively, when we load the combatant info at first we could check expansion at that point and assign to a new field to avoid confusion. i.e just add a soulbindTraits field and populate it with info.artifact if expansion = shadowland

Dambroda (commented):
one question, one comment, otherwise LGTM

src/parser/core/Combatant.ts
Pewtro: I planned on just deleting them when we swap over to non-prepatch since they're then irrelevant

src/parser/core/Combatant.ts
Pewtro: Interesting idea, which I think we should probably do if they keep the naming scheme once SL goes live and it's not just some temporary beta-naming-scheme, because it just doesn't make sense to have them be called artifact and heartOfAzeroth 😄

Dambroda (approved):

Pewtro:
Let's gooooo, time to implement all these conduits and soulbinds and what else easier

PR #4069, opened 2020-11-08
[Retribution] Empyrean Power Tracker

LordSkeletor:
-Empyrean Power tracker based on Divine Purpose tracker. 
-Azerite trait version does inflate the procs but that will stop being an issue in a few weeks.
image
image


src/parser/paladin/retribution/modules/talents/EmpyreanPower.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const BUFF_TIME: number = 15000 * .95;//add buffer since log events lmao
+const TRACK_BUFFER = 500;
+
+class EmpyreanPower extends Analyzer {
+
+  averageTimeTillBuffConsumed: number = 0;
+
+  hasProc: boolean = false;
+  procsWasted: number = 0;
```
Dambroda, line 26: When setting a field to a default value that correctly implicitly types, you don't need to explicitly set the type. This might get linted out later, so it might potentially save you some time later on if the rule ever comes through.

```typescript
myField: number = 0; // Ok, but not needed
myField = 0; 

myField: boolean = false; // Ok, but not needed
myField = false;
```

src/parser/paladin/retribution/modules/talents/EmpyreanPower.tsx
```diff
@@ -0,0 +1,103 @@
...
+      return;
+    }
+
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.CRUSADER_STRIKE), this.castCounter);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(SPELLS.DIVINE_STORM), this.divineStormDamage);
+    this.addEventListener(Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.EMPYREAN_POWER_BUFF), this.applyBuff);
+    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.EMPYREAN_POWER_BUFF), this.removeBuff);
+  }
+
+  castCounter(){
```
Dambroda: Spacing after all of the method parenthesis in the file

Dambroda (approved):
Thank you! Two minor optional code style changes.

PR #4073, opened 2020-11-08
[Enhancement] Display Maelstrom Weapon statistics: stacks gained, used and time spent at max stacks

darya-talanina:
Related to https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3222

Used [this logs](https://www.warcraftlogs.com/reports/YWJTjbZVCrGgH7vB) to verify calculations.

Screenshot 2020-11-11 at 21 50 46
Screenshot 2020-11-11 at 21 50 52


src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
```diff
@@ -0,0 +1,231 @@
...
+    );
+
+    const MAELSTROM_WEAPON_SPENDERS: Spell[] = [
+      SPELLS.CHAIN_HEAL,
+      SPELLS.CHAIN_LIGHTNING,
+      SPELLS.HEALING_SURGE,
+      SPELLS.LIGHTNING_BOLT,
+    ]
+
+    MAELSTROM_WEAPON_SPENDERS.forEach(spell => {
```
Sharrq: addEventListener accepts arrays, so you can do the following

this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(MAELSTROM_WEAPON_SPENDERS), this.castMaelstromWeaponSpender);

src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
```diff
@@ -0,0 +1,231 @@
...
+class MaelstromWeapon extends Analyzer {
+  protected stacksGained = 0;
+  protected stacksUsed = 0;
+  protected stacksExpired = 0;
+  
+  protected currentStacks = 0;
+
+  protected cappedIntervals: Intervals;
+
+  get MAX_STACKS_SPENT_PER_CAST() {
```
Sharrq: this and MAX_STACKS can be constants, and should be above the class (below the imports)

const MAX_STACKS_SPENT_PER_CAST = 5
const MAX_STACKS = 10

src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
```diff
@@ -0,0 +1,231 @@
...
+      this.removeAllStacks
+    );
+
+    // This only occurs when player casts a spender, because in case stacks expire, 'removebuff' event occurs.
+    this.addEventListener(
+      Events.removebuffstack.by(SELECTED_PLAYER).spell(SPELLS.MAELSTROM_WEAPON_BUFF),
+      this.spendStacks
+    );
+
+    const MAELSTROM_WEAPON_SPENDERS: Spell[] = [
```
Sharrq: this should be above the class with the other constants

src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  removeAllStacks(event: RemoveBuffEvent) {
+    if (this.currentStacks === 0) {
+      // stacks were spend on a spender spell
+      return;
+    }
+
+    if (this.currentStacks === 10) {
```
Sharrq: use your MAX_STACKS constant instead of the number

src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
```diff
@@ -0,0 +1,231 @@
...
+class Intervals {
+  protected intervals: Interval[];
+
+  constructor() {
+    this.intervals = [];
+  };
+
+  startInterval(timestamp: number) {
+    if (this.isLastIntervalInProgress) {
+      console.log('Intervals: cannot start a new interval because one is already in progress.')
```
Sharrq: console logs should only be run if a "debug mode" is enabled.

So you can do 'debug && console.log(...' and add a debug = false constant, or remove the logs.

src/interface/icons/Uptime.js
```diff
@@ -3,7 +3,12 @@ import React from 'react';
 // https://thenounproject.com/term/duration/370713/
 // duration by Bohdan Burmich from the Noun Project
 const Icon = ({ ...other }) => (
-  <svg xmlns="http://www.w3.org/2000/svg" viewBox="10 10 80 80" className="icon" {...other}>
+  <svg
```
Sharrq: im assuming Prettier automatically formatted this. No real reason to change this from how it was before.

Sharrq (changes requested):

src/interface/icons/Uptime.js
darya-talanina: I would argue that any code style changes that do not make the linter fail should be accepted.
But I can change it back as you ask.

src/parser/shaman/enhancement/modules/core/MaelstromWeapon.tsx
darya-talanina: Done.
Could you please explain when should we define constants outside of class and when can we do it inside?

darya-talanina:
@Sharrq Thank you for the review! I implemented you change suggestions and also left a couple of questions for you.
Could you answer them and if everything looks good, approve the pull request please?

Sharrq:
I wasnt sure about the Interval/Intervals classes that were added to the bottom of that, so I asked one of the admins to take a look at it.

Their response was that those do not need to be their own classes as they can be done with getters and that while there may be one or two exceptions around, in general there should be one class per file and that if it absolutely needed it's own class then it should be in it's own file.

darya-talanina:
@Sharrq surely it could be done via getters and setters on the `MaestromWeapon` class: but the resulting code would be harder to understand and to reuse.
I find that the logic that `Interval` and `Intervals` classes encapsulate doesn't belong to the `MaelstromWeapon` analyzer, hence separate classes. `MaelstromWeapon` analyzer doesn't have to know the details of interval manipulation. It only needs to use its API.

Having this logic in dedicated classes allows to keep only maelstrom weapon-specific logic in the analyzer.
It could also be useful in case we need to share it with other analyzers.

I moved `Interval` and `Intervals` classes to separate files.

Sharrq (approved):

PR #4079, opened 2020-11-16
[Core] Filtered Active Time/Downtime

Sharrq:
Open to suggestions, this is working at the moment but seems to be skewed by some wonky things going on with Haste in Prepatch (Im assuming some haste values are using pre-squish haste amounts).

This adds the ability to get the active time/downtime for a specfied time frame. So for example, i could get the player's uptime while Combustion was active.

It has some other uses that I plan on looking into (Maybe forced downtime on bosses?)


Pewtro (approved):
A bit hard to verify all the math without loading it up locally - but this looks okay to me!

Sharrq:
im pretty confident in the math and sketched out a lot of it as it turned out to be pretty complicated to have the math work properly.

There might be a simpler way to do it, but i couldnt find one.

Issue #3664, opened 2020-07-25
"Be Well Prepared" checklist for Shadowlands


Pewtro:
## Existing Features
- [x] Update list of best enchants
- [x] Update list of potions
- [x] Add check for flask
- [x] Add check for best flask (same as enchants)
- [x] Add check for foodbuff
- [x] Add check for best foodbuff (same as enchants)
## New Features
- [ ] Add check for gems (X/Y gem slots not empty)
- [ ] Add check for best gems (same as enchants)
- [x] Sharpening Stones https://shadowlands.wowhead.com/spells/professions/blacksmithing/shadowlands-plans/weapons-mods
- [x] Weight Stones https://shadowlands.wowhead.com/spells/professions/blacksmithing/shadowlands-plans/weapons-mods
- [x] Weapon Oils
- [ ] Legendary equipped check

Sharrq:
Should add sharpening stones and weapon oils and such to this too

Pewtro:
Weightstones are also a thing https://shadowlands.wowhead.com/spells/professions/blacksmithing/shadowlands-plans/weapons-mods

PR #4080, opened 2020-11-16
[Core] Weapon Oils & Sharpening Stones + Weapon Enchants

Sharrq:
Added Weapon Oils and Sharpening Stones to the Checklist and added Weapon Enchants to the Enchant Checker.

image


Pewtro:
If you fix the conflicts this should be fine (just override the bot commit probably)

Pewtro:
Oh - and also remember to tick away at https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3664. 

I think this PR is missing these two btw - They are worse versions than the one you have implemented so maybe we should have enchanted vs best enchants here too? 
https://shadowlands.wowhead.com/spell=307719/porous-weightstone
https://shadowlands.wowhead.com/spell=307717/porous-sharpening-stone

Sharrq:
I couldn't find those two in the blacksmithing unlearned and they weren't on the consumables vendor so I thought they got removed. I'll add them. Maybe I just missed them

Sharrq:
So i went to add the non max Sharpening Stone/Weight stone, but from looking at Enchants .... it defines the list of "Min Enchants" but never actually uses it. 

It is just checking to see if an enchant is on there at all (i.e. if item.permanentEnchant exists/has a value). So im not sure that we need the Min item list for Enchants or Oils/Stones?? Cause Oils/Stones is doing the same thing ... if item.temporaryEnchant exists/has a value then it has something on it.

abelito75:
Fixed the changelog for you buddy <3

abelito75:
I might have left a hanging character in the change log : ^) but also these things

Edit: i just put a 0 in front of the 1 and now its angry. Just remove that and you gucci on the changelog... stupid linter can't you read abe

Sharrq:
@Pewtro are these ok for me to go ahead and merge it? Not sure if my comments about the lower stones are enough to satisfy you or not

PR #4083, opened 2020-11-19
3801 - replaced the deprecated statistics of the resto druid module 

leozhekov:
Small comparison can be seen here:
image


abelito75 (approved):
Thanks for all the hard work!

PR #4085, opened 2020-11-20
Convert AtonementHealingBreakdown to functional component 

tompeng:
For #3800.

Converted a small component `AtonementHealingBreakdown` to functional component.
Made use of hooks `useMemo` to memoize the table body, since the calculation seemed somewhat expensive.
Added a type for `AtonementHealingDone.bySource`.

I'm still learning more about the codebase behind this awesome project. Any feedback would be appreciated, thanks.


MartijnHols (approved):
Excellent, thanks!

I don't think the useMemo will actually do anything, since this should only be re-rendered when the toggle is triggered and it should therefore always re-render as well. It doesn't hurt here though.

PR #4097, opened 2020-11-25
[Unholy Death Knight] Unholy Spellbook changes

Khazakdk:
A couple of core impacting changes in this one

- Added a general SPELLS file for covenant signature abilities
- Set timeline to filter out CASTS_THAT_ARENT_CASTS to prevent the hidden rune spells from cluttering the timeline


Juko8 (approved):
👍

PR #4099, opened 2020-11-28
[Core] Delete BFA & Add New Dungeons

Sharrq:
- Deleted all of the BFA Modules (Left the ITEM/SPELL IDs for now until things like the Character Tab can be updated)
- Added configs and backgrounds for the Shadowlands Dungeons


Pewtro:
you should be playing shadowlands not here! 😄

abelito75:
300 files yikes from me dawg. But probably fine if you fix the snapshot

PR #4101, opened 2020-11-30
[HPriest] Fixing bug with POM tracking

TJMoats:
I think this was broken when someone updated the module to use the event system. Easy thing to miss :D


abelito75 (approved):

abelito75:
Thanks

PR #4104, opened 2020-12-01
[HPriest] 9.0.2

TJMoats:


Pewtro:
Was this actually the only changes necessary to bump holy from 8.2.5 to 9.0.2? No changed recommendation/talents etc? 😮

PR #4107, opened 2020-12-02
[DisciplinePriest] Fix Schism damage multiplier to 25%

VladimirMakaev:
Noticed this when was looking through the code. It was changed from 40% to 25%


abelito75:
You shouldn't delete .env.local.example and hit up a change log and you'll be looking good

VladimirMakaev:
Cleaned up my commit

Pewtro:
Thanks!

PR #4118, opened 2020-12-08
Quicker build workflow

MartijnHols:


Pewtro:
The changes to this caused 502 errors. Loading up the commit before this allows logs to be loaded again, so reverting for now.

PR #4119, opened 2020-12-08
[Core] Goodbye BFA

Pewtro:
The great purging of BFA


Pewtro:
Fuck it, let's break this site like every other wow 3rd party site, yeeeet.

PR #4132, opened 2020-12-09
Fix a bunch of spells

MartijnHols:


MartijnHols:
integration tests are broken in shadowlands too so this doesn't make it worse I don't think

PR #4144, opened 2020-12-10
[Warlock] Fix some afflication issues

kevindqc:
Update spells to take into account that:
- Unstable Affliction is no longer a soul shard spender, so it should have higher uptime.
- Unstable Affliction can't be stacked anymore - only one at a time.
- The Contagion passive 10% increased damage taken when someone has Unstable Affliction is gone.
- Shadow Embrace is now baseline.
- Some mana costs were changed and spells taken off the GCD

Added Night Fae spells.

Fixed an issue (can be seen on [this report](https://www.wowanalyzer.com/report/z4kpCD2VxqbKNHga/5-Normal+Shriekwing+-+Kill+(5:43)/Maleficien/standard) for Unstable Affliction, can see on [the logs](https://www.warcraftlogs.com/reports/z4kpCD2VxqbKNHga/#fight=5&type=summary&source=12&view=events) here that UA ticks at 00:00:01.685, but there's no "is afflicted by" event) where the uptime was getting messed up in some cases where the Apply Debuff event was nowhere to be seen
- When we see a damage event from a dot, but haven't seen an apply debuff before, we will fabricate one with the timestamp set to the beginning of the fight


MartijnHols (approved):
Great start! Thanks a lot.

PR #4152, opened 2020-12-12
[Core] Make Missing Cloak enchants a "minor" issue

Sharrq:
Made the suggestion for missing a cloak enchant a minor issue instead of major.


abelito75:
I'd argue its pretty major for healers to not have a free 30 leech

PR #4165, opened 2020-12-13
[HPriest] bugfixes

TJMoats:
These two modules were broken, now they are fixed!

3giXAKp


abelito75:
update the change log and we blast

Pewtro:
Thanks!

PR #4167, opened 2020-12-14
Druid - Update SpellID for Berserk

jwmclark:
Warcraft logs is tracking Berserk usage as spell ID 106951 (which triggers 343216.)

Prior to this change, Berserk usage would show as 0 uses.

Example log: https://www.warcraftlogs.com/reports/jh6FMGrfqgtNBCmQ#fight=16

Berserk as spell ID 343216:
Berserk-343216

Berserk as spell ID 106951:
Berserk-106951


Pewtro:
There are modules that rely on the spell id being provided being the buff id, so there should be a split of BERSERK and BERSERK_BUFF and then change the SPELLS.BERSERK values to the relevant one wherever they are in the codebase.

jwmclark:
> There are modules that rely on the spell id being provided being the buff id, so there should be a split of BERSERK and BERSERK_BUFF and then change the SPELLS.BERSERK values to the relevant one wherever they are in the codebase.

I would agree with your assessment but unless I am reading the logs incorrectly, WCL does not seem to be respecting the actual spell ID of the buff and simply tracking as the ability.

See: [Warcraft Logs Link](https://www.warcraftlogs.com/reports/jh6FMGrfqgtNBCmQ#fight=16&source=2&type=auras&pins=0%24Separate%24%23244F4B%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%24195018127.0.0.Druid%24false%24106951&ability=106951&view=events)

In this case We can see usage based on WCL tracking [106951](https://www.wowhead.com/spell=106951/berserk) as gained and faded. Since only [106951](https://www.wowhead.com/spell=106951/berserk) is tracked we can only assume [343216](https://www.wowhead.com/spell=343216/berserk) is active. 

Due to this, wouldn't splitting Berserk into the Ability and the Buff would result in the ability usage being tracked but not the buff being applied?

If I am misunderstanding how the ID is parsed and used in the modules, I will dive deeper.

Pewtro:
Buff and cast id is the same in this case, so it's fine - thanks! 😄

PR #4170, opened 2020-12-15
[BRM] Fix the Light Brewing talent ID

mathieu-carlier:
Fix the ID of the light brewing talent, that wasn't up to date. 
Note that this will also fix the CD calculation of Purifying and Celestial Brew.
Live:
image
Fixed:
image

### DIFF
--- a/src/parser/monk/brewmaster/integrationTests/example.test.ts.snap
+++ b/src/parser/monk/brewmaster/integrationTests/example.test.ts.snap
@@ -170,7 +170,7 @@ exports[`Brewmaster Monk integration test: example log BrewCDR matches the stati
         <div
           className="value"
         >
-          26.30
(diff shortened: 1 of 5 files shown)


src/parser/monk/brewmaster/integrationTests/example.test.ts.snap
```diff
@@ -1381,7 +1381,7 @@ exports[`Brewmaster Monk integration test: example log CastEfficiency matches th
               }
             >
               22
-              /36
+              /32
```
emallson, line 1384: We will need a new test log for this. Looks like the old one does not have the new talent ID.

emallson (approved):
Approving for now. We will need a new test log for Brewmaster. I'll file an issue for it.

PR #4174, opened 2020-12-15
[Druid / Feral] Update Bloodtalons for Shadowlands

RobinKa:
Previously Bloodtalons was generated by using (instant) regrowths with Predatory Swiftness. Now it's generated by using 3 different combopoint generating abilities within 4 seconds and spent only on Rip, Ferocious Bite and Primal Wrath (AOE Rip). I'm new so let me know what else needs to be changed and I'll get on it. I assume anything involving Predatory Swiftness could be removed as it's a healing (or utility) only thing now.

The underlying Bloodtalons code should still work fine mostly from what I can tell as it only checks that the Bloodtalons buff is used by one of the spenders.

**Still to do**

~~Need to make Primal Wrath be considered as a spender.~~

Also should probably update the text to give suggestions on which 3 spells to use for generation (as far as I can tell from reading the Feral discord, sometimes it is best to use Thrash, and sometimes not, depending on gear / talents / conduits).

Furthermore refreshing the buff should not be counted as wasted if eg. refreshing rake, and using shred and brutal slash which would also regenerate the buff.


src/parser/druid/feral/modules/features/checklist/Module.js
```diff
@@ -87,7 +85,6 @@ class Checklist extends BaseChecklist {
           moonfireDowngrade: this.moonfireSnapshot.downgradeSuggestionThresholds,
 
           // bloodtalons
-          predatorySwiftnessWasted: this.predatorySwiftness.suggestionThresholds,
           bloodtalonsWasted: this.bloodtalons.suggestionThresholds,
```
Pewtro, line 88: If you are removing predatory swiftness suggestions from the checklist - does this mean that this is no longer an interesting metric at all, or? 

If it is no longer an interesting metric at all - the suggestions and thresholds should probably be removed from the PredatorySwiftness file too. 

If it is an interesting metric, but you are removing for some other reason, then we should address that issue instead of removing them from the checklist.

RobinKa: The only use for it now would be to use instant-cast entangling roots if some mechanic requires it, or to heal with instant regrowth when you have downtime which should be almost 0 if owlweaving, which is going into moonkin to cast sunfire (and starsurge while heart of the wild is active). Without owlweaving (a ~2-5% damage loss afaik) there is quite a lot of downtime (~30% of the time spent waiting) where spending a GCD on the instant cast heal would still be of minor use. For damage there isn't any use for it anymore.

Do you only want suggestions that increase damage or also other minor benefits such as this healing? If those minor non-damage benefits aren't interesting anything involving Predatory Swiftness could be removed I assume.

Pewtro: Sounds like something that could initially be removed to me. If owlweaving gets to stay around, maybe some implementation could be made to make an owlweaving build and a non-owlweaving build. We have some support to make different builds like that (holy paladin with glimmer vs no glimmer, mages with no icelance vs regular etc).

RobinKa:
This is mostly done now I think and the other changes such as suggestions for which 3 combo point-generating abilities to use to generate bloodtalons should probably be separate PRs.

src/CHANGELOG.tsx
```diff
@@ -13,6 +13,7 @@ export default [
   change(date(2020, 12, 19), 'Adjust our weak Potion suggestion to only show if a weak potion was actually used', Putro),
   change(date(2020, 12, 17), 'Added phase information for Castle Nathria bosses', Sharrq),
   change(date(2020, 12, 16), 'Fix character API realm slugs for non-English realms.', Zerotorescue),
+  change(date(2020, 12, 15), 'Update Feral Bloodtalons for Shadowlands', Tora),
   change(date(2020, 12, 15), <> Add support for the Engineering scopes, <ItemLink id={ITEMS.ENCHANT_WEAPON_OPTICAL_TARGET_EMBIGGENER.id} /> and <ItemLink id={ITEMS.ENCHANT_WEAPON_INFRA_GREEN_REFLEX_SIGHT.id} />, to the enchant checker. </>, Putro),
```
Pewtro: Your changelogs should go in the feral druid changelog

Pewtro (changes requested):
If you move your changelog to the feral specific one this seems good to go

src/CHANGELOG.tsx
RobinKa: Whoops, thanks!

Pewtro (approved):

Pewtro:
In changelog you should only use <> when needing to use JSX elements such as SpellLink or ItemLink etc - but not blocking over that

Pewtro:
Thanks for the help on this!

PR #4179, opened 2020-12-17
[#3799] Converting indexbyID to TypeScript

### LINKED ISSUE #4188: [Arcane] ArcanePower.tsx : does not count bad arcane power casts!
if statement on line 116 is never fulfilled and therefore bad casts during Arcane Power are never counted.

can be seen in [Wowanalyzer Vashtii](https://wowanalyzer.com/report/3F4jnmTdXVDh8kzt/3-LFR+Huntsman+Altimor+-+Kill+(6:16)/Vashtii/standard/timeline) @ around 0:12 where supernova was used with arcane power

### LINKED ISSUE #4189: [Elemental Shaman] Stormkeeper not showing correct DPS on statistics page
as evident on [the analysis of top 1 elemental shaman statistics page](https://wowanalyzer.com/report/KnG7AYB83bCJgTQj/11-Mythic+Huntsman+Altimor+-+Kill+(4:57)/Raime/standard/statistics) 

I realise the spec/class is unmaintained atm

image

CasperKjaerhus:
SPELLS, SPECS, ITEMS and anything else that uses indexbyID now uses a generic typed version of it. #3799

This did however lead me on a debug hunt through many files where legendaries were used implicitly, i tried to fix the best of my abilities. Due to being new to the codebase some fixes *could* be logically wrong so please review this PR carefully. 

Also fixes #4189 and fixes #4188


src/common/ItemLink.tsx
```diff
@@ -37,7 +37,7 @@ const ItemLink = (props: Props) => {
   } else if (props.details) {
     quality = Math.max(props.details.itemLevel >= 370 ? 4 : 3, props.details.quality);
   } else {
-    quality = ITEMS[id] ? ITEMS[id].quality : 0;
+    quality = ITEMS[id] ? 0 : 0;
```
abelito75: I feel like this line does nothing

Pewtro:
Hey, first of all thanks for the PR!

Whatever in this PR is forcing ``as X`` and ``brynadorsMight.bonusID ? brynadorsMight.bonusID : 0`` everywhere shouldn't be in the PR imo, that's not what we would want to have to do everywhere in the code.
These modules should very much crash if those values are not set, so we can fix the issues. Allowing them to run with some fallback value will merely cause inaccuracies and hide potential errors. This goes for all the instances of it throughout this PR pretty much.

src/parser/core/Combatant.ts
```diff
@@ -509,7 +509,7 @@ class Combatant extends Entity {
         ability: {
           abilityIcon: buff.icon.replace('.jpg', ''),
           guid: buff.ability,
-          name: spell ? spell.name : undefined,
+          name: spell ? spell.name : "undefined",
           type: 0,
```
Pewtro: this should very much be undefined without ""

CasperKjaerhus:
> Hey, first of all thanks for the PR!
> 
> Whatever in this PR is forcing `as X` and `brynadorsMight.bonusID ? brynadorsMight.bonusID : 0` everywhere shouldn't be in the PR imo, that's not what we would want to have to do everywhere in the code.
> These modules should very much crash if those values are not set, so we can fix the issues. Allowing them to run with some fallback value will merely cause inaccuracies and hide potential errors. This goes for all the instances of it throughout this PR pretty much.

This is 100% true, I thought about throwing an exception instead or would you rather i log the error to console?

src/parser/monk/mistweaver/modules/features/MistweaverHealingEfficiencyTracker.tsx
```diff
@@ -79,7 +79,7 @@ class MistweaverHealingEfficiencyTracker extends HealingEfficiencyTracker {
 
   getSoothingMistDetails(spellInfo: SpellInfoDetails) {
     // the default tracker gets the healing of the soothing mists, but only the mana for the first cast. Every tick costs mana.
-    spellInfo.manaSpent = this.soothingMist.soomTicks * SPELLS.SOOTHING_MIST.manaCost;
+    spellInfo.manaSpent = this.soothingMist.soomTicks * (SPELLS.SOOTHING_MIST.manaCost ? SPELLS.SOOTHING_MIST.manaCost : 1);
```
abelito75: This would break the accuracy of healer efficiency Tracker if manaCost is ever null. I'd rather have it error out for 30 minutes and fix it manually

src/parser/monk/mistweaver/modules/shadowlands/legendaries/AncientTeachingsoftheMonastery.tsx
```diff
@@ -23,7 +24,8 @@ class AncientTeachingsoftheMonastery extends Analyzer {
    */
   constructor(options: Options) {
     super(options);
-    this.active = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.ANCIENT_TEACHINGS_OF_THE_MONASTERY.bonusID);
+    const ANCIENT_TEACHINGS_OF_THE_MONASTERY = SPELLS.ANCIENT_TEACHINGS_OF_THE_MONASTERY as LegendarySpell;
```
abelito75: I'd rather have it crash so I can manually fix it.

src/parser/monk/mistweaver/modules/shadowlands/legendaries/CloudedFocus.tsx
```diff
@@ -28,7 +29,8 @@ class CloudedFocus extends Analyzer {
 
   constructor(options: Options) {
     super(options);
-    this.active = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.CLOUDED_FOCUS.bonusID);
+    const CLOUDED_FOCUS = SPELLS.CLOUDED_FOCUS as LegendarySpell;
```
abelito75: I'd rather have it crash so I can manually fix it.

src/parser/monk/mistweaver/modules/talents/Lifecycles.tsx
```diff
@@ -60,8 +60,8 @@ class Lifecycles extends Analyzer {
       this.castsNonRedViv += 1;
       return;
     }
-    this.manaSaved += SPELLS.VIVIFY.manaCost * LIFECYCLES_MANA_PERC_REDUCTION;
-    this.manaSavedViv += SPELLS.VIVIFY.manaCost * LIFECYCLES_MANA_PERC_REDUCTION;
+    this.manaSaved += (SPELLS.VIVIFY.manaCost ? SPELLS.VIVIFY.manaCost : 1) * LIFECYCLES_MANA_PERC_REDUCTION;
```
abelito75: If these values are ever null and it defaults to 1 it will just provide inaccurate information making the whole statistic pointless and misleading. Since it wouldn't crash people won't come ask for it to be fixed either and just assume `damn lifecycles is literally the worst talent in the game and even playing it to the mas it doesn't save me more than 100 mana` on top of then maybe going into peak and making me have to deal with them there

src/parser/monk/mistweaver/modules/talents/Lifecycles.tsx
```diff
@@ -78,15 +78,15 @@ class Lifecycles extends Analyzer {
     // Checking for chiji stacks and determine mana reduction
     const chijiStacksAtEnvCast = this.selectedCombatant.getBuff(SPELLS.INVOKE_CHIJI_THE_RED_CRANE_BUFF.id, event.timestamp)?.stacks;
     if (!chijiStacksAtEnvCast) {
-      this.calculateEnvManaSaved(SPELLS.ENVELOPING_MIST.manaCost);
+      this.calculateEnvManaSaved(SPELLS.ENVELOPING_MIST.manaCost ? SPELLS.ENVELOPING_MIST.manaCost : 0);
```
abelito75: If these values are ever null and it defaults to 1 it will just provide inaccurate information making the whole statistic pointless and misleading. Since it wouldn't crash people won't come ask for it to be fixed either and just assume damn lifecycles is literally the worst talent in the game and even playing it to the mas it doesn't save me more than 100 mana on top of then maybe going into peak and making me have to deal with them there

src/parser/monk/mistweaver/modules/talents/Lifecycles.tsx
```diff
@@ -78,15 +78,15 @@ class Lifecycles extends Analyzer {
...
       return;
     }
     //check for free cast from chiji
     if (chijiStacksAtEnvCast === MAX_CHIJI_STACKS) {
       return;
     }
     //have to do this weird because blizzard decided to make each chiji stack reduce the mana cost by 1001 instead of and exact 33%
-    const modifiedManaCost = SPELLS.ENVELOPING_MIST.manaCost - (CHIJI_MANA_SAVED_PER_STACK * chijiStacksAtEnvCast);
+    const modifiedManaCost = (SPELLS.ENVELOPING_MIST.manaCost ? SPELLS.ENVELOPING_MIST.manaCost : 0) - (CHIJI_MANA_SAVED_PER_STACK * chijiStacksAtEnvCast);
```
abelito75: If these values are ever null and it defaults to 1 it will just provide inaccurate information making the whole statistic pointless and misleading. Since it wouldn't crash people won't come ask for it to be fixed either and just assume damn lifecycles is literally the worst talent in the game and even playing it to the mas it doesn't save me more than 100 mana on top of then maybe going into peak and making me have to deal with them there

src/parser/core/Combatant.ts
CasperKjaerhus: Yes, it requires that i set name as nullable, would that be preffered?

src/parser/monk/mistweaver/modules/talents/Tier45Comparison.tsx
```diff
@@ -152,7 +152,7 @@ class Tier45Comparison extends Analyzer {
       //life cycles reduces mana cost of two spells if you casted the other before hand
       //so best = (x-1) * VivCost * LifcylesReduction + x * EnvCost * LifcylesReduction = (best + ReducedViv) / ReducedEnv = x
       //x-1 since you viv first in all fights
-      this.lifecycles.requiredEnvs = Math.ceil((this.best.manaFrom + SPELLS.VIVIFY.manaCost * SPELLS.LIFECYCLES_VIVIFY_BUFF.manaPercRed) / SPELLS.ENVELOPING_MIST.manaCost * SPELLS.LIFECYCLES_ENVELOPING_MIST_BUFF.manaPercRed);
+      this.lifecycles.requiredEnvs = Math.ceil((this.best.manaFrom + Number(SPELLS.VIVIFY.manaCost ? SPELLS.VIVIFY.manaCost : 1)) / (SPELLS.ENVELOPING_MIST.manaCost ? SPELLS.ENVELOPING_MIST.manaCost : 1) * 1);
```
abelito75: This isn't accurate. The math is just wrong now.

src/parser/monk/mistweaver/modules/talents/Tier45Comparison.tsx
```diff
@@ -166,8 +166,8 @@ class Tier45Comparison extends Analyzer {
   //anaylze current play style and see how much mana they would have gained from this talent
   generateSotc() {
     const sotcBlackOutKicks = this.abilityTracker.getAbility(SPELLS.BLACKOUT_KICK_TOTM.id).damageHits || 0;
-    const manaPercentFromSotc = sotcBlackOutKicks * SPELLS.SPIRIT_OF_THE_CRANE_BUFF.manaRet;
-    const rawManaFromSotc = manaPercentFromSotc * this.manaTracker.maxResource;
+    const manaPercentFromSotc = 0.065 ;
```
abelito75: I know we want to remove .manaRet from the spell files but i don't think hard coding it here is the correct move. It should go in the constant file for mw

src/parser/monk/mistweaver/modules/talents/Tier45Comparison.tsx
```diff
@@ -177,7 +177,7 @@ class Tier45Comparison extends Analyzer {
   generateManaTea() {
     const fightLength = (this.owner.fight.end_time - this.owner.fight.start_time) / 1000;
     const manaTeasPossible = (Math.ceil(fightLength / 90) || 1);
-    const manaPerDuration = (this.totalManaSpent() / fightLength) * SPELLS.MANA_TEA_TALENT.duration / 1000;//duration of mana Tea
+    const manaPerDuration = (this.totalManaSpent() / fightLength) * (SPELLS.MANA_TEA_TALENT.duration ? SPELLS.MANA_TEA_TALENT.duration : 1) / 1000;//duration of mana Tea
```
abelito75: Would provide inaccurate info if mana tea's duration is ever not defined

src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
```diff
@@ -7878,6 +7878,29 @@ exports[`Mistweaver Monk integration test: example log Tier45Comparison matches
...
+      With your current playstyle you are not using the most effective tier 45 talent. 
+      <ForwardRef
+        id={210802}
+      />
+       is better based off of how you played.
+    </React.Fragment>,
+    "stat": <React.Fragment>
+      25k{"id":"monk.mistweaver.suggestions.tier45Talent.efficiency","message":"mana returned through"}Spirit of the Crane
+       (
+      Spirit of the Crane would have returned 98k
```
abelito75: this makes me believe something broke

src/parser/monk/mistweaver/modules/talents/Tier45Comparison.tsx
abelito75: as a secondary note its not .065 as that would be 6.5%. It should be .0065 as that would be the correct .65%
image

abelito75 (changes requested):
Thanks for the PR

src/common/ItemLink.tsx
CasperKjaerhus: Yeah, i was confused as well. Quality doesn't seem to do anything.. I have no idea why i didn't just set it equal to 0 though.

Dambroda:
I also had converted indexById to TS at one point, but I personally am of the opinion that converting it causes more problems than it's worth. It doesn't really add any useful typechecking (the casting you do could be added right now anyway, without converting indexById), and breaks some functionality that helps in development (at least for me). 

For example, right now if I'm working in a file that uses SPELLS.ARCANE_POWER, I can goto -> declaration and it actually takes me to where the constant is declared. After converting, the same operation instead takes me to indexById, which is completely useless to me. I don't know if there is a more elegant way to convert it that preserves this functionality or if it's something I would just have to adapt to. However, looking at this PR and all of the other bits of casting that are going on, I feel like perhaps rather than converting indexById and introducing more code smell we might be better served in redesigning how we declare and access spell constants. Curious to hear other thoughts.

On a completely separate point, please split up your PRs when dealing with multiple issues; small fixes get lost in huge 84 file PRs when we could have merged those fixes right away instead 😄

CasperKjaerhus:
>However, looking at this PR and all of the other bits of casting that are going on, I feel like perhaps rather than converting indexById and introducing more code smell we might be better served in redesigning how we declare and access spell constants. Curious to hear other thoughts.

> On a completely separate point, please split up your PRs when dealing with multiple issues; small fixes get lost in huge 84 file PRs when we could have merged those fixes right away instead 😄

I totally agree, redoing how spells are accessed is ultimately the correct way to go. IndexById is a hack that shouldn't exist, but it does.

However, as i stated in initial PR description, typing IndexById is exactly what lead me to discover the issues that also got fixed in this PR (which is also why i didn't make multiple PRs). Therefore i think typing this, for now, is the correct move since obviously the type errors are hard to discover when making changes to old code etc..

Pewtro:
I strongly disagree on this being the correct thing to do as it makes coding in modules quite obnoxious having to cast all these things every time for no real gain anywhere AND **not** being able to jump straight to definitions is a 100% gamebreaker for me. 

IMO, whatever issues you found (which issues even? Most of these changes seem either wrong or caused by your changes?) should be fixed in a separate PR and this PR closed because this typing should not happen as it is now - it should be a part of an eventual SPELLS rewrite some time far out in the future.

CasperKjaerhus:
> (which issues even? Most of these changes seem either wrong or caused by your changes?) 

#4188 and #4189 was discovered because of this.
You're right Mistweaver errors was cause by my own changes, and I'll be fixing those errors for this PR.

> I strongly disagree on this being the correct thing to do as it makes coding in modules quite obnoxious having to cast all these things every time for no real gain anywhere AND not being able to jump straight to definitions is a 100% gamebreaker for me.

You might be right, I'm not sure. I just think non-typed language introduces bugs that are very hard to track and requires the developer to be extra vigilant; error-prone. 

>  it should be a part of an eventual SPELLS rewrite some time far out in the future.

While I agree that a SPELLS/ITEMS rewrite is much more desireable, I don't think this it should be far out in the future.

abelito75:
>You might be right, I'm not sure. I just think non-typed language introduces bugs that are very hard to track and requires the developer to be extra vigilant; error-prone.

idk about other projects but the majority of issues in wowA are pretty easy to track down and the ones that are not don't come from typing issues from my experience

src/common/ItemLink.tsx
CasperKjaerhus: I ended up giving Item a quality? property

CasperKjaerhus:
Alright, so I backtracked a lot and changed how IndexById works and how i typed it. I am new to the codebase, and the previous comments on this PR really helped me understand how you work with the codebase itself, thank you! As of now, "go to declaration" within editor works as intended, along with giving access to data right in the editor. Anything indexById related now is typed by the declaration of the spells/items/etc themself. I think many of the previous gripes people had with the previous typed implementation of IndexById should now be gone, but, you're of course welcome to come with additional criticism :)



Since type context is needed I had to modify SafeMerge a tad. It now allows for setting the wanted return type. This does, in my opinion, look *VERY* ugly (screenshot below) and is not very intuitive. If someone has more experience in TS and has an idea on how to mitigate this, please feel free to do so. Maybe even removing safemerge, since right now it isn't really being used. However, I do think that this is more desirable than not having types.



Also very sorry with 199 files changed.

Changes:

General:
* Typed IndexById -> IndexById now returns only a new object with id indexers; This means that when IndexById is used, it will have to be merged into existing object in order for that object to be indexable by id
* Implemented the newer IndexById changes into where IndexById was used.
* Gave many arrays and/or object types
* Some articles had age old items that are now removed from ITEMS, i set the id to -1. Are these articles even used/available anymore??

This actually made me discover multiple bugs/Typing errors in code

* /parser/rogue/assassination/..../Duskwalkers_Patch:
	* Duskwalkers_Patch had no bonusId in LegendarySpell; Gave it undefined since no bonusId exists afaik
	* Corrected Misspelling of EXHILARTATION (was EXHILERTATION) 
* /parser/shaman/elemental/..../StaticDischarge
	* Corrected STATIC_DISCHARGE to STATIC_DISCHARGE_TALENT
* /parser/shaman/enhancement/..../Flametongue
	* Flametongue has apparently been changed and doesn't provide a buff anymore, This code just doesn't work and probably hasn't since Shadowlands release. I changed FLAMETONGUE_BUFF to FLAMETONGUE_ATTACK for now. It should be noted that it still doesn't do anything meaningful.
* /parser/shaman/enhancement/..../FlametongueRefresh
	* Updated SPELLS.FLAMETONGUE_BUFF to SPELLS.FLAMETONGUE_ATTACK
* /parser/Death Knight/Shared/.../Superstrain
	* Corrected Frost_Death_Knight (was Frost_Death_Knigh)

If this PR doesn't get accepted, above changes should get fixed in another

abelito75:
This PR really seems to be escaping scope quite quickly

PR #4186, opened 2020-12-17
[Outlaw Rogue] Between the eyes and dispatch fixes

FrozenReality:
Between the eyes no longer requires the Ruthless Precision buff to be active, instead it should be used whenever off cooldown, instead of dispatch.


src/parser/rogue/outlaw/modules/features/Checklist/Component.js
```diff
@@ -44,7 +44,7 @@ const OutlawRogueChecklist = ({ combatant, castEfficiency, thresholds }) => {
       </Rule>
       <Rule
         name="Use your finishers efficiently"
-        description={<>Your two damaging finishers should typically be used at maximum combo points. If you have <SpellLink id={SPELLS.RUTHLESS_PRECISION.id} /> active you should prioritize <SpellLink id={SPELLS.BETWEEN_THE_EYES.id} /> as your damaging finisher.</>}
+        description={<>Your two damage finishers should typically be used at maximum combo points. Whenever <SpellLink id={SPELLS.BETWEEN_THE_EYES.id} /> is off cooldown, you should always prioritize it over <SpellLink id={SPELLS.DISPATCH.id} />.</>}
       >
```
Pewtro, line 48: This should definitely say "damaging" for grammar.

src/parser/rogue/outlaw/modules/spells/BetweenTheEyesDamageTracker.js
```diff
@@ -9,13 +9,7 @@ class BetweenTheEyesDamageTracker extends FilteredDamageTracker {
...
-    // isOnCooldown returns true when this event is a BTE cast, but we want to keep those casts too
-    if (event.ability.guid !== SPELLS.BETWEEN_THE_EYES.id && this.spellUsable.isOnCooldown(SPELLS.BETWEEN_THE_EYES.id)) {
-      return false;
-    }
-
-    const hasRuthlessPrecision = this.selectedCombatant.hasBuff(SPELLS.RUTHLESS_PRECISION.id);
-    return hasRuthlessPrecision;
+    return !this.spellUsable.isOnCooldown(SPELLS.BETWEEN_THE_EYES.id);
   }
 }
```
Pewtro, line 14: This doesn't seem right - what are you trying to achieve here?

src/parser/rogue/outlaw/modules/spells/Dispatch.js
```diff
@@ -38,11 +38,11 @@ class Dispatch extends Analyzer {
 
   constructor(...args) {
     super(...args);
-    this.betweenTheEyesDamageTracker.subscribeInefficientCast([SPELLS.DISPATCH], (s) => `Between The Eyes should be prioritized as your spender during Ruthless Precision`);
+    this.betweenTheEyesDamageTracker.subscribeInefficientCast([SPELLS.DISPATCH], (s) => `Between The Eyes should be prioritized as your damage finisher.`);
   }
```
Pewtro, line 42: damaging finisher.

src/parser/rogue/outlaw/modules/spells/BetweenTheEyes.js
```diff
@@ -1,49 +0,0 @@
...
-import Analyzer from 'parser/core/Analyzer';
-
-import DamageTracker from 'parser/shared/modules/AbilityTracker';
-
-import { t } from '@lingui/macro';
-
-import BetweenTheEyesDamageTracker from './BetweenTheEyesDamageTracker';
-
-class BetweenTheEyes extends Analyzer {
-  get thresholds() {
```
Pewtro, line 15: Since you say that BTE should be prioritized over Dispatch in the other files - why is this file being removed instead of being reworked - surely it is still interesting to see BTE information on the checklist / some sort of statistic on this?

Pewtro (changes requested):

src/parser/rogue/outlaw/modules/spells/BetweenTheEyes.js
FrozenReality: I felt like anything I reworked here would just be the inverse of the dispatch information "You used BTE over Dispatch 95%" and would not add anything extra other than bloat.

src/parser/rogue/outlaw/modules/spells/BetweenTheEyesDamageTracker.js
FrozenReality: This is to track how many times dispatch was used when BTE was not on CD, in other words, whenever BTE should've been used. This is my first PR, so I'm still learning how the code base works, if this should be done in another way I'd be happy to hear it.

src/parser/rogue/outlaw/modules/spells/Dispatch.js
FrozenReality: Changed.

src/parser/rogue/outlaw/modules/features/Checklist/Component.js
FrozenReality: Changed.

src/parser/rogue/outlaw/modules/spells/BetweenTheEyesDamageTracker.js
Pewtro: The module is called Between The Eyes Damage tracker - surely it should not be tracking Dispatch casts?

tyndivelspaz: This should only be tracking Between the Eyes damage. If we want to track improper Dispatch casts we can do that in a separate module.

src/parser/rogue/outlaw/modules/spells/BetweenTheEyes.js
tyndivelspaz: We should probably keep this, as it is a learning tool, and help newcomer Outlaw Rogues with this knowledge. And make sure the Dispatch module conveys the same message.

Evolixe:
> Between the eyes no longer requires the Ruthless Precision buff to be active, instead it should be used whenever off cooldown, instead of dispatch.

I mean it kinda does though. BTE damage when it doesn't crit is very whack and absolutely not worth it compared to a Dispatch. Nevermind the fact that Dispatch can proc Count the Odds and BTE can't. Whats important is that you keep the BTE buff up. Not that you use it on CD without ruthless precision. Because you actually don't want to do that.

FrozenReality:
Seeing as I broke more than I fixed I'm declining the PR.

Issue #1133, opened 2018-01-03
Make a core version of Guardian Druid's AntiFillerSpam


MartijnHols:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/master/src/Parser/Druid/Guardian/Modules/Features/AntiFillerSpam.js

I reckon this would be a great module for most tank/dps specs and healers to a lesser extent. Extending it with a tab or maybe an indicator in the spell timeline would be even greater.

PR #4193, opened 2020-12-18
[Druid] Fix shadowlands guardian compiler errors

abbottmg:
This corrects two compiler errors affecting the druid/guardian module hierarchy:

1. The Frenzied Regeneration spell module was attempting to read a Combatant's trait information, which was recently removed
2. The `Abilities` loader module was not finding the guardian-specific `Ability` subtype, which adds properties necessary for the Anti-Filler Spam feature module to operate. I believe this compiler error only occurs in development mode.

The first error required simply removing the calculation of a defunct Legion trait's effect on healing, which removed the need to reference the missing trait accessor.

The second error required modifying the core `Abilities` `loadSpellbook` method to dynamically discover the `Ability` subtype to construct using the class's static `ABILITY_CLASS` property, rather than statically calling the `Ability` constructor. This `ABILITY_CLASS` property is overridden by the guardian `Abilities` object to point at [guardian-specific Ability](src/parser/druid/guardian/modules/Ability.js).

The comments noting a special requirement for copy-pasting the `Ability` constructor let me on a small wild goose chase (the duplicated constructor was never getting called because `loadSpellbook` wasn't looking for it). Once I did find the root issue, I noticed that the duplicate constructor was actually redundant, at least based on the current implementation of `propTypes` and `Ability._setProps`. As such, I removed the redundant override as well as the comments, and replaced the latter with a comment explaining how to override `propTypes` to ensure `_setProps` has the information it needs within a single call.

All that said, I am not particularly familiar with this codebase so I understand if you know of a reason why the copy-paste constructor pattern should remain.

### DIFF
--- a/src/CONTRIBUTORS.ts
+++ b/src/CONTRIBUTORS.ts
@@ -1065,6 +1065,7 @@ export const Xcessiv: Contributor = {
     link: 'https://worldofwarcraft.com/en-us/character/us/thrall/xcessiv',
   }],
 };
+
 export const Tora: Contributor = {
   nickname: "Tora",
   github: "RobinKa",
@@ -1077,4 +1078,19 @@ export const Tora: Contributor = {
   links: {
     "Website": "https://warlock.ai"
   },
+};
+
+export const Kettlepaw: Contributor = {
+  nickname: 'Kettlepaw',
+  github: 'abbottmg',
+  discord: 'abbott#2506',
+  mains: [{
+    name: 'Caeldrim',
+    spec: SPECS.GUARDIAN_DRUID,
+    link: 'https://worldofwarcraft.com/en-us/character/us/wyrmrest-accord/caeldrim',
+  }, {
+    name: 'Kettlepaw',
+    spec: SPECS.BREWMASTER_MONK,
+    link: 'https://worldofwarcraft.com/en-us/character/us/wyrmrest-accord/kettlepaw',
+  }],
 };
--- a/src/parser/core/modules/Abilities.ts
+++ b/src/parser/core/modules/Abilities.ts
@@ -81,7 +81,9 @@ class Abilities extends Module {
     this.loadSpellbook(this.spellbook());
   }
   loadSpellbook(spellbook: SpellbookAbility[]) {
-    this.abilities = spellbook.map(options => new Ability(this, options));
+    // Abilities subtypes may want to construct a particular subtype of Ability
+    const ability_class = (this.constructor as typeof Abilities).ABILITY_CLASS;
+    this.abilities = spellbook.map(options => new ability_class(this, options));
     this.activeAbilities = this.abilities.filter(ability => ability.enabled);
   }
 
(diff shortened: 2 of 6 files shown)


abbottmg:
I am interested in writing integration tests that cover these two errors, especially the first error, which will be a problem any time defunct expansion features are removed from the core modules. Both of these errors surfaced in their respective types' constructors, and I am not familiar enough with Typescript to know how to adapt that to the behavioral testing tool I see in use. I'd appreciate advice on this front.

abbottmg:
Oh, and I suppose I should also mention that I am tentatively interested in maintaining the guardian modules, although I won't claim to be an expert in the spec.

Pewtro:
Hey, thanks for your PR! Guardians really need the help 😅 

Guardian druid is **ancient** code when it comes to wowanalyzer so it probably has a lot of old coding practices that are now deprecated. Are you sure that the antifiller stuff should still be in there? No other tank uses it / have a need for it, so should guardians?

If it's unnecessary, rather it be deleted than we alter the core even if the changes seem fine as is.

abbottmg:
I can only say I personally did use the anti-filler feature in BFA (because again I'm not an expert and hit swipe too often while panicking) and that #1133 exists to consider porting the feature for use with other specs. I thought Brewmaster had a similar module in BFA as well, because I remember seeing I hit Tiger Palm too frequently.

But core would have to change in either case, IMHO. The core `Abilities` class does currently expose an `ABILITY_CLASS` property that seems intended for this specific use of discovering which `Ability` subtype to construct; it just wasn't using checking that property at the moment. If we decide the filler module should go away and no other module should expect to subclass `Ability` in this way, we should at least strip `ABILITY_CLASS` from core.

I don't consider either a very big change personally, but I'm obviously a newcomer to this codebase. I decided it was fair to do the dynamic type lookup in [`loadSpellbook`](src/parser/core/modules/Abilities.ts) because we're already doing the same thing in the related core `Ability._setProps` and it makes sense that the Ability-loader class might want to know how to load more specialized Ability types.

I don't know if other modules use different mechanisms to extend the data associated with an Ability. I would be happy to work on adapting the filler module to use this other mechanism to replace more deprecated functionality. **But** I would personally recommend doing that in a follow-up issue/PR, so the big scary bar associated with these execution errors can go away from the live site while I figure out how to make that change (which would include stripping out whatever bits of core we decide should be considered deprecated, and reverting this dynamic `ABILITY_CLASS` lookup).

**EDIT:** Oh and to clarify, part of the reason I want the scary bar gone is to hopefully encourage people to file more bug reports against Guardian. I don't see a whole lot of issues for work people would like to see on Guardian, and part of the reason I decided to go ahead and figure out the Frenzied Regen bug was because I noticed Earthwarden was showing 0 mitigated attacks and wasn't sure if that was because it was downstream of the broken FR module. Turns out that seems to be independently broken. That's next on my list to tackle, but I figured either way I should wait to publish a branch until I know whether I'm reproducing the new contributor entry separately or can rely on this one being in trunk lol

Pewtro:
Yeah the change itself is fine - it was more so to reflect on if it was still a relevant aspect of Guardian as it hasn't been touched in ages. If you fix the contributors file it should be good to go

abbottmg:
Merge conflict resolved

Pewtro:
Cheers! 🥇

Issue #3721, opened 2020-09-20
[Shadowlands][Mage] Legendaries


Sharrq:
Other related issues:
- #3657 Core Class Changes
- #3675 Conduits/Soulbinds

**Shared**
- [ ] [Discliplinary Command](https://shadowlands.wowhead.com/spell=327365/disciplinary-command) *Check number of procs to see if they are actually activating it or not*
- [ ] [Expanded Potential](https://shadowlands.wowhead.com/spell=327489/expanded-potential) *check number of times proc'd*
- [ ] [Grisly Icicle](https://shadowlands.wowhead.com/spell=333393/grisly-icicle) *probably dont need anything for this, but maybe damage contribution?*
- [ ] [Triune Ward](https://shadowlands.wowhead.com/spell=333373/triune-ward) *amount of damage absorbed?*

**Arcane**
- [x] [Arcane Bombardment](https://shadowlands.wowhead.com/spell=332892/arcane-bombardment) *damage contribution?*
- [x] [Arcane Harmony](https://shadowlands.wowhead.com/spell=332769/arcane-harmony) *Damage contribution?*
- [ ] [Siphon Storm](https://shadowlands.wowhead.com/spell=332928/siphon-storm) *Check to see if charges were wasted by using Evo at 3 or 4 charges?*
- [ ] [Temporal Warp](https://shadowlands.wowhead.com/spell=327351/temporal-warp) *check to see if they used their extra lust?*

**Fire**
- [x] [Fevered Incantation](https://shadowlands.wowhead.com/spell=333030/fevered-incantation) *Damage contribution?*
- [x] [Firestorm](https://shadowlands.wowhead.com/spell=333097/firestorm) *Number of procs, usage during the 5sec buff?*
- [ ] [Molten Skyfall](https://shadowlands.wowhead.com/spell=333167/molten-skyfall) *Count number of Meteor procs*
- [ ] [Sun King's Blessing](https://shadowlands.wowhead.com/spell=333313/sun-kings-blessing) *Check to make sure they are activating it as much as they can, check usage during the 6sec Combust, maybe check to make sure they are setting up for it properly? ... gonna need to ask TCs about this*

**Frost**
- [x] [Cold Front](https://shadowlands.wowhead.com/spell=327284/cold-front) *Count number of Frozen Orbs this provided*
[...]

PR #4197, opened 2020-12-20
[Mage/Core] CD Active Time, Glacial Fragments, Typos

Sharrq:
- Added support for Glacial Fragments legendary (Towards #3721 )
- Removed a suggestion from Rune of Power that suggested Incanters flow for players not using RoP properly.
- Added a module to Frost/Fire/Arcane to track the player's active time during Icy Veins/Combustion/Arcane Power
- Removed a getter from Firestorm that wasnt used
- Added the buff from Firestorm to Buffs
- Fixed a bug in CombustionSpellUsage that was marking Fireball casts that started before Combustion started but ended after Combustion started as being cast during Combustion
- Added the start/end channel for Shifting Power so it shows the full channel in the timeline (and doesnt count as downtime)
- Fixed an issue in FilteredActiveTime that was not correctly counting active time.


src/parser/core/Events.ts
```diff
@@ -216,6 +216,9 @@ export interface EndChannelEvent extends Event<EventType.EndChannel> {
   start: number;
   duration: number;
   beginChannel: BeginChannelEvent;
+  trigger?: {
+    timestamp: number;
+  };
 }
```
Pewtro, line 222: where are these event changes used? (including the added ones below)

abelito75: If this is only one field then why are we making a whole new json for it?

Sharrq: They are used in filtered active time.

There are more under trigger, I just only needed the one and I didn't add the rest

PR #4199, opened 2020-12-20
[Core] More Warning Messages

Sharrq:
Added warning messages for when Advanced Combat Logging is not enabled and if there are no dungeon/boss encounters in the log.

Advanced Combat Logging
image

No Encounters Found
image


MartijnHols (approved):
Nice, thanks!

PR #4202, opened 2020-12-20
Windfury Totem

joshinat0r:
image

Adds a statistic for extra damage by windfury.
Is tracking the time between melee hits, every melee hit within 200ms is counted as windfury proc.
Only works 100% reliable for specs that use one 2h only and account for parry haste.

Is only showing for specs that are whitelisted and if the totem had some uptime since we dont know what the group composition looked like.
Meaning:
- fury would never show up
- unholy would display the stat if the player had uptime on windfury, regardless of how many damage it added)
- mistweaver would display the stat if the player had uptime on windfury, regardless of how many damage it added)


MartijnHols:
Hey since this PR has been stale in draft mode for more than a month I am going to close it. Please open a new PR when you're ready to have it merged.

PR #4203, opened 2020-12-20
[Death Knight] Venthy Covenant statistic & superstrain damage fix

joshinat0r:
- Adds a statistic for the Venthyr covenant ability
image
image

- Adds damage gained by Superstrains RP part (by taking the average cost of DS/DC/FS)
- [Blood] Reactivated Ossuary and made the checklist use its efficiency instead of uptime
- [Blood] Added covenant abilities to the spellbook


Pewtro (approved):
I hate the superstrain eventlisteners but can't see many other possibilities than the way you did it if it is to be in a shared module - looks g2g!

PR #4205, opened 2020-12-21
Fix IDs and cooldowns changed for Shadow Priest in 9.0.2

DylanDirlam:
Minor changes that were made after pre-patch when this stuff was last updated. Cooldowns and spell IDs


src/parser/priest/shadow/modules/Abilities.tsx
```diff
@@ -360,8 +359,14 @@ class Abilities extends CoreAbilities {
...
         category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
-        cooldown: 100,
+        cooldown: 60,
         enabled: combatant.hasCovenant(COVENANTS.NECROLORD.id),
+        castEfficiency: {
+          suggestion: true,
+          recommendedEfficiency: 0.8,
+          averageIssueEfficiency: 0.7,
+          majorIssueEfficiency: 0.6,
+        },
```
Pewtro, line 369: i didn't know we still had these and not purely the "recommendedEfficiency" one, but if it works it works!

PR #4211, opened 2020-12-22
[Core] Add Darkmoon Deck: Voracity & Power Infusion to Haste

Pewtro:
Also fixes/cleans some random thingies that I found

image


src/parser/shared/modules/items/shadowlands/crafted/DarkmoonDeckVoracity.tsx
```diff
@@ -0,0 +1,214 @@
...
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE()}
+        size="flexible"
+        category={STATISTIC_CATEGORY.ITEMS}
+        dropdown={
+          <table className="table table-condensed">
+            <thead>
+              <tr>
+                <th>Name</th>
```
abelito75: I feel like name is weird here. Wouldn't card be a better name

Pewtro: Sure

abelito75 (commented):
Looks good but i will say there is a lot of other random stuff. Probably could have a better PR name or split it up next time

PR #4213, opened 2020-12-22
Making types accessible throughout codebase

CasperKjaerhus:
Before src\common\indexById transformed the input object and returned the same object, but mutated to be indexable by number. Now, it returns a new object that is also indexable by number, without mutating the input object. Along with that src\common\safeMerge was changed to retain the type given to it through generics.

The responsibility of merging the returned object from indexById with the original object now relies on the caller instead of the function itself.

Now, what does all this change mean? This means that the typedata can be made available all the way through the codebase. For example, its possible to see exactly what data is on every property of SPELLS, right in the editor, whereas compared to before, spells would be given 'any' type. Also Typeerrors will be caught during typecheck. **big**

It should be noted that above is also correct for specs, covenants, soulbinds, pets and items!

This is a rather large PR due to the amount of typing i had to do for it to compile. Which is a good thing, since multiple type errors was discovered. These all _probably_ worked at some point, but because of changes to other parts of the codebase, during past commits, which wasn't typed, these errors were never thrown and were silent.

TLDR; spells/items/pets/covenants/soulbinds are now accessible throughout codebase!

Type-error fixes/changes:
* src\parser\deathknight\shared\covenants\SwarmingMist.tsx
	* SPELLS.FROST_STRIKE does not exist (anymore?) Changed to FROST_STRIKE_CAST (Hopefully this is correct; someone with more knowledge about the spec please check)
* src\parser\monk\mistweaver\modules\talents\Tier30Comparison.tsx
	* Changed SPELLS.LIFECYCLES_ENVELOPING_MIST_BUFF.manaPercRed -> LIFECYCLES_MANA_PERC_REDUCTION
* src\parser\monk\windwalker\modules\covenants\WeaponsOfOrder.tsx
	* Fixed hasBuff call parameter from spell to the spells id. (hasBuff doesn't take type spell)
	* ^ twice
* src\parser\rogue\assassination\modules\spells\shadowlands\legendaries\DuskwalkersPatch.tsx
	* Asserts that DUSKWALKERS_PATCH.bonusID is not undefined (which it is 100% right now)
	* Fixed spelling of EXHILARATION (btw these errors will type error from now on and are VERY easy to catch)
* src\parser\shaman\enhancement\modules\core\Flametongue.tsx
	* Changed SPELLS.FLAMETONGUE_BUFF -> SPELLS.FLAMETONGUE_ATTACK since SPELLS.FLAMETONGUE_BUFF does not exist anymore
	* Changed SPELLS.FLAMETONGUE -> SPELLS.FLAMETONGUE_ATTACK since SPELLS.FLAMETONGUE does not exist anymore
* src\parser\shaman\enhancement\modules\core\FlametongueRefresh.tsx
	* Same as above
* src\parser\hunter\survival\constants.tsx
	* SURVIVAL_CHAKRAM_TYPES is now an array of spells since it is used like one (incorrectly?) in other files
* src\parser\shaman\elemental\modules\talents\StaticDischarge.tsx
	* Changed SPELLS.STATIC_DISCHARGE -> to SPELLS.STATIC_DISCHARGE_TALENT since apparently it was moved to talent sometime

Other changes are basicly just "gave type to x" and can be seen in commit c8d93fc


src/articles/2018-01-31-1st-Anniversary/index.tsx
```diff
@@ -138,7 +137,7 @@ const Article = (props: ArticleProps) => {
 
           <Image source={v031} description="Holy Paladin mastery effectiveness calculator v0.3.1 statistics at 18 Mar 2017" /><br />
 
-          The implementation of <ItemLink id={ITEMS.DRAPE_OF_SHAME.id} /> included a large part of the work needed for adding items, so it was possible to quickly add statistics for the similar legendaries Ilterendi, Crown Jewel of Silvermoon and Velen's Future Sight in the next few days.
+          The implementation of <ItemLink id={-1} /> included a large part of the work needed for adding items, so it was possible to quickly add statistics for the similar legendaries Ilterendi, Crown Jewel of Silvermoon and Velen's Future Sight in the next few days.
```
MartijnHols:
```suggestion
          The implementation of Drape of Shame included a large part of the work needed for adding items, so it was possible to quickly add statistics for the similar legendaries Ilterendi, Crown Jewel of Silvermoon and Velen's Future Sight in the next few days.
```

In articles for now we just swap in the name when necessary.

MartijnHols:
Hey since this PR has been stale in draft mode for more than a month I am going to close it. Please open a new PR when you're ready to have it merged.

PR #4221, opened 2020-12-23
Add Dark Thoughts module for Shadow Priest

DylanDirlam:
# Dark Thoughts
* Passive: "For each of your Shadow damage over time effects on the target, your Mind Flay and Mind Sear have a 3% chance to trigger a Dark Thought."
* Dark Thought: "Increases the number of charges of Mind Blast by 1, Mind Blast can be cast instantly, and can be cast while channeling Mind Flay or Mind Sear."

The module tracks the number of procs you got during the fight and amount used. Losing a proc of this is pretty dumb during a fight because it's instant and can be cast during filler channels.

image
image


src/parser/priest/shadow/modules/features/DarkThoughts.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  onBuffRemoved() {
+    if (!this.eventHistory.last(1, 100, Events.cast.by(SELECTED_PLAYER).spell(SPELLS.MIND_BLAST))) { // If MB is not instant, it's not a proc
+      return;
+    }
+
+    this.procsUsed += 1;
+  }
+
+  getProcsWasted() {
```
Sharrq: i assume this should be 

get procsWasted() {
    return this.procsGained - this.procsUsed;
}

and then this.procsWasted / this.procsGained in actual under the suggestion threshold

src/parser/priest/shadow/modules/features/DarkThoughts.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  suggestions(when: When) {
+    when(this.suggestionThresholds)
+      .addSuggestion((suggest, actual, recommended) => suggest(<>You wasted {formatPercentage(actual)}% of <SpellLink id={SPELLS.DARK_THOUGHTS.id} /> procs. </>)
+          .icon(SPELLS.DARK_THOUGHTS.icon)
+          .actual(
+            t({
+              id:'priest.shadow.suggestions.darkThoughts.efficiency',
+              message: `Wasted ${formatPercentage(actual)}% of Dark Thought procs. Remember that the proc allows you to cast Mind Blast instantly during Mind Flay, Mind Sear, or while moving and should never be left unused.`
```
Sharrq: Your translated suggestion has more info than the actual suggestion text. Might want to add that second sentence to the actual suggestion above there.

Keep in mind that each suggestion should aim to explain what you did, what you should have done, and why.

Sharrq (changes requested):

src/parser/priest/shadow/modules/features/DarkThoughts.tsx
DylanDirlam: Thanks, I hadn't really ever known what was required for a suggestion so good to know. 👍

DylanDirlam: Rewrote the suggestion based on your recommendations. Also switched over to using actual values of procs instead of percentage because it's more relevant than a random %.

image

src/parser/priest/shadow/modules/features/DarkThoughts.tsx
```diff
@@ -0,0 +1,87 @@
...
+import EventHistory from 'parser/shared/modules/EventHistory';
+
+// Example log: /report/JAPL1zpDfN7W8wck/33-Heroic+The+Council+of+Blood+-+Kill+(5:46)/Mayrim/standard/statistics
+class DarkThoughts extends Analyzer {
+  static dependencies = {
+    eventHistory: EventHistory,
+  };
+  protected eventHistory!: EventHistory;
+
+  procsGained: number = 0;
```
Sharrq: you dont need to specify that they are numbers if you are assigning a value to it immediately, it will figure that out itself

Sharrq (changes requested):
one minor change and then its good to go.

PR #4241, opened 2020-12-23
Add Cooldowns section for Shadow Priest

DylanDirlam:
# Cooldowns for Shadow Priest
* Pre-Shadowlands, cooldowns for priest weren't really a thing. Now in SL there are Power Infusion and Void Eruption as 2 main real cooldowns. I have added them to the standard Cooldown section built into WoWA.
* I also added Boon of the Ascended (Kyrian covenant ability) which is a 3 min CD which lasts 10 seconds and changes up the rotation during that time.
* Reviewed with theorycrafter from Warcraft Priests discord to confirm that these were all CDs that should be displayed


Sharrq (approved):
Fix conflicts and it should be fine

DylanDirlam:
Conflicts fixed!

PR #4242, opened 2020-12-23
Havoc talents and spells update for SL

philipsabri:
This is my first PR so I appreciate feedback.

Changelog:
- Added Glaive Tempest
- Dark Slash becomes Essense Break
- Fel Mastery removed
- Nemesis removed

Known bugs:
~~Essence Break not showing damage in statistics.~~
image

Thanks to [Dadwald](https://worldofwarcraft.com/en-us/character/us/area-52/dadwald) for helping me with some test data for Essence Break


Pewtro:
Definitely fix that bug in this PR.

src/common/SPELLS/demonhunter.ts
```diff
@@ -158,6 +158,11 @@ const spells = {
     name: 'Immolation Aura First Strike',
     icon: 'ability_demonhunter_immolation',
   },
+  IMMOLATION_AURA_SECOND_STRIKE: {
```
philipsabri: Im a little bit confused what to name this spell when looking att line 151 and 156

philipsabri (commented):
> Definitely fix that bug in this PR.

Fixed

src/common/SPELLS/demonhunter.ts
Pewtro: Give them a name that fits what they do, I don't think first and second strike fit a proper descriptive name for the abilities.
It could be IMMOLATION_AURA_FIRST_HIT and IMMOLATION_AURA_DAMAGE or whatever those IDs actually do. 

First strike sounds like some effect from a talent or trait or whatever.

philipsabri: Thanks for feedback! Changed the names and now I understand what both these spells actually do. I also noticed it didnt show up properly in Statistic - Talents. Also fixed that.

src/common/SPELLS/demonhunter.ts
```diff
@@ -153,11 +153,16 @@ const spells = {
     name: 'Immolation Aura',
     icon: 'ability_demonhunter_immolation',
   },
-  IMMOLATION_AURA_FIRST_STRIKE: {
+  IMMOLATION_AURA_FIRST_STRIKE_DPS: {
```
philipsabri: Used to be "IMMOLATION_AURA_FIRST_STRIKE_DPS" when looking at old comments, fits more that way IMO.

src/common/SPELLS/demonhunter.ts
```diff
@@ -153,11 +153,16 @@ const spells = {
...
     id: 258921,
     name: 'Immolation Aura First Strike',
     icon: 'ability_demonhunter_immolation',
   },
+  IMMOLATION_AURA_BUFF_DPS: {
```
philipsabri: Followed the pattern on line 156, and also noticed old comments that used this name.

src/common/SPELLS/demonhunter.ts
Pewtro: What does "First Strike" mean here? Is it the damage on use, or?

src/parser/demonhunter/havoc/CHANGELOG.js
```diff
@@ -1,6 +1,7 @@
-import { LeoZhekov } from 'CONTRIBUTORS';
+import { LeoZhekov, flurreN } from 'CONTRIBUTORS';
 import { change, date } from 'common/changelog';
 
 export default [
   change(date(2020, 10, 30), 'Updated the deprecated StatisticBox elements with the new Statistic ones.', [LeoZhekov]),
+  change(date(2020, 12, 23), 'Updated spells and talents for SL', [flurreN]),
 ];
```
Pewtro, line 7: Put this on top of the other changelog entry

src/parser/demonhunter/havoc/modules/Abilities.js
```diff
@@ -211,6 +211,20 @@ class Abilities extends CoreAbilities {
...
+        cooldown: 20,
+        gcd: {
+          base: 1500,
+        },
+        castEfficiency: {
+          suggestion: true,
+          recommendedEfficiency: 0.95,
+          extraSuggestion: `This is a great AoE damage spell, but also does a great damage on single target. You should cast it as soon as it gets off cooldown. The only moment you can delay it's cast is if you already expect an add wave to maximize it's efficiency and damage output.`,
+        },
+      },
```
Pewtro, line 227: If it does great damage in both AoE and Single-Target, probably no need to mention it. I'd write something like the below to keep it short and simple.
```
extraSuggestion: `The only time you should delay casting <SpellLink id={SPELLS.GLAIVE_TEMPEST_TALENT.id} /> is when you're expecting adds to spawn soon.",
```

src/common/SPELLS/demonhunter.ts
philipsabri: This is the first stike (aoe dmg), it does directly when you activate.
image

-> this is the buff which does aoe dmg after the strike
image

Pewtro: I feel like IMMOLATION_AURA_INITIAL_HIT is more telling, and use DAMAGE instead of DPS imo.

philipsabri: Hit implies that two things are making contact, I would say this is more a strike. BUT you could also say a strike also need to make contact to something.
I agree on the DAMAGE instead of DPS tho.

yajinni: I want to add to this, I believe it was called dps because both havoc and veng could have this spell but the first strike spell Id was different based on spec. Something you want to check. Probably should have been called havoc instead of dps.

src/common/SPELLS/demonhunter.ts
yajinni: Here the dps once again refered to the fact the spell Id is deferent for havoc vs veng. If this continues to be the case then yeah change to to havoc instead of dps.

src/common/SPELLS/demonhunter.ts
philipsabri: Looked up what they actually do and we dont have to seperate them with _HAVOC and _VENGENCE anymore. Only the last one for vengence.

_From combat log_
HAVOC
**258920 - first cast
258921 - initial hit
258922 - buff aoe**

VENGENCE
**258920 - first cast
258921 - initial hit
258922 - buff aoe**
320334 - (talent)infernal armor, extra dmg vs melee

We could go with your initial_hit,
```
IMMOLATION_AURA
IMMOLATION_AURA_INITIAL_HIT
IMMOLATION_AURA_BUFF
```
Maybe
```
IMMOLATION_AURA
IMMOLATION_AURA_INITIAL_HIT
IMMOLATION_AURA_DAMAGE 
```
?

Do we still add _DAMAGE on the initial and buff?

yajinni: I would prefer the first one with _damage

philipsabri: Changed it to this
```
IMMOLATION_AURA
IMMOLATION_AURA_INITIAL_HIT_DAMAGE
IMMOLATION_AURA_BUFF_DAMAGE
```

PR #4246, opened 2020-12-24
Add insanity statistics to Shadow Priest modules

DylanDirlam:
* Insanity is a much more important resource to not lose out on in Shadowlands.
* Adds insanity statistics to the talents to make it easier for the person to compare how much insanity talents are generating and where they might be losing some.

### Before:
image

### After:
image

### New: Added suggestion if you are overcapping insanity. 
Note that I had to do this by calculating the time spent channeling compared to the insanity generation rate while channeling as the wasted amount coming from the Energize Event was always showing 0 no matter what.
image

### Before:
image

### After:
image


src/parser/priest/shadow/modules/talents/ShadowCrash.tsx
```diff
@@ -2,39 +2,47 @@ import React from 'react';
...
+import Insanity from 'interface/icons/Insanity'
 
 // Example Log: /report/zgBQ3kr6aAv19MXq/22-Normal+Zul+-+Kill+(2:26)/3-Selur
 class ShadowCrash extends Analyzer {
   static dependencies = {
     abilityTracker: AbilityTracker,
   };
+
   damage: number = 0;
+  insanityGained: number = 0;
```
Sharrq: dont need to specify that these are numbers if you are assigning a value to it immediately. it will figure that out itself.

src/parser/priest/shadow/modules/talents/ShadowCrash.tsx
```diff
@@ -2,39 +2,47 @@ import React from 'react';
...
   }
 
   get averageTargetsHit() {
     return this.totalTargetsHit / this.abilityTracker.getAbility(SPELLS.SHADOW_CRASH_TALENT.id).casts;
   }
 
-  onShadowCrashDamage(event: DamageEvent) {
+  onDamage(event: DamageEvent) {
     this.totalTargetsHit += 1;
     this.damage += event.amount;
```
Sharrq: make sure you are including absorbed damage as well

`this.damage += event.amount + {event.absorb || 0);`

src/parser/priest/shadow/modules/talents/VoidTorrent.tsx
```diff
@@ -37,33 +42,56 @@ class VoidTorrent extends Analyzer {
...
+
+  get insanityOvercapped() {
+    // Can't use the Energize event's waste because it shows 0 for some reason even if you overcap, so we calculate based on the time channeling vs. how much sanity you should've gotten from it
+    return ((this.totalChannelingTime / 1000) * VOID_TORRENT_INSANITY_PER_SECOND) - this.insanityGained;
+  }
+
+  get interruptThreshold() {
     return {
-      actual: this.totalWasted,
+      actual: this.timeWasted,
```
Sharrq: So for this threshold and the one below it. You should try to avoid using thresholds based on the sum of something over the course of the fight whenever possible.

What i mean by that, is say you make 5 mistakes in a fight, dont set the thresholds to 5 mistakes, 10 mistakes, 15 mistakes, instead use something like a percentage of mistakes, or mistakes per minute, or something like that.

the reason is because on a short fight, it will be a lot harder to trigger the advice, and on a long fight it will be a lot easier to trigger it. 

That said, some things are like .. this should NEVER happen and then it is ok, but if its something like it can happen but minimize it ... then you want to use different thresholds.

Sharrq (changes requested):

src/parser/priest/shadow/modules/talents/VoidTorrent.tsx
DylanDirlam: That makes sense, the values sounded good on paper but a 3 min fight vs 10 min fight would make a huge difference. Have updated it to use percentage calculations.

src/parser/priest/shadow/modules/talents/ShadowCrash.tsx
```diff
@@ -2,37 +2,45 @@ import React from 'react';
...
 
   get averageTargetsHit() {
     return this.totalTargetsHit / this.abilityTracker.getAbility(SPELLS.SHADOW_CRASH_TALENT.id).casts;
   }
 
-  onShadowCrashDamage(event: DamageEvent) {
+  onDamage(event: DamageEvent) {
     this.totalTargetsHit += 1;
-    this.damage += event.amount;
+    this.damage += event.amount + (event.absorbed || 0);
```
Sharrq, line 39: sorry my bad. its event.absorb, not absorbed.

src/parser/priest/shadow/modules/talents/VoidTorrent.tsx
```diff
@@ -4,31 +4,36 @@ import SPELLS from 'common/SPELLS/index';
...
   return Math.round(seconds * 10) / 10;
 }
 
-// Example Log: /report/hmJqLPZ7GVgY1CNa/16-Normal+Fetid+Devourer+-+Kill+(1:52)/44-베시잉/events
+// Example Log: /report/jcHgC7bfNAMqtRFv/12-Mythic+Hungering+Destroyer+-+Kill+(6:41)/Adoraci/standard/statistics
 class VoidTorrent extends Analyzer {
 
   _previousVoidTorrentCast: any;
-  damage = 0;
+  damage: number = 0;
```
Sharrq: ditch the :number

src/parser/priest/shadow/modules/talents/VoidTorrent.tsx
```diff
@@ -73,31 +109,51 @@ class VoidTorrent extends Analyzer {
     this._previousVoidTorrentCast = null;
   }
 
-  onVoidTorrentDamage(event: DamageEvent) {
+  onDamage(event: DamageEvent) {
     this.damage += event.amount || 0;
```
Sharrq: `this.damage += event.amount + (event.absorb || 0);`

Sharrq (changes requested):
just a couple more things. Some of these were in there before you made changes, but might as well fix them while we are here

src/parser/priest/shadow/modules/talents/ShadowCrash.tsx
DylanDirlam: Is it not absorbed? Looking at other damage events where they have the same logic, it uses `event.absorbed`

src/parser/priest/shadow/modules/talents/VoidTorrent.tsx
DylanDirlam: Updated this one with `absorbed` as well like above, if the `absorb` is indeed correct then I can switch it over to that, but based on searching around `absorbed` looks correct.

Sharrq: hmmm ... i was looking at some damage events recently and the property was called absorb. ill double check ... maybe something changed

Sharrq: Its absorb
image

src/parser/priest/shadow/modules/talents/ShadowCrash.tsx
Sharrq: yea im wrong. ignore me

DylanDirlam:
@Sharrq Merge conflicts are fixed again and ready to go

Sharrq (approved):

PR #4250, opened 2020-12-24
Druid Dark Titan's Lesson

vexxra:
Lifeblooms with DTL have their own spell id and therefore are currently not tracked at all. Having them on multiple targets also messes with some stuff. Fixed everything but the Photosynthesis tracker, that gets a bit more complicated.


src/common/SPELLS/druid.ts
```diff
@@ -203,6 +203,12 @@ const spells = {
     icon: 'inv_misc_herb_felblossom',
     manaCost: 800,
   },
+  LIFEBLOOM_DTL_HOT_HEAL: {
```
abelito75: This should go in the legendary file

src/parser/druid/restoration/modules/features/Lifebloom.js
```diff
@@ -16,7 +16,8 @@ import { t } from '@lingui/macro';
 
 class Lifebloom extends Analyzer {
   get uptime() {
-    return this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id);
+    // DTL Lifeblooms are on two targets, getBuffUptime sums both up
+    return this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id) + (this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_DTL_HOT_HEAL.id) / 2);
```
abelito75: couldn't this make uptime 200% which wouldn't make any sense?

abelito75 (changes requested):

src/parser/druid/restoration/modules/features/Lifebloom.js
vexxra: I could've reformatted this. Both DTL hots are LIFEBLOOM_DTL_HOT_HEAL, that's why I divided it by 2.
With DTL the first value should be 0, without DTL the second one.

src/parser/druid/restoration/modules/Abilities.js
```diff
@@ -182,6 +182,16 @@ class Abilities extends CoreAbilities {
           SPELLS.LIFEBLOOM_BLOOM_HEAL.id,
         ],
       },
+      {
+        spell: SPELLS.LIFEBLOOM_DTL_HOT_HEAL,
```
abelito75, line 187: Just thought about it but you need to check if they have DTL on here and on normal life bloom or you get this 
image

src/parser/druid/restoration/modules/features/Lifebloom.js
```diff
@@ -16,7 +16,10 @@ import { t } from '@lingui/macro';
 
 class Lifebloom extends Analyzer {
   get uptime() {
-    return this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id);
+    // Only either LIFEBLOOM_HOT_HEAL or LIFEBLOOM_DTL_HOT_HEAL can be up (with or without the DTL legendary), but
+    // DTL Lifeblooms (LIFEBLOOM_DTL_HOT_HEAL) are on two targets so their BuffUptime need to behalved for a percentage
+    return this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_HOT_HEAL.id)
+      + (this.combatants.getBuffUptime(SPELLS.LIFEBLOOM_DTL_HOT_HEAL.id) / 2);
```
abelito75: You don't need to divide by two here. combatants.getBuffUptime gives you what you want even if two are out at the same time.

abelito75 (changes requested):

src/parser/druid/restoration/modules/features/Lifebloom.js
vexxra: checked the wrong getBuffUptime function... fixed

PR #4258, opened 2020-12-25
Allow Echoing Shock to duplicate Earthquake

HawkCorrigan:
Echoing Shock should always be spend on either Lava Burst or Earthquake.

Original implementation only allowed Lava Burst.


abelito75 (approved):

src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
```diff
@@ -41,7 +41,7 @@ class EchoingShock extends Analyzer {
 
     this.casts += 1;
 
-    if (event.ability.guid !== SPELLS.LAVA_BURST.id) {
+    if (event.ability.guid !== SPELLS.LAVA_BURST.id && event.ability.guid !== SPELLS.EARTHQUAKE.id) {
```
abelito75, line 44: Im pretty sure this needs to be || not &&

abelito75 (changes requested):

src/parser/shaman/elemental/modules/talents/EchoingShock.tsx
Dambroda: With `||`, every cast would be a bad cast :D `&&` is correct

Dambroda (approved):

PR #4274, opened 2020-12-28
[Core] Fix the changelog dropdown displaying [object Object]

DylanDirlam:
The `<select>` tag does not allow rendering of React components, so can't use the `<Trans>` component in here. Looking at other people having the same issue, using the direct `i18n._` call to translate works as it renders the text alone. [Reference thread](https://github.com/lingui/js-lingui/issues/655)

The issue is occurring under https://wowanalyzer.com/about

## Before
image

## After
image

### DIFF
--- a/src/interface/ChangelogPanel/ChangelogPanel.tsx
+++ b/src/interface/ChangelogPanel/ChangelogPanel.tsx
@@ -3,6 +3,7 @@ import React, { useState } from 'react';
 import CORE_CHANGELOG from 'CHANGELOG';
 import AVAILABLE_CONFIGS from 'parser';
 import { Trans } from '@lingui/macro';
+import { i18n } from '@lingui/core';
 
 import Changelog from '../Changelog';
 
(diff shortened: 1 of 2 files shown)


src/interface/ChangelogPanel/ChangelogPanel.tsx
```diff
@@ -26,7 +27,7 @@ const ChangelogPanel = () => {
           onChange={(e) => setChangelogType(Number(e.target.value))}
         >
           <option value={0}>
-            <Trans id="interface.changelogPanel.option.core">Core</Trans>
+            {i18n._('interface.changelogPanel.option.core')}
```
MartijnHols: Since Lingui v3 i18n._ has been replaced with the `t` macro. It is used all over the codebase for an example.

DylanDirlam: Ah, yeah I have seen that all over, didn't realize it was the same thing. Updated!

MartijnHols (approved):
Thanks for the fix!

PR #4279, opened 2020-12-28
[BrM] Add support for Chi Burst & Chi Wave

### LINKED ISSUE #4180: [BrM] Fix Chi Wave cooldown
Some people are actually picking the talent now. It isn't properly configured in the BrM abilities file.

JoeyBG:
Should fix #4180 too.

Hi! I was thinking of looking at warlocks next, but I'd rather start small :)

#### Results

Changelog
image

Checklist
image
image

Suggestions
image
image


JoeyBG:
When looking for sample reports, I found an issue with [this log](https://wowanalyzer.com/report/gjAXYRzm4cMqtFrJ/9-Mythic+Huntsman+Altimor+-+Kill+(5:32)/Sinzhu/standard).

I'll try to see if I can figure out what is crashing, but it's not related to what I changed (considering it's happening in prod as well)

src/parser/monk/brewmaster/modules/Abilities.ts
```diff
@@ -56,6 +56,30 @@ class Abilities extends CoreAbilities {
...
+      {
+        spell: SPELLS.CHI_WAVE_TALENT,
+        category: Abilities.SPELL_CATEGORIES.ROTATIONAL,
+        cooldown: 15,
+        enabled: combatant.hasTalent(SPELLS.CHI_WAVE_TALENT.id),
+        castEfficiency: {
+          suggestion: true,
+        },
+        gcd: {
+          base: 1500,
```
emallson: chi wave has a 1s hasted GCD AFAIK

emallson (changes requested):

src/parser/monk/brewmaster/modules/Abilities.ts
JoeyBG: I copied the values from MW/WW. Do you know if it should be the same for these specs?

JoeyBG: I just finished some basic testing in-game. Seems like it's 1.5 for MW, and 1.0 for BM/WW. Good catch!

emallson (approved):
The GCD for CW *might* be 1000 hasted (not static) but this is close enough tbh. I haven't rechecked whether its hasted since Sinzhu posted about it in early Nov.

PR #4283, opened 2020-12-28
Fixes #4220

### LINKED ISSUE #4220: Page Crash
https://wowanalyzer.com/report/P7ZpbvM6gxWCDaRJ/2-Normal+Shriekwing+-+Kill+(6:12)
Click on Annei and the webpage and big page error pops up.

Scotsoo:


Pewtro:
Feel like this issue should be fixed in WeaponEnchantChecker where the error is instead of editing PerformanceThresholds as a whole.

Scotsoo:
> Feel like this issue should be fixed in WeaponEnchantChecker where the error is instead of editing PerformanceThresholds as a whole.

I can do, but overall it's wrong to treat an unknown as a boolean value. So if style is number, then we should check this as a number?

emallson (commented):
while this may technically fix the crash in #4220, i don't believe that this is the approach that we want to take. throwing an error there is intentional. once on production, we should not be getting NaNs in thresholds. this approach means that we will not get notifications about possible analysis errors.

instead of masking the NaN, we should fix the broken code that is generating it.

Sharrq:
I was able to fix this locally, so if you havent already fixed it in WeaponEnhancementChecker, then i can PR mine.

PR #4284, opened 2020-12-29
[Demon Hunter - Vengeance] Some more modernization

ryanadean:
## What this does
- Updates Soul Cleave reporting
- Updates resource tracking to use Fury instead of Pain
- Updated Sigil of Flames display for Kyrian Covenants
- Updated Demon Spikes triggers
- Implemented (but left disabled) an Infernal Strikes module (see below)

## What's with the Infernal Strikes module?
During implementation and testing of this module, I noticed that _even in the direct WoW combat log_ Infernal Strikes has several odditiies.

- It DOES NOT have successful cast events
- It DOES NOT have failed cast events for "missing" (aka not damaging any targets/using for mobility)
- It DOES have damage events
- It DOES have failed cast events for things like "can't find a pathway" or "not yet available"

A bug report was filed with Blizzard and jPutro also may have tugged at some strings to see if we can't get this fixed up.


ryanadean:
Fixed various minor branch issues from moving from one repo to another. Should be good for review now.

PR #4291, opened 2020-12-29
[Castle Nathria] Soul infusion icon fix and channeld in Timeline

philipsabri:
Before:
image

Now:
image

Adding the spell to SPELLS did not fix the icon bug, there is a whole discussion about it in discord, but this is how it looks when I comment it out from BAD_ICONS to prove it.
image


philipsabri:
I had to update the test with these commands, if I did something wrong I will fix it asap
`yarn test`
`u`

src/parser/shared/modules/spells/SoulInfusion.tsx
```diff
@@ -0,0 +1,43 @@
+import SPELLS from 'common/SPELLS';
+import CoreChanneling from 'parser/shared/modules/Channeling';
+import Events, {RemoveBuffEvent, CastEvent} from 'parser/core/Events';
+import { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+
+class SoulInfusion extends CoreChanneling {
+
+  constructor(options: Options) {
+    super(options);
+    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER), this.onRemoveBuff);
```
Khazakdk:
```suggestion
    this.addEventListener(Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.SOUL_INFUSION), this.onRemoveBuff);
```

You can do this to filter the events sent to the listener and remove the need to check the spell id first thing in the listener

Khazakdk (dismissed):
The snapshot changes here are concerning.  The downtime statistic changes drastically for several specs, which shouldn't be happening unless they all happen to use SKS for their snapshots but that shouldn't be the case given our disclaimer on SKS.

src/parser/shared/modules/spells/SoulInfusion.tsx
Sharrq: pretty sure i used Hungering/Shriek/etc for the mage ones. and it seems like its removing an entire section from the mage ones?

Sharrq: yea i used Shriek for all 3 mage specs. so those snapshots should not be getting changed at all

philipsabri: Rewrote my bad code @Khazakdk 

I have no idea about the tests tho, reverted them incase I should add or change something @Sharrq

Sharrq: So you have to do the tests, cause the PR wont pass without them

The problem is something in your changes is causing other spec's snapshots to change, and it probably shouldnt.

Sharrq:
Much better.

FYI @Pewtro your hunter tests are getting updated in here cause they are based on Sun King for Multi Target. not sure if you wanted to check to make sure its ok.

The rest of it looks fine to me

src/parser/shared/modules/spells/SoulInfusion.tsx
philipsabri: Should be better now!

src/parser/shared/modules/spells/SoulInfusion.tsx
```diff
@@ -0,0 +1,35 @@
...
+/* Soul Infusion for Sun king's Salvation */
+/* Example Data: https://Wowanalyzer.com/report/g4Pja6pLHnmQtbvk/32-Normal+Sun+King's+Salvation+-+Kill+(10:14)/Pjurbo/standard/timeline */
+
+class SoulInfusion extends CoreChanneling {
+
+  constructor(options: Options) {
+    super(options);
+    
+    const boss = this.owner.boss;
+    this.active = (boss === bosses.bosses["SunKingsSalvation"]) ? true : false;
```
philipsabri: Could this be done better?

Sharrq: Im sure we could add a BOSSES file to src/game/shadowlands and make that simpler. I cant think of a way to make it any simpler with what is already available.

i dont think there are many, if any, instances where we have analysis for a specific encounter so i dont think we ever made anything for it.

Sharrq: actually you could probably do something like

`this.active = getBossName(this.owner.fight, false) === "Sun King's Salvation"`

not sure that that is any cleaner tho

philipsabri: > Im sure we could add a BOSSES file

> i dont think there are many, if any, instances where we have analysis for a specific encounter so i dont think we ever made anything for it.

Yeah, BOSSES file would make this look the cleanest, but if its not going to get used maybe its not worth to implement it.

Pewtro: Is there a reason to create a ternary for this? Isn't it enough to simply do ``this.active = (boss === bosses.bosses["SunKingsSalvation"])``?

Pewtro:
> Much better.
> 
> FYI @Pewtro your hunter tests are getting updated in here cause they are based on Sun King for Multi Target. not sure if you wanted to check to make sure its ok.
> 
> The rest of it looks fine to me

Yeah seems fine to me

src/parser/shared/modules/spells/SoulInfusion.tsx
abelito75: Does `this.active = (boss === bosses.bosses["SunKingsSalvation"]` === `(boss === bosses.bosses["SunKingsSalvation"]) ? true : false;`? Meaning there is no need to do the whole ? : aspect

philipsabri: Fixed. Overthinked it for sure.

Pewtro:
Thanks!

PR #4294, opened 2020-12-30
Convert affliction to Typescript

JoeyBG:
This seems big enough, I will finish the work in another PR.

Here is a screenshot of the statistics tab.

image

Regarding `UptimeBar`, I removed the configurable height since it was set to 100% everywhere now. I also changed the background to something a little more visible.

What's left to do to have a working MVP:
- Malefic Rapture analyzer
- Update Soul Conduit logic to use Malefic Rapture instead of Unstable Affliction
- Add minimal support for covenant spells to Darkglare


src/parser/core/Events.ts
```diff
@@ -720,9 +714,11 @@ export interface SoulbindTrait {
   icon: string,
 }
 
+export type ConduitRank = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15;
+
 export interface Conduit {
```
Pewtro: is it really necessary to do this? Seems to be more of an annoyance in the long run. 
We already map explicitly to those values in the relevant function, so it should be unnecessary to have a type for it beyond the basic number.

src/parser/core/Combatant.ts
```diff
@@ -6,7 +6,15 @@ import { Enchant } from 'common/ITEMS/Item';
...
-import { Buff, CombatantInfoEvent, Conduit, EventType, Item, SoulbindTrait } from 'parser/core/Events';
+import {
+  Buff,
+  CombatantInfoEvent,
+  Conduit,
+  ConduitRank,
+  EventType,
+  Item,
+  SoulbindTrait,
+} from 'parser/core/Events';
```
Pewtro: should keep this one-lined

src/parser/paladin/holy/modules/talents/RuleOfLaw.js
```diff
@@ -63,7 +63,6 @@ class RuleOfLaw extends Analyzer {
               uptimeHistory={history}
               start={this.owner.fight.start_time}
               end={this.owner.fight.end_time}
-              style={{ height: '100%' }}
             />
           </div>
```
Pewtro: Shouldn't holy paladins stay untouched in this PR, so style should still be usable in the uptimebar? Can just make it default to 100%, if that's the generic case right now - but imo it should stay as a possibility to alter that.

src/parser/warlock/affliction/integrationTests/example.test.ts.snap
```diff
@@ -5241,15 +5340,17 @@ exports[`Affliction Warlock integration test: example log ShadowEmbrace matches
...
               </sup>
             </small>
           </dfn>
           <br />
           0
-           DPS 
+           DPS
+           
           <small>
             0.00
```
Pewtro, line 5348: What's with all the weird additional breaklines here? They shouldn't be there

Pewtro (changes requested):
A couple of things

src/parser/core/Events.ts
JoeyBG: I ended up doing that in order to avoid typescript errors when using [this map](https://github.com/JoeyBG/WoWAnalyzer/blob/chore/affliction-ts/src/parser/warlock/affliction/modules/features/Abilities.ts#L8).

I think in general, the more precise you get with typing, the more value you get from it, such as, for instance, forgetting a conduit rank at some point in the future. The explicit declaration will make it error out.

src/parser/core/Combatant.ts
JoeyBG: I think Prettier did that automatically 🤔

src/parser/paladin/holy/modules/talents/RuleOfLaw.js
JoeyBG: I somewhat disagree. The UptimeBar doesn't _look right_ when `height` is not 100%. I'm sure that at some point there was a valid use case for it, but instead of fixing css that will end up not being used, I'd much rather enforce it to a sane value.

I can remove the change from the holy pal module though, but IMHO it's just keeping dead code there for the sake of it at this point.

src/parser/warlock/affliction/integrationTests/example.test.ts.snap
JoeyBG: Fixed! Thanks. I think those were inserted by error when converting from `js` to `tsx`.

src/parser/core/Events.ts
Pewtro: The codebase uses arrays for effects of conduits if you look at existing implementations 😄

src/parser/paladin/holy/modules/talents/RuleOfLaw.js
Pewtro: Hence adding the default value but letting others change it if they want to

src/parser/core/Events.ts
JoeyBG: Oh! I didn't see that. Any example I can steal from?? I didn't write the above code, but I can refactor it if it's a better way to do it

JoeyBG: Nevermind, I got it. Fix incoming :)

JoeyBG: Fixed by fb9e0f147c9f48b9ede1217ca7a1d7ad021a337e

src/parser/paladin/holy/modules/talents/RuleOfLaw.js
JoeyBG: Done in 088af66c699795a7009fc633bc8449658bce0940

Pewtro (approved):
Thanks!

PR #4297, opened 2020-12-31
[Balance Druid] Typescript and Integration Tests

Sharrq:
- Updated Balance Druid to Typescript
- Updated EarlyDotRefreshes to Typescript
- Removed the Empowerments (Were removed with Shadowlands)
- Added Integration Tests
- Updated the example log

Planning on updating their Spellbook soon, just didnt want to throw it into this


src/parser/druid/balance/modules/features/AlwaysBeCasting.tsx
```diff
@@ -0,0 +1,28 @@
...
+          .icon('spell_mage_altertime')
+          .actual(<Trans id="shared.suggestions.alwaysBeCasting.downtime">{formatPercentage(actual)}% downtime</Trans>)
+          .recommended(`<${formatPercentage(recommended)}% is recommended`)
+          .regular(recommended + 0.15).major(recommended + 0.2));
+    }
+  }
+
+  statisticOrder: any = STATISTIC_ORDER.CORE(1);
+}
+
```
Pewtro, line 28: statisticOrder is deprecated, and ideally we shouldn't be using magic numbers for downtime as it can just follow the regular usage of thresholds that we find elsewhere. This also allows you to omit .regular().major()

src/parser/druid/balance/modules/features/Checklist/Component.tsx
```diff
@@ -97,11 +97,9 @@ const BalanceDruidChecklist = ({ combatant, castEfficiency, thresholds }) => {
       </Rule>
       <Rule
         name="Do not overcap your resources"
-        description="You should try to always avoid overcapping your Astral Power and your solar and lunar empowerments. Sometimes you can not avoid overcapping both of them. In that case, you should prioritize spending the Astral Power."
+        description="You should try to always avoid overcapping your Astral Power. Sometimes you can not avoid overcapping both of them. In that case, you should prioritize spending the Astral Power."
       >
         <Requirement name="Astral Power efficiency" thresholds={thresholds.astralPowerEfficiency} />
```
Pewtro, line 102: "both of them" in that description seems to refer to empowerments maybe? so the suggestion should be slightly rewritten as a consequence of it being removed.

src/parser/druid/balance/modules/features/Checklist/Module.tsx
```diff
@@ -44,7 +51,7 @@ class Checklist extends BaseModule {
         thresholds={{
           ...this.preparationRuleAnalyzer.thresholds,
 
-          downtime: this.alwaysBeCasting.suggestionThresholds,
+          downtime: this.alwaysBeCasting.downtimeSuggestionThresholds,
           cancelledCasts: this.cancelledCasts.suggestionThresholds,
           moonfireUptime: this.moonfireUptime.suggestionThresholds,
```
Pewtro: At the moment the module doesn't actually have any downtimeSuggestionThresholds as mentioned earlier.

src/parser/druid/balance/modules/features/EarlyDotRefreshesInstants.tsx
```diff
@@ -87,14 +89,14 @@ class EarlyDotRefreshesInstants extends CoreEarlyDotRefreshesInstants {
   static dots = DOTS;
 
   // Check for Stellar Drift on both the cast event and the next event, since it might have expired mid GCD.
-  couldCastWhileMoving(castEvent, endEvent) {
+  couldCastWhileMoving(castEvent: CastEvent, endEvent: CastEvent) {
     if (this.selectedCombatant.hasBuff(SPELLS.STELLAR_DRIFT.id, castEvent.timestamp) && this.selectedCombatant.hasBuff(SPELLS.STELLAR_DRIFT.id, endEvent.timestamp)) {
       return SPELLS.STELLAR_DRIFT.name;
```
Pewtro, line 94: is endEvent always a CastEvent? The description just says "next event"

src/parser/shaman/elemental/integrationTests/example.test.ts.snap
```diff
@@ -5988,7 +5988,7 @@ exports[`Elemental Shaman integration test: example log matches the checklist sn
                   }
                 >
                    
-                  0.00%
+                  6.67%
```
Pewtro: Elemental shamans being changed - is that intended?

Pewtro (changes requested):
Couple of things

src/parser/shaman/elemental/integrationTests/example.test.ts.snap
Sharrq: not sure. maybe they use Early Dot Refreshes. Ill double check to see if anything changed functionally in there

src/parser/druid/balance/modules/features/EarlyDotRefreshesInstants.tsx
Sharrq: when i looked it seemed like it was only cast events, but ill double check

src/parser/shaman/elemental/integrationTests/example.test.ts.snap
Sharrq: So, the snapshot they are using is from Eternal Palace (Orgozoa), and when i look in the snapshot to get the WCL report so i can load up that fight and see what is going on, the report doesnt exist anymore, and has a weird link.

https://www.warcraftlogs.com/reports/TEST/#fight=18&source=23&type=damage-done

So i dont know what happened to make their snapshot change, but i also cant dig into that log to check ...

src/parser/druid/balance/modules/features/EarlyDotRefreshesInstants.tsx
Sharrq: It looks like end event is indeed always a cast event. But i can change it to AnyEvent just in case?

Pewtro: Rather we keep it CastEvent if it seems to always be a cast 😄 That'll catch some errors for us if they are there

src/parser/shaman/elemental/integrationTests/example.test.ts.snap
Pewtro: Well that WCL link would've never worked 😄 
Let me upload a fresh one to them, and then we can compare changes between them properly.

Sharrq: Seems like with the new integration test nothing changed due to my changes

src/parser/druid/balance/modules/features/EarlyDotRefreshesInstants.tsx
Sharrq: changed it back to CastEvent

Pewtro (approved):

PR #4298, opened 2020-12-31
[MW/Core] Fixed calls to calculateStatPercentage with negative rating (kind of)

abelito75:
Also rewrote mastery stat weight for mw... I hope

### DIFF
--- a/src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
+++ b/src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
@@ -7174,7 +7174,7 @@ exports[`Mistweaver Monk integration test: example log StatValues matches the st
             <td
               className="text-right"
             >
-              Infinity
+              10.21
             </td>
             <td
               style={
--- a/src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
+++ b/src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
@@ -7174,7 +7174,7 @@ exports[`Mistweaver Monk integration test: example log StatValues matches the st
             <td
               className="text-right"
             >
-              10.21
+              Infinity
             </td>
             <td
               style={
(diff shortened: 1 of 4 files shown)


src/parser/monk/mistweaver/integrationTests/example.test.ts.snap
```diff
@@ -7174,7 +7174,7 @@ exports[`Mistweaver Monk integration test: example log StatValues matches the st
             <td
               className="text-right"
             >
-              10.21
+              Infinity
```
MartijnHols: Uh?

abelito75: Yeah. I saw that too but couldn't reproduce it on any logs from the current xpack. I'm assuming it has to do with inflated stats from last xpack and the new stats scaling values

MartijnHols: I reckon it has to do with a bug

abelito75: In all reality having a leech weight of 10.21 was already insane making me think the test log is wonky

Pewtro: Infinity is very possible if the person has way too high secondaries. 
https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4298/files#diff-6ce7a480e8b1ceb541faf90324e8bac5c822eff9595ff54c29836619c43abf98R534

MartijnHols: It doesn't go from 10 to Infinity without a serious bug, weirder when all the bug is supposed to fix is handling "negative" stat values

abelito75: If anything its from the mistweaver code I changed since it only effected mistweaver

PR #4299, opened 2021-01-01
[Havoc Demon Hunter] Updated suggestions for Demonic, Blade Dance and Death Sweep

philipsabri:
- Converted Demonic ability script to Typescript
- Updated suggestions for Blade Dance and Death Sweep to show only when using the right talents.
- Updated Suggestions for the double use of Death Sweep + Demonic to only show when using the right talents
- Added a **new** suggestion for doing a bad cast with Death Sweep and/or Blade Dance on Single target

image

Bad cast on timeline
image


src/parser/demonhunter/havoc/modules/Abilities.js
```diff
@@ -42,9 +42,9 @@ class Abilities extends CoreAbilities {
           base: 1500,
         },
         castEfficiency: {
-          suggestion: combatant.hasTalent(SPELLS.FIRST_BLOOD_TALENT.id),
-          recommendedEfficiency: 0.95,
-          extraSuggestion: <>This should be part of your single target rotation due to the <SpellLink id={SPELLS.FIRST_BLOOD_TALENT.id} /> talent. This includes the <SpellLink id={SPELLS.DEATH_SWEEP.id} /> casts since they are the same ability and share their cooldowns.</>,
+          suggestion: combatant.hasTalent(SPELLS.FIRST_BLOOD_TALENT.id) || combatant.hasTalent(SPELLS.TRAIL_OF_RUIN_TALENT.id),
+          recommendedEfficiency: 0.85,
+          extraSuggestion: <>This should be part of your single target rotation due to using either or both of the <SpellLink id={SPELLS.FIRST_BLOOD_TALENT.id} /> and <SpellLink id={SPELLS.TRAIL_OF_RUIN_TALENT.id} /> talent. This includes the <SpellLink id={SPELLS.DEATH_SWEEP.id} /> casts since they are the same ability and share their cooldowns.</>,
```
abelito75: This sentence is strange to me. Just reads weird

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  badCast = 0;
+  hitCount = 0;
+  hitTimeStamp: number = 0;
+  firstHitTimeStamp: number = 0;
+  strikeTime: number = 1000;
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.spec === SPECS.HAVOC_DEMON_HUNTER;
```
abelito75: You don't have to check for spec

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (!this.active || this.selectedCombatant.hasTalent(SPELLS.TRAIL_OF_RUIN_TALENT) || this.selectedCombatant.hasTalent(SPELLS.FIRST_BLOOD_TALENT)) {
```
abelito75: This says 
if the yare not demon hunter or if they have talent a or talent b active then die

abelito75: You should set active to true or false if the file is active or not meaning put the `hasTalent` calls before the if and set active to whatever it should be

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+        return;
+    }
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell(SPELLS.DEATH_SWEEP_DAMAGE), this.onDamage);
```
abelito75: If all of these listeners go down to a single method then you could put all the spells in the `.spell` call by putting them in an array like this `.spell([SPELLS.DEATH_SWEEP_DAMAGE, BLADE_DANCE_DAMAGE, ...)`

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+      },
+      style: ThresholdStyle.NUMBER,
+    };
+  }
+
+  badCast = 0;
+  hitCount = 0;
+  hitTimeStamp: number = 0;
+  firstHitTimeStamp: number = 0;
+  strikeTime: number = 1000;
```
abelito75, line 26: Is this always one second or does it change with haste?

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+        this.badCast += 1;
+      }
+      this.firstHitTimeStamp = this.hitTimeStamp //Timestamp for first hit in strike
+      this.hitCount = 0;
+    }
+    this.hitCount += 1;
+  }
+
+  onFightEnd(event: FightEndEvent){
+    if (this.hitCount < 5){
```
abelito75: Is it accurate to say 
If the last cast of the fight is blade dance but the fight finished before blade dance could finish it was a bad cast?
if so then cool if not then this is going to inflate your bad casts by 1

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { ThresholdStyle, When } from 'parser/core/ParseResults';
+import SpellLink from 'common/SpellLink';
+
+//For Blade dance and Death Sweep
+class BladeDance extends Analyzer {
+  get suggestionThresholds() {
+    return {
+      actual: this.badCast,
+      isGreaterThan: {
+        minor: 0,
```
abelito75: This probably shouldn't be 0 as you could get mechanics that require you to move, boss movement, or phasing that could make you miss a few hits. A little bit of lee way is a nice way to combat fight rng

abelito75: Typically if something is mess up once and get yelled at it has to be something that fight rng can't mess with

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+        average: 0,
+        major: 1,
+      },
+      style: ThresholdStyle.NUMBER,
+    };
+  }
+
+  badCast = 0;
+  hitCount = 0;
+  hitTimeStamp: number = 0;
```
abelito75: The only place you use this is in a listener meaning it doesn't need to be global.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onFightEnd(event: FightEndEvent){
+    if (this.hitCount < 5){
+      //Check last strike
+      this.badCast += 1;
+    }
+  }
+
+  suggestions(when: When) {
+    when(this.suggestionThresholds)
+    .addSuggestion((suggest, actual, recommended) => suggest(<>You should not cast <SpellLink id={SPELLS.BLADE_DANCE.id} /> or <SpellLink id={SPELLS.DEATH_SWEEP.id} /> on single target when you are not using <SpellLink id={SPELLS.FIRST_BLOOD_TALENT.id} /> or <SpellLink id={SPELLS.TRAIL_OF_RUIN_TALENT.id} /> as a talent,</>)
```
abelito75: Doesn't end with a period.

src/parser/demonhunter/havoc/modules/talents/Demonic.tsx
```diff
@@ -42,12 +45,13 @@ class Demonic extends Analyzer {
     this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.DEATH_SWEEP), this.onDeathSweepCast);
   }
 
-  onEyeBeamCast(event) {
+  onEyeBeamCast(event: CastEvent) {
     const hasMetaBuff = this.selectedCombatant.hasBuff(SPELLS.METAMORPHOSIS_HAVOC_BUFF.id, event.timestamp - 1000);
 
-    if (hasMetaBuff) {
+    if (hasMetaBuff || !(this.selectedCombatant.hasTalent(SPELLS.TRAIL_OF_RUIN_TALENT) || this.selectedCombatant.hasTalent(SPELLS.FIRST_BLOOD_TALENT))) {
```
abelito75: since talents can't change throughout the fight i'd put these two calls in the constructor and make them global variables

src/parser/demonhunter/havoc/modules/talents/Demonic.tsx
```diff
@@ -42,12 +45,13 @@ class Demonic extends Analyzer {
...
       return;
     }
+
     this.eyeBeamCasts += 1;
     this.eyeBeamTimeStamp = event.timestamp;
```
abelito75, line 57: I can't comment on the line but im pretty sure that deathsweepsInMetaCounter will never be undefined now meaning this is dead code

src/parser/demonhunter/havoc/modules/talents/Demonic.tsx
```diff
@@ -63,14 +67,14 @@ class Demonic extends Analyzer {
     this.deathsweepsInMetaCounter = 0;
   }
 
-  onDeathSweepCast(event) {
+  onDeathSweepCast(event: CastEvent) {
     if (this.eyeBeamTimeStamp !== undefined && (event.timestamp - this.eyeBeamTimeStamp) < META_BUFF_DURATION_EYEBEAM) {
```
abelito75: `this.eyeBeamTimeStamp !== undefined` can never be undefined

abelito75 (changes requested):

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
Pewtro: Why would you exit out of this module if you have those talents?

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: I couldnt find a way to calculate it properly with haste, but its max 1 second, and its impossible to cast it again during that time.

abelito75: Anything with haste is just `duration / (1 + %haste)`

src/parser/demonhunter/havoc/modules/Abilities.js
philipsabri: Could just remove the text about the talents tbh if its makes it weird. But I feel the need to say its because you are using FIRST_BLOOD_TALENT **and/or** TRAIL_OF_RUIN_TALENT, could'nt find a better way to do it than saying "using either or both of the..." and not just point at First Blood as they did before.

But I can look into it.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: Yeah ik, but how do I get the correct haste value? I know you can do that in the Abilities script, but I couldnt find out where you actually get the value from and use it in a spell script.

Saw some spell scripts using functions to get the haste value, but it feels like Im overdoing it for something as this simple.

src/parser/demonhunter/havoc/modules/talents/Demonic.tsx
philipsabri: Fixed

src/parser/demonhunter/havoc/modules/talents/Demonic.tsx
philipsabri: Moved it into `talentsCheck`, line 31. If this is what u meant by moving them into the constructor?

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: I agree, the people who use this wrong will do it more than once, and if u have bad rng and it mess you up it should be minor. Updated the numbers.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: I would say if you do blade dance in the end of the fight and it doesnt hit enough, its a bad cast because you have other spells available to you that could do more damage. If it would be because you need 5 more fury for your main non-cd abiliity, melee + demons bite would do more damage in the same amount of time.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: Changed `this.active = ...` to check you not having those talents instead.

@Pewtro because these bad casts etc should not be calculated if you have any of those 2 talents. These spells should only be in your rotation if you have one of those talents.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
abelito75: import statTracker and do
statTracker.hastePercentage(statTracker.currentHasteRating));

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
abelito75: lit

abelito75 (changes requested):

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
philipsabri: image
(last hit - first hit) = total time
373823500-373822541=959

1000 / (1 + 0.17303030303030303) = 852.492895893

According to this its not affected by haste. Looks more like its just max 1 second and it depend on where targets are located(in the hit area) and how many during the last hit, as other strikes have different total times, but no one is matching the haste calculation.

src/parser/demonhunter/havoc/modules/Abilities.js
philipsabri: image

This is the best I can come up with. Im not the best with sentences.

src/parser/demonhunter/havoc/modules/spells/BladeDance.tsx
abelito75: Lit

PR #4300, opened 2021-01-01
[Core] Character Page Legendary, Conduit, Covenant & Soulbinds update

Pewtro:
Before:
image

After:
image


src/interface/report/Results/Character/PlayerInfo/Gear.tsx
```diff
@@ -11,16 +11,15 @@ interface Props {
 
 const Gear = (props: Props) => {
   const { gear } = props;
-  const EPIC_ITEMS_ILVL = 340;
+  const EPIC_ITEMS_ILVL = 200;
```
abelito75: I won't stop the PR but this should probably be placed somewhere else since it is used in multiple locations

abelito75: Also this one is more of a blocker isn't epic items purple items? Meaning 184 would be the item level threshold not 200? (mythic dungeons)

abelito75 (changes requested):
Might not actually require changes

src/interface/report/Results/Character/PlayerInfo/Gear.tsx
Pewtro: Darkmoon decks are 200 and blue

Pewtro: Donesies

abelito75 (approved):
Love it

src/interface/report/Results/Character/PlayerInfo/Gear.tsx
abelito75: I know you did done it but It feels like Dark moon decks are the exception not the standard :P (which i guess is what you thought)

PR #4301, opened 2021-01-02
[Resto Druid] add lifebloom uptime to the efflo lifebloom checklist suggestion

bcrabbe:
before
Screen Shot 2021-01-02 at 09 12 23

after
Screen Shot 2021-01-01 at 23 50 30


abelito75:
Yeah. That was weird how it was in the module file but not on the checklist... seems like someone removed it at some point but nice catch

PR #4307, opened 2021-01-03
Added Protection Paladin Abilities

5chmidt:
Added ability tracking for Divine Toll and Final Stand for protection paladin.

image

I believe this are cooldowns that are not effected by haste and that is how I coded them, however I am seeing errors for them being cast early.  I did add the 30% CD reduction from Unbreakable Spirit, but maybe I am missing something here.
> 17:55.359 (module: SpellUsable) Final Stand 204079 was cast while already marked as on cooldown. time passed: 800 cooldown remaining: 209200 expectedDuration: 210000

Here is my [test log](https://www.warcraftlogs.com/reports/wBxNPJ3Tg2v4FVKH#fight=last&type=casts&source=3)

### DIFF
--- a/src/parser/paladin/protection/modules/Abilities.js
+++ b/src/parser/paladin/protection/modules/Abilities.js
@@ -1,7 +1,6 @@
 import SPELLS from 'common/SPELLS';
 
 import CoreAbilities from 'parser/core/modules/Abilities';
-import COVENANTS from 'game/shadowlands/COVENANTS';
 //import SpellLink from 'common/SpellLink';
 
 class Abilities extends CoreAbilities {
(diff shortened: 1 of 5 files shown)


src/parser/paladin/protection/modules/Abilities.js
```diff
@@ -242,6 +242,31 @@ class Abilities extends CoreAbilities {
         },
         enabled: combatant.hasTalent(SPELLS.REPENTANCE_TALENT.id),
       },
+      {
+        spell: SPELLS.DIVINE_TOLL,
```
abelito75: All you need to do to add this is add `DivineToll.ts` to the CombatLogParser

abelito75 (changes requested):

src/parser/paladin/protection/modules/Abilities.js
5chmidt: Good catch, I have corrected it to use DivineToll.ts.
Thanks,

abelito75:
Just as a heads up you *can* extend the Divine Toll module in prot paladin if you want to do more indepth stuff with it. Just an easy way to make sure its there and what not

PR #4313, opened 2021-01-08
[Hunter] Remove some naughty : any types

Pewtro:


Pewtro:
And God said let there not be any any's!

PR #4319, opened 2021-01-10
[Warlock] Finish Affliction Typescript conversion

JoeyBG:
The only meaningful change is Vile Taint support for Darkglare damage calculations.

Soul Conduit is still in javascript, but it will need to be mostly rewritten.

There also many whitespaces added back, in #4294 I broke a few things, I have a better understanding of why they were there now :D

image


emallson (approved):
none of the snapshots really changed. that must mean its correct, right?

PR #4322, opened 2021-01-10
[Arms Warr] Update execute range for venthyr covenant

Chasson1992:
### Update execute range for Venthyr covenant
Venthyr covenant for warriors replaces execute with Condemn that allows the target to be executed above 80% health. Fixing the ExecuteRange tracker for arms warriors to reflect this.

Below is an example that shows the execute range working with the Mortal Strike tracker in the timeline for a target above 80% health. Also verified that this doesn't break the trackers for non-venthyr warriors.

ExecuteRange


src/parser/warrior/arms/modules/core/Execute/ExecuteRange.js
```diff
@@ -15,14 +17,16 @@ const EXECUTE_RANGE_MASSACRE = 0.35;
...
 
-  execRange = (this.selectedCombatant.hasTalent(SPELLS.MASSACRE_TALENT_ARMS.id) ? EXECUTE_RANGE_MASSACRE : EXECUTE_RANGE);
   enemyMap = {};
 
   isExecPhase = false;
   execPhaseStart = 0;
   execPhaseDuration = 0;
   lastHitInExecPhase = 0;
 
+  lowerThreshold = (this.selectedCombatant.hasTalent(SPELLS.MASSACRE_ARMS_TALENT.id) ? EXECUTE_RANGE_MASSACRE : EXECUTE_RANGE);
```
abelito75: This isn't the right spell

src/parser/warrior/arms/modules/core/Execute/ExecuteRange.js
```diff
@@ -15,14 +17,16 @@ const EXECUTE_RANGE_MASSACRE = 0.35;
  */
 class ExecuteRangeTracker extends Analyzer {
```
abelito75, line 18: tbh should probably just swap this to use the ExecuteHelper module

abelito75 (changes requested):

src/parser/warrior/arms/modules/core/Execute/ExecuteRange.js
Chasson1992: Fixed in latest commit

src/parser/warrior/arms/modules/core/Execute/ExecuteRange.js
Chasson1992: I figured, I saw Kill Shot for hunters was using the helper.

There was a lot I wanted to clean up in this class but I figured I would get this small change in first and then submit a larger PR that would update all of that. Also, since this class technically works I didn't know if we wanted to change everything to use the helper.

src/parser/warrior/arms/modules/core/Execute/ExecuteRange.js
Chasson1992: Fixed in latest commit

abelito75 (approved):

PR #4338, opened 2021-01-15
[Shared] Update CastEfficiency to fix negative return values

Khazakdk:
For spells with CDR effects and a baseline CD longer than the fight duration, it was possible to get negative cast efficiency values.  [Army of the Dead](https://www.wowhead.com/spell=42650/army-of-the-dead) and [Army of the Damned](https://www.wowhead.com/spell=276837/army-of-the-damned) is an example where this was an issue.  

This was caused by the timeOffset used to account for the CD extending past the end of the fight using the full 480 second duration, even if the fight itself was less than that.  The offset would be bigger than the time spent on CD and cause a negative value to propagate forward to the rest of the calculations.

To fix this,  I added a condition before timeOffset to check if averageCooldown exists.  If it does, I substituted the averageCooldown in for the static coolDown.  This will affect any spell that has a dynamic CD and I requested review from those with affected snapshots in case this changes makes those specs less accurate.  If it's negatively affecting other specs, I can move this within unholy only.  

Verified my math with the following usecase:
https://www.warcraftlogs.com/reports/vAxYVyrP7wmLqQj8/#fight=20&source=88&type=casts

In a 247 second fight, AOTD was cast at 3 seconds.  With 50 casts of Death Coil after that, the CD of AOTD was reduced from 480 seconds to 230 seconds.  247 - 230 - 3 - ~1.3 (GCD) = ~13 seconds where Army was available and ~234 seconds unavailable.  234/247 = .9474, which is within .1% of what the code calculated.

image


Sharrq:
Once i merge #4339 you will have to update the mage snapshots on your PR. Found an issue in Combustion's CDR that might be skewing yours.

Pewtro (approved):
Makes sense

Sharrq:
Warlock is fine. Ill double check Fire after you update from shadowlands to get the changes i made to Combustion

PR #4340, opened 2021-01-15
[Core] Missing Encounters Warning

Sharrq:
Updated the warning message that showed if there were no boss encounters in a log to make it show always and recommend clicking Refresh if something is missing.

Warning Message (Updated)
image


Pewtro:
I don't think we should show this always.. Seems distracting for the, I'm guessing, silent majority of users who don't report issues with it in Discord

Sharrq:
Im also seeing in places like the mage discord "Wowanalyzer hasnt uploaded the fights yet" as an assumption for why stuff is missing.

Also i remember when i added this before a couple ppl commented in Discord that it should be shown always. Hence why i did that

MartijnHols (approved):

Issue #3781, opened 2020-10-02
[Shadowlands] Spellbook updates for unmaintained specs


Sharrq:
Along with the Shadowlands release, we should also at the very least update the spellbooks for the specs that are not currently maintained so that the specs are at least semi functional

The below list is based on the last time the spec was updated (Patch Level).

- [x] Havoc Demon Hunter
- [x] Vengeance Demon Hunter
- [x] Balance Druid
- [x] Guardian Druid
- [x] Restoration Druid
- [x] Protection Paladin
- [x] Shadow Priest
- [x] Outlaw Rogue
- [x] Assassination Rogue
- [x] Subtlety Rogue
- [x] Affliction Warlock
- [x] Destruction Warlock
- [x] Demonology Warlock

DylanDirlam:
Shadow priest can be marked off here.

abelito75:
Closing this.

PR #4349, opened 2021-01-16
[Druid] Add Shadowlands conduits information for Druid

tiborhercz:
Added the Conduits for all Druid Specs

Towards #3781

### DIFF
--- a/src/parser/druid/feral/CHANGELOG.js
+++ b/src/parser/druid/feral/CHANGELOG.js
@@ -4,7 +4,7 @@ import { Abelito75, Zeboot, LeoZhekov, Tora, Xcessiv, Tiboonn } from 'CONTRIBUTO
 import { change, date } from 'common/changelog';
 
 export default [
-  change(date(2021, 1, 16), <>Add class conduits</>, Tiboonn),
+  change(date(2021, 1, 16), <>Added spell information for conduits</>, Tiboonn),
   change(date(2020, 12, 15), <>Update Bloodtalons for Shadowlands</>, Tora),
   change(date(2020, 12, 14), <>Corrected tracking of Berserk cooldown usage.</>, Xcessiv),
   change(date(2020, 11, 6), <>Replaced the deprecated StatisticBoxes with the new Statistic and reworked SavageRoar module.</>, LeoZhekov),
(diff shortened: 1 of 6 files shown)


src/parser/druid/feral/CHANGELOG.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
-import { Abelito75, Zeboot, LeoZhekov, Tora, Xcessiv } from 'CONTRIBUTORS';
+import { Abelito75, Zeboot, LeoZhekov, Tora, Xcessiv, Tiboonn } from 'CONTRIBUTORS';
 import { change, date } from 'common/changelog';
 
 export default [
+  change(date(2021, 1, 16), <>Add class conduits</>, Tiboonn),
```
Sharrq: I would remove the fragments from these  (The <></>) since it is just text.

And to be clear, i would change it to "Added spell information for conduits", cause adding the spell IDs and icons and such to SPELLS doesnt actually add the conduits to anything, it just allows them to be added to stuff

Sharrq (changes requested):

src/parser/druid/feral/CHANGELOG.js
tiborhercz: Good feedback! Thanks 
I already noticed the React error, fixed that.
Also updated the change log message.

Wanted to start with something simple to get familiar with the project and the code, to maybe later update the Balance druid class :)

src/parser/druid/feral/CHANGELOG.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
-import { Abelito75, Zeboot, LeoZhekov, Tora, Xcessiv } from 'CONTRIBUTORS';
+import { Abelito75, Zeboot, LeoZhekov, Tora, Xcessiv, Tiboonn } from 'CONTRIBUTORS';
 import { change, date } from 'common/changelog';
 
 export default [
+  change(date(2021, 1, 16), <>Added spell information for conduits</>, Tiboonn),
```
Sharrq: these fragments too and then it should be good to go.

In general, if its just text, just wrap it with ' '. If it has something like SpellLink or something like that, then use <> </>

tiborhercz: makes sense!

Sharrq (approved):

PR #4354, opened 2021-01-17
[Arms Warrior] Updated Mortal Strike module and added a Deep Wounds early refresh during execution phase

Otthopsy:
Mortal Strike module :

- Track if Mortal Strike has been casted too much or too few during execution phase (as it is only used to reapply Deep Wounds)
- Changed the suggestion for the cast of Mortal Strike during execution phase
- Updated the checklist module of Mortal Strike to take into account the changes

Deep wounds :

- Added an EarlyDotRefresh to show whether or not the refresh of Deep Wounds during execution phase should be improved (as you need to always have the dot on the target during this phase). Only Mortal Strike will and should trigger the file
- Added a suggestion to reflect the previous point
- The EarlyDotRefreshes is the core for the Deep Wounds file, I just added the lines 119 to 121 to track only the cast during the execution phase

Also, added my name to the Changelog and Contributors


MartijnHols (approved):
Looking good, thanks!

MartijnHols:
Not sure why status checks aren't running...

PR #4355, opened 2021-01-17
[Druid] Update balance druid spells

tiborhercz:
Updated Spells for Balance druid and changed all occurences for 'Solar Wrath' and 'Lunar Strike'.

In the `UnempoweredLunarStrike.tsx` file I only changed the names of the Solar Wrath and Lunar Strike for now, and kept the file name the same.
The `UnempoweredLunarStrike.tsx` file will be replaced with UnempoweredStarfire module and a additional UnempoweredWrath module will be created.

Towards #3781


MartijnHols (approved):
Thanks! Just need to fix the status checks and this is good to go.

tiborhercz:
@MartijnHols Fixed the status checks.

PR #4357, opened 2021-01-18
[BRM] Stormstouts Last Keg's legendary

mathieu-carlier:
Add support for Stormstouts Last Keg's legendary for BRM and module in the stats to calculate the 30% damage gains

image


DylanDirlam (approved):
Everything looks good to me, @emallson maybe just double check since brewmaster.

MartijnHols (approved):
Looks good to me, thanks!

PR #4361, opened 2021-01-19
[Shadow Priest] Death and Madness talent threshold suggestion

DylanDirlam:
* Adds a suggestion with a low threshold for Death and Madness resets
* Main goal is to point out if someone took the talent on a single target or low multi-target fight and is getting little/no value

image


MartijnHols (approved):
Thanks!

PR #4368, opened 2021-01-22
[Shadow Priest] Add insanity generation tracking 

DylanDirlam:
* Adds insanity waste tracker - includes void torrent which doesn't show waste in log
* Adds suggestion for overcapping
* Adds checklist rule since it's so important

Do note that the screenshots below were me maxing out insanity for 45 seconds with a bunch of random abilities.. so it's intentionally showing bad here.

image
image
image


MartijnHols (approved):
Thanks!

PR #4369, opened 2021-01-23
[Shadow Priest] Add proper execute support for Shadow Word: Death

DylanDirlam:
## New Shadow Word: Death module
* Makes use of execute helper to determine number of casts while in execute range
  * Since shadow is relatively unique where it can be used above execute threshold, but it's not as effective, I added a method that returns the number of casts not in execute threshold
* Calculates `maxCasts` from execute, but also adds the number of casts above execute threshold otherwise they could end up with 9/6 casts if they had 3 above execute. This way it would show as 9/9 to account for the above execute casts.
* Cast efficiency is only 80% because depending on fight length, there may only be 4 possible casts in a 3-4 minute fight.

image

## Moved some checklist rules around.
* Adds SW: Death into core rotation
* Moves void torrent/shadow crash into rotational instead of cooldowns
* Added Power Infusion to the check list cause I forgot it before
### Before
image

### After
image


MartijnHols (approved):
Looks good to me. Very nice description as well. Thanks.

PR #4374, opened 2021-01-23
[Priest] Add 2 shadow legendaries and 1 priest legendary

DylanDirlam:
## General Changes
* Added Power Infusion to the global cooldown throughput tracker as it can be cast on any player for huge damage or healing boosts

### Twins of the Sun Priestess (All Specs)
* Calculates number of bad casts
  * A bad cast is a cast on themself with the legendary. Since using PI on an ally gives both of them PI, there is no case where they should ever cast it on themself. Even giving the worst dps the PI is still a net gain.
* Suggests they use on ally if they have any bad casts

image
image

### Eternal Call to the Void (Shadow)
* Legendary gives random proc when using mind flay or mind sear. No other interaction is required.
* Gives DPS gain statistic from using the legendary

image

### Talbadar's Stratagem
* Legendary gives bonus damage if the target has Vampiric Touch, Shadow Word: Pain, and Devouring Plague
* Gives DPS gain statistic from using the legendary

image

### DIFF
--- a/src/common/SPELLS/shadowlands/legendaries/priest.ts
+++ b/src/common/SPELLS/shadowlands/legendaries/priest.ts
@@ -21,7 +21,23 @@ const legendaries: SpellList<LegendarySpell> = {
     bonusID: 6973,
   },
   //endregion
-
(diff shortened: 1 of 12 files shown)


MartijnHols (approved):
LGTM. Thanks!

PR #4376, opened 2021-01-23
[Core] Fix crash on About tab of unsupported specs

DylanDirlam:
Page was crashing on some unsupported specs when their patch was set to null:
https://wowanalyzer.com/report/MyhaxNdgVTJt6qrp/16-Mythic+Huntsman+Altimor+-+Kill+(6:06)/Artillocrani/standard/about

image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -9,6 +9,7 @@ import { change, date } from 'common/changelog';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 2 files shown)


MartijnHols:
Thanks for the PR. I fixed the issue in https://github.com/WoWAnalyzer/WoWAnalyzer/commit/ea87ed9843e38079f467f4808bd20abb97106365. Considering parseVersionString is typed, I prefer to solve usages than adding checks for falsey values.

PR #4377, opened 2021-01-24
[Shadow Priest] Add Searing Nightmare talent module

DylanDirlam:
## Searing Nightmare
* Adds timeline support for using searing nightmare during mind sear channel
* Adds statistic that shows amount of dps done and average targets hit
* Adds suggestion if the number of average targets hit is less than 4
  * Shadow Priest TC accepted values are devouring plague up to 3 targets. 4+ targets is Searing Nightmare
  * Also doubles as a check if the person forgot to switch talents since if they don't cast it at all, it will be 0 average targets hit

image


MartijnHols (approved):
Nice, thanks!

PR #4378, opened 2021-01-24
[Core] Add a warning if there are filtered results on a fight

DylanDirlam:
* Adds a `filteredResults` option to encounters
* Displays a warning that results on WarcraftLogs are filtered but the filtered numbers are not reflected in WoWAnalyzer

There have been multiple occasions where someone will ask in Discord why their dps is wrong on certain fights and it's due to the filtering difference. I looked into getting the filtered values from wclogs but the filtering is done on the table and raw events are never touched, so it is not an easy task at this time.

In Nathria, there is filtered damage in Huntsman and Council. In the past with Nya'lotha for example, there was filtered damage in Hivemind, Carapace, and N'zoth.

image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -9,6 +9,7 @@ import { change, date } from 'common/changelog';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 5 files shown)


MartijnHols:
I feel like we are overusing this warning which is making it less effective. Since this is primarily relevant to the statistics, and even more specifically the throughput statistics, maybe we should show this warning just above them instead (with the same 2 column width)?

PR #4380, opened 2021-01-24
Fix fight downtime toggle

MartijnHols:
It did not work at all in the previous version as its events were never bound. Converted it to TS and fixed the issue.


emallson (approved):

emallson:
Looks like the test changes are uptime related, though I'll admit i didn't read through all of them.

PR #4382, opened 2021-01-25
[BRM] Add Fortifying Ingredients, Grounding Breath & Harm Denial in the Log Parser

mathieu-carlier:
- Adding the existing conduits into the BRM Combat log parser
- Converting conduits files to TSX


MartijnHols (approved):
Thanks!

PR #4390, opened 2021-01-28
[Elemental Shaman] Add check for suboptimal chain lighting

defunes43:
It is less optimal to cast Chain Lightning than Lightning Bolt when there are less than 3 targets.

Adds a new module to check for that.


MartijnHols (approved):
This seems like a very good analyzer. Thanks!

PR #4393, opened 2021-01-29
[Shaman] Convert Maelstrom management to Typescript

defunes43:
Converted Maelstrom management to typescript
- Renamed files.
- Strongly typed variables 
-- Exported the ResourceUpdate type from the ResourceTracker, as it was needed.
-- Added reference to react-vega, used to draw the chart.


analysis/shaman/src/Maelstrom.tsx
```diff
@@ -13,7 +15,7 @@ const COLORS = {
   WASTED_MAELSTROM_BORDER: 'rgba(255, 90, 160, 1)',
 };
 
-const Maelstrom = props => {
+const Maelstrom = (props: any) => {
```
procy-dev, line 18: Instead of any, would it be better to declare an interface? I believe since there is a propTypes declared on line 114 you can just declare an interface there instead and expect the props be be of the Maelstrom.

rp4rk:
Doesn't look like this is still being worked on, going to close it - lmk if that changes :)

PR #4398, opened 2021-02-09
Convert Dungeons to TS

procy-dev:
Converted dungeon files to TS and added Dungeon interface in raids/index.ts


src/game/raids/index.ts
```diff
@@ -38,6 +38,15 @@ export interface Phase extends PhaseConfig {
   start: number[];
   end: number[];
 }
+export interface Dungeon {
```
procy-dev, line 41: Wasn't entirely sure if this was the best place to declare this interface

src/game/raids/index.ts
```diff
@@ -38,6 +38,15 @@ export interface Phase extends PhaseConfig {
...
+  id: number;
+  name: string;
+  background?: string;
+  backgroundPosition?: string;
+  headshot?: string;
+  icon?: string;
+  fight: unknown;
```
procy-dev, line 48: I saw that the Raid interface declares this as an EncounterConfig but since there is only an empty set of { } I figured I'd put this as unknown right now.

MartijnHols (approved):
Sweet, thanks!

PR #4401, opened 2021-02-13
Analyzer for using three starsurges after entering an eclipse

DerAluDev:
When using the Balance of All Things legendary one should cast three consecutive Starsurges when entering a new eclipse (www.dreamgrove.gg/balance/2020-12-08-9.0_faq#15-how-do-i-play-with-hahahugoshortcode-s29-hbhb).

Not utilizing the legendary effect properly is one of the biggest mistakes new players make.

This Module checks if the first three casts of after entering an eclipse have been Starsurges. 

In a multitarget fight you usually do not want to use your Astral Power on Starsurges in order to have enough AP to maintain full Starfall uptime. Thus the Analyzer tries to not analyze multitarget fights. The used solution for this might not be optimal. 

More informations on this in the BalanceOfAllThingsOpener.tsx file.

In order for the analysis to work properly, I fixed some things on the way, namely:

1. Added Convoke the Spirits spell 
2. Removed GCD from Celestial Alignment and Incarnation, as those got removed in Shadowlands

As I am new to contributing to this project, I am thankful for any advice and feedback.

Screenshots:
image

Correct BOAT usage:
image

Incorrect BOAT usage:
image


analysis/druidbalance/src/modules/Abilities.tsx
```diff
@@ -165,6 +160,22 @@ class Abilities extends CoreAbilities {
...
+        gcd: {
+          base: 1500,
+        },
+        castEfficiency: {
+          suggestion: true,
+          recommendedEfficiency: 0.85,
+          majorIssueEfficiency: 0.80
+        },
+        timelineSortIndex: 12,
+      },
```
Ciuffi: I'm wondering if this will interfere with this block here in the convoke module: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/05bef0cec0d64bd8a9c355dfb02061935b87d368/analysis/druid/src/ConvokeSpirits.tsx#L58-L69

Ciuffi (commented):
Other than a small comment, this looks fine to me. I am also new to this project however, so I take my opinion with a grain of salt.

analysis/druidbalance/src/modules/Abilities.tsx
DerAluDev: Thanks for the reminder, it does indeed lead to the following:
image

While I am not quite sure if the above way of injecting it into the options.abilities might lead to problems later on (Spec specific requirements on efficiency for example), I can resolve it for the time being by removing my convoke definition on the Abilities.tsx and inserting the ConvokeSpirits.tsx statistic (which should be done later on either way).

Sharrq:
I'm looking on mobile so I haven't been able to dive through all of it, but noticed a couple things.

Did you test this on a multi target fight to make sure it was ignoring it like you want it to? It could be because I'm on mobile but I don't think the starfall stuff is working (I might be wrong)

And zero doesn't like using variable++ to increment stuff. Use += 1 instead

DerAluDev:
Thank you a lot for your feedback @Sharrq 
I did test it on multiple Multitarget logs (Huntsman, Denathrius, Stone Legions, Counciö) and so far it worked on any of them.

So far it even worked for the Council fights where you play full singletarget in the beginning and start switching to AoE later on, when the adds spawn. 

Im changing the variable increment right now.

analysis/druidbalance/src/modules/Abilities.tsx
DylanDirlam: Injecting it into the `options.abilities` is fine, it doesn't really make much of a difference. Injecting it like that is generally used in cases where a deeper analysis is required on the `maxCasts` and you need more calculation to go into it (ie: murder of crows resets on target's death, void bolt only available in voidform).

DylanDirlam:
Programming-wise, everything is good and a great first PR. 

I did run into the issue of aoe not being detected and I have an idea (albeit a bit complicated) for a solution. Firstly, [here is the log/pull](https://www.warcraftlogs.com/reports/FNDJwZcKjYpGTXQV#fight=7&type=damage-done&source=14) I was looking at. Around the 35 seconds mark, the boomkin did not have starfall up when the add was still alive, but was using starfire to cleave it during the boat window.

Possible Solution
* Create an array for the abilities used during boat. 
* Use the timestamp of the cast as the identifier for each one. 
* Use the damage events and `EventHistory` to search for the timestamp of the last cast within ~150ms to match the cast to the damage events.
* Add the hit to the damage event for the cast. 
* Determine if it was a single target or multi-target cast, likely when the boat window ends since you then know how many targets were hit.

I don't know boomkin very well, so if these would still be considered bad casts that's fine. Just looking at an approach which isolates each individual boat window and checks them.

DerAluDev:
I discovered, that there are two different buffs for Balance of All Things, depending on which eclipse you entered. This lead to the prior implementation to only detect half of the buff windows.

Therefor I added both buffs to being tracked. 
Referencing: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4401/commits/f0d9b1b7cd9eaeaa8093688a6e8bec47544a0734

DerAluDev:
> Programming-wise, everything is good and a great first PR.
> 
> I did run into the issue of aoe not being detected and I have an idea (albeit a bit complicated) for a solution. Firstly, [here is the log/pull](https://www.warcraftlogs.com/reports/FNDJwZcKjYpGTXQV#fight=7&type=damage-done&source=14) I was looking at. Around the 35 seconds mark, the boomkin did not have starfall up when the add was still alive, but was using starfire to cleave it during the boat window.
> 
> Possible Solution
> 
> * Create an array for the abilities used during boat.
> * Use the timestamp of the cast as the identifier for each one.
> * Use the damage events and `EventHistory` to search for the timestamp of the last cast within ~150ms to match the cast to the damage events.
> * Add the hit to the damage event for the cast.
> * Determine if it was a single target or multi-target cast, likely when the boat window ends since you then know how many targets were hit.
> 
> I don't know boomkin very well, so if these would still be considered bad casts that's fine. Just looking at an approach which isolates each individual boat window and checks them.

Thank you very much for your feedback and I am sorry for the improper testing if I havent found an example like this.

I took your advice and implemented something similiar. The new implementation looks for how many targets got hit at all during the BOAT-window, not accounting for the issuing spell. So far for my (hopefully this time sufficient testing), it works and accounts for each Buff-window on its own, rather than for the whole fight.

I'd be happy to receive feedback from you.

analysis/druidbalance/src/modules/features/BalanceOfAllThingsOpener.tsx
```diff
@@ -0,0 +1,173 @@
...
+      this.onApplyBuff,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.BALANCE_OF_ALL_THINGS_SOLAR),
+      this.onRemoveBuff,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.BALANCE_OF_ALL_THINGS_LUNAR),
+      this.onRemoveBuff,
+    );
```
DylanDirlam, line 52: For future reference, spell can accept an array of abilities so this could be condensed into:
```ts
Events.removebuff.by(SELECTED_PLAYER).spell([SPELLS.BALANCE_OF_ALL_THINGS_SOLAR, SPELLS.BALANCE_OF_ALL_THINGS_LUNAR])
```

DylanDirlam (approved):

PR #4409, opened 2021-02-21
[Balance] Unempowered spells

tiborhercz:
Add modules for Starfire/Wrath to track empowered casts.
It will take into account that there is always two empowered casts needed to get into eclipse.


MartijnHols (approved):
Thanks!

PR #4413, opened 2021-02-24
[druid] fix convoke crash

acornellier:
Fixes this bug reported in Discord: https://wowanalyzer.com/report/rfVHDZPQKAJc29bv/1-Mythic++Profondeurs+Sanguines+-+Kill+(34:44)/Natshsram/standard/overview

The druid starting casting convoke and immediately died, causing `startTracking` to be called, incrementing `this.cast`, then `stopTracking` to be called, but `this.whatHappendIneachConvoke[this.cast]` hadn't been initialized yet. I moved the initialization to `startTracking` to avoid any future issues like this.


DylanDirlam (approved):
Thanks for the fix and welcome to the project!

PR #4417, opened 2021-02-25
TS functional news components & related fixes

acornellier:
There were a couple bugs I found while refactoring the news components. I've put them in separate commits to make them clear.
1. Routing was not working - try https://wowanalyzer.com/news/Shadowlands+Prepatch
2. Articles were being filtered twice, resulting in them being hidden unless near the top

Performance on my PC went from taking 7ms to load `<News/>` to 3ms, so that's nice :)

Unfortunately, git has marked the files as deleted and added instead of modified... Any idea on how to fix this? I even did the `git mv` as a separate commit.

To get a proper diff in github, click on the commit instead of viewing all changes, and disable whitespace: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4417/commits/4a9137227468da67c2e5b9766b9a9478e670eaef?diff=split&w=1


MartijnHols (approved):
Sweet, thanks!

PR #4427, opened 2021-03-07
Updated Soul of the Forest talent for Shadowlands + statistic

DerAluDev:
Since Shadowlands this talent now gives +50% Astral Power generation when using Wrath while in Solar Eclipse and +150% AoE damage for Starfire while in Lunar Eclipse. 

In BFA it used to reduce AP cost on Starfall. I removed this calculation.

Additionally I added a statistic to track how much additional AP you gained from using this talent.
image

I did not include the statistic for the additional AoE Damage gained in this PR, in order to not bloat up it up and keep it small.


MartijnHols (approved):
Thanks!

PR #4428, opened 2021-03-07
[Mage] 9.0.5 Update

Sharrq:
Added/Changed the following to support 9.0.5

- Added support for Disciplinary Command and a check to ensure the buff covers 100% of Combustion (Requested by Mage Discord)
- Added a check for the amount of delay between using Combustion and letting your pre-cast finish (Requested by Mage Discord)
- Adjusted the Combustion Stat Box to show the Active Time and Pre-Cast Delay as well as a breakdown for each cast
- Added Disciplinary Command to the timeline buffs and Casts that arent casts (it was causing cancelled casts)
- Updated Arcane Harmony and Mirrors of Torment based on 9.0.5 changes.


MartijnHols (approved):
Nice, thanks!

PR #4429, opened 2021-03-07
Implemented suggestions in timeline for unempowered Wrath and Starfire

DerAluDev:
To get into an Eclipse you need to cast either two unempowered Starfire or Wrath, depending. Currently those two cast get marked as wrong in the timeline.

Old:
image

New:
image


MartijnHols (approved):
Thanks!

PR #4430, opened 2021-03-07
Add precommit hook with automatic code style fixes

### LINKED ISSUE #403: Implement Prettier
PRs are too often about code style, this is a waste of time. With Prettier this is automatically resolved.

I implemented Prettier in a React project in the past and code quality only degraded because it had terrible JSX support. I think this has since been improved enough, but not sure.

### LINKED ISSUE #4088: Update to react-scripts v4
The react-scripts v4 (create-react-app v4) provides several new development features as well as support for dropping the React import.

- Update `react-scripts` version constraint to `^4.0.0`
- Verify that development still works and automatically reloads the part of the page being changed. Take extra care verifying that analyzers properly trigger a re-analysis, since react-refresh might not work well with our project. If that is the case, consider disabling react-refresh as it has little benefit for us unless we build custom support for our analyzers.
- Update `@emico/eslint-config` version constraint to `^3.2.0`
- Fix new eslint issues

MartijnHols:
Fixes #4088 and #403


Khazakdk:
I get the argument for having an automatic code formatter, but the code outputted by prettier is just awful.  If there still isn't a way to disable it inserting linebreaks I think its a bad idea.  

The original issue calls out style comments as being a problem during PRs but looking over PRs merged this year, that doesn't seem to be much of an issue.  I see a couple of comments about it but most merged PRs that get multiple comments had issues deeper than line length.

Sharrq:
I agree with Khazzak. The line breaking is really really annoying and I would much rather prefer that it did not do it.

Like, im all for letting it fix indenting, and add semi colons and commas to stuff, or swapping out double quotes for single quotes .... all that stuff is fine .... But please for the love of god stop line breaking everything or splitting shit into multiple line stuff. It makes it harder for me to quickly read through stuff and understand it when its all spaced out vertically

MartijnHols:
There are many reasons for introducing Prettier, see https://prettier.io/docs/en/why-prettier.html for an extensive overview of them.

The goal isn't to get the prettiest code and no one will argue that Prettier achieves that. I was initially just as opposed to it but after working with it for nearly two years, I can read code outputted by Prettier just as well as code optimally formatted manually. If you're unfamiliar with it, that might take a while. Prettier is very popular in the JS community, so many developers are already familiar with its code style while my or your personal code style will have a learning curve for *everyone*.

Not seeing comments about code style in PRs does not mean there weren't any issues. I have merged many PRs with misplaced spaces, inconsistent indenting, mixed quotation marks, multiple expressions on a single line and many other issues because I do not want to hold up the PR process over these kinds of problems. The PR process is already slow and those kinds of comments are always frustrating.

You can find many other arguments for why Prettier is better than a manual code style at its site; https://prettier.io/docs/en/why-prettier.html. One of the important ones for me is https://prettier.io/docs/en/why-prettier.html#writing-code; `I want to write code. Not spend cycles on formatting.` One of the important ones for this project is that inconsistent code style is tech debt which needs to be reduced.

DylanDirlam:
I still agree that Prettier is a good change. It takes getting used to but then it just becomes a "type code and forget about having to format". It really is a matter of opinion in a lot of cases. To me, the change from:
```ts
const runicCorruptionContribution = this.addPassiveAccelerator(SPELLS.RUNIC_CORRUPTION.id, passiveRunesGained, passiveRunesWasted, RUNIC_CORRUPTION_INCREASE);
```
to
```ts
  const runicCorruptionContribution = this.addPassiveAccelerator(
    SPELLS.RUNIC_CORRUPTION.id,
    passiveRunesGained,
    passiveRunesWasted,
    RUNIC_CORRUPTION_INCREASE,
  );
```
looks much better and easier to read.

The only thing I notice that really bugs me is files like this where it re-alphabetizes all the imports when I had them organized by category. https://github.com/WoWAnalyzer/WoWAnalyzer/blob/eee58b9442ac8e3f0452995bfc0f2034ded2eff0/analysis/priestshadow/src/CombatLogParser.tsx

Sharrq:
I get that it's an opinion type thing and some people find it easier to read it when it's split across multiple lines..... But there are also us that find it harder to read stuff when it's like that.

But ultimately sounds like zero doesn't give a fuck what we think about it and is gonna do it anyway. So fuck us I guess.

Sharrq:
And again, I don't care about the other formating things mentioned in that wall of text zero, the question marks and missing quotes and such I'm fine with.

The splitting things into multiple lines and line breaks I am not

Khazakdk:
image

Adding to Sharrq's point, the above screenshot is an example of what I take issue with.  On the left you can see at a glance each event listener is starting the same way so your eyes immediately move to the end of the lines where you see the differences.  On the right, you have to read each line of each listener.  It takes me noticeably longer figure out whats going on.  

I disagree with the idea that code formatting is tech debt that needs addressing right now.  Spec support is already fairly dismal right now and I don't see inserting a bunch of line breaks and making code harder to read doing anything to fix that.  Stuff like converting to WCL v2 should take precedence over screwing with the readability of every spec when it comes to paying off tech debt.  Or updating our wiki to make spec contributions easier for those new to the project.

Sharrq:
> I disagree with the idea that code formatting is tech debt that needs addressing right now. Spec support is already fairly dismal right now and I don't see inserting a bunch of line breaks and making code harder to read doing anything to fix that. Stuff like converting to WCL v2 should take precedence over screwing with the readability of every spec.

100% this

acornellier:
Personally, I like the line splitting. Having to scroll horizontally to read code is very frustrating, especially when I'm working on my smaller screen and want to split vertically.

The example that Khazakdk showed is a great example. To me, the diff looks like this.
Screen Shot 2021-03-10 at 09 26 04

I'm all for the change, including the line splitting.

MartijnHols:
> I get that it's an opinion type thing and some people find it easier to read it when it's split across multiple lines..... But there are also us that find it harder to read stuff when it's like that.

The opposite is also true; long lines are harder to scan and read than if it gets separated by a newline. Just look at the two screenshots in this thread; it's much easier to scan and find the event handlers being added that if you first have to find the comma to know where the event handler starts. Again, please do check out https://prettier.io/docs/en/why-prettier.html for the many reasons why Prettier is better for a project, even if the different code style requires some getting used to and might not always lead to more readable code (although in my opinion both screenshots so far are better in the *after*).

> Adding to Sharrq's point, the above screenshot is an example of what I take issue with. On the left you can see at a glance each event listener is starting the same way so your eyes immediately move to the end of the lines where you see the differences. On the right, you have to read each line of each listener. It takes me noticeably longer figure out whats going on.

See above. On the right you can find the handler without first having to find where the second argument starts. This gets more important when the left side either has a variable amount of parameters or not all event listeners have the same amount of parameters (e.g. spell is omitted).

> But ultimately sounds like zero doesn't give a fuck what we think about it and is gonna do it anyway. So fuck us I guess.

The only feedback so far have been about the code style which is purely subjective. Prettier is going to take the subjectiveness out of code style. There are many other real reasons why Prettier is good for a project, please do check out https://prettier.io/docs/en/why-prettier.html.

> I disagree with the idea that code formatting is tech debt that needs addressing right now. Spec support is already fairly dismal right now and I don't see inserting a bunch of line breaks and making code harder to read doing anything to fix that. Stuff like converting to WCL v2 should take precedence over screwing with the readability of every spec when it comes to paying off tech debt. Or updating our wiki to make spec contributions easier for those new to the project.

Making it easier to get started in the codebase and develop things are steps towards making it easier for others to build spec support. No matter what I do there will always be something else that also needs to be done. I can't let that hold me back or nothing will get done ever. Prioritizing is hard, and doing anything at all gets hard when anything I do to try to reduce the tech debt and improve things structurally gets this much push back.

I do not see any significant gains to updating to WCL v2. I reckon adopting that would be akin to continuing to build new features without ever cleaning up. That's what I've done in the past few years and I regret that because of the tech debt it has left us with. It's time I clean up the mess I caused and make the codebase easier to maintain so the debt grow is reduced and easier to get started with for new contributors.

I will merge this now since the only feedback has been about the code style outputted by Prettier. Code style is subjective, and Prettier will permanently end the code style debate. It will take some getting used to, I have experienced that myself, but it's not as bad as you're making it out to be as you can see from the two other comments in the thread, my plea, and the thousands of other people using the Prettier code style in their projects. In addition none seemed to have an issue with all the other benefits it has which in my opinion are more significant.

Thanks for your feedback, it is appreciated.

PR #4432, opened 2021-03-08
Convert most Report related components to Typescript

acornellier:
This started off as a small project and ended up escalating very quickly. There's a LOT of files changed, I apologize for that. What I really wanted to do was convert the Loaders into functional components with hooks and `useSWR`. But I didn't feel comfortable doing that without proper typing, so I decided to simply convert to TS first, then make the actual logical changes later.

The good news is that the changes are almost entirely type related. I made very few logic changes - only when absolutely necessary, or if it would have been just horrible or wrong code without the change.

Changes
- Introduce `Report` and `CharacterProfile` types
- Convert nearly every file in `src/interface/report` and a few other common files to TS. There are a few remaining large files in `Results` and `Results/Timeline`
- Convert most of the redux store to TS
- Fix various types in existing components exposed by the above changes (like `Fight` -> `WCLFight`)


MartijnHols (approved):
Awesome, thanks a lot!

PR #4436, opened 2021-03-11
[DRAFT] useSWR for loading reports

acornellier:
This is not ready to merge, but I'd like active contributors to take a look and let me know what you think before I go on doing any more work on this.

I'd like to propose using [swr](https://swr.vercel.app/) for our API requests. It's a popular, modern, well-supported library for this purpose. It comes with many features, but the ones being used in this PR are
- state management using the hook
- error handling
- mutation (refresh)


emallson:
What is the goal of this change? does it introduce a fix for something like the report refreshing (which is probably the number 1 thing asked about in `#support`)? or is it just code cleanup?

emallson:
o.o

PR #4446, opened 2021-03-20
Astral Power usage recommendations now consider if Balance of All Things legendary is used

ghost:
When using the Balance of All Things legendary it is a top priority to cast three consecutive Starsurges when entering an eclipse. In order to have enough Astral Power for this, one is expected to overcap it a bit.

I looked at several top logs to decide on the numbers, they reached from 4% to 8% overcapped AP.


MartijnHols (approved):
Thanks!

PR #4447, opened 2021-03-20
Fixes WoWAnalyzer#4445 : FoodChecker now checks druid conduit Born Anew

DerAluDev:
When a druid has the https://www.wowhead.com/spell=341280/born-anew conduit and Battle-revives someone, the revived player gets a special Well Fed buff, gaining 30 Primary Stat.


MartijnHols (approved):
Thanks!

Issue #4433, opened 2021-03-09
Add the Guardian Druid Checklist


Phatso1973:
Currently there is no checklist available for guardian druid and should be added.

PR #4451, opened 2021-03-25
Added a basic checklist for guardian druid

Phatso1973:
This is for issue #4433.  I added a basic checklist and changed the touched files over to typescript.  From here I will need to investigate and fix Ironfur, Frenzied Regen, Incarnation, and EarthWarden as they seem to be showing bad data, or no data.


analysis/druidguardian/src/modules/features/AlwaysBeCasting.tsx
```diff
@@ -0,0 +1,29 @@
+import { t } from '@lingui/macro';
+import { formatPercentage } from 'common/format';
+import { When } from 'parser/core/ParseResults';
+import CoreAlwaysBeCasting from 'parser/shared/modules/AlwaysBeCasting';
+import React from 'react';
+
+class AlwaysBeCasting extends CoreAlwaysBeCasting {
```
emallson, line 7: generally you don't need to override this unless you are adding custom functionality. what differs between this and the base class?

emallson (commented):
small comment that i'd like clarified. otherwise it looks fine for a start

analysis/druidguardian/src/modules/features/AlwaysBeCasting.tsx
Phatso1973: Thanks for the review @emallson.  So I added this override here because I wish to expand on the ABC suggestions area for guardian with some tips on how to fill the gaps with MF/IF, and Swipe.  This is a precursor to more changes I am working on.

PR #4459, opened 2021-04-06
Discipline Priest Maintenance

rp4rk:
# Description
Updates Discipline with a fancy new event emitting Spirit Shell module that allows for previous modules to work seamlessly between standard Atonement <--> Spirit Shell.

Read comments for reasoning, this is a special one


src/parser/priest/discipline/constants.ts
```diff
@@ -42,7 +65,39 @@ export const POWER_WORD_SHIELD_ATONEMENT_DUR = 15000;
 export const POWER_WORD_RADIANCE_ATONEMENT_DUR = 9000;
 export const SHADOW_MEND_ATONEMENT_DUR = 15000;
 
-// https://www.wowhead.com/spell=137032/discipline-priest
-export const DISC_PRIEST_DAMAGE_REDUCTION = 0.67;
```
rp4rk: No longer used

src/parser/priest/discipline/modules/core/AtonementAnalyzer.ts
```diff
@@ -0,0 +1,129 @@
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
```
rp4rk: One of the two core drivers behind a lot of the spec functionality, emits AtonementEvents as pairs with a source DamageEvent and a resulting HealEvent. Has a dependency on the Spirit Shell module to emit events with the same contract regardless of final tier talent use.

src/parser/priest/discipline/modules/core/SpiritShell.ts
```diff
@@ -0,0 +1,299 @@
+import SPELLS from 'common/SPELLS';
```
rp4rk: This is a special one, there is no combat log event for adding more absorption to the Spirit Shell effect - so we reverse engineer that using the `unmitigatedAmount` field combined with knowledge of disc modifiers.

This module emits Spirit Shell specific events which are then picked up by the AtonementAnalyzer module which is used as a source for (hopefully) all atonement analysis across the disc spec.

analysis/priest/src/Mindgames.tsx
```diff
@@ -45,7 +45,10 @@ class Mindgames extends Analyzer {
   constructor(options: Options) {
     super(options);
 
-    this.active = this.selectedCombatant.hasCovenant(COVENANTS.VENTHYR.id);
```
rp4rk, line 48: The generic priest Mindgames has a bunch of mixed concerns that really don't mesh well with the Discipline implementation

analysis/priestdiscipline/src/CONFIG.tsx
```diff
@@ -7,10 +7,10 @@ import CHANGELOG from './CHANGELOG';
 
 const config: Config = {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [Khadaj, Oratio],
+  contributors: [Khadaj, Oratio, Reglitch],
   // The WoW client patch this spec was last updated.
-  patchCompatibility: null,
-  isPartial: true,
+  patchCompatibility: '9.0.5',
```
rp4rk, line 12: Spec is around 90% of the way there, it's definitely at the point you can use it to analyze choices. More stuff on the way!

PR #4461, opened 2021-04-07
Update the RtB Tracker for Outlaw Rogue.

### LINKED ISSUE #4239: [SUGGESTIONS] Roll the Bones
- Use first Roll the Bones 1 Second before Pull (2 seconds if talented into Marked for Death)

- Reroll single True Bearing, Buried Treasure and Grand Melee as soon as possible.

- Reroll other buffs with 3 seconds left.

- In the current RTB implementation, unsure what exactly “low/mid/high value efficiency” is describing. The player either has to reroll ASAP, or they don’t, depending on what buffs are active. Perhaps track how long RTB was off cooldown when a single True Bearing, Buried Treasure or Grand Melee are active -- instances where they should have rerolled.

> Outdated script seems to work okay for the basic needs, but the buff priority had changed according to the following logic:
> 
> (cumulative value) ≤ 2 - recast
> (cumulative value) > 2 - wait out full duration
> 
> Broadside (BS) = 3
> True Bearing (TB) = 3
> Ruthless Precision (RP) = 2
> Skull and Crossbones (SC) = 2
> Buried Treasure (BT) = 1
> Grand Melee (GM) = 1
> 
> However, if Sleight of Hand conduit is being used, the values change to:
> 
> BS = 2
> TB = 2
> Rest stays the same
> 
> (This is because if using the conduit, we should always try to get 2+ buffs but not BT+GM)
-- @tomovka

alex-bau:
* Update High Value buffs to Broadside and True Bearing. The buff prio
  changed with SLs, but this isn't reflected in the analyzer.

* Use values and a sum to track if a roll is low or high values (as
  opposed to ifing for high value rolls).

* Incorporate Sleight of Hand into the sims by punishing people who hold
  onto single value rolls (even if they _should_ be high value rolls).

* Punish holding onto GM+BT (the two worst buffs for OLW) if rolled
  together.

### DIFF
--- a/analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
+++ b/analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
@@ -9,9 +9,23 @@ import { ROLL_THE_BONES_BUFFS, ROLL_THE_BONES_DURATION } from '../../constants';
 
 export const ROLL_THE_BONES_CATEGORIES = {
   LOW_VALUE: 'low',
-  MID_VALUE: 'mid',
   HIGH_VALUE: 'high',
 };
+
+let BROADSIDE_VALUE = { sleight_of_hand: 2, base: 3 };
+let TRUE_BEARING_VALUE = { sleight_of_hand: 2, base: 3 };
+let RUTHLESS_PRECISION_VALUE = { sleight_of_hand: 2, base: 2 };
+let SKULL_AND_CROSSBONES_VALUE = { sleight_of_hand: 2, base: 2 };
+let BURIED_TREASURE_VALUE = { sleight_of_hand: 1, base: 1 };
+let GRAND_MELEE_VALUE = { sleight_of_hand: 1, base: 1 };
+
+var BUFF_VALUE_BY_ID = {};
+BUFF_VALUE_BY_ID[SPELLS.BROADSIDE.id] = BROADSIDE_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.TRUE_BEARING.id] = TRUE_BEARING_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.RUTHLESS_PRECISION.id] = RUTHLESS_PRECISION_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.SKULL_AND_CROSSBONES.id] = SKULL_AND_CROSSBONES_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.BURIED_TREASURE.id] = BURIED_TREASURE_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.GRAND_MELEE.id] = GRAND_MELEE_VALUE;
 
 // e.g. 1 combo point is 12 seconds, 3 combo points is 24 seconds
 const PANDEMIC_WINDOW = 0.3;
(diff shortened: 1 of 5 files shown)


alex-bau:
This implements the the prio logic mentioned in https://github.com/WoWAnalyzer/WoWAnalyzer/issues/4239#issue-774022537 wrt buff Prio and SoH logic. Removing the mid roll buffs (which didn't really make sense) makes it so we either keep buffs for the whole duration or we penalise the user for not re-rolling ASAP.

analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
```diff
@@ -9,10 +9,24 @@ import { ROLL_THE_BONES_BUFFS, ROLL_THE_BONES_DURATION } from '../../constants';
...
+let GRAND_MELEE_VALUE = { sleight_of_hand: 1, base: 1 };
+
+var BUFF_VALUE_BY_ID = {};
+BUFF_VALUE_BY_ID[SPELLS.BROADSIDE.id] = BROADSIDE_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.TRUE_BEARING.id] = TRUE_BEARING_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.RUTHLESS_PRECISION.id] = RUTHLESS_PRECISION_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.SKULL_AND_CROSSBONES.id] = SKULL_AND_CROSSBONES_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.BURIED_TREASURE.id] = BURIED_TREASURE_VALUE;
+BUFF_VALUE_BY_ID[SPELLS.GRAND_MELEE.id] = GRAND_MELEE_VALUE;
+
```
emallson, line 29: use `const` instead of `let` if you're not going to be reassigning the value. `var` should not be used in general, use `let` or `const` instead.

`var` has global or function scope, not lexical scope (as used by most languages by default), which can lead to very surprising behaviors.

in this particular case, it looks like these could all be `const`, but i'm not going to nitpick as long as the use of `var` is changed.

analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
```diff
@@ -53,14 +67,18 @@ class RollTheBonesCastTracker extends Analyzer {
   }
 
   categorizeCast(cast) {
-    if (
-      cast.appliedBuffs.some(
-        (buff) => buff.id === SPELLS.RUTHLESS_PRECISION.id || buff.id === SPELLS.GRAND_MELEE.id,
-      )
-    ) {
+    var combat_buffs_value = 0;
```
emallson: use `let` here

emallson (changes requested):
couple uses of `var` that should be changed to `let` or `const`. Otherwise looks good :+1:

analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
alex-bau: TIL, thanks for the rundown :)

analysis/rogueoutlaw/src/modules/features/RollTheBonesCastTracker.js
```diff
@@ -12,14 +12,14 @@ export const ROLL_THE_BONES_CATEGORIES = {
...
-
-var BUFF_VALUE_BY_ID = {};
+const BROADSIDE_VALUE = { sleight_of_hand: 2, base: 3 };
+const TRUE_BEARING_VALUE = { sleight_of_hand: 2, base: 3 };
+const RUTHLESS_PRECISION_VALUE = { sleight_of_hand: 2, base: 2 };
+const SKULL_AND_CROSSBONES_VALUE = { sleight_of_hand: 2, base: 2 };
+const BURIED_TREASURE_VALUE = { sleight_of_hand: 1, base: 1 };
+const GRAND_MELEE_VALUE = { sleight_of_hand: 1, base: 1 };
+
+let BUFF_VALUE_BY_ID = {};
```
emallson, line 22: fun fact: assigning props to an object works on `const` objects. `const` only controls whether you can use `=`. for example:

```js
const a = {};
a.foo = 1;
```
works but
```js
const a = null;
a = 1;
```
does not.

not gonna block the review over it but its good to know

emallson (approved):

PR #4466, opened 2021-04-10
[Druid] Restore anti-filler reporting

abbottmg:
Remove null default for antiFillerSpam prop which was being set *after*
the superclass's constructor was called, probably in part because this class
is not Typescript.

As indicated in previous discussion, it's probably best to stop jumping
through hoops to get this additional data into SpellbookAbility child objects
and just provide it as a parallel type created within guardiandruid, but
that is a bigger change and I am still learning the architecture of this
project, so I thought it worthwhile to stabilize this feature in the
meantime.

### DIFF
--- a/analysis/druidguardian/src/CHANGELOG.tsx
+++ b/analysis/druidguardian/src/CHANGELOG.tsx
@@ -2,6 +2,7 @@ import { change, date } from 'common/changelog';
 import { Kettlepaw, Zeboot, g3neral, Tiboonn, Buudha } from 'CONTRIBUTORS';
 
 export default [
+  change(date(2020, 4, 7), 'Updated class syntax to re-enable AntiFillerSpam info', Kettlepaw),
   change(date(2021, 4, 7), 'Correct reporting for Earthwarden absorb events', Kettlepaw),
   change(date(2021, 3, 25), 'Added basic checklist section to be expanded on, and upgraded touched files to Typescript', Buudha),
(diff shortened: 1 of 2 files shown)


analysis/druidguardian/src/CHANGELOG.tsx
```diff
@@ -2,6 +2,7 @@ import { change, date } from 'common/changelog';
 import { Kettlepaw, Zeboot, g3neral, Tiboonn, Buudha } from 'CONTRIBUTORS';
 
 export default [
+  change(date(2020, 4, 7), 'Updated class syntax to re-enable AntiFillerSpam info', Kettlepaw),
```
DylanDirlam: We have a time traveler! :wink:

rp4rk (approved):
LGTM :)

analysis/druidguardian/src/CHANGELOG.tsx
abbottmg: Ha I decided to wait for my other 4/7 change to get merged and then forgot to actually cut this PR for a bit. I don't mind postdating it!

DylanDirlam: I was mostly joking about the 2020, that's all 😛 I've done it before

abbottmg: Oh no! I didn't even notice that

PR #4468, opened 2021-04-12
[Priest] Fae Guardians 9.0.5 changes and DR correction

DylanDirlam:
The DR calculation was not calculating properly, this fixes it and updates to the new 20% DR from fae guardians changed in 9.0.5. Calculations below.

OLD - 10% DR calculation used:
https://www.warcraftlogs.com/reports/Q8MBTt6WVAXwLZNx#fight=last&type=auras&source=7&ability=327694
image

Now if we use a wclogs filter to damage taken to the player when he had faerie guardians up: [Link](https://www.warcraftlogs.com/reports/Q8MBTt6WVAXwLZNx#fight=last&type=damage-taken&source=7&pins=2%24Separate%24%23244F4B%24damage%241%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%240%24and%24auras-gained%240%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24327694%24true%24false%2463)

The total damage is 264.4k. Using the previous calculation there is no way that 10% DR would be 204k.

NEW - 20% DR w/ new calculation used, half it to get the old value of what it should be:
image

### DIFF
--- a/analysis/priest/src/FaeGuardians.tsx
+++ b/analysis/priest/src/FaeGuardians.tsx
@@ -4,6 +4,7 @@ import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import COVENANTS from 'game/shadowlands/COVENANTS';
 import SPECS from 'game/SPECS';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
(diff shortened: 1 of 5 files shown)


rp4rk (approved):
LGTM, nice catch with the damage taken calculation

PR #4469, opened 2021-04-12
[Rogue] Fixed a bug where master assassin was showing 0 good casts always

sspeaks:
Before:
image

After:
image

Log source: https://www.warcraftlogs.com/reports/yW6nMV97kpTBFGwb#fight=10

Character name: Foxypoke


DylanDirlam (approved):
Thanks for the contribution!

PR #4470, opened 2021-04-14
Updated StatTracker (in core) and Mastery (in druidrestoration) to TypeScript

kfinch:
Updated StatTracker and (druidrestoration) Mastery to TS. Getting things setup in WebStorm also pushed some changes to the .idea folder (which specifically handles WebStorm formatting stuff) - I think all the changes are correct with regards to our existing ESLint / TSLint / Prettier, and so should remain in this PR.


kfinch:
Ah, the Brewmaster integration test was an effective canary - something from StatTracker reporting wrong caused it. I'll investigate and submit a fix.

.idea/compiler.xml
```diff
@@ -2,6 +2,6 @@
 <project version="4">
   <component name="TypeScriptCompiler">
     <option name="recompileOnChanges" value="true" />
-    <option name="typeScriptCompilerParams" value="--noEmit" />
+    <option name="typeScriptCompilerParams" value="--noEmit --target es5 --jsx react" />
```
kfinch, line 5: I needed to add this for it to build in WebStorm, so if nothing else from the .idea changes make it in, this should.

.idea/codeStyles/Project.xml
```diff
@@ -22,7 +22,6 @@
     <JSCodeStyleSettings version="1">
```
tyndivelspaz, line 22: I don't use anything by this company anymore but should we be updating these any `.idea` files?

.idea/compiler.xml
tyndivelspaz: This is an exception to my previous comment about changing the `.idea` files.

analysis/druidrestoration/src/modules/core/Mastery.ts
```diff
@@ -0,0 +1,334 @@
...
+    } else {
+      this.totalNoMasteryHealing += healVal.effective;
+    }
+  }
+
+  onAbsorbed(event: AbsorbedEvent): void {
+    this.totalNoMasteryHealing += event.amount;
+  }
+
+  onFightEnd(_: FightEndEvent): void {
```
tyndivelspaz, line 120: Do we need to pass an event argument to this method since we aren't using it and the listener is already listening for the FightEndEvent?

.idea/codeStyles/Project.xml
kfinch: I'm honestly not sure, it's possible this should even be added to the gitignore

analysis/druidrestoration/src/modules/core/Mastery.ts
kfinch: I think you're right that JS allows this to be written with no arguments. It was a stylistic choice to use the underscore indicating "I acknowledge I was passed the event, but I won't be using it"

tyndivelspaz: I wasn't sure if the linter would complain about it or not. I've seen it done both ways and have no preference, was more of a 'appease the linter overlord' comment.

kfinch: The linter automatically runs on pull requests. The "all checks have passed" on this PR indicates it's passed all linting.

.idea/codeStyles/Project.xml
kfinch: Zero says he prefers the .idea to stay for more uniformity with users of WebStorm. I backed out some of the changes I didn't intend but left in ones that look valuable.

emallson (approved):

PR #4471, opened 2021-04-17
[Rogue][Refactor] Update Outlaw Rogue to Typescript

hpabst:
## Notes
- Update existing Outlaw Rogue modules to Typescript in anticipation of additional future work.
- This PR is a no-op in terms of behaviour change.

## Testing
- `yarn build && yarn test:integration`

## Screenshots
- Testing log used: https://www.warcraftlogs.com/reports/KVA8F7phqwagt9Nm#fight=4&type=damage-done&source=15&translate=true

### Before Changes:
WowAnalyzerRefactorBefore

### After Changes:
WowAnalyzerRefactorAfter


analysis/rogueoutlaw/src/modules/features/Finishers.tsx
```diff
@@ -31,7 +31,7 @@ class Finishers extends FinisherTracker {
     return points;
   }
 
-  extraSuggestion() {
+  extraSuggestion(): React.ReactElement | any {
```
kfinch: how is `X | any` different from `any`? What type are you returning here that requires use of any?

analysis/rogueoutlaw/src/modules/spells/OpportunityDamageTracker.ts
```diff
@@ -3,7 +3,7 @@ import SPELLS from 'common/SPELLS';
 import { FilteredDamageTracker } from '@wowanalyzer/rogue';
 
 class OpportunityDamageTracker extends FilteredDamageTracker {
-  shouldProcessEvent(event) {
+  shouldProcessEvent(event: any): boolean {
```
kfinch: Should be able to use base Event type here instead of `any`

analysis/rogueoutlaw/src/modules/spells/RollTheBonesCounter.tsx
```diff
@@ -12,11 +12,12 @@ class RollTheBonesCounter extends Analyzer {
   static dependencies = {
     rollTheBonesCastTracker: RollTheBonesCastTracker,
   };
+  protected rollTheBonesCastTracker!: RollTheBonesCastTracker;
 
   rolltheBonesBuffDistributionChart() {
     const castTracker = this.rollTheBonesCastTracker;
 
-    const distributionObj = castTracker.rolltheBonesCastEvents.reduce((buffCount, cast) => {
+    const distributionObj = castTracker.rolltheBonesCastEvents.reduce((buffCount: any, cast) => {
```
kfinch: Should be `buffCount: RTBCast[]` right? Maybe could use a better var name too.

analysis/rogueoutlaw/src/CombatLogParser.ts
```diff
@@ -9,7 +9,6 @@ import {
   EssenceOfBloodfang,
   Flagellation,
   InvigoratingShadowdust,
-  Sepsis,
```
kfinch, line 12: Why removed?

kfinch (commented):
Overall looks good. A couple places where I think `any` could be narrowed down. Also you didn't address why Sepsis module was removed.

analysis/rogueoutlaw/src/CombatLogParser.ts
hpabst: This is an oversight. This PR started with adding onto the Sepsis module, then converting that into Typescript, then converting all of Outlaw into Typescript. I'll readd it in the following revision.

analysis/rogueoutlaw/src/modules/spells/RollTheBonesCounter.tsx
hpabst: Nope, `buffCount` is an indexable object of how many times a "buff level" from Roll the Bones procced. Essentially a `{[index : number] : number}`. I'll put this as the type rather than `any` to be more specific.

analysis/rogueoutlaw/src/modules/spells/OpportunityDamageTracker.ts
hpabst: Done

analysis/rogueoutlaw/src/modules/features/Finishers.tsx
hpabst: The issue is that `extraSuggestion()` here overrides `extraSuggestion()` from `FinisherTracker` in the base Rogue module that has not yet been converted  to Typescript. I can narrow this down to `React.ReactElement | string` by converting that module to Typescript.

kfinch: Leaving it as `any` is fine if you don't want to deal with in this PR, just leave a TODO comment

hpabst: Already did the conversion, it's a small module.

kfinch (approved):
noice

PR #4473, opened 2021-04-20
Update README to use current profile api key wording

wmavis:


MartijnHols (approved):
Lgtm thanks. Just needs that changelog entry

PR #4477, opened 2021-04-25
[Rogue] Add CD efficiency and follow-up ability tracking for Sepsis

### LINKED ISSUE #4233: [SUGGESTIONS] Sepsis
Use on cooldown without overcapping Combo Points. Use Ambush without overcapping as the use of an stealth ability granted when Sepsis runs out

hpabst:
### Notes
- Add CD efficiency tracking to Sepsis analyzer.
- Add CD reduction tracking to Sepsis analyzer.
- Add new analyzer, `StealthAbilityFollowingSepsis` to track proper stealth ability usage following Sepsis DoT completion.

### Screenshots:
image
image
image

### DIFF
--- a/analysis/rogue/src/covenants/nightfae/Sepsis.tsx
+++ b/analysis/rogue/src/covenants/nightfae/Sepsis.tsx
@@ -36,17 +36,16 @@ class Sepsis extends Analyzer {
       Events.damage.by(SELECTED_PLAYER).spell(SPELLS.SEPSIS_POISON),
       this.onPoisonDamage,
     );
-    this.addEventListener(Events.removedebuff.spell(SPELLS.SEPSIS), this.onDebuffFade);
+    this.addEventListener(
+      Events.removedebuff.by(SELECTED_PLAYER).spell(SPELLS.SEPSIS),
+      this.onDebuffFade,
+    );
   }
 
   onDebuffFade(event: RemoveDebuffEvent) {
     if (this.lastSepsisCast == null) {
       // This should never happen and indicates a bug if it does.
       debug && console.warn('Sepsis debuff faded before cast recorded.');
-      return;
-    }
-    if (event.sourceID !== this.selectedCombatant.id) {
-      // Debuff faded from Sepsis cast by another Rogue.
       return;
     }
     const lastCastTimestamp = this.lastSepsisCast.timestamp;
--- a/analysis/rogue/src/covenants/nightfae/Sepsis.tsx
+++ b/analysis/rogue/src/covenants/nightfae/Sepsis.tsx
@@ -2,30 +2,66 @@ import { formatNumber } from 'common/format';
 import SPELLS from 'common/SPELLS';
 import COVENANTS from 'game/shadowlands/COVENANTS';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { DamageEvent } from 'parser/core/Events';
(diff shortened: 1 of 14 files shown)


hpabst:
For reference, these logs were used in testing:

https://www.warcraftlogs.com/reports/q2TQCpfX3KB7trFY#fight=6&type=damage-done&source=14
https://www.warcraftlogs.com/reports/HRN6fDc7bnATqwL1#fight=77&type=damage-done&source=746
https://www.warcraftlogs.com/reports/XybvWxgnZBPzwHFV#fight=18&type=damage-done&source=149

analysis/rogue/src/covenants/nightfae/Sepsis.tsx
```diff
@@ -2,30 +2,66 @@ import { formatNumber } from 'common/format';
...
+  onDebuffFade(event: RemoveDebuffEvent) {
+    if (this.lastSepsisCast == null) {
+      // This should never happen and indicates a bug if it does.
+      debug && console.warn('Sepsis debuff faded before cast recorded.');
+      return;
+    }
+    if (event.sourceID !== this.selectedCombatant.id) {
+      // Debuff faded from Sepsis cast by another Rogue.
+      return;
+    }
```
kfinch: Can skip this check by adding a `by` check to the EventListener, like is already present on the other event listeners.

analysis/rogue/src/covenants/nightfae/StealthAbilityFollowingSepsis.tsx
```diff
@@ -0,0 +1,114 @@
...
+        >
+          <BoringSpellValue
+            spell={SPELLS.SEPSIS}
+            value={`${this.badCasts.length}`}
+            label="Bad Casts after Sepsis Finishes"
+          />
+        </Statistic>
+      </>
+    );
+  }
```
kfinch: Am I correctly reading that this will show even if you have zero bad casts? It might look a little confusing in that case.

kfinch (changes requested):
Overall looks good, just a couple comments.

analysis/rogue/src/covenants/nightfae/StealthAbilityFollowingSepsis.tsx
tyndivelspaz: I don't think hiding this on 0 bad casts would be beneficial. I'd like to think that seeing I had 0 bad casts would give immediate positive reinforment.

kfinch: I guess I was thinking the tooltip would be weird, but I suppose you're right. It's up to @hpabst to change or leave as is.

tyndivelspaz (approved):
All looks good to me. You can address the tooltip comment or leave it be. I think it can go either way

analysis/rogue/src/covenants/nightfae/Sepsis.tsx
hpabst: Will resolve in the next revision.

analysis/rogue/src/covenants/nightfae/StealthAbilityFollowingSepsis.tsx
hpabst: I agree in that the tooltip could be kind of weird along with the dropdown if there are 0 bad casts. I'll adjust this in the next revision to remove the dropdown and have a different tooltip if there are 0 bad casts.

kfinch (approved):
noice

PR #4479, opened 2021-04-26
Events: Add space to Demon Hunter Class

nicholasRutherford:
# Description

In [SPECS.ts](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/src/game/SPECS.ts#L711) they use `Demon Hunter` (with a space) which was causing the[ class string comparison in ReportRaidBuffList.tsx](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/src/interface/ReportRaidBuffList.tsx#L55) to fail. This now matches the class string formatting of Death Knight as well (which was fixed in cd93544).

# Motivation
I noticed that the Chaos Brand raid buff was incorrectly reporting. For [this report](https://wowanalyzer.com/report/7ZwWpynfCtRaQMTA/8-Normal+Sludgefist+-+Kill+(4:35)) it was saying 0, even though we had three demon hunters.

<details>
<summary>Current production screenshot</summary>

image
(Note the 0 for Chaos Brand despite having three demon hunters)
</details>

<details>
<summary>Local fix screenshot</summary>

image
(Note the 3 for Chaos Brand for having three demon hunters)
</details>

# Testing
I checked the player selection screen locally. I also searched for everywhere else we use `Class`, which looks to only be here. So everything should be fine? Let me know if there is more I should be testing!

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -1,7 +1,7 @@
 import { change, date } from 'common/changelog';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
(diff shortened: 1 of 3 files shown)


kfinch:
Good catch, thanks for the contribution. All pull requests require an entry in the CHANGELOG.tsx, so please add that and then we can merge.

kfinch (approved):

PR #4481, opened 2021-04-28
Converted HotTracker and all the rest of the Resto Druid modules to TypeScript

kfinch:
I've converted HotTracker and all the remaining Resto Druid modules to TypeScript. There should be no functional changes here. A lot of HotTracker is no longer being used, but I kept it mostly intact because I plan to start using it again.

Some oddities were required in this conversion, they will be commented on below.


.idea/compiler.xml
```diff
@@ -2,6 +2,6 @@
 <project version="4">
   <component name="TypeScriptCompiler">
     <option name="recompileOnChanges" value="true" />
-    <option name="typeScriptCompilerParams" value="--noEmit --target es5 --jsx react" />
+    <option name="typeScriptCompilerParams" value="--noEmit --target es5 --jsx react --esModuleInterop true" />
```
kfinch, line 5: WebStorm doesn't seem to like to read the tsconfig.json and looks at its own .idea file. Copied more of that in here to make it work properly.

analysis/druidrestoration/src/constants.ts
```diff
@@ -40,6 +40,7 @@ export const ABILITIES_AFFECTED_BY_HEALING_INCREASES_SPELL_OBJECTS = [
   //TODO - blazyb double check if any non resto druid spells scales with healing increases.
 ];
 
+// TODO this is OBE in Shadowlands
```
kfinch, line 43: Will address this in future work.

analysis/druidrestoration/src/modules/core/Rejuvenation.ts
```diff
@@ -33,36 +33,43 @@ class Rejuvenation extends Analyzer {
...
     mastery: Mastery,
   };
+
+  protected healingDone!: HealingDone;
+  protected mastery!: Mastery;
+
   totalRejuvsCast = 0;
 
-  constructor(options) {
+  constructor(options: Options) {
```
kfinch, line 42: Commented out unused bits (below) to avoid "unused" warning. Will be revived in future work.

analysis/monkmistweaver/src/modules/core/HotTrackerMW.tsx
```diff
@@ -23,37 +23,39 @@ class HotTrackerMW extends HotTracker {
     this.upwellingActive = selectedCombatant.hasTalent(SPELLS.UPWELLING_TALENT.id);
   }
 
-  calculateMaxDuration(event: ApplyBuffEvent & { selectedCombatant: Combatant }) {
+  calculateMaxDuration(event: ApplyBuffEvent, combatant: Combatant) {
```
kfinch: tacking an extra field onto the existing passed Event made no sense and is very awkward in TypeScript. I instead just made the combatant an extra parameter.

analysis/monkmistweaver/src/modules/core/HotTrackerMW.tsx
```diff
@@ -23,37 +23,39 @@ class HotTrackerMW extends HotTracker {
...
     ) {
       return (REM_BASE_DURATION + TFT_REM_EXTRA_DURATION) * 2;
     }
     return REM_BASE_DURATION * 2;
   }
 
-  calculateMaxRemDuration(
-    event: (ApplyBuffEvent | RefreshBuffEvent) & { selectedCombatant: Combatant },
-  ) {
+  calculateMaxRemDuration(event: ApplyBuffEvent | RefreshBuffEvent, combatant: Combatant) {
```
kfinch: tacking an extra field onto the existing passed Event made no sense and is very awkward in TypeScript. I instead just made the combatant an extra parameter.

analysis/monkmistweaver/src/modules/core/HotTrackerMW.tsx
```diff
@@ -23,37 +23,39 @@ class HotTrackerMW extends HotTracker {
...
+  calculateMaxDuration(event: ApplyBuffEvent, combatant: Combatant) {
     const spellId = event.ability.guid;
     if (spellId === SPELLS.ENVELOPING_MIST.id) {
-      return (ENV_BASE_DURATION + (this.mistwrapActive ? MISTWRAP : 0)) * 2;
+      return (
+        (ENV_BASE_DURATION + (combatant.hasTalent(SPELLS.MIST_WRAP_TALENT.id) ? MISTWRAP : 0)) * 2
+      );
     }
     if (spellId === SPELLS.ESSENCE_FONT_BUFF.id) {
-      return (EF_BASE_DURATION + (this.upwellingActive ? UPWELLING : 0)) * 2;
```
kfinch, line 32: A very strange issue in Typescript was making `this` undefined inside this function. I couldn't figure how to resolve it so I simply changed references to `this` to get at talents through the passed in Combatant instead

src/parser/shared/modules/HotTracker.ts
```diff
@@ -18,27 +28,25 @@ const extensionDebug = false; // logs pertaining to extensions
 const applyRemoveDebug = false; // logs tracking HoT apply / refresh / remove
 const healDebug = false; // logs tracking HoT heals
 
-class HotTracker extends Analyzer {
+abstract class HotTracker extends Analyzer {
```
kfinch, line 57: This whole class is a mess and really needs a refactor (or just replacement). A lot of its features are currently unused. I intend to do work on it in the future, but for now I preserved as much of it as I could and relearned a lot of what was going on in the process of adding types.

src/parser/shared/modules/HotTracker.ts
```diff
@@ -81,23 +89,24 @@ class HotTracker extends Analyzer {
...
+      const c: Combatant = this.selectedCombatant;
+      maxDuration = maxDurationFunc(event, c);
     }
 
     const newHot = {
       start: event.timestamp,
       end: event.timestamp + hotDuration,
       originalEnd: event.timestamp + hotDuration,
       spellId, // stored extra here so I don't have to convert string to number like I would if I used its key in the object.
-      guid: this.currentGuid,
```
kfinch: this was some sort of 'counting' functionality that was totally unused and served no clear purpose, hence why I removed it.

src/parser/shared/modules/HotTracker.ts
```diff
@@ -530,21 +540,74 @@ class HotTracker extends Analyzer {
...
+// A mapping from spell ID to that spell's HotInfo
+export type HotInfoMap = { [key: number]: HotInfo };
+
+// information about a Heal over Time spell specific to tracking
+export interface HotInfo {
+  duration: number; //HoT's base duration, in ms
+  tickPeriod: number; //HoT's base period between ticks, in ms
+  maxDuration?: (e: ApplyBuffEvent, c: Combatant) => number; // TODO used by MW - description?
+  bouncy?: boolean; // true iff a hot will bounce to another target on any event
+  durationConditions?: (e: ApplyBuffEvent | RefreshBuffEvent, c: Combatant) => number; // TODO used by MW - description?
```
kfinch: I think Abelito was the one who added this functionality - not super clear on why it's needed or how it works, but I tried to leave it untouched. I confirmed that the Mistweaver module still returns the same results afterwards.

kfinch (commented):
Comments on weirdness:

src/parser/shared/modules/HotTracker.ts
abelito75: I'm not working on this project anymore but you can look at HotTrackerMW.tsx for an example of durationConditions. Not all hots have a 1 constant duration. Thunder focus Tea increases Renewing Mists duration by 10 seconds.
Then Rising mist has a limit of only being able to double its init duration so TFT -> rem would have a max of 60 seconds compared to the normal maximum of 40. 
So this was added to the core incase blizzard ever uses it again or if you want to do pandemic checking or something generically

src/parser/shared/modules/HotTracker.ts
abelito75: This was more for debugging purposes than anything else as having to find a specific hot that is breaking this can be painful.

src/parser/shared/modules/HotTracker.ts
abelito75: As a note. Combatant needs to be passed in or it will not be able to find it in the function its calling.

src/parser/shared/modules/HotTracker.ts
kfinch: ahh that makes sense - maybe I'll find myself adding it again some day

src/parser/shared/modules/HotTracker.ts
kfinch: Thanks for giving the context. I spent some more time looking at it and had mostly figured this out but wasn't confident enough to write the definitive docs for it. I'll update the docs with your info.

And yeah, I discovered the thing with Combatant - but I think it was harmless to make Combatant a separate field rather than adding it to the Event. TypeScript gets upset when you try to tack things to defined objects.

kfinch:
Based on additional context from Abelito, I refactored the maxDuration to be a bit cleaner. Discovered the functions don't actually need the ApplyBuff event because the spell info is known based on the HoT it's being called on. I was also able to roll conditional duration into the duration field itself.

hpabst (approved):
LGTM in terms of a TS conversion, but as stated in discord I have little understanding of healer mechanics and thus the how and why of HotTracker. If you're concerned about changes in behavior from your changes, you might want a second reviewer.

PR #4482, opened 2021-05-01
[Rogue][Outlaw] Add Celerity Legandary Analyzer

### LINKED ISSUE #4229: [SUGGESTIONS] Implement Celerity Legendary
You should use Adrenaline Rush on cooldown, except when you already have Adrenaline Rush running (which only happens with celerity) because it refreshes the duration instead of extending it

hpabst:
### Notes
- Add Celerity legendary tracker to track "bad casts" of Adrenaline Rush when the legendary has already provided the buff.
- Resolves #4229 

### Testing Logs
https://www.warcraftlogs.com/reports/qDjpBv219KHR8XWP#fight=3&type=damage-done&source=8

### Screenshots

image

### DIFF
--- a/analysis/rogueoutlaw/src/CHANGELOG.js
+++ b/analysis/rogueoutlaw/src/CHANGELOG.js
@@ -6,6 +6,7 @@ import React from 'react';
 
 
 export default [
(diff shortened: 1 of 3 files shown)


kfinch (approved):
Look good to me

PR #4488, opened 2021-05-08
Updates to HotTracker Extensions and Resto Druid Flourish

kfinch:
Major changes by class:

HotTracker
* Completely changed the method HotTracker uses to attribute Extensions. It should now be more understandable and should give more accurate results when HoTs are removed early (which happens more often now that Swiftmend consumes HoTs).
* Added the `noHaste` and `noPandemicRefresh` flags to HotTracker's HotInfos, allowing me to better represent some HoTs.

HotTrackerRestoDruid
* Added extra handling to allow attributions to include the healing boost from Mastery

Tranquility
* Tranquility direct heals refresh the Tranq HoT but weren't producing refreshbuff events. Added a module to fabricate them.

Flourish
* Previously was outputting an incredible amount of non-helpful information that made both the tooltip and the *code* very hard to read. I removed the unneeded stuff and reworded what was left to make it more understandable.
* Added code to handle the special case when Convoke the Spirits casts a Flourish - healing from that is now tracked separately. 
* Screenshots of the Flourish tooltip change are below. Note that the numbers being different is expected - the per cast numbers from the table in the old version were wrong, and the updates to HotTracker extension attribution and mastery attribution changed the overall numbers.

OLD:
flourish-old-stat
flourish-old-tooltip

NEW:
flourish-new-stat
flourish-new-tooltip


src/parser/shared/modules/HotTracker.ts
```diff
@@ -303,18 +311,18 @@ abstract class HotTracker extends Analyzer {
...
       end: event.timestamp + hotDuration,
       originalEnd: event.timestamp + hotDuration,
-      spellId, // stored extra here so I don't have to convert string to number like I would if I used its key in the object.
-      name: event.ability.name, // stored for logging
-      ticks: [], // listing of ticks w/ effective heal amount and timestamp, to be used as part of the HoT extension calculations
-      attributions: [], // The effect or bonus that procced this HoT application. No attribution implies the spell was hardcast.
-      extensions: [], // The effects or bonuses that caused this HoT to have extended duration. Format: { amount, attribution }
-      boosts: [], // The effects or bonuses that caused the strength of this HoT to be boosted for its full duration.
-      healingAfterOriginalEnd: 0, //healing this hot did after the base duration due to extensions
-      maxDuration, //the true max duration that extensions will obey by (looking at rising mist)
```
kfinch, line 317: These fields already documented in the definition of the Tracker interface

analysis/druidrestoration/src/modules/features/Tranquility.ts
```diff
@@ -0,0 +1,49 @@
...
+      Events.heal.by(SELECTED_PLAYER).spell(SPELLS.TRANQUILITY_HEAL),
+      this.onTranqHeal,
+    );
+  }
+
+  /**
+   * When tranquility directly heals it refreshes the HoT, but no refresh buff event is emitted.
+   * This fabricates the event.
+   */
+  onTranqHeal(event: HealEvent) {
```
abelito75: This is probably because it uses the `Buff Stack Gained` event instead. Since you're not putting a new buff on them nor are you refreshing one.
image

kfinch: Interesting, not sure how I missed those events the first time. However, observation of buff remove times suggests the Buff Stack Gained events actually *are* refreshing the duration.

abelito75: Yeah. Buff stack events have the power to do that. You should update the hot tracker to handle buff stack events instead of throwing a new fab event imo

kfinch: Yeah you're probably right, I'll make the change

PR #4502, opened 2021-05-17
Feral Druid - Fixed display issues in energy cap tracker

kfinch:
The existing Feral Druid energy cap tracker was confusing and appears to have its numbers inverted.

Below is what it currently looks like, note that 75.72% is actually the percent time NOT capped.
image

I swapped out the stats box for the same one currently used by Rogue and WW Monk modules, as I like the look of it. Below is what the same log looks like now:
image

Note the numbers don't quite match because the calculation used is slightly changed - time based instead of amount based.

### DIFF
--- a/analysis/druidferal/src/modules/features/EnergyCapTracker.js
+++ b/analysis/druidferal/src/modules/features/EnergyCapTracker.js
@@ -1,12 +1,14 @@
 import { t } from '@lingui/macro';
-import { formatPercentage } from 'common/format';
+import { formatPercentage, formatDuration } from 'common/format';
 import SPELLS from 'common/SPELLS';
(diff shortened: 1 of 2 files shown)


analysis/druidferal/src/modules/features/EnergyCapTracker.js
```diff
@@ -28,13 +30,13 @@ const RESOURCE_REFUND_ON_MISS = 0.8;
  * Reduced drain cost from Berserk/Incarnation on Ferocious Bite is already applied in the log.
  */
 class EnergyCapTracker extends RegenResourceCapTracker {
-  get percentCapped() {
+  get percentNotCapped() {
```
kfinch, line 33: The getter is actually calculating percent not capped, hence the name change.

PR #4508, opened 2021-05-22
Made Heart of the Wild a shared ability and fixed an issue with cast effic display in Feral

kfinch:


Dambroda (approved):

kfinch:
The integration failure appears to be an additional minor suggestion in Boomkin that I actually intended, so I'll just update the expected

PR #4510, opened 2021-05-23
rebuilt ExpelHarm Normalizer from core EventOrderNormalizer

Carglass:
The expelHarmNormalizer was reordering the events around a cast of Expel Harm in a custom way. It was hard to debug, and there were some issues so I rewrote it using the EventOrderNormalizer abstract class. I also converted the file to typescript.

Now Orbs are correctly counted as being "attracted"  by Expel Harm.

I tested on some personal logs. Please let me know if I should look into integration tests for the normalizer.


analysis/monkbrewmaster/src/normalizers/ExpelHarm.ts
```diff
@@ -0,0 +1,45 @@
...
+      CelestialFortuneHealOrder,
+      ExpelHarmHealOrder,
+      ...GIFT_OF_THE_OX_SPELL_IDS.map((id) => ({
+        beforeEventId: SPELLS.EXPEL_HARM.id,
+        beforeEventType: EventType.Cast,
+        afterEventId: id,
+        afterEventType: EventType.Heal,
+        bufferMs: 50,
+        anyTarget: true,
+        updateTimestamp: true,
```
kfinch: Instead of doing this, you should pass the array of spellIds into `afterEventId` (which supports arrays). I think for performance reasons its preferable to use an array in one EventOrder rather than multiple EventOrder specs.

kfinch (changes requested):
Looks good except for the one comment. Just to double check, did you run this locally and confirm that it fixes the issue?

Also, as the autobuild shows, you need to add a changelog entry.

analysis/monkbrewmaster/src/normalizers/ExpelHarm.ts
Carglass: I merged the gift of the ox ids into one event order. I also tried to add the other spells into one event order, but somehow, it would not work, I did not dig too much :man_shrugging:

Carglass:
I created a contibutor and an entry in the changelog, the overlay indicated invalid contributor, maybe something more needs to be done?

And I tested locally, the bug was fixed, I confirmed on 2-3 parses

kfinch:
It appears you have figured it out

kfinch (approved):

PR #4513, opened 2021-05-26
Flash concentration

Carglass:
I created an Analyzer for Flash Concentration Holy Priest Legendary, here is what it adds to Holy Priest
 * - statistic: HPS added to Heal Casts.
 * - statistic: Cast Time gained on Heal thanks to FC
 * - suggestion: when too many FC Drops.
 * - suggestion: when too many wasteful Flash Heal Casts.

### DIFF
--- a/analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
+++ b/analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
@@ -99,27 +99,48 @@ class FlashConcentration extends Analyzer {
     };
   }
 
+  get wastefulFlashHealSuggestion() {
+    return {
+      actual: this._wastefulCasts,
+      isGreaterThan: {
+        minor: 0,
+        average: (2 * this.owner.fightDuration) / 1000 / 60,
+        major: (4 * this.owner.fightDuration) / 1000 / 60,
+      },
+      style: ThresholdStyle.NUMBER,
+    };
+  }
+
   suggestions(when: When) {
     when(this.dropSuggestion).addSuggestion((suggest, actual) =>
       suggest(
         <span>
-          You dropped <SpellLink id={SPELLS.FLASH_CONCENTRATION.id} /> stacks {actual} times. Try to
-          keep it up at all times.
+          You dropped your <SpellLink id={SPELLS.FLASH_CONCENTRATION.id}></SpellLink> stacks{' '}
+          {actual} times. Try to keep it up at all times.
         </span>,
       )
         .icon(SPELLS.FLASH_CONCENTRATION.icon)
-        .staticImportance(SUGGESTION_IMPORTANCE.MAJOR),
+        .staticImportance(SUGGESTION_IMPORTANCE.MAJOR)
+        .actual(<>You lost your stacks {actual} times</>)
+        .recommended(<>It is recommended to maintain 5 stacks at all times</>),
     );
 
-    // when(this._drop_suggestion).addSuggestion((suggest) =>
-    //   suggest(
-    //     <span>
-    //       You used too many <SpellLink id={SPELLS.FLASH_HEAL.id} /> Jun!
-    //     </span>,
-    //   )
-    //     .icon(SPELLS.FLASH_CONCENTRATION.icon)
-    //     .staticImportance(SUGGESTION_IMPORTANCE.MAJOR),
-    // );
+    when(this.wastefulFlashHealSuggestion).addSuggestion((suggest, actual) =>
+      suggest(
+        <span>
+          You used too many unnecessary <SpellLink id={SPELLS.FLASH_HEAL.id} /> while you had 5
+          stacks of <SpellLink id={SPELLS.FLASH_CONCENTRATION.id}></SpellLink>
+        </span>,
+      )
+        .icon(SPELLS.FLASH_CONCENTRATION.icon)
+        .actual(
+          <>
+            You used {actual} <SpellLink id={SPELLS.FLASH_HEAL.id} /> while there were more than 5
+            seconds left of <SpellLink id={SPELLS.FLASH_CONCENTRATION.id} />,
+          </>,
+        )
+        .recommended(<>No inefficient cast is recommended</>),
+    );
   }
 }
 
--- a/analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
+++ b/analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
@@ -1,10 +1,11 @@
 import SPELLS from 'common/SPELLS';
 import { SpellLink } from 'interface';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { CastEvent } from 'parser/core/Events';
+import Events, { CastEvent, HealEvent } from 'parser/core/Events';
 import { ApplyBuffStackEvent, ApplyBuffEvent, RemoveBuffEvent } from 'parser/core/Events';
 import SUGGESTION_IMPORTANCE from 'parser/core/ISSUE_IMPORTANCE';
 import { When, ThresholdStyle } from 'parser/core/ParseResults';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
 import ItemHealingDone from 'parser/ui/ItemHealingDone';
 import Statistic from 'parser/ui/Statistic';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
@@ -31,6 +32,13 @@ class FlashConcentration extends Analyzer {
   _wastefulCasts = 0;
   // a counter for when FC stacks are dropped
   _totalDrops = 0;
+  // tracker for number of current stacks
+  _currentStacks = 5;
+  // cumulated gained healing and cast time on every buffed Heal Cast, plus the totals
+  _totalHeal = 0;
+  _gainedHeal = 0;
+  _gainedCastTime = 0;
+  _totalCastTime = 0;
 
   constructor(options: Options) {
     super(options);
@@ -51,6 +59,14 @@ class FlashConcentration extends Analyzer {
       Events.cast.by(SELECTED_PLAYER).spell(SPELLS.FLASH_HEAL),
       this.onByPlayerFlashHealCast,
     );
+    this.addEventListener(
+      Events.heal.by(SELECTED_PLAYER).spell(SPELLS.GREATER_HEAL),
+      this.onByPlayerHealAmount,
+    );
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.GREATER_HEAL),
+      this.onByPlayerHealCast,
+    );
   }
 
   /**
@@ -65,6 +81,7 @@ class FlashConcentration extends Analyzer {
     } else {
       this._isFullStack = false;
     }
+    this._currentStacks = event.stack;
   }
 
   /**
@@ -79,6 +96,7 @@ class FlashConcentration extends Analyzer {
   onByPlayerFlashConcentrationInitialBuff(event: ApplyBuffEvent) {
     if (event.prepull !== true) {
       this._isFullStack = false;
+      this._currentStacks = 1;
     }
   }
 
@@ -88,6 +106,7 @@ class FlashConcentration extends Analyzer {
   onByPlayerFlashConcentrationBuffRemoval(event: RemoveBuffEvent) {
     this._totalDrops += 1;
     this._isFullStack = false;
+    this._currentStacks = 0;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+      }
+      this._lastFullStackRefresh = event.timestamp;
+    }
+  }
+
+  onByPlayerHealAmount(event: HealEvent) {
+    this._totalHeal += event.amount;
+    this._gainedHeal +=
+      (event.amount * (this._currentStacks * 0.03)) / (1 + this._currentStacks * 0.03);
+    this._gainedCastTime += this._currentStacks * 200;
```
kfinch: All spell or analysis related constant numbers should be at the top of the file above the class instead of hardcoded. This makes it easier to see what the numbers refer to and also easier to update if a spell gets buffed or nerfed. In this case the `0.03` and `200` should be moved into consts.

See Flourish.tsx for an example of this.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+    );
+  }
+
+  get dropSuggestion() {
+    return {
+      actual: this._totalDrops,
+      isGreaterThan: {
+        minor: 0,
+        average: 0,
+        major: 0,
```
kfinch, line 259: Is letting the buff drop so bad that it happening even once warrants a major suggestion? It might be worth spreading this out a bit (unless it really is that bad).

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+          You used too many unnecessary <SpellLink id={SPELLS.FLASH_HEAL.id} /> while you had 5
+          stacks of <SpellLink id={SPELLS.FLASH_CONCENTRATION.id}></SpellLink>
+        </span>,
+      )
+        .icon(SPELLS.FLASH_CONCENTRATION.icon)
+        .actual(
+          <>
+            You cast {actual} times while there were more than 5 seconds left of{' '}
+            <SpellLink id={SPELLS.FLASH_CONCENTRATION.id} />,
+          </>,
```
kfinch: So it's always bad to cast Flash Heal at max stacks? Is this because at max stacks Heal is a strictly better spell? If so, you should explicitly say that in the suggestion.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+    if (this._isFullStack) {
+      if (event.timestamp - this._lastFullStackRefresh < 15000) {
+        this._wastefulCasts += 1;
+      }
+      this._lastFullStackRefresh = event.timestamp;
+    }
+  }
+
+  onByPlayerHealAmount(event: HealEvent) {
+    this._totalHeal += event.amount;
```
kfinch: Absorbed healing shows up in the optional `absorbed` field and is typically effective healing, so when tallying we use `event.amount + (event.absorbed || 0)`

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+    this._gainedHeal +=
+      (event.amount * (this._currentStacks * 0.03)) / (1 + this._currentStacks * 0.03);
```
kfinch: You probably want to use the `calculateEffectiveHealing` helper method here, as it also accounts for overhealing properly.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+        ? event.channel.beginChannel.timestamp
+        : event.timestamp);
+  }
+
+  statistic() {
+    // HealHPSNoFC gives the HPS which would have been obtained by casting the same number of heals without Flash Concentration Buff
+    const HealHPSNoFC =
+      (this._totalHeal - this._gainedHeal) / (this._totalCastTime + this._gainedCastTime);
+    // BuffedHealHPS gives the total HPS of Heal in this fight
+    const BuffedHealHPS = this._totalHeal / this._totalCastTime;
```
kfinch: I think this calculation is really sketchy and you shouldn't be trying to mash together the heal gain and cast times like this. Healers are mana limited and so we can't directly attribute cast time like that.

I'm up for more discussion on this, but I think you split the healing gained from the buff and the cast time gains separately. Alternately, you could pretend a fully buffed Heal is like a stronger and cheaper Flash Heal and tally it that way.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,228 @@
...
+      >
+        <BoringSpellValueText spell={SPELLS.FLASH_CONCENTRATION}>
+          <ItemHealingDone
+            // approximate considering the player would have cast heal without FC
+            amount={(BuffedHealHPS - HealHPSNoFC) * this._totalCastTime}
+            approximate
+          />
+          <br />
+          {this._gainedCastTime / 1000} sec saved on <SpellLink id={SPELLS.GREATER_HEAL.id} /> casts
+        </BoringSpellValueText>
```
kfinch: What does this look like? Can you attach a screenshot?

kfinch (changes requested):

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: well, I am more on the coder side than pure theorycrafter, but I play holy priest, and loosing your stacks is not only a pure loss of HPS, but also a mess to rebuild asap while also healing efficiently, and a huge mana loss. So I guess it is pretty bad overall ^^ I was hoping more expert holy players could give their point of view during the review still.

Carglass: image
from [mechanicalpriest.com](https://mechanicalpriest.com/holy-priest-healing-with-flash-concentration/)

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: So, no, but this is a good enough approximation would be my 2-cents (with a little tweak, see below). There are two cases when it is ok to cast Flash Heal before the buff is about to drop:
- When you have 2 stacks of surge of light (which can def happen, truthfully I forgot about this case, but would fit if I made the minor threshold at 1 per minute, I think, alternatively, we could check for it, although, it is more complex)
- When someone will die in between the time Flash Heal would have healed and the time Heal heals, which is a very very narrow window at 5 stacks, so I chose to ignore that, because it is pretty complex and not very impactful imo.
That's another case where a more experienced point of view would help for sure

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: I'll look into this helper :+1:

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
kfinch: >     * When you have 2 stacks of surge of light (which can def happen, truthfully I forgot about this case, but would fit if I made the minor threshold at 1 per minute, I think, alternatively, we could check for it, although, it is more complex)
It's very easy to check, try `this.selectedCombatant.getBuffStacks(SPELLS.SURGE_OF_LIGHT_BUFF.id) >= 2`

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: I am expecting discussion on this one for sure. I put it into the tooltip, but Flash Concentration fundamentally changes the Gameplay of Holy priest, so hard to compare. I thought about your proposal to compare to Flash heal casts, but it sounded weirder in my mind, because that's another spell, and Heal replaces more then only Flash Heals casts under Flash Concentration (see below). 
My first iteration was to have the gain in hps from using the same amount of heal casts, unbuffed, without taking into account the cast time gain. And to be honest it was irrelevant. like 150HPS, which is misleading. 
so what I did hear is compare what the HPS produced by the same amount of Heal Casts would have been without FC, to what we see with FC. And then I give what it means gained HPS over the total time spent casting Heal. 
It is absolutely not perfect, but that is the most relevant calculation I could find. the reality is that using Flash Concentration makes Heal replace:
- Flash Heal
- Heal 
- Prayer of Healing
- Renew ...
So the impact is very hard to value :man_shrugging: 
I am open to suggestion for sure. More theorycrafting experience would be very valuable here as well

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: Here goes :)
image

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
kfinch: Generally I think if you can't make a reasonable statistic, you shouldn't. It's possible there's just no satisfying thing to display for this legendary, and that's OK. You could just add in the suggestions and leave it at that, or add a less ambitious statistic instead.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
kfinch: Having `40.6 sec saved on Heal casts` split over two lines doesn't look great. You could fit it on one line using `<small>` tags to produce the muted text like you see with `11.82% of total`

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: oh yes indeed, thanks! will implement it for sure then and I think this issue here is mostly :+1:

Carglass:
I should have fixed everything but https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4513#discussion_r640172893
to be continued :)

Carglass:
image
image

kfinch (approved):

kfinch:
Changes look good. Let me know when you've updated the display to your liking and I'll look again.

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  onByPlayerHealAmount(event: HealEvent) {
+    this._totalHeal += event.amount + (event.absorbed || 0);
+    this._gainedHeal += calculateEffectiveHealingStacked(
+      event,
+      FLASH_CONCENTRATION_HEAL_BUFF,
+      this._currentStacks,
+    );
```
kfinch: Wait a sec just realized this is wrong. `calculateEffectiveHealingStacked` is the attribution for the last stack, while I think you want attributions for all the stacks. `calculateEffectiveHealing(event, FLASH_CONCENTRATION_HEAL_BUFF * this._currentStacks)` is what you want here

kfinch (changes requested):

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
Carglass: let me check again when I get a chance :) I was not sure myself between both tbh :sweat_smile:

Carglass:
From my discussions on the Priest Discord, here is what I have left to do (not including bugs or your comments that may come):
- Drop the HPS in the stat box, mostly irrelevant
- Document that I do NOT create a wasteful stack when there are 2 SOL procs, people got confused by that.
- Raise the Threshold for bad FH casts, to account for the few use cases where it make sense (ignore 1 or 2 per min)
- Have number of Heals cast per number of stacks in Statistic
- Compare number of Flash Heal to Heal under buff in Statistic

analysis/priestholy/src/modules/shadowlands/items/FlashConcentration.tsx
kfinch: Those methods were in dire need of a docs update so I just put up a PR for it. If you've got time could you take your new-person eyes and look at this PR? https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4516  - let me know if you can understand how they work from the new docs.

kfinch:
Looks like you're missing a buffer div or something on the pie chart - there's no margin on it. I haven't used these charts before but check how others are doing it.
image

kfinch (approved):

PR #4516, opened 2021-05-27
Rewrote documentation for all 'calculateEffective...' functions and removed references to Legion spells and abilities. Functionality is unchanged.

kfinch:
Also switched them to TypeScript.


kfinch:
Also fixed a bug in ShiveringCore where the damage amount was being passed to calculateEffectiveDamage instead of the DamageEvent

PR #4524, opened 2021-06-03
Refactored formatDuration to take milliseconds LIKE EVERYONE ELSE

kfinch:
I couldn't take it anymore. All of the other format functions take milliseconds but formatDuration takes seconds, and 95% of the people calling it were passing milliseconds and dividing by 1000 in the call. This is the refactor to fix it.


MartijnHols:
Nice fix, thanks!

PR #4528, opened 2021-06-08
Add support for TBC analysis

MartijnHols:
This is based around a concept where we can have any number of configs in the app and the config determines if it's applicable to a player/fight/expansion. Analysis folders / configs will need to become standalone packages. I'll call each instance of this a "config" further down, though I am still looking for a better name since config is generally just the object with settings while it's being used here to include anything in the analysis folder.

In the future I'd like to:
- remove `defaultModules` from `core/CombatLogParser`, instead in a config we can do `import DefaultModules from '@wowanalyzer/shadowlands'` and spread that in the modules object. This gives more freedom (e.g. you can stop spreading or `delete` modules you don't want) and makes configs more standalone while reducing the spaghetti in the core
- SPELLS become config specific, no more global spells object
- remove SPECS file, it's a reference from core to configs which shouldn't exist (core should not have any config-specific code)
- stat tracker buffs list should become expansion specific (and disappear from core/shared)
- Allow for multiple configs per spec
- lots more

But all of that is a ton of work and refactoring which is risky and I'll only try to do when necessary and/or low risk.

More pragmatic I'll probably do:

- set up basic Hunter analysis with a working timeline and perhaps statistics to indicate auto shot clipping and such that are essential to the spec
- fix cross-expansion issues such as tooltips not working, SPELLs not existing
- clean up the checklist. As mentioned many times in Discord I want to make it possible to have different levels of spec-support. Checklist being the highest, but earlier steps where there are just some statistics should be encouraged too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
         id: cache
         with:
           path: node_modules
-          key: ${{ runner.os }}-node_modules3-${{ hashFiles('yarn.lock') }}
-          restore-keys: ${{ runner.os }}-node_modules3-
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('package.json', 'yarn.lock') }}
+          restore-keys: ${{ runner.os }}-node_modules-
       - name: Get yarn cache directory path
         if: steps.cache.outputs.cache-hit != 'true'
         id: yarn-cache-dir-path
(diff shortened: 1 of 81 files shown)


kfinch (commented):
It looks like a 'healthstone' module was added in all the updated integration tests - was this intended?

MartijnHols:
An issue was fixed where the Healthstone module did not properly detect the presence of a Warlock in the group.

kfinch (approved):
`analysis/tbchunter/...` isn't what I expected, shouldn't there be `analysis/the-burning-crusade/hunter` instead? I guess this is all temp so it doesn't matter much.

MartijnHols:
`analysis/*` are basically all configs available with their appropriate name. I prefer prefixing over nesting, so tbchunter is intended.

PR #4531, opened 2021-06-10
Add elemental blast spell id, update lightning shield spell id, update Elemental shaman patch version

kedearian:
I added the elemental blast spell id (missing previously), update the lightning shield spell id, and bumped the version for elemental shaman to 9.0.5. I left it as partial support, my local testing show it to be working correctly.

Also updated the changelogs since I missed that step in my previous pull request.


kedearian:
Spell ID existed under talents, not sure why it was throwing errors that it didn't exist. I'll re-evaluate and submit if needed. Local testing didn't show an issue with this change which is weird.

PR #4532, opened 2021-06-11
Updated Events Tab with spellIds and additional event type coloring

kfinch:
image


kfinch:
Drain event appearance with new color and amount coloring:
image

kfinch:
Also added amount info to stacking events
image

Dambroda (approved):

PR #4534, opened 2021-06-12
Fixed event order normalizer unit test and test run scripts

kfinch:
Also fixed an issue where the testing script definitions were causing parser tests not to run


MartijnHols (approved):
Thanks!

PR #4543, opened 2021-06-22
modified checlist for enduring blow gameplay

Carglass:
Using the Enduring Blow Legendary promises to be mainstream in 9.1 when NF becomes the better covenant for Arms Warriors. The main gameplay impact of using it is that Mortal Strike should be used on cooldown at any time, even during the execute phase.

I modified the checklist to add Mortal Strike to the normal rotational abilities list, with a not explaining the change with this legendary. I also removed the Mortal Strike Efficiency box when using this legendary, since those rules do not apply anymore.


MartijnHols (approved):
Thanks!

PR #4547, opened 2021-06-24
[Warlock][Destruction] Convert to Typescript

ogunb:
Hi! 
I'm a Destro player and wanted to contribute. This seemed like a good issue to start.

I need feedback for `FabricatedEvent`.I wanted to use `Event` type initially but `timestamp` property was not nullable so default assignment failed. Since there may be some .js files that doesn't input it, I didn't want to remove it.
And there was a todo on event type seeming off, I felt like this was a good solution. Let me know if there is a better way.


analysis/warlockdestruction/src/modules/soulshards/SoulShardTracker.ts
```diff
@@ -120,7 +134,7 @@ class SoulShardTracker extends ResourceTracker {
     super.onCast(event);
   }
 
-  onEvent(event) {
+  onEvent(event: Event<string>) {
```
kfinch: I think you want `AnyEvent` here

src/parser/core/modules/EventEmitter.ts
```diff
@@ -16,6 +16,11 @@ type BoundListener<ET extends EventType, E extends MappedEvent<ET>> = {
   listener: EventListener<ET, E>;
 };
 
+interface FabricatedEvent extends Omit<Event<any>, 'timestamp'> {
+  timestamp?: number | undefined;
+  [prop: string]: any;
+}
```
kfinch: I'm unclear why you're doing this - which FabricatedEvent doesn't have a timestamp? I see you mentioned problems with this in the PR description, could you go in to more detail about what's going wrong?

kfinch (changes requested):
I scanned over what you've done here and it overall looks good. Just see comments below.

src/parser/core/modules/EventEmitter.ts
ogunb: This is the error I get before changing the type:
Screen Shot 2021-06-25 at 17 54 18

Same thing happens for current:
Screen Shot 2021-06-25 at 17 56 47

This seems ok though:
Screen Shot 2021-06-25 at 18 00 27

I've gone over some other classes and they seemed to implemented it with variables like so:

```ts
    const evt: AtonementAnalyzerEvent = {
      timestamp: spiritShellEvent.timestamp,
      targetID: spiritShellEvent.targetID,
      healEvent: spiritShellEvent,
      damageEvent: this.atonementSource,
      sourceID: spiritShellEvent.sourceEvent.sourceID,
      provenance: SourceProvenance.SpiritShell,
      type: AtonementAnalyzer.atonementEventFilter.eventType,
      buffDetails,
      expirationDelta: buffDetails.baseExpirationTimestamp - spiritShellEvent.timestamp,
    };

    this.eventEmitter.fabricateEvent(evt);
```

I'm not that fluent with TS, could this be the problem here?:
https://stackoverflow.com/questions/61698807/interesting-behaviour-object-literal-may-only-specify-known-properties

---

If I remove Omit and just extend Event, as far as TS concerns, timestamp should never be undefined therefor doesn't require a default assignment.
Screen Shot 2021-06-25 at 18 07 43

These are the problems I run into. So 2 more possible solutions I see here:
- Just implement it as an event variable and revert changes in EventEmitter.ts.
- Remove default assignment as it should never be undefined.

kfinch: I still haven't had time to look really closely at what's happening, but it sounds like wires got crossed somewhere. I'd like you to revert the changes to EventEmitter - an Event, even a fabricated one, should always have a timestamp. If you're sure the timestamp you're using will never be undefined, I'd prefer you make whatever assertions are needed when the Event is created instead of altering Event's signature.

ogunb: Okay, I'll revert the EventEmitter changes and go with an event variable so that TS doesn't complain.

ogunb: Done. I reverted the emitter changes.

kfinch: Looks good. Odd that assigning it to a separate var was all it took to calm the typechecker...

kfinch (approved):
Thanks for the contribution!

Issue #3801, opened 2020-10-04
Convert StatisticBox to Statistic


MartijnHols:
`StatisticBox` is deprecated but there are a lot of older files still using it. Please help us convert these to using `Statistic` instead. `Statistic` provides more flexibility.

Because there are a lot of older files in the project still using `StatisticBox`, this is a lot of work that can not be done in one go. This issue serves to point anyone interested in the right direction.

Example PR: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/3786/files#diff-50f1fd019ed3b46be191d743d38f701cR204

PR #4550, opened 2021-06-26
Removed StatisticBox reference and replaced with STATISTIC_ORDER

Soulhealer95:
Following: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/3801
> StatisticBox is deprecated but there are a lot of older files still using it. Please help us convert these to using Statistic instead. Statistic provides more flexibility.

I found a reference to the now deprecated StatisticBox in a magearcane file. It wasn't instantiating the class, only the STATISTIC_ORDER component, so I didn't have to replace any instances of StatisticBox directly but rather just change the import for the STATISTIC_ORDER to point to the file of that name instead of via StatisticBox. 

Website seemed to compile and load without any noticeable issues. 

`yarn test` couldn't find any tests to run 
$ yarn test
yarn run v1.22.10
$ react-app-rewired test
No tests found related to files changed since last commit.


Soulhealer95:
added myself to contributors and updated change log.

kfinch (approved):
Thx

Issue #3728, opened 2020-09-21
Update Retribution Paladin for Shadowlands


Juko8:
Changes Retribution needs for shadowlands, this will be updated as blizzard releases new things or implements them.

### Core
- [x] Update Spellbook
- [x] Update cast efficiency
- [x] Update Hammer of Wrath to use new execute module
- [ ] Update Execution Sentence module
### Covenants
- [ ] Kyrian
- [ ] Venthyr
- [ ] Necrolord
- [ ] Night Fae
### Legendaries
- [x] Final Verdict
- [ ] The Mad Paragon
- [ ] Relentless Inquisitor
- [ ] Tempest of the Lightbringer
### Conduits
- [ ] Expurgation
- [ ] Templar's Vindication
- [ ] Truth's Wake
- [ ] Virtuous Command

abelito75:
@Juko8 I think i made a HoW module for you in shared

PR #4552, opened 2021-06-27
Updates to execution sentence for Shadowlands.

SethEArnold:
Removed indirect damage calculations as they are no longer necessary and updated tool tip to match.

Seems to have also fixed a bug with not counting Final Verdict's consuming of judgement debuffs. Not sure why, something to do with removing the event listener.


SethEArnold:
For task in #3728

MartijnHols (approved):
Nice, thanks!

PR #4554, opened 2021-06-30
Balance Druid - More Checklist Updates

kfinch:
* Fixed outdate handling of Eclipse spell haste
* Merged DoT rules in checklist
* Added new Eclipse rule that checks filler casts and starsurge usage, and added/updated related suggestions
* Separated AsP overcap with a rule for during Eclipse (when it's extra bad to overcap)
image


MartijnHols (approved):
Thanks!

PR #4556, opened 2021-06-30
Add functional low rank casts suggestion

MartijnHols:
Added functional low rank spells suggestions with a shared functional cast count stat


MartijnHols:
Marked as draft: need to figure out how to filter by player.

PR #4558, opened 2021-07-02
Guardian convoke support

abbottmg:
The first cut of this is a copy from the Feral module with Feral Frenzy ripped out. I don't know of any damage abilities that weren't already covered by that module but I'd definitely appreciate a double-check.

I'm also interested in parsing healing done and damage prevention applied via ironfur stacks if feasible, but that could be a separate PR.


MartijnHols (approved):
LGTM Thanks!

PR #4559, opened 2021-07-03
Add functional buff handling

MartijnHols:
# Includes #4556

- debuffs are handled the same as buffs
- `buffApplications` for tracking when buffs were applied. Should probably only be used by other shared stats
- `buffUptimeTotal` for total buff uptime (including overlap)
- `buffUptimeDistinct` for distinct buff uptime
- renamed stats to metrics to differentiate it more and fits better as well


src/parser/core/ParseResults.tsx
```diff
@@ -339,13 +340,13 @@ export type When = <T extends ValidThresholds>(threshold: T) => GenericSuggestio
 class ParseResults {
   tabs: ParseResultsTab[] = [];
   statistics: React.ReactElement[] = [];
-  issues: Issue[] = [];
+  issues: Array<Issue | SuggestionData> = [];
```
kfinch, line 343: What's with the move from `[]` to `Array<>` - I'm not too familiar with the pro/con

src/parser/core/metric.ts
```diff
@@ -0,0 +1,28 @@
+import { AnyEvent } from './Events';
+import Ability from './modules/Ability';
+
+export interface Info {
+  playerId: number;
+  abilities: Ability[];
+  fightStart: number;
+  fightEnd: number;
+}
```
kfinch, line 9: This seems good to start and we should also be able to easily add things to this later - I can see us wanting encounter information as well

src/parser/core/metric.ts
```diff
@@ -0,0 +1,28 @@
...
+export type Metric<Value = any> = (events: AnyEvent[], info: Info, ...args: any[]) => Value;
+
+const metric = <T extends any[], U>(fn: (events: AnyEvent[], ...args: T) => U) => {
```
kfinch, line 26: Seems like just what we need - for the sake of future devs could add docs explaining what this is and when it should be used?

src/parser/shared/metrics/buffApplications.ts
```diff
@@ -0,0 +1,82 @@
...
+import metric from 'parser/core/metric';
+
+/**
+ * Returns the index of the last element in the array where predicate is true, and -1
+ * otherwise.
+ * Source: https://stackoverflow.com/a/53187807/684353
+ * @param array The source array to search in
+ * @param predicate find calls predicate once for each element of the array, in descending
+ * order, until it finds one where predicate returns true. If such an element is found,
+ * findLast immediately returns that element index. Otherwise, findLastIndex returns undefined.
```
kfinch: Out of date docs - this returns the object or undefined, not the index

src/parser/shared/metrics/buffApplications.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Returns an object with all buff applications. Should probably not be used
+ * directly in configs, but only internally by other shared stats.
+ */
+const buffApplications = (events: AnyEvent[]) =>
+  events.reduce<{
+    [spellId: number]: {
+      [sourceId: number]: Array<{
```
kfinch, line 34: What's the motivation behind indexing by spellId and sourceId but not by target? It seems like it would be much easier to detect buff times per target if we indexed by target. Also, won't almost everything have sourceId === the player?

src/parser/shared/metrics/buffApplications.ts
```diff
@@ -0,0 +1,82 @@
...
+        targetId: number;
+        targetInstance?: number;
+        start?: number;
+        end?: number;
```
kfinch, line 38: I could also see myself adding refreshTimes to this as well, but that can be a future upgrade

src/parser/shared/metrics/buffApplications.ts
```diff
@@ -0,0 +1,82 @@
...
+      );
+
+      if (item) {
+        item.end = event.timestamp;
+      } else {
+        sourceSpellBuffs.push({
+          targetId: event.targetID,
+          targetInstance: event.targetInstance,
+          start: undefined,
+          end: event.timestamp,
```
kfinch, line 75: Is this the start of an alternate way of handling pre-cast buffs? "Missing start time means the buff was present on pull"

src/parser/shared/metrics/buffUptimeDistinct.ts
```diff
@@ -0,0 +1,33 @@
...
+    const end = buff.end ?? fightEnd;
+    lastEnd = end;
+
+    const duration = Math.max(0, end - distinctStart);
+
+    return uptime + duration;
+  }, 0);
+};
+
+export default metric(buffUptimeDistinct);
```
kfinch, line 33: Won't this not work right with `metric`? The memoization code only checks that the events match in order to return the cached value, but `buffUptimeDistinct` could be called with different spellIds.

src/parser/core/metric.ts
MartijnHols: I'm unsure about all the fight props so far, seems like `Info` is going to keep on growing like this and that's not great. Might switch this around a bit later. I considered making our own `Fight` type that's basically a cleaned up version of WCL's response.

src/parser/core/ParseResults.tsx
MartijnHols: Simple types use `[]` and complex types `Array<x>`. Eslint will yell about it.

src/parser/shared/metrics/buffApplications.ts
MartijnHols: target also has a targetinstance, which can vary but targetId can be repeated. If the player is affected by someone else's buff, like Bloodlust, the source will be someone else. There are quite a few buffs like that.

src/parser/shared/metrics/buffApplications.ts
MartijnHols: This has always been the case afaik. I believe we normalize these applybuffs in nowadays, but that wouldn't have been necessary with this solution. Now it's fixed twice, which ain't bad.

src/parser/shared/metrics/buffUptimeDistinct.ts
MartijnHols: Indeed, that's a WIP.

src/parser/core/metric.ts
kfinch: including sub-types like `Fight` might be the way, yeah. I don't think `Info` needs to grow a lot, but there is metadata that needs to be communicated somehow, and `Info` is probably the vehicle for that.

src/parser/shared/metrics/buffApplications.ts
kfinch: I know about targetInstance, in existing code I try to abstract that out with `encodeTargetString`, but some code forgets to use it. I fixed a bug with that just last week. Would you consider using something like `encodeTargetString` to index this object?

src/parser/shared/metrics/buffUptimeDistinct.ts
MartijnHols: Good catch though!

src/parser/shared/metrics/buffApplications.ts
MartijnHols: Might also be a separate metric.

src/parser/core/metric.ts
MartijnHols: I am hesitant to provide all data that we have like report and fight in info. Some things are duplicated or slightly different. Current approach is to add things as they're needed and then refactor once it becomes clear what is needed as a whole. So far I've been adding more and more fight props (in the other PRs), so that will probably need to be added one way or another.

src/parser/shared/metrics/buffApplications.ts
MartijnHols: It can be just as easy to detect buff times with this array if we provide a selector function. What remains is a performance question, which might be too early.

I was hesitant to use `encodeTargetString` because it would add complexity while this solution works as well. I think it makes sense to go source -> spell -> target since that reflects how they're cast (player casts spell on target). We're also pretty likely to be looking at buffs cast *by* the player, though perhaps then we'd look at the cast event instead. Maybe it should be changed around after all.

src/parser/core/metric.ts
MartijnHols: Will work on that later.

src/parser/shared/metrics/buffApplications.ts
MartijnHols: Will work on that later.

PR #4561, opened 2021-07-05
Fixed a bug where erroneous costs could be assigned to pre-cast spells

kfinch:
Caused by classResource field being copied with cost included into pre-cast spells - see bug in action here with a potion being listed as costing Astral Power: https://wowanalyzer.com/report/vra49mt8BjdPwDGZ/23-Mythic+Lady+Inerva+Darkvein+-+Kill+(5:33)/Coolbeanss/standard/astral-power-usage

### DIFF
--- a/src/parser/shared/normalizers/PrePullCooldowns.js
+++ b/src/parser/shared/normalizers/PrePullCooldowns.js
@@ -140,14 +140,15 @@ class PrePullCooldowns extends EventsNormalizer {
          * information more accurately.
          */
         if (precastClassResources === null && event.classResources) {
-          debug && console.debug('Setting prepull class resources (but stripping costs) to:', event.classResources);
-          precastClassResources = event.classResources;
-          // make a new class resources with costs stripped - have to deep copy to avoid messing with the original cast event
-          precastClassResources = precastClassResources.map(classResource => {
-            const noCostClassResource = Object.assign({}, classResource);
-            noCostClassResource.cost = undefined;
-            return noCostClassResource;
-          });
+          debug &&
+            console.debug(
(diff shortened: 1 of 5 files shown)


MartijnHols:
Hey are the precommit hooks not running automatically for you? Is this something I need to investigate?

src/parser/shared/normalizers/PrePullCooldowns.js
```diff
@@ -140,8 +140,14 @@ class PrePullCooldowns extends EventsNormalizer {
          * information more accurately.
          */
         if (precastClassResources === null && event.classResources) {
-          debug && console.debug('Setting prepull class resources to:', event.classResources);
+          debug && console.debug('Setting prepull class resources (but stripping costs) to:', event.classResources);
           precastClassResources = event.classResources;
+          // make a new class resources with costs stripped - have to deep copy to avoid messing with the original cast event
+          precastClassResources = precastClassResources.map(classResource => {
+            const noCostClassResource = Object.assign({}, classResource);
```
MartijnHols: In the modern JS version we use, you can write this like so:

```
const noCostClassResource ={
  ...classResource,
  cost: undefined,
};
```

With that this entire section could become:

```
precastClassResources = precastClassResources.map((classResource) =>({
  ...classResource,
  cost: undefined,
}))
```

kfinch: Right I forgot you could do that, will change it

kfinch:
Looks like the integration test diff is because this bug was manifesting in a very minor way in the MM hunter log, so I'll just update the test

PR #4563, opened 2021-07-05
Fixed bug Where final verdict wasn't being counted as consuming judgement debuffs

SethEArnold:
Fixed bug Where final verdict wasn't being counted as consuming judgement debuffs. This occured because final verdict was pushed to the list of holy power consumers in the final verdict module. Then if the judgement module ran before the final verdict module it would run using a list that didn't include final verdict because the list hadn't been updated yet.

Fixed by removing the push from the final verdict module and adding final verdict to the list in the constants.tsx file.

Example/steps to reproduce:
https://www.wowanalyzer.com/report/AwkDTXcGrKLM3jhQ
Go to wipe 16
Go to Fashathus - the ret paladin
Go to statistics. It should say 22 judgement debuffs consumed
Refresh the page
It will now say 3 judgement debuffs consumed
This happens because for some reason when navigating straight to the page it always bugs but when going from the wipes page it usually doesn't bug


SethEArnold:
The bug is inconsistent to reproduce. I was able to consistently get it to show the right number with the steps I listed but now it always displays 3, aka the wrong amount.

MartijnHols (approved):
Thanks!

PR #4567, opened 2021-07-10
Updated timePeriods helper with more functionality, and added new functional activeTime module

kfinch:
This is incomplete, posting as a draft for review.

My approach is based on `mergeTimePeriod` code that I wrote for a previous module which effectively takes a bunch of possibly overlapping time periods and spits out the time periods that form the logical union of the inputs. When calculating active time, rather than trying to interleave the handling of GCD and channeling, I simply calculate their time periods separately and then union them at the end.

So far, I've used testing code to compare the new and current active time calculations - the new active time is consistently a bit less than the current one and I'm not yet sure why.


src/parser/shared/modules/AlwaysBeCasting.tsx
```diff
@@ -48,6 +49,18 @@ class AlwaysBeCasting extends Analyzer {
...
+
+  onFightEnd() {
+    const info = {
+      abilities: this.owner.getModule(Abilities).abilities,
+      playerId: this.owner.selectedCombatant.id,
+      fightStart: this.owner.fight.start_time,
+      fightEnd: this.owner.fight.end_time,
+    };
+    const newActiveTime = activeTime(this.owner.eventHistory, info)
+    console.log(`Fight Total Time:${this.owner.fightDuration} / Active Time Old:${this.activeTime} / Active Time New:${newActiveTime}`);
```
kfinch: How do I call a functional module from a non-functional one? Seems helpful for the transition period, but this doesn't look good.

src/parser/core/metric.ts
```diff
@@ -1,11 +1,19 @@
...
+  /** The ID of the selected player */
   playerId: number;
+  /** The selected player's spellbook */
   abilities: Ability[];
+  /** The timestamp of the selected fight's start */
   fightStart: number;
+  /** The timestamp of the selected fight's end */
   fightEnd: number;
+  /** The combatantinfo for the selected player */
+  selectedCombatant?: CombatantInfoEvent;
```
MartijnHols, line 14: This should always be set, right?

src/parser/core/metric.ts
```diff
@@ -1,11 +1,19 @@
...
+  /** The combatantinfos for all players in the encounter, indexed by playerId */
+  combatants?: { [playerId: number]: CombatantInfoEvent };
```
MartijnHols, line 16: This should always be set, right?

src/parser/core/timePeriods.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * For example, a buff's duration might be an open time period with respect to the encounter,
+ * where the buff might not have a start time if it was applied before the encounter
+ * and might not have an end time if it was still active when the encounter ended.
+ * In this case, the fight's duration is the enclosing time period.
+ */
+export type OpenTimePeriod = Partial<ClosedTimePeriod>;
+
+/** Returns the total duration of the given time period(s).
+ * Overlap will NOT be considered, use union first if non-overlapping duration is required. */
```
MartijnHols, line 28: Isn't overlap actually considered though? If I understand this correctly any overlap will both increase the total.

src/parser/shared/metrics/activeTime.ts
```diff
@@ -0,0 +1,70 @@
+import { AnyEvent, EventType } from 'parser/core/Events';
+import metric, { Info } from 'parser/core/metric';
+import { ClosedTimePeriod, duration, intersection, union } from 'parser/core/timePeriods';
+
+// TODO possible extensions:
+//   * channel and GCD time periods marked with spell, allowing additional filtering
+
+/** Time periods the player was channeling during the encounter */
+export const channelTimePeriods = metric((events: AnyEvent[], info: Info): ClosedTimePeriod[] => {
```
MartijnHols, line 9: I think it would be nice to put this in a separate file. It can probably be used for multiple purposes.

src/parser/shared/metrics/activeTime.ts
```diff
@@ -0,0 +1,70 @@
...
+  const results: ClosedTimePeriod[] = events.reduce((acc: ClosedTimePeriod[], event) => {
+    if (event.type === EventType.BeginChannel && !event.isCancelled) {
+      uncoveredUncancelledBeginChannelTime = event.timestamp;
+    } else if (event.type === EventType.EndChannel) {
+      acc.push({ start: event.beginChannel.timestamp, end: event.timestamp });
+      uncoveredUncancelledBeginChannelTime = undefined;
+    }
+    return acc;
+  }, []);
+  // handle special case of channel in progress when encounter ends
```
MartijnHols, line 27: Ooh this is a tough one. For active time I imagine you'd want to calculate this as downtime, seeing as it's practically a cancelled cast that didn't land. It might help to mark this special case with an `end: null`?

src/parser/core/metric.ts
kfinch: Yeah, but when I made it mandatory it broke your unit tests and my new unit tests because I had no good way of getting the combatant. What I probably actually need to do is make it mandatory and then mock the field out during testing.

src/parser/core/metric.ts
kfinch: Yes, see above comment (obviously this one's easy to mock because I can just do `[]`)

src/parser/core/timePeriods.ts
kfinch: Err yeah this is a weird double negative - I just need to reword this. You're correct, overlap will both increase the total.

src/parser/shared/metrics/activeTime.ts
kfinch: Ah yeah I wasn't sure how much you wanted to break this up by file. I had been putting logically connected functions together even though here you can see each outward facing function is exported. Is there any mechanical difference to separating them or is it just style?

src/parser/shared/metrics/activeTime.ts
kfinch: I don't feel I need a special time period because it's not this metric's job to judge that, it just wants to know when the player was 'active' at channeling - so capping at the encounter end makes sense to me and also makes the downstream processing a lot more straightforward.

I did consider counting this channel as 'cancelled', all I'd have to do is just remove this special case handling. It didn't feel 'fair' to the player, though.

src/parser/core/metric.ts
MartijnHols: What unit tests did it break? I may have to do more Pick<>ing in my metrics.

kfinch: I guess the tests didn't break, they no longer type-check because the `Info` object created for testing doesn't have a selectedCombatant, and I couldn't figure out a quick way to mock one in. The solution is clearly to actually mock one in.

MartijnHols: I think it may be better to have these metrics only Pick the fields they use. That way we can actually see what tests need which fields and know what to test.

MartijnHols: Ah it's in this PR: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4562/files#diff-b4f839e317b6747bff68bd8b7fda426890d7b0c51c90e933259b74d3c985a89d

kfinch:
I got a little burned out and didn't work much on this, will be back to finish it out in a few days.

src/parser/shared/metrics/activeTime.ts
MartijnHols: I think it's fair since players can play around it by casting an instant cast or auto attacking. Casting a 3 sec cast while the boss is dying in 1.5 is the wrong decision.

src/parser/shared/metrics/activeTime.ts
MartijnHols: Just style. I like to think of separate files as separate features that can be used by itself, and when channelTimePeriods can have its own live outside of active time, I think it makes sense as a separate file.

src/parser/core/metric.ts
MartijnHols: This should be cleaned up in shadowlands now.

MartijnHols:
> I got a little burned out and didn't work much on this, will be back to finish it out in a few days.

Oof. No worries - take your time.

emallson:
Closing due to (almost) 1yr of inactivity

PR #4569, opened 2021-07-14
Updated combatantinfo to match changes to WCL API

kfinch:


kfinch:
The expected results of integration tests didn't change, so I updated the test inputs by search replacing the combatants.json to substitute in the new field names

PR #4571, opened 2021-07-20
TBC-Priest: Initial Buildout

TJMoats:
This PR adds very basic support for Priests in TBC. No custom modules are built out, but the page loads, you get some basic metrics, and the mana efficiency tab loads. 

I also fixed a crash on the character sheet for TBC characters.

6yM92Cv
FJDjiLI
5TfyZvJ
Juhq91V
ydbH1AF
hZgpFvO


analysis/monkmistweaver/src/integrationTests/example.test.ts.snap
```diff
@@ -6035,7 +6035,7 @@ exports[`Mistweaver Monk integration test: example log MistweaverHealingEfficien
           <td
             className="text-right"
           >
-            2080000
+            2.08m
```
MartijnHols, line 6043: this entire file diff seems troublesome

analysis/tbcpriest/src/SPELLS.ts
```diff
@@ -0,0 +1,94 @@
+const spells = {
```
MartijnHols: Hey, real nice that you're colocating your spells in your spec folder, but you no longer should have to specify the name/icon for each spell. Just an id should be sufficient, the rest of the info is either pulled from the log or from the WCL API. Less work for us! See tbc-hunter for an example.

analysis/tbcpriest/src/modules/features/HealingEfficiencyBreakdown.tsx
```diff
@@ -0,0 +1,80 @@
+import BaseHealingEfficiencyBreakdown from 'parser/core/healingEfficiency/HealingEfficiencyBreakdown';
+import React from 'react';
+import Toggle from 'react-toggle';
+
+class HealingEfficiencyBreakdown extends BaseHealingEfficiencyBreakdown {
```
MartijnHols: What did you change here from the base? Is this not already in the base?

MartijnHols (changes requested):
Lots of extra fixes - thanks! LGTM apart from the snapshot.

analysis/monkmistweaver/src/integrationTests/example.test.ts.snap
TJMoats: The current version of the mana efficiency module only includes spells that have mana costs defined on the spell. I changed it so that it grabs the mana costs from the casts. There is more data being shown for mana efficiency now. 

I also fixed an issue with some infinite values breaking the chart, and added some number formatting to make it cleaner. 

The various changes in this snapshot are from all of this ^^

analysis/tbcpriest/src/modules/features/HealingEfficiencyBreakdown.tsx
TJMoats: I removed the duplicate code.

analysis/tbcpriest/src/SPELLS.ts
TJMoats: How do you get a reference to the icon? I'm used to the SPELLS.SPELL_NAME.icon bit for adding icons to analyzers.

TJMoats:
@MartijnHols I added a comment about the snapshot and removed some of the redundant code.

analysis/tbcpriest/src/SPELLS.ts
MartijnHols: If you use the SpellIcon component it will automatically fetch the icon for the spell if it wasn’t in the log.

analysis/monkmistweaver/src/integrationTests/example.test.ts.snap
MartijnHols: That's not the only issue in it though, such as further below has a NaN and Infinity whereas it had numbers before, and the color changed:

image

PR #4592, opened 2021-08-14
Add initial TBC (arcane) mage module

BjarkeM:
I've created an initial template for very basic functionality for arcane mages in TBC classic.

It's been tested on a few of my gruul/mag and karazhan logs and I think I covered all of the spells that will be used in the spellbook.

No useful suggestions are made as it's my first foray into the codebase, but I added relevant CDs to the timeline and statistics.

image
image
image
image
image


MartijnHols (approved):
This looks like a very complete initial setup. Short of suggestions and statistics I think you actually got everything in one go. Thanks!

PR #4593, opened 2021-08-19
[Death Knight] Deaths due module

SpencerP23:
Added statistics module for Deaths Due, which measures:

- Avg strength gained throughout encounter (main)
- Buff uptime (useful for blood since it can have 100% uptime)
- Avg Stacks throughout encounter (This is technically avg strength/2 but still neat to see)
- Avg Stacks per cast (Useful for Frost/unholy as they aim to hit 4 stacks each cast)

As well as two suggestions:

- For blood only: Death's Due Buff uptime (essentially just moved it from the spell)
- For all specs: Wasted cast warning (casts where no buff stack is gained from using the rune spender ability within it)

A little info on how the handleWindow works in DeathsDue (in case the comments aren't helpful enough): The ground effect lasts 10s, and the buff lasts 12s, so however many stacks are at the end of the window is how many stacks its "worth." Since I had trouble getting buff stacks in the future, I decided to get the buff stacks in the past. So each time the EventHandler is called, it takes the stack count when the last window ended and preps for the next by marking when this window will end. So fightEnd would just measure the last cast, but I also put something in to not measure the last cast if they don't get a full window.

This one is different from handleStacks because handleWindow gets us the stacks just at the end of the cast, while handleStacks measures buff throughout the encounter for stack uptime purposes.

I know barely anyone plays NF DK but dangit I do and more should it's fun.


analysis/deathknight/src/DeathsDue.js
```diff
@@ -0,0 +1,244 @@
+import { t } from '@lingui/macro';
```
MartijnHols, line 1: Please consider writing new modules in TypeScript. We've been migrating slowly to TypeScript for many benefits, and I reckon it would really enhance your coding experience and is worth the trouble of learning.

analysis/deathknight/src/DeathsDue.js
```diff
@@ -0,0 +1,244 @@
...
+    //get stuff set up for the current cast, if this is a cast
+    if (event.type === EventType.Cast) {
+      this.casts += 1;
+      this.groundEffectEnd = event.timestamp + GROUND_EFFECT_DURATION;
+    }
+  }
+
+  get averageStacks() {
+    let avgStacks = 0;
+    this.stacks.forEach((elem, index) => {
```
MartijnHols, line 102: Writing this as a `reduce` could be slightly cleaner, but this works too.

analysis/deathknight/src/DeathsDue.js
```diff
@@ -0,0 +1,244 @@
...
+          <>
+            <strong>Uptime: </strong> {formatPercentage(this.uptime)}% <br />
+            <strong>Avg Stacks Throughout Encounter: </strong> {this.averageStacks.toFixed(1)}
+            <br />
+            <strong>Avg Stacks Per Cast: </strong> {this.avgStackPerCast.toFixed(1)}
+          </>
+        }
+      >
+        <BoringSpellValueText spellId={SPELLS.DEATHS_DUE.id}>
+          <>
```
MartijnHols, line 235: This fragment can be omitted.

analysis/deathknight/src/DeathsDue.js
```diff
@@ -0,0 +1,244 @@
...
+      },
+      style: 'percentage',
+    };
+  }
+
+  suggestions(when) {
+    if (this.selectedCombatant.spec === SPECS.BLOOD_DEATH_KNIGHT) {
+      when(this.uptimeSuggestionThresholds).addSuggestion((suggest, actual, recommended) =>
+        suggest(
+          <>
```
MartijnHols, line 184: Please use the `<Trans id="x">` component for strings so they can be localized.

MartijnHols (approved):
Sorry for the late review. I am not going to hold off on merging this any longer, so these comments are just for future reference. There's also not a big deal. Thanks for contributing!

PR #4594, opened 2021-08-27
Shadow priest: Updated module with patch 9.1 changes

marson:
I implemented the changes to the Talbadars Stratagem legendary and the Dissonant echoes conduit from the 9.1 patch and bumped the supported patch version to 9.1.0.

Question:
Should I create 1 changelog entry describing all the changes or 3 changelog entries 1 for each change?


DylanDirlam (approved):
All looks good! For change log, I've seen most people do individual entries for each change and that's what I prefer. I don't think there's a required way to do that though. Welcome aboard :)

PR #4601, opened 2021-09-06
TbcDruid: Initial Build Out

TJMoats:
This PR gets druid to be clickable, but doesn't implement any custom analyzers. Those will come shortly.


MartijnHols:
I assume the FLASH_HEAL is just a placeholder. LGTM. Thanks!

PR #4607, opened 2021-09-28
Affliction warlock updated shadow embrace suggestions and dark glare cooldown

Pirrang:
For 9.1 Shadow embrace has become a talent instead of a default spell, and Dark Caller has become a default spell instead of a talent. Leading to wrong suggestion regarding the shadow embrace uptime
image


MartijnHols (approved):
Nice, thanks!

PR #4615, opened 2021-10-01
Fix splitting ice img

npaganini:
Fixed the frost mage talent `splitting ice` as it was missing the icon in the `character/ZONE/Realm/character-name/` page.

BEFORE
Screen Shot 2021-10-01 at 16 01 30

AFTER
Screen Shot 2021-10-01 at 16 01 51


Sharrq:
This change will break the functionality of Frost Mage. Also that ID doesnt seem to exist??? https://www.wowhead.com/spell=429385

npaganini:
> This change will break the functionality of Frost Mage. Also that ID doesnt seem to exist??? https://www.wowhead.com/spell=429385

Hello there @Sharrq 
Forgive me, first time contributing to this project. I was using the icon id instead of the spell id, I had misunderstood. I correctly fixed the icon without altering the spell. Functionality should work now

Sharrq:
If it still fixes the icon without changing the id, then it's fine.

I don't work here anymore, I just didn't want to see the only functioning mage spec get broken because of a simple mistake.

npaganini:
> If it still fixes the icon without changing the id, then it's fine.
> 
> I don't work here anymore, I just didn't want to see the only functioning mage spec get broken because of a simple mistake.

Great! @emallson who can approve the PR?
Is the PR missing something? Please let me know

emallson (approved):

PR #4622, opened 2021-10-08
Shadowlands Arms minor spells changes

elasticspoon:
Took Avatar off GCD and Sweeping Strikes is half GCD. 
Implemented seasoned soldier passive - auto crits generate 30% more rage. 

I plan to fix a bunch more stuff but starting small here.

### DIFF
--- a/analysis/warriorarms/src/modules/features/RageTracker.js
+++ b/analysis/warriorarms/src/modules/features/RageTracker.js
@@ -24,7 +24,11 @@ class RageUsage extends ResourceTracker {
   }
 
   onDamage(event) {
-    this.processInvisibleEnergize(SPELLS.RAGE_AUTO_ATTACKS.id, RAGE_PER_MELEE_HIT);
+    if (event.hitType === 2) {
+      this.processInvisibleEnergize(SPELLS.RAGE_AUTO_ATTACKS.id, RAGE_PER_MELEE_HIT * 1.3);
+    } else if (event.hitType === 1) {
+      this.processInvisibleEnergize(SPELLS.RAGE_AUTO_ATTACKS.id, RAGE_PER_MELEE_HIT);
+    }
   }
 }
 
--- a/analysis/warriorarms/src/modules/features/RageTracker.js
+++ b/analysis/warriorarms/src/modules/features/RageTracker.js
@@ -1,4 +1,5 @@
 import SPELLS from 'common/SPELLS';
+import HIT_TYPES from 'game/HIT_TYPES';
 import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import { SELECTED_PLAYER } from 'parser/core/Analyzer';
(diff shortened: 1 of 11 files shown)


analysis/warriorarms/src/modules/features/RageTracker.js
```diff
@@ -24,7 +24,11 @@ class RageUsage extends ResourceTracker {
   }
 
   onDamage(event) {
-    this.processInvisibleEnergize(SPELLS.RAGE_AUTO_ATTACKS.id, RAGE_PER_MELEE_HIT);
+    if (event.hitType === 2) {
```
emallson: you should use the `HIT_TYPE` enum (`import HIT_TYPES from 'game/HIT_TYPES'`) instead of a magic number here:

```suggestion
    if (event.hitType === HIT_TYPES.CRIT || event.hitType === HIT_TYPES.BLOCKED_CRIT) {
```

analysis/warriorarms/src/modules/features/RageTracker.js
```diff
@@ -24,7 +24,11 @@ class RageUsage extends ResourceTracker {
...
+      this.processInvisibleEnergize(SPELLS.RAGE_AUTO_ATTACKS.id, RAGE_PER_MELEE_HIT * 1.3);
+    } else if (event.hitType === 1) {
```
emallson: same thing here

```suggestion
    } else if (event.hitType === HIT_TYPE.NORMAL) {
```

or simply

```suggestion
    } else {
```

if there are no other conditions where the hit might (or might not) generate rage.

emallson (changes requested):
looks great overall, minor request in `RageTracker` to avoid scattering magic numbers around.

analysis/warriorarms/src/modules/features/RageTracker.js
elasticspoon: Sure, I added HIT_TYPE.

analysis/warriorarms/src/modules/features/RageTracker.js
elasticspoon: Based on looking at a few logs I think melee hits only give rage if they are of hit types of NORMAL or CRIT. PARRY does not and I have not seen other types of hits in pve logs. So I don't think the if statement can be simplified.

elasticspoon:
I don't know if I should keep adding more small ish stuff to this pull request or not. But I also added some files for SOD items mainly to implement the haste for Old Warrior Soul trinket properly.

emallson (approved):

emallson:
> I don't know if I should keep adding more small ish stuff to this pull request or not. But I also added some files for SOD items mainly to implement the haste for Old Warrior Soul trinket properly.

Generally best to keep 1 PR for 1 feature, but this was a small thing so its okay. Thanks!

PR #4623, opened 2021-10-09
APL Checker

emallson:
This PR introduces a new core metric based on checking spell casts against an APL. This metric is more robust against difficult-to-measure fight downtime than cast efficiency, while also allowing automatic annotation of incorrect spell casts on the timeline with little additional effort.

At this time, I have not yet completed a proof-of-concept for the Brewmaster APL but it is in-progress. The APL checker works and has tests.

## Background

One of the constant issues over the past half-decade of working on WoWAnalyzer has been the reliance on *Cast Efficiency* as a metric. While this is a useful metric, it breaks down in automated use in a couple of ways:

1. Reasonable targets for cast efficiency vary with both fight and strategy within a fight. This has led to the development of complex (and time-intensive) methods to automatically strip out downtime that have never been *quite* reliable enough to make the default view mode.
2. Setting targets for cast efficiency is hard in a vacuum. Most classes have complex interactions between spells that mean it is *sometimes* correct to delay a spell. This leads to most spells being set to a simple 80%-ish cast efficiency as a "good enough" value that will flag egregious outliers but isn't very useful for catching minor mistakes.

In particular, I'd say one of the biggest issues right now is that the default metric we use comes with so many caveats that a novice user can't really interpret the results in detail. If they're making egregious errors, then we have can give high-level feedback---but more moderate errors are out of reach.

Both the Guardian Druid and Brewmaster Monk analyzers have special analysis for "Filler" spells that rely on very basic APL checking (i.e. you should never cast Tiger Palm when Keg Smash or Blackout Kick are available). This is used to automatically flag "bad casts" of these spells, but (as this PR will hopefully show) this idea can be extended to handle more complex APLs than that.

## Goals

I would like this to eventually become the "default" metric for most analysis for DPS, as well as for the DPS portion of Tank specs. The theory community is generally very comfortable with APLs as a concept, and automating a bunch of the features (like a basic checklist, timeline automation, dynamic cast efficiency thresholds) based on an APL built from (for example) a Wowhead or Icy-Veins guide is much more approachable.

So, the goals are:

- Easy to get started with
- Accurate with minimal fiddling
- Able to handle complex conditions & APLs

The immediate output I expect from this is timeline annotation, though I'd like to use it for other features as well.

## Non-Goals

- Fully replace Cast Efficiency metrics. This idea can never replace cast efficiency for healers---heals don't follow an APL. Tanks may be able to move on from CE in most situations but will likely still need to use it in a few outliers. The goal is *not* to base everything on an APL, but to build features based on it where it makes sense.
- Fully replace Downtime metrics. APLs don't measure whether you spent a lot of time between spell casts. Any APL usage would have to be in tandem with downtime measurement.
- Use SimC APLs. SimC APLs tend to be hyper-specific to the properties of the simulation itself, and aren't going to be usable as-is in WoWA even if we could parse them. Part of the work of theorycraft is taking the SimC APL and synthesizing it into an understandable list of rules that is usable in practice.

## Features

- [x] Validate each spell cast against a priority list of spells, listing the cast events that don't follow the priority.
- [x] Support "conditional rules" such as "this spell has priority when a buff is active"
- [x] Automatically annotate the timeline with incorrect spell casts
- [x] (future?) Automatically generate checklist section with APL check results

## Implementation Notes

The APL checker itself is pretty straightforward: given a list of rules, check every spell cast to see if the current highest-priority rule matches. Only spells that are present in the list of rules are checked, which helps avoid some issues with casts that aren't casts (others, like channeling Arcane Missiles, will of course need a normalizer).

The more complicated part is handling *conditional rules*. At the core, each rule can have a condition that determines if the rule applies to a cast event. This is done via a `Condition` object with a `validate(state, event)` method. 

For the `Condition` objects, I used an external-state pattern that I've had success with elsewhere: each `Condition` has pure methods to initialize, update, and consume a state object that is tracked by the `aplCheck` function. This means that multiple `aplCheck`s can run concurrently (though I don't know why you would), and there are no issues with sharing conditions across many rules.


emallson:
Preview of the checklist

Screenshot from 2021-10-16 18-46-46

The timeline is also automatically annotated.

The complete APL that is used for this is:

```tsx
export const apl = build([
  SPELLS.BONEDUST_BREW_CAST,
  { spell: SPELLS.KEG_SMASH, condition: buffPresent(SPELLS.WEAPONS_OF_ORDER_BUFF_AND_HEAL) },
  { spell: SPELLS.BREATH_OF_FIRE, condition: hasLegendary(SPELLS.CHARRED_PASSIONS) },
  SPELLS.KEG_SMASH,
  SPELLS.BLACKOUT_KICK_BRM,
  SPELLS.FAELINE_STOMP_CAST,
  SPELLS.BREATH_OF_FIRE,
  {
    spell: SPELLS.RUSHING_JADE_WIND_TALENT,
    condition: buffMissing(SPELLS.RUSHING_JADE_WIND_TALENT, {
      timeRemaining: 2000,
      duration: 6000,
    }),
  },
  { spell: SPELLS.SPINNING_CRANE_KICK_BRM, condition: buffPresent(SPELLS.CHARRED_PASSIONS_BUFF) },
  SPELLS.CHI_WAVE_TALENT,
  SPELLS.CHI_BURST_TALENT,
  {
    spell: SPELLS.SPINNING_CRANE_KICK_BRM,
    condition: optional(
      hasConduit(SPELLS.WALK_WITH_THE_OX),
      <>
        It is worthwhile to cast <SpellLink id={SPELLS.SPINNING_CRANE_KICK_BRM.id} /> over{' '}
        <SpellLink id={SPELLS.TIGER_PALM.id} /> when using this conduit <em>if</em> doing so would
        get you an extra cast of <SpellLink id={SPELLS.INVOKE_NIUZAO_THE_BLACK_OX.id} /> that lines
        up with incoming damage. We cannot check this automatically, and be warned that it is a
        small defensive loss due to the loss of Brew cooldown reduction.
      </>,
    ),
  },
  SPELLS.TIGER_PALM,
]);
```

The checklist rule can then be added by using the `aplCheck` method in the Checklist module and passing the results to `<AplRule {...props} />` (see the Brewmaster code for worked example).

The timeline annotation is easiest to implement as a functional suggestion, though unfortunately this runs the APL checker twice. It is quite fast (30ms on my machine), but this is on the list of improvements for later.

```tsx
export const check = aplCheck(apl);

const brmApl = (): WIPSuggestionFactory => (events, info) => {
  const { violations } = check(events, info);
  annotateTimeline(violations);

  return undefined;
};

export default brmApl;
```

Dambroda:
The code itself all looks good, some theoretical questions that might make implementation of this hard for some specs is AoE APL depending on how many enemies you hit. For frost mage for example, we often check for hitting two targets with cleave for some suggestions. For now is this APL just going to assume a single target rotation, or are there plans to expand the conditions such that it could support a feature like that? (From my perusal it looks like it would require a bit of a refactor to change how conditions work to not only use cast events)

emallson:
> The code itself all looks good, some theoretical questions that might make implementation of this hard for some specs is AoE APL depending on how many enemies you hit. For frost mage for example, we often check for hitting two targets with cleave for some suggestions. For now is this APL just going to assume a single target rotation, or are there plans to expand the conditions such that it could support a feature like that? (From my perusal it looks like it would require a bit of a refactor to change how conditions work to not only use cast events)

My idea there had been to introduce a "lookahead" feature. Since this is a functional analyzer, it actually gets the full event list. So the idea would be that on a condition you would add a `lookahead: number` key telling it how many seconds you wanted to look ahead and then you'd get the events from `currentTime` to `currentTime + lookahead` as an extra parameter during validation.

Then, for example, the condition for aoe targets would look ahead `n` seconds for `m` distinct target IDs being hit.

So it'll require a small refactor, but just a small one.

PR #4625, opened 2021-10-11
Fix crashes in TBC Warlock

emallson:
two issues:

1. Curse modules were not added to CLP
2. Curse modules used non-existant spells


Talador12:
LGTM - The prior PR was my first change to this project. Good fixes!

PR #4626, opened 2021-10-13
Localization and TS updates

ckaczor:
- Convert Armory icon file from JS to TSX
- Localization in Armory icon file
- Localization of Armory text

A first round of changes to make sure I'm on the right track.  I'll do more (in small-ish batches) once this is approved.

### DIFF
--- a/src/interface/icons/Armory.tsx
+++ b/src/interface/icons/Armory.tsx
@@ -1,0 +1,15 @@
+import { t } from '@lingui/macro';
+import React from 'react';
+
+import ArmoryLogo from './images/Armory-logo.png';
+
+const Icon = (props: any) => (
+  <img
+    src={ArmoryLogo}
+    alt={t({ id: 'interface.armory.alt', message: 'World of Warcraft Armory' })}
+    className="icon"
+    {...props}
+  />
+);
+
+export default Icon;
--- a/src/interface/icons/Armory.tsx
+++ b/src/interface/icons/Armory.tsx
@@ -3,7 +3,9 @@ import React from 'react';
 
 import ArmoryLogo from './images/Armory-logo.png';
 
-const Icon = (props: any) => (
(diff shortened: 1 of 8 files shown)


src/interface/icons/Armory.tsx
```diff
@@ -0,0 +1,15 @@
+import { t } from '@lingui/macro';
+import React from 'react';
+
+import ArmoryLogo from './images/Armory-logo.png';
+
+const Icon = (props: any) => (
```
emallson: when typing a wrapper component like this, you should use something like:

```suggestion
type Props = Omit<React.ComponentPropsWithoutRef<'img'>, "src" | "alt" | "className">;
const Icon = (props: Props) => (
```

that way users of the icon get direct typechecking support

emallson (changes requested):
Good start! Would appreciate using a type for the `Icon` props instead of `any`.

Also: are there uses of the `<ArmoryIcon />` component that *aren't* immediately followed by the ` Armory` text? if not it may make sense to make a shortcut since the `Trans` version is a little more involved. That doesn't need to be in this PR though.

src/interface/icons/Armory.tsx
ckaczor: Thanks for the feedback - I just pushed a change with proper typing.

It looks like DEFAULT_BUILD.tsx uses ArmoryIcon without any text afterwards.  I'm not quite sure what you mean by shortcut - I'm pretty new to React so maybe that's something I haven't come across yet.

Thanks!

emallson: By a shortcut i just meant a component that had the text and `<Trans>...</Trans>` element as part of the output, but since it is sometimes used without that part it doesn't need it. Thanks!

emallson (approved):

PR #4629, opened 2021-10-15
Add support for Fiery Soul legendary to Vengeance Demon Hunter, and fix Fury Usage costs

WarcraftYax:
## Fiery Soul

Fiery Soul is run by most Vengeance Demon Hunters, and it reduces Fiery Brand cooldown based on consuming Soul Fragments with Soul Cleave. Because this cooldown reduction wasn't supported in WoWA, the Fiery Brand timeline would often look as such:

image

And the Abilities section could show uptimes of >100% with the user casting more times than what the analyzer thought was possible:

image

I updated `SoulFragmentsConsume` to fabricate an event with `EventEmitter` when fragments are consumed, and added a `FierySoul` module to listen to these events (if the user has the legendary equipped) and reduce the cooldown using the `SpellUsable` module.

Now the Timeline and Abilities sections make more sense:

image

image

I also added a simple statistic to show the user how much cooldown reduction they are gaining from the legendary:

image

I used this log mostly for testing: `/report/nbmwNtLjcByDg71r/24-Mythic+Guardian+of+the+First+Ones+-+Kill+(5:07)/Xyä/standard/overview`

## Fury Usage

Separately, I made a quick fix to the `FuryUsage` module as it was dividing Fury costs by 10 unnecessarily.

Before:

image

After:

image


src/parser/core/modules/EventEmitter.ts
```diff
@@ -289,7 +289,10 @@ class EventEmitter extends Module {
     }
   }
   // todo double check this 'event' shape... seems wrong
-  fabricateEvent(event: { type: EventFilter<any> | string }, trigger: any = null) {
+  fabricateEvent(
+    event: { type: EventFilter<any> | string; [additionalProperties: string]: any },
```
WarcraftYax, line 293: TypeScript wasn't happy when this was just `{ type: EventFilter<any> | string; }`, so I added the `additionalProperties`. This makes sense to me, and I see there is a todo comment above the method... if this is correct then I can remove the todo. If it's not then I can fix it however it needs to be fixed!

emallson: i'm not familiar enough with the context of that TODO so I'd need to double-check. Seems like `event` should actually just be an `Event<any>`? not sure. Not your problem though

emallson (approved):

PR #4633, opened 2021-10-18
Warrior APL initial + additional APL conditionals to make it work

elasticspoon:
Basically, I was trying to set up an APL for arms and there were a few conditions that I needed but they were not implemented. I gave implementing them a shot, but I'm sure they could be done better. 

I don't know if a pull request is the best way to do this or if an issue is better but whatever. 

Stuff I tried my hand at adding:

- spellCharges - spell priority can change based on how many charges you have of it. Ex: overpower higher priority if capped on charges. 
- buffStacks - I made lessThanBuffStacks and moreThanBuffStacks. sometimes the buff being present is not enough and you cast the ability at a certain amount of stacks. maybe this can be rolled into buff present somehow?

This is other stuff I ran into needing for arms APL and general suggestions: 
- buffPresent - buffMissing has a time buffer of 100ms if the cast itself applies a buff. buff present should have the same for if the spell is consuming a buff. EX: casting mortal strike when its cost is reduced. 
- classResource - some spells are only cast when high or low on resource. some way to track it, maybe based on energize events? 
- debuffPresent - same as buff but on spell target. (maybe im missing something and this is already in?)
- targetInExecute - pretty self explanatory. execute rotations change. 
- timeRemainingOnCd - again sorta complex but some spells only get cast with x seconds of another spell coming off cooldown.


emallson:
Hey! these look like a good start, but i'm a little wary of adding conditions that aren't used yet because of how quirky the logs can be. Code that appears correct can be slightly off due to weird buff ordering etc. That said, you need quite a lot of new conditions for the Arms APL so I'm sympathetic.

I will be working on the protection paladin APL later this week, which will need:

- `classResources`
- something for delaying wings until seraphim is available (similar to `timeRemainingOnCd`

so that should help lighten the load for Arms.

I'm also going to put together a wiki page with instructions for use.

emallson:
I added `inExecute` and `hasResources` in #4642 @elasticspoon

elasticspoon:
Built out the majority of Arms APL. Still a few things that need adjusting but most of the logic is correct. 
image


Needs:  
- `remaingTimeOnCD `
- `debuffPresent`
- a lot of the code would probably be cleaner with a `not` conditional

Stuff in my commits. Sorry there are so many. 
- fixed `buffStacks` conditional to work with range
- fixed `chargeStacks` to work with range
- added `hasTalent`
- added `noTalent` and `noLegendary` 
- added a warrior specific execute conditional for venthyr and massacare 

Other Todos:
- Fix the front end
- Add dots (idk if I want to add them to APL or just modules)

Still needs to be made prettier. And there is a shit ton of sloppy/ reused code. 

Also the `formatRange` function used for the `hasResource` and is probably worth shoving somewhere to be reused.

analysis/warriorarms/src/modules/checklist/Component.tsx
```diff
@@ -101,7 +121,7 @@ const ArmWarriorChecklist = ({ combatant, castEfficiency, thresholds }: Checklis
         {combatant.hasLegendaryByBonusID(SPELLS.ENDURING_BLOW.bonusID) && (
           <AbilityRequirement spell={SPELLS.MORTAL_STRIKE.id} />
         )}
-      </Rule>
+      </Rule> */}
```
emallson: probably worth deleting instead of commenting out. git will always have the old version anyway

analysis/warriorarms/src/modules/core/AplCheck.tsx
```diff
@@ -0,0 +1,88 @@
...
+        timeRemaining: 2400,
+        duration: 8000,
+      }),
+      cnd.hasConduit(SPELLS.ASHEN_JUGGERNAUT),
+    ),
+  },
+  //2 Overpower Charges
+  {
+    spell: SPELLS.OVERPOWER,
+    condition: cnd.spellCharges(SPELLS.OVERPOWER, { atLeast: 2, atMost: 2 }),
```
emallson, line 23: we should either add an `exactly: 2` option or adjust the range formatter to handle the case where `atLeast === atMost`. i can handle it after a merge or you can, up to you.

analysis/warriorarms/src/modules/core/AplCheck.tsx
```diff
@@ -0,0 +1,88 @@
...
+    spell: SPELLS.MORTAL_STRIKE,
+    condition: cnd.and(
+      cnd.buffStacks(SPELLS.OVERPOWER, { atLeast: 2 }),
+      cnd.buffStacks(SPELLS.EXPLOITER, { atLeast: 2 }),
+      inExecute(),
+    ),
+  },
+  {
+    spell: SPELLS.MORTAL_STRIKE,
+    condition: cnd.and(cnd.buffStacks(SPELLS.OVERPOWER, { atLeast: 2 }), outsideExecute()),
```
emallson: I guess this is why you said a `not` would be useful? I think that sounds fine.

src/parser/shared/metrics/apl/conditions/buffStacks.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (range.atLeast !== undefined && range.atMost !== undefined) {
+    return `${range.atLeast}-${range.atMost}`;
+  } else if (range.atLeast !== undefined) {
+    return `at least ${range.atLeast}`;
+  } else if (range.atMost !== undefined) {
+    return `at most ${range.atMost}`;
+  } else {
+    return '';
+  }
+}
```
emallson: instead of copying this, you should import it. if the compiler complains because of a circular dep, we can move it to a `range.ts` file or something.

src/parser/shared/metrics/apl/conditions/hasLegendary.tsx
```diff
@@ -18,3 +18,18 @@ export default function hasLegendary(legendary: LegendarySpell): Condition<boole
     ),
   };
 }
+
+export function hasNoLegendary(legendary: LegendarySpell): Condition<boolean> {
```
emallson: another prime use for `not`

src/parser/shared/metrics/apl/conditions/hasTalent.tsx
```diff
@@ -0,0 +1,33 @@
...
+    validate: (state, _event) => state,
+    describe: (tense) => (
+      <>
+        you {tenseAlt(tense, 'have', 'had')} <SpellLink id={talent.id} /> talented
+      </>
+    ),
+  };
+}
+
+export function hasNoTalent(talent: Spell): Condition<boolean> {
```
emallson: 100% sold on `not`.

src/parser/shared/metrics/apl/conditions/spellCharges.tsx
```diff
@@ -0,0 +1,41 @@
...
+  if (range.atLeast !== undefined && range.atMost !== undefined) {
+    return `${range.atLeast}-${range.atMost}`;
+  } else if (range.atLeast !== undefined) {
+    return `at least ${range.atLeast}`;
+  } else if (range.atMost !== undefined) {
+    return `at most ${range.atMost}`;
+  } else {
+    return '';
+  }
+}
```
emallson: same as previous comment about `formatRange`

analysis/warriorarms/src/normalizers/BattlelordBuff.ts
```diff
@@ -0,0 +1,26 @@
...
+    afterEventType: EventType.RemoveBuff,
+    bufferMs: 50,
+    anyTarget: true,
+  },
+];
+
+/**
+ * Ensures that the Battlelord Buff is removed after mortal strike is cast.
+ */
+class BattlelordNormalizer extends EventOrderNormalizer {
```
emallson, line 20: is this normalizer needed if I adjust the `buffPresent` condition as you suggested? maybe still useful for other modules?

analysis/warriorarms/src/modules/core/inExecuteWarrior.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { Condition, tenseAlt } from 'parser/shared/metrics/apl';
+import React from 'react';
+
+const EXECUTE_RANGE = 0.2;
+const EXECUTE_RANGE_MASSACRE = 0.35;
+const EXECUTE_RANGE_VENTHYR = 0.8;
+
+export type TargetHealthRange = { [targetId: number]: number; [threshold: string]: number };
+
+export function inExecute(): Condition<TargetHealthRange> {
```
emallson: hadn't thought about the above/below executes. MM has (or had?) a similar one. maybe we make the standard `inExecute` accept a range that defaults to `{ atMost: 0.2 }` and then the warrior one with Condemn becomes `or(healthInRange({ atMost: 0.2 }), healthInRange({ atLeast: 0.8 })`? I kinda like the `inExecute` name even if that is no longer exactly accurate though...

emallson (commented):
On the whole, this is great! Curious what you think of the APL system so far.

I think that adding a `not` condition as you suggested would do a lot to reduce the need for custom conditions. Goal is to basically never need to write your own condition, only to combine them using `and`, `or`, `not`.

That said, I don't think we need to hold up your PR for further adjustments to the conditions if the APL analysis itself is ready to go up. We can make those internal changes later. The only thing I'd really say blocks this right now is the copy/paste of `formatRange`, which is an easy enough fix.

analysis/warriorarms/src/modules/core/inExecuteWarrior.tsx
elasticspoon: After messing around for a while the idea of making the argument for `inExecute` be a range just resulted in too many issues if someone added a top and bottom end. I ended up modifying the `inExecute` to accept a negative value and use that as a upper threshold.

analysis/warriorarms/src/normalizers/BattlelordBuff.ts
elasticspoon: potentially. I don't there is any harm to it. But the change  would solve future issues for classes  with no normalizers.

analysis/warriorarms/src/modules/core/AplCheck.tsx
elasticspoon: I adjusted the `formatRange` to cover that situation. and stuck it in `index.ts`. no idea if there is a better place for it.

analysis/warriorarms/src/modules/checklist/Component.tsx
elasticspoon: done

emallson (approved):
looks good! thanks!

analysis/warriorarms/src/modules/core/inExecuteWarrior.tsx
emallson: ah makes sense. using and/or for it works and probably makes `inExecute` simpler too

analysis/warriorarms/src/modules/core/AplCheck.tsx
emallson: index is good for now. still figuring out the organization of this stuff

analysis/warriorarms/src/normalizers/BattlelordBuff.ts
emallson: i will plan to get that change in with some future APL updates

emallson:
(don't forget a changelog entry!)

elasticspoon:
Overall, the APL is 90% functional. The other 10% is held back by not having a `hasDebuff`. 
I have probably missed some glaring typos but oh well. At this point this can probably be merged?

Stuff I did:
- implemented `not` (descriptions are a bit sus)
- fixed other conditions and APL to include not
- added upper threshold option to `inExecute` maybe not done the prettiest way
- move `formatRange` into `index.ts`

Stuff left:
- conditionals based on dots. future problem. 

Also:
`not` is still sloppy. not really a functionality issue but the text output does not English well. I'm not really sure how to append a word to the description of other conditionals and make the English make sense consistently. 
Ex: 
6. Cast Slam when you have Fervor of Battle talented - good English
6. Cast Slam when don't you have Fervor of Battle talented - bad English

emallson:
for the `not` wording: it probably makes sense to just have an optional param to override the description. i don't think english as a language is structured enough to make that something we can automate for every case. what you've got now is a good enough start that i don't think we need to wait for that. the next person to use it can add a parameter.

PR #4634, opened 2021-10-18
Small localization changes and another batch of basic TypeScript conversions/improvements

ckaczor:
Last batch of updates to the icons with some associated localization changes.


src/localization/en/messages.json
```diff
@@ -475,6 +475,9 @@
   "interface.specListItem.unmaintained": "CURRENTLY UNMAINTAINED",
   "interface.statistics.components.gauge.high": "High",
   "interface.statistics.components.gauge.low": "Low",
+  "interface.warcraftLogs": "Warcraft Logs",
+  "interface.warcraftLogs.timeline": "Warcraft Logs timeline",
+  "interface.wipefest": "Wipefest",
```
emallson: these are proper names so I don't believe that they get translated. the "timeline" part might, but checking the WCL website they don't translate their name on other language variants.

emallson (changes requested):
Type changes look good. I don't believe the translation changes are needed because generally proper names don't get localized in this context. `interface.warcraftLogs.timeline` can get localized since it is more than just the name, but for now lets leave the `Trans` off of WCL/Wipefest.

src/localization/en/messages.json
ckaczor: Good point about the proper names - I've reverted the changes back to just TS changes and any localization (like "timeline") can come in another PR later.

emallson (approved):

PR #4637, opened 2021-10-19
Sanctum of Domination boss info and images

DylanDirlam:
Adds all 10 bosses, their phases, vantus runes, and images for the 9.1 raid


src/game/raids/sanctumofdomination/SylvanasWindrunner.ts
```diff
@@ -0,0 +1,82 @@
...
+        difficulties: [
+          DIFFICULTIES.NORMAL_RAID,
+          DIFFICULTIES.HEROIC_RAID,
+          DIFFICULTIES.MYTHIC_RAID,
+        ],
+        filter: {
+          type: EventType.Health,
+          guid: 175732,
+          health: 74.5,
+        },
```
emallson, line 76: this would probably be more reliable if based on the Blasphemy cast (on Mythic you're not likely to reach 74.5% in P2, at least during prog), but this is a much needed PR so i don't see a reason to hold up the rest of the changes for such an improvement

src/game/raids/sanctumofdomination/FatescribeRohKalo.ts
```diff
@@ -0,0 +1,74 @@
...
+        difficulties: [
+          DIFFICULTIES.NORMAL_RAID,
+          DIFFICULTIES.HEROIC_RAID,
+          DIFFICULTIES.MYTHIC_RAID,
+        ],
+        filter: {
+          type: EventType.ApplyBuff,
+          ability: {
+            id: SPELLS.EXTEMPEROUS_FATE.id,
+          },
```
emallson, line 67: Extemporous Fate is applied about a minute into P3, so this should likely be HP-based instead. However, similar to the Sylvanas comment I think its okay to have a minor inaccuracy that we can fix in a followup PR.

emallson (approved):
Overall looks good, a couple changes that I think are needed to improve phase accuracy but not worth holding up the PR for:

- Fatescribe P3
- Sylvanas P3

PR #4653, opened 2021-11-03
[MW] More Clean up

abelito75:
A lot of this is general clean up and moving panels around. A lot of these panels being moved around is more specific to theory-crafting or just things that are nearly impossible to calculate without wowA. 
It has created a new tab header called `SPECIFIC` which can contain them without forcing them to the bottom.
I attempted to use PANEL before hand to put it in the DETAIL section but it has weird rendering rules and was either putting them at the VERY BOTTOM of the webpage, unreadable or in the mana pool graph (combining the two infographics). 
The idea is just for SPECIFIC things to the spec that will never be able to be shared but also not general enough to the player to put near the top. 
I'm 100% open to re-naming it.

If you do look at the fallen order stuff realize that NONE of those spells have actual cast events so you gotta cheese it by looking at buff applied/refreshed events


analysis/monkmistweaver/src/modules/shadowlands/covenant/CloneAbilityAverage.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { SpellIcon, SpellLink } from 'interface';
+import Analyzer, { Options, SELECTED_PLAYER, SELECTED_PLAYER_PET } from 'parser/core/Analyzer';
+import Events, { ApplyBuffEvent, RefreshBuffEvent, SummonEvent } from 'parser/core/Events';
+import BoringValueText from 'parser/ui/BoringValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import React from 'react';
+
+class CloneAbilityAverage extends Analyzer {
```
Pewtro: I would probably call this something else like FallenOrderCraneAverage to let a poor future developer know what this is about <3

Pewtro (commented):
LGTM once CloneAbilityAverage is namechanged, not sure I love SPECIFIC as a section name in the statistics, but I can't think of something much better for theorycrafty stuff

analysis/monkmistweaver/src/modules/shadowlands/covenant/CloneAbilityAverage.tsx
emallson: agreed here

emallson:
> LGTM once CloneAbilityAverage is namechanged, not sure I love SPECIFIC as a section name in the statistics, but I can't think of something much better for theorycrafty stuff

what if we just called it `THEORYCRAFT`?

Pewtro:
seems good to me

PR #4655, opened 2021-11-05
Verify patch 9.1.5 compatibility and bump support

darkpsy3934:
After looking at patch notes and confirming no changes have been done since 9.0.5, I have determined the analysis is correct for 9.1.5.  Bumped patch support to 9.1.5, added to CONTRIBUTORS, and added changelog entry.

### DIFF
--- a/analysis/deathknightfrost/src/CHANGELOG.tsx
+++ b/analysis/deathknightfrost/src/CHANGELOG.tsx
@@ -6,6 +6,7 @@ import React from 'react';
 
 
 export default [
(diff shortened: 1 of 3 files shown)


abelito75:
Are we sure rotation/covenant choices haven't changed over that time frame?

darkpsy3934:
To my knowledge, they have remained the same.

HawkCorrigan:
You're gonna need to import your "name" in the changelog as well.

emallson (approved):
AFAIK from people in guild the only real change from 9.1.5 is the ability to play other covenants where it makes sense, so this seems fine.

PR #4656, opened 2021-11-05
Filled out/added some stats for tbc resto shaman.

Arbixal:
- Filled out ability info for various resto shaman spells

- Added Mad Alchemist Potion as a valid Combat Potion

- Added Earth Shield and Water Shield Statistic and Recommendations (based on the retail resto shaman one)

image
image

- Added Chain Heal statistic and recommendations (based on the retail resto shaman one)

image
image
image
image

- Added Mana Tide totem statistic.

image
image

- Whitelisted some lower rank spells that are common-place for resto shaman use.


analysis/tbcshaman/src/CONFIG.tsx
```diff
@@ -35,7 +35,7 @@ const config: Config = {
   builds: {
     [Build.DEFAULT]: {
       url: 'standard',
-      name: '20/41/0',
+      name: '0/5/56',
```
emallson, line 38: that is a pretty large change in the default talent build. was there a meta shift or was this a typo, etc?

analysis/tbcshaman/src/modules/spells/ChainHeal.tsx
```diff
@@ -0,0 +1,264 @@
...
+      timestamp: currentCast.timestamp,
+      castNo: this.castIndex,
+      hits: this.buffer.filter((event) => event.type === EventType.Heal).length,
+    };
+    this.buffer = [];
+  }
+
+  suggestions(when: When) {
+    const suggestedThreshold = this.suggestionThreshold;
+    if (isNaN(suggestedThreshold.actual)) {
```
emallson: instead of using `isNaN` here, you should check the condition that you actually care about (which I think is `this.casts === 0`?)

emallson (changes requested):
1 change (the `isNaN` thing) and one clarification (whats up with the talent build change?)

it doesn't look like changing the build name is a functional change, only a label change, so I'm wondering if that was just a copy/paste error when Khadaj set this up.

analysis/tbcshaman/src/CONFIG.tsx
Arbixal: I can only assume a copy-pasta. There isn't really a known/common 20/41/0 build for shaman at all.
image
Priest however does have a 20/41/0 build.

emallson: :+1:

emallson (approved):

PR #4657, opened 2021-11-05
Allow SpellHistory to have prepull events in it

Pewtro:
On a report like [this](https://wowanalyzer.com/report/Mwf9KAbVP8JNhdHG/26-Mythic+Guardian+of+the+First+Ones+-+Kill+(4:25)/Schmub%C3%A4rchi/standard/statistics). 
image

We end up in a scenario where Double Tap is correctly registered as precast. 
This means that the begincooldown event has been triggered correctly. That's why the cast efficiency bar graph correctly can show 98% since it uses timeOnCooldown/timeInCombat (+some GCD tom foolery). 
However as you can see the casts still show as 4 out of 5 possible. The maxCasts looks at fightDuration/spellCooldown which means a 4m25s fight lets you have 5 1 minute cooldown casts off. The casts however looks at spellhistory in order to get the casts amount which currently do NOT allow for prepull events. 
This mean we end up looking at 4 casts done out of possible 5, even though the player DID get the possible 5 casts. 
The metrics are comparing completely different things, timeOnCooldown accounts for us triggering the cooldown during prepull which fixes cast efficiency graph, but then we don't care about what happens in prepull for the casts comparison which doesn't make much sense. 

This fix allows prepull spell events to go into the spellhistory, as it seems like there is no reason they shouldn't be in there. 

As you can see from the test snapshots this also fixes issues with pot usage precombat, and by extension also any other precombat spell issue where that cast event wasn't counted as a cast. 

An alternate fix is to subtract 1 from maxCasts if precast, but to me that would display the wrong information. In this case it is correct that there were 5 possible casts, and not 4 as we'd show in that case. 

The above report ends up looking like this with this fix: 
image

On Live we correctly show the spells in the timeline: 
image
We also include it in our (fabricated casts enabled) view:
image

Nothing is changed in that regard with this fix:
image
image

Edit: I know it looks like there's a double cooldown event for that Double Tap, will dig into that but it's unrelated to this issue as I could showcase the same issue with any other prepull spell.

### DIFF
--- a/analysis/hunterbeastmastery/src/integrationTests/beastMasteryIntegrationTests.test.ts.snap
+++ b/analysis/hunterbeastmastery/src/integrationTests/beastMasteryIntegrationTests.test.ts.snap
@@ -13221,7 +13221,7 @@ exports[`Beast Mastery Hunter integration test: Single Target CastEfficiency mat
                 }
               }
             >
-              0.00
+              0.19
             </td>
             <td
               className="text-right"
@@ -13231,7 +13231,7 @@ exports[`Beast Mastery Hunter integration test: Single Target CastEfficiency mat
                 }
               }
             >
-              0
+              1
               /1
                casts
             </td>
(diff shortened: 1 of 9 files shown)


abelito75:
Please don't make a PR unless you have a valid description. If there isn't one by the end of the decade I'm going to unmerge this PR and delete your github account

emallson (approved):

PR #4659, opened 2021-11-06
Add present/missing debuff, fractional spell charges and remaining cooldown time as APL conditions 

Pewtro:
Adds some new conditions to play with
- debuffPresent - follows same logic as buffPresent but needed some additional control to handle multiple targets being capable of having debuffs
- debuffMissing - follows same logic as buffMissing but needed some additional control to handle multiple targets being capable of having debuffs
- spellFractionalCharges - lets you define that a given spell must have between atLeast and atMost fractional spellcharges (e.g. 1.4 charges of Barbed Shot) 
- spellCooldownRemaining - lets you define that a given spell must have between atLeast and atMost milliseconds remaining of cooldown

Added a new helper called formatTimestampRange to extend the formatRange to also allow for timestamp formatting, which is initially used in spellCooldownRemaining.

I moved around a few things as they were being reused - if the new file placements don't make sense, let me know and I'll happily move em around again. 

I implemented both spellCooldownRemaining and spellFractionalCharges in the BM APL checker - originally I wanted to tuse the debuffPresent and debuffMissing ones but I forgot that tar trap is stupid and doesn't work on bosses so it was wildly inaccurate, and I didn't want to hack some "doesn't apply to bosses" when I could just write the apl condition differently.


src/parser/shared/metrics/apl/conditions/not.tsx
```diff
@@ -2,11 +2,15 @@ import React from 'react';
...
-export default function not(cnd: Condition<any>): Condition<any> {
+export default function not(cnd: Condition<any>, showDont: boolean = true): Condition<any> {
   return {
     ...cnd,
     key: `not-${cnd.key}`,
     validate: (state, event, spell) => !cnd.validate(state, event, spell),
-    describe: (tense) => <>don't {cnd.describe(tense)}</>,
+    describe: (tense) => (
+      <>
+        {showDont && "don't"} {cnd.describe(tense)}
```
emallson, line 12: per our discussion in discord i'm probably gonna mess around with alternate ways of doing this, but we can merge this PR as-is and I can refactor later if i find something i really like.

emallson (approved):
looks like you need to update snapshots but otherwise :+1:

PR #4665, opened 2021-11-08
Clean up cache actions

MartijnHols:
I used custom cache actions to improve performance beyond what is possible with the standard actions/cache action from GitHub. I made PRs to the original repo, but they has been ignored for up to a year so I closed them and started my own fork in which I cleaned everything up. This moves the temporary messy actions to a more permanent solution.

In case you want to know more about my caching solution, I wrote docs for a change: https://github.com/MartijnHols/actions-cache


emallson:
This looks good to me. You need a changelog entry though :P

MartijnHols:
I thought we could merge regardless of the changelog check passing :o

emallson:
nope, changelog entry is a required check

emallson (approved):

PR #4669, opened 2021-11-11
[Resto druid] Removed healer stat weights

abelito75:
Removing healer stat weights in favor of QElive


emallson:
doesn't even build smdh @abelito75

abelito75:
> doesn't even build smdh @abelito75

I know. I was doing this when work called... Also one random generic conduit is made druid specific zzz

emallson:
ngl unless that conduit actually changes gameplay you can just remove it as well

abelito75:
I talked to voulk. He said keep so I keep
Also I did another PR for this and it's already in

emallson:
Fair enough

PR #4670, opened 2021-11-11
[Hunter] Update Rylakstalkers Piercing Fangs

Pewtro:
A simple bugfix to correctly attribute damage to this legendary. since beast cleave can't crit but the increased damage makes beast cleave do more damage hence the increased damage from the crit should be attributed to this legendary.


abelito75:
Looks good to me

PR #4685, opened 2021-11-14
Rewrite of channel event fabrication, and added helper class for inserting fabricated events during a Normalizer's processing

kfinch:
This PR just adds the helper, no uses yet. Sharrq expressed interest in using a Normalizer to insert events, so I expect that to be the first usage. In the future, I believe it's a more durable design to insert fabricated events in a Normalizer instead of an Analyzer, so I expect this to get further usage.


src/parser/core/insertEvents.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @return the analysis events with the inserted events added in order
+ */
+export function insertEvents(events: AnyEvent[], insertEvents: AnyEvent[]): AnyEvent[] {
+  // sort the insertEvents by timestamp and mark them fabricated
+  insertEvents.forEach((e) => (e.__fabricated = true));
+  insertEvents.sort((a, b) => a.timestamp - b.timestamp);
+  // iterate through analysis events, inserting insertEvents as we go into new fixedEvents list
+  let currInsertIndex = 0;
+  const fixedEvents: AnyEvent[] = [];
+  events.forEach((event: AnyEvent) => {
```
emallson: i'm curious whether this is faster than throwing them at the end and running `sortBy` on the timestamps.

kfinch: I think it's faster as long as `insertEvents` is smaller than `events` (a reasonable assumption). After `insertEvents` gets sorted, this is just the merging of two sorted lists, which is O(n).

Just using sortBy would be simpler, that's for sure. My only worry is we'd have to make sure the sort is stable, because some analyzers rely on ordering of events with the same timestamp.

kfinch:
Ok this properly ready for review.

The main thing here is a complete rewrite of how Channeling is handled, which can be seen in the new unified Channeling class. The vast majority of the file changes here are removal of the old channeling classes (which were per-spec), and updating of all the imports.

This PR also includes a new 'InsertableEventsWrapper' class which is designed to help Normalizers fabricate events more easily.

kfinch:
Known issue: the 'active time' calculation for channeling still isn't quite right, it's often incrementing active time with both a GCD and the Channel, even though they overlap. On balance, things are more accurate, but I'd like to get around to fixing this. I just want to get this PR reviewed and in because it's been going on too long.

src/parser/core/Events.ts
```diff
@@ -163,9 +163,17 @@ export type AnyEvent<
 > = MappedEventTypes[ET];
 
 export interface Ability {
+  /** The ability's name */
   name: string;
+  /**
+   * The ability's ID - this uniquely identifies the ability,
```
abelito75, line 169: This is literally the most minor thing event but if we are adding comments we should probably mention that
A single spell can have multiple GUIDs
A single GUID can handle multiple parts of a spell
And that this alone might not be all the information you need about a spell

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+ * Channelled spells are handled inconsistently in the events, but this normalizer aims to uniformly
+ * represent them with a BeginChannel when started and and EndChannel when stopped (regardless of if
+ * the spell finished or it was stopped early). There are a variety of ways channeled spells show
+ * in events, and this normalizer allows special case handling to be registered for each.
+ */
+class Channeling extends EventsNormalizer {
+  /**
+   * Listing of all special case handlers for channels
+   */
+  static CHANNEL_SPECS: ChannelSpec[] = [
```
abelito75, line 40: I wonder if we could have this defined in the spec's analysis area ie (analysis\mnokmistweaver\normalizers\MWChanneling.ts)

I just have never been a fan of big master lists so its w/e either way

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+    // Priest
+    buffChannelSpec(SPELLS.DIVINE_HYMN_CAST.id),
+    nextCastChannelSpec(SPELLS.PENANCE_CAST.id),
+    buffChannelSpec(SPELLS.VOID_TORRENT_TALENT.id),
+    buffChannelSpec(SPELLS.MIND_FLAY.id), // TODO double check ID
+    buffChannelSpec(SPELLS.MIND_SEAR.id), // TODO double check ID
+    // Rogue
+    // Druid
+    buffChannelSpec(SPELLS.CONVOKE_SPIRITS.id),
+    // Monk
```
abelito75, line 59: Does this work with pets too? 
Invoke Yu'lon, the Jade Serpent currently does all of **HER** healing though Soothing Breath which is a 4 second channel
Fallen Order currently channels Soothing Mist OR casts Enveloping Mist

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    const channelState: ChannelState = {
+      unresolvedChannel: null,
+      eventsInserter,
+    };
+
+    // for each event, check if there is a special case handler for that GUID -
+    // if so - call the handler, if not - call the default handling
+    events.forEach((event: AnyEvent, index: number) => {
+      if (!HasSource(event) || event.sourceID !== this.selectedCombatant.id) {
```
abelito75, line 116: If this does plan to work with pets you will need to factor in another field
I believe we don't currently map it but it should be sourceInstance
which would be which instance of the pet it is. This would/could be something major for Venthyr Mistweaver and Demo lock 

You can actually see this field being used in other locations in the codebase...mainly where summoning/tracking pets are concerned

abelito75 (commented):
This isn't a real review based on the actual coding practices but rather a slew of questions

src/parser/core/Events.ts
kfinch: Yeah I can add to it. I added these comments just because I was trying to figure out what `ability.type` was and so after hunting down what it was I decided to comment the whole object.

src/parser/shared/normalizers/Channeling.ts
kfinch: I definitely was considering it, and could change it to be that way. I mostly have it in master list form because it was easier to mock up that way. I plan to at least make `CHANNEL_SPECS` passed in instead of static so it's extensible.

src/parser/shared/normalizers/Channeling.ts
kfinch: As currently set up it does not, but I think it would be possible to make it handle pets if I extended `ChannelState` to track per entity. It's generally important to filter by source because if you don't, you get screwed up by Cast events that target the player.

Is there value and supporting pets though? The primary purpose of the channel events is to support ActiveTime calculations, and I thought that wasn't really relevant for pets - I could be mistaken though as I've never played a pet class.

src/parser/shared/normalizers/Channeling.ts
kfinch: Yeah if I can be convinced it's valuable to support pets, I can see where I would extend this.

emallson: i'm not aware with any pets with channels except succubus CC, but that doesn't impact downtime calculations so i think its fine.

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+    events: AnyEvent[],
+    eventIndex: number,
+    channelState: ChannelState,
+  ): void {
+    if (event.type === EventType.BeginCast) {
+      // we're starting a new generic hardcast
+      cancelCurrentChannel(event, channelState);
+      beginCurrentChannel(event, channelState);
+    } else if (event.type === EventType.Cast) {
+      // TODO this won't catch pre-cast channels
```
emallson, line 144: is it even possible to precast a channel? like, won't the first application initiate combat? i guess its a question of whether that hit gets logged which...maybe?

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+  defaultHandler(
+    event: AnyEvent,
+    events: AnyEvent[],
+    eventIndex: number,
+    channelState: ChannelState,
+  ): void {
+    if (event.type === EventType.BeginCast) {
+      // we're starting a new generic hardcast
+      cancelCurrentChannel(event, channelState);
+      beginCurrentChannel(event, channelState);
```
emallson, line 142: if the `defaultHandler` gets triggered, will this not insert a begincast for non-channeled casts?

src/parser/shared/normalizers/Channeling.ts
```diff
@@ -0,0 +1,344 @@
...
+    // if so - call the handler, if not - call the default handling
+    events.forEach((event: AnyEvent, index: number) => {
+      if (!HasSource(event) || event.sourceID !== this.selectedCombatant.id) {
+        return; // only concerned with events coming from selected player
+      }
+      const handler = HasAbility(event) ? this.channelSpecMap[event.ability.guid] : undefined;
+      if (handler) {
+        handler(event, events, index, channelState);
+      } else {
+        this.defaultHandler(event, events, index, channelState);
```
emallson, line 123: is this going to trigger the default channel handler for non-channeling events?

emallson (commented):
couple questions about the `defaultHandler` possibly triggering on non-channel events. i will actually run this PR in the morning but if you happen to have a ready answer before then that works too

src/parser/shared/normalizers/Channeling.ts
abelito75: There are some pets that have channels like Crane clones for all monks, yu'lon the jade serpent mistweaver
And i'd imagine there are some for demo and maybe water elemental
They all won't affect downtime calculation but it could yield some useful information about the pet AI

abelito75: Yes it is possible to precast a channel and more logs exist than you would assume with it. 
Mistweavers with essence font and marksmen hunters with aimed shot

src/parser/shared/normalizers/Channeling.ts
kfinch: I'm gonna say this should go live without pet support, and it can be added later if needed.

src/parser/shared/normalizers/Channeling.ts
kfinch: This is in the defaultHandler, the comment is really referring to hardcasts. Basically, if a player starts a cast before the pull but finishes it after, we'll see the cast event but not the begincast, and so this handler will believe it's an instant and handle it accordingly.

This will cause only small inaccuracy, and the handling (and comment) are carryovers from the previous handler, so my new code is at least not any worse.

src/parser/shared/normalizers/Channeling.ts
kfinch: I'm not sure what you're asking here. The purpose of this module is that both regular hardcasts and channels are shown in the same way by the fabricated `beginchannel` and `endchannel` events. Regular hardcasts always start with a `begincast` event, and so this logic puts a `beginchannel` in the same place. This is intended behavior, and it's the same thing the module currently does on live.

src/parser/shared/normalizers/Channeling.ts
kfinch: Yes. Judging by this comment and the one above, I think you have a misconception about what the `Channeling` module does (both in this rewrite and on live). Its purpose is to fabricate events that handle regular cast time spells and channeled spells in the same way, making it easier for users of these events to display when the player is casting and when he isn't.

It is the design intent of this module that regular cast time spells will have a `beginchannel` and `endchannel` event associated with them - this is laid out in the first paragraph of the class docs.

Cast time spells are pretty consistent in the original events out of WCL - a `begincast` at the start and a `cast` when it finishes - the purpose of the `defaultHandler` function is to handle these consistent baseline cases. Channeled spells are both different from cast time spells and also not internally consistent, hence the registered handlers for each of them.

emallson: Ah, I'd missed that in the module description. While I think that having a consistent "actively casting/channeling" event marker is valuable, this currently breaks the timeline (non-channels are shown as channels, see below). 

Screenshot from 2021-12-14 11-27-48

Going to re-review the ABC changes and see if they require the change to channeling events. If so, I see a couple of options:

1. Add a prop to the channel events to indicate default vs non-default handler (not a fan of this).
2. Fabricate an "activitystart" / "activityend" event pair that is used for ABC instead of repurposing the "beginchannel" / "endchannel" pair.

If ABC doesn't require the change to casts having channeling events then I think we just don't emit channeling events for non-channels for the time being.

emallson: Interestingly, I think the ABC code was broken for non-channel casts that were longer than the GCD prior to this. The GCD code doesn't appear to handle cast times (which would be a bad place to put it, but I needed to check) and the ABC code doesn't handle casts (only GCD and channel). This resulted in it being subtly wrong since few casts are longer than a GCD.

Another alternative is removing the CSS for channels. I don't know that they really need to be demarcated from casts, and the timeline suffers from high levels of visual noise that this could help with.

emallson: Screenshot from 2021-12-14 12-26-04
actually prod looks like this right now. not sure why? fucking weird.

i think we just remove the channeling css and call it good. everything else seems fine.

kfinch: This is what I've been trying to say! I duplicated the behavior that's currently on prod. On prod, beginchannel and endchannel events are fabricated for regular hard casts, and those events are used by ABC to get active time right.  This is not a bug, this is intended behavior.

emallson: okay i'm a dumdum. gonna merge this and we'll do the channeling timeline change later because wow that looks weird

PR #4688, opened 2021-11-18
Fixed Rising Sun Kick usage for Mistweaver monks

trevorm4:
Fixed a bug in RisingSunKick.tsx where it was not properly restarting the cooldown when a Rising Sun Kick was cast because of a blackout kick reset.

Also added myself to Contributors


abelito75:
Good shit.

Issue #4666, opened 2021-11-09
[RFC] Automatically detect the closest build


Arbixal:
### Background
**Note:** I'm coming in to this code-base pretty fresh having not gone through previous iterations, so this is based on my current understanding.

Current functionality allows for multiple builds to be defined for a config, but it will always select a given default unless specified in the url. This causes confusion with users that don't realize that either they need to manually change builds by clicking on the relevant icon, or that their spec is not yet supported.

This functionality seems to apply to the Burning Crusade classes/specs as the previously implemented class/spec analysis for Retail appear to be broken down into one class/spec per analysis.

IF multiple builds exist in a given analysis it would be ideal if the closest build to the selected player's build would be automatically selected.

#### WCL Spec
A couple of options are available for selecting the correct build, depending on how tightly coupled we are (or want to be) to WCL given specs.

WCL provide a number of specs per class and perform ranking calculations based on those different specs. Upgrading to the V2 api for WCL would give access to that information and allow for selection of build based on a given spec. It is probably worth noting that it is not just limited to 3 specs per class.

#### Match talent description
If we don't want to line up directly with whichever spec WCL has classified a player as, the only other option I can see is to compare talent descriptions. WCL provide each players talent selections as the standard "0/5/56" notation. Our builds also make use of this notation for the build name (though not enforced?)

One way of finding the "closest" match is sum the difference of the spec breakdown. E.g.
```
"0/5/56" vs "0/9/52" 
= |0-0| + |5 - 9| + |56-52| 
= 0 + 4 + 4 
= 8
```
If the number is substantially different then you know it isn't a good match.

For example comparing shaman specs:
![image](https://user-images.githubusercontent.com/1750811/140909755-c6210666-ebad-40c2-847b-b887a116a04c.png)

#### Another option?
Are there other options I haven't thought about?

PR #4689, opened 2021-11-23
Automatic build selection

Arbixal:
First cut of automatically selecting the closest build (mostly for TBC) based on my suggestions in #4666.
If this isn't the best approach, I'm open to changes.

Comparison is done for all the builds in the relevant config and the talents given by WCL. The builds are sorted in order of closest match and then the first one is selected. 

If a build differs too much (50 talents points) then it is considered not a close enough match. This essentially means 25 points in a different tree as it will differ by 25 points in both trees which adds together to make 50 points of difference.

I also had to adjust the points given to some other builds (druid, paladin, rogue warlock) as they seemed to be copy-pasta.

I'm also thinking of splitting out the talent points from the name of the build as there is nothing forcing the name to adhere to "X/Y/Z" format that I am assuming.

### DIFF
--- a/analysis/tbcdruid/src/CONFIG.tsx
+++ b/analysis/tbcdruid/src/CONFIG.tsx
@@ -37,7 +37,7 @@ const config: Config = {
   builds: {
     [Build.DEFAULT]: {
       url: 'standard',
-      name: '20/41/0',
+      name: '0/0/61',
       icon: <Icon icon="spell_nature_healingtouch" />,
       visible: true,
     },
--- a/analysis/tbcdruid/src/CONFIG.tsx
+++ b/analysis/tbcdruid/src/CONFIG.tsx
@@ -38,6 +38,7 @@ const config: Config = {
     [Build.DEFAULT]: {
       url: 'standard',
       name: '0/0/61',
+      talents: [0, 0, 61],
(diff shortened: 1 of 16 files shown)


emallson:
> I'm also thinking of splitting out the talent points from the name of the build as there is nothing forcing the name to adhere to "X/Y/Z" format that I am assuming.

this would be good

overall, I think this looks like a good first pass at it and we'll see how it ends up working in practice.

I do think that we should seriously investigate just using the WCL builds, but I'm not sure if there are multiple "builds" within the same WCL "spec" (e.g. two different resto druid builds in TBC that could both be called resto but need different analysis) that would break relying solely on that.

Arbixal:
> I do think that we should seriously investigate just using the WCL builds, but I'm not sure if there are multiple "builds" within the same WCL "spec" (e.g. two different resto druid builds in TBC that could both be called resto but need different analysis) that would break relying solely on that.

Yeah. There isn't currently a need for it perhaps. Although, I can already see that there is a fury warrior and a deathwish fury warrior build. Both would count as fury but one has an extra cooldown that might want to be tracked. I feel like there might be some changes that could change the analysis slightly, but is it enough to justify this approach?

PR #4690, opened 2021-11-23
[Mistweaver] added checking for essence font cancellation

trevorm4:
Added a suggestion to MW analysis to make sure the player doesn't ever cancel EF early with a 100ms buffer

### DIFF
--- a/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
+++ b/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
@@ -62,22 +62,10 @@ class EssenceFont extends Analyzer {
     );
     this.addEventListener(Events.EndChannel.by(SELECTED_PLAYER), this.handleEndChannel);
     this.upwelling = Number(this.selectedCombatant.hasTalent(SPELLS.UPWELLING_TALENT.id));
-    this.spells = this.spellGuids;
   }
 
   get efHotHealing() {
     return this.efHotHeal;
-  }
-
-  get spellGuids() {
-    const result: any = {};
-    for (const value of Object.values(baseSpells)) {
-      result[value['id']] = value['name'];
-    }
-    for (const value of Object.values(talents)) {
-      result[value['id']] = value['name'];
-    }
-    return result;
   }
 
   get efHotOverhealing() {
--- a/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
+++ b/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
@@ -6,7 +6,13 @@ import talents from 'common/SPELLS/talents/monk';
 import { SpellLink } from 'interface';
 import { SpellIcon } from 'interface';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { ApplyBuffEvent, CastEvent, HealEvent, RefreshBuffEvent } from 'parser/core/Events';
+import Events, {
+  ApplyBuffEvent,
+  CastEvent,
+  HealEvent,
+  RefreshBuffEvent,
+  EndChannelEvent,
+} from 'parser/core/Events';
 import { ThresholdStyle, When } from 'parser/core/ParseResults';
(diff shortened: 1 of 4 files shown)


abelito75:
This doesn't handle well If they have upwelling and it should be added to the checklist

trevorm4:
Fixed upwelling handling

trevorm4:
Added EF to checklist, currently with greater than 0 until I do another PR for greater than fix in `performanceForThresholds.ts`

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -76,7 +93,25 @@ class EssenceFont extends Analyzer {
...
+        major: 0,
+      },
+      style: ThresholdStyle.NUMBER,
+    };
+  }
   castEssenceFont(event: CastEvent) {
+    let extra_secs = 0;
+    if (this.hasUpwelling) {
+      extra_secs = Math.max(
+        (event.timestamp - (this.last_ef + SPELLS.ESSENCE_FONT.cooldown * 1000)) / 6000,
```
abelito75: Shouldn't this be math.min?
Since if they hit ON CD wouldn't think code just auto pick 3?

abelito75: Does the spell files actually have cooldown?
This is referring to 
common/spell/monk.ts

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -46,6 +61,8 @@ class EssenceFont extends Analyzer {
       Events.refreshbuff.by(SELECTED_PLAYER).spell(SPELLS.ESSENCE_FONT_BUFF),
       this.refreshEssenceFontBuff,
     );
+    this.addEventListener(Events.EndChannel.by(SELECTED_PLAYER), this.handleEndChannel);
```
abelito75: You should be able to do `.spell(SPELLS.ESSENCE_FONT)` after .by so filter for just EF end cast events

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -109,6 +144,23 @@ class EssenceFont extends Analyzer {
     this.targetOverlap += 1;
   }
 
+  handleEndChannel(event: EndChannelEvent) {
+    if (
+      event.ability.guid === SPELLS.ESSENCE_FONT.id &&
```
abelito75: You shouldn't do this check. Let the event filter handle it

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -109,6 +144,23 @@ class EssenceFont extends Analyzer {
...
+      event.duration < this.expected_duration - this.cancelDelta
+    ) {
+      console.log(
```
abelito75: Console should only print if it's a major system breaking issue
You can add a debug boolean set to false if you want the next developer to have a debug toggle if you want BUT console is holy when it goes to prod

abelito75 (changes requested):
Just some things

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
trevorm4: Yeah it should be

trevorm4: Ah no it doesn't, think I confused Abilities.tsx in mw folder with the other monk file

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
trevorm4: Removed

emallson:
looks like changes were addressed. merging

abelito75:
Dammit emallson. I was gonna double check

PR #4692, opened 2021-11-26
Implement greaterThanOrEqual and lessThanOrEqual for checklists

trevorm4:
Added `performanceForGreaterThanOrEqualThresholds` and `performanceForLessThanOrEqualThresholds` to `performanceForThresholds.ts`, allowing for >= and <= to be used in checklists.

Also, modified `EssenceFont.tsx` in MW analysis module to change the threshold for cancellations (as discussed in previous PR)

### DIFF
--- a/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
+++ b/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
@@ -61,7 +61,10 @@ class EssenceFont extends Analyzer {
       Events.refreshbuff.by(SELECTED_PLAYER).spell(SPELLS.ESSENCE_FONT_BUFF),
       this.refreshEssenceFontBuff,
     );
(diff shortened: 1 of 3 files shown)


analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -99,12 +99,13 @@ class EssenceFont extends Analyzer {
   get suggestionThresholdsCancel() {
     return {
       actual: this.cancelled_ef,
-      isGreaterThan: {
-        major: 0,
+      isGreaterThanOrEqual: {
+        major: 1,
```
emallson, line 103: these are a equivalent. what led to the change?

emallson (approved):
going to approve since it ultimately seems like it'll close an awkward inconsistency between systems, but I'm a little confused why so much energy was spent on this to do something that is equivalent to what we had already. does this fix a tooltip or something?

PR #4694, opened 2021-11-26
Fix compatibility with NodeJS 16

jander99:
This PR fixes a couple of issues when using NodeJS 16 LTS

- Fix compatibility issue with `node-sass` and Node by bumping version to one know to work with NodeJS 16
- NodeJS 15 changed the way unhandled rejections work. Instead of being warned, they are now thrown. A Promise in `src/localization/I18nProvider.tsx` did not have a corresponding `.catch()` so now we log the error if it occurs.

### DIFF
--- a/src/localization/I18nProvider.tsx
+++ b/src/localization/I18nProvider.tsx
@@ -47,7 +47,6 @@ const I18nProvider = ({ children }: Props) => {
       })
       .catch((error) => {
         console.error('Unable to set locale', error);
-        return;
       });
   }, [locale, activeLocale, setActiveLocale]);
 
--- a/src/localization/I18nProvider.tsx
+++ b/src/localization/I18nProvider.tsx
@@ -39,11 +39,16 @@ const I18nProvider = ({ children }: Props) => {
       return;
     }
 
-    loadCatalog(locale).then(({ messages }) => {
-      i18n.load(locale, messages);
-      i18n.activate(locale);
-      setActiveLocale(locale);
-    });
+    loadCatalog(locale)
+      .then(({ messages }) => {
+        i18n.load(locale, messages);
+        i18n.activate(locale);
+        setActiveLocale(locale);
+      })
+      .catch((error) => {
+        console.error('Unable to set locale', error);
+        return;
+      });
   }, [locale, activeLocale, setActiveLocale]);
 
   if (!activeLocale && process.env.NODE_ENV !== 'test') {
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,0 @@
-trailingComma: "all"
-semi: true
-singleQuote: true
-printWidth: 100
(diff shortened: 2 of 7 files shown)


jander99:
In order to fix a breaking change to `node-sass`, `yarn.lock` needed to be regnerated. It seems that minor version upgrades to `eslint` have introduced additional checks that now produce warnings, specifically around the use of "for..in" loops. 

> for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array or use for..of

I'm confident most of the warnings can be remedied with a little bit of refactoring (i.e. following [this pattern](https://www.30secondsofcode.org/articles/s/eslint-refactor-for-in)) however those changes should be in a different pull request. I will investigate which devDependency needs to have its version downgraded in order to make the linting step pass with the current code. Once this PR is accepted, we can version-bump and fix up the other `eslint` (and `prettier`) recommendations.

jander99:
This PR also seems to address [this card](https://github.com/WoWAnalyzer/WoWAnalyzer/projects/16#card-72456689) in the Internals project.

src/localization/I18nProvider.tsx
```diff
@@ -39,11 +39,16 @@ const I18nProvider = ({ children }: Props) => {
...
-      i18n.activate(locale);
-      setActiveLocale(locale);
-    });
+    loadCatalog(locale)
+      .then(({ messages }) => {
+        i18n.load(locale, messages);
+        i18n.activate(locale);
+        setActiveLocale(locale);
+      })
+      .catch((error) => {
```
MartijnHols, line 48: This is currently only executed on the client side, do the rejection handling of node should not be relevant making this catch unnecessary. I think we can maintain the old behavior.

MartijnHols (changes requested):
Thanks for the PR! This seems like a good step towards. Eventually we’d like to remove node-sass altogether and maybe in the mean time switch todart-sass, but for now node16 compat is a welcome addition. 

Just one thing;

src/localization/I18nProvider.tsx
emallson: I believe that this comes up in tests. Going to pull the branch and check but currently when running tests there are (sometimes many) uncaught i18n-related throws.

emallson: this does affect tests (positively---a test that had been consistently failing in my dev environment works as a result of this change)

jander99: I would need to figure out how to mock enough objects to where the promise would succeed. For now, I wanted to get the render test in `src/Root.test.tsx` to pass. In my research of the error message generated when upgrading from node 14.x to node 16.x I found [this article](https://betterprogramming.pub/whats-new-in-node-js-15-fc24e87e2590) which suggested that the affected change in warning threshold was added in node 15. 

In reviewing this file, I did notice that the extra return statement inside the catch is not needed as the other statements successfully cover the branching with a return statement that either creates the component or returns null anyway.

emallson (approved):

PR #4695, opened 2021-11-27
Bump React + TypeScript + Babel versions

emallson:
The typescript version bump exposed a number of places where we had potential type issues, which have been fixed as part of this.

I also fixed the `for ... in` issue that was noted by @jander99


emallson:
@Pewtro would like it if you can pull this branch and just confirm that everything works on your machine after doing a `yarn install`.

this bumps versions on a bunch of stuff and i am mostly concerned with making sure people's dev environments continue working without issue

Pewtro:
I would maybe change our eslintrc file to include the extends mentioned here https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md#when-not-to-use-it rather than just turning off the rule. 

Probably same end-result but seems like that's recommended way.

Pewtro (approved):
Compiles fine on both my windows and mac on different node versions. 
Seems solid - loaded up a few random logs on both without issues

PR #4697, opened 2021-11-30
Added EF cancellation to timeline

trevorm4:
Added EF cancellation to timeline, so that you can see which EF you cancelled in specific

Courtesy of Abe

### DIFF
--- a/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
+++ b/analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
@@ -37,7 +37,8 @@ class EssenceFont extends Analyzer {
   cancelled_ef: number = 0;
   hasUpwelling: boolean = false;
   cancelDelta: number = 100;
-  last_ef: number = 0;
+  last_ef_time: number = 0;
+  last_ef: any = null;
   protected haste!: Haste;
   constructor(options: Options) {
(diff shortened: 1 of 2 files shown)


abelito75:
Thanks for making this change for me. I was a bit lazy to boot up the server and make a new branch since mine is currently flooded with fallen order memes

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
```diff
@@ -37,7 +37,8 @@ class EssenceFont extends Analyzer {
   cancelled_ef: number = 0;
   hasUpwelling: boolean = false;
   cancelDelta: number = 100;
-  last_ef: number = 0;
+  last_ef_time: number = 0;
+  last_ef: any = null;
```
abelito75: BAD
NEVER DO ANY. @Pewtro WILL MURDER YOU IF YOU DO
You should do CastEvent | null

abelito75 (changes requested):
ANY BAD

analysis/monkmistweaver/src/modules/spells/EssenceFont.tsx
trevorm4: Fixed

PR #4701, opened 2021-12-05
[Windwalker Monk] T28 4-Set

Hurstilol:
Its a bit early but ¯\_(ツ)_/¯


abelito75 (changes requested):
I'm too lazy to review this right now in depth but i do want changes so don't merge

emallson:
> I'm too lazy to review this right now in depth but i do want changes so don't merge

PR #4711, opened 2021-12-11
[Monk] Module for NF Legendary

Hurstilol:
Small module that calculates the damage gain from the 8% debuff applied to enemies through the nightfae legendary 'Faeline Harmony' in the statistics


abelito75:
I'll look at this in a bit

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import React from 'react';
+import Enemies from 'parser/shared/modules/Enemies';
+
+
+
+const MOD = 0.08;
```
Pewtro: Please give consts a more descriptive name to indicate what this MOD does for easier maintainability for someone else in the future

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+ 
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE()}
+        size="flexible"
+		category={STATISTIC_CATEGORY.COVENANTS}
+        tooltip={
+          <>
+            Total damage done by the 8% increase: {formatNumber(this.totalDamage)}
```
Pewtro: This could be a bit more descriptive to indicate where the 8% comes from (a spell link?) and you can tie the number to the const above to make it less bug prone in the future `formatPercentage(MOD)`

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+      return;
+    }
+    const target = this.enemies.getEntity(event);
+    if (target !== null && target.hasBuff(SPELLS.FAELINE_HARMONY_DEBUFF.id, event.timestamp)) {
+      this.totalDamage += calculateEffectiveDamage(event, MOD);
+    }
+  }  
+  get dps() {
+    return (this.totalDamage / this.owner.fightDuration) * 1000;
+  }
```
Pewtro: This function is unnecessary if you follow the below suggestion

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+
+ 
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE()}
+        size="flexible"
+		category={STATISTIC_CATEGORY.COVENANTS}
```
Pewtro: indentation 
image

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+        tooltip={
+          <>
+            Total damage done by the 8% increase: {formatNumber(this.totalDamage)}
+            <br />
```
Pewtro: unnecessary break line if there's no content underneath

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+          </>
+        }
+      >
+        <BoringSpellValueText spellId={SPELLS.FAELINE_HARMONY_DEBUFF.id}>
+          <img src="/img/sword.png" alt="Damage" className="icon" /> {formatNumber(this.dps)} DPS{' '}
+          <small>
+            {formatPercentage(this.owner.getPercentageOfTotalDamageDone(this.totalDamage))} % of
+            total
+          </small>
```
Pewtro:
```suggestion
          <ItemDamageDone amount={this.totalDamage}/>
```

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+        this.onAffectedDamage,
+      );
+    }
+  }
+  onAffectedDamage(event: DamageEvent) {
+    if (event.targetIsFriendly) {
+      return;
+    }
+    const target = this.enemies.getEntity(event);
+    if (target !== null && target.hasBuff(SPELLS.FAELINE_HARMONY_DEBUFF.id, event.timestamp)) {
```
Pewtro: If I were to write this I would probably do const enemy = this.enemies.getEntity(event); 
if(!enemy || !enemy.hasBuff(...)) { return } 

to avoid potential nesting if statements in the future - but this is also fine however you can simplify it like this: 

```suggestion
    if (target && target.hasBuff(SPELLS.FAELINE_HARMONY_DEBUFF.id)) {
```

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
```diff
@@ -0,0 +1,82 @@
...
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import React from 'react';
+import Enemies from 'parser/shared/modules/Enemies';
+
+
+
+const MOD = 0.08;
+
+class FaelineStompWindwalker extends Analyzer {
```
Pewtro: Shouldn't this simply be called FaelineStomp? We're already in the windwalker directory and then it matches with the file naming

analysis/monkwindwalker/src/CombatLogParser.ts
```diff
@@ -8,6 +8,7 @@ import Buffs from './modules/Buffs';
 import Channeling from './modules/core/Channeling';
 import GlobalCooldown from './modules/core/GlobalCooldown';
 import WeaponsOfOrderWindwalker from './modules/covenants/WeaponsOfOrder';
+import FaelineStompWindwalker from './modules/covenants/FaelineStomp';
```
Pewtro:
```suggestion
import FaelineStomp from './modules/covenants/FaelineStomp';
```

analysis/monkwindwalker/src/CombatLogParser.ts
```diff
@@ -71,6 +72,7 @@ class CombatLogParser extends CoreCombatLogParser {
     fallenOrder: FallenOrder,
     weaponsOfOrder: WeaponsOfOrderWindwalker,
     faelineStomp: FaelineStomp,
+    faeexposure: FaelineStompWindwalker,
```
Pewtro:
```suggestion
    faelineStomp: FaelineStomp,
```

Pewtro (changes requested):
Hey thanks for the PR, I got a few comments that should simplify it up a bit, and a few stylistic ones - but other than that the core logic seems good

analysis/monkwindwalker/src/modules/covenants/FaelineStomp.tsx
Hurstilol: Changed the naming since it was a bit misleading. Its not the stomp this module is about and more the dmg from the legendary. So both modules, the original nf ww and the new one should exist at the same time.

Pewtro: In that case it should be called FaelineHarmony to match the legendary?

Hurstilol: Yes, renamed it already but put 'dmg' at the end to make clear this is only for the dmg calculation (in case abe wants to make a module for the calculation of the healinggain through this legendary)

Pewtro: I mean it's already under windwalker, so for whatever healing done by the windwalker it should just be included in this file and then rendered alongside the damage in a <ItemHealingDone/>. 

If it's a generic legendary that can do both and doesn't change much between the specs it should go under the monk file and account for both things

PR #4712, opened 2021-12-12
[Hunter / APL] Add a last spell cast condition and implement it for Marksmanship hunters

Pewtro:
Pretty simple condition that allows you to check if the last spell cast was a specific spell. 
I'd imagine this might also be useful for monks and their combo thingy, or other generic "don't cast this spell twice in a row ever", or "always cast this spell twice" like is the case for Marksmanship at times when specced into a talent. 

image


src/parser/shared/metrics/apl/conditions/lastSpellCast.tsx
```diff
@@ -0,0 +1,36 @@
+import type Spell from 'common/SPELLS/Spell';
+import { SpellLink } from 'interface';
+import { EventType } from 'parser/core/Events';
+
+import { Condition } from '../index';
+
+export function lastSpellCast(spell: Spell): Condition<boolean> {
+  let sourceID: number;
```
emallson, line 8: you know, i should have thought of (ab)using closures to track state instead of having this extra state parameter. On the one hand, explicit vs implicit + testability. on the other, one fewer parameter 🤔

emallson (approved):

PR #4721, opened 2021-12-18
[Monk] Added Calculated Strikes Module

Hurstilol:
more or less just copy paste stuff from jukos motc module ¯\_(ツ)_/¯


emallson (approved):
At the end of the day, this should *probably* just be part of the MotC module (would be less added code and would produce results that are more accurate---this approach will produce slight overestimates for both MotC and CS). However, it will also be easier to strip this out in 10 months when we're prepping for 10.0 so 🤷

PR #4722, opened 2021-12-21
[Monk] Module for CO conduit

Hurstilol:


analysis/monkwindwalker/src/modules/covenants/CoordinatedOffensive.tsx
```diff
@@ -0,0 +1,116 @@
...
+        .by(SELECTED_PLAYER)
+        .spell([SPELLS.STORM_EARTH_AND_FIRE_EARTH_SPIRIT, SPELLS.STORM_EARTH_AND_FIRE_FIRE_SPIRIT]),
+      this.trackSummons,
+    );
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER_PET).spell(SPELLS.MELEE),
+      this.handleMelee,
+    );
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER_PET).spell(ABILITIES_CLONED_BY_SEF),
```
emallson, line 54: should this also trigger on damage from the player while SEF is active?

emallson (commented):
one (potentially important) question, otherwise looks good

analysis/monkwindwalker/src/modules/covenants/CoordinatedOffensive.tsx
Hurstilol: Nope, CO only buffs dmg from the 2 clones that are not the player

PR #4725, opened 2021-12-26
Added Wasteland Statistic 

abelito75:
This stats hows how much wastelands 6% (self buff) provided on the fight
and since im a gamer I also added the value OTHER peoples 3% buff gave you

THIS DOESN'T include the 3% buffs you provide to allies


src/parser/core/calculateEffectiveDamage.ts
```diff
@@ -12,6 +12,21 @@ import { DamageEvent } from 'parser/core/Events';
...
+  const effectiveDamage = damageIncrease - overKill;
+
+  return Math.max(0, effectiveDamage);
+}
+
+export function calculateOverkillDamage(event: DamageEvent, increase: number): number {
+  const overKill = event.overkill || 0;
+  const raw = (event.amount || 0) + (event.absorbed || 0);
+  const damageIncrease = raw - raw / (1 + increase);
+  const effectiveDamage = damageIncrease - overKill;
```
emallson, line 27: should this just call `calculateEffectiveDamage`?

src/parser/core/calculateEffectiveDamage.ts
```diff
@@ -12,6 +12,21 @@ import { DamageEvent } from 'parser/core/Events';
  * @return the amount of damage attributable on the given damage event from the given boost
  */
 export default function calculateEffectiveDamage(event: DamageEvent, increase: number): number {
+  const overKill = event.overkill || 0;
   const raw = (event.amount || 0) + (event.absorbed || 0);
```
emallson, line 16: this should use `overkill` right? total damage is `amount + absorbed + overkill`

src/parser/core/calculateEffectiveHealing.ts
```diff
@@ -28,3 +28,35 @@ export default function calculateEffectiveHealing(
...
+  relativeHealIncrease: number,
+): number {
+  const amount = event.amount;
+  const absorbed = event.absorbed || 0;
+  const overheal = event.overheal || 0;
+  const raw = amount + absorbed + overheal;
+  const relativeHealingIncreaseFactor = 1 + relativeHealIncrease;
+  const healingIncrease = raw - raw / relativeHealingIncreaseFactor;
+  const effectiveHealing = healingIncrease - overheal;
+  const effectiveOverhealing = healingIncrease - effectiveHealing;
```
emallson, line 59: if the entire `effectiveHealing` is overhealing, then `effectiveHealing` will be negative and thus `effectiveOverhealing` will become larger than `healingIncrease`, inflating the reported overhealing.

Using `calculateEffectiveHealing` corrects for this and avoids code duplication.

src/parser/core/calculateEffectiveDamage.ts
```diff
@@ -12,6 +12,21 @@ import { DamageEvent } from 'parser/core/Events';
...
+  return Math.max(0, effectiveDamage);
+}
+
+export function calculateOverkillDamage(event: DamageEvent, increase: number): number {
+  const overKill = event.overkill || 0;
+  const raw = (event.amount || 0) + (event.absorbed || 0);
+  const damageIncrease = raw - raw / (1 + increase);
+  const effectiveDamage = damageIncrease - overKill;
+
+  const effectiveOverkill = damageIncrease - effectiveDamage;
```
emallson, line 29: similar sign issue as noted for `effectiveOverhealing`.

emallson (changes requested):
some math issues, and you need to update snapshots.

PR #4726, opened 2021-12-27
Rjw bc anom bad (not his fault. He isn't a coder)

abelito75:
We were not counting pre-casts or refresh buffs 
(tbh when you first thing of this ability you don't think refreshes are possible bc duration = cooldown but its hasted)


abelito75:
Shit was stuck. Re-running in tomorrow land

PR #4730, opened 2022-01-01
Sonar Fixes, Part 1

jander99:


local-pipeline.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env sh
```
emallson, line 1: i like the concept, though this won't work for our windows folks unless they're using WSL i think. i'll probably iterate on this later.

emallson (approved):
Overall looks good. Can you link the scanner you're using here? If its the kind of static code checker I'm thinking of you're likely to deal with a LOT of false positives from the dataflow tracing.

local-pipeline.sh
jander99: That's true, I wouldn't know where to begin with a PowerShell equivalent and my .bat file experience is from the 90s 😬

jander99:
> Overall looks good. Can you link the scanner you're using here? If its the kind of static code checker I'm thinking of you're likely to deal with a LOT of false positives from the dataflow tracing.

https://sonarcloud.io/project/overview?id=jander99_WoWAnalyzer

Let me know if you can't access the report, it's their free offering and it's against my fork. Not sure what kind of access I can provide or whether you have to sign up for an account (via Github OAuth).  I haven't tried to understand what it would take to get Github Actions to run a scan yet as it installed a webhook against PRs and the main branch.

PR #4732, opened 2022-01-03
[TBCShaman] Adding a totem tracker for all TBC shaman specs.

TJMoats:
Currently there is no easy way to know how well you are managing your totems in TBC. 

This module adds a statistic display for each of the totem elements:

Screen Shot 2022-01-02 at 10 07 07 PM
Screen Shot 2022-01-02 at 10 07 11 PM
Screen Shot 2022-01-02 at 10 07 16 PM
Screen Shot 2022-01-02 at 10 07 21 PM

I've also added a checklist item to keep your totems up:

Screen Shot 2022-01-02 at 10 06 52 PM

And lastly, I've added a grounding totem tracker. This would have been useful when Vashj was a hard fight, but I wasn't shamaning back then :D

Screen Shot 2022-01-02 at 11 46 11 PM

There is some room for improvements in tracking what you are gaining from each totem type (healing done, mana regen, damage delt), but that can be handled in another PR.


analysis/tbcshaman/src/modules/features/TotemTracker.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const duration = Math.min(possibleDuration, maxDuration);
+    this.totemElementEvents[element][
+      this.totemElementEvents[element].length - 1
+    ].dismissedAt = timestamp;
+    this.totemElementEvents[element][
+      this.totemElementEvents[element].length - 1
+    ].dismissReason = reason;
+    this.totemElementEvents[element][
+      this.totemElementEvents[element].length - 1
+    ].duration = duration;
```
emallson:
```suggestion
    totemEvent.dismissedAt = timestamp;
    totemEvent.dismissReason = reason;
    totemEvent.duration = duration;
```

no need to duplicate this indexing

analysis/tbcshaman/src/modules/spells/totems/BaseTotem.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { ThresholdStyle } from 'parser/core/ParseResults';
+import BoringValue from 'parser/ui/BoringValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+
+import { TotemElements, TOTEMS_BY_ELEMENT } from '../../../totemConstants';
+import TotemTracker from '../../features/TotemTracker';
+
+export default class BaseTotem extends Analyzer {
```
emallson, line 15: not a required change but: thoughts on structuring as a single analyzer with multiple suggestion `get`ters + multiple statistics returned (using a fragment in `statistic`) rather than base class + a subclass per element?

I can't foresee a reason to vary the behavior here in a way that has the subclass be more than a fancy way of setting a parameter.

emallson (changes requested):
one minor code-smell fix (see suggestion) and a comment. other than that looks good

analysis/tbcshaman/src/modules/spells/totems/BaseTotem.tsx
TJMoats: I'm planning on adding element specific extras that didn't make sense to have in every totem element. 

Water totems should show mana/health restores
Fire totems should show how much damage they deal
Earth can show how much damage was reduced by stoneskin
Air can show spells grounded

I did the grounding totem as a one off, but I think I like the idea of having it as a part of the entire air totem package. I could go either way on that one.

analysis/tbcshaman/src/modules/features/TotemTracker.tsx
TJMoats: Good callout, for some reason I had it in my head that it'd do the assignment by val and not by ref

analysis/tbcshaman/src/modules/spells/totems/BaseTotem.tsx
emallson: ah, that makes more sense. carry on :+1:

analysis/tbcshaman/src/modules/features/TotemTracker.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    const totemEvent: TotemEvent = this.totemElementEvents[element][
+      this.totemElementEvents[element].length - 1
+    ];
+    const possibleDuration: number = timestamp - totemEvent.summonedAt;
+    const maxDuration: number = (TotemDurations as any)[totemEvent.totemSpellId] as number;
+    const duration = Math.min(possibleDuration, maxDuration);
+    totemEvent.dismissedAt = timestamp;
+    totemEvent.dismissReason = reason;
+    totemEvent.duration = duration;
```
emallson:
```suggestion
    totemEvent.duration = duration;
```

looks like prettier is catching some trailing spaces. not sure if thats from me or not, rip.

PR #4740, opened 2022-01-16
Fixed a bug where AdaptiveSwarm statistic wasn't counting Feral Moonfire or Feral Thrash. Also fixed some rank issues

kfinch:


analysis/druid/src/shadowlands/AdaptiveSwarm.ts
```diff
@@ -124,7 +127,7 @@ class AdaptiveSwarm extends Analyzer {
     if (this.hasEvolvedSwarm) {
       this._evolvedSwarmPeriodicBoost =
         EVOLVED_SWARM_EFFECT_BY_RANK[
-          this.selectedCombatant.conduitRankBySpellID(SPELLS.EVOLVED_SWARM.id)
+          this.selectedCombatant.conduitRankBySpellID(SPELLS.EVOLVED_SWARM.id) + 2 // +2 because we're assuming its empowered... TODO better handling?
```
emallson, line 130: my thinking was that in 9.2 we just assume everything is empowered. nbd starting now, but this should get rolled into `conduitRankBySpellID` at that time rather than throwing +2 a bunch of places

emallson (approved):

PR #4748, opened 2022-01-23
[TBC] Paladin ret: Adding Seal of Command talent suggestion

llanas:
Adding Suggestion based on usage of high Seal of Command rank


emallson (approved):
looks good, thanks 👍

PR #4756, opened 2022-02-05
Adding a module to better suggest Bladestorm use for Arms warriors

Chasson1992:
Changing the Bladestorm suggestion for Arms warrior by removing the spell from castEfficieny tracker and creating a new module.

Bladestorm rules for Arms are generally as follow:

1. Single Target - Only used if you are rage starved
2. Multi Target - Dont overlap with Sweeping Strikes, Bladestorm doesnt benefit from sweeping strikes
3. Multi Target - Align with big cooldowns such Warbreaker/Avatar

I did this by hooking into 3 events, the cast, on damage and the end of the channel. 

Cast event sets the current event.

On damage will check unqiue Target hits and holds their IDs in an array.

End of the channel triggers the event analysis by checking if this was a single target situation or multi target situation by getting the amount of unique enemies hit. Then checks the rules stated above  to mark this as a bad cast and provide rationale in the event reasoning if nessicary.

Added some "forgiveness" to the cooldown alignment and sweeping strikes falling.

Example Log: 
Single Target
report/h2pfRkbBtGmw8cVg/2-Normal+Shriekwing+-+Kill+(5:49)/Carrottopp/standard/timeline
Multi Target
report/WRqkNnxQV13A2vbg/45-Mythic+Sylvanas+Windrunner+-+Kill+(13:51)/2-Hellring/standard

### DIFF
--- a/analysis/warriorarms/src/CHANGELOG.tsx
+++ b/analysis/warriorarms/src/CHANGELOG.tsx
@@ -5,6 +5,7 @@ import { Abelito75, carglass, Carrottopp, Otthopsy, bandit } from 'CONTRIBUTORS'
 import { SpellLink } from 'interface';
 
 export default [
+  change(date(2022, 2, 5), <>Added a module for better suggestions of <SpellLink id={SPELLS.BLADESTORM.id} icon /> usage.</>, Carrottopp),
   change(date(2021, 11, 8), <>Fixed SpellLink issue for Signet Of Tormented Kings.</>, Abelito75),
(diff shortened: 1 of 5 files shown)


emallson (approved):
Looks good. My only comment is that this is going to overwrite the `inefficientCastReason` text. I suppose that is probably okay, and definitely not a reason to block it. But something to think about.

You can look at the Brewmaster [PurifyingBrew](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/analysis/monkbrewmaster/src/modules/spells/PurifyingBrew.tsx) for an example of handling multiple messages.

PR #4761, opened 2022-02-10
fix: Only run import/no-extraneous-dependencies in CI

MartijnHols:
This speeds up local ESLint by about 40%. I don't think [this rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md) gets triggered very often, since it only checks for missing dependencies and spec devs rarely change dependencies. Still I think it's a good rule to keep enabled in CI, since it does ensure the required dependencies are listed and avoids issues in the future when an unrelated dependency gets updated.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 const CHECK_CODESTYLE = process.env.CODE_STYLE === 'true';
+const CI = Boolean(process.env.CI);
 
 module.exports = {
   extends: ['@emico/eslint-config', 'plugin:react/jsx-runtime'],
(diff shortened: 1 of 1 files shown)


emallson (approved):
still not a HUGE fan of only running this in CI, but you're right that it is virtually never triggered so it probably won't come up.

MartijnHols:
Alternatively we could run it on CLI and in CI, but not during precommit and in the builder?

emallson:
I think we start here and deal with that if/when it actually happens

PR #4763, opened 2022-02-10
Update react-scripts (improves dev performance!)

MartijnHols:
# Depends on #4762 

Had to update vega to fix several type errors.

Fixed the iterator.js warning in vega with a patch. The file doesn't exist anyway, so nothing is actually changed.

Modernized tsconfig.json

# Don't merge until someone with Windows confirmed this is also working for them

I got burned once before and since this has a similar risk, let's make sure it doesn't happen again. 🔥


emallson:
Looks like the version of jest pulled in by this version bump causes problems with our integration test setup. That setup honestly needs updating as well, but it does mean that we can't easily merge it.

MartijnHols:
I’ll take a look when I have time, shouldn’t be too big of an issue

MartijnHols:
More importantly: this still needs to be tested on Windows

Pewtro:
Windows 11: 
Yarn install fine
Yarn start is fine - can load logs without issues
Yarn build seems fine too

MartijnHols:
@Pewtro thanks!

I think this is good to go then.

One thing I should mention is that Lingui relies on babel macros which aren't supported by SWC and prevents us from using it. This is unfortunate, as SWC would be a huge performance improvement. It doesn't seem likely this will be resolved quickly, but hopefully it will be resolved eventually. A lot of people are switching to SWC and finding this restriction.

PR #4767, opened 2022-02-16
Fixing CanceledCasts module

TJMoats:
I guess a linter got to this code?

### DIFF
--- a/analysis/tbcpriest/src/CHANGELOG.tsx
+++ b/analysis/tbcpriest/src/CHANGELOG.tsx
@@ -6,6 +6,11 @@ import * as SPELLS from './SPELLS';
 
 export default [
   change(
+    date(2022, 2, 16),
+    'TBC Priest Fixing CanceledCasts module.',
+    Khadaj,
+  ),
+  change(
     date(2021, 9, 29),
     'TBC Priest Fixing mana efficiency and adding CanceledCasts module.',
     Khadaj,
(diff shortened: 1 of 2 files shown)


emallson:
`for...of` should work fine here. what happened?

TJMoats:
for...of works if you don't index the object. I will switch it to use for of.

emallson:
That is weird. I guess for of converts keys to strings? May be better to just switch this to `Object.values(this.canceledCasts).reduce((a, b) => a + b));` (and a `.map` call in the latter section) rather than fighting the linter on this.

TJMoats:
`for...in` uses the index of the thing you are iterating over
`for...of` uses the value of the thing

There is no (easy) way to get an index out of `for...of` which is why this used `for...in` initially.

```
const foo = {1: "a", 2: "b", 3: "c"}
for (const bar in foo){
  console.log(bar) // outputs 1 2 3
  console.log(foo[bar]) // outputs "a" "b" "c"
}

// foo is not iterable because we are using an object, not an array. Objects are not iterable by default.
// This will throw an error
for (const baz of foo){
  console.log(baz)
}
```

```
const foo_array = ["a", "b", "c"]
for (const bar in foo_array){
  console.log(bar) // outputs 0 1 2
  console.log(foo_array[bar]) // outputs "a" "b" "c"
}

// foo is iterable because we are using an array.
// This will not throw an error
for (const baz of foo_array){
  console.log(baz)  // outputs "a" "b" "c"
}
```

emallson:
Ah, that'd explain it. I don't use for loops that often in js anymore...partly because of nonsense like this.

PR #4776, opened 2022-02-26
Arms warrior analysis ready for 9.2

Chasson1992:
Bumping arms warrior compatibility for 9.2

Also, file was typescript but wasn't using the Config type from parser. So updated that too.


emallson (approved):
changelog entry then you're good 👍

PR #4779, opened 2022-03-01
 fixed bug in Rough Outlaw module stating incorrect high value buffs.

bhawkins6177:
This is a bug fix for issue ##4570.

I changed the high value buffs listed as Ruthless Precision and Grand Melee to True Bearing and Broadside.

### DIFF
--- a/analysis/rogueoutlaw/src/CHANGELOG.js
+++ b/analysis/rogueoutlaw/src/CHANGELOG.js
@@ -1,10 +1,11 @@
 import { change, date } from 'common/changelog';
 import SPELLS from 'common/SPELLS';
-import { Putro, Tyndi, Zeboot, Canotsa, Hordehobbs, Akai, ab } from 'CONTRIBUTORS';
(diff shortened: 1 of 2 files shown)


emallson (commented):
The code changed here looks fine, but before we merge I want to check: does this line up with what the analyzer actually does?

I *think* it is (broadside and true bearing have the highest values in `RollTheBonesCastTracker.tsx`). I just want to confirm because it is better for *both* analysis and text to be wrong (because then it is clear that it is wrong) than for the text to be right but the analysis to be wrong.

bhawkins6177:
> The code changed here looks fine, but before we merge I want to check: does this line up with what the analyzer actually does?
> 
> I _think_ it is (broadside and true bearing have the highest values in `RollTheBonesCastTracker.tsx`). I just want to confirm because it is better for _both_ analysis and text to be wrong (because then it is clear that it is wrong) than for the text to be right but the analysis to be wrong.

I see, I just checked the RollTheBonesCastTracker.tsx . What other way can I use to check the value?

emallson:
That seems right; I'll get it merged then :)

PR #4784, opened 2022-03-05
[Frost Mage] 9.2 Update

Sharrq:
Updated Frost Mage to 9.2. This includes the following changes

- Added support for Comet Storm. Verifies that Comet Storm was shattered or hit multiple targets (AOE)
- Adjusted the CDR that Blizzard provided to Frozen Orb from 500ms to 300ms
- Updated Integration Tests and the Example Log


analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
```diff
@@ -0,0 +1,45 @@
...
+export const HITS_TARGET = 'HitsTarget';
+
+const EVENT_LINKS: EventLink[] = [
+  {
+    linkRelation: FROM_HARDCAST,
+    linkingEventId: SPELLS.COMET_STORM_DAMAGE.id,
+    linkingEventType: EventType.Damage,
+    referencedEventId: SPELLS.COMET_STORM_TALENT.id,
+    referencedEventType: EventType.Cast,
+    forwardBufferMs: CAST_BUFFER_MS,
```
kfinch: Unless I'm mistaken, the cast event will always be before the damage events, meaning `forwardBufferMs` should be zero. Leaving it at 2.5 seconds is going to cause a lot of false positives once the set bonus comes into play.

analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
```diff
@@ -0,0 +1,45 @@
...
+    backwardBufferMs: CAST_BUFFER_MS,
```
kfinch, line 19: 2500 ms for the back buffer is almost certainly too large - you should do some logs analysis and find a more reasonable number (or prove that it actually does need to be 2500). A too large number will cause false positives once the set bonus comes into play.

analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+export function isFromHardcast(event: DamageEvent): boolean {
+  return HasRelatedEvent(event, FROM_HARDCAST);
+}
+
+export function hardcastTargetsHit(event: CastEvent): number {
+  return GetRelatedEvents(event, HITS_TARGET).length;
+}
```
kfinch: I see this isn't being used, and in the CometStorm module you are manually looking up 'HitsTarget' links. You should instead delete `.length` and make this return `DamageEvent[]`. Then you could use this in the CometStorm module.

kfinch (changes requested):
I reviewed only the Comet Storm stuff here. Looks like it should overall work, but I suggested some changes to tighten it up.

analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
kfinch: I realized this would actually have to be `AnyEvent[]` because of how `GetRelatedEvents` is implemented, but you get the idea

analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
```diff
@@ -0,0 +1,45 @@
...
+    referencedEventId: SPELLS.COMET_STORM_TALENT.id,
+    referencedEventType: EventType.Cast,
+    forwardBufferMs: CAST_BUFFER_MS,
+    backwardBufferMs: CAST_BUFFER_MS,
+    anyTarget: true,
+    reverseLinkRelation: HITS_TARGET,
+  },
+];
+
+class CometStormLinkNormalizer extends EventLinkNormalizer {
```
kfinch, line 26: Please add a brief couple sentence comment on what this Normalizer does.

analysis/magefrost/src/normalizers/CometStormLinkNormalizer.ts
Sharrq: I did confirm via logs. I tried 2000 and there was stuff getting cut off from it
image

Keep in mind that when you cast Comet Storm (Or Meteor for Fire), it doesnt do anything for like a second other than start the animation up in the sky ... and then it drops. The above example is 2393ms

kfinch: Oh I knew there was a delay but I didn't realize it was that long - this is enough proof for me you can leave it as is

kfinch (approved):

PR #4787, opened 2022-03-08
Fury Covenant updates

tyndivelspaz:
Added Fury's Condemn, and other covenant spells to spellbook. Bumped to partial 9.2 support.


analysis/warrior/src/Condemn.ts
```diff
@@ -8,10 +7,12 @@ class Condemn extends Analyzer {
     abilities: Abilities,
   };
 
+  condemnSpellId: number = 0;
+
   constructor(options: Options) {
     super(options);
 
-    this.active = this.selectedCombatant.hasCovenant(COVENANTS.VENTHYR.id);
+    this.active = false; // this.selectedCombatant.hasCovenant(COVENANTS.VENTHYR.id);
```
emallson: if this module is just going to be disabled, it should simply be removed.

emallson (changes requested):
looks good, aside from one change (removing the Condemn module since it is unneeded)

analysis/warrior/src/Condemn.ts
tyndivelspaz: Ah right. D'oh! Removed.

emallson (approved):

PR #4788, opened 2022-03-13
[Core] Add has2Piece and has4Piece functions to combatant information

Pewtro:
- Adds a simple has2Piece() and has4Piece() check to see if a player has 2 or 4 piece active.
- Removes punchcard logic from combatant since it is no longer used 

# How to use 
If you want to be explicit in your code:
```
import { MONK_TIER_ID } from 'common/ITEMS/shadowlands'; 
<----> 
this.active = this.selectedCombatant.has2Piece(MONK_TIER_ID); 
```
If you don't want additional imports we will look for the setID for you based on current active spec:
```
this.active = this.selectedCombatant.has2Piece();
```


src/parser/core/Combatant.ts
```diff
@@ -513,6 +468,20 @@ class Combatant extends Entity {
 
   // endregion
 
+  // region Tier
+  get tierPieces() {
+    return [this.head, this.shoulder, this.chest, this.legs, this.hands];
+  }
+
+  has2Piece(setId: number) {
```
abelito75: Is there a real need for setId to be passed in?
We can get the specID of the player here couldn't here and then filter from that?

Pewtro: passing setId is now optional, if not passed, the tier setId will be calculated by mapping spec to class to setId

emallson (approved):

PR #4796, opened 2022-03-16
MW 4set init

abelito75:
Duplicate spells are same spell different source/spellID. They are truly different.
image


Pewtro:
If I was you I'd try and group those same spells by using their name or similiar to group em up, then just whatever main spell id

PR #4806, opened 2022-03-24
Fixed degraded experience when FO isn't cast at all in the pull

trevorm4:
https://www.warcraftlogs.com/reports/7fj8WxvpR2TBncVY#fight=4&type=healing&source=13 offending log (doesn't cast FO at all)

### DIFF
--- a/analysis/monkmistweaver/src/modules/shadowlands/covenant/FallenOrderAverageHPOfTargetOnCast.tsx
+++ b/analysis/monkmistweaver/src/modules/shadowlands/covenant/FallenOrderAverageHPOfTargetOnCast.tsx
@@ -34,6 +34,7 @@ class FallenOrderAverageHPOfTargetOnCast extends Analyzer {
   }
 
   healListener(event: CastEvent) {
+    console.log(event);
     const hpPercent = event.hitPoints! / event.maxHitPoints!;
     this.hpOfTargetOnCast.push(hpPercent);
     if (event.ability.guid === SPELLS.FALLEN_ORDER_ENVELOPING_MIST.id) {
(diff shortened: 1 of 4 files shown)


analysis/monkmistweaver/src/modules/shadowlands/covenant/FallenOrderAverageHPOfTargetOnCast.tsx
```diff
@@ -65,9 +65,12 @@ class FallenOrderAverageHPOfTargetOnCast extends Analyzer {
   }
 
   statistic() {
-    const averageOfAll = this.getAverage(this.hpOfTargetOnCast);
```
abelito75, line 68: I'd much rather have an if statement that disables the stat if the length is zero than 20 billion ternaries

abelito75 (changes requested):

analysis/monkmistweaver/src/modules/shadowlands/covenant/FallenOrderCraneAverage.tsx
```diff
@@ -53,6 +53,9 @@ class FallenOrderCraneAverage extends Analyzer {
   }
 
   statistic() {
+    if (this.mwClones === 0 || this.soomCasts === 0 || this.mwClones === 0) {
```
abelito75: 2 things
1) you are checking for mwClones === 0 twice for no reason

2) You shouldn't worry about soom or enm casts. If I'm reading this code correctly those being 0 won't make a crash and on top of that if someone summons 5 clones and get 0 enm and/or soom casts that would be a valid stat

abelito75 (changes requested):
Almost there... I think

PR #4808, opened 2022-03-24
Untempered Dedication Implementation

xizbow1:
Added support for untempered dedication in conduit list, and added analysis for it. Stole a good bit of code from the Flash Concentration module for Holy Priest, as they are very similar.


rp4rk (approved):
lgtm

analysis/paladinholy/src/modules/items/UntemperedDedication.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { When, ThresholdStyle } from 'parser/core/ParseResults';
+
+class UntemperedDedication extends Analyzer {
+  // Warcraft Logs seems to have issues with the stack status at fight start
+  // we do not know how many stacks the character has and how long it has until expiration
+  // in WoWA we have a prepull apply buff with the same lack of info
+  // Realistically, people start at 5 stacks, or _isFullStack is going to turn to false on the first applyBuff or applyBuffStack
+  // the only issue is that the first LOTM may be considered wasteful when it is not.
+  _isFullStack = true;
+  // There is also no buff event generated when LOTM is cast while there are already 5 stacks
```
abelito75: There is. Its called refreshbuff

analysis/paladinholy/src/modules/items/UntemperedDedication.tsx
```diff
@@ -0,0 +1,111 @@
...
+  // so we are maintaining the date of the last refresh here.
+  _lastFullStackRefresh = 0;
+  // a counter for when LOTM stacks are dropped
+  _totalDrops = 0;
+  // tracker for number of current stacks
+  _currentStacks = 5;
```
abelito75: I wouldn't assume you start at 5.
The vast majority of players are not going to start at five

If you are really concerned about the edge case I'd capture the first n LoTM until we see a buff applied/buff stack gained/buff refreshed/buff removed event and then deal with them

analysis/paladinholy/src/modules/items/UntemperedDedication.tsx
```diff
@@ -0,0 +1,111 @@
...
+    this._totalDrops += 1;
+    this._isFullStack = false;
+    this._currentStacks = 0;
+  }
+
+  get dropSuggestion() {
+    return {
+      actual: this._totalDrops,
+      isGreaterThan: {
+        minor: 0,
```
abelito75: This is too low. 
Since this completely relies on holy power generation and dumping anytime there is force downtime in a fight you have a chance to drop it. You'd rather want a % uptime here or better yet % LoTM casts buffed vs unbuffed IE 20/25 = bad

analysis/paladinholy/src/modules/items/UntemperedDedication.tsx
```diff
@@ -0,0 +1,111 @@
...
+      },
+      style: ThresholdStyle.NUMBER,
+    };
+  }
+
+  suggestions(when: When) {
+    when(this.dropSuggestion).addSuggestion((suggest, actual) =>
+      suggest(
+        <span>
+          You dropped your Untempered Dedication stacks {actual} times. Try to keep it up at all
```
abelito75: TECHNICALLY having the stacks up 100% of the time is completely pointless. The only time it matters is if you're gonna LoTM...

analysis/paladinholy/src/modules/items/UntemperedDedication.tsx
```diff
@@ -0,0 +1,111 @@
+import SPELLS from 'common/SPELLS';
```
abelito75: This isn't about this line but rather the whole file.
Why is it in the items folder when its a conduit? Wouldn't it make more sense to put it in a conduits folder... which i'd imagine would be under the shadowlands folder

analysis/paladinholy/src/CombatLogParser.ts
```diff
@@ -109,6 +110,9 @@ class CombatLogParser extends CoreCombatLogParser {
     // Covenant
     ashenHallow: AshenHallow,
     divineToll: DivineToll,
+
+    // Items
```
abelito75: Why does this have the item tag when its a conduit

abelito75 (changes requested):
This is just a first pass so there still might be a some more things to fix later

analysis/paladinholy/src/modules/shadowlands/conduits/UntemperedDedication.tsx
```diff
@@ -0,0 +1,60 @@
...
+import SpellLink from 'interface/SpellLink';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
+import { RemoveBuffEvent } from 'parser/core/Events';
+import SUGGESTION_IMPORTANCE from 'parser/core/ISSUE_IMPORTANCE';
+import { When, ThresholdStyle } from 'parser/core/ParseResults';
+
+class UntemperedDedication extends Analyzer {
+  // a counter for when UD stacks are dropped
+  private totalDrops = 0;
```
abelito75: don't need the private modifier here

analysis/paladinholy/src/modules/shadowlands/conduits/UntemperedDedication.tsx
```diff
@@ -0,0 +1,60 @@
...
+  suggestions(when: When) {
+    when(this.dropSuggestion).addSuggestion((suggest, actual) =>
+      suggest(
+        <span>
+          You dropped your <SpellLink id={SPELLS.UNTEMPERED_DEDICATION.id} /> stacks {actual} times.
+          Try to keep it up at all times, as this is a fairly large loss of healing.
+        </span>,
+      )
+        .icon(conduits.UNTEMPERED_DEDICATION.icon)
+        .staticImportance(SUGGESTION_IMPORTANCE.MAJOR)
```
abelito75: This shouldn't be here. Like at all i'm pretty sure

This basically said `if there is a problem. make it major` Which this metric is very prone to things like boss downtime meaning making it major always feels wrong

abelito75 (changes requested):

analysis/paladinholy/src/modules/shadowlands/conduits/UntemperedDedication.tsx
```diff
@@ -0,0 +1,60 @@
...
+import SpellLink from 'interface/SpellLink';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
+import { RemoveBuffEvent } from 'parser/core/Events';
+import { When, ThresholdStyle } from 'parser/core/ParseResults';
+
+class UntemperedDedication extends Analyzer {
+  // a counter for when UD stacks are dropped
+  totalDrops = 0;
+  conduitRank = 0;
```
abelito75: you don't use this variable anyway
kill it

analysis/paladinholy/src/modules/shadowlands/conduits/UntemperedDedication.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasConduitBySpellID(SPELLS.UNTEMPERED_DEDICATION.id);
+    this.conduitRank = this.selectedCombatant.conduitRankBySpellID(SPELLS.UNTEMPERED_DEDICATION.id);
```
abelito75: you don't use this at all. remove it

abelito75 (changes requested):
almost there. Looks a lot better

emallson (approved):
abe's comments were addressed, merging :+1:

PR #4810, opened 2022-03-26
some Blood DK updates for SFO

mfield:
Plenty left to do before calling BDK supported for 9.2, just dipping my toes in for now. Thresholds are super arbitrary, would love some guidance there...

- add CRW module
- Bonestorm has a purpose on ST with tier


analysis/deathknightblood/src/modules/features/BoneShieldTimesByStacks.js
```diff
@@ -44,9 +64,35 @@ class BoneShieldStacksBySeconds extends Analyzer {
...
+    }
     this.lastBoneShieldStack = currentStacks(event);
   }
 
+  trackApplication(event) {
+    // Can't track this with buff applied events in case the player
+    // refreshes at 10 stacks.
+    const isMarrow = (event.ability.guid === SPELLS.MARROWREND.spellID);
+    const isDRW = (event.ability.guid === SPELLS.DANCING_RUNE_WEAPON.spellID);
+    const hasCRW = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.CRIMSON_RUNE_WEAPON.bonusID);
```
emallson: just merged something into the main `shadowlands` branch that changes how legendary checks worked. you'll want to pull that into your branch and change this to

```suggestion
    const hasCRW = this.selectedCombatant.hasLegendary(SPELLS.CRIMSON_RUNE_WEAPON);
```

analysis/deathknightblood/src/modules/features/BoneShieldTimesByStacks.js
```diff
@@ -58,6 +104,28 @@ class BoneShieldStacksBySeconds extends Analyzer {
     });
     return avgStacks;
   }
+
+  reduceDRWCooldown(stackDiff) {
+    if (!this.selectedCombatant.hasLegendaryByBonusID(SPELLS.CRIMSON_RUNE_WEAPON.bonusID)) {
```
emallson: rather than repeatedly checking this, check it in the constructor and then disable the whole module if the legendary is missing (if the module is only intended to be active with CRW equipped). for example:

```js
if(!this.selectedCombatant.hasLegendary(SPELLS.CRIMSON_RUNE_WEAPON)) {
   this.active = false;
   return;
}
```
before registering your event listeners.

analysis/deathknightblood/src/modules/items/CrimsonRuneWeapon.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+class CrimsonRuneWeapon extends Analyzer {
+  static dependencies = {
+    boneShieldTimesByStacks: BoneShieldTimesByStacks,
+  };
+
+  constructor(options: Options) {
+    super(options);
+
+    const active = this.selectedCombatant.hasLegendaryByBonusID(SPELLS.CRIMSON_RUNE_WEAPON.bonusID);
```
emallson: this will need to change after merging `shadowlands` into your branch.
```suggestion
    const active = this.selectedCombatant.hasLegendary(SPELLS.CRIMSON_RUNE_WEAPON);
```

analysis/deathknightblood/src/modules/items/CrimsonRuneWeapon.tsx
```diff
@@ -0,0 +1,97 @@
...
+    this.active = active;
+    if (!active) {
+      return;
+    }
+  }
+
+  get reduction() {
+    // @ts-ignore
+    return this.boneShieldTimesByStacks.totalDRWCooldownReduction / 1000;
```
emallson, line 30: if you have time, then you can convert the `BoneShieldTimesByStacks` module to ts. Otherwise, you can use `as` to convert the unknown type to a known type.  rather than using `@ts-ignore`.

analysis/deathknightblood/src/modules/items/CrimsonRuneWeapon.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  get wastedReduction() {
+    // @ts-ignore
+    return this.boneShieldTimesByStacks.totalDRWCooldownReductionWasted / 1000;
```
emallson, line 34: ditto here

emallson (changes requested):
3 main things you need to do:

1. Run prettier (will fix 99% of the lint errors). `yarn prettier --fix` should do the trick iirc. Your editor can probably also run prettier for you on save.
2. Replace the `@ts-ignore` lines with typecasts as needed, or convert `BoneShieldTimesByStacks` to typescript so that casts aren't needed
3. Merge `shadowlands` into your branch, then update the legendary checks to use the `hasLegendary` method. This is necessary because of changes to support the Unity legendaries.

Once those are done, we're good to merge 👍

analysis/deathknightblood/src/modules/features/BoneShieldTimesByStacks.js
mfield: Bone shield buff management is relevant regardless of talent/legendary, dancing rune weapon cooldown is added when you have the CRW legendary on. I was copying the pattern from https://github.com/WoWAnalyzer/WoWAnalyzer/blob/shadowlands/analysis/deathknightblood/src/modules/runicpower/RunicPowerTracker.js where cooldown reduction is tracked for vamp blood if red thirst talent is selected. If you'd rather see the CDR tracking from CRW in a module of its own I could do that instead.

analysis/deathknightblood/src/modules/items/CrimsonRuneWeapon.tsx
mfield: Abelito helped me find a good solution for this in discord :)

analysis/deathknightblood/src/modules/features/BoneShieldTimesByStacks.js
emallson: :+1: that is fine, it just looked like it *might* be turning off (effectively) all the logic of the module here---if it isn't then how you have it is good

emallson:
One more thing, you'll need to add a changelog entry to `analysis/deathknightblood/src/CHANGELOG`. you will need to add yourself to the `CONTRIBUTORS` file as well. we ask for this so we make sure that people get credit for the work they do :+1:

mfield:
Bookkeeping sorted. I've asked for a 2nd pair of eyes in Acherus, hoping they'll stop by to double check me on the BDK specifics.

emallson:
> Bookkeeping sorted. I've asked for a 2nd pair of eyes in Acherus, hoping they'll stop by to double check me on the BDK specifics.

Did you hear back from them? We can merge this whenever you give me the thumbs up

mfield:
Thanks for the poke, I've passed it along. I actually have another branch ready to PR based off this, so let's merge and if they have comments I can put the fixes in that diff.

PR #4822, opened 2022-03-30
[Fury Warrior] Natures Fury and Spear of Bastion Update

tyndivelspaz:
Added tracking for: 
- Nature's Fury legendary 
- Spear of Bastion ability
- Deathmaker Legendary power


analysis/warriorfury/src/modules/shadowlands/covenants/SpearOfBastion.tsx
```diff
@@ -0,0 +1,37 @@
...
+    this.active = this.selectedCombatant.hasCovenant(COVENANTS.KYRIAN.id);
+
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(SPELLS.SPEAR_OF_BASTION_ENERGIZE),
+      this.onSpearDamage,
+    );
+  }
+
+  onSpearDamage(event: DamageEvent) {
+    this._spearOfBastionDamage += event.amount;
```
abelito75: you should also look at event.absorbed so damage to shields is counted (as that is still effective damage)

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,57 @@
...
+    );
+
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(SPELLS.NATURES_FURY_DAMAGE),
+      this.onNaturesFuryDamage,
+    );
+  }
+
+  onAftershockDamage(event: DamageEvent) {
+    this._ancientAftershockDamage += event.amount;
```
abelito75: event.absorbed here too

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  onNaturesFuryDamage(event: DamageEvent) {
+    this._naturesFuryDamage += event.amount;
```
abelito75: event.absorbed here too

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+class NaturesFury extends Analyzer {
+  _ancientAftershockDamage: number = 0;
+  _naturesFuryDamage: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.active =
+      this.selectedCombatant.hasCovenant(COVENANTS.NIGHT_FAE.id) &&
+      (this.selectedCombatant.hasLegendary(SPELLS.NATURES_FURY) ||
```
abelito75: you don't need to check for unity

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+
+class NaturesFury extends Analyzer {
+  _ancientAftershockDamage: number = 0;
+  _naturesFuryDamage: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.active =
```
abelito75: In all reality should the whole module be disabled if they don't have the legendary?
From what it seems this stat shows both for the covenant in general and the legendary... might want to break it up?

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,57 @@
...
+      this.selectedCombatant.hasCovenant(COVENANTS.NIGHT_FAE.id) &&
+      (this.selectedCombatant.hasLegendary(SPELLS.NATURES_FURY) ||
+        this.selectedCombatant.hasLegendary(SPELLS.UNITY));
+
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(SPELLS.ANCIENT_AFTERSHOCK),
```
abelito75: you can do [SPELLS.ANCIENT_AFTERSHOCK, SPELLS.ANCIENT_AFTERSHOCK_DOT] to save some space 
iirc

analysis/warriorfury/src/modules/talents/Siegebreaker.tsx
```diff
@@ -28,7 +28,9 @@ class Siegebreaker extends Analyzer {
 
   constructor(options: Options) {
     super(options);
-    this.active = this.selectedCombatant.hasTalent(SPELLS.SIEGEBREAKER_TALENT.id);
+    this.active =
```
abelito75: Why is this legendary required to show the stat?

src/common/SPELLS/shadowlands/legendaries/warrior.ts
```diff
@@ -77,6 +82,23 @@ const legendaries = {
...
+    bonusID: 7471,
+    name: "Nature's Fury",
+    icon: 'ability_ardenweald_warrior',
+  },
+  NATURES_FURY_DAMAGE: {
+    id: 354163,
+    name: "Nature's Fury",
+    icon: 'ability_ardenweald_warrior',
+  },
+  UNITY_WARRIOR: {
```
abelito75: WCL now sends Unity over just as the correct covenant legendary. No need to add unity into the code base

abelito75 (changes requested):

analysis/warriorfury/src/modules/talents/Siegebreaker.tsx
tyndivelspaz: The Fury legendary Deathmaker has a chance to apply Siegebreaker. I figured it would be best to utilize the module already created rather than create a duplicate.

abelito75: I'd assume people would want to see the difference between `Deathmaker Siegebreaker` and regular `Siegebreaker`

tyndivelspaz: Is it safe to assume a player would never use Deathmaker and choose the Siegebreaker talent. They both apply the same debuff spell id

abelito75: I would never assume anything like that

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
```diff
@@ -0,0 +1,55 @@
...
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+
+class NaturesFury extends Analyzer {
+  _ancientAftershockDamage: number = 0;
+  _naturesFuryDamage: number = 0;
+  _hasLegendary: boolean = this.selectedCombatant.hasLegendary(SPELLS.NATURES_FURY);
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasCovenant(COVENANTS.NIGHT_FAE.id) || this._hasLegendary;
```
emallson: why does this key off of the covenant as well as the legendary? we can't rely on people having Unity (unfortunately), so if this is assuming that if you're NF you'll have Unity and thus Nature's Fury Equipped, it needs to change.

You can just check for the presence of the legendary, no need to check covenant as well.

emallson (commented):
one last check before we merge

analysis/warriorfury/src/modules/shadowlands/legendaries/NaturesFury.tsx
tyndivelspaz: Ahh right. No need to check for covenant and legendary.

Fixed. Separated Ancient Aftershock from the Nature's Fury module.

emallson (approved):

PR #4823, opened 2022-03-30
Resto Druid - added Tranquility to throughtput tracker tab

kfinch:


abelito75:
Why does tranq need to be added to the throughput tracker?

emallson:
Tree of Life and Flourish are in it. Resto seems to be using it for healing cds generally. Seems fine?

abelito75:
Just seems weird to me to include something you can't do anything during to it...
I press button. I am now stuck doing that action. The action has ended, throughput tracker tracking it (with the provided config) also ended

PR #4831, opened 2022-04-03
[Mage] Additional Features and Updates

Sharrq:
Made the following changes

- Adjusted Radiant Spark to no longer allow greater than the threshold of casts per Radiant Spark (4 with Harmonic Echo, 5 without)
- Adjusted Radiant Spark to also accept the AOE use of Radiant Spark if it hits 3+ targets
- Added a statistic to Fire to show the spells that were used during Combustion
- Updated the integration tests and sample report for Fire
- Added a Munched Procs module for Frost and added it to the Checklist
- Updated the integration tests for Frost


analysis/magearcane/src/modules/Checklist/Component.tsx
```diff
@@ -113,7 +113,7 @@ const ArcaneMageChecklist = ({ combatant, castEfficiency, thresholds }: Checklis
           />
           <Requirement
             name="Radiant Spark Utilization"
-            tooltip="Since Arcane Blast hits very hard when at 4 Arcane Charges, you should use Radiant Spark's damage increase to make Arcane Blast hit even harder. Every time you cast Radiant Spark, you should cast 5 Arcane Blasts (4 if using the Harmonic Echo/Unity Legendary) before Radiant Spark ends."
+            tooltip="Since Arcane Blast hits very hard when at 4 Arcane Charges, you should use Radiant Spark's damage increase to make Arcane Blast hit even harder. Every time you cast Radiant Spark, you should cast 5 Arcane Blasts (4 if using the Harmonic Echo/Unity Legendary) before Radiant Spark ends. Alternatively, if there is 3 or more targets, you can use Arcane Explosion, Arcane Orb, and Arcane Barrage instead of Arcane Blast."
```
Pewtro:
```suggestion
            tooltip="Since Arcane Blast hits very hard when at 4 Arcane Charges, you should use Radiant Spark's damage increase to make Arcane Blast hit even harder. Every time you cast Radiant Spark, you should cast 5 Arcane Blasts (4 if using the Harmonic Echo/Unity Legendary) before Radiant Spark ends. Alternatively, if there are 3 or more targets, you can use Arcane Explosion, Arcane Orb, and Arcane Barrage instead of Arcane Blast."
```

Pewtro (approved):
Ready to merge whenever you're ready

PR #4833, opened 2022-04-03
Dependabot version bumps

jander99:


Pewtro:
Minimist update was intentionally ignored here: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4827 
I can't see a dependabot alert for ansi-regex in this repo.

jander99:
I guess I should make it a habit to go through all the closed PRs of this repo instead of relying on Dependabot on my fork. I can't see any Dependabot alerts for this repo. Here's the [PR](https://github.com/jander99/WoWAnalyzer/pull/14) from my fork.

emallson:
Generally, any dependabot alerts that only impact local development are ignored. Almost all of them are trivial regex explosion exploits that aren't relevant for the place the libraries are used. If a dependabot alert is triggered for something in production or as part of the build process, then its worth us looking at.

PR #4834, opened 2022-04-03
Mikanikos CDR for SpellUsable

mfield:
live: 
image
after:
image

Will changelog after https://github.com/WoWAnalyzer/WoWAnalyzer/pull/4810 so I don't conflict myself in contributors.


src/parser/shadowlands/modules/Mikanikos.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+  }
+
+  onCast(event: CastEvent) {
+    this.casts.push([event, 0]);
+  }
+
+  onDebuff(event: ApplyDebuffEvent) {
+    if (this.casts.length === 0) {
+      this.warn(event.ability.name, 'applied before first', this.classAbility.name, 'cast');
```
Pewtro: This isn't something we can do much about - so I'd prefer if we put it behind a bool like const debug = false; and then debug && this.warn(...). 

There's already a lot of stuff in the console that we _can_ do something about (haste, cooldowns, etc)

src/parser/shadowlands/modules/Mikanikos.ts
```diff
@@ -0,0 +1,92 @@
...
+  classAbility: SpellInfo;
+
+  constructor(options: Options) {
+    super(options);
+
+    const classId = getClassBySpecId(this.selectedCombatant.specId);
+    this.classAbility = SPELLS_BY_CLASS[classId];
+
+    const active =
+      classId < Infinity &&
```
Pewtro: You're doing a check here for classId < Infinity - but you're using it as an index above, might make sense to do this before that index usage, not that it matters much but 😄

src/parser/shadowlands/modules/Mikanikos.ts
```diff
@@ -0,0 +1,92 @@
...
+  [CLASSES.MONK]: SPELLS.WEAPONS_OF_ORDER,
+  [CLASSES.PALADIN]: SPELLS.DIVINE_TOLL,
+  [CLASSES.PRIEST]: SPELLS.BOON_OF_THE_ASCENDED,
+  [CLASSES.ROGUE]: SPELLS.ECHOING_REPRIMAND,
+  [CLASSES.SHAMAN]: SPELLS.VESPER_TOTEM,
+  [CLASSES.WARLOCK]: SPELLS.SCOURING_TITHE,
+  [CLASSES.WARRIOR]: SPELLS.SPEAR_OF_BASTION,
+};
+
+class Mikanikos extends Analyzer {
```
Pewtro: This should probably be called EffusiveAnimaAccelerator as that is what the module is for. Then it would fit under shadowlands/covenants/kyrian or something similarly named

src/parser/shadowlands/modules/Mikanikos.ts
mfield: i tried this but i think it looks better to keep all the conditions in the one if statement.

src/parser/shadowlands/modules/covenants/kyrian/EffusiveAnimaAccelerator.ts
```diff
@@ -0,0 +1,95 @@
...
+      return;
+    }
+    const [lastCast, cdr] = this.casts[this.casts.length - 1];
+    const baseCd = this.abilities.getExpectedCooldownDuration(lastCast.ability.guid, lastCast);
+    if (baseCd == null) {
+      debug && this.warn('no cd for', lastCast.ability.name);
+      return;
+    }
+
+    if (cdr >= baseCd / 3) {
```
Pewtro: the 3 here and the 15 down below are magic numbers and should be taken out into a const to indicate what they mean (I know it's that every CDR proc is 1/15 of base CD and then it can proc up to 5 times aka 1/3 of base CD, but it's not intuitive by reading the code unless you know what it does).

e.g. the 3 here could probably be converted to some tracker that counts amount of times it got reduced and if that's = 5 (with the 5 being a const MAX_REDUCTION_PROCS = 5 or similiar) then doesn't allow future reduction -- that's more clear than check cdr vs baseCD / 3.

Pewtro (changes requested):

src/parser/shadowlands/modules/covenants/kyrian/EffusiveAnimaAccelerator.ts
mfield: 👍 can do

src/parser/shadowlands/modules/covenants/kyrian/EffusiveAnimaAccelerator.ts
```diff
@@ -0,0 +1,98 @@
...
+  [CLASSES.MONK]: SPELLS.WEAPONS_OF_ORDER,
+  [CLASSES.PALADIN]: SPELLS.DIVINE_TOLL,
+  [CLASSES.PRIEST]: SPELLS.BOON_OF_THE_ASCENDED,
+  [CLASSES.ROGUE]: SPELLS.ECHOING_REPRIMAND,
+  [CLASSES.SHAMAN]: SPELLS.VESPER_TOTEM,
+  [CLASSES.WARLOCK]: SPELLS.SCOURING_TITHE,
+  [CLASSES.WARRIOR]: SPELLS.SPEAR_OF_BASTION,
+};
+
+const CDR_PER_APPLICATION = 1 / 15;
```
Pewtro, line 32: If it were me I'd throw in some comments explaining why 1/15 is the correct number here, so a short example or explanation would be good 👍🏼

src/parser/shadowlands/modules/covenants/kyrian/EffusiveAnimaAccelerator.ts
```diff
@@ -0,0 +1,98 @@
...
+    super(options);
+
+    const classId = getClassBySpecId(this.selectedCombatant.specId);
+    this.classAbility = SPELLS_BY_CLASS[classId];
+
+    const active =
+      classId < Infinity &&
+      this.classAbility != null &&
+      this.selectedCombatant.hasCovenant(COVENANTS.KYRIAN.id) &&
+      this.selectedCombatant.hasSoulbind(SOULBINDS.FORGELITE_PRIME_MIKANIKOS.id);
```
Pewtro: If you don't want to move the classId check up above where you use it, then it's pointless to have here. 
Then your classAbility check is enough, and it could be simplified down to the below because you know if classAbility is not undefined, then it's fine and if they have mikanikos then they're kyrian by default
```suggestion
    const active = this.classAbility && this.selectedCombatant.hasSoulbind(SOULBINDS.FORGELITE_PRIME_MIKANIKOS.id);
```

Pewtro (changes requested):

src/parser/shadowlands/modules/covenants/kyrian/EffusiveAnimaAccelerator.ts
mfield: :+1:

Pewtro:
Thanks for this, good work!

Issue #4782, opened 2022-03-03
Enhancement shaman showing 'cheap weapon enhancement'


kedearian:
https://wowanalyzer.com/report/GrdjhBnxXbwqF6vW/13-Normal+Dausegne,+the+Fallen+Oracle+-+Kill+(3:41)/Zeashaman/standard

example, it's showing that windfury/flametongue are cheap enhancements. This should hopefully be a easy fix.

danny-janse:
I believe this issue was resolved back in april. I have yet to run into this issue myself.

ToppleTheNun:
This issue should be able to be closed; the linked report is showing as not cheap.

PR #4837, opened 2022-04-06
Add Flametongue and Windfury weapon to max enhancement list

sorrellp:
This is a fix for issue #4782. 

Enhancement shamans are unable to use weapon oils so flametongue and windfury weapon should count as max weapon enhancements.


emallson (approved):
Thanks for the fix!

PR #4841, opened 2022-04-07
[Warrior][Fury] Fix bad whirlwind fraction suggestion text

AndreasAlbert:
In the current implementation before this pull request, the whirlwind suggestion text makes no sense. For example, see this screenshot, from [here](https://wowanalyzer.com/report/39Gaqg1fN7JdPrvM/1-Normal+Halondrus+the+Reclaimer+-+Kill+(5:39)/Cmb/standard/overview):

image

The text seems to indicate that there are 0% bad casts (which would be great!) and that one should have >90% of bad casts (what?!). This seems to be a simple mixup of using the fraction of good casts = 1 - fraction of bad casts.

In this PR, I suggest to simply flip the fraction in the suggestion text and leave all other parts of the analysis untouched.


emallson (approved):
good catch! thanks!

Issue #4541, opened 2021-06-19
[RFC] Functional analyzers


MartijnHols:
> We did some brainstorming on this on Discord and I hope this RFC will flesh out the idea, get more people behind it and serve as a good design document in case the implementation takes a while.

Glossary

- **Functional**: [This article](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-functional-programming-7f218c68b3a0) does a good job explaining it. In short it can be summarized to be pure functions while avoiding state*.
- **Pure functions**: a function which given the same input always returns the same output and has no side-effect. `(a) => a * 5` is pure, `(a) => a + this.previousValue` is not and neither is `(a) => { window.something = 'something'; return a * 5; }`. When passing objects, you should treat them as immutable and return a new instance. e.g. `(a) => { return { ...a, prop: 1 } }` and not `(a) => { a.prop = 1; return a }`.

*Avoiding state
> Let's recap: state is where you have a variable that remembers the last value of something. In a log this might be the Haste level of someone at the current time, which can be accessed today through `this.haste.current` given a dependency on the `Haste` module.

The main reason for the event parsing setup currently implemented in WoWAnalyzer is that it matches what a player would experience in game. If you're at 2:00 in a boss fight your Haste could be 10%, boss at 88% health, you are affected by the Bloodlust buff, etc. Current we let you read this state through various methods, such as `this.haste.current` for the state. This works pretty well for matching what someone would experience in-game but isn't great for code quality and maintenance.

## What is a functional analyzer?

A fully functional analyzer is a function that **takes events** and **outputs a value** without the usage of any dependencies or mutating any state. The events are passed as an argument; `events: Event[]` which can be reduced; `events.reduce(x)`.

Writing a functional analyzer to sum the total damage of a certain ability would be easy:

```ts
const totalFireballDamage = (events: Event[]) => events.reduce((sum, event) => {
  if (event.type === DamageEventType && event.ability.guid === FIREBALL_ID) {
    sum += event.amount
  }

  return sum
})
```

> We would make composers available for commonly used analysis so you won't have to write this code yourself every time but can instead do `const totalFireballDamage = totalDamage(FIREBALL_ID)`.

This matches the "pure function" definition earlier: given the same events it will always output the same damage done. It does not get affected by the state of anything.

[...]

PR #4848, opened 2022-04-08
Integrate functional analyzers into the class-based analyzer pipeline

emallson:
Related: #4541

This replaces the current implementation of functional analyzers (which is a bespoke addition to the analysis pipeline) with an implementation that generates an anonymous class from a functional analyzer and then passes it through the existing class-based analyzer pipeline.

In the process, I also caught an issue in our integration test system that wasn't triggering the event sorting code that had been added to the functional section of the `CombatLogParser` when running integration tests and thus had to update some snapshots once this was moved to a location that *was* reached during integration tests. The code there was a hack in the first place and should've been done better but :shrug: it wasn't.

As I mentioned [here](https://github.com/WoWAnalyzer/WoWAnalyzer/issues/4541#issuecomment-1087762260), the goal is not to eliminate functional analyzers but to allow them to exist within the existing infrastructure without special-cased support in `CombatLogParser`. This allows us to use & experiment with them without having to worry about ramifications for the core analysis system.

There is still one piece of "special-cased" support: I added the ability to return `Suggestion[]` from the `suggestions()` method, which also applies to class analyzers but is currently not used by any. Could be useful if we want to strip away some of the magic around suggestions.

(The branch name is related to other stuff that I was planning to do at the same time, like eventdispatch -> normalizers, but this ended up being large enough that I wanted to PR it solo)


src/parser/core/Analyzer.ts
```diff
@@ -50,3 +56,93 @@ class Analyzer extends EventSubscriber {
 }
 
 export default Analyzer;
+
+type ConstructedDependency<T> = T extends new (options: Options) => infer R ? R : never;
+
+type InjectedDependencies<Deps extends Dependencies> = {
+  [Key in keyof Deps]: ConstructedDependency<Deps[Key]>;
+};
```
emallson, line 64: also just wanna say: the typescript type system is pretty neat. there are still pretty severe limitations around dynamically generated class properties, but outside of those we can actually do neat things with dependencies.

if `Deps` is the dependency map you'd supply to `static dependencies`, then `InjectedDependencies<Deps>` is a map with each key being a constructed instance of the dependency. of limited type-strength right now due to the aforementioned issues with dynamic class props, but a small step is a step nonetheless.

src/parser/core/Analyzer.ts
```diff
@@ -50,3 +56,93 @@ class Analyzer extends EventSubscriber {
...
+    eventList: AnyEvent[] = [];
+
+    constructor(options: Options) {
+      super(options);
+
+      eventFilters.forEach((filter) => this.addEventListener(filter, this.appendEvent));
+    }
+
+    private appendEvent(event: AnyEvent) {
+      this.eventList.push(event);
```
emallson, line 97: this does make a shallow copy of the event list, which is not ideal. I was lazy with the APL checker's `EventFilter` and left them at the default `EventFilter.any` even though they need something like `[EventFilter.cast, EventFilter.updatespellusable, EventFilter.begincast, ...]`. I need to find or write a function that applies an `EventFilter` to an array of events.

MartijnHols:
I have a review pending, I just need to go over all my comments again to sanity check them when I am les tired. Will do that when it's not night time and I have time. If that takes too long just ping and I'll dump it as-is.

emallson:
> I have a review pending, I just need to go over all my comments again to sanity check them when I am les tired. Will do that when it's not night time and I have time. If that takes too long just ping and I'll dump it as-is.

no rush, this isn't pressing.

PR #4854, opened 2022-04-12
[Enhancement Shaman] Add support for Witch Doctor's Wolf Bones and Fix Maelstrom Weapon Tracking

sorrellp:
Adding support for enhances most popular legendary Witch Doctor's Wolf Bones which decreases the cooldown of Feral Spirit each time a stack of maelstrom is gained. I also updated the logic for tracking the maelstrom weapon stacks as the total amount gained wasn't matching up with warcraftlogs. 

For example using Werdup's log here https://www.warcraftlogs.com/reports/H1zBDXwVnyL4PqZ6#fight=9&type=auras&source=2&ability=344179 we can see he gained 495 maelstrom weapon stacks during this encounter. Viewing the same on WoWAnalyzer shows 296 gained:
image

After my changes this should reflect more accurately:
image

This mainly was caused by the maelstrom weapon module not tracking the maelstrom weapon RefreshBuff events. If the selected player is already at the max stacks on maelstrom weapon a applybuff/applybuffstack event would not be triggered so we need to look for refreshbuff events in that scenario. Another scenario is when a stack of maelstrom weapon is gained during the same timestamp that a spender was used in this case a applybuff/applybuffstack event is also not always fired and we need to look for the refreshbuff event is this scenario as well.


HawkCorrigan:
Looks good at a first glance

Would like someone more active to look over it before merging though.

analysis/shamanenhancement/src/modules/shadowlands/legendaries/WitchDoctorsWolfBones.tsx
```diff
@@ -0,0 +1,90 @@
...
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.OPTIONAL(10)}
+        size="flexible"
+        category={STATISTIC_CATEGORY.ITEMS}
+      >
+        <BoringSpellValueText spellId={SPELLS.WITCH_DOCTORS_WOLF_BONES.id}>
+          <>
+            <CooldownIcon /> {this.totalCdrGained / 1000}s <small> of Feral Spirit CDR</small>
```
abelito75: Should use `formatDuration` here

analysis/shamanenhancement/src/modules/shadowlands/legendaries/WitchDoctorsWolfBones.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <br />
+            <CooldownIcon /> {this.totalCdrWasted / 1000}s{' '}
```
abelito75: Same here

abelito75 (changes requested):

analysis/shamanenhancement/src/modules/shadowlands/legendaries/WitchDoctorsWolfBones.tsx
sorrellp: Both places are using formatDuration now

PR #4855, opened 2022-04-15
[Windwalker Monk] Coordinated Offenive Fixate

oBusk:
So I found it interesting that you need to Fixate to gain the bonus damage from the Coordinated Offensive conduit and I've been staring at WowAnalyzer recently so felt like trying to add some suggestions and felt like making sure you fixate if you have the conduit is probably a good tip?

I managed to cobble together a checklist item for it. Messed around a bit about adding a statistics for it too. Not sure how the suggestions work! 

Please let me know what you think! 

Have a good day!

image
image


oBusk:
I considered to keep track of the average fixate delay. So the time from SEF cast to Fixate and set some thresholds for it. Hypothetically it's plausible to use 2 GCD to spread Mark of the Crane, but any longer than that is kind of throwing away dps. I've gone with a "percentage of SEF time that is fixated" right now.

emallson:
I did a merge in the github ui, it broke, i'll fix it later today :)

oBusk:
I think percentage of fixate makes the most sense to have thresholds for. Average fixate delay feels unintuitive.

I also removed the tracking of "missed damage". While I like the idea of tracking damage you're missing to be able to say exactly how much damage you're missing out on, I'm not sure where it makes sense to show it. So for reference, a earlier iteration looked like this. 

image

If anyone thinks this is a good idea or an idea how to show the data somewhere nice (statistics, checklist or suggestion), let me know.

emallson (approved):

emallson:
> I also removed the tracking of "missed damage". While I like the idea of tracking damage you're missing to be able to say exactly how much damage you're missing out on, I'm not sure where it makes sense to show it. So for reference, a earlier iteration looked like this.

You could have it in the statistic drawer (child elements of the statistic show up in a drawer) if you want to have it accessible but not highlight it. I don't think it *needs* to be there, so I'm happy to merge without it.

oBusk:
I don't think we need to show how much damage someone is missing, since it's hypothetical anyway. 

Somethings that I'd like a second opinion on:

I've added a new item into interface here, "ConduitLink" which uses context to produce correctly scaled links/tooltips of conduits. Does this seem like a bad idea to clutter up the directories more? I just found it so cumbersome to create the link and get the itemlevel or rank every time. But if we're not sure about it I can change the analyzer and component to just create spellinks using rank.

oBusk:
Has some interesting idea for how to track 
https://discord.com/channels/114786249750872069/434767646978998282/965733644696174622

Don't merge yet

emallson:
Converting to a draft because of your "Don't merge yet" comment. when its ready you can convert back

oBusk:
We now track the percentage of the _possible_ damage gain, recommend thresholds for this percentage.

Current thresholds major if below 70%, average if below 80%, minor if below 95%. 

## Checklist
image

## Suggestions
image

## Statistics

I've kept the statistics-box simple but with an informational tooltip 
image

emallson:
@oBusk is this ready for draft to be toggled off and reviewed for a merge?

oBusk:
@abelito75 [mentioned](https://discord.com/channels/316864121536512000/316864385052049420/967110449478922260) that it's not great to have a tab with a single check like in this case. But I can't figure out a better solution. The suggestions was either to put the "fixate" check under the "Use Cooldowns Effectively" section, but without being able to explain, just asking them to fixate seems a bit confusing. The other suggestion was to have a "use SEF effectively" section. But if you don't have CO, that section would only have a single check.

If we're ok with the current UI, this is ready for review again @emallson

emallson (approved):

emallson:
I'm not one to let the perfect stand in the way of the good, so this works for me 👍

PR #4858, opened 2022-04-15
[meta] Force test locale to avoid snapshot mismatches

oBusk:
The tests for WoWAnalyzer/WoWAnalyzer#4856 failed because of snapshot changes. When running locally on my machine that uses Swedish number formatting (`1 000 000`), the snapshots had lots of unnecessary changes where for DPS in statistics and similiar because they [rely on `Number.toLocaleString()`](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/7ec1a9ea1198ad7bb85d9ac2efed5ccfe0bf2201/src/common/format.ts#L6) which resulted in different strings on my machine compared to when the snapshots were created.

I felt like this would be a common problem and tried to figure out a way to force the locale of Node and/or Jest on runtime but couldn't really find any solution to this. Instead I added a separate environment variable called "LOCALE" which let's you force the locale when building/running the app. This can then be overriden when running the tests, using a setupTests script. Then I changed all uses of `toLocaleString()` to send in the potential env.LOCALE. If it's not defined, `toLocaleString()` will just use the users locale as usual, but if it's defined, that locale is used instead. That way we can ensure that when the tests run, we always use `en-US` and get consistent snapshots, whatever machine you run on.


oBusk:
I added a globalSetup and new `jest.setup.js` file, but there is already a [`setupTests.js`](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ee89b88/src/setupTests.js), but I don't really understand what decides that this file is included, so I didn't touch it.

oBusk:
The PR checks failed because I haven't added a changelog entry.. Should there be changelog entries for completely technical changes like this?

emallson:
We do typically do changelog entries for purely technical changes, yea. In this case we could override that check but its up to you.

oBusk:
Seems unnecessary to me. No user should ever notice the change.

abelito75:
It not just a change log for end users IMO

emallson:
The Official Reason:tm: is that it just gives public credit to contributors. Things like the git commit log etc are useful for a history of changes but hidden from the public eye. That's all. If you don't want a changelog entry I'm not fussed about merging it without one.

abelito75:
Lets not forget that the git commit log of a file can be destroyed

emallson:
so can the changelog list :)

oBusk:
I think it's fair enough that the changelog is not only for end-users so I added a changelog entry! :)

emallson (approved):

PR #4863, opened 2022-04-16
Holy power per minute

xizbow1:
If there's a way to just add a changelog for "Paladin" I would love that


abelito75 (approved):
lgtm
Simple but effective. nice

PR #4867, opened 2022-04-18
[Windwalker Monk] Fix tracking of 4set

oBusk:
I realised that I was being dumb and needed to do a bit more to track damage that has been boosted by Primordial Power.

Also updated ww lists to track damage events from chi wave/burst and faeline stomp because I noticed they were missing.


analysis/monkwindwalker/src/modules/tier/PrimordialPotential.tsx
```diff
@@ -54,48 +57,58 @@ class PrimordialPotential extends Analyzer {
...
+          {event.meta.enhancedCastReason}
+          <br />
+          {reason}
+        </>
+      ) : (
+        reason
+      );
+    } else {
+      this.wasLastCastPowered.delete(event.ability.guid);
     }
```
emallson: flip the `if` on this. generally prefer early return over wrapping an entire method in an `if`.

```suggestion
    if (!this.selectedCombatant.hasBuff(SPELLS.PRIMORDIAL_POWER_BUFF.id)) {
      this.wasLastCastPowered.delete(event.ability.guid);
      return;
    }

    this.wasLastCastPowered.add(event.ability.guid);

    this.numberPoweredCasts += 1;

    if (!this.poweredCasts[event.ability.guid]) {
      this.poweredCasts[event.ability.guid] = 0;
    }

    this.poweredCasts[event.ability.guid] += 1;

    event.meta = event.meta || {};
    event.meta.isEnhancedCast = true;
    const reason = (
      <>
        This cast was empowered by <SpellLink id={SPELLS.PRIMORDIAL_POWER_BUFF.id} /> (4-set
        bonus)
      </>
    );
    event.meta.enhancedCastReason = event.meta.enhancedCastReason ? (
      <>
        {event.meta.enhancedCastReason}
        <br />
        {reason}
      </>
    ) : (
      reason
    );
```

analysis/monkwindwalker/src/castDamage.ts
```diff
@@ -0,0 +1,34 @@
...
+  [SPELLS.WHIRLING_DRAGON_PUNCH_TALENT, [SPELLS.WHIRLING_DRAGON_PUNCH_TALENT_TICK]],
+  [SPELLS.SPINNING_CRANE_KICK, [SPELLS.SPINNING_CRANE_KICK_DAMAGE]],
+  [SPELLS.CHI_BURST_TALENT, [SPELLS.CHI_BURST_TALENT_DAMAGE]],
+  [SPELLS.CHI_WAVE_TALENT, [SPELLS.CHI_WAVE_TALENT_DAMAGE]],
+  [SPELLS.RISING_SUN_KICK, [SPELLS.RISING_SUN_KICK_SECOND]],
+  [
+    SPELLS.FAELINE_STOMP_CAST,
+    [SPELLS.FAELINE_STOMP_PULSE_DAMAGE, SPELLS.FAELINE_STOMP_DAMAGE_AND_HEAL],
+  ],
+]);
```
emallson, line 23: since this is internal and you always use the `entries` anyway *and* you already have it written in entries format, should this just be `const entries = [ .... ];`?

Not sure either way. Up to you how you want to do it.

emallson (changes requested):
1 small change + a comment then good to go

analysis/monkwindwalker/src/castDamage.ts
oBusk: Not sure from github interface which element you're referring to..

`castDamageMap` is a `Map` so I can map from object to object.

`entries` is to get an array from the `Map` since the `Iterator` that `Map.entries()` returns does not have `.map()`.

`castToDamage` and `damageToCast` are simple ID->`Spell` dictionaries to easily get `Spell` with all meta data with IDs.

oBusk:
We now track the percentage of the _possible_ damage gain, recommend thresholds for this percentage.

Current thresholds major if below 70%, average if below 80%, minor if below 95%. 

## Checklist
image

## Suggestions
image

## Statistics

I've kept the statistics-box simple but with an informational tooltip 
image

analysis/monkwindwalker/src/castDamage.ts
emallson: so my thinking is that `castDamageMap` is never actually used as a map (it is not exported, and its only use here is in defining the `entries` const).

so why not change it to:


```ts
const entries = [
  [SPELLS.FISTS_OF_FURY_CAST, [SPELLS.FISTS_OF_FURY_DAMAGE]],
  [SPELLS.WHIRLING_DRAGON_PUNCH_TALENT, [SPELLS.WHIRLING_DRAGON_PUNCH_TALENT_TICK]],
  [SPELLS.SPINNING_CRANE_KICK, [SPELLS.SPINNING_CRANE_KICK_DAMAGE]],
  [SPELLS.CHI_BURST_TALENT, [SPELLS.CHI_BURST_TALENT_DAMAGE]],
  [SPELLS.CHI_WAVE_TALENT, [SPELLS.CHI_WAVE_TALENT_DAMAGE]],
  [SPELLS.RISING_SUN_KICK, [SPELLS.RISING_SUN_KICK_SECOND]],
  [
    SPELLS.FAELINE_STOMP_CAST,
    [SPELLS.FAELINE_STOMP_PULSE_DAMAGE, SPELLS.FAELINE_STOMP_DAMAGE_AND_HEAL],
  ],
];
```

and ditch the map entirely?

Then you use `entries` as you currently are for constructing `castToDamage` and `damageToCast`

emallson: this is really a nitpick, not going to block merging over it :) you should be able to click the merge button since I've approved the PR. if not, ping me when you want to merge

emallson (approved):

analysis/monkwindwalker/src/castDamage.ts
oBusk: That's definitely a good point. But I'd maybe argue that a `Map` is a bit of a clearer structure for someone looking at the code, than just an array of arrays and objects? The syntax for creating Maps are a bit awkward, and I really wish I could make the objects in an easier way than going via entries to array... 

I feel like if you look at the code, a `Map`, while somewhat arcane, at least has a clear purpose, while the type `[Spell, Spell[]][]` is kind of nonsense lol.

Originally I created the map and then util methods that took `id` and then `find()` the entries, but that felt like unnecessary looping for each spell.

oBusk: Dropping the Map for just `entries` is a small perf. optimization at dev logic cost, which I don't think is worth it in this case.

emallson: :+1:

PR #4868, opened 2022-04-18
Fix small error about non-boolean properties being boolean value.

oBusk:
* "Recieved 'true' for a non-boolean attribute `colored`. On home page.
  image

  The icons take `colored` as a boolean, but then spread all `props` on the `<svg>` which then puts that boolean colored on the svg, which doesn't support that and throws a debugger error.


oBusk:
No need for a changelog

MartijnHols (approved):
Thanks!

PR #4869, opened 2022-04-18
Refactor report component to functional component

oBusk:
> Original Context:
>
> https://discord.com/channels/316864121536512000/316864385052049420/965443315216580698
> https://discord.com/channels/316864121536512000/316864385052049420/965639628185038868

So the current solution has problems of relying on rendering components which updates the state as a sideeffect, causing react to throw errors. 

On solution is this, which is to engage in the modernisation of the Reports component to be a function component which uses Hooks for state, rather than the state it was previously. This implementation seems to work fine to me, but would love any feedback on it.

I'm mostly unsure about structuring of files etc. I suspect the bigger hooks could have lots of the parsing broken out into methods which fits into `src/parser` somewhere while the hooks, only dealing with calling other hooks to manage state, could stay in `src/interface` somewhere.


oBusk:
The current structure, which as far as I can tell, should always have caused state issues, was introduced in https://github.com/WoWAnalyzer/WoWAnalyzer/commit/68a2b572fc237832062173285ea90efe5ec5f90f.

emallson:
adding myself as a reviewer so I remember, but not getting around to reviewing this tonight

emallson (approved):
discussed in discord. there were some issues loading reports with missing character profiles, and then it was possible to infinite loop the event parser due to how it used async inside `useEffect`. fixed those things, works well

PR #4870, opened 2022-04-19
Proxy SPELLS object to throw when accessing a missing spell

emallson:
This will help avoid things like `EffusiveAnimaAccelerator` ending up in production and sneakily breaking all Kyrian logs.

The changes are:

- When using the `SPELLS.NAME` or `SPELLS[spellId]` forms, an error is thrown if the spell is not found. This error is not intended to be caught.
- If you want to get a spell by `NAME` or `spellId` and expect that the spell may not be present, use `SPELLS.maybeGet(...)` instead. `SPELLS.maybeGet` will never throw.

I spent a good deal of time looking at fixing `SPELLS` to use the correct types instead of `any`, but it just causes so many spurious type errors to do it right because of incorrect/inspecific types used across the codebase. There are a few actual issues in there as well, but they're a minority. Since we want to switch from the global `SPELLS` to spec/class `SPELLS` for 10.0 I just reverted those changes in favor of this 90% solution (which unfortunately still doesn't give autocompletion, but it will catch typos / bad spell names early!)

(the two snapshot changes are not related to this---they were just missing in the `shadowlands` branch)

### DIFF
--- a/analysis/paladinholy/src/integrationTests/example.test.ts.snap
+++ b/analysis/paladinholy/src/integrationTests/example.test.ts.snap
@@ -6354,6 +6354,69 @@ Array [
 ]
 `;
 
+exports[`Holy Paladin integration test: example log HolyPowerPerMinute matches the statistic snapshot 1`] = `
+<div
+  className="col-lg-3 col-md-4 col-sm-6 col-xs-12"
+>
+  <div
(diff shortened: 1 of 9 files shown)


abelito75 (approved):
LGTM

PR #4871, opened 2022-04-20
[Core] Update talent code

abelito75:
This is silly as its two things in one which isn't how PRs should be but I found the vantus issue after commit the talent change and am lazy


src/parser/core/Combatant.ts
```diff
@@ -130,7 +130,7 @@ class Combatant extends Entity {
   }
 
   _getTalent(row: number) {
-    return this._talentsByRow[row];
+    return [...this._talentsByRow][row];
```
emallson: this spread / array construction likely eliminates all of the gain from using a set for the `.has(...)`.

src/parser/core/Combatant.ts
```diff
@@ -167,11 +167,7 @@ class Combatant extends Entity {
     if (spellObj.id) {
       spellId = spellObj.id;
     }
-    return Boolean(
-      Object.keys(this._talentsByRow).find(
-        (row: string) => this._talentsByRow[Number(row)] === spellId,
-      ),
-    );
```
emallson, line 174: if the performance of this is a problem, we should store both `_talentSet: Set<number>` and `_talentsByRow: Record<number, number>` and use them as appropriate---trading memory for runtime.

if we don't want to store it twice, then simplifying this to `return Object.values(this._talentsByRow).find(id => id === spellId) !== undefined;` is likely faster and also simpler code.

emallson (commented):
so i guess I have two questions:

1. is this actually a net performance improvement? `_getTalent` is likely slower this way, while `hasTalent` is faster.
2. how large is the gain on the paths where `hasTalent` is used?

I don't think this implementation of `_getTalent` is particularly good. `hasTalent` likely vastly outweighs it in terms of usage, so this *probably* is a net gain

if performance is an issue here, I do think we just store it twice and move on since the memory footprint is negligible. that said: i also don't think the performance gain is likely to be that big so I'm kinda meh on this

src/parser/core/Combatant.ts
abelito75: This is one thing I feared but this method is almost never called 
If I did my checking correctly the only places that directly call_getTalent is the `lvFooTalent` methods which those are called a grand total of 5 times throughout the code base specifically in spec analyzers.
If anything this set of code `_getTalent` -> `getlvFooTalent` should probably just be deleted all in all

Looking forwards to dragon flight (which is probably not the correct call) since the tree is now not deterministic as before this code would most likely die there.

abelito75:
I will admit that one thing I didn't test was true performance gains on this verse the old method. Which is something I should have done. 

The eyeball test was used (very poor) but we know that `.has` is a O(1) function while `.find` should be O(n)

One thing that this is also doing is preventing performance issues (which was not tested with actual results) in the future for dragon flight but as I said in another comment, this is probably something I shouldn't be trying to optimize for as we don't have data on the log yet.

abelito75:
I'm actually going to remove the talent row stuff right now and just update the few places that use it. 
normalizing the codebase away from these dated calls is probably a smart thing

emallson:
That makes sense to me

abelito75:
Okay. I think I got it all and found a fun edge case for in untyped land for the SPELLS change the other day

Which this time I confirmed the spell issue in prod

emallson (approved):
love it

PR #4873, opened 2022-04-21
Enhancement shaman feral spirit stats

sorrellp:
I was curious how often the two set bonus was proccing for enhancement and what the distribution of the elemental spirits buff was so I decided to add stats for both. 

Elemental Spirits is easy to track just by looking at buffs. 

For the tier set bonus the wolves summoned by the set bonus are summoned under separate ids than when wolves are summoned by the Feral Spirit spell. I added these under the ability section for shaman and added _TIER to the names of the each.

image
image


emallson (approved):
👍 looks good

PR #4874, opened 2022-04-22
[WW Monk] Further fixes to PP tracking

oBusk:
@oBusk is a terrible programmer and should be kept outside of furnished rooms.


oBusk:
* Fixed mistake which caused all the spells where the damage is a different spell than the cast to not be tracked. This is the main reason why the tracking is so low.

  | Before | After
  |:-:|:-:|
  | image | image |
  
* Track missing damage of Fist of the White Tiger.

* Changed how all the Level 15 talents are tracked. The cast is not "Powered", instead the damage is increased if you have the buff, whenever it happens. This probably doesn't affect the amount of damage.

* Added tracking of Jade Ignition explosion "Chi Explosion"

* Updated the copy text. I wanted to clarify that this is the number of casts, and not a breakdown of the damage increase.

  | Before | After |
  |:-:|:-:|
  | image | image |

emallson (approved):

PR #4875, opened 2022-04-22
Changed Shadowmeld to be a minor suggestion only for Feral

kfinch:
also fixed a tooltip


analysis/druidferal/src/modules/racials/Shadowmeld.js
```diff
@@ -29,8 +30,8 @@ class Shadowmeld extends Analyzer {
     return {
       actual: this.correctUses / this.possibleUses,
       isLessThan: {
-        minor: 0.9,
-        average: 0.8,
+        minor: 0.7,
```
abelito75, line 33: Should probably leave these the same? Since it changes the color of the bar in the checklist right?

kfinch: True but it's way too strict towards the current importance of Shadowmeld, I'll change them to be different numbers but lower

PR #4880, opened 2022-04-24
Fix some blood dk rune cd errors

mfield:
Still missing something since I can't get rid of all of them, but there's fewer now.
Getting the precise values on OWS seems to help a bit, so it seems worth to hard code since it's still so prevalent in 9.2.


emallson (approved):

Khazakdk:
@mfield I am a little concerned about the huge swing in the rune efficiency stat on the unholy snapshot.  Did you look into that at all?

mfield:
@Khazakdk you're right i think i see what's up. i'll open another branch soon with a fix.

PR #4882, opened 2022-04-26
[DK] eternal hunger and withering plague conduits

mfield:
I could pull out this pet tracking somewhere... not sure where it would go though.


emallson (approved):

analysis/deathknight/src/shadowlands/conduits/EternalHunger.tsx
```diff
@@ -0,0 +1,73 @@
...
+    super(options);
+    this.conduitRank = this.selectedCombatant.conduitRankBySpellID(SPELLS.ETERNAL_HUNGER.id);
+
+    if (!this.conduitRank) {
+      this.active = false;
+      return;
+    }
+    this.increase = ETERNAL_HUNGER_BASE_INCREASE + this.conduitRank * ETERNAL_HUNGER_PER_RANK;
+
+    // Can't filter this .by(SELECTED_PLAYER_PET) since that only includes pets present
```
Khazakdk, line 30: I know this module was made for blood, but I think you need a listener for .by(SELECTED_PLAYER_PET) in conjunction with your summons tracking here.  The permanent unholy pet does not have a summon event in normal gameplay and so its damage will never be picked up.

mfield: There's 2 conditions in onDamage: `this.isPet(event) || this.owner.byPlayerPet(event)` the second one should account for permanent pets. I looked at a couple unholy logs and they're calculating about 3x as much dps from the conduit as my Blood logs are, so gut check passes. Did you have a log you were worried about?

Khazakdk: I did some more looking into the numbers and EH is just not as good at higher ranks as I thought it was.  It was coming out at 5-6% on unholy logs.  Comparing logs with 5 minute kill times to sims shows a similar percentage, close enough that the permanent pet would have to be picked up by wowa.

PR #4884, opened 2022-04-27
fix rune regen tracking from #4880

mfield:
Mostly a revert. analysis/deathknightblood/src/modules/Abilities.js is the thing i was missing before, so having the acceleration thing in SpellUsable was causing Runic Corruption to get 4x instead of 2x.


emallson (approved):

Khazakdk:
Thanks!  Unholy snapshot rune usage is back to the original values.

PR #4888, opened 2022-04-29
[Monk] Dampen Harm Evaluation

renanthera:
After substantial amounts of pain and spamming abe some *really* dumb questions, I have most of a Dampen Harm value evaluator implemented.

I (we) ran into a roadblock though with maximum HP tracking, though. The current solution is to skip any events that omit the maxHitPoints field, such as SPELL_ABSORBED events. In the attached sample log, that leads to a discrepancy of approximately 4k and 7 skipped events.

https://www.warcraftlogs.com/reports/YzTmv7yjdNXaBfQM#fight=2&type=damage-taken&source=5&view=events&pins=2%24Off%24%23244F4B%24damage%24-1%240.0.0.Any%240.0.0.Any%24true%24230908150.0.0.Monk%24true%24365853%7C366933%7C366849%7C361816%7C361818%7C368660%7C362407%7C1%7C363020%7C368428%7C362394%7C362545%7C362772%7C362773%24and%24auras-gained%24-1%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24122278%24true%24false


analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,59 @@
...
+    super(options);
+    if (!this.selectedCombatant.hasTalent(SPELLS.DAMPEN_HARM_TALENT.id)) {
+      this.active = false;
+      return;
+    }
+    this.addEventListener(Events.damage.to(SELECTED_PLAYER), this.damageReduction);
+  }
+
+  damageReduction(event: DamageEvent) {
+    if (event.ability.guid !== SPELLS.STAGGER_TAKEN.id) {
```
abelito75: Nested If statements are nasty. You should change this to 
```
if(event.ability.guid === SPELL.STAGGER_TAKEN.id) {
   return;
}
if(this.selectedCombatant.hasBuff(...){...}
```

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,59 @@
...
+      if (this.selectedCombatant.hasBuff(SPELLS.DAMPEN_HARM_TALENT.id)) {
+        this.hitsReduced += 1;
+        const h = event.amount || 0;
+        const a = event.absorbed || 0;
+        const o = event.overkill || 0;
+        // TODO: ACTUAL MAX HP TRACKING
+        // event.maxHitPoints is insufficient as many events such as SPELL_ABSORBED are missing it and Keg of the Heavens is cursed.
+        const maxHP = event.maxHitPoints || 0;
```
abelito75: This should probably be caught at the top of the function with an if like
```
if (!event.maxHitPoints) {
    return;
}
```

abelito75 (changes requested):
Some small readability changes

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  damageReduction(event: DamageEvent) {
+    if (event.ability.guid !== SPELLS.STAGGER_TAKEN.id) {
+      if (this.selectedCombatant.hasBuff(SPELLS.DAMPEN_HARM_TALENT.id)) {
+        this.hitsReduced += 1;
+        const h = event.amount || 0;
+        const a = event.absorbed || 0;
+        const o = event.overkill || 0;
+        // TODO: ACTUAL MAX HP TRACKING
```
abelito75: You can kind of sneakily do this by tracking all TO damage/healing events. you can find some niche instances where it goes HP buff->full absorb where you won't have updated info but that shouldn't happen all the time TBH... especially since if youre the brew with 4set you should be having happening to you all the time

renanthera:
Latest commit causes it to not skip any DH dr'd hits, but uses the last known max HP. In this case, filtering only events during DH is undesirable under the situation that the first hit in a DH window is missing max HP information. If you were to have a max hp event, max hp update, then DH dr'd event without max HP information, it will be wrong until the next max hp event occurs.

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,73 @@
...
+  skip = 0;
+  currentMaxHP = 0;
+
+  constructor(options: Options) {
+    super(options);
+    if (!this.selectedCombatant.hasTalent(SPELLS.DAMPEN_HARM_TALENT.id)) {
+      this.active = false;
+      return;
+    }
+    this.addEventListener(Events.damage.to(SELECTED_PLAYER), this.updateMaxHP);
```
abelito75, line 26: If we want higher coverage of this we should add `Events.heal.to` to the list
Up to you if you want to add this or not (it should make it more accurate*)

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,73 @@
...
+      if (hitSize >= maxHP / 2) {
+        drdh = 0.5;
+      } else {
+        drdh = 0.6 - 0.5 * Math.sqrt(0.64 - (6 * hitSize) / (5 * maxHP));
+      }
+      this.damageReduced += hitSize / (1 - drdh) - hitSize;
+    }
+  }
+
+  updateMaxHP(event: any) {
```
abelito75: This parameter should have a type. You can do things like `event: DamageEvent | HealEvent | ResourceChangeEvent | DrainEvent ` to encompass all types 

I will require a type here before merging

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Events, { DamageEvent } from 'parser/core/Events';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+
+class DampenHarm extends Analyzer {
+  hitsReduced = 0;
+  damageReduced = 0;
+  skip = 0;
```
abelito75: I don't believe this is used at all?
If not remove please

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,73 @@
...
+      return;
+    }
+    if (this.selectedCombatant.hasBuff(SPELLS.DAMPEN_HARM_TALENT.id)) {
+      this.hitsReduced += 1;
+      const h = event.amount || 0;
+      const a = event.absorbed || 0;
+      const o = event.overkill || 0;
+      // TODO: ACTUAL MAX HP TRACKING
+      // event.maxHitPoints is insufficient as many events such as SPELL_ABSORBED are missing it and Keg of the Heavens is cursed.
+      const maxHP = event.maxHitPoints || this.currentMaxHP;
```
abelito75: maxHP has a chance to equal 0 if the first damage event to the player is fully absorbed (and.. the player didn't receive any damage or healing before hand)
A simple `if (maxHP === 0) {return;}` will catch this edge case. 

It might seem minor or unlikely to happen but right when you don't code an edge case it decides to poke its nasty head out and fuck your code up

abelito75 (changes requested):
Almost there. Looks great though. Thanks a ton of this module!

analysis/monk/src/DampenHarm.tsx
renanthera: Done

analysis/monk/src/DampenHarm.tsx
renanthera: Done, was a relic of counting skipped hits with the previous version.

analysis/monk/src/DampenHarm.tsx
renanthera: Done.

analysis/monk/src/DampenHarm.tsx
```diff
@@ -0,0 +1,80 @@
...
+  constructor(options: Options) {
+    super(options);
+    if (!this.selectedCombatant.hasTalent(SPELLS.DAMPEN_HARM_TALENT.id)) {
+      this.active = false;
+      return;
+    }
+    this.addEventListener(Events.damage.to(SELECTED_PLAYER), this.updateMaxHP);
+    this.addEventListener(Events.resourcechange.by(SELECTED_PLAYER), this.updateMaxHP);
+    this.addEventListener(Events.drain.by(SELECTED_PLAYER), this.updateMaxHP);
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.updateMaxHP);
```
abelito75: iirc this should be .to not .by
As the maxHP of the event would be of the target not the source (which while you're not doing a ton of external healing it can mess things up) 
This way you will also capture the maxHP due to things like druid hots, paladin beacon etc etc
More events with the data

abelito75 (changes requested):
I think this is the last thing and we are ready to rock. Thanks a ton again

analysis/monk/src/DampenHarm.tsx
renanthera: Drain also needed to be swapped to "to", which I also did. You know, in case you run WoWA on a TBC dungeon which still has mana burn and stuff...?

PR #4890, opened 2022-04-29
Fixes to Player Info, Encounter Stats and Character Parses page

oBusk:
Saw some issues on the "PlayerInfo"/"EncouterStats" part of the parse. aka the "Character" tab

Fixed trinkets that are not in hardcoded files having broken icons

| Before | After | 
|:-:|:-:|
| image | image |

Fix conduits not having working links. I expect it's intended to show the name of the conduit next to it, instead of just icons like it was. Bonus: now also shows tooltip/link to the correct rank of conduit.

| Before | After |
|:-:|:-:|
| image | image |
| image | image |

Also fixes broken trinket icon and legendary link/icon on character parses page

| Before | After |
|:-:|:-:|
| image | image |
| image | image |

### Bonus

I was seeing errors about duplicate `key=2544` and managed to track it down to the fact that two different bosses was rendering with id 2544. I found that the jailer encounter had the wrong id, the same as Prototype Pantheon. Not sure where these IDs are used, but feels like it could/should manifest in more places than just rendering the boss selection list on the character parses page.

### DIFF
--- a/src/game/raids/sepulcher/Jailer.ts
+++ b/src/game/raids/sepulcher/Jailer.ts
@@ -3,7 +3,7 @@ import { Boss } from 'game/raids';
 import Background from './images/Jailer.jpg';
 
 const Jailer: Boss = {
-  id: 2544,
+  id: 2537,
   name: 'The Jailer, Zovaal',
   background: Background,
   icon: 'inv_achievement_raid_progenitorraid_jailer',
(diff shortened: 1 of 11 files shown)


oBusk:
I think there's a bug in the fetching of legendary icons. If both spell icons are in the local cache, the state will not update and the component will not render them. Will push fix in  a bit

oBusk:
I realised the API response already have the icon information? So to me it looks like the API request to figure it out seems unnecessary? I've ripped all that code out for now. Maybe there's some corner case where it's needed, let me know and I'll revert those changes to only use API when the icon is missing.

src/game/raids/sepulcher/Jailer.ts
```diff
@@ -3,7 +3,7 @@ import { Boss } from 'game/raids';
 import Background from './images/Jailer.jpg';
 
 const Jailer: Boss = {
-  id: 2544,
+  id: 2537,
```
emallson, line 6: good catch :+1:

emallson (approved):
pulled up the changes locally, looks good :+1:

PR #4894, opened 2022-05-02
[Core] Event Type Check

abelito75:
This will cause your dev environment to break if it detects event types it doesn't know (forces standardization) but does nothing besides a console message in prod


src/interface/report/hooks/useEventParser.ts
```diff
@@ -121,6 +121,23 @@ const useEventParser = ({
...
       //sort now normalized events to avoid new fabricated events like "prepull" casts etc being in incorrect order with casts "kept" from before the filter
       .sort((a, b) => a.timestamp - b.timestamp);
+
+    // verify events
+    const types = new Set<string>(Object.values(EventType));
+    // cycle through events
+    result.forEach((event) => {
+      // do we have the type?
+      if (!types.has(event.type)) {
+        if (process.env.NODE_ENV === 'production') {
```
emallson: imho we should gate this whole loop behind the `!== production` flag. don't forsee us really using this check in production but it'll loop over all events every time. i guess its not a big perf cost, but still

abelito75: I was thinking of that as well but this is similar (IMO) to the spell check you did a few weeks ago so I followed the logic there
The total logic behind that decision might be wrong though

abelito75: I think I will go back and change it since while its very minor for performance in prod it isn't logical to be there

emallson: true. there is added cost to every `SPELLS` index with the change I made, and there are enough `SPELLS.whatever` accesses flying around that that adds up. i'll probably pull this branch in the morning and do a quick sanity check to see how long it actually takes to run it so we know whether these kinds of things are worth optimizing out in prod

abelito75: If we desire performance gains I think one thing we should look at is time lost due to console logging while in prod
I don't think the vast majority of messages/warnings/errors produce in the console in prod are used... even by us developers

PR #4898, opened 2022-05-03
added tooltip

oisinod:
just adding a tooltip to explain the hps values for the conduit


analysis/priestdiscipline/src/modules/shadowlands/conduits/SwiftPenitence.tsx
```diff
@@ -118,6 +119,13 @@ class SwiftPenitence extends Analyzer {
         position={STATISTIC_ORDER.OPTIONAL(13)}
         size="flexible"
         category={STATISTIC_CATEGORY.COVENANTS}
+        tooltip={
+          <>
+            The top value represents healing from <SpellLink id={SPELLS.ATONEMENT_BUFF.id} />, while
+            the lower value is the healing from directly healing a target with{' '}
+            <SpellLink id={SPELLS.PENANCE_HEAL.id} />.
+          </>
+        }
```
emallson, line 128: this should probably be expressed in the statistic itself, but the tooltip is better than it'd been so I'm going to go ahead and merge.

emallson (approved):

PR #4900, opened 2022-05-03
Fade dr

oisinod:
adding this conduit to show dr values


analysis/priest/src/TranslucentImage.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(options: Options) {
+    super(options);
+    this.conduitRank = this.selectedCombatant.conduitRankBySpellID(SPELLS.TRANSLUCENT_IMAGE.id);
+
+    if (!this.conduitRank) {
+      this.active = false;
+      return;
+    }
+    this.conduitIncrease = TRANSLUCENT_IMAGE_INCREASE[this.conduitRank];
```
emallson: this should use `parser/core/conduitScaling` with the appropriate rank 1 value (which based on the other file is i think `6`?) unless the scaling doesn't follow regular conduit scaling (`rankOneValue * (1 + (rank - 1) * 0.1)`)

analysis/priestdiscipline/src/constants.ts
```diff
@@ -138,3 +138,21 @@ export const SWIFT_PENITENCE_INCREASE = [
...
+  9,
+  9.6,
+  10.2,
+  10.8,
+  11.4,
+  12,
+  12.6,
+  13.2,
+  13.8,
+];
```
emallson: this is unnecessary if you use `conduitScaling`

emallson (changes requested):
small change for conduit scaling.

i also merged the main branch in to catch it up with some other changes

oisinod:
should have all those changes addressed

emallson (approved):

PR #4902, opened 2022-05-04
added mana saved to Penitent one legendary

oisinod:
mana value for this leggo now available


emallson (approved):

emallson (commented):
actually, does this work correctly if the TPO buff expires naturally due to duration? seems like it would count the mana saving even though no penance was cast?

emallson (approved):

PR #4908, opened 2022-05-05
feat: add blood-spattered scale & codex of the first technique tracking

ljosberinn:
image
practically this (except no uptime) for both. scale also tracks uses / missed casts

### DIFF
--- a/src/parser/shadowlands/modules/items/dungeons/CodexOfTheFirstTechnique.tsx
+++ b/src/parser/shadowlands/modules/items/dungeons/CodexOfTheFirstTechnique.tsx
@@ -1,0 +1,85 @@
+import ITEMS from 'common/ITEMS';
+import SPELLS from 'common/SPELLS';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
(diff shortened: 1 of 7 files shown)


src/parser/shadowlands/modules/items/dungeons/CodexOfTheFirstTechnique.tsx
```diff
@@ -0,0 +1,85 @@
...
+    (options.buffs as Buffs).add({
+      spellId: SPELLS.CODEX_OF_THE_FIRST_TECHNIQUE_HEALING.id,
+      timelineHighlight: true,
+    });
+
+    (options.abilities as Abilities).add({
+      spell: SPELLS.CODEX_OF_THE_FIRST_TECHNIQUE_HEALING.id,
+      category: Abilities.SPELL_CATEGORIES.ITEMS,
+      gcd: null,
+    });
```
emallson, line 41: this *might* add the trinket to the abilities list on the stats page, but i'm not sure. gonna merge and find out bc its such a low-impact error if that is the case

emallson (approved):

PR #4909, opened 2022-05-06
feat: support ascended vigor

ljosberinn:
has one issue: its indiscernible whether the healing increase is actually applied to your own healing (tanks, leech, etc.) or to healing you receive from any external source (Urh, healers, etc.). 

because of that I've added this `displayPercentage` boolean to `ItemHealingDone` because it would always compare against your personal healing done and thus show misleading info. in my screenshot, approx 80% of the healing comes from myself. 

happy to adjust if theres a way to separate these.

image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -58,6 +58,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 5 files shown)


emallson (approved):

emallson:
the `displayPercentage` is maybe not a 100% optimal way of working around that, but i don't have better ideas so it stands

PR #4910, opened 2022-05-07
[Core] Condensing Math functions into one file

abelito75:
There were a handful of files that contained 1 function being exported as default. I don't really see a reason for this so i put them into 1 file. The idea is any generic function can go here.

File name needs to be changed. Current idea is like `EventBoostLib` (its very java-y)

But the idea of core functions being grouped in 1 location instead of 9 files is important to me


emallson (approved):
:+1: this also helps with the annoyance from auto-import not totally working rn, one easier-to-remember import instead of 9 imports

emallson:
@abelito75 i'm gonna kick this one back to you for one more tweak since I merged stuff into it that used the methods you moved.

abelito75:
You are a demon

PR #4913, opened 2022-05-09
Add extraattacks event for windfury totem etc.

oBusk:
I ran into an un-defined eventtype so I added it.


abelito75:
C H A N G E L O G

Dambroda (approved):

PR #4916, opened 2022-05-10
fixing penance bolts for trinkets/legendary interactions

oisinod:


analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -7,33 +7,16 @@ import BoringSpellValue from 'parser/ui/BoringSpellValue';
...
+  _missedBolts = 0;
+  _casts = 0;
 
   constructor(options) {
     super(options);
 
-    // Castigation Penance bolt count to 4 (from 3)
-    this._boltCount = this.selectedCombatant.hasTalent(SPELLS.CASTIGATION_TALENT.id) ? 4 : 3;
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.PENANCE_CAST), this.onCast);
     this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDamage);
```
abelito75: we can add `.spell(SPELLS.PENANCE)` to this event listener to filter out all non penance damage events

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -7,33 +7,16 @@ import BoringSpellValue from 'parser/ui/BoringSpellValue';
...
     this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.onHeal);
```
abelito75: We can add `.spell(SPELLS.PENANCE_HEAL)` to this event listener to filter out non penance healing events

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -48,19 +31,35 @@ class Penance extends Analyzer {
...
 
-    this.eventGrouper.processEvent(event);
-
-    event.penanceBoltNumber = this.currentBoltNumber;
+    event.penanceBoltNumber = this._boltCount;
+    this._boltCount += 1;
   }
 
   onHeal(event) {
     if (!Penance.isPenance(event.ability.guid)) {
```
abelito75: If we add the `.spells` to the eventListeners we can remove this if statement

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -48,19 +31,35 @@ class Penance extends Analyzer {
       return;
```
abelito75: If we add the `.spells` to the eventListeners we can remove this if statement

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -48,19 +31,35 @@ class Penance extends Analyzer {
...
+  onCast(event) {
+    this._casts += 1;
+    this._missedBolts += this._expectedBoltCount - this._boltCount;
+    this.calculateExpectedBolts(event);
+    this._boltCount = 0;
+  }
 
-    event.penanceBoltNumber = this.currentBoltNumber;
+  calculateExpectedBolts(cast = null) {
+    this._expectedBoltCount = this.selectedCombatant.hasTalent(SPELLS.CASTIGATION_TALENT.id)
```
abelito75: This should be done in the constructor. Talents can't change mid combat so this doesn't need to be done more than once

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -48,19 +31,35 @@ class Penance extends Analyzer {
...
+
+  onCast(event) {
+    this._casts += 1;
+    this._missedBolts += this._expectedBoltCount - this._boltCount;
+    this.calculateExpectedBolts(event);
+    this._boltCount = 0;
+  }
 
-    event.penanceBoltNumber = this.currentBoltNumber;
+  calculateExpectedBolts(cast = null) {
```
abelito75: Only one function calls this and both are pretty simple in terms of total lines of code so we should probably inline this. 
IE moving the `if(this.selectedCombatant.hasBuff...` into this function

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -48,19 +31,35 @@ class Penance extends Analyzer {
...
     }
 
-    this.eventGrouper.processEvent(event);
+    event.penanceBoltNumber = this._boltCount;
+    this._boltCount += 1;
+  }
+
+  onCast(event) {
+    this._casts += 1;
+    this._missedBolts += this._expectedBoltCount - this._boltCount;
```
abelito75: If you determine missed bolts of the last cast when the new cast happens then you run into the issue of never checking the last cast of the fight

analysis/priestdiscipline/src/modules/spells/Penance.js
```diff
@@ -7,33 +7,16 @@ import BoringSpellValue from 'parser/ui/BoringSpellValue';
...
-  get currentBoltNumber() {
-    return [...this.eventGrouper].slice(-1)[0].length - 1; // -1 here for legacy code
-  }
-
-  _boltCount = 3;
-  hits = 0;
-  eventGrouper = new EventGrouper(PENANCE_MINIMUM_RECAST_TIME);
+  _expectedBoltCount = 0;
+  _boltCount = 0;
+  _missedBolts = 0;
```
abelito75: I'd suggest having a `totalExpectedBolts` which would be the total number of bolts expected on a fight and then a `totalBoltHits` which would be the actual bolt hits counted in a log
This means we can derive the total missed bolts by doing `totalExpectedBolts - totalBoltHits`

abelito75 (changes requested):

PR #4924, opened 2022-05-13
remove dollar sign typo

mireie:
simple typo correction, updated spec. Not sure a changelog is necessary at this point.


emallson (approved):
This looks good! If you can add a changelog entry, then we can get it merged

PR #4934, opened 2022-05-21
[Core] Non timebased buff tracker

abelito75:


emallson:
This looks sane to me. Not convinced we should merge w/o first actually using it in an analyzer to make sure it doesn't come with problems/quirks of its own

abelito75:
I made it without attaching anything to it so it wouldn't be bogged down or held up by any analyzer slog.
I can totally tie it to one real quick if desired... 
I might also expand on this a bit to make it so toy with buffs from different combatants n what not tracked too

emallson:
my concern is that we try to use this and then realize that we're just trading one ordering issue for another

emallson:
Thinking about it, having the ability to choose your trade-offs is probably fine but i do want this to be used somewhere before it is merged so that we don't end up with (another) analysis tool that just rots in the codebase

emallson:
Closing for now until we find a place to use it. Keep it in your back pocket @abelito75

PR #4937, opened 2022-05-26
Earthbreakers impact

oBusk:
First of, adds Earthbreaker's Impact (https://www.wowhead.com/item=188264) as an Ability and Buff to show up in timeline.

image

Also shows stats in statistics for Earthbreaker's Impact

image

Damage includes a breakdown of the damage 

image

Also just shows an explanation what the average weak points mean.

image


src/parser/shadowlands/modules/items/raid/sepulcherofthefirstones/EarthbreakersImpact.tsx
```diff
@@ -0,0 +1,329 @@
...
+  getStats() {
+    const stats = this.sessions
+      // Summarize stats for each session
+      .map((session) => {
+        const timestamps = Object.keys(session).map(Number);
+
+        const tickTimestamps = findPeroidicNumbers(timestamps, TICK_MS);
+        const weakPointTimestamps = timestamps.filter(
+          (timestamp) => !tickTimestamps.includes(timestamp),
+        );
```
emallson, line 210: so if i'm understanding this right, the reason this is done is because the weak points are the same spell id as the regular ticks? and you want to separate them?

oBusk: Yeah exactly. It's all damage with the same id. So I try to find the timestamps that is every 2 seconds so that I can identify the weak points.

PR #4947, opened 2022-06-06
Fix scalding brew module not working.

oBusk:
The enemies modules was updated to track instances of enemies, which ScaldingBrew was not updated to handle.

| Before | After |
|-|-|
|image | image |
| image | image |

### DIFF
--- a/analysis/monkbrewmaster/src/CHANGELOG.tsx
+++ b/analysis/monkbrewmaster/src/CHANGELOG.tsx
@@ -17,6 +17,7 @@ import { SpellLink } from 'interface';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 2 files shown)


abelito75 (approved):
Good catch. I bet there are more of these out there

oBusk:
@abelito75 I felt like the `enemies` array probably should not be public since it requires understanding of how the keys are built, and that the enemies module should just expose methods that can be called to get enemies. I only found [one other](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/a7f9b23/analysis/paladinprotection/src/modules/talents/BlessedHammerDamageReduction.tsx#L40) module that used the enemies array directly, so I intended to change that and make the array private to avoid this in the future, but then I realised that the blessed hammer looks at the source rather than the target... so then I considered if the enemies module should have methods for getting target/source from events, but then I don't know the expectations of the `Entities` classes so then I chickened out 🙃

PR #4948, opened 2022-06-06
Remove outdated info

MartijnHols:
Since I stopped playing a long time ago, none of it was useful anymore.


emallson (approved):
😢 but approved

PR #4951, opened 2022-06-10
[Healer] Triage Heal now works on small time windows

abelito75:
Before the triage heal tab only worked on singular heal events which would make certain spells that proc other spells not count. 
For instance whenever a mistweaver hits vivify they will also proc gust of mist on the target but as 2 different events
So the log would look like
Vivify cast successful on Abelito75
Vivify healed Abelito75 for 200
Gust of Mist healed Abelito75 for 200

Which means the triage heal tab would count them as two different events so even effective triage healing isn't accurately counted
IE
player at 10% HP
Threshold to show up is 15% hp from heal
Vivify heals for 10%
Gust of Mist heals for 10% 

That Vivify cast effectively did 20% of their hp but since its two events the tab would say `nope both of these 10% events are too low`

Old: 
image

New: 
image

Duplicate spells are because multiple hit them at the same time. I believe that is actually helpful to see here

`import Slider from 'rc-slider';` throws a `Could not find a declaration file for module 'rc-slider'.` error and I do not know how to fix it. Please help.


abelito75:
DO NOT MERGE till we fix the 
```
/* eslint-disable */
// @ts-ignore
import Slider from 'rc-slider';
/* eslint-enable */
```
Issue

src/parser/shared/modules/features/LowHealthHealing/Component.tsx
```diff
@@ -1,30 +1,128 @@
...
+        array = [];
+        comboMappy.set(key, array);
+      }
+
+      for (let outer = 0; outer < value.length; outer += 1) {
+        const outerEvent = value[outer];
+
+        const currentTimeStamp = outerEvent.timestamp;
+        // small buffer for log weirdness far too low for another cast
+        const upperRange = currentTimeStamp + 100;
```
kfinch: can you make this 100 ms buffer a constant at top of file just so it's clear at a glance what the buffer is

PR #4952, opened 2022-06-11
Reduced prio

xizbow1:
I have received lots of feedback from people in HOF guilds and others that the recommended CPM of a spells in way too high, with most of them directed at judgment. This is likely due to the fact that hpal is much more gcd-locked playing necrolord, maraads, and (sometimes) beacon of virtue, leading to not many globals to press things like judge.


abelito75 (approved):
lgtm

PR #4956, opened 2022-06-16
Added timeline and simple stat box support for The First Sigil. Added timeline item for Ravenous Frenzy buff.

kfinch:
Makes the timeline a little busy for these CD popping Balance Druids - but it was requested for analysis so here it is. Also added Balance Incarn to buff tracking because it was missing even though Celestial Alignment was there...

image


kfinch:
Discovered I somehow ended up with two slightly different contributor entries, so I fixed that too

PR #4962, opened 2022-06-24
added soothing shade

oisinod:


emallson (approved):
kind of awkward to have a soulbind trait be a spec module, but mastery is...special so its probably fine.

PR #4963, opened 2022-06-25
added general draven effects for disc, also marked disc as up to date

oisinod:


analysis/priestdiscipline/src/CHANGELOG.tsx
```diff
@@ -5,6 +5,7 @@ import { SpellLink } from 'interface';
 
 
 export default [
+  change(date(2022,25,6), <>General Draven HPS increases added</>, Hana),
```
kfinch: Date format should be year/month/day - apparently it just adds up all the months and renders this as being in 2024 LOL 
image

kfinch (changes requested):
I tested this out and it looks good. Just need to fix the date format in changelog.

kfinch (approved):

PR #4964, opened 2022-06-26
Brewmaster Guide Prototype

emallson:
<details>
<summary>Mythic Jailer Wipe</summary>

</details>

<details open>
<summary>LFR Jailer Wipe</summary>

</details>

<details>
<summary>Another Mythic Jailer Wipe</summary>

</details>

# Overview

This PR is the product of discussions in `#maintainers`, primarily with @kfinch. The core concept is a rethinking of the summary/checklist/overview page---moving away from abstracted, aggregated metrics into identifying concrete individual problems and examining the details of individualized cooldown uses as a means to deal with the long-standing problem we have with context.

I could write a ton on the philosophy behind this, but the tl;dr is that there are a couple central goals:

1. Refocus the summary page on individual moments in the fight where we can (hopefully) be confidently correct
2. Put more details & context on the summary page than is currently presented by the checklist/suggestions

In some ways, this moves WoWA more towards an "interactive guide" but the core conceit is still using log data to analyze the performance of spells/talents/items (for theorycrafters) as well as individual performance (for players).

This is currently hidden behind a "View Prototype" link on the existing summary page. It is non-invasive and (should be) non-breaking.

The goal right now is to get something live so that I can show it to people and collect feedback.

# The Brewmaster "Guide"

The images above show what is currently implemented for Brewmaster. It is not finished, but should show the what kind of analysis we can include here.

- The **Shuffle** subsection shows a breakdown not only of how many hits were mitigated by shuffle, but also _which_ hits were unmitigated, as well as a chart highlighting the most serious problems seen. The current checklist only shows the number of hits mitigated.

- The **Purifying Brew** subsection shows both the number of charges used (compared against possible charges used), as well as the (likely) reason for each cast. For casts where no reason was determined, it flags a problem. It also identifies gaps where there are possible purifies and plots out how much you could've mitigated if you'd pushed your buttons more.

- The **Invoke Niuzao** section provides a breakdown of Invoke usage. This cooldown is really difficult to analyze at the fight level because you have to use it when you're actually tanking _and_ it is worthwhile to lose overall casts in order to get casts while you're taking more damage. This shows a list of each cast (Weapons of Order gives you 1/2 of an Invoke, which is why it is listed here), as well as a basic checklist you should hit in order to get good usage and a chart detailing your Stagger level, Purifying Brew usage, and Stomp damage.

# Review Tips

- There are not a lot of core changes. The `Overview` and core `CombatLogParser` code is it (aside from random bugfixes that I found along the way)
- There is some shared guide-related code in the `interface/guide` folder. This is intended for use across specs.
- The Brewmaster-specific code is all under `analysis/monkbrewmaster`. There are a few things here that will likely move to `interface/guide` before all is said and done, but not all of it is ready for that yet.

# Prototyping a Guide for Your Spec

1. Write a React component that has the type `Guide<typeof YourCombatLogParser>`, where `YourCombatLogParser` is the CLP class for your spec. The spec modules, `info` map, and event list are provided as props.
2. add `static guide = Guide` to your `CombatLogParser` class.

There are very few docs and the current implementation still has a bunch of rough edges.  There are `Section` / `SubSection` and `ProblemList` components that can be used by anyone, but a number of other things (like the cooldown list) haven't been generalized yet due to time constraints.

# Next Steps

tl;dr of my thoughts on what will happen with this after deployment:

- Show this around to BrM people and get feedback.
- Iterate implementation and direction of this.
- Build a library of plug-and-play components for maintainers to rapidly build out spec support.

## Removing This

If we decide this isn't the way we want to go, it will be relatively easy to back out the changes by simply deleting all the guide files, then reverting the changes to the base CLP and common Overview component.


kfinch:
Overall love it - don't have time to do a full review just yet. I did load this on my local machine and put in a few of my guild main tank's logs so that I could play around with it live instead of just looking at the pictures.

I have a few points of fairly small feedback specific to your Brewmaster prototype - I'll come back later with more general feedback.
* Shuffle 'problem point' graph is missing a label on the y-axis. I assume based on context it's player health, but it should still have a label.
* Purifying Brew 'problem point' graph suggests the player should purify death 🤣
just-purify-death-bro
* The green and red dots on the 'problem point' graph are a little too small IMO - it's difficult to mouse over them for a tooltip.
* For the shuffled hits timeline, things get a little wonky when there are a lot of hits, especially with the mouse-overs. Probably not a big deal though
many-shuffles

Overall this was already helpful for me because I've always found this tank hard to heal and the Guide revealed that he's not using Purifying Brew nearly enough.

kfinch (commented):
Ok I was able to look over more of what's happening - and I had a head start because of course I already looked it all over when building my Resto Druid guide prototype.

From a code perspective, I don't really have any complaints and want this in mainline ASAP so I can put my own prototypes live. However, I have two big requests before this goes in: documentation and generalization. 

I'd like to see every new module in core get documented such that a developer who wants to use the module could read the docs and have a reasonable idea on how to proceed. `Problem.tsx` is particularly in need of this - from running your prototype I see what it does, and looking at the code it's clear that the 'Problem' code is building it, but I'd probably have spend half an hour scrutinizing your implementation in the BrM guide to figure out exactly how it's working. 

Generalization - I think the components you built that aren't directly tied to BrM should go in core, even if you don't see an immediate need for them. I'm thinking `HitsList` `MicroTimeline` and `ProblemList` here - I can already see exactly how I'd use them in a Guardian module because Ironfur is so similar to Shuffle in how you want to play with it.

emallson:
> Shuffle 'problem point' graph is missing a label on the y-axis. I assume based on context it's player health, but it should still have a label.

fixed :)

> Purifying Brew 'problem point' graph suggests the player should purify death rofl

can you link me that log? hard finding a case like that

> The green and red dots on the 'problem point' graph are a little too small IMO - it's difficult to mouse over them for a tooltip.

Increased size from 50 to 75.

> For the shuffled hits timeline, things get a little wonky when there are a lot of hits, especially with the mouse-overs. Probably not a big deal though

this is actually an issue with lords of dread because some of the dots aren't flagged as ticks in the combat log. i added them to the ignore list manually and that removes several hundred hits :)

> I'd like to see every new module in core get documented such that a developer who wants to use the module could read the docs and have a reasonable idea on how to proceed. Problem.tsx is particularly in need of this - from running your prototype I see what it does, and looking at the code it's clear that the 'Problem' code is building it, but I'd probably have spend half an hour scrutinizing your implementation in the BrM guide to figure out exactly how it's working.

This is a very reasonable requirement. I'll work on that.

> Generalization - I think the components you built that aren't directly tied to BrM should go in core, even if you don't see an immediate need for them. I'm thinking HitsList MicroTimeline and ProblemList here - I can already see exactly how I'd use them in a Guardian module because Ironfur is so similar to Shuffle in how you want to play with it.

I would like to do this in a follow-up PR. With us putting more react-component development in the hands of contributors, I would like to introduce either a scoped css or css-in-js library like https://vanilla-extract.style/ or https://styled-components.com/ to make it easier to manage css across contributors. Each of those has a good chunk of css attached that is currently only loosely-scoped that I would like to fix.

kfinch:
> can you link me that log? hard finding a case like that

I think I found which pull this was from: https://www.warcraftlogs.com/reports/kcK6tWz4VhQxwq8G#fight=29&type=deaths

kfinch:
> I would like to do this in a follow-up PR. With us putting more react-component development in the hands of contributors, I would like to introduce either a scoped css or css-in-js library like https://vanilla-extract.style/ or https://styled-components.com/ to make it easier to manage css across contributors. Each of those has a good chunk of css attached that is currently only loosely-scoped that I would like to fix.

That's fine with me. I was thinking the same thing regarding CSS management, and you certainly know more about doing that than I do.

emallson:
@kfinch added a bunch of doc comments. lmk what you think.

the `interface/guide/index.tsx` file has a module comment but not all editors support that so I might move it to the `Guide` or `GuideProps` type.

kfinch (approved):

PR #4965, opened 2022-06-27
Fire Mage Shared Code POC

Sharrq:
Proof of concept for creating shared code and functions to minimize the amount of repeated code and to simplify the process of adding modules to specs


analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+    eventHistory: EventHistory,
+  };
+  protected abilityTracker!: AbilityTracker;
+  protected eventHistory!: EventHistory;
+
+  castBreakdownDuringBuff(buff: SpellInfo) {
+    /**
+     * @param buff the spell object for the buff
+     * @returns an array containing each unique spell cast and the number of times it was cast
+     */
```
kfinch: Function docs should be directly above the function declaration, not below. This has actual consequences in many IDEs which won't recognize these docs when placed below.

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+        }
+      });
+    return castArray;
+  }
+
+  countTotalCasts(spell: SpellInfo) {
+    return this.abilityTracker.getAbility(spell.id).casts;
+  }
+
+  countEventsDuringBuff(buff: SpellInfo, eventType: string, cast?: SpellInfo | undefined) {
```
kfinch: On `cast?: SpellInfo | undefined` - the `?` implies that it can be undefined, you don't need to also include the `| undefined`

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+    );
+    const filteredCasts: BeginCastEvent[] = [];
+    events &&
+      events.forEach((cast) => {
+        filteredCasts.push(cast);
+      });
+    return filteredCasts;
+  }
+
+  highlightTimeline(
```
kfinch: Should add docs here, and also it's unclear from function name that this marks cast inefficient (when 'enhanced' is also an option)

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+        }
+      });
+    return filteredCasts;
+  }
+
+  getBeginCastEvents(
+    count?: number | undefined,
+    startTimestamp?: number | undefined,
+    duration?: number | undefined,
+    spell?: SpellInfo | undefined,
```
kfinch: The `?` all imply these can be undefined, you don't need to make it explicit

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+        if (CASTS_THAT_ARENT_CASTS.includes(cast.ability.guid)) {
+          return;
+        } else {
+          filteredCasts.push(cast);
+        }
+      });
+    return filteredCasts;
+  }
+
+  getBeginCastEvents(
```
kfinch: It's not clear to me that this needs to be provided as a 'standard check' - it seems like it's just using `eventHistory.last` with the type locked as `BeginCastEvent`. Why not just use `eventHistory.last`?

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const filteredCasts: CastEvent[] = [];
+    events &&
+      events.forEach((cast) => {
+        if (CASTS_THAT_ARENT_CASTS.includes(cast.ability.guid)) {
+          return;
+        } else {
+          filteredCasts.push(cast);
+        }
+      });
+    return filteredCasts;
```
kfinch: You can use an Array `filter` to do this a lot more cleanly: The following line has the same effect: `return events.filter(e => !CASTS_THAT_ARENT_CASTS.includes(e.ability.guid));`

You also don't need to do the `events &&` check - it looks to me like `eventHistory.last` can't return undefined

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+    castEvents &&
+      castEvents.forEach((c) => {
+        const index = castArray.findIndex((arr) => arr.includes(c.ability.guid));
+        if (index !== -1) {
+          castArray[index][1] += 1;
+        } else {
+          castArray.push([c.ability.guid, 1]);
+        }
+      });
+    return castArray;
```
kfinch: I really don't like this 2D array return type. The 2nd dimension of the array is being used like an unlabeled object where the 0 index is implicitly the spellId and the 1 index is implicitly count. This should explicitly be an object instead. This will have the additional effect of making the downstream code more readable:
```
const castArray: { spellId: number, count: number }[] = [];
    castEvents &&
      castEvents.forEach((c) => {
        const index = castArray.findIndex((arr) => arr.spellId === c.ability.guid);
        if (index !== -1) {
          castArray[index].count += 1;
        } else {
          castArray.push({ spellId: c.ability.guid, count: 0 });
        }
      });
```

analysis/mage/src/StandardChecks.tsx
```diff
@@ -0,0 +1,161 @@
...
+     * @param cast an optional cast spell object to count. Omit or leave undefined to count all casts
+     */
+    const events = this.getEventsDuringBuff(buff, eventType, cast);
+    return events.length;
+  }
+
+  getEventsDuringBuff(buff: SpellInfo, eventType: string, spell?: SpellInfo | undefined) {
+    /**
+     * @param buff the spell object for the buff
+     * @param eventType the type of event that you want to search for. i.e. "cast", "begincast", EventType.Cast, EventType.BeginCast, etc.
```
kfinch: This is implying it will work for any eventType when in fact it only works for Cast and BeginCast. If you want it to work for every event type obviously need to find something better than the switch statement. If you're fine with it being only those two types, you can actually restrict this in the type annotation like so: `eventType: 'cast' | 'begincast'`

Sharrq:
Thanks, some of those things I planned on doing just haven't gotten to it yet (adding the docs to every function). And some of that I've been redoing as I've been going.

Like the one you mentioned that only accepts cast and begin cast atm, that's because so far that's all I've needed. I don't know what all it will or won't need by the end, so it may get redone all depending.

I learned about filter not long ago and I've already started moving some stuff to using that instead, I just haven't committed those changes yet

Sharrq:
I am going to work on cleaning this up and finalizing it with the modules that I have already redone instead of waiting till the entire spec is done ... that way I can get it moved over to the dragonflight branch and continue working on it from there alongside pruning stuff for dragonflight.

So feel free to start reviewing and providing feedback so I can get them implemented while i work on cleaning it up. Below are a few things that I am planning on finishing while I clean this up.

- Going to probably change the name from standardChecks ... maybe sharedCode or something since that is what we have been calling it
- Going to move most of the existing standardChecks/sharedCode stuff into a Core module ... possibly extending that with a class module for some of the things that might be more mage specific
- Changelogs and Integration Tests of course
- Need to do a pass on everything I redid to make sure the results match what is on live currently (unless the rewrite is more accurate than live, which ended up being the case in a few places i think)
- Going to make sure the docs are complete/accurate/worded well.

I will take this out of Draft since I am ready for people to take a look at provide feedback. But please don't merge it yet as I still need to verify stuff. 

It is very possible there is a simpler way to do some of the stuff im doing, so if so then let me know. I am not a developer so some of this is stuff I have kinda figured out along the way.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+        e.meta = e.meta || {};
+        e.meta.isInefficientCast = true;
+        e.meta.inefficientCastReason = tooltip;
+      });
+    } else {
+      event.meta = event.meta || {};
+      event.meta.isInefficientCast = true;
+      event.meta.inefficientCastReason = tooltip;
+    }
+  }
```
emallson: this should be a standalone function instead of in an analyzer, since it doesn't use any analyzer-only things like dependencies.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+      : new EventFilter(eventType).by(source);
+    const events = searchBackwards
+      ? this.eventHistory.last(count, duration, eventFilter, startTimestamp)
+      : this.eventHistory.next(count, duration, eventFilter, startTimestamp);
+
+    const filteredEvents = events.filter((e) =>
+      HasAbility(e) ? !CASTS_THAT_ARENT_CASTS.includes(e.ability.guid) : true,
+    );
+    return filteredEvents;
+  }
```
emallson: this should likely go on `EventHistory` instead of a generic `SharedCode` module.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+   */
+  getEvents<ET extends EventType>(
+    searchBackwards: boolean = true,
+    eventType: ET,
+    spell?: SpellInfo | SpellInfo[],
+    count?: number,
+    startTimestamp: number = this.owner.fight.end_time,
+    duration?: number,
+    includePets: boolean = false,
+  ): Array<MappedEvent<ET>> {
```
emallson: it is usually better to have an `options` object rather than a large number of optional/default parameters. positional parameters (especially ones that share a type but not a meaning, like `startTimestamp` and `duration`) are error prone in large quantities.

```suggestion
  type EventSearchOptions = {
    searchBackwards?: boolean;
    spell?: SpellInfo | SpellInfo[];
    count?: number;
    startTimestamp?: number;
    duration?: number;
    includePets?: boolean;
  }
  
  getEvents<ET extends EventType>(
    eventType: ET,
    { 
      searchBackwards, 
      spell, 
      count, 
      startTimestamp, 
      duration, 
      includePets
    }: EventSearchOptions = {}
  ): Array<MappedEvent<ET>> {
```

then callers use `getEvents(eventType, { spell: whatever })` and leave off the ones they don't care about. note that when taking this approach, you need to use `?` or `??` in the method body to default parameters, since providing default values in the params for this will be overwritten by the object.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * @param event the event that you want to check the target's health on.
+   * @returns the target's health percentage (in decimal, i.e. 0.75 = 75%)
+   */
+  getTargetHealth(event: AnyEvent) {
+    const target =
+      (event.type === EventType.Cast || event.type === EventType.BeginCast) && HasTarget(event)
+        ? encodeTargetString(event.targetID, event.targetInstance)
+        : undefined;
```
emallson: if you don't have a target, you can immediately return

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+   * @param eventType the type of event that you want to search for. i.e. EventType.Cast, EventType.BeginCast, etc. Use EventType.Event for all events.
+   * @param cast an optional cast spell object to count. Omit or leave undefined to count all casts
+   * @returns the number of events found
+   */
+  countEventsByBuff(buffActive: boolean, buff: SpellInfo, eventType: EventType, cast?: SpellInfo) {
+    const events = buffActive
+      ? this.getEventsByBuff(true, buff, eventType, cast)
+      : this.getEventsByBuff(false, buff, eventType, cast);
+    return events.length;
+  }
```
emallson: generally, I don't think we need additional methods for `countX` when the body is (almost) just `X.length`.

in the case of `countEventsByBuff``, you will only ever call it with `buffActive` being true (if your code cares about events while the buff was up) or false (for the opposite)---it will never vary from call to call, so the ternary won't come up in caller code.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * @param buff the spell object for the proc's buff.
+   * @param spenderSpell the spell object (or an array of spell objects) that are used to spend the proc.
+   * @returns a numeric count of expired procs
+   */
+  countExpiredProcs(buff: SpellInfo, spenderSpell: SpellInfo | SpellInfo[]) {
+    return this.getExpiredProcs(buff, spenderSpell).length;
+  }
```
emallson: similar comment for this `countX` method

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+        EventType.Cast,
+        spenderSpell,
+        1,
+        e.timestamp + 1,
+        250,
+      )[0];
+      return !castEvent;
+    });
+    return filteredEvents;
+  }
```
emallson: this implementation is a little bit spooky to me because of how it could theoretically blow up by calling `getEvents` for every event, but I can't really think of a realistic scenario where that causes problems.

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+   */
+  countExpiredProcs(buff: SpellInfo, spenderSpell: SpellInfo | SpellInfo[]) {
+    return this.getExpiredProcs(buff, spenderSpell).length;
+  }
+
+  /**
+   * @param event the event that you want to check the target's health on.
+   * @returns the target's health percentage (in decimal, i.e. 0.75 = 75%)
+   */
+  getTargetHealth(event: AnyEvent) {
```
emallson:
```suggestion
  getTargetHealth(event: AnyEvent): number | undefined {
```

src/parser/shared/modules/EventHistory.ts
```diff
@@ -82,6 +96,92 @@ class EventHistory extends Module {
...
+        };
+      }
+    }
+
+    let history = this.owner.eventHistory.filter((event) => filter(event));
+    if (count && count < history.length) {
+      history = history.slice(count);
+    }
+    return history as E[];
+  }
```
emallson, line 164: it looks like these two methods (`last` and `next`) can be largely combined. that is:

- Create a new `protected buildFilter(...)` method that takes `initialFilter` in place of the timestamp parameter (+ the other provided parameters)
- Call this from `last` and `next`, like so:

```ts
public next(/* ... */): E[] {
    let filter = (event: AnyEvent) => true;

    if (fromTimestamp) {
      const prevFilter = filter;
      filter = (event) => {
        if (!event.timestamp || event.timestamp < fromTimestamp) {
          return false;
        }
        return prevFilter(event);
      };
    }

    filter = this.buildFilter(filter, /* ... */);

    let history = this.owner.eventHistory.filter((event) => filter(event));
    if (count && count < history.length) {
      history = history.slice(count);
    }
    return history as E[];
}
```

this cuts out ~70 lines from each method

src/parser/core/SharedCode.tsx
```diff
@@ -0,0 +1,231 @@
...
+    spell?: SpellInfo | SpellInfo[],
+  ): Array<MappedEvent<ET>> {
+    const events = this.getEvents(true, eventType, spell);
+    const filteredEvents = events.filter((e) =>
+      buffActive
+        ? this.selectedCombatant.hasBuff(buff.id, e.timestamp - 1)
+        : !this.selectedCombatant.hasBuff(buff.id, e.timestamp + 1),
+    );
+    return filteredEvents;
+  }
```
emallson: I would turn this into two methods, tbh. `getEventsWithBuff` and `getEventsWithoutBuff`. slight code duplication, but it is easier to tell what is going on without having to consult the documentation for what the boolean parameter means.

emallson (changes requested):
I looked at this a bit ago and wasn't entirely sure how I wanted to handle it. You're on the right track, but I think the implementation details aren't the direction that we necessarily want to go.

What the code here points to is gaps in our common, core API for interacting with Event history more than anything, I think. To that end, I've suggested some specific changes to the implementation.

More broadly, I think it'd be good if you took the code in the core `SharedCode` module and divided it up into:

1. Things that go in the core `EventHistory` (mainly `getEvents`, `getEventsByBuff` could as well)
2. Things that go in a mage-specific shared module (most of the other methods)
3. Other methods that eventually go in core places.

For point 3:

I think that the `highlightInefficientCast` method can go in `parser/results/Timeline/Cast` (where it could be imported as `import { highlightInefficientCast } from 'parser/results/Timeline/Cast`---now that we have auto-imports working again this should be much easier).

I think the `getTargetHealth` method will eventually move to a dedicated `HealthTracker` module that implements this + other related things (I've needed this a lot for tank code), but for now can live in your mage code.

One of the areas that I'm planning to work on as we head towards Dragonflight is trying to provide better common code support---a lot of that initially will be for new DF stuff (like talents) and new features (like the guide stuff), but it will also mean looking at frequent tasks that currently take a lot of effort and making them easier (like the `EventHistory` calls that you've added here).

PR #4970, opened 2022-07-04
Added generic 'Buff Count Graph' plot generator, and added an implementation to Resto Druid.

kfinch:
image


analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
```diff
@@ -0,0 +1,49 @@
...
+    this.selectedCombatant.hasTalent(SPELLS.FLOURISH_TALENT) &&
+      castSpecs.push({ spells: SPELLS.FLOURISH_TALENT, color: '#ffbb22' });
+    this.selectedCombatant.hasCovenant(COVENANTS.NIGHT_FAE.id) &&
+      castSpecs.push({ spells: SPELLS.CONVOKE_SPIRITS, color: '#0000bb' });
+    return castSpecs;
+  }
+
+  statistic() {
+    return (
+      <Panel title="Hot Graph" position={100} explanation={<></>}>
```
abelito75: **This does not block this PR**

Should we try to make a better way for generating position numbers for specific parts of the `Details`
I know personally I put `position={100}` to get it above all the mana detail stuff but it took a bit of guesswork to figure out what number worked... I'd imagine if we start adding more graphs or class/spec specific information in the detail area it could be rather annoying to get it between two things

src/parser/shared/modules/BuffCountGraph.tsx
```diff
@@ -0,0 +1,439 @@
...
+      });
+    });
+    return lookup;
+  }
+
+  /** Converts the external specification type to the internal tracking type */
+  _convertToInternalTracker(spec: GraphedSpellSpec): GraphedSpellInternalTracker {
+    const spells: Spell[] = Array.isArray(spec.spells) ? spec.spells : [spec.spells];
+    const name: string = spec.name !== undefined ? spec.name : spells[0].name;
+    const color: string = this._convertColor(spec.color);
```
abelito75, line 387: There should probably be a check here or somewhere else to confirm the developer actually put it in as #rrggbb... I know its stated that you should be a simple check that is a few lines could save a lot of pain later

analysis/druid/src/shadowlands/ConvokeSpirits.tsx
```diff
@@ -227,9 +227,10 @@ class ConvokeSpirits extends Analyzer {
     );
   }
 
-  onConvoke(_: ApplyBuffEvent) {
+  onConvoke(event: ApplyBuffEvent) {
```
abelito75, line 230: did you mean to add convoke stuff to this PR?

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+      castSpecs.push(this.convokeSpec);
+      this.convokeWithFlourishSpec = { name: 'Convoke w/ Flourish', spells: [], color: '#00aacc' };
+      castSpecs.push(this.convokeWithFlourishSpec);
+    }
+    return castSpecs;
+  }
+
+  statistic() {
+    return (
+      <Panel title="Hot Graph" position={100} explanation={<></>}>
```
abelito75: It might be helpful to add a small explanation on what you're meant to be able to view with this graph. If not then people might start asking why other hots are not on the graph

abelito75: This isn't a blocker for this PR but we should probably try to find a better way to do position instead of abstract numbers... maybe subgrouping in the major groupings? Who knows but finding the correct number to put it in details and then where you want it is 100% guess work right now

src/parser/shared/modules/BuffCountGraph.tsx
```diff
@@ -0,0 +1,440 @@
...
+    this._onBuffChanged(event, 1);
+  }
+
+  onBuffRemoved(event: RemoveBuffEvent | RemoveDebuffEvent) {
+    this._onBuffChanged(event, -1);
+  }
+
+  _onBuffChanged(event: AbilityEvent<any>, change: number) {
+    const applicableTrackers = this.buffTrackerLookup[event.ability.guid];
+    if (applicableTrackers === undefined) {
```
abelito75: Could simplify this to `if(!applicableTrackers)` if I'm not mistaken

src/parser/shared/modules/BuffCountGraph.tsx
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  /**
+   * Adds a vertical rule line for the given tracker at the given timestamp
+   * @param trackerName the name of the tracker this rule line is for -
+   *     will determine the line's color and its name in the legend
+   * @param timestamp the timestamp to add the rule
+   */
+  addRuleLine(trackerName: string, timestamp: number) {
+    const tracker = this.ruleLineTrackers.find((t) => t.name === trackerName);
```
abelito75, line 109: Would making a map with the keys the name bet quicker? reducing this O(n) to O(1)?

src/parser/shared/modules/BuffCountGraph.tsx
```diff
@@ -0,0 +1,440 @@
...
+ *
+ * Buffs to graph are specified by implementing the {@link buffSpecs} function to return
+ * information of the buffs to show and how they should appear in the resulting graph.
+ * Optionally the implementer can also override the {@link castRuleSpecs} function to cause
+ * vertical rule lines to be drawn on the timestamp of certain spell casts.
+ *
+ * If more complex logic is desired for the vertical rule lines, this can be accomplished as follows:
+ * Return a cast rule spec in {@link castRuleSpecs} with the desired name and color, but with an
+ * empty array passed for the 'spells' field. Then, the implementing Analyzer should call
+ * {@link addRuleLine} with the tracker's name at the timestamps where rule lines should appear.
```
abelito75, line 31: :P You should add an author tag and claim it for yourself :D

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return buffSpecs;
+  }
+
+  castRuleSpecs(): GraphedSpellSpec[] {
+    const castSpecs: GraphedSpellSpec[] = [];
+    castSpecs.push({ spells: SPELLS.TRANQUILITY_CAST, color: '#bbbbbb' });
+    this.selectedCombatant.hasTalent(SPELLS.FLOURISH_TALENT) &&
+      castSpecs.push({ spells: SPELLS.FLOURISH_TALENT, color: '#ffbb22' });
+    if (this.selectedCombatant.hasCovenant(COVENANTS.NIGHT_FAE.id)) {
+      this.convokeSpec = { name: 'Convoke', spells: [], color: '#0000bb' };
```
abelito75: Shouldn't these just be in line with the push?
so 
`castSpecs.push({ name: 'Convoke', spells: [], color: '#0000bb' });

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  buffSpecs(): GraphedSpellSpec[] {
+    const buffSpecs: GraphedSpellSpec[] = [];
+    buffSpecs.push({
+      spells: [SPELLS.REJUVENATION, SPELLS.REJUVENATION_GERMINATION],
+      color: '#900090',
+    });
+    buffSpecs.push({ spells: SPELLS.WILD_GROWTH, color: '#00b000' });
+    this.selectedCombatant.hasTalent(SPELLS.CENARION_WARD_TALENT) &&
```
abelito75: I might be insane but an If statement would probably be easier to read here... More LoC but we don't play golf

abelito75 (commented):
My mind isn't in the right state to fully review this so here is what my mind found/thought of right now

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
kfinch: yeah it's definitely annoying and something to look at, but as you say - out of scope for this PR. I don't have any better ideas in mind unfortunately

src/parser/shared/modules/BuffCountGraph.tsx
kfinch: As I accidentally discovered, putting in the wrong format doesn't cause a crash, it just makes the color end up black in the graph, which hopefully the dev will notice. Still, I'll add a quick regex check and log an error if it's wrong too.

analysis/druid/src/shadowlands/ConvokeSpirits.tsx
kfinch: Yeah, I actually need the timestamp info in Convoke in order to easily add the flourish vs. no-flourish info in the new HotGraph. While I was there I added timestamp to Convoke's display because I had been intending to do that for a while.

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
kfinch: Will do.

kfinch: see above comment on the topic

src/parser/shared/modules/BuffCountGraph.tsx
kfinch: tru tru

src/parser/shared/modules/BuffCountGraph.tsx
kfinch: Only would improve perf if this gets used lots of time, but it won't because this is the method to add vertical cast rule lines, of which there will never be more than a dozen I think.. won't change because this will add complexity for unnoticeable gain.

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
kfinch: This is a relic of me mixing up the way I was doing things due to problems with variable load order - I'll fix it

analysis/druidrestoration/src/modules/features/HotCountGraph.tsx
kfinch: Yeah you're probably right

src/parser/shared/modules/BuffCountGraph.tsx
kfinch: Ehhh I haven't done that with my other stuff, and future changes could make it OBE. I'll let `git blame` do the work.

emallson (approved):
my only comment really would be about the color stuff---we may get more readable charts if we sacrifice using a fixed color per spell and instead use a [categorical color scale](https://vega.github.io/vega/docs/schemes/#categorical). However, that definitely doesn't block this and there are questions I'd have w/ that on how to make sure that e.g. Regrowth gets the same color when `buffSpecs` varies from one player to another.

abelito75:
Maybe we should make a map in the spec's constant file that is just spell/spellID: color
then you use those for all your color needs... I might do that right now for mistweaver in general

emallson:
That solves consistency across charts, I was more thinking-out-loud about making those colors perceptually distinct / nicely contrasting (which would be the reason to use one of vega's color schemes)

kfinch:
Yeah I think I'd like to keep developer's ability to pick colors, and I did specifically choose these colors to somewhat match colors in spell icons. I definitely could adjust what I have to make it look better. I wonder if we could make a "palette" of colors that look nice on WoWA and still allow dev to choose from them.

kfinch:
I'm gonna put this in for now, but I'm definitely open to enhancing this later with better color palette / more features.

PR #4973, opened 2022-07-09
WIP Disc priest guide 

oisinod:


kfinch:
First glance at the content, then I'll get to the code. Bear in mind I don't play discipline priest, but perhaps this makes me a good test because I'll see how much I learn.

I want to emphasize that this looks good and none of the feedback items here are "demands" - if you disagree with my suggestions feel free to push back on them

image
I can see from the code there are a lot of conditions you use to decide if the ramp is good, but the actual text description of what to do is pretty cursory - perhaps expand on this?

image
I see your SpellIcons are noLink - I had to look up a bunch of these to figure out what they were. Obviously this will be less of an issue for someone who actually plays the class, but surely allowing mouseover link would do no harm here

image
`You didn't refresh your dot at the start of the ramp.` - what's "my DoT"? Be specific about what spell you're expecting. If it changes based on talent, you can use an if clause to show the right one. Also, mousing over this highlights a Shadow Word: Pain cast - is this not a DoT? Why doesn't that count?

image
Probably shouldn't repeat the same problem multiple times in the same Evang cast (time off GCD showing up twice here). Perhaps show only once per cast and say something like `It's very important to chain cast in the seconds leading up to Evangelism. The red arrows in the above sequence indicate gaps where you weren't casting!`. Also, the mouseovers highlight the previous cast in red, which is confusing to me - the prev cast isn't the problem, the gap is the problem.

Finally, while the mouseover highlighting and icon size-change is a great effect, the size-change causes the section to jitter in a way I think doesn't look good. Either remove the size-change, or fix something in the timeline div height so that the jitter doesn't happen.

analysis/priestdiscipline/src/Guide.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import type CombatLogParser from './CombatLogParser';
+import { EvangelismApplicators } from './modules/problems/ramps';
+
+export default function Guide({
+  modules,
+  events,
+  info,
+}: GuideProps<typeof CombatLogParser>): JSX.Element {
+  // console.log(modules.ramps.analyzeRamps);
```
kfinch, line 13: I like to add a `const debug = false` at the top and then leave select debug logging statements in like `debug && console.log( ...`. This way you can avoid spamming production, but if there's a problem somewhere down the road all you have to do is change the flag on your local.

analysis/priestdiscipline/src/modules/problems/ramps/analyzer.ts
```diff
@@ -0,0 +1,155 @@
...
+  index: number;
+  problemType: string;
+  timestamp: string;
+  gcdTime?: number;
+};
+
+type EvangelismRampType = {
+  rampInfo: CastEvent[];
+  evangelismTimestamp: string;
+};
```
kfinch, line 40: Brief comment description on what these are and what the fields mean may be helpful

analysis/priestdiscipline/src/modules/problems/ramps/analyzer.ts
```diff
@@ -0,0 +1,155 @@
...
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.EVANGELISM_TALENT),
+      this.onCast,
+    );
+    this.evangelismRamps = [];
+    this.rampProblems = [];
+  }
+
+  onCast(event: CastEvent) {
+    const rampHistory = this.eventHistory
+      .last(30, 17000, Events.cast.by(SELECTED_PLAYER))
```
kfinch, line 69: 30 / 17000? What's this about

kfinch (commented):
A few comments on the code. Would have to look more at the components you built to see which could be generalized and how. The cast sequence with mouseover highlighting is very neat.

kfinch:
.. I looked closer at the casting gap thing and I think you have a bug with its handling of cast time spells. See this:
image
and here's what it looks like in the events (filtered to only include cast and begincast):
gap-cast-time-bug
That looks like one GCD after PW:S and before the begincast - I think you're not properly counting the time during the cast.

oisinod:
I am going to remake this from scratch, using some of the code and trying to make it a bit cleaner. Closing this PR

PR #4977, opened 2022-07-14
Resto Druid Guide

kfinch:
Resto Druid guide - including several new general purpose elements: `CooldownBar`, `GradiatedPerformanceBar`, and `PerformanceBoxRow`.

The Guide so far includes a 'Core Spells' and a 'Healing Cooldowns' section. It is more wide-ranging than the Brewmaster and Discipline guides, but less 'mechanically fancy' - I designed to the new visual elements to be more general purpose usable and easy to drop in.

I have future plans to expand this guide but I wanted to drop the PR now so I can get more feedback from the Resto Druid community and also to prevent the size of this PR from getting out of hand.

My future plans include:
* Adding a 'Talent / Legendary evaluation' section
* Adding individual CD use evaluation dropdowns, similar to the checklist design in the Brewmaster guide

image


analysis/druidrestoration/src/Guide.tsx
```diff
@@ -0,0 +1,237 @@
...
+      {/*    <SpellLink id={SPELLS.SPRING_BLOSSOMS_TALENT.id} /> is the standard pick for Raiding and*/}
+      {/*    Mythic+. Make sure to maintain high <SpellLink id={SPELLS.EFFLORESCENCE_CAST.id} /> uptime*/}
+      {/*    and you'll get strong extra healing from this talent.*/}
+      {/*    <br />*/}
+      {/*    <br />*/}
+      {/*    <SpellLink id={SPELLS.OVERGROWTH_TALENT.id} /> is terrible do not pick it.*/}
+      {/*  </SubSection>*/}
+      {/*  <SubSection title="Tier 7 - CHOSEN TALENT GOES HERE"></SubSection>*/}
+      {/*  <SubSection title="Legendary - CHOSEN LEGENDARY GOES HERE"></SubSection>*/}
+      {/*</Section>*/}
```
kfinch, line 234: I started making an outline for this section but decided I should save it for a future PR, hence the commented out block.

analysis/druidrestoration/src/modules/features/PrematureRejuvenations.tsx
```diff
@@ -28,92 +36,123 @@ class PrematureRejuvenations extends Analyzer {
   static dependencies = {
     healingDone: HealingDone,
     combatants: Combatants,
+    hotTracker: HotTrackerRestoDruid,
```
kfinch, line 37: Refactored this class to use the the more detailed info from `HotTrackerRestoDruid` rather than trying to generate it itself.

analysis/druidrestoration/src/modules/features/WildGrowth.tsx
```diff
@@ -1,32 +1,28 @@
...
 const OVERHEAL_THRESHOLD = 0.5;
 
 /**
  * Tracks stats relating to Wild Growth
- *
- * This modules functioning relies on normalizers that ensure:
- * * Wild Growth cast will always come before Wild Growth HoT apply
- * * Wild Growth HoT apply will always come before Wild Growth first heal
  */
 class WildGrowth extends Analyzer {
```
kfinch, line 27: Refactored this to use `EventLinkNormalizer` (the new way I've been doing things) instead of event order tracking

kfinch (commented):
Some comments on changes to non-guide parts.

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -0,0 +1,237 @@
...
+          {modules.lifebloom.subStatistic()}
+        </SubSection>
+        <SubSection title="">
+          <b>
+            <SpellLink id={SPELLS.EFFLORESCENCE_CAST.id} />
+          </b>{' '}
+          is extremely mana efficient if you're good about placing it where raiders are standing.
+          Under the boss is usually a safe bet. Aim to keep it active at all times.
+          <br />
+          {modules.efflorescence.subStatistic()}
```
emallson, line 98: i know this is taking the efflo substat, but this should really be colored differently on the guide for usability reasons. everywhere else on the guide, red = bad. but the efflo one has a reddish pink = good

analysis/druidrestoration/src/modules/features/RegrowthAndClearcasting.tsx
```diff
@@ -220,6 +235,30 @@ class RegrowthAndClearcasting extends Analyzer {
...
+      } else if (rgCast.reason === 'ns') {
+        message = "Free due to Nature's Swiftness";
+      } else if (rgCast.reason === 'clearcast') {
+        message = 'Free due to Clearcasting';
+      } else if (rgCast.reason === 'abundance') {
+        message = `Cheap due to ${rgCast.abundanceStacks} stacks of Abundance`;
+      } else if (rgCast.reason === 'triage') {
+        message = 'Cast full price on a low health target';
+      } else if (rgCast.reason === 'bad') {
+        message = 'Cast full price on a high health target';
```
emallson, line 252: showing the target name here would be helpful for quickly understanding the cast (e.g. if it shows your tank's name, then that is different than maybe showing yourself or another healer)

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -0,0 +1,237 @@
...
+        </SubSection>
+        <SubSection title="">
+          <b>
+            <SpellLink id={SPELLS.SWIFTMEND.id} />
+          </b>{' '}
+          is our emergency heal and it isn't very mana efficient - normally it should only be used
+          to prevent an eminent death. However, if using{' '}
```
emallson:
```suggestion
          to prevent an imminent death. However, if using{' '}
```

emallson (commented):
Couple of notes. willing to merge so you can more easily show people for feedback:

- <s>The HoT graph's legend is difficult to match with the vertical lines. The colors seem off. Possibly an issue with opacity of the legend's color indicators?</s> i just pushed a fix

- It would be nice to have a table/accordion to summarize cooldown ramps. is there a "good enough" threshold? could be a table like "Cooldown | Rejuv Count | WG Count | Mastery Stacks" or an accordion like I've got for brew Invoke Niuzao depending on how much detail/context you need to show for each. might not be necessary and would add a lot of space since you've already got the two big charts

- I notice there is nothing about mana here. Is that not relevant? for example if you use Swiftmend on CD for procs do you run the risk of going OOM?

analysis/druidrestoration/src/Guide.tsx
kfinch: I was kinda thinking the same thing, but wasn't sure on a good solution

analysis/druidrestoration/src/Guide.tsx
kfinch: oops good catch

analysis/druidrestoration/src/modules/features/RegrowthAndClearcasting.tsx
kfinch: I considered adding more info, like the target health / name / overheal percent, but decided to not overload it. I'll take another look - also is there an easy way to get target name from targetID?

kfinch:
>     * The HoT graph's legend is difficult to match with the vertical lines. The colors seem off. Possibly an issue with opacity of the legend's color indicators? i just pushed a fix
I was aware of this but couldn't figure out how to fix it - thanks 😄 

>     * It would be nice to have a table/accordion to summarize cooldown ramps. is there a "good enough" threshold? could be a table like "Cooldown | Rejuv Count | WG Count | Mastery Stacks" or an accordion like I've got for brew Invoke Niuzao depending on how much detail/context you need to show for each. might not be necessary and would add a lot of space since you've already got the two big charts
I absolutely plan to do this, will come in a following PR. Didn't want to let this PR get too huge.

>     * I notice there is nothing about mana here. Is that not relevant? for example if you use Swiftmend on CD for procs do you run the risk of going OOM?
I should actually update the Swiftmend blurb because it's less inefficient than I realized - it's kinda hard to OOM as Resto Druid unless you're actually spamming spells on people who don't need it. I cover mana a bit in both the Rejuv section (with high overheal / early refresh) and the Regrowth section. I'd consider adding an explicit "mana" section, but would have to think how to display it.

PR #4978, opened 2022-07-14
[Havoc DH] add Sinful Brand uptime tracker

ToppleTheNun:
Add Sinful Brand uptime tracker as it's where the majority of HDH damage comes from in raid.

Change the recommendation of casting Sinful Brand on cooldown as it's actually harmful if you're using double legendary (will add a PR for this check later).

**Sinful Brand DoT Uptime included in Checklist**
image

**Substatistic for Sinful Brand DoT Uptime**
image

**Non-Venthyr Havoc DH**
image

### DIFF
--- a/analysis/demonhunterhavoc/src/CONFIG.js
+++ b/analysis/demonhunterhavoc/src/CONFIG.js
@@ -1,3 +1,4 @@
+import { ToppleTheNun } from 'CONTRIBUTORS';
 import Expansion from 'game/Expansion';
 import SPECS from 'game/SPECS';
 
@@ -5,7 +6,7 @@ import CHANGELOG from './CHANGELOG';
 
 export default {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [],
+  contributors: [ToppleTheNun],
   expansion: Expansion.Shadowlands,
   // The WoW client patch this spec was last updated.
   patchCompatibility: null,
@@ -43,7 +44,8 @@ export default {
     </>
   ),
   // A recent example report to see interesting parts of the spec. Will be shown on the homepage.
-  exampleReport: '/report/TGzmk4bXDZJndpj7/6-Heroic+Opulence+-+Kill+(8:12)/5-Tarke',
+  exampleReport:
+    '/report/Y9Z4RLcnNbz2TkAX/11-Mythic+Skolex,+the+Insatiable+Ravener+-+Kill+(5:02)/Felahunter',
 
   // Don't change anything below this line;
   // The current spec identifier. This is the only place (in code) that specifies which spec this parser is about.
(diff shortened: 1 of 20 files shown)


analysis/demonhunterhavoc/src/modules/covenants/SinfulBrandUptime.tsx
```diff
@@ -0,0 +1,83 @@
...
+          t({
+            id: 'demonhunter.havoc.suggestions.sinfulBrand.uptime',
+            message: `${formatPercentage(actual)}% Sinful Brand uptime`,
+          }),
+        )
+        .recommended(`>${formatPercentage(recommended)}% is recommended`),
+    );
+  }
+
+  subStatistic() {
```
kfinch, line 65: Is this getting rendered? `subStatistic` doesn't get placed on the page on its own, it has to be called from somewhere else. If you want a standalone stats box, use `statistic()`

kfinch (changes requested):
Overall looks good, but a screenshot would help to show what this looks like (especially because I think your stats box might not be rendered, see the comment below)

analysis/demonhunterhavoc/src/modules/covenants/SinfulBrandUptime.tsx
ToppleTheNun: That's what I get for copy and pasting from the CorruptionUptime implementation. 😄

analysis/demonhunterhavoc/src/CONFIG.js
```diff
@@ -1,14 +1,15 @@
+import { ToppleTheNun } from 'CONTRIBUTORS';
 import Expansion from 'game/Expansion';
 import SPECS from 'game/SPECS';
 
 import CHANGELOG from './CHANGELOG';
 
 export default {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [],
+  contributors: [ToppleTheNun],
```
kfinch, line 9: Just so you know, you aren't required to put yourself here. If you do, you'll probably get people asking for more changes. If you're fine with that and intend to add more Havoc DH stuff then certainly you can leave this here.

analysis/demonhunterhavoc/src/CONFIG.js
```diff
@@ -1,14 +1,15 @@
...
   expansion: Expansion.Shadowlands,
   // The WoW client patch this spec was last updated.
-  patchCompatibility: null,
+  patchCompatibility: '9.2.5',
```
kfinch: I haven't reviewed the content from this spec recently - are things up to date across the board? If you're confident we can put this through as well, but I don't want it to be automatic.

kfinch:
Apart from the above comments I think this is ready to go in, just double checking

analysis/demonhunterhavoc/src/CONFIG.js
ToppleTheNun: I'll remove this check for the moment and then do a quick run through after my TS PR to make sure everything's good to go for 9.2.5/9.2.7.

analysis/demonhunterhavoc/src/CONFIG.js
ToppleTheNun: I intend to do some more maintenance of the spec; there's some things that are missing that I'd love to improve.

kfinch (approved):

PR #4982, opened 2022-07-17
Evoker Boilerplate

tyndivelspaz:
Adding boilerplate for the Evoker class to allow for updates to be made as Alpha and beta move forward. Also to allow for faster creation of modules when that time comes.

Additions include:
- Talents
- Core Spells
- `analysis` folder for base class and both devastation and preservation specs


emallson:
Closed by #5063

PR #4986, opened 2022-07-19
Enhancement Shaman - Hot Hand - Lava Lash CD interaction

danny-janse:
The Hot Hand talent should decrease the cooldown on Lava Lash by 75% while the buff is up.
When the buff expires the cooldown shoots back to the original 100% if the spell was not yet off cooldown.

These changes made to the Hot Hand make it so that 99% of the time the cooldown is correctly reduced in combination with Hot Hand.


analysis/shamanenhancement/src/modules/talents/HotHand.tsx
```diff
@@ -21,19 +23,63 @@ const HOT_HAND = {
...
+
+  reduceLavaLushCDOnCast() {
+    if (this.selectedCombatant.hasBuff(SPELLS.HOT_HAND_BUFF.id)) {
+      this.reduceLavaLushCDOnBuff();
+    }
+  }
+
+  restoreLavaLushCDRemoveBuff() {
+    if (this.spellUsable.isOnCooldown(SPELLS.LAVA_LASH.id)) {
+      this.spellUsable.cooldownRemaining(SPELLS.LAVA_LASH.id);
```
emallson: this won't modify the cooldown remaining, unless there is some very strange implicit behavior going on behind the scenes.

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
```diff
@@ -21,19 +23,63 @@ const HOT_HAND = {
...
     );
   }
 
+  calcLavaLushCd75pct() {
+    return (9 / (1 + this.haste.current)) * 1000;
+  }
+
+  reduceLavaLushCDOnBuff() {
+    if (this.spellUsable.isOnCooldown(SPELLS.LAVA_LASH.id)) {
+      this.spellUsable.reduceCooldown(SPELLS.LAVA_LASH.id, this.calcLavaLushCd75pct());
```
emallson: i feel like we had something for accelerated cooldowns but skimming through fire mage and death knight stuff it looks like the DK one is DK-specific and the fire one just alters the cooldown when cast during combustion (which is not 100% correct but pretty damn close for the abilities in question).

@kfinch or @Pewtro might know more? this set/reset approach is probably fine for a spell without charges.

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
```diff
@@ -21,19 +23,63 @@ const HOT_HAND = {
...
   protected buffedLavaLashDamage: number = 0;
 
   constructor(options: Options) {
     super(options);
 
     this.active = this.selectedCombatant.hasTalent(SPELLS.HOT_HAND_TALENT.id);
 
+    this.addEventListener(
+      Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.HOT_HAND_BUFF),
+      this.reduceLavaLushCDOnBuff,
```
emallson: these should be `LavaLash` not `LavaLush`

emallson (changes requested):

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
danny-janse: I actually thought of another method which is to overwrite the cooldown noted generated in the spellbook using the following function in the function in the core Abilities module. 

This example is however for max charges. I'm just unsure if this is the right way to go about it. 

```ts
updateAbilityCharges(spellId: number, charges: number) {
  const index = this.activeAbilities.findIndex((ability) => {
    if (ability.spell instanceof Array) {
      return ability.spell.includes(spellId);
    } else {
      return ability.spell === spellId;
    }
  });

  if (index === -1) {
    return;
  }

  this.activeAbilities[index].charges = charges;
}
```

The way i sumbled upon this is because i noted that the ertain abilities have multiple charges based on talents but also buffs, eg; shadow priets's Mind Blast. This does work if its a talent, but not a buff like Voidform.

Edit:
Not sure if i make sense here. Its a solid 36C in my computer room atm so my brain is mushy  :D

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
danny-janse: Oh my! Thanks for noting this! Changes have been made.

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
danny-janse: Oh no, this code is obsolete. I have removed it.

analysis/shamanenhancement/src/modules/talents/HotHand.tsx
danny-janse: After a discussion about this on discord changes have been made to handle the oncast part through the spellbook. Sref adviced to keep the 'artificial' cooldown reduction when the hothands buff is applied.

emallson (approved):

PR #4989, opened 2022-07-19
Holy Priest - Remove GDC on Fade

danny-janse:
Fade does not trigger a GCD.


emallson (approved):
need a changelog entry but otherwise this is good

danny-janse:
> need a changelog entry but otherwise this is good

I always seem to forget the changelog 😄

PR #4990, opened 2022-07-20
Remove deprecated TraitStatisticBox

danny-janse:
TraitStatisticBox was in use for only two (now) talents of the holy paladin.
Converting it to <Statistic /> has let to no visual changes.

This is inspired by issue #3801


emallson (changes requested):
changes look okay, needs snapshot update (`yarn test -u`)

danny-janse:
> changes look okay, needs snapshot update (`yarn test -u`)

Is this something that should be done before a review in the future? Or rather after a review?

emallson:
> > changes look okay, needs snapshot update (`yarn test -u`)
> 
> Is this something that should be done before a review in the future? Or rather after a review?

either is fine. if you do it before review we can see the diff of the snapshot. if you do it after review, we can see the diff in the failed test run.

doing it before review means we can merge immediately if everything else looks good, so i'd slightly prefer that.

PR #4992, opened 2022-07-20
Remove deprecated TalentStatisticBox

danny-janse:
The usage of <TalantStatisticBox> is deprecated, this PR converts all usages of this to <Statistic>.

It was quite the tedious process to find a log using a specific talent in order to compare the before and after of the conversion. I noticed that most if not all of these talents are part of 'unsupported' specialisations.

This is inspired by issue #3801


ToppleTheNun:
The Havoc DH replacements are coming in #4996 as well, fwiw.

emallson (changes requested):
changes look okay to me, but you need to update the snapshots (`yarn test -u`)

emallson (approved):

PR #4994, opened 2022-07-20
Replace cookies with localStorage API

danny-janse:
Moderen browsers have an alternative to browser-only cookies. This would be through the localStorage API; [caniuse.com](https://caniuse.com/?search=localStorage) | [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)

There were three instances of cookie usage as far as i could find;
* The Privacy Page
* Language selector
* Report History.

The privacy page makes direct usage of the cookies using `document.cookie`. Whereas the later two used the `universal-cookie` package. 

Implementing this new method gets rid of a package but comes with a con; Language preference and Report History will be reset.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
     "redux-thunk": "^2.3.0",
     "sass-text-stroke": "1.2.0",
     "swr": "^0.5.6",
-    "universal-cookie": "^4.0.4",
     "vega": "~5.21.0",
     "vega-lite": "^5.2.0"
   },
--- a/src/interface/reducers/language.ts
+++ b/src/interface/reducers/language.ts
@@ -1,21 +1,16 @@
 import { SET_LANGUAGE } from 'interface/actions/language';
 import { AnyAction } from 'redux';
-import Cookies from 'universal-cookie';
 
-const cookies = new Cookies();
-const COOKIE_NAME = 'LANGUAGE';
-const cookieOptions = {
-  path: '/',
-  maxAge: 86400 * 365, // 1 year
-};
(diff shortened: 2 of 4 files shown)


ToppleTheNun:
Would it be worth converting the cookies to localStorage for a bit so that the preference and history aren't reset for everybody?

danny-janse:
> Would it be worth converting the cookies to localStorage for a bit so that the preference and history aren't reset for everybody?

@ToppleTheNun What you mean to say is to run the system in hybrid mode for a while. Lets say, untill Dragonflight launches and then remove cookie support all together? I'm a fan of this.

Pewtro:
These things should, imo, stay as cookies. Cookies serve a different purpose than localStorage - else cookies would have been replaced ages ago.

I don’t see the issue being solved by converting from cookie to localStorage, and the change can open to XSS attacks or similiar which you can’t through Secure/http-only cookies. We also lose the expirationDate functionality of cookies. Now we probably don’t store much that could be exploited, but what are we gaining from this change?

danny-janse:
> These things should, imo, stay as cookies. Cookies serve a different purpose than localStorage - else cookies would have been replaced ages ago.
> 
> I don’t see the issue being solved by converting from cookie to localStorage, and the change can open to XSS attacks or similar which you can’t through Secure/http-only cookies. We also lose the expirationDate functionality of cookies. Now we probably don’t store much that could be exploited, but what are we gaining from this change?

Truth be told, I don't see any advantages myself. i was scanning the core code and saw it as an opportunity. I'll close the PR :)

PR #5000, opened 2022-07-22
add create event type

ToppleTheNun:
add create event type

create event is triggered at least by Demonic Circle
for Warlocks

include Demonic Circle creation tracker for demo-ing
the create event

image


src/parser/core/Events.ts
```diff
@@ -96,6 +96,7 @@ export enum EventType {
   FeedHeal = 'feed_heal',
 
   // Warlock
+  Create = 'create',
```
abelito75: This is a generic non-warlock specific event wow has. It should be placed at the top and not under a specific class tag (very w/e since its just location in a enum)
Since I can't comment on un-edited lines you will also need to add this to MappedEventTypes which will require you to make a interface for it with the correct fields.

abelito75 (changes requested):
You also need a changelog event but thanks for the find!

src/parser/core/modules/EventEmitter.ts
```diff
@@ -325,6 +325,11 @@ class EventEmitter extends Module {
       );
     }
 
+    if (event.type === 'create') {
```
abelito75: Not sure if we still need this console event haha

analysis/warlock/src/DemonicCirclesCreated.tsx
```diff
@@ -0,0 +1,38 @@
...
+  constructor(options: Options) {
+    super(options);
+    this.addEventListener(
+      Events.create.by(SELECTED_PLAYER).spell(SPELLS.DEMONIC_CIRCLE_SUMMON),
+      this.handleCreateEvent,
+    );
+  }
+
+  handleCreateEvent(event: CreateEvent) {
+    console.log('Handling create event', event);
```
abelito75: Probably don't need this log statement... if you want it put it behind a debugg check
Gotta keep that production console log clean

src/parser/core/Events.ts
```diff
@@ -851,6 +853,24 @@ export interface SpendResourceEvent extends Event<EventType.SpendResource> {
   __fabricated: true;
 }
 
+export interface CreateEventTarget {
```
abelito75: Is this the same type of Target as all the other Targets in the Event.ts file? Like BeginCastEvent/BaseCastEvent
if so it might be logical to normalize them as well

abelito75 (changes requested):

analysis/warlock/src/DemonicCirclesCreated.tsx
ToppleTheNun: Whoops, my bad, will remove it.

src/parser/core/modules/EventEmitter.ts
ToppleTheNun: Definitely not lol

src/parser/core/Events.ts
ToppleTheNun: It's slightly different from what I could tell, but will double check

src/parser/core/Events.ts
```diff
@@ -380,7 +390,7 @@ export interface BaseCastEvent<T extends string> extends Event<T> {
   sourceInstance?: number;
   sourceIsFriendly: boolean;
   spellPower?: number;
-  target?: { name: 'Environment'; id: -1; guid: 0; type: 'NPC'; icon: 'NPC' };
+  target?: CastTarget;
```
emallson, line 393: hmm, strange that these were using a const interface like that. good catch. some of these legitimately don't ever have targets (`begincast` doesn't afaik) but `cast` certainly does

emallson (approved):

PR #5002, opened 2022-07-23
Death Knight Runeforge enchants

danny-janse:
I updated the list for enchant names in #4995 but only now found out that the Runeforging enchants were missing.
Also noticed that one of the icons was not working, that is resolved aswell.

Screenshot_1Screenshot_2


Pewtro:
Thanks!

PR #5014, opened 2022-07-28
[RFC][Dragonflight] Initial implementation of talent tree generation

Pewtro:
This aims to generate every spell from the class and spec trees in Dragonflight. 

The talents are strictly typed so you'll have autocompletion, you'll have type checking where appropriate and all the other regular goodies from TypeScript. 

I'm not sure about the maxRanks/reqPoints/spelltype/talentType that have been added.
I can see value in the maxRanks one for use in analysis - but I'm not sure about the others, since it generally doesn't tell us anything, that's mostly UI related. 

# Future
If we are happy with this setup I will remove src/common/SPELLS/talents folder which will break every current talent implementation, so I'm a little unsure on the best procedure there. 
Here are my initial possible avenues: 
- I can do find and replace for any talent and patch them to my best guess of the new variant - but some of those will inevitably have changed also making the module inaccurate or unable to pop up if the talent ID is different now. 
- I can also just let everything be broken until someone fixes it since we are on a highly volatile non-prod branch anyways
- I can fix some of the major issues (for example in changelogs), and let spec modules be broken until a maintainer gets to run them through


scripts/talents/talent-tree-types.ts
```diff
@@ -0,0 +1,103 @@
...
+  HolyPower,
+  AlternatePower,
+  Maelstrom,
+  Chi,
+  Insanity,
+  Unused14PowerType,
+  Unused15Powertype,
+  ArcaneCharges,
+  Fury,
+  Pain,
```
ToppleTheNun, line 20: If this is for Vengeance DH, they actually use Fury now, barring a rename in Dragonflight.

abelito75: putput is just exporting all the resource types from the database. If its being used or note isn't really that important :P

.gitignore
```diff
@@ -8,10 +8,7 @@
 /build
 
 # scripts
-/scripts/conduits/data/
-raidbotsConduitsInfo.json
-soulbindconduititem.json
-soulbindconduitrank.json
+/scripts/talents/*.json
```
ToppleTheNun, line 11: Should the generated TS files be included in the gitignore as well?

Pewtro: No - the idea is that we know what information is available for everywhere, but anyone can run the script easily to update between patches

kfinch:
Seems pretty nice. Only thing I could see being annoying is if a talent changes name and regenerating the talents causes breakage, but should be an easy search-replace.

emallson:
> I can also just let everything be broken until someone fixes it since we are on a highly volatile non-prod branch anyways

This is my preferred approach. Changelogs will be getting cleared for DF so we don't have 2 yrs of mostly-irrelevant SL changes referencing dead spells.

emallson (approved):

Pewtro:
> > I can also just let everything be broken until someone fixes it since we are on a highly volatile non-prod branch anyways
> 
> This is my preferred approach. Changelogs will be getting cleared for DF so we don't have 2 yrs of mostly-irrelevant SL changes referencing dead spells.

Deleting spells/TALENTS now then

Pewtro:
Build is now naturally failing but so it be

danny-janse:
> Build is now naturally failing but so it be

I wrote a script that replaces the old with the new format throughout all the files. Do I make a pull request to your fork?

Current status is 461 mapped and 144 talents unmapped. Which is to be expected since not all talent trees are out or some of the talents were removed/replaced with new ids.

Pewtro:
I think it’s better we explicitly require folks to go through specs and fix issues rather than assume talents function the same way.

The CI checks can be overridden so that’s not too bad

danny-janse:
> I think it’s better we explicitly require folks to go through specs and fix issues rather than assume talents function the same way.
> 
> The CI checks can be overridden so that’s not too bad

What does CI stand for?

Not sure if we think of the same thing when I mention 'mapping'. To put it into perspective: https://github.com/djanse/WoWAnalyzer/commit/1af476cc622b6248eedf4fe7fc29377d1ae877ee
image

Naturally I agree with your statement, we shouldn't asume that everything works the same going into dragonflight.
I'm unaware how switching expansions is usually approached by WOWA, but i do believe the replacement shown above will save a bunch of time for everyone.

Pewtro:
CI is continous integration (so the build checks on PRs) - the idea, to me at least, is that having them broken forces a contributor/maintainer to consider the contents of the file. It would also let them import talents whichever way they prefer, like talents.X or TALENTS_CLASS.X.

emallson:
Closed by #5061

emallson:
Closed as part of #5061

PR #5016, opened 2022-08-01
[Resto Shaman] Cleanup

niseko:
Removed a couple modules that were inaccurate enough to be more of a hindrance than help and looked over the rest to make sure they're still up to date. Set the resto shaman compatibility to current patch.


abelito75 (approved):
LGTM

PR #5018, opened 2022-08-04
Resto Druid Guide update + 1.0

kfinch:
Filled out cooldowns section of Guide, and overhauled / added detail basically everywhere else, all in response to feedback from Dreamgrove. After this PR, I consider the Guide to be in '1.0' state.

image


kfinch:
I know this is an incredibly large amount of changes - most of it is adding in all the display elements to the Resto Guide itself and also doing a lot of backend work in the Resto modules so I'd have the info I needed.

The part I'm most interested in a review of are my updates / changes to the core Guide code.

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -139,7 +148,7 @@ function EfflorescenceSubsection({ modules, events, info }: GuideProps<typeof Co
       is extremely mana efficient if you're good about placing it where raiders are standing. Under
       the boss is usually a safe bet. While it's acceptable to let it drop during heavy movement,
       you should otherwise aim to keep it active at all times.
-      <br />
+      <p />
```
emallson: why are you using empty paragraph tags here? if you want the paragraph styling, you should wrap the content in `p`. If you just want blank space, `br` was correct (or you could use `p` or a `span` with margins set)

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -150,6 +159,11 @@ function SwiftmendSubsection({ modules, events, info }: GuideProps<typeof Combat
   const hasVi = info.combatant.hasLegendary(SPELLS.VERDANT_INFUSION);
   const has4p = info.combatant.has4Piece();
   const procCount = (hasSotf ? 1 : 0) + (hasVi ? 1 : 0) + (has4p ? 1 : 0);
+  const chartDescription = ` - ${
+    hasVi ? 'Blue is great (extended high value HoTs), ' : ''
+  }Green is a fine cast, ${procCount > 0 ? 'Yellow' : 'Red'} is a non-triage (>50% health) cast${
+    hasVi ? '' : ' that removes a WG or Rejuv'
+  }${procCount ? ' (still acceptable for generating procs)' : ''}`;
```
emallson: this has a LOT of conditional behavior packed into a tiny space. should probably just be 2 strings and you pick between if `procCount != 0`.

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -343,7 +374,24 @@ function CooldownBreakdownSubsection({
...
+function CooldownExpandable(): JSX.Element {
+  const [isExpanded, setIsExpanded] = useState(false);
+  return (
+    <ControlledExpandable
+      header={<SectionHeader>Convoke lul</SectionHeader>}
+      element="section"
+      expanded={isExpanded}
+      inverseExpanded={() => setIsExpanded(!isExpanded)}
+    >
+      ARG BLARG
```
emallson: nice lol

analysis/druidrestoration/src/Guide.tsx
```diff
@@ -373,25 +376,334 @@ function CooldownBreakdownSubsection({
...
+                      }
+                    >
+                      <span>
+                        <InformationIcon />
+                      </span>
+                    </Tooltip>
+                  </>
+                ),
+                result: <PassFailCheckmark pass={cast.form === 'Tree of Life'} />,
+              });
```
emallson: this should *probably* not be done inline. my personal preference would be for this to be a component that you pass the props necessary for rendering (e.g. `cast.form`, `restoCast.recentlyFlourished`). Long-term, we can build a standard `CooldownChecklist` component that consumes props that are very similar to what you're using here, and then you can use a wrapper that looks something like:

```tsx
const ConvokeCastChecklist = (props) => (
  <CooldownChecklist>
    <ChecklistRow>
      <ChecklistLabel tooltip={...}>
        Sync with <SpellLink id={SPELLS.RESTO_DRUID_TIER_28_4P_SET_BONUS.id} />
      </ChecklistLabel>
      <ChecklistResult>
       <PassFailCheckmark pass={props.cast.form === 'Tree of Life'} />
      </ChecklistResult>
    </ChecklistRow>
   ...
  </CooldownChecklist>
);
```

that said, **i don't want you to re-invent the whole module for the purposes of this PR.** would rather a smaller change like extracting this logic to a function so that this method isn't so incredibly long.

src/interface/guide/index.tsx
```diff
@@ -219,12 +219,14 @@ export function PassFailBar({ pass, total }: { pass: number; total: number }) {
   );
 }
 
+export const PerfectMark = () => <i className="glyphicon glyphicon-ok-circle perfect-mark" />;
+export const GoodMark = () => <i className="glyphicon glyphicon-ok good-mark" />;
+export const OkMark = () => <i className="glyphicon glyphicon-asterisk ok-mark" />;
+export const BadMark = () => <i className="glyphicon glyphicon-remove bad-mark" />;
```
emallson, line 225: love this. i was constantly looking up these values

src/interface/report/Results/Overview.tsx
```diff
@@ -17,20 +17,32 @@ interface Props {
...
 }) => (
   <a style={{ justifySelf: 'end' }} href="#prototype" onClick={() => setGuideMode(!guideMode)}>
-    {guideMode ? 'Return to Normal View' : 'View Prototype'}
+    {defaultGuide
+      ? guideMode
+        ? 'View Old Version'
+        : 'Return to Guide View'
+      : guideMode
+      ? 'Return to Normal View'
+      : 'View Prototype'}
```
emallson, line 33: nested ternaries are a little spicy but this isn't too bad.

emallson (changes requested):
overall looks good. i toyed around with it a bit on my guild's rdruid. i think the main thing (and referring back to that convo with @Pewtro today) is that there is a TON of stuff currently being done inline in the guide. we're definitely still learning how to organize the code for these, but not having quite so much done inline in conditionals is i think a fairly straightforward win for maintainability.

analysis/druidrestoration/src/Guide.tsx
kfinch: Mostly because I'm a heathen who doesn't know how to do inline styling well >_<

I starting using them because I liked the look of the "line-and-a-half" break it gave me - bigger than one `br` but smaller than two `br`. It sounds like I could get the same effect by wrapping in a paragraph tag, or maybe the margins set thing you suggested?

analysis/druidrestoration/src/Guide.tsx
kfinch: You're right, I'll expand this a bit so it's more readable what's going on

analysis/druidrestoration/src/Guide.tsx
kfinch: ohhh there's a commit I forgot to push it looks like - I was wondering how placeholder stuff made it through

analysis/druidrestoration/src/Guide.tsx
kfinch: Yeah my whole 'cooldown dropdowns' section is a huge mess - I'll do mitigation stuff for this PR and then I like the look of what you posted for future work

src/interface/report/Results/Overview.tsx
kfinch: I hesitated before doing this I swear! But yeah I think in this particular case it's not bad

kfinch:
All issues should now be fixed

emallson (approved):
much more nicely organized 👍

PR #5019, opened 2022-08-07
convert Buff and Buffs to TypeScript

ToppleTheNun:
convert Buff and Buffs to TypeScript to make the
future PR for supporting Fated Raids easier.

changes to getOptionalModule and getModule are necessary
to enable getting modules with dependencies in their
constructor, as TypeScript doesn't allow accessing the
derived properties from the constructor.
see https://github.com/Microsoft/TypeScript/issues/6110.


src/parser/core/modules/Buffs.ts
```diff
@@ -47,30 +68,25 @@ class Buffs extends Module {
...
-  activeBuffs = [];
-  constructor(options) {
-    super(options);
-    this.loadBuffs(this.buffs());
-  }
-  loadBuffs(buffs) {
+  loadBuffs(buffs: SpellbookBuff[]) {
     this.activeBuffs = buffs
-      .map((options) => new this.constructor.BUFF_CLASS(options))
+      .map((options) => new Buffs.BUFF_CLASS(options))
```
emallson: since this is referencing a static member of this class, it can just be replaced with:
```suggestion
      .map((options) => new Buff(options))
```

src/parser/core/modules/Buffs.ts
```diff
@@ -47,30 +68,25 @@ class Buffs extends Module {
...
   }
 
   /**
    * Add a buff to the list of active buffs.
    * @param {object} options An object with all the properties and their values that gets passed to the Buff class.
    */
-  add(options) {
-    const buff = new this.constructor.BUFF_CLASS(options);
+  add(options: SpellbookBuff) {
+    const buff = new Buffs.BUFF_CLASS(options);
```
emallson: same thing here:
```suggestion
    const buff = new Buff(options);
```

src/parser/core/modules/Buffs.ts
```diff
@@ -13,29 +25,38 @@ import Buff from './Buff';
  */
 class Buffs extends Module {
   static dependencies = {
-    haste: Haste,
     abilities: Abilities,
+    haste: Haste,
   };
   static BUFF_CLASS = Buff;
```
emallson: which allows removing this
```suggestion
```

emallson (changes requested):
minor changes then we're good. we could remove proptypes from the `Buff` object but they're actually used by a live codepath so there's more work there that can be addressed later

PR #5020, opened 2022-08-07
rebrand Buffs to Auras

ToppleTheNun:
rebrand Buffs to Auras and add support for tracking debuffs.

add the Fated raid debuffs to the timeline tracker. actual
support for additional Fated raid effects will come later.

**Current Version of Debuff UI**
image


kfinch (commented):
This has a lot of copy-paste from Buffs, maybe this should be even more tightly integrated with the existing Buffs module - do we even need a separate analyzer? Do we need a separate type for debuffs? Why not just include debuffs under the existing buffs structure, and update relevant listeners to include the additional event types?

I'd say don't change anything yet, I want more opinions on this.

ToppleTheNun:
> This has a lot of copy-paste from Buffs, maybe this should be even more tightly integrated with the existing Buffs module - do we even need a separate analyzer? Do we need a separate type for debuffs? Why not just include debuffs under the existing buffs structure, and update relevant listeners to include the additional event types?
> 
> I'd say don't change anything yet, I want more opinions on this.

I don't disagree; especially since this original case is for debuffs that are actually buffs. I think we'd still need some way to denote weird cases like these Fated infusions: currently handling it via `timelineBuff` on the `SpellbookDebuff` interface.

emallson:
> I don't disagree; especially since this original case is for debuffs that are actually buffs. I think we'd still need some way to denote weird cases like these Fated infusions: currently handling it via `timelineBuff` on the `SpellbookDebuff` interface.

definitely understandable, but probably best for everyone's sanity not to have two distinct but almost identical modules for `Buffs` and `Debuffs`. I see that it looks like you've started putting some debuff code in the Buffs module. Is the Debuffs module intended to be dead now?

EDIT: if you want to make the name clear, you could rename it to 'Auras' as well, though I don't think that is necessary

ToppleTheNun:
> > I don't disagree; especially since this original case is for debuffs that are actually buffs. I think we'd still need some way to denote weird cases like these Fated infusions: currently handling it via `timelineBuff` on the `SpellbookDebuff` interface.
> 
> definitely understandable, but probably best for everyone's sanity not to have two distinct but almost identical modules for `Buffs` and `Debuffs`. I see that it looks like you've started putting some debuff code in the Buffs module. Is the Debuffs module intended to be dead now?
> 
> EDIT: if you want to make the name clear, you could rename it to 'Auras' as well, though I don't think that is necessary

The ones that are in there were added mostly to have a way to test the changes to buffs. I think I'll just have the one Auras module add listen for ApplyDebuff and RemoveDebuff.

analysis/warriorfury/src/modules/features/Buffs.ts
```diff
@@ -1,8 +1,8 @@
 import SPELLS from 'common/SPELLS';
 import BLOODLUST_BUFFS from 'game/BLOODLUST_BUFFS';
-import CoreBuffs from 'parser/core/modules/Buffs';
+import CoreAuras from 'parser/core/modules/Auras';
 
-class Buffs extends CoreBuffs {
+class Buffs extends CoreAuras {
   buffs() {
```
emallson: this should be `auras()` now, correct?

analysis/warriorprotection/src/modules/features/Buffs.ts
```diff
@@ -1,8 +1,8 @@
 import SPELLS from 'common/SPELLS';
 import BLOODLUST_BUFFS from 'game/BLOODLUST_BUFFS';
-import CoreBuffs from 'parser/core/modules/Buffs';
+import CoreAuras from 'parser/core/modules/Auras';
 
-class Buffs extends CoreBuffs {
+class Buffs extends CoreAuras {
   //TODO: Get shrike to fill this out since I'm lazy
   buffs() {
```
emallson: same

analysis/druidferal/src/modules/Buffs.js
```diff
@@ -1,8 +1,8 @@
 import SPELLS from 'common/SPELLS';
 import BLOODLUST_BUFFS from 'game/BLOODLUST_BUFFS';
-import CoreBuffs from 'parser/core/modules/Buffs';
+import CoreAuras from 'parser/core/modules/Auras';
 
-class Buffs extends CoreBuffs {
+class Buffs extends CoreAuras {
   buffs() {
```
emallson: also this

analysis/paladinretribution/src/modules/Buffs.js
```diff
@@ -1,8 +1,8 @@
 import SPELLS from 'common/SPELLS';
 import BLOODLUST_BUFFS from 'game/BLOODLUST_BUFFS';
-import CoreBuffs from 'parser/core/modules/Buffs';
+import CoreAuras from 'parser/core/modules/Auras';
 
-class Buffs extends CoreBuffs {
+class Buffs extends CoreAuras {
   buffs() {
```
emallson: and this

emallson (changes requested):

emallson:
looks like the refactoring tool missed the Buffs.js files in the conversion

PR #5027, opened 2022-08-09
Shadow - Mindblast dynamic max charges.

danny-janse:
* Implement Voidform's max charge increase on Mind Blast. Aswell as its cooldown reset.
* Implemented Dark Thought's max charge increase on Mind Blast.
* Added two fabricated events for changing the maximum charges.


src/parser/core/modules/Abilities.ts
```diff
@@ -135,6 +152,61 @@ class Abilities extends Module {
...
+    this.eventEmitter.fabricateEvent(
+      {
+        type: EventType.MaxChargesDecreased,
+        timestamp: event.timestamp,
+        spellId: spellId,
+        by: decreaseBy,
+      },
+      event,
+    );
+  }
```
kfinch: I'd prefer `increaseBy` and `decreaseBy` were required fields, just because I saw them being called in Dark Thoughts and immediately thought "INCREASED BY HOW MUCH?". IMO, defaulting to 1 hurts readability.

kfinch (commented):
Looks pretty good - have you tested this with real logs to ensure this behaves as expected?

danny-janse:
> Looks pretty good - have you tested this with real logs to ensure this behaves as expected?

Thanks! I'm pretty proud of it. Yeah it works in-line with what is experienced in-game.

Did alot of debugging. The hardest thing to figure out was the `chargesOnCooldown` part. It would lead to a charge being on cooldown eventhough it wasn't.

src/parser/core/modules/Abilities.ts
danny-janse: I made them mandatory as you suggested.

kfinch (approved):

PR #5031, opened 2022-08-12
Remove deprecated statisticOrder property on all Analyzers.

danny-janse:
There wasn't alot of them left so i decided to completely remove them.

* Moved the value of `statisticOrder` to the `Statistic(Box).position` if it didnt already have one.
* `AlwaysBeCasting` had its own alternative; `position`
* Removed it from `ResourceUsage` is only used by`FocusUsage` but doesnt overwrite the default order of 12 so I defaulted to this.

### DIFF
--- a/analysis/druidbalance/src/modules/features/CancelledCasts.tsx
+++ b/analysis/druidbalance/src/modules/features/CancelledCasts.tsx
@@ -2,7 +2,6 @@ import { t } from '@lingui/macro';
 import { formatPercentage } from 'common/format';
 import { ThresholdStyle, When } from 'parser/core/ParseResults';
 import CoreCancelledCasts from 'parser/shared/modules/CancelledCasts';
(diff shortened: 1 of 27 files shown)


emallson (approved):
looks reasonable to me. does the `STATISTIC_ORDER` const look like it can also be removed?

danny-janse:
> looks reasonable to me. does the `STATISTIC_ORDER` const look like it can also be removed?

Not to my knowledge because `STATISTIC_ORDER` is actually used in every Statistic that has a set position.
It allows maintainers to set a perfered order of the Statistic Modules shown on the statistic tab.

danny-janse:
Altho i believe that you could argue that hte order of modules in the CombatLogParser is the default unmanipulated order?

emallson:
yea not gonna touch it then :thumbsup:

PR #5032, opened 2022-08-12
Update patch information - Old Patch Warning

danny-janse:
This hasn't been updated since hte launch of BFA. eventhough its a great notification. eg with the recent trinket nerfs that were implemented at the start of season 4.  This is what it would look like;

image


emallson (approved):
another good catch :+1:

Issue #3803, opened 2020-10-04
Internationalization of old strings


MartijnHols:
A lot of older files were made without i18n support. We need to add i18n support to those files. Please help us update all older files. This does not require any WoW or spec knowledge.

First import i18n.
```
import { i18n } from 'interface/RootLocalizationProvider';
```

then

Props:
```
error="The API is down."
```
=>
```
error={i18n._(t('visualPath.apiDown.error')`The API is down.`)}
```
Strings:
```
.actual(`${formatPercentage(actual)}% overhealing`)
```
=>
```
.actual(i18n._(t('visualPath.overhealing.actual')`${formatPercentage(actual)}% overhealing`))
```
Components:
```
<Link to="/">Home</Link>
```
=>
```
<Link to="/"><Trans id="visualPath.home">Home</Trans></Link>
```

Please use ids for all i18n strings (older i18n code has omitted this, that will need to be updated separately). The id should be `visualPath.id`, e.g. the "search" button in the report selector on the home page would be `home.reportSelector.search`.

Tutorials:

- https://lingui.js.org/tutorials/react.html#introducing-internationalization
- https://lingui.js.org/tutorials/react-patterns.html
- https://lingui.js.org/tutorials/javascript.html#localizing-your-app

jafowler:
@MartijnHols Hi! Is there a list for the `visualPath.id`? For example looking at the following error : 
`<FullscreenError
          error="An API error occured."
          details="Something went wrong talking to our servers, please try again."
          background="https://media.giphy.com/media/m4TbeLYX5MaZy/giphy.gif">`

Would that just be something like `visualPath.api.error'? 

Or

`<FullscreenError
          error="A wild INTERNET EXPLORER appeared!"
          details="This browser is too unstable for WoWAnalyzer to work properly."
          background="https://media.giphy.com/media/njYrp176NQsHS/giphy.gif">`

Not sure what that would classify as other than maybe a `visualPath.notSupported.error` or something of the like, but i don't know if there is a pre-determined list for these.

MartijnHols:
It doesn't really have to be perfect, just give a translator an idea where they can find a certain string. Since the API error can occur on any page, I would use `fullScreenError.error`, but `home.api.error` would be fine too.

jafowler:
@MartijnHols would you prefer updates in phases? Or one giant lump of a PR?

MartijnHols:
[...]

PR #5034, opened 2022-08-13
Internationalization - general DK strings

sMteX:
Part of #3803. Looking for feedback if doing it this way is ok, in which case I'd continue onto whatever more I'll find


analysis/deathknight/src/RuneBreakdown.js
```diff
@@ -38,134 +39,126 @@ class RuneBreakdown extends ResourceBreakdown {
...
-                      />
-                    </td>
-                  </tr>
-                ))}
+            {generated?.map((ability) => (
+              <tr key={ability.abilityId}>
+                <td style={{ width: '30%' }}>
+                  {ability.abilityId === SPELLS.RUNE_1.id && (
+                    <>
+                      <SpellIcon id={ability.abilityId} noLink />
```
danny-janse: Could you add `{' '}` between the icon and text so that they dont stick together?

analysis/deathknight/src/RuneBreakdown.js
```diff
@@ -38,134 +39,126 @@ class RuneBreakdown extends ResourceBreakdown {
...
+                    {resourceName} wasted
+                  </Trans>
                 </TooltipElement>
               </th>
             </tr>
           </thead>
           <tbody>
-            {generated &&
-              generated
-                .filter((ability) => ability.abilityId === SPELLS.RUNE_1.id)
```
danny-janse, line 57: I'm guessing the order of the rune generators doesnt really matter?

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
```diff
@@ -91,39 +91,53 @@ class RuneOfHysteria extends Analyzer {
...
       wastedSpell = (
-        <>
+        <Trans id="deathknight.shared.runeOfHysteria.statistic.wastedSpell">
           , losing out on {this.wastedDeathStrikes} {SPELLS.DEATH_STRIKE.name}
-        </>
+        </Trans>
+      );
+    }
+
+    // Lingui doesn't fare well with nested Trans, maybe even in combination with conditionals
```
danny-janse: I'm unsure myself if this is the right way to go about it, The maintainers might give a better 2nd opinion on this.

danny-janse (commented):
Looks overall good to me. A thing that caught my eye are the larger translations eg. tooltips. It might be hard to translate because of this. Aside from this a few pointers down below.

analysis/deathknight/src/RuneBreakdown.js
sMteX: The rune generators are already being sorted (in `ResourceBreakdown.prepareGenerated`) and while the previous structure intentionally puts the passive rune generation first (even if it wasn't the top), by the look of it it **massively dwarfs** other rune generation methods for DKs so I'm tempted to leave it as it is like this.

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
sMteX: To be fair, I've never worked with Lingui before, so maybe I'm missing some construct which it can work with.

emallson: rather than using `<Trans />` for this, you may be able to use the `t(...)` macro directly. however, that may have its own issues with nesting. I do see it used in conditionals in (for example) `ReportDurationWarning.tsx`

analysis/deathknight/src/RuneBreakdown.js
```diff
@@ -38,134 +39,126 @@ class RuneBreakdown extends ResourceBreakdown {
...
         </table>
         {showSpenders && (
           <table className="data-table">
             <thead>
               <tr>
-                <th>Ability</th>
-                <th colSpan="2">{resourceName} spent</th>
-                <th colSpan="2">Casts</th>
+                <th>
+                  <Trans id="deathknight.shared.runeBreakdown.ability">Ability</Trans>
```
emallson: common labels like this should really be a part of a common set of translation strings. we've been pretty lax about that in the past because some translations are better than none, but having to translate `Ability` for every module that uses the label is not ideal

sMteX: 👍🏼 That thought occurred to me too, I'll go through it and extract some things that could be common across specs.

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
sMteX: Yes, I've tried looking through the repo for `<Trans />` usages, and while I haven't looked through everything, just about.. 2/3s of it, nowhere I've seen nested `<Trans>` components. However using `t()` inside `<Trans>` seemed alright, so I've refactored and pushed changes with that.

emallson (approved):

PR #5037, opened 2022-08-14
Updated core DK modules to TS

sMteX:


analysis/deathknight/src/shadowlands/covenants/DeathsDue.tsx
```diff
@@ -221,7 +231,7 @@ class DeathsDue extends Analyzer {
         .recommended(`${recommended} recommended`),
     );
   }
-  statistic() {
+  statistic(): JSX.Element {
```
emallson: adding types to overridden methods isn't necessary (but you don't need to go back and remove them from this PR)

emallson (approved):

emallson:
looks like merge conflicts from the i18n PR

analysis/deathknight/src/shadowlands/covenants/DeathsDue.tsx
sMteX: Yea, I don't have any better explanation than "I felt it needed type annotation", despite already being correctly suggested 😄 Will get rid of it in places where it's not needed.

Issue #2386, opened 2018-10-05
Steal xivanalysis' dependency usage eslint rule


MartijnHols:

MartijnHols:
https://github.com/xivanalysis/linter/blob/master/lib/rules/no-unused-dependencies.js

abelito75:
Is this still something desired?

abelito75:
So like... does that PR mean we are done with this issue?

PR #5039, opened 2022-08-14
Add local eslint rule to require spreading parent dependencies

### LINKED ISSUE #2703: Add an eslint rule that modules defining `dependencies` must always spread parent's dependencies
Not spreading a parent's dependencies leads to bugs.

It would be very nice if it excluded modules just extending Analyzer/Module/EventSubscriber, but not essential.

sMteX:
Fixes #2703

At least I tried to 😄 I don't know what's your stance to adding new dependencies but hear me out on this one. ESlint deprecated a CLI option allowing you to specify additional local rules (`--rulesdir`) in favor of using npm plugins. This package works around it by installing a package that has single LOC:
```js
module.exports = require('../../.eslintplugin')
```
Essentially a npm package that just imports a `.eslintplugin.js` or `.eslintplugin/index.js` from your root, allowing you to have "local" rules.

ESLint then works by analyzing and reacting to AST of the code, so my reasoning (partially described in the code) was this:

1. Look for `extends X` in class declaration - if it doesn't extend, or extends few whitelisted classes, ignore
2. check if there is `static dependencies = {...}` - if it doesn't, ignore
3. if the dependencies are empty => spread parent dependencies
4. if the dependencies aren't empty but don't contain the spread => spread parent dependencies **first**
5. if the dependencies aren't empty and DO contain the spread but not at the top => move the spread to the top

Everything is very literal and hardcoded, meaning if we decided to rename the `dependencies` property, the plugin wouldn't work. Also, the plugin works on file-by-file basis, so it has no way of knowing, if the parent even **has** dependencies to spread, though **not sure how to address that** (truth be told anything that eventually extends `Module` has them, which probably is a majority).

Also, invaluable tools/links in case someone develops additional rules:
- https://astexplorer.net/
- https://eslint.org/docs/latest/developer-guide/working-with-rules


ToppleTheNun:
I would recommend a slightly different approach: add an `eslint-plugin` package and use that instead of `eslint-plugin-local`. This will allow you to add even more rules (e.g., #2386 or #2387) in a flexible manner.

Example of what an index.js could look like for this plugin package:
```javascript
const spreadParentDependencies = require('./spread-parent-dependencies');

module.exports = { ...spreadParentDependencies };
```

sMteX:
Well, AFAIK there's nothing stopping you from including more rules. The structure of the file I added is 
```js
exports.rules = {
   'name-of-the-rule': {
    ..
    }
}
```
So I'm assuming there's nothing stopping you from adding another rule after that but I haven't tested it. Also, considering my package imports from a path, we could add entire folder with rules and merge them in `index.js` too.

emallson:
I went ahead and converted this to a package since it was just moving files around.

emallson (approved):

emallson:
Closing/reopening to try to retrigger the build action

PR #5041, opened 2022-08-15
Converting SPELLS and ITEMS to Typescript

sMteX:
So apparently, typing SPELLS and ITEMS and `indexById` with Typescript opened a whole can of worms. I tried fixing things to the best of my knowledge but it's very likely I missed something or fixed it wrongly.

One big issue I have is `Item` interface (`parser/core/Events` vs `common/ITEMS/Item`) which is used interchangeably but doesn't match at all.


ToppleTheNun:
This implementation of `indexById` mostly works, but breaks the auto-completion `indexById` currently does (at least in WebStorm).

This implementation also works but keeps the auto-completion (assuming all of the current usages of `indexById` are updated suitably):

**src/common/indexById.ts**
```typescript
interface BaseIndexableObj {
  id: number;
  __ignoreDuplication?: boolean;
}

const typedKeys = <T>(obj: T) => Object.keys(obj) as Array<keyof typeof obj>;

const indexById = <Idx extends BaseIndexableObj>() => <T extends { [name: string]: Idx }>(
  arg: T,
): {
  [key in keyof typeof arg | number]: Idx;
} => {
  const indexedByNameAndId: {
    [key in keyof typeof arg | number]: Idx;
  } = { ...arg };
  typedKeys(arg).forEach((key) => {
    const value = arg[key];
    if (process.env.NODE_ENV === 'development') {
      if (
        indexedByNameAndId[value.id] &&
        !indexedByNameAndId[value.id].__ignoreDuplication &&
        !value.__ignoreDuplication
      ) {
        throw new Error(`A spell with this ID already exists: ${value.id},${key}`);
      }
    }
    indexedByNameAndId[value.id] = value;
  });
  return indexedByNameAndId;
};

export const asIndexableList = <Idx extends BaseIndexableObj>() => <T extends { [name: string]: Idx }>(
  arg: T,
) => arg;

export default indexById;
```

**src/common/SPELLS/demonhunter.ts**
```typescript
import { asIndexableList } from 'common/indexById';
import Spell from 'common/SPELLS/Spell';

const spells = asIndexableList<Spell>()({
  // Common Spells
  GLIDE: {
    id: 131347,
    name: 'Glide',
    icon: 'ability_demonhunter_glide',
  },
  CONSUME_SOUL: {
    id: 228532,
    name: 'Consume Soul',
    icon: 'ability_warlock_improvedsoulleech',
  },
  FELBLADE_DAMAGE: {
    id: 213243,
    name: 'Felblade',
    icon: 'ability_demonhunter_felblade',
  },
  // ...
});

export default spells;
```

**src/common/SPELLS/index.ts**
```
const ABILITIES = {
  // ...
  ...DEMON_HUNTER
  // ...
} as const;

// ...
const InternalSpellTable = indexById<Spell>()(ABILITIES);
```

Once all of the spread objects in `ABILITIES` use `asIndexableList`, you can perform auto-completion on both `ABILITIES` and `InternalSpellTable`:
image
image

src/common/indexById.ts
```diff
@@ -0,0 +1,28 @@
...
+        throw new Error('A spell with this ID already exists: ' + ability.id + ',' + key);
+      }
+    }
+    obj[ability.id] = ability;
+  });
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore - couldn't get a combination of [key: string | number]: T + count: number on the return type
+  // it will be assigned in runtime but anywhere it's needed, @ts-ignore it
+  return Object.assign(obj, { count });
```
ToppleTheNun: Since this is only used in one place (`ThroughputPerformance.tsx`), I think it might be better to export the spec count in the `game/SPECS` file directly instead of requiring the use of `@ts-ignore`.

src/common/SPELLS/index.ts
```diff
@@ -77,16 +77,18 @@ const ABILITIES = {
...
 // We should type indexById properly some day to make this standard.
 // And then fix all those errors.
 // Which will prevent bugs.
-const InternalSpellTable = indexById(ABILITIES);
+const InternalSpellTable = indexById(ABILITIES) as SpellCollection;
 // assignment is used here to avoid potential performance pitfalls when
 // compiling the spread operator on large objects.
-InternalSpellTable.maybeGet = (key: string | number | undefined): Spell | undefined =>
-  key ? InternalSpellTable[key] : undefined;
+InternalSpellTable.maybeGet = (key) => (key ? InternalSpellTable[key] : undefined);
```
ToppleTheNun: Could this be defined as a separate export and the current uses of `SPELLS.maybeGet` refactored to use the separate export instead? This was very much relying on the fact that `indexById` was previously returning `any/unknown`.

sMteX:
With the help of @ToppleTheNun we even managed to strongly type `safeMerge()` leading to us having strongly typed `ITEMS` and `SPELLS` with actual autocompletion. This allowed us to even catch some typos that went unnoticed!

A little downside is that indexing e.g. `SPELLS` by some variable either needs it to be number (indexing by spell ID) or some kind of TS ignore when indexing by the `UPPERCASE_SNAKE_CASE` name (because it doesn't have string *index* type anymore). You can use `maybeGetSpell()` which would suit well though.

emallson:
Been meaning to comment on this. That is a good workaround to get autocompletion; i'd prototyped something similar before but it required omitting the type on the class/spec files which was not ideal because it allowed putting broken objects in.

danny-janse:
> With the help of @ToppleTheNun we even managed to strongly type `safeMerge()` leading to us having strongly typed `ITEMS` and `SPELLS` with actual autocompletion. This allowed us to even catch some typos that went unnoticed!
> 
> A little downside is that indexing e.g. `SPELLS` by some variable either needs it to be number (indexing by spell ID) or some kind of TS ignore when indexing by the `UPPERCASE_SNAKE_CASE` name (because it doesn't have string _index_ type anymore). You can use `maybeGetSpell()` which would suit well though.

This sounds like something that will get us all excited! Nice work, cant wait for it to work!

sMteX:
Somehow all hell broke loose with failing integration tests which I'll try to take a look at tomorrow and I'm not *completely* sure if I handled all fixes correctly (I'm *very sorry* to anyone reviewing this), but it's gonna be nice when all is finally resolved and merged.

sMteX:
Alright, looks like this is mostly ready for review. Any maintainer's eyes welcome.

danny-janse:
I really love the change where auto-completion starts working. But im running into a few issues here;

* I cant boot up the branch because several items dont exist during runtime. eg; `ITEMS.RAESHALARE_DEATHS_WHISPER.id` Seems to be specific to either this bow or <ItemLink> in general.

* Looking up the spell and item definitions by clicking on them previously opened the definition of that specifc spell. Now however it opens the `indexById` definition or the interfaced passed to `indexById`. its not that big of an issue really but it will take take a little more time correcting/looking up the id's of spells within WoWA.

sMteX:
The issue with not being able to start was caused by a typo in `safeMerge()` that resulted in those items not being present. That much is fixed now. But truth be told, the other issue might be specific to Webstorm, because even on a branch that doesn't have these updates, VSCode doesn't allow me to Ctrl Click on an individual spell to get to the definition. Fastest way imo would be to global search the "spell name" string.

djanse:
Seems to be some weird stuff going on with the timelines; 
image

analysis/deathknight/src/shadowlands/legendaries/Superstrain.tsx
```diff
@@ -34,7 +34,7 @@ class Superstrain extends Analyzer {
     }
 
     if (
-      [SPECS.BLOOD_DEATH_KNIGHT, SPECS.UNHOLY_DEATH_KNIGHT].includes(this.selectedCombatant.spec)
+      [SPECS.BLOOD_DEATH_KNIGHT, SPECS.UNHOLY_DEATH_KNIGHT].includes(this.selectedCombatant.spec!)
```
Pewtro: Most linters don't allow non-null assertion operators, so I'm not sure if it's something we should allow

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
```diff
@@ -25,7 +26,7 @@ class RuneOfHysteria extends Analyzer {
   constructor(options: Options) {
     super(options);
 
-    const active = this.selectedCombatant.hasWeaponEnchant(SPELLS.RUNE_OF_HYSTERIA);
+    const active = this.selectedCombatant.hasWeaponEnchant(SPELLS.RUNE_OF_HYSTERIA as Enchant);
```
Pewtro: I really dislike this typecasting done in many places, it opens users to errors by wrongly typecasting something as you're essentially forcing typescript to accept a given value as an Enchant even if it doesn't fit an Enchant.

analysis/druidrestoration/src/modules/shadowlands/conduits/GroveInvigorationResto.tsx
```diff
@@ -43,7 +43,7 @@ class GroveInvigorationResto extends Analyzer {
   }
 
   get avgRatingGain() {
-    const buffAtt = this.mastery.getBuffBenefit(SPELLS.REDIRECTED_ANIMA.id);
+    const buffAtt = this.mastery.getBuffBenefit(SPELLS.REDIRECTED_ANIMA.id.toString());
```
Pewtro: This looks suspiciously incorrect... Why does getBuffBenefit take the id as a string?

analysis/druidrestoration/src/modules/shadowlands/items/Tier28_2pc.tsx
```diff
@@ -34,11 +34,11 @@ class Tier28_2pc extends Analyzer {
   }
 
   get directHealing() {
-    return this.mastery.getDirectHealing(SPELLS.RENEWING_BLOOM.id);
+    return this.mastery.getDirectHealing(SPELLS.RENEWING_BLOOM.id.toString());
   }
```
Pewtro: The same question why strings on getDirectHealing and getMasteryHealing

analysis/druidrestoration/src/modules/talents/Abundance.tsx
```diff
@@ -73,7 +73,7 @@ class Abundance extends Analyzer {
       this.stacks.reduce(function (a, b) {
         return a + b;
       }, 0) / this.stacks.length || 0;
-    const avgManaSaings = SPELLS.REGROWTH.manaCost * avgManaSavingsPercent;
+    const avgManaSaings = SPELLS.REGROWTH.manaCost! * avgManaSavingsPercent;
```
Pewtro: Another instance of non-null assertion that we should probably consider a bit more if we want to allow

analysis/huntersurvival/src/modules/core/SpellUsable.tsx
```diff
@@ -15,10 +15,12 @@ class SpellUsable extends CoreSpellUsable {
...
   bombResets = 0;
 
+  raptorMongooseVariantsIds = RAPTOR_MONGOOSE_VARIANTS.map((spell) => spell.id);
+
   onCast(event: CastEvent) {
     const spell = event.ability;
     if (this.selectedCombatant.hasLegendary(SPELLS.RYLAKSTALKERS_CONFOUNDING_STRIKES_EFFECT)) {
-      if (RAPTOR_MONGOOSE_VARIANTS.includes(spell)) {
+      if (this.raptorMongooseVariantsIds.includes(spell.guid)) {
         this.lastPotentialTriggerForBombReset = event;
```
Pewtro, line 24: Why this change?

scripts/conduits/conduitSpells.ts
```diff
@@ -1381,4 +1381,4 @@ export default {
     name: 'Adaptive Armor Fragment',
     icon: 'inv_belt_plate_ardenweald_d_01',
   },
-};
```
Pewtro, line 1384: This file is mostly meant as a place you copy from, you can't reference from it

scripts/talents/generateTalents.js
```diff
@@ -297,16 +297,18 @@ Object.keys(talents).forEach((classId) => {
...
+const talents = asIndexableList<Spell>()(
 ${Object.keys(spellList)
   .map(
     (spec) => `\n  //${spec}${visualizeTalentTree(talentTrees, spec)}
 ${printTalents(spellList, spec)}`,
   )
   .join('\n')}
-};
+});
+
```
Pewtro, line 310: This file will be deleted in the Dragonflight branch, so any change associated with talents in this PR will also be discarded on that branch

src/interface/SpellLink.tsx
```diff
@@ -7,7 +8,7 @@ import SpellIcon from './SpellIcon';
 import useSpellInfo from './useSpellInfo';
 
 interface Props extends Omit<React.HTMLAttributes<HTMLAnchorElement>, 'id'> {
-  id: number;
+  id: number | Spell;
```
Pewtro: An id can be a full Spell object? That doesn't seem correct

src/parser/shadowlands/modules/items/dungeons/ShadowgraspTotem.tsx
```diff
@@ -14,7 +14,7 @@ import Statistic from 'parser/ui/Statistic';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
 import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
 
-const ITEM: Item = ITEMS.SHADOWGRASP_TOTEM;
+const ITEM: Item = ITEMS.SHADOWGRASP_TOTEM as any;
```
Pewtro: as any?

src/parser/shared/modules/SpellUsable.ts
```diff
@@ -393,8 +393,8 @@ class SpellUsable extends Analyzer {
       type: EventType.UpdateSpellUsable,
       ability: {
         guid: spellId,
-        name: spell ? spell.name : undefined,
-        abilityIcon: spell ? spell.icon : undefined,
+        name: spell?.name || '',
+        abilityIcon: spell?.icon || '',
```
Pewtro: can also be written as `spell.name ?? ''`. Not a requirement at all, just a note, if you'd prefer something like that

Pewtro (commented):
A few comments here and there, some of them (the .toString() and the non-null assertion are only put a few places but they go again quite a few times.). 

Major thing I guess is that the DF branch will throw out any change to talents done here with it's own TALENTS/ folder instead of SPELLS/talents, and brings it's own strict typing method through utilising `type EnforceObjectType<T> = <V extends T>(v: V) => V;` and `createTalentList: EnforceObjectType<{ [key: string]: Talent }> = (v) => v;`

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
emallson: this is definitely something where the merging into the SPELLS object needs to be smarter---it *is* possible to not erase the types of individual properties during the merge, but it requires some clever omission of types to let TS infer the exact values.

src/common/indexById.ts
```diff
@@ -0,0 +1,43 @@
...
+interface NameIndexed<T> {
+  [name: string]: T;
+}
+
+// For ease of use we may want to be able to both access items by code names for completion/compile time checking (e.g. `ABILITIES.AURA_OF_MERCY_TALENT`) and by spell id for runtime evaluation (e.g. `ABILITIES[183415]`)
+const indexById = <Idx extends BaseIndexableObj>() => <T extends NameIndexed<Idx>>(
+  arg: T,
+): {
+  [key in keyof typeof arg | number]: Idx;
+} => {
```
emallson: the typing of this method is what erases the specific types of keys, changing them from e.g. `Enchant` to `Spell`. This can likely be recast as:

```ts
function indexById<E extends BaseIndexableObj, T extends Record<string, BaseIndexableObj>>(originalMap: T): T & { [id: number]: E }
```

but i'll fiddle around with the specifics of this later.

analysis/deathknight/src/shadowlands/legendaries/Superstrain.tsx
sMteX: To be fair, I wasn't sure to do in places dealing with specs. `Combatant.spec` is deprecated, but `Combatant.player.icon` is supposed to replace spec? I haven't tested if the values match. I can likely take a look at this.

analysis/deathknight/src/runeforges/RuneOfHysteria.tsx
sMteX: I'm afraid that's outside of my abilities (even current implementation is courtesy of @ToppleTheNun) but if you're able to come up with that, I'll more than gladly include it.

analysis/druidrestoration/src/modules/shadowlands/conduits/GroveInvigorationResto.tsx
sMteX: `getBuffBenefit()` and its sister methods take `healId: string` that's an index into `MasteryAttributionsBySpell` object (`[key: string]: MasterySpellAttribution`) and the `SPELLS.x.id` is a number. I guess before when it was `any` it did the conversion implicitly

Pewtro: surely the key should just be a number if it's always spellid?

analysis/druidrestoration/src/modules/talents/Abundance.tsx
sMteX: I agree but not sure how to approach it here. `Spell.manaCost` is optional, so is it better to do `(SPELLS.REGROWTH.manaCost || 0) * avgManaSavingsPercent` instead? I did verify if the actual property exists for that specific object.

This applies for `manaCosts` and other properties in other places too (saw something with monks and priests)

analysis/huntersurvival/src/modules/core/SpellUsable.tsx
sMteX: `RAPTOR_MONGOOSE_VARIANTS` is a `Spell[]`, but `event.ability.spell` is of type `Ability`, couldn't compare with `includes`

scripts/conduits/conduitSpells.ts
sMteX: Fair, I didn't see the comment while copypasting the same thing across a billion files 😄

src/interface/SpellLink.tsx
sMteX: In a lot of places, `SpellLink` is used with the entire `Spell` object (in fact, about 48 places). Even the comment mentions it 😄 I could go through all of them specify the `.id` and this could go away, if that's better.

src/parser/shadowlands/modules/items/dungeons/ShadowgraspTotem.tsx
sMteX: Hmm, I initially meant it as a hack around inconsistencies of `Item` interfaces, but it might not even be needed seeing as this `ITEM` const is used only for its item ID. Will refactor

abelito75:
I'm removing myself from `required` feedback. Please don't readd me

this sounds super toxic... Its not meant to be. Just don't have time to commit good feedback or work on wowA currently

analysis/druidrestoration/src/modules/shadowlands/conduits/GroveInvigorationResto.tsx
sMteX: I suppose it could, it seems to be indexed by numbers mainly. Will request a review from @kfinch then 😄

kfinch: I think when I wrote the Mastery module, we were still on JS and I didn't realize I was allowed to key things as numbers. Leaving this as `toString` for now is fine, and I'll fix this in a follow-on PR.

sMteX: I tried to fix that for you, if you could take a look at the newest commit @kfinch

kfinch: Ah yeah that should be all it takes, thanks!

analysis/druidrestoration/src/modules/talents/Abundance.tsx
ToppleTheNun: One way that this _could_ be resolved is using a discriminated union, but that ends up being a much larger refactor. Example, just for completeness:

```typescript
interface BaseSpell {
  id: number;
  name: string;
  icon: string;
  cost: number;
}

interface ManaSpell extends BaseSpell {
  resource: RESOURCE_TYPES.MANA;
}

interface RageSpell extends BaseSpell {
  resource: RESOURCE_TYPES.RAGE;
}

// other spell cost types

type Spell = ManaSpell | RageSpell /* add additional types here */;
```

Usage would be somewhat similar to how it is now, but every spell would need to specify its resource and the cost.

sMteX: Oh sweet, the thing we did with the Spell typings make this non-null assertion obsolete since it knows `REGROWTH` has mana cost

sMteX:
I think this is ready for another round of review @Pewtro @emallson . I'm still aware of the failing integration test on surv hunter and will try to figure out why.

sMteX:
So the different snapshot was likely caused by this, in Survival Hunter `SpellUsable`
image

The original impl was comparing `Ability` object with `Spell` array, whereas in the new, we're comparing numbers. Verified by putting a breakpoint inside the condition, old one was never fired whereas the new one fires all the time (after all it's Raptor/Mongoose strike, one of the rotational abilities). The mismatching ability was Pheromone Bomb and this looks like implementing some cooldown resetting effect on the bombs.

emallson (approved):
if the survival snapshot change looks okay to @Pewtro then I think this is good to go

Pewtro (approved):
I could've sworn Rylaks worked in the past, but I'll assume this is fine too

kfinch (approved):
Don't have time to fully look this over, but based on what I see and what I've read in the Discord, this looks good to me

PR #5045, opened 2022-08-17
Bad Sundering casts

danny-janse:
One wouldn't think it'd happen but, yet as seen in [this report](https://www.warcraftlogs.com/reports/kAjvaxWdPbpnf8cK#fight=1&type=damage-done&source=12&ability=197214), sundering can miss!

Tricky with sundering is that the damage can be dealt to multiple enemies. So you cant just do `this.casts - this.hits`. Which let me to the solution of recording all cast events as a miss and removing them if damage was dealt within the same timespan.

* This adds a suggestion, Kinda hard to balance but in general its not the end of the world when it happens. just shouldn't happen concistantly.
image

* And marks the cast as bad on the timeline
image

* Tooltip on how many misses. 
image

so sorry about the amount of commits...


danny-janse:
Checking the appropriate treshold as before this can be merged. its more of a minor thing to miss it.

danny-janse:
This is ready for review now.

sMteX:
LGTM 👍🏼

emallson (approved):

PR #5046, opened 2022-08-18
Resolve NaN item level

danny-janse:
Noticed that ~~bugged~~ characters without gear would result in calculating NaN for item level. See ReportCompensition and PlayerTile for [this report](https://wowanalyzer.com/report/J64knxvhHqVfWAb1/3-Heroic+Vigilant+Guardian+-+Kill+(5:37))

This changes it so that
* Players without gear have an iLvl of 0 (zero)
* The average raid comp ilvl is no longer based on them

image


emallson (approved):
ah yes, combatantinfo bugs. how we love these :upside_down_face:

PR #5048, opened 2022-08-19
SPELL_CATEGORY enum

danny-janse:
\+ label Core

While working on priority suggestions for Enhancemnt Shamen i noticed that the category was a 'computed' string per spell in the spellbook. And would be computed in every compare done to it. This in itself seemed inlogical/inefficient. Apparently i wasn't the only one who thought this as someone suggested in the a comment `TODO - Enum?`.

This is also more type gated since the previous setup allowed for any string to be used.  For both setting the the value and comparing. eg; `HealingEfficiencyTracker.ts:152`

Most of the files are a simple search-replace. Files with notable changes are;
* `src/parser/ui/CastEfficiency.ts`
* `src/parser/core/SPELL_CATEGORY.ts`
* `src/parser/core/Ability.ts`


src/parser/core/SPELL_CATEGORY.ts
```diff
@@ -0,0 +1,74 @@
...
+        message: `Damaging Spell`,
+      });
+    case SPELL_CATEGORY.CONSUMABLE:
+      return t({
+        id: 'core.abilities.spellCategories.consumable',
+        message: `Consumable`,
+      });
+  }
+
+  return 'Unknown Spell Category';
```
sMteX: This could be `default` case in the switch. Also missing translation here.

src/parser/core/modules/Abilities.ts
```diff
@@ -88,6 +40,7 @@ class Abilities extends Module {
     // Abilities subtypes may want to construct a particular subtype of Ability
     const abilityClass = (this.constructor as typeof Abilities).ABILITY_CLASS;
     this.abilities = spellbook.map((options) => new abilityClass(this, options));
+    console.log(this.abilities);
```
sMteX: Forgotten log here

sMteX (commented):
I didn't really review all the changes in various Abilities, but as you said, it might be just replace `Abilities.SPELL_CATEGORIES` -> `SPELL_CATEGORY` + import 😄 Otherwise it seems fine apart from those 2 tidbits

src/parser/core/modules/Abilities.ts
danny-janse: Oops completely missed that. Thanks for spotting this.

src/parser/core/SPELL_CATEGORY.ts
danny-janse: I'm personally against translating this as it is primairly a developers thing to notice when changing anything in this file. 
What's a maintainers perspective on return statement? Should this be in or outside the switch-statement?

emallson: We shouldn't translate strings that are for devs only IMO. extra work for no value since nobody should ever see them.

As far as the return vs default, either is fine with me. `default` might be a little cleaner but i'm not going to fuss over that.

emallson: also i did the same thing for reviewing as @sMteX so as long as snapshots still pass i think this is okay after deciding what you want to do for default vs return

danny-janse: Changed to a default: return in favor of argument of cleaner. Thanks both of you! snapshots are pretty much the same. I had an error during development but fixed to code to be the same result again.

danny-janse:
Oh let met checkout the branch locally to fix lint issues

emallson (approved):

PR #5051, opened 2022-08-20
SpellUsable overhaul

kfinch:
Putting up a mostly complete draft of an overhaul of SpellUsable. The goals were to add tons of documentation and to reduce the amount of ambiguous behavior - the public interface is largely the same.

Key Changes:
* Tons of documentation on everything
* Added support for 'modRate' changes
* Removed superfluous fields from `UpdateSpellUsableEvent` and renamed the kept ones to make them more clear
* Changed `UpdateSpellUsableEvent.trigger` to `updateType`, better reflecting what it actually is
* updateType `AddCooldownCharge` renamed to `UseCharge`, better reflecting what it actually is
* Re-ordered the fields on some of the public functions to allow for easier calling

What still needs to be done:
* More testing to make sure nothing broke
* Handling for 'filtered info events' that apparently come up with analyzing by phase (do we do this anymore?)

What could be done but maybe I leave it for another PR?
* Implement default handling for common CD special cases in order to reduce the amount of copy pasted code in specs


src/parser/core/modules/Ability.ts
```diff
@@ -29,9 +29,12 @@ export interface SpellbookAbility<TrackedAbilityType extends TrackedAbility = Tr
...
-   * for more complicated calls or even to check for buffs. Parameters
-   * provided: `hastePercentage`, `selectedCombatant`
+   * The cooldown of a spell at the time of the cast. Unlike most other durations in WoWA,
+   * this is in *seconds, NOT milliseconds*. This can be the direct number, or it can be a function
+   * for cooldowns that scale with haste, and can also branch based on combatantinfo.
+   * @param haste the player's haste at the time of spell cast, as a proportion of
+   *   normal casting speed (20% haste will be passed as 1.20).
+   * @param trigger TODO what the hell is this and what is it for?
    */
   cooldown?: ((haste: number, trigger?: AnyEvent) => number) | number;
```
kfinch: I still can't figure out who wants the `trigger` in cooldown - if the answer is "no one", I'd like to remove it

kfinch:
Some common things spec extensions of SpellUsable implement that I'd like to make standardized:
* Cooldown starts on something other than the cast event (e.g. healer dispel only goes on cooldown when Dispel actually happens)
* Cooldown has a reset based on an undetectable proc (lots of cases of this)

src/parser/core/modules/Ability.ts
emallson: no tests fail from removing it, and typecheck passes with minimal changes (removing references to the parameter from the `Ability` and `Abilities` modules in core). I'm good removing it.

src/interface/report/Results/Timeline/Cooldowns.tsx
```diff
@@ -32,7 +32,7 @@ class Cooldowns extends PureComponent<Props> {
     return (
       <Lane
         key={spellId}
-        spellId={spellId}
+        // spellId={spellId}
```
emallson: deleto

emallson (approved):
Assuming snapshots pass, I think this looks great. First-class support for modrate is awesome and a major gap in the existing core. The amount of documentation is wonderful and will hopefully make it easier for people to correctly use `SpellUsable`. Changing the internal representation to better match the info that we actually need for the inner workings of the core is also a nice boon :+1: 

so yea, if tests pass I honestly think this is good to go.

emallson:
> Handling for 'filtered info events' that apparently come up with analyzing by phase (do we do this anymore?)

we do still support this

PR #5052, opened 2022-08-21
Rewrote BDK modules to Typescript

sMteX:
As an added bonus, translated all strings too. Functionality should be mostly the same, I tried to simplify some things I spotted too.


emallson (approved):
:+1:

it would be easier to review in the future if you do a handful of modules at a time instead of a whole class--especially since the introduction of translations rewrites a lot of the snapshot

PR #5055, opened 2022-08-22
Death Knight translations

sMteX:
Plus occasional convert to Typescript


emallson:
@sMteX looks like there are a bunch of conflicts with the generated JSON translation files. Not sure if that is indication of their being conflicts between this and your BDK PR?

Issue #5050, opened 2022-08-20
What to do with mega objects ITEMS and SPELLS?


sMteX:
There has been a discussion over the items, provoked by PR #5041. For the discussion to not get lost in #development, I'm starting this issue.

Known issues and why it's a good idea to figure something out with it:
- while the referenced PR managed to strongly type the objects (so now the autocompletion offered by IDEs should work), *indexing* the object is quite simply a nightmare (pressing Ctrl + Space on `SPELLS` takes me upwards to several seconds)
- specific modules (**not** core ones) rarely need to have reference to any kind of spells, usually just a tiny subset (e.g. spec/class spells, or just dungeon trinkets etc.)
- additional properties like `icon` or `name` are also very rarely used (mostly for UI components like `SpellLink` or debugging reasons (`debug && console.log(SPELLS[spellId].name)`))
- indexing by ID itself is also seldom used (few modules use that, but it can be refactored to a local `Spell[]` of those concerned spells, and indexing just a few objects, instead of 3000+), but is still quite often in core modules

Mainly a few ideas I get regarding this, feedback is welcome:
- spec modules should be fine with importing just a subset of SPELLS/ITEMS they need (`import SPELLS from 'common/SPELLS/druid'`)
- I think we should also be fine with even removing the objects themselves and moving to spell IDs only as that is usually enough
    - if additional properties are needed (like `manaCost`, `effectId` for enchants), I think a separate exported object should be pretty clear 
    ```ts
    const manaCosts = { 
      [SPELLS.VIVIFY]: 1500, 
      ... 
    } as const;
    export default manaCosts;
   ```
- for those few places where getting name/icon just by spell ID without any knowledge of what it could be (so for core modules), maybe we can use some external API for that? `SpellLink` already has a fetch for that kind of info (but primarily uses `SPELLS` still, while it's fetching in the background)

sMteX:
Copying @emallson's thoughts from Discord:

> - splitting SPELLS is good, progress on that is desireable
> - we effectively never use the name/icon data from SPELLS except when debugging locally, so I could see us dropping it.
> - switching the type ffrom Spell to number would be a loss
> - pulling spell data from dbc would be awesome, but likely hard to effectively integrate into the build process

emallson:
[...]

PR #5056, opened 2022-08-23
[Demon Hunter] update modules to use class/spec specific spells

ToppleTheNun:
update to use class/spec specific SPELLS modules instead of all-inclusive SPELLS object,
as discussed in #5050.


src/common/SPELLS/shadowlands/legendaries/index.ts
```diff
@@ -12,7 +11,7 @@ import SHAMAN from './shaman';
 import WARLOCK from './warlock';
 import WARRIOR from './warrior';
 
-const legendaries = spellIndexableList({
+const legendaries = {
```
emallson: you should apply the `legendarySpellIndexableList` here, instead of in the per-class list.

src/common/SPELLS/shadowlands/legendaries/demonhunter.ts
```diff
@@ -1,4 +1,6 @@
-const legendaries = {
+import { legendarySpellIndexableList } from 'common/SPELLS/Spell';
+
+const legendaries = legendarySpellIndexableList({
```
emallson: this should just be the bare list of legendary objects.

emallson (changes requested):
mostly looks okay, one comment on the way you reapplied the `spellIndexableList` method, which will cause non-DH legendaries to become unindexable by id.

I think this PR is an interesting look at what a transition to these objects means. Would like to see more `import { FOO } from SPELLS` in new modules but for transitioning existing code I think this approach is fine.

emallson (approved):

PR #5059, opened 2022-08-25
Cooldown History

kfinch:
A simple module to allow for querying the cooldown history. I tested it locally but don't have any direct uses for it yet - Sharrq requested this feature and suggested he would use it.


Sharrq:
Looks good to me

Sharrq (approved):

kfinch:
I included a minor fix in this PR where cooldown end after a haste change is rounded to the nearest millisecond - this is what's causing the very very tiny difference in integration test results

Issue #5049, opened 2022-08-19
Move away from a monorepo setup for analysis folders


Pewtro:
## Why? 

I love monorepos. I use them often, but the way it's set up in this repo makes very little sense to me. We have 57 folders in the analysis folder each being their own package, but we don't utilise the individual package for anything. We don't have a setup where we can release an individual package and have unique versioning for each package. So why is it in a monorepo package structure? It just doesn't make sense to me to introduce more complexity with that structuring to have, to my knowledge, no gain. If we are not planning to start properly utilising having multiple packages in the nearest future, then I don't see why we have separated out source code from the source code folder. I don't why we have created all those packages when all we do is build it all together at once anyways, and deploy all of it every time anyways. The entire point of monorepos is to make it easier to design and maintain microservices, but that isn't what WoWAnalyzer is - it's all deployed as one. Not to mention that having all these packages also leads to a bunch of issues with autocompletion and auto importing as seen in #4905. 

## My suggestion

Return to having class based folder structure under the src folder, so we'd have something like this (analysis folder can be called whatever): 

> src/
> ├─ analysis/
> │  ├─ class/
> │  │  ├─ spec/
> │  │  ├─ shared
> │  ├─ class2/
> │  │  ├─ spec2
> │  │  ├─ shared2

This leads to less folder bloat, it's easier to focus your work on a single area when working on a class as a whole. 

## What about classic(s)? 
As for what to do with TBC? Probably just remove it on DF branch because then WotLK is out, and then it can live together with Shadowlands on that branch. It seems incorrect to try and support multiple expansions in the same repo. But they could also just live in a tbc/ folder, the important bit is to move back into src and reduce the package/folder bloat.

ToppleTheNun:
> As for what to do with TBC? Probably just remove it on DF branch because then WotLK is out, and then it can live together with Shadowlands on that branch. It seems incorrect to try and support multiple expansions in the same repo. But they could also just live in a tbc/ folder, the important bit is to move back into src and reduce the package/folder bloat.

If we maintain branches for each expansion (see Legion due to Legion Timewalking), would it make sense to have classic also be in its own branch?

sMteX:
Agreed and one point I thought of - if part of the appeal of monorepo is nice import paths like `@wowanalyzer/druid`, we could achieve that with just aliases in tsconfig for example.

Pewtro:
[...]

Issue #4905, opened 2022-05-04
Fix auto-completion / auto-import


emallson:
Currently, auto-completion doesn't work in most files unless the object has been imported in one of your open files.

There are two causes of this:

1. The sub-packages don't work with autocomplete unless you create a `tsconfig.json` with `{}` as its content
2. Sub-folders don't work with autocomplete. Part of the issue is linking and autocompletion turning off for anything that is in the `node_modules` folder, but there are other issues here as well that need to be sorted.

This is a major annoyance, and has a larger impact on newer contributors that won't know everything that is available to them in the codebase.

This is *done* when VSCode auto-imports work across the project. Check that it works in:

- [ ] `analysis/hunterbeastmastery/src/modules/dummy.ts`
- [ ] `src/parser/core/dummy.ts`
- [ ] `src/dummy/dummy.ts`
- [ ] `src/dummy.ts`

This may involve refactoring the project back into a monopackage.

PR #5061, opened 2022-09-05
DF Re-org

emallson:
This incorporates all the changes from @Pewtro's PR, plus a reorganization back into a mono-package (see #5049). I haven't fully tested that this resolves the issues with auto-imports (see #4905) but will before merging.

In the process of doing this, I discovered that no tests were running in the `analysis` folder *except* for integration tests. This is now fixed.

As part of prep for DF, there are several additional TEMPORARY changes:

- All integration test files have been removed. New integration tests will need to be set up for supported specs.
- All broken tests have been annotated with `.skip` (+ other changes needed to satisfy the linter/tsc). These will need to be re-enabled later. most of the tests that have broken are broken due to talent changes.
- All retail specs have been fully disabled. This means that TypeScript will not load or parse the files, they will not appear on the site in dev or production mode, etc. In order to re-enable a spec, uncomment its import and export in `src/parser/index.ts`.
  - This required setting the analysis folder to be excluded in `tsconfig.json`, which is not ideal long-term. Exclusions do not apply to files that are imported from a non-excluded file, so this effectively excludes everything that isn't included in the bundle. We will want to remove this exclusion before release.

## Testing this Branch

1. Pull the branch
2. `yarn install`
3. `yarn start`

Make sure to run `yarn test` as well.

We need a windows user to pull this branch and make sure that I didn't break anything with paths.

## Enabling Support for a DF Spec

1. Remove the `exclude` line for the spec and class from `tsconfig.json`.
2. Un-comment the import/export lines for the spec in `src/parser/index.ts`.
3. Fix all missing/invalid talent imports. Delete liberally.
4. Re-enable all of the spec's skipped tests and fix any issues with them. Delete if they test now-dead code.
5. You are now prepared to work on Dragonflight support :)


emallson:
@Pewtro it wouldn't let me set the merge target to your branch, so this is a fully separate PR.

rp4rk (approved):
full send

kfinch (approved):
I mean sure why not lol

danny-janse:
Windows (11) user here;

* Yarn properly installs
* Application starts
* Following classic classes do not work;
  * Hunter 
image
  * Other classes had no issues

`yarn test` summary
image

I did noticed that the analysis folder in the root still exists after checking out this PR. This is due to the node_modules folders created for each subfolder. All of these folders were empty and could be safely removed.

Everything else seems to work as intended.

emallson:
I also tested and this does appear to fix the auto-import issues

kfinch:
Windows user here. I just pulled the branch and used `yarn start` and `yarn test`. 'Start' launched the website without issues, and all the (activated) tests passed.

When looking at the code in WebStorm I am having a lot of import errors however.

kfinch:
After removing the exclude in tsconfig, my IDE loads spec stuff properly

emallson:
After digging around, import errors seen by @kfinch and @darthelit seem to be related to the tsconfig exclusion of spec folders. Once you remove the exclusions, they go away. Moved that step up to #1 in the instructions. going to merge

PR #5065, opened 2022-09-07
add 'export const' for every generated talent

ToppleTheNun:
enables the use of single imports like the below while maintaining existing behavior:
```tsx
import {THE_HUNT_TALENT} from 'common/TALENTS/demonhunter';
```

### DIFF
--- a/scripts/talents/generate-talents.ts
+++ b/scripts/talents/generate-talents.ts
@@ -5,6 +5,7 @@ import {
   createTalentKey,
   csvToObject,
   findResourceCost,
+  printTalentConstExports,
   printTalents,
   readCsvFromUrl,
   readJsonFromUrl,
@@ -126,7 +127,9 @@ async function generateTalents() {
     fs.writeFileSync(
       `./src/common/TALENTS/${lowerCasedClassName}.ts`,
       `// Generated file, changes will eventually be overwritten!
-import { createTalentList } from './types';
+import { createTalentList, Talent } from './types';
+
+${printTalentConstExports(talentObjectByClass[lowerCasedClassName])}
 
 const talents = createTalentList({${printTalents(talentObjectByClass[lowerCasedClassName])}
   });
--- a/scripts/talents/talent-tree-helpers.ts
+++ b/scripts/talents/talent-tree-helpers.ts
@@ -59,6 +59,23 @@ export async function readCsvFromUrl(url: string) {
   return res;
 }
 
+export function printTalentConstExports(talentObj: ITalentObjectByClass[string]) {
+  if (!talentObj) {
+    return "\n//Class doesn't exist in data yet\n";
+  }
+  return Object.entries(talentObj)
+    .map(([specName, specTalents]) => {
+      const header = `\n  //region ${specName}`;
+      const footer = '\n  //endregion';
+      const talents = Object.keys(specTalents)
+        .map((talent) => `export const ${talent}: Talent = ${JSON.stringify(specTalents[talent])};`)
+        .join('\n');
+
+      return [header, talents, footer].join('\n');
+    })
+    .join('\n');
+}
+
 export function printTalents(talentObj: ITalentObjectByClass[string]) {
   if (!talentObj) {
     return "\n//Class doesn't exist in data yet\n";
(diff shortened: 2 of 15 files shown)


emallson:
I like this in theory (I say, having suggested directly this approach on your previous DH pr), im going to play around with the PR locally. but it has occurred to me that since autocompletion and auto import work everywhere now, that we may suddenly have thousands of spells in our auto import suggestions if we do this :thinking:

ToppleTheNun:
> I like this in theory (I say, having suggested directly this approach on your previous DH pr), im going to play around with the PR locally. but it has occurred to me that since autocompletion and auto import work everywhere now, that we may suddenly have thousands of spells in our auto import suggestions if we do this 🤔

I'll play around a bit with it and see how it feels in some of the DH analyzers that I'm already working with; I think this is a nice middle ground of the two approaches where we can have the individual imports while maintaining the class/spec specific SPELLS objects.

I might also mess around with removing the large object as well; it might not be necessary anymore if we're `export const`-ing everything.

I do think we might miss out on some type enforcement by not using the `createTalentList` function in the file.

ToppleTheNun:
> I'll play around a bit with it and see how it feels in some of the DH analyzers that I'm already working with; I think this is a nice middle ground of the two approaches where we can have the individual imports while maintaining the class/spec specific SPELLS objects.

It's _very_ fast.

ToppleTheNun:
Example auto-completion with "DEMONIC", which is a fairly popular prefix. It was done in under half a second:
image

ToppleTheNun:
I will also say that it feels much better to do the direct imports in #5066.

Pewtro:
> > I like this in theory (I say, having suggested directly this approach on your previous DH pr), im going to play around with the PR locally. but it has occurred to me that since autocompletion and auto import work everywhere now, that we may suddenly have thousands of spells in our auto import suggestions if we do this 🤔
> 
> I'll play around a bit with it and see how it feels in some of the DH analyzers that I'm already working with; I think this is a nice middle ground of the two approaches where we can have the individual imports while maintaining the class/spec specific SPELLS objects.
> 
> I might also mess around with removing the large object as well; it might not be necessary anymore if we're `export const`-ing everything.
> 
> I do think we might miss out on some type enforcement by not using the `createTalentList` function in the file.

I’d be pretty opposed to removing the talents object. 

To expand a little bit, providing options to contributors is fine. However when playing around with the new talents structure (either all, or that of a specific class) I found there was no need to drill further down after playing with both spec specific and every spell exported locally, that’s why I didn’t do it. 

I, personally, disagree with adding the export const to every spell object because it’s a lot of unnecessary boilerplate imo that clutters regular auto-completion. 

The boilerplate is something I can live with because of the generated nature of these files, but I’d be really miffed if that became the standard in spells too. 

The cluttering of auto-completion… much less so.

Thus making the exported const the ONLY option, is just not a tenable solution to me

ToppleTheNun:
I'm inclined to agree as doing the export for every single spell will take an obscene amount of time.

kfinch:
I just played around with it and dislike the `export const` for purely aesthetic reasons:
```
const PERIODIC_DAMAGE: SpellInfo[] = [
  SPELLS.ENTANGLING_ROOTS,
  SPELLS.RIP,
  SPELLS.RAKE, // adaptive swarm also boosts the direct damage, so no need for 'tick' differentiation
  MASS_ENTANGLEMENT_TALENT,
  SPELLS.THRASH_FERAL,
  SPELLS.MOONFIRE_FERAL,
  SPELLS.MOONFIRE_DEBUFF,
```

ToppleTheNun:
> I just played around with it and dislike the `export const` for purely aesthetic reasons:
> 
> ```
> const PERIODIC_DAMAGE: SpellInfo[] = [
>   SPELLS.ENTANGLING_ROOTS,
>   SPELLS.RIP,
>   SPELLS.RAKE, // adaptive swarm also boosts the direct damage, so no need for 'tick' differentiation
>   MASS_ENTANGLEMENT_TALENT,
>   SPELLS.THRASH_FERAL,
>   SPELLS.MOONFIRE_FERAL,
>   SPELLS.MOONFIRE_DEBUFF,
> ```

One benefit that the `export const` has over the current implementation (likely to be fixed) is that the talents returned from the `talents` export are currently coming back as mapped types and thus can't be assigned to anything that's expecting a `Spell`.

ToppleTheNun:
Closing this as it has more downsides than upsides.

PR #5066, opened 2022-09-08
[WIP] initial pass at DH updates

ToppleTheNun:
update all analyzers to use new generated talents

add talent scaling constants for shared, havoc, veng

update CONFIG.tsx for havoc and vengeance

clean out old changelog entries pre-2022

migrate all demon hunter spells into DH SPELLS file

currently dependent on #5065


ToppleTheNun:
Closing this as #5067 does the same thing with less changes.

PR #5070, opened 2022-09-09
Enhancement Shaman - Initial Dragonflight swipe

danny-janse:
\+ label Shaman, Dragonflight

This is an initial swipe for the dragonflight exp.

* Removed SPELL entries that have TALENT entries.
* Replaced `SPELLS` with `TALENTS` where required.
* Removed, removed talents (so akward to write)
* Re-organized the shared folder.
* Updated the Spellbook as much as possible
* Made sure each talent has a `this.active` property assigment and `return` statement
* Disabled some code that is _shared with/borrowed from_ restoration
* Disbaled Apl Check
  * Need to figure out how to pass in the Talent object

The app starts without trouble but crashes with Shadowlands logs being parsed. 
I've not been able to obtain/create a log for Dragonflight.


ToppleTheNun:
> Need to figure out how to pass in the Talent object

This is an issue currently with how `asRestrictedTable` types the return. Currently the easiest way is to just use `.id` where possible.

emallson (approved):

PR #5075, opened 2022-09-13
Dragonflight - activate Resto Druid

kfinch:
This just removes some shadowlands only stuff and fixes all the talents and makes sure it builds. I made some updates for the way things will work in DF, but still more work to be done there. This also includes a partial reorganization of the folder structure which I will continue in future PRs.

I regenerated only the Druid talents so as not to interfere with custom changes made to other classes talent trees.


src/analysis/retail/druid/restoration/modules/features/Checklist/Component.tsx
```diff
@@ -11,7 +10,9 @@ import GenericCastEfficiencyRequirement from 'parser/shared/modules/features/Che
 import Requirement from 'parser/shared/modules/features/Checklist/Requirement';
 import Rule from 'parser/shared/modules/features/Checklist/Rule';
 import PropTypes from 'prop-types';
+import { TALENTS_DRUID } from 'common/TALENTS';
 
+// TODO remove in favor of Guide (once UI for this figured out)
```
emallson, line 15: fwiw i went ahead and nuked the brew checklist. don't need to for resto (yet) but it made my life a lot easier

emallson (approved):

PR #5083, opened 2022-09-23
convert CharacterParses to TypeScript

ToppleTheNun:
also make regions slightly safer.

### DIFF
--- a/src/common/regions.ts
+++ b/src/common/regions.ts
@@ -11,6 +11,11 @@ export function isRegion(region: string | undefined): region is Region {
   return ALL_REGIONS.includes(region as Region);
 }
 
+/**
+ * Not every region is supported by the Blizzard API (and, by extension, the bits of the
+ * WoWA API that use the Blizzard API). This is used to check that a region is supported before
+ * sending a request.
+ */
 export function isSupportedRegion(region: string | undefined): region is SupportedRegion {
   // the 'as' below is just to allow the use of .includes
   return SUPPORTED_REGIONS.includes(region as SupportedRegion);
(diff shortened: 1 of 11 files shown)


src/common/regions.ts
```diff
@@ -0,0 +1,17 @@
...
+// WoWAnalyzer/Blizzard API supported regions
+export const SUPPORTED_REGIONS = ['EU', 'US', 'TW', 'KR'] as const;
+export type SupportedRegion = typeof SUPPORTED_REGIONS[number];
+
+export function isRegion(region: string | undefined): region is Region {
+  // the 'as' below is just to allow the use of .includes
+  return ALL_REGIONS.includes(region as Region);
+}
+
+export function isSupportedRegion(region: string | undefined): region is SupportedRegion {
```
emallson, line 19: this should get a docstring and/or name update clarifying that this is related to API support, not analysis support.

```suggestion
/**
 * Not every region is supported by the Blizzard API (and, by extension, the bits of the WoWA API that use the Blizzard API). This is used to check that that a region is supported before sending a request.
 */
export function isSupportedApiRegion(region: string | undefined): region is SupportedRegion {
```

src/common/typeGuards.ts
```diff
@@ -0,0 +1,3 @@
+export function isDefined<T>(t: T | undefined): t is T {
```
emallson, line 1: location of this feels weird, but i like the way it simplifies the code that uses it so it stands for now i think

emallson (changes requested):
just a change to make sure the supported regions code doesn't get misunderstood, then we're good imo

emallson (approved):

PR #5089, opened 2022-09-26
Dragonflight Resto Druid - Implemented several new talents and updated Guide strings.

kfinch:
Resto Druid analyzer now mostly supported, so I bumped version to '10.0'. There's still a few missing things and obviously I'll have to stay on top of any further changes during beta, but I think we're in a good enough state for this to be useful for Dragonflight players.

This pull request also includes a reorganization of talent stat boxes, and an update to the stat rating per percent constants (to match level 70 players)

Most of the existing guide and statistics didn't need to be changed - here are some screenshots of changed bits:
image
image
verdancy_wip2


kfinch:
Just gonna merge this actually to submit to Dreamgrove people for feedback

PR #5094, opened 2022-09-27
APL Guide Section

emallson:
image
image

## What is this?

This adds a shared `AplSectionData` component that uses an [apl checker](https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/APLCheck) to automatically generate suggestions.

*Note that the changing talent trees on beta interact badly with the APL code---if you load an old log it may not correctly detect that you have a talent, and then the APL can break in weird ways since it won't require you to cast a spell that you don't have.*

In particular, it includes:

- The `AplSectionData` component, which shows the summary of the priority list, detection of some basic problems, and narrowed cast timelines for mistakes.
- An API for adding your own APL-based suggestion generators beyond the basic ones included by default.
- The introduction of [emotion](https://emotion.sh/docs/introduction) into the codebase for future use in styling. This has been discussed in `#admin` in Discord previously.
- Monk talent updates
- The addition of APL support for "multi-spell rules" where it is correct to cast one of several spells. This ended up having a...larger cost than I'd expected and could be persuaded to remove it, since I'm not sure anyone needs it besides Brewmaster.

(Monk talent updates + adding emotion account for a large fraction of the added lines)

Not included:

- The experimental timeline rewriter that I'd been working on (and shown off in discord). This has too many failure modes to release right now, even though it would *often* produce useful output.

## Why is this?

When I added the [apl checker](https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/APLCheck) last year, a large part of my motivation was to remove the need to hand-tune `castEfficiency` suggestions, implement manual filler cast analysis (see e.g. the Swipe analyzer for Guardian and Tiger Palm analyzer for Brewmaster), and other related busywork.

The central idea is that for *many* (not all, *many*) specs, an APL can give you high-quality data about which spell casts were correct. However, the checklist output that has [previously been used](https://user-images.githubusercontent.com/4909458/137603890-9753e06b-a6f8-4a9e-a2a1-b6431f8749b7.png) leaves a lot to be desired---it is difficult to determine what you should change in order to improve the results.

This guide section helps address both of these problems:

1. It can generate a LOT of useful analysis for APL-driven specs just by writing out the APL. This removes a lot of boilerplate, and means that maintainers don't need to try to fine-tune cast efficiency---reducing the work required.<br/>*This should also avoid situations where differences in fight structure, such as downtime and intermissions, totally break cast efficiency suggestions.* 
2. It dramatically improves on the previous checklist output of the APL by focusing on a small number of actionable steps the player can take to improve at their rotation. Further, these steps are tied to individual mistakes---so the player can quickly jump to the exact spot that they made a mistake and see exactly what happened.

### Important Note

**This tool is not intended to be used by every spec.** Not every spec lines up neatly with an APL---for example, some have fairly rigid cast sequences that map poorly to this model. The idea is that the specs that *can* use it get a lot of value from it for little additional effort on their part.

## Future Work

This is "complete enough" that I want to go ahead and get it in. However, there are still some things I'd like to do in the future:

- Add range support to the APL checker (i.e. don't penalize you for casting a ranged spell instead of a melee spell when the boss is far away)
- Add deep links from examples into the full timeline / Warcraft Logs.
- Add a little bit more context to the timeline shown here (it is currently minimal because there is a LOT that goes into rendering the timeline)

## Reviewing This

I recommend starting with the `README.md` file that has been added. It is an intro to how you'd use this in your own summary page and tries to explain some core concepts.

After that, look at the code within the `src/interface/guide/shared/Apl` folder. This is the bulk of the new stuff and (hopefully) should be relatively well-structured.

Finally, there are some changes to the implementation of the APL checker. These are *almost all* plumbing to support spell lists, but there is some reorganization that I added while working on the timeline rewriter that didn't get stripped back out because it made the code more readable.


emallson:
image

added a small details subsection below the APL. was thinking that I want something that shows more absolute performance so people can see the bar go up as they get better.

src/analysis/retail/monk/brewmaster/modules/problems/InvokeNiuzao/index.tsx
```diff
@@ -248,59 +249,65 @@ export function InvokeNiuzaoSection({
...
+        most powerful damage cooldowns in the game&mdash;and one of the most dangerous. Using this
+        ability, we can efficiently turn massive amounts of damage taken into massive amounts of
+        damage dealt.
       </p>
       <p>
-        The level 58 upgrade to this ability causes <SpellLink id={talents.INVOKE_NIUZAO_THE_BLACK_OX_BREWMASTER_TALENT.id} /> to
-        add 25% of the purified damage to the next <SpellLink id={SPELLS.NIUZAO_STOMP_DAMAGE.id} />{' '}
-        within 6 seconds. Niuzao casts <SpellLink id={SPELLS.NIUZAO_STOMP_DAMAGE.id} /> when he is
-        summoned, and then every 5 seconds after that, for a total of 5
+        The level 58 upgrade to this ability causes{' '}
```
kfinch, line 267: Presumably the "level 58" text doesn't apply to the talent

src/parser/shared/metrics/apl/range.ts
```diff
@@ -0,0 +1,325 @@
...
+export const defaultRange = (info: PlayerInfo): number => {
+  const role = info.combatant.spec?.role;
+  const spec = info.combatant.specId;
+  switch (role) {
+    case ROLES.DPS.MELEE:
+    case ROLES.TANK:
+      return DEFAULT_MELEE_RANGE;
+    case ROLES.DPS.RANGED:
+      return DEFAULT_RANGED_RANGE;
+  }
```
kfinch: Is there a way to specify a different default for a whole spec, and make it talent dependent? The Druid class tree has a talent that increases the range of everything by 6 yards. Basically all Guardian and Feral druids want this, so typically they're playing with a melee range of 11 yds.

On a related note, I see what these calculations are doing, but I don't understand where/how the results are consumed to shape the output.

kfinch (approved):
LGTM, just a few questions

src/analysis/retail/monk/brewmaster/modules/problems/InvokeNiuzao/index.tsx
emallson: Yea, that section needs a pretty big overhaul actually because of the 10.0 changes. Haven't gotten to it yet, since it also isn't clear whether the talent its talking about is going to remain (it is the *only* "X, Rank 2" talent that didn't get a new name in the last pass---which makes me think it might be getting removed)

src/parser/shared/metrics/apl/range.ts
emallson: > Is there a way to specify a different default for a whole spec, and make it talent dependent? The Druid class tree has a talent that increases the range of everything by 6 yards. Basically all Guardian and Feral druids want this, so typically they're playing with a melee range of 11 yds.

I had thought about putting something on the Abilities module for the default range. Maybe I move this there, and they can conditionally set it based on talents?

> On a related note, I see what these calculations are doing, but I don't understand where/how the results are consumed to shape the output.

The main point that it is used is the [`isInRange`](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5094/files#diff-fd5c9ce915391510495d8ef8437a8555029c44bc880a4bbbafc9e24f44727e86R263) call here. The init/update state calls are [here](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5094/files#diff-fd5c9ce915391510495d8ef8437a8555029c44bc880a4bbbafc9e24f44727e86R457) and [here](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5094/files#diff-fd5c9ce915391510495d8ef8437a8555029c44bc880a4bbbafc9e24f44727e86R381)

src/interface/guide/shared/Apl/rules.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const AplRuleList = styled.ol`
+  padding-left: 1.5rem;
+`;
+
+const AplListItem = styled.li<{ highlighted?: boolean; muted?: boolean }>`
+  opacity: ${(props) => (props.muted ? 0.5 : 1)};
+
+  ${(props) =>
+    !props.highlighted
```
Pewtro: Personally I'd prefer `props.highlighted && `` ` instead of a ternary.

src/parser/shared/metrics/apl/range.ts
```diff
@@ -0,0 +1,325 @@
...
+  TargettedEvent,
+} from 'parser/core/Events';
+import { encodeTargetString } from 'parser/shared/modules/Enemies';
+import { PlayerInfo } from './index';
+
+const debug = false;
+
+export const Range = {
+  Melee: 5,
+  FullRanged: 40,
```
Pewtro: Add support for Evoker immediately with their 25y range (i think thats still it,no?)? 😄

Pewtro (approved):
Only two minor things non blocking, and then I'll probably have way more questions once I find time to mess with the implementation itself 😸

src/parser/shared/metrics/apl/range.ts
emallson: I think I'm going to try to move this into something in spec modules so that it can handle druid shenanigans (range increases, also rogues have that right?!), so Evokers can do their own thing there :P

PR #5099, opened 2022-10-01
Feral Druid file re-org and Guide outline. Added natural regen handling to ResourceTracker.

kfinch:


src/parser/core/Events.ts
```diff
@@ -455,7 +469,7 @@ export interface HealEvent extends Event<EventType.Heal> {
   /** If the event is a tick of a HoT or HoT like object */
   tick?: boolean;
   resourceActor: number;
```
emallson:
```suggestion
  resourceActor: ResourceActor;
```
should be able to set the type of these fields to `ResourceActor` since you set the numeric enum values, right?

src/parser/core/Events.ts
```diff
@@ -666,7 +680,10 @@ export interface ResourceChangeEvent extends Event<EventType.ResourceChange> {
   /** The amount of wasted resource gain (overcapped). */
   waste: number;
   otherResourceChange: number; // defaults to 0?
-  resourceActor: number; // 1 = source, 2 = target? used for classResources/hitpoints/etc, not resourceChange fields I think
+  /** Shows whether the source or the target is being referred to by
+   *  the classResources, hitPoints, etc. See {@link ResourceActor}. */
+  resourceActor: number;
```
emallson:
```suggestion
  resourceActor: ResourceActor;
```

src/parser/core/Events.ts
```diff
@@ -689,7 +706,10 @@ export interface DrainEvent extends Event<EventType.Drain> {
   resourceChange: number;
   resourceChangeType: number;
   otherResourceChange: number;
+  /** Shows whether the source or the target is being referred to by
+   *  the classResources, hitPoints, etc. See {@link ResourceActor}. */
   resourceActor: number;
```
emallson:
```suggestion
  resourceActor: ResourceActor;
```

src/parser/core/getResourceSpent.ts
```diff
@@ -0,0 +1,18 @@
+import { CastEvent } from 'parser/core/Events';
+import { Resource } from 'game/RESOURCE_TYPES';
+
+/** Returns the amount of resource spent by a cast event, returning zero if the
+ *  resource isn't included in classResources.
+ * @param event the event to get resource cost from
+ * @param resource the resource to check. See {@link RESOURCE_TYPES}
+ * @return the amount of the resource spent
+ */
+function getResourceSpent(event: CastEvent, resource: Resource): number {
```
emallson, line 10: good add :+1:

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -7,76 +7,189 @@ import Events, {
...
+  // END override values
+
+  /** Time ordered list of resource updates */
+  resourceUpdates: ResourceUpdate[] = [];
+  /** Tracked builders, indexed by spellId */
+  buildersObj: { [index: number]: BuilderObj } = {};
+  /** Tracked spenders, indexed by spellId */
+  spendersObj: { [index: number]: SpenderObj } = {};
+
   constructor(options: Options) {
```
emallson, line 159: it would be a breaking change to have the constructor set the params that are required, right?

like:

```tsx
type ResourceTrackerParams = Pick<ResourceTracker, 'resource' | 'maxResource'> & Partial<Pick<ResourceTracker, 'baseRegenRate' | 'isRegenHasted'>>;

constructor(options: Options, params: ResourceTrackerParams) {
  this.resource = params.resource;
  this.maxResource = params.maxResource;
  this.baseRegenRate = params.baseRegenRate ?? this.baseRegenRate;
  this.isRegenHasted = params.isRegenHasted ?? this.isRegenHasted;

  // ...the rest of the constructor
}
```

emallson (approved):
got some comments but nothing blocking

src/parser/core/Events.ts
kfinch: .. oh probably, I think I hesitated because I wasn't sure what would happen if the raw data gave us a number that wasn't expected by the enum

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
kfinch: It's certainly breaking for all the existing extenders of this class, but wouldn't be too challenging a refactor. I was unsure if I'd have access to `selectedCombatant` when building this way, because for example Feral EnergyTracker needs to check a talent rank to figure out what the baseRegenRate is

kfinch: In fact I think I explicitly had this problem when designing BuffCountGraph - buffSpecs used to be passed in the constructor but it didn't work because I needed dynamic combatantInfo stuff

kfinch:
Merging this - can mess with the constructor later if we can find a way that works right

emallson:
:+1: sounds good to me

PR #5101, opened 2022-10-02
Feature/disc talents second pass

oisinod:


src/common/SPELLS/priest.ts
```diff
@@ -73,20 +73,20 @@ const spells = spellIndexableList({
...
     icon: 'spell_holy_penance',
-    manaCost: 800,
+    manaCost: 4000,
   },
   PENANCE_CAST: {
     id: 47540,
     name: 'Penance',
     icon: 'spell_holy_penance',
-    manaCost: 800,
+    manaCost: 4000,
```
emallson, line 89: these mana costs should be in the talent data, right? or is penance not a talent for disc?

emallson (commented):
one comment but otherwise looks good

src/common/SPELLS/priest.ts
oisinod: It's baseline :)

emallson: unfortunate

PR #5102, opened 2022-10-03
update Specializations page for Dragonflight

### LINKED ISSUE #5091: Update Specializations Page
Currently, the [specs](https://dragonflight.wowanalyzer.com/specs) page has a jumble of all specs. This would benefit greatly from better organization.

- [x] Divide into two sections: Classic & Retail
- [x] Organize Retail specs into rows
- [ ] Move unsupported *classes* to the bottom of the list. (Unsupported specs of a supported class should stay in their row)

ToppleTheNun:
separate specs into Retail and Classic sections

group specs by class, order by support and name

image
image

closes #5091


kfinch:
I like it, but also I will add Guardian Druid in the near future and I'd like to know what that's going to look like

ToppleTheNun:
It would end up looking like this but with the Guardian paw instead of the leaf:
image

emallson:
We can do something to either fit 4 on the row for Druid, or do 2 rows of 2 so it isn't imbalanced. (but that can be done later)

emallson:
Going to merge, but reopen the issue with a note about Druid stuff (+ another thing I want to explore but may not ultimately do w/ showing specs that are totally unsupported)

PR #5113, opened 2022-10-05
[Talents] Only use Shared/SpecName when necessary

Pewtro:


src/common/TALENTS/demonhunter.ts
```diff
@@ -287,544 +287,512 @@ const talents = createTalentList({
     icon: 'ability_demonhunter_concentratedsigils',
     maxRanks: 1,
   },
+  FODDER_TO_THE_FLAME_TALENT: {
+    id: 391429,
+    name: 'Fodder to the Flame',
+    icon: 'ability_maldraxxus_demonhunter',
+    maxRanks: 1,
+  },
+  ELYSIAN_DECREE_TALENT: {
```
ToppleTheNun, line 296: Both DH specs have access to both Fodder to the Flame and Elysian Decree in their spec trees and at least Elysian Decree has different spell IDs.

Havoc Elysian Decree: 390163
Vengeance Elysian Decree: 389815

kfinch: On my local I'm seeing these two talents as having the same ID, which is probably why the script merged them. Do we know which of these two options is more up to date?

ToppleTheNun: The odd thing is that it looks like the actual talent on beta is using this ID (`390163`) but the logs on WCL have two separate IDs for the casts:

https://www.warcraftlogs.com/reports/vQFkL6GbpzyHCwmc#boss=-2&difficulty=0&wipes=1&source=10&type=casts&ability=389815
https://www.warcraftlogs.com/reports/pB6yDRqZNThjCxVc#boss=-2&difficulty=0&wipes=1&source=28&type=casts&ability=390163

ToppleTheNun: I'm mostly bringing this up because we've noted previously that the talent usually has the spell ID of the cast. I can take note of this and cover it in the DH `SPELLS` object.

kfinch: Ah yeah I checked those logs and it is very odd that they have different cast IDs. I did notice that in those logs on the summary page, we can see the both players had the same Talent ID - meaning our generation is correct.

I think your proposal to handle this in the DH `SPELLS` object is the correct solution.

Pewtro: I think this is a case of talent and cast/damage spells being different :) 
Looking at wowhead [havoc talent calc](https://www.wowhead.com/beta/talent-calc/demon-hunter/havoc/DAAANURVmEQiAUAEAQBA), the havoc talent has the spellid 391429. 
The same happens on [vengeance talent calc](https://www.wowhead.com/beta/talent-calc/demon-hunter/vengeance/DAAAOVVFUIgQQoAQIAQBA), where the talent also has the same spellid 391429. 

Either it's bugged in-game where the havoc talent points to the vengeance spell by accident (wowhead has parsed in `Requires Demon Hunter (Vengeance)`), or it's the same spellid for the talent but it then activates different cast/damage spells based on specs - both wouldn't be surprising.

PR #5115, opened 2022-10-05
convert ShearFracture to use EventLinkNormalizer

ToppleTheNun:
It does change how bad casts are detected (likely in a good way) so the numbers may end up different than they are on live.

**Changed version**
image

**Live version**
image
https://dragonflight.wowanalyzer.com/report/vQFkL6GbpzyHCwmc/12-Heroic+Kurog+Grimtotem+-+Wipe+2+(8:00)/Devonhunter/standard/overview

### DIFF
--- a/src/analysis/retail/demonhunter/vengeance/normalizers/ShearFractureNormalizer.ts
+++ b/src/analysis/retail/demonhunter/vengeance/normalizers/ShearFractureNormalizer.ts
@@ -1,0 +1,65 @@
+import SPELLS from 'common/SPELLS/demonhunter';
+import TALENTS from 'common/TALENTS/demonhunter';
+import { CastEvent, EventType, GetRelatedEvents, RemoveBuffStackEvent } from 'parser/core/Events';
+import { Options } from 'parser/core/Module';
+import EventLinkNormalizer, { EventLink } from 'parser/core/EventLinkNormalizer';
+
+/*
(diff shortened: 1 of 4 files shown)


src/analysis/retail/demonhunter/vengeance/normalizers/ShearFractureNormalizer.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const EVENT_LINKS: EventLink[] = [
+  {
+    linkRelation: GENERATED_SOUL_FRAGMENT,
+    referencedEventId: [SPELLS.SHEAR.id, TALENTS.FRACTURE_VENGEANCE_TALENT.id],
+    referencedEventType: EventType.Cast,
+    linkingEventId: SPELLS.SOUL_FRAGMENT.id,
+    linkingEventType: EventType.Cast,
+    forwardBufferMs: SOUL_GENERATE_BUFFER,
+    backwardBufferMs: SOUL_GENERATE_BUFFER,
```
kfinch: If my reading of the behavior is correct, the long delay always comes *after* the Shear / Fracture cast - does this work with forward buffer of zero, but backwards buffer the same? Am I misunderstanding what's going on?

ToppleTheNun: Setting it to a `forwardBufferMs` of `0` shows as having 64 bad casts, which is more accurate. I'm still trying to run down another false positive in one of the logs I'm working with, but this in itself should be enough.

PR #5117, opened 2022-10-08
Baseline working MW with no compilation errors

trevorm4:
Untangled various files with dead spell id's etc. There are a bunch of shadowlands specific stuff that will take a while to untangle  due to nature of the MW directory setup, so going to limit the scope of this PR.


emallson (approved):
looks good, thanks!

PR #5118, opened 2022-10-08
Feature/aegis of wrath first pass

oisinod:
Will need an update when the 4p module is created


src/analysis/retail/priest/discipline/modules/spells/AegisOfWrath.tsx
```diff
@@ -0,0 +1,82 @@
...
+        event.targetID === shield.targetID &&
+        event.timestamp > shield.timestamp &&
+        shield.timestamp + POWER_WORD_SHIELD_DURATION_MS > event.timestamp
+      ) {
+        const shieldAmount = shield.absorb || 0;
+        const extraShield = shieldAmount - shieldAmount / 1.5;
+        this.bonusShielding += extraShield - (event.absorb || 0);
+      }
+    });
+  }
```
emallson, line 57: just an OCD note that this is fairly inefficient since it will reprocess every application on every expiration, but it is probably MORE inefficient to use `splice` to remove the old entry every time. and that this inefficiency is very minor in the scheme of things.

emallson (approved):

PR #5120, opened 2022-10-09
Reorganized MW files and added support for existing talents

trevorm4:
Reorganized files and added support for already implemented talents such as clouded focus, uplifted spirits (formerly rsr), FLS, etc

Deleted shadowlands legendary spell info to avoid duplication as well

RM is still not working as it requires some clever solutions to the problem that RM does not extend any REM's procced by talents (which seems like a very hard problem to solve on the surface)

The actual mechanics of talents was implemented with help from Sweggles as he has done extensive testing on beta


trevorm4:
Oops forgot changelong entry, will add

emallson (approved):

PR #5125, opened 2022-10-09
[Arcane Mage] Decrashify for Dragonflight

Sharrq:
Just fixed the talent spells so it doesnt crash and activates the spec. Going to do actual DF changes after this.


emallson (approved):

Sharrq:
I need to refix the talent links, looks like the names changed since i did this

PR #5130, opened 2022-10-11
Power word shield module

oisinod:
this handles both the 4p and aegis of wrath contributions to power word shield.

unfortunately due to how connected they are it seemed based on conversations with a few people to have these handled in the same module. this will be the case for other effects such as Divine Aegis and Weal and Woe which will be handled in future PRs


src/analysis/retail/priest/discipline/modules/spells/PowerWordShield.tsx
```diff
@@ -0,0 +1,181 @@
...
+    return crit;
+  }
+
+  onShieldExpiry(event: RemoveBuffEvent) {
+    this.shieldApplications.forEach((shieldApplication) => {
+      if (
+        event.targetID === shieldApplication.event.targetID &&
+        event.timestamp > shieldApplication.event.timestamp &&
+        shieldApplication.event.timestamp + POWER_WORD_SHIELD_DURATION_MS > event.timestamp
+      ) {
```
emallson: for both this and `onPWSAbsorb` I would recommend converting `shieldApplications` to an object or map keyed by `targetID`. For example:

```ts
private shieldApplications: Map<number, ShieldInfo> = new Map();
```

then instead of using this foreach you do:

```ts
const info = this.shieldApplications.get(event.targetID);
if (!info || 
    info.event.timestamp > event.timestamp || // not sure how this happens? fabrication stuff?
    info.event.timestamp + POWER_WORLD_SHIELD_DURATION_MS < event.timestamp
    ) {
    return;
    }
 ```
 
 this will be significantly faster on higher group sizes / longer fights.

**Review by emallson — CHANGES_REQUESTED** [review]

File: src/analysis/retail/priest/discipline/modules/spells/PowerWordShield.tsx [hunk 2]
```diff
@@ -104,48 +118,60 @@ class PowerWordShield extends Analyzer {
...
-        totalShielded -=
-          totalShielded >= shieldOfAbsolutionBonus ? shieldOfAbsolutionBonus : totalShielded;
-        this.aegisValue += totalShielded;
-      }
-    });
+    this.handleRemovedShield(event);
+  }
+
+  onShieldRefresh(event: RefreshBuffEvent) {
+    this.handleRemovedShield(event);
```
emallson, line 125: is there always an apply after the refresh? this will handle the removal but not apply a new pw:s. example: https://www.warcraftlogs.com/reports/ycBxvL38YaWtGfkC#fight=30&type=auras&ability=17&view=events&source=411

oisinod: no, you're right, so I should add this also to the refreshbuff method?
```javascript
    if (this.shieldApplications.get(event.targetID)) {
      this.shieldApplications.set(event.targetID, null);
    }
    this.shieldApplications.set(event.targetID, {
      event: event,
      shieldOfAbsolutionValue: this.shieldOfAbsolutionValue,
      healing: 0,
      crit: this.critCheck(event),
    });
    this.shieldOfAbsolutionValue = 0;

Comment (emallson) [reply src/analysis/retail/priest/discipline/modules/spells/PowerWordShield.tsx:125]: this could maybe be:

```ts
this.handleRemovedShield(event);
this.onShieldApplication(event);
```
which would require adjusting the type of the 2nd call but should work i think?

oisinod:
Change adressed @emallson

Let me know if that's what you meant :)

emallson (approved):

PR #5144, opened 2022-10-13
[Stats] Update stat scaling for DF

Pewtro:
Currently uses the prepatch numbers for rating per % with a commented out value for usage @ 70 after prepatch ends


Pewtro:
The combatantinfo event doesn't carry player level so I couldn't make it scale by players easily, and setting defaults doesn't seem to give any value as then we just need to find all usages and update, versus just replacing in the `statBaselineRatingPerPercent` object.

kfinch:
I guess this will be fine - I'd prefer not to merge this until the Pre-patch so that the scaling will still work right for Beta logs

emallson:
We *could* do this based on talent points spent since the game literally will not let you apply a build that doesn't spend all your points, but tbh i think this is fine for prepatch. I can create a task to go in and swap these numbers out when we go live.

Pewtro:
> I guess this will be fine - I'd prefer not to merge this until the Pre-patch so that the scaling will still work right for Beta logs

Then you should probably do a separate PR to fix them to proper values since crit/haste is swapped around currently. not that it will matter _much_ but 🤷🏼

kfinch:
Here we goooooooooooooooooooooo

PR #5147, opened 2022-10-14
Clean up duplicate spells/talents for mistweaver

trevorm4:
As suggested in previous PR, bunch of duplicate spells/talents needed to be deleted. Got rid of conduit analysis files that aren't needed anymore and deleted some conduit files as analysis on them is not needed.


src/analysis/retail/monk/mistweaver/CHANGELOG.tsx
```diff
@@ -6,6 +6,7 @@ import { SpellLink } from 'interface';
 
 
 export default [
+  change(date(2022, 10, 13), <>Cleaned up MW spells/talents files</>, Trevor),
   change(date(2022, 10, 9), <>Added Secret Infusion haste buff and fixed Uplifted Spirits</>, Trevor),
   change(date(2022, 10, 8), <>Added support for existing MW talents</>, Trevor),
```
kfinch: I think it's acceptable to clean up the pre-DF stuff in changelog when updating a spec.

src/analysis/retail/monk/mistweaver/CONFIG.tsx
```diff
@@ -9,7 +9,7 @@ const config: Config = {
   contributors: [],
   expansion: Expansion.Dragonflight,
   // The WoW client patch this spec was last updated.
-  patchCompatibility: null,
+  patchCompatibility: '10.0.2',
   isPartial: false,
```
kfinch: If DF isn't fully supported yet, either set `isPartial` to true, or leave patchCompatability as 9.2.7 (analysis will run but will display 'not updated for latest patch' message)

trevorm4: Set isPartial to true

PR #5148, opened 2022-10-14
Update RM implementation to account for talent procs

trevorm4:
Rising Mist module currently blindly extends all HoTs but this is not reality. ReM and EnvM from talents (Dancing Mists, Misty Peaks, and Rapid Diffusion) do not get extended by ReM. This diff adds a module to link Buff applications/removal to their original application. Iff a ReM can be linked back to a hardcast then it will be extended by ReM. This framework also adds support for adding attributions for various talents, tho atm only hardcast is added to the hot object.


src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+    referencedEventType: [EventType.RemoveBuff],
+    forwardBufferMs: 0,
+    backwardBufferMs: 500,
+    anyTarget: true,
+    additionalCondition(linkingEvent, referencedEvent) {
+      return (
+        (linkingEvent as ApplyBuffEvent).targetID !== (referencedEvent as RemoveBuffEvent).targetID
+      );
+    },
+  },
```
kfinch: Are bounces able to cause a RefreshBuff?

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+  // link renewing mist removal to its application event
+  {
+    linkRelation: BOUNCED,
+    linkingEventId: [SPELLS.RENEWING_MIST_HEAL.id],
+    linkingEventType: [EventType.RemoveBuff],
+    referencedEventId: SPELLS.RENEWING_MIST_HEAL.id,
+    referencedEventType: [EventType.ApplyBuff],
+    forwardBufferMs: 0,
+    backwardBufferMs: 65000,
+  },
```
kfinch: Well this is novel, but probably works. How was the 65 second buffer chosen?

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+    referencedEventId: [
+      TALENTS_MONK.RISING_SUN_KICK_TALENT.id,
+      TALENTS_MONK.ENVELOPING_MIST_TALENT.id,
+      SPELLS.ENVELOPING_MIST_TFT.id,
+    ],
+    referencedEventType: [EventType.Cast],
+    forwardBufferMs: 500,
+    backwardBufferMs: 500,
+    anyTarget: true,
+  },
```
kfinch: Big possibility of false positive here - I'm pretty sure if player casts Enveloping Mists -> Renewing Mist, both spells will go off at the same time, and then this link will catch the hardcast.

While it's not documented, the eventlinks are applied in order so I think you could add an additional condition to this link to check that it doesn't already have a FROM_HARDCAST link. You'd be breaking new ground for this normalizer, but it was a use case I had in mind.

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+    linkRelation: FROM_MISTY_PEAKS,
+    linkingEventId: [TALENTS_MONK.ENVELOPING_MIST_TALENT.id],
+    linkingEventType: [EventType.ApplyBuff],
+    referencedEventId: SPELLS.RENEWING_MIST_HEAL.id,
+    referencedEventType: [EventType.Heal],
+    anyTarget: true,
+    forwardBufferMs: 50,
+    backwardBufferMs: 50,
+  },
+];
```
kfinch: Again I'd think about the possibility of false positives here although I haven't thought this all the way through

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+    // see if ancestor event can be linked to a hardcast
+    if (minEvent.type === EventType.RemoveBuff) {
+      return isFromHardcast(minEvent as RemoveBuffEvent);
+    } else {
+      return isFromHardcast(minEvent as ApplyBuffEvent);
+    }
+  }
+  return false;
+}
```
kfinch: Big yikes, but I think I understand what you're doing here. Looks like you're resolving the 'double link' problem with logic here, then tracing back bounces if you need to? This seems viable, but I'd like you to add a bit more docs just so future maintainers can figure out the reasoning.

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,173 @@
...
+    linkingEventId: [SPELLS.RENEWING_MIST_HEAL.id],
+    linkingEventType: [EventType.ApplyBuff],
+    referencedEventId: [
+      TALENTS_MONK.RISING_SUN_KICK_TALENT.id,
+      TALENTS_MONK.ENVELOPING_MIST_TALENT.id,
+      SPELLS.ENVELOPING_MIST_TFT.id,
+    ],
+    referencedEventType: [EventType.Cast],
+    forwardBufferMs: 500,
+    backwardBufferMs: 500,
```
kfinch: Does the buffer really need to be this big, and in both directions?

kfinch (changes requested):
Looks good, but I'd like you to take a second look on the things I've commented on

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
trevorm4: Talked to Sweggles and he said 77k ms is a good safe upper bound as you can get a bunch of extensions from casting EF and FLS EF on cd, so gonna change to that. With just baseline tft rem and max RM extension the max is 60 seconds, but each EF bolt hit adds another second.

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
trevorm4: Yep technically if every single person in the group has a rem then yes, should add RefreshBuffEvent alongside Apply's.

kfinch:
I discussed strategies for this in DM convo with Trevor - brainstorming solutions

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
trevorm4: Addressing this one by adding an additionalCondition that it cannot be linked to a hardcast as suggested in DMs, which should address false positive issue

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
trevorm4: Added comments and split out spooky logic to a helper function to explain this one as it is a very necessary step.

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
trevorm4: Reducing buffer

src/analysis/retail/monk/mistweaver/modules/spells/RisingMist.js
```diff
@@ -121,23 +125,36 @@ class RisingMist extends Analyzer {
...
   }
 
+  hasAttribution(attributions, name) {
+    let foundAttribution = false;
+    attributions.forEach(function (attr) {
+      if (attr.name === name) {
+        foundAttribution = true;
+      }
+    });
+    return foundAttribution;
```
kfinch, line 142: Can do same thing with `return Boolean(attributions.find(att => att.name === name));`

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -0,0 +1,185 @@
...
+    if (minEvent.type === EventType.ApplyBuff) {
+      return isFromHardcast(minEvent as ApplyBuffEvent);
+    } else if (minEvent.type === EventType.RemoveBuff) {
+      return isFromHardcast(minEvent as RemoveBuffEvent);
+    } else {
+      return isFromHardcast(minEvent as RefreshBuffEvent);
+    }
+  }
+  return false;
+}
```
kfinch, line 183: This is pretty scary, but if it's working I guess??

kfinch:
After discussion, will merge this now and look for avenues for improvement in follow-on PR

kfinch (approved):

PR #5154, opened 2022-10-14
Holy priest new talent and updated cooldown view

emilnormann:
Apotheosis and salvation are now tracked in the cooldown throughput window.
Added support for 'Healing Chorus' talent.


kfinch (approved):
LGTM

PR #5162, opened 2022-10-15
Holy priest tier implemented

emilnormann:
2 and 4 set has been implemented. Right now they are by default not shown on page because their activation relies on 'has2Piece' and 'has4Piece' which are not updated for Dragonflight.
Talent 'translucent image' has also been updated for Dragonflight


src/analysis/retail/priest/holy/CombatLogParser.tsx
```diff
@@ -107,7 +111,6 @@ class CombatLogParser extends CoreCombatLogParser {
     Lightweaver: Talents.BottomRow.Lightweaver,
 
     divineImage: Talents.BottomRow.DivineImage,
-    twinsOfTheSunPriestess: TwinsOfTheSunPriestess,
```
emallson, line 110: did you mean to remove this?

emilnormann: I removed it because I think it provides very little value in the statistics tab since it is basically the amount of times you cast PI.
So if anything I think it should be in the overview tab. But the functionality is still there so it is easy to insert again.

oisinod: I'd just say that the Twins module has a suggestion if you didn't use it on others while its active which is IMO somewhat useful to show

emallson: i'm okay leaving it out for now. gonna go ahead and merge since there is other stuff we want here.

PR #5166, opened 2022-10-15
[DF] Sub Rogue initial fixes

sMteX:
As usual, mostly renaming stuff from Spells to Talents, but also occasional fix (some conduits have become talents, some modules are no longer shared etc). Will likely need additional work to get really up to par, but it should load without errors.


emallson (commented):
looks like you have a type error on the shadowstep talent, but other than that this looks good

kfinch:
LGTM, though some Talents that were Conduits are fixed in code but still remain in 'shadowlands/conduits' folder - I recommend doing the big folder rearrange next. Fine for that to be in following PR.

kfinch (approved):

PR #5168, opened 2022-10-15
add support for Stoke the Flames

ToppleTheNun:


emallson (approved):

src/analysis/retail/demonhunter/vengeance/modules/talents/StokeTheFlames.tsx
```diff
@@ -0,0 +1,41 @@
...
+      return;
+    }
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(SPELLS.FEL_DEVASTATION_DAMAGE),
+      this.onDamage,
+    );
+  }
+
+  onDamage(event: DamageEvent) {
+    this.addedDamage += calculateEffectiveDamage(event, STOKE_THE_FLAMES_INCREASE);
```
danny-janse, line 29: Wouldn't a tank also be interested in how much extra healing they gained from this?

danny-janse (approved):
Just a slight addition you could potentiolly make to make the module more complete.

src/analysis/retail/demonhunter/vengeance/modules/talents/StokeTheFlames.tsx
ToppleTheNun: The answer is "yes but also no" because the majority of the healing that Fel Devastation does is from leech instead of the ability itself; I'll add it as well because it is useful.

ToppleTheNun: Update, it actually doesn't affect the healing outside of leech.

PR #5170, opened 2022-10-15
Druid Feral - Overhauled Guide's 'Resource Use' section

kfinch:
image


src/analysis/retail/druid/feral/Guide.tsx
```diff
@@ -158,3 +177,21 @@ function CooldownGraphSubsection({ modules, events, info }: GuideProps<typeof Co
...
+  grid-template-columns: 1fr max-content;
+  align-content: center;
+  align-items: center;
+`;
+
+const SideBySidePanels = styled.div`
+  display: grid;
+  grid-template-columns: minmax(50%, max-content) 1fr;
+  grid-column-gap: 1em;
+`;
```
kfinch, line 197: Better place for these than copy-pasta'd silly name?

src/interface/guide/index.tsx
```diff
@@ -308,6 +308,11 @@ export const GoodMark = () => <i className="glyphicon glyphicon-ok good-mark" />
 export const OkMark = () => <i className="glyphicon glyphicon-asterisk ok-mark" />;
 export const BadMark = () => <i className="glyphicon glyphicon-remove bad-mark" />;
 
+export const PerfectColor = '#2090c0';
+export const GoodColor = '#4ec04e';
+export const OkColor = '#ffc84a';
+export const BadColor = '#ac1f39';
```
kfinch: These are copied from the same colors in CSS - is it possible for them to be accessible from TS and CSS without copying them?

ToppleTheNun: This seems like a possibly good case for CSS variables.

```tsx
export const PerfectColor = getComputedStyle(document.documentElement).getPropertyValue('--guide-perfect-color');
```

```scss
$perfect-perf-color: #2090c0;

:root {
  --guide-perfect-color: #{perfect-perf-color};
]
```

kfinch: Seems to work, thanks!

PR #5182, opened 2022-10-16
add support for Fiery Brand mitigation

ToppleTheNun:
image
image

### DIFF
--- a/src/parser/shared/modules/Enemies.ts
+++ b/src/parser/shared/modules/Enemies.ts
@@ -42,6 +42,7 @@ class Enemies extends Entities<Enemy> {
   }
 
   /**
+   * Retrieves the enemy that was the target of given event.
    * @param {object} event
    * @returns {Enemy|null}
    */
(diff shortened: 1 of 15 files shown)


src/parser/shared/modules/Enemies.ts
```diff
@@ -72,6 +73,42 @@ class Enemies extends Entities<Enemy> {
...
 
+  /**
+   * Retrieves the enemy that was the source of given event.
+   * @param {object} event
+   * @returns {Enemy|null}
+   */
+  getSourceEntity(event: AnyEvent): Enemy | null {
+    if (!HasTarget(event)) {
+      return null;
+    }
```
kfinch: `!HasSource(event)) {` instead?

ToppleTheNun: Good catch!

src/parser/shared/modules/Enemies.ts
```diff
@@ -72,6 +73,42 @@ class Enemies extends Entities<Enemy> {
...
+  getSourceEntity(event: AnyEvent): Enemy | null {
+    if (!HasSource(event)) {
+      return null;
+    }
+    if (event.sourceIsFriendly) {
+      return null;
+    }
+    if (!event.sourceID) {
+      return null;
+    }
```
kfinch: `!HasSource(event)` and `!event.sourceID` are redundant

ToppleTheNun: Also makes the `"sourceInstance" in event` check unnecessary

kfinch (approved):

PR #5187, opened 2022-10-16
Misty peaks analysis for mistweaver

trevorm4:
Adding misty peaks analysis enabled by the rising mist PR. Also fixed some overcounting that was in the Rising Mist diff.


trevorm4:
Closing. The hot tracker has short comings with the new renewing mist and abe & I agreed that we probably need to write an entirely separate module for tracking mistweaver hots as there are too many random buff apply/remove's with new MW talents to expect the existing module to work properly.

PR #5193, opened 2022-10-18
add Convoke Spirits to Balance Druid Abilities

ToppleTheNun:
I noticed this while trying to validate another change that I was making.


kfinch (approved):
I think this happened because I originally had it in shared, but then removed it from there and forgot to re-add it everywhere..

PR #5194, opened 2022-10-18
use generated talent info for PlayerInfoTalents

ToppleTheNun:
this prevents us from running into a situation
where spells are unknown when rendering on the
character page.

Note that there are still some weird cases where a
talent might not appear due to naming conflicts.
Example of this is War Machine for Warrior. I'm
still looking into a possible fix for this, but this
should be a good starting place.

Log: https://dragonflight.wowanalyzer.com/report/zg96K4TcxkrBWP32/3-Normal+Terros+-+Kill+(5:37)/Daemondie/standard/character

**Before**
image

**After**
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -67,6 +67,7 @@ import { TALENTS_SHAMAN } from 'common/TALENTS';
 
 // prettier-ignore
 export default [
+  change(date(2022, 10, 17), 'Use generated talent info for character pages.', ToppleTheNun),
   change(date(2022, 10, 16), 'Update SpellLink to accept Spell instances.', ToppleTheNun),
   change(date(2022, 10, 16), 'Add ability to determine if wearing T29 2pc or 4pc.', ToppleTheNun),
   change(date(2022, 10, 16), 'Adding sourceInstance to BuffEvent', Jonfanz),
--- a/src/common/TALENTS/maybeGetTalent.ts
+++ b/src/common/TALENTS/maybeGetTalent.ts
@@ -1,0 +1,42 @@
+// We currently only need to fetch talents by IDs from the large object.
+import { indexOnlyById } from 'common/indexById';
+import { Talent } from 'common/TALENTS/types';
+import {
+  TALENTS_WARRIOR,
+  TALENTS_PALADIN,
+  TALENTS_HUNTER,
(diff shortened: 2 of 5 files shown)


Pewtro:
If we do it like this I think we should remove the autogenerated `index.ts` file as it's becoming more complex for no real gain

ToppleTheNun:
I think moving this snippet to a separate file would work as well, without removing the auto-generated index.ts:
```ts
// We currently only need to fetch talents by IDs from the large object.
const TALENTS = {
  ...indexOnlyById<Talent, typeof TALENTS_WARRIOR>(TALENTS_WARRIOR),
  ...indexOnlyById<Talent, typeof TALENTS_PALADIN>(TALENTS_PALADIN),
  ...indexOnlyById<Talent, typeof TALENTS_HUNTER>(TALENTS_HUNTER),
  ...indexOnlyById<Talent, typeof TALENTS_ROGUE>(TALENTS_ROGUE),
  ...indexOnlyById<Talent, typeof TALENTS_PRIEST>(TALENTS_PRIEST),
  ...indexOnlyById<Talent, typeof TALENTS_DEATH_KNIGHT>(TALENTS_DEATH_KNIGHT),
  ...indexOnlyById<Talent, typeof TALENTS_SHAMAN>(TALENTS_SHAMAN),
  ...indexOnlyById<Talent, typeof TALENTS_MAGE>(TALENTS_MAGE),
  ...indexOnlyById<Talent, typeof TALENTS_WARLOCK>(TALENTS_WARLOCK),
  ...indexOnlyById<Talent, typeof TALENTS_MONK>(TALENTS_MONK),
  ...indexOnlyById<Talent, typeof TALENTS_DRUID>(TALENTS_DRUID),
  ...indexOnlyById<Talent, typeof TALENTS_DEMON_HUNTER>(TALENTS_DEMON_HUNTER),
  ...indexOnlyById<Talent, typeof TALENTS_EVOKER>(TALENTS_EVOKER),
} as const;

export const maybeGetTalent = (key: number | undefined): Talent | undefined =>
  key ? TALENTS[key as any] : undefined;
```

ToppleTheNun:
In thinking about it, I decided I liked having it be a separate file.

emallson (approved):
looks like putput's comments are addressed, and the changes look fine to me :+1:

PR #5195, opened 2022-10-18
Guide core rearrange

kfinch:
Moved several of the UI components that were in Guide's index into their own files. Changes to Druid and Monk files are entirely fixing the imports - no user facing changes have been made.

@emallson does this folder structure make sense to you? Is it cool to have emotion components in a file like that, or should they just be in line with whatever is using it? Is there a better name for that file than `GuideDivs`


Pewtro:
Would it make sense to put them inside a guide/components folder, potentially even a folder per component to let .css files and everything related to a component be isolated?

emallson:
> Would it make sense to put them inside a guide/components folder, potentially even a folder per component to let .css files and everything related to a component be isolated?

I like `guide/components` a lot. IMO for the folder thing---either do:

- `Foo.tsx` if there is only a single file for it, or
- `Foo/index.tsx` and `Foo/whatever.css` etc for multiple files.

then you always import `interface/guide/components/Foo`

kfinch:
Ok I did the components and `Foo/index.tsx` thing as suggested, Will merge when this passes checks.

PR #5197, opened 2022-10-18
Feature/demo-in-dragonflight

jharrell:
Going from Affi to Demo was a real wakeup call. I tried to do the bare minimum to get this spec working in DF and we should treat this spec as pretty much perma-broken until someone with a lot more Demo knowledge than I can fix.


ToppleTheNun:
Can you add a screenshot of it loading a demo log?

jharrell:
image

@ToppleTheNun attached! I had to tweak a few things. It looks like the character tab isn't working but I'm guessing that's because the report is old? It's failing on `PlayerInfoTalents` component.

ToppleTheNun:
Thanks for the screenshot! I'll double check on the character page thing shortly.

ToppleTheNun (approved):
After checking it out locally and rebasing for the fix for character pages, this is good.

PR #5155, opened 2022-10-15
Resolve Errors for Dragonflight Elemental Shaman

omegabiscuit:
Enabled Elemental Shaman for Dragonflight and fixed any errors that appeared. Overall summary of changes are:
- Imported Talents and changed abilities that will be talents in DF to use TALENTS list
- Removed dead talents Earthen Rage, Echoing Shock, Static Discharge, and Unlimited Power. 
- With the removal of Unlimited Power, I removed UnlimitedPowerTimesByStacks as well

PR #5198, opened 2022-10-18
Cleanup remaining shaman spells

danny-janse:
Removed a bunch of remaining spells in the spells/shaman file that are now talents. This has become heavily Elemental Shaman focused.

A few aditional adjustments since i had noticed them while making the above changes.
* Enable Elemental Shaman on the specs page.
* Add an enabled variable to the Elemental Shaman abilities which are now a talent.
* Add active checks to the Elemental Shaman modules.
* Add Paladin as a class that has a CR.
* Add Evoker as a class that has a Bloodlust like ability.


src/analysis/retail/shaman/elemental/modules/Buffs.tsx
```diff
@@ -1,5 +1,5 @@
 import SPELLS from 'common/SPELLS';
-import TALENTS from 'common/TALENTS/shaman';
+import TALENTS, { TALENTS_SHAMAN } from 'common/TALENTS/shaman';
```
emallson:
```suggestion
import TALENTS from 'common/TALENTS/shaman';
```
these imports are the same object. can merge (or keep `TALENTS_SHAMAN`, up to you---just better not to have 2 names for the same thing)

src/analysis/retail/shaman/elemental/modules/Buffs.tsx
```diff
@@ -15,7 +15,7 @@ class Buffs extends CoreAuras {
         timelineHighlight: true,
       },
       {
-        spellId: SPELLS.FIRE_ELEMENTAL.id,
+        spellId: TALENTS_SHAMAN.FIRE_ELEMENTAL_TALENT.id,
```
emallson:
```suggestion
        spellId: TALENTS.FIRE_ELEMENTAL_TALENT.id,
```

src/analysis/retail/shaman/elemental/modules/checklist/Component.tsx
```diff
@@ -1,5 +1,5 @@
 import SPELLS from 'common/SPELLS';
-import TALENTS from 'common/TALENTS/shaman';
+import TALENTS, { TALENTS_SHAMAN } from 'common/TALENTS/shaman';
```
emallson: same comment here---should use a single name

src/analysis/retail/shaman/elemental/modules/features/AlwaysBeCasting.tsx
```diff
@@ -1,6 +1,6 @@
 import { formatPercentage } from 'common/format';
 import SPELLS from 'common/SPELLS';
-import TALENTS from 'common/TALENTS/shaman';
+import TALENTS, { TALENTS_SHAMAN } from 'common/TALENTS/shaman';
```
emallson: and here

emallson (changes requested):
didn't comment on every instance, but there are a number of files importing `TALENTS_SHAMAN` as `TALENTS_SHAMAN` *and* `TALENTS`. Should edit to only use a single name to avoid confusion (either is fine)

danny-janse:
> didn't comment on every instance, but there are a number of files importing `TALENTS_SHAMAN` as `TALENTS_SHAMAN` _and_ `TALENTS`. Should edit to only use a single name to avoid confusion (either is fine)

I've changed every instance to TALENTS since this is what was primairly used by omegabiscuit in #5155

PR #5200, opened 2022-10-19
fix rsk uptime calculations for mw

trevorm4:
SpellUsable rewrite broke RSK module because we manually began/ended cooldown for rsk but the module does that for us already which led to bad things


kfinch (approved):
We discussed this - fix makes sense

PR #5201, opened 2022-10-19
Fix ATOTM rsk event handler, fix RSK uptime calculation and rename RSK damage event 

trevorm4:
- Renamed the RSK damage event from SPELLS.RISING_KICK_SECOND to SPELLS.RISING_KICK_DAMAGE (old name meant zero sense to any human looking at it)
- Removed rsk event fabrication RSK module as its handled by SpellUsable now (also fixed in SL branch #5200)
- Fixed ATOTM module as the event listener was listening to talent ID on a damage event listener (instead of the RSK damage id)


src/analysis/retail/monk/mistweaver/modules/spells/RisingSunKick.tsx
```diff
@@ -41,9 +41,6 @@ class RisingSunKick extends Analyzer {
 
     if (this.lastBOK > this.lastRSK) {
       this.rskResets += 1;
-
-      this.spellUsable.endCooldown(TALENTS_MONK.RISING_SUN_KICK_TALENT.id);
-      this.spellUsable.beginCooldown(event, TALENTS_MONK.RISING_SUN_KICK_TALENT.id);
```
kfinch, line 46: I thought you added these lines recently - is this a merge mistake?

trevorm4: The first PR was on the shadowlands branch, but this is for DF branch. Is there some syncing across them that I should've been using?

kfinch: No - I might be getting mixed up I thought these lines were supposed to be added not removed

trevorm4: Nope, they are what was causing the issue. It was restarting the CD here, but also getting restarted in SpellUsable.tsx which is what causes the issue.

kfinch: Apologies then, in you go

PR #5202, opened 2022-10-19
regenerate talents to support different spell IDs

ToppleTheNun:
warrior has a couple of talents that are in the
class tree but have different spell IDs
depending on the spec (e.g., War Machine,
Thunder Clap). This was causing issues on the
talent page. this should resolve it, though it
may cause issues if talents are regenerated.


danny-janse (approved):
This looks quite alright to me. I'm glad to see more spells returned that felt missing including Enhancements Ascendance (quite likely unrelated to your change). Not merging this yet myself because i'd like some else to have a look at it too :)

emallson:
> This was causing issues on the
talent page. this should resolve it, though it
may cause issues if talents are regenerated.

To clarify this: you mean that if other PRs regenerate talents before this is merged, it may create conflicts? or what other issues do you anticipate?

ToppleTheNun:
> To clarify this: you mean that if other PRs regenerate talents before this is merged, it may create conflicts? or what other issues do you anticipate?

I anticipate issues caused by the order in which the talents JSON is processed. For instance, if this iteration had Fury Warrior go first (it did), then `TALENTS_WARRIOR.WAR_MACHINE_TALENT` will be the Fury version. If this gets run again but Arms is the iteration that goes first, then `TALENTS_WARRIOR.WAR_MACHINE_TALENT` will be the Arms version, `TALENTS_WARRIOR.WAR_MACHINE_ARMS_TALENT` will disappear, and `TALENTS_WARRIOR.WAR_MACHINE_FURY_TALENT` will be added.

One thought I had on this was to force some kind of ordering on the processing.

emallson:
I see. We likely need a different approach then. If the same talent exists
in two spec trees, each should have the spec name in it imo. If the
previous was a class talent then it puts us in a little but of a word spot

This can be done by adding the key to a conflict list after first
encountering a conflict and then rekeying the previous talent seen.

Pewtro:
I have a separate fix locally that should alleviate this issue without caring for processing ordering

ToppleTheNun:
Putput's version does what this PR did in a better way, so closing it in favor of that one.

PR #5203, opened 2022-10-19
Added new ExplanationAndData panel, and used it in Feral Druid Guide's 'Core Rotation' section

kfinch:
Starting with this, but considering future work with a global cookie-saved toggle to 'reduce explanations'. ExplanationAndData could then be modified to detect this flag and hide the explanation column.

image


src/analysis/retail/druid/feral/Guide.tsx
```diff
@@ -64,15 +65,19 @@ function CoreRotationSection({ modules, events, info }: GuideProps<typeof Combat
...
-      {modules.ferociousBite.guideSubsection}
-      {modules.rakeUptime.guideSubsection}
-      {info.combatant.hasTalent(TALENTS_DRUID.LUNAR_INSPIRATION_TALENT) &&
-        modules.moonfireUptime.guideSubsection}
-      {info.combatant.hasTalent(TALENTS_DRUID.ADAPTIVE_SWARM_TALENT) &&
-        modules.adaptiveSwarm.guideSubsection}
+      <ExplanationAndData
+        rows={[
+          info.combatant.hasTalent(TALENTS_DRUID.BLOODTALONS_TALENT) &&
+            modules.bloodtalons.explanationAndData,
```
emallson: i think my main thought on this is that this top-level high-order component feels out of place. the same thing seems like it could be accomplished by having the `guideSubsection` methods produce something like:

```tsx
<NarrowWideColumns>
  <Explanation>
   explanation goes here
  </Explanation>
  {data goes here}
</NarrowWideColumns>
```

without needing the top-level component or passing generated components around as the `{ explanation, data }` objects.  

if we want to move towards the explanation toggle, this could become:

```tsx
<ExplanationRow>
  <Explanation>
   explanation goes here
  </Explanation>
  {data goes here}
</ExplanationRow>
```

which could automatically do the right thing when the explanation is hidden.

but: you can basically do that with the current system if you wanted so i'm not fundamentally opposed.

kfinch: I think my main motivation for doing it this way was I figured I'd make future layout changes and didn't want to have to update things in each module individually - but I do take your point. I kinda like your idea with `ExplanationRow` and `Explanation`, I might switch to that

kfinch: Ok I refactored let me know what you think

PR #5210, opened 2022-10-20
WOTLK: Adding Classic Specific Changelog

TJMoats:


emallson:
It's been a long week but I don't see an issue with having these in the core changelog. Is the concern that retail players won't want to look through the classic changelog entries?

I'm not gonna black separating it, I just either missed or forgot the discussion about this change

TJMoats:
That was my concern, but if you don't think it needs to be split out, I'm cool with that as well.

PR #5212, opened 2022-10-20
WOTLK: Adding Classic Specific Spell Defs

TJMoats:
We have various places in code where we display spell information. This information is generally manually created per class for retail. Normally, spells will be looked up against our manually generated list of spells. If the spell isn't defined, we make a call to the Blizz api to get the spell information. If there is no spell in our defs, and there is no spell in the api, the spell shows up as an unknown spell. This happens infrequently in retail, and when it does, we can just manually add the spell to the list.

Classic has no spell list defined, and no api to fallback on. That makes it so there are a ton of spells that show up as unknown. Even when the spells *do* show up, they are using retail spell info (See: [Hymn of Hope](https://www.wowhead.com/wotlk/spell=64901/hymn-of-hope) vs [Symbol of Hope](https://www.wowhead.com/spell=64901/symbol-of-hope)) We don't currently have any way to cleanly define spells that are in classic, but not in retail. This PR remedies this by adding a classic specific spell module.

## Considerations

- We want to change the existing functionality as little as possible, so all of the changed functions should default to `DragonFlight`.
- We want this to be done as close to the spell fetch as possible
- There are a number of spells that currently work by looking at the retail spells. This isn't a problem if the spell hasn't changed, but we will see some additional `Unknown Spell`s until we add them to the classic spell list.
- We don't want to have to define every spell in the game, so there will still be a call to the api if we don't have an override.

Before 

WjaK40a

After

fBdDDmS


src/common/SPELLS/classic/index.ts
```diff
@@ -0,0 +1,77 @@
...
+          `Attempted to retrieve invalid or missing spell from SPELLS: ${String(
+            prop,
+          )}. If this is expected, use SPELLS.maybeGet.`,
+        );
+      }
+    }
+
+    return value;
+  },
+});
```
emallson, line 61: this should probably just become a function that is used by both classic and non-classic SPELLS tables. just `(table) => new Proxy(table, ....);` and then call it on each.

src/common/SPELLS/index.ts
```diff
@@ -91,8 +93,13 @@ const SPELLS = new Proxy(InternalSpellTable, {
 });
 
 export default SPELLS;
-export const maybeGetSpell = (key: string | number | undefined): Spell | undefined =>
-  key ? InternalSpellTable[key as any] : undefined;
+
+export function maybeGetSpell(key: string | number | undefined, expansion = Expansion.Dragonflight): Spell | undefined {
```
emallson, line 97: note for later: we probably want an enum to represent which *game* you're looking at (Retail or Classic) instead of the expansion to avoid having to change all the calls to this later.

you don't have to worry about fixing that. i just am making a note so i remember

emallson (changes requested):
one small change to deal with some duplicated code, otherwise good

src/common/SPELLS/index.ts
TJMoats: We can add that to the expansion context. You can have it provide `isClassic` and `isRetail` pretty easily.

src/analysis/retail/priest/discipline/modules/spells/Expiation.tsx
```diff
@@ -53,15 +53,15 @@ class Expiation extends Analyzer {
       return;
     }
 
-    if (damageEvent.ability.guid === SPELLS.EXPIATION_DAMAGE.id) {
+    if (damageEvent?.ability.guid === SPELLS.EXPIATION_DAMAGE.id) {
```
TJMoats: Manually typing the spells bit pointed this out as an error.

src/common/SPELLS/classic/index.ts
TJMoats: I tried to do this change, but I'm afraid I can't make heads or tails of the expected typing here. I got the linter and node to stop complaining, but there is a test failing and I'm not sure how to fix it.

emallson:
Okay. Don't let it block this. Can merge without that. I think when we next
do a TS version bump this machinery mostly goes away anyway

PR #5220, opened 2022-10-22
Fix EF cancellation when buffed by TFT

trevorm4:
TFT can now buff EF and it halves the channel duration, however we hardcoded expected duration. This diff just divides expected duration by 2 if player is buffed by TFT

image
Still works on a bad log, but is fixed for a good log

image


src/analysis/retail/monk/mistweaver/modules/spells/EssenceFontCancelled.tsx
```diff
@@ -44,6 +47,10 @@ class EssenceFontCancelled extends Analyzer {
       );
     }
     this.expected_duration = (3000 + extra_secs * 1000) / (1 + this.haste.current!);
+    // TFT ef has half duration
+    if (this.combatants.players[event.sourceID].hasBuff(TALENTS_MONK.THUNDER_FOCUS_TEA_TALENT.id)) {
```
danny-janse: `this.combatants.players[event.sourceID]` should be `this.selectedCombatant`.

danny-janse (changes requested):

trevorm4:
Changed to use selected combatant

PR #5225, opened 2022-10-22
[Talents] Fix issues with duplicate talent names in class trees

Pewtro:
A different approach to https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5202 that shouldn't care about processing order. 

If it detects a duplicate talentname in a class tree, all those talents get keyed with spec name (example that War Machine is now renamed WAR_MACHINE_{SPEC}_TALENT for all three specs)


ToppleTheNun:
This does what I needed in a better way, so thank you for that!

src/common/TALENTS/warrior.ts
```diff
@@ -337,6 +379,30 @@ const talents = createTalentList({
     icon: 'ability_creature_cursed_04',
     maxRanks: 1,
   },
+  WAR_MACHINE_TALENT: {
```
ToppleTheNun: This is the Arms version of the talent, as a note.

Pewtro: oh no did war machine still end up weird, not like this

ToppleTheNun (approved):

PR #5232, opened 2022-10-23
add Soul Fragment buff stack tracking to Guide

ToppleTheNun:
image

### DIFF
--- a/src/parser/shared/modules/BuffStackTracker.ts
+++ b/src/parser/shared/modules/BuffStackTracker.ts
@@ -1,0 +1,107 @@
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Spell from 'common/SPELLS/Spell';
+import Events, { ApplyBuffEvent, ChangeBuffStackEvent, RemoveBuffEvent } from 'parser/core/Events';
+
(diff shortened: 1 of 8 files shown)


src/parser/shared/modules/BuffStackTracker.ts
```diff
@@ -0,0 +1,107 @@
...
+  current: number;
+};
+
+type BuffStackChange =
+  /** Player gained a buff stack */
+  | 'apply'
+  /** Player lost a buff stack */
+  | 'remove';
+
+const DEBUG = true;
```
kfinch: false for production please

kfinch (changes requested):
Ah, a version of the ResourceGraph for BuffStack count - I think this is a good idea. Personally I think I need to push viewing improvements to the ResourceGraph to make it more usable, but it's a good placeholder graphic until can figure out how to make it better. Any improvements there could be carried over to this graph, so I have no issues. 

Ready to merge except for fixing the debug flag

kfinch:
The big thing I want to add to the ResourceGraph is some way to highlight in red places where waste happens, if I can get that working I'll carry it over to this too

kfinch (approved):

PR #5233, opened 2022-10-23
Vivacious Vivification module for mistweaver

trevorm4:
Vivacious vivification is a buff you get every 10 seconds that grants an instant vivify. When receiving the buff, the next buff cd starts and if you consume the instant vivify after the cd is done, then you will instantly get a new buff. However, if its been over 10 seconds since the last apply (you haven't consumed the first free vivify and 2nd free vivify is ready), you are essentially wasting buff uptime.

This module adds tracking for overcapping the buff and also shows healing done from instant vivifies.

Suggestion 
image

Panel in statistics tab
image


src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
```diff
@@ -0,0 +1,136 @@
...
+        </>,
+      )
+        .icon(TALENTS_MONK.VIVACIOUS_VIVIFICATION_TALENT.icon)
+        .actual(
+          `${formatDuration(this.wastedTime) + ' '}${t({
+            id: 'monk.mistweaver.suggestions.vivaciousVivification.wastedTime',
+            message: `of overcapping`,
+          })}`,
+        )
+        .recommended(`<${formatNumber(recommended / 1000)} seconds of overcapping is recommended`),
```
ToppleTheNun, line 110: You could use `formatDuration` here, though that might change the output format.

src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
```diff
@@ -0,0 +1,136 @@
...
+        tooltip={
+          <ul>
+            <li>Total Healed: {formatNumber(this.totalHealed)} </li>
+            <li>Total Instant Casts: {this.totalCasts}</li>
+          </ul>
+        }
+      >
+        <TalentSpellText talent={TALENTS_MONK.VIVACIOUS_VIVIFICATION_TALENT}>
+          <ItemHealingDone amount={this.totalHealed} />
+          <br />
```
ToppleTheNun: You don't need the `<br />` here if you're not adding an additional item.

src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
```diff
@@ -0,0 +1,136 @@
...
+  wastedTime: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS_MONK.VIVACIOUS_VIVIFICATION_TALENT);
+    if (!this.active) {
+      return;
+    }
+    this.addEventListener(Events.death.by(SELECTED_PLAYER), (event) => {
+      this.lastApply = 0;
```
ToppleTheNun, line 33: Good catch here.

ToppleTheNun (changes requested):
Just one extra `<br />` cleanup and this should be good to go.

src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
trevorm4: That will give 0:30, but I think just having 30 is a bit cleaner for this suggestion.

ToppleTheNun (approved):

PR #5238, opened 2022-10-23
Implement Yulon's Whisper module for mistweaver

trevorm4:
Added module for Yulon's whisper which applies a 3 tick hot on up to 6 people in front of you when you press tft. This module adds a suggested avg of 5 or above and adds suggestion if necessary

Suggestion
image
Statistic
image


kfinch (commented):
Is it always 3 ticks, or can this number go up at high haste? From where I'm sitting, it would make more sense to count `ApplyBuff` and `RefreshBuff` events rather than the heal - then you wouldn't have to deal with the tick counting stuff.

trevorm4:
> Is it always 3 ticks, or can this number go up at high haste? From where I'm sitting, it would make more sense to count `ApplyBuff` and `RefreshBuff` events rather than the heal - then you wouldn't have to deal with the tick counting stuff.

It is always 3 ticks, it is not hasted. The issue is that this hot actually doesn't have an applybuff/removebuff, only healevent for some reason, so I had to resort to this approach.

kfinch:
> > Is it always 3 ticks, or can this number go up at high haste? From where I'm sitting, it would make more sense to count `ApplyBuff` and `RefreshBuff` events rather than the heal - then you wouldn't have to deal with the tick counting stuff.
> 
> It is always 3 ticks, it is not hasted. The issue is that this hot actually doesn't have an applybuff/removebuff, only healevent for some reason, so I had to resort to this approach.

B L I Z Z A R D

PR #5243, opened 2022-10-25
WOTLK: Updating Food Checker

TJMoats:
I've added all of the foods for Wotlk classic, and modified the suggestions to recommend other food when the food is objectively better.

<details>
<summary>No Food</summary>

iXqGvM5
</details>

<details>
<summary>Low Tier Food</summary>

WGq7f6I
</details>

<details>
<summary>Debug Mode</summary>

1sOPSP6
</details>


src/common/ITEMS/classic/others.ts
```diff
@@ -0,0 +1,6 @@
+import { itemIndexableList } from 'common/ITEMS/Item';
```
TJMoats, line 1: I left this in more as a "this is how you can add additional items to classic". We can remove the other item list if anyone feels strongly about it, but I felt that leaving this here sets up a pattern for others to follow.

src/common/ITEMS/index.ts
```diff
@@ -33,6 +34,7 @@ const ITEMS = {
     SHAMAN,
     WARLOCK,
     WARRIOR,
+    CLASSIC
```
TJMoats, line 37: I didn't bother breaking the classic item list into its own module like we did for the spell list. There isn't any overlap in ID's for items that I know of, so I didn't see any value of doing so for the classic items. There may be an argument for optimization, but we aren't adding very many items in classic for now.

kfinch (approved):

PR #5245, opened 2022-10-25
WOTLK: Updating Potion Recommendations

TJMoats:
<details>
<summary>With Potion Use</summary>

f09SJbW
</details>
<details>
<summary>Without Potion Use</summary>

JaPDnJC
</details>

PR #5247, opened 2022-10-25
WOTLK: Adding Priest Spell Defs

TJMoats:
Should be merged after https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5245


src/parser/shared/modules/StatTracker.ts
```diff
@@ -671,16 +671,16 @@ class StatTracker extends Analyzer {
       if (returnRatingForNextPercent) {
         //Returns the rating needed for 1% at current rating levels
         return (
-          (baselineRatingPerPercent / (1 - penaltyThresholds[idx - 1].penaltyAboveThis) / coef) *
+          (baselineRatingPerPercent / (1 - (penaltyThresholds[idx - 1].penaltyAboveThis || 0)) / coef) *
```
TJMoats: Negative index crashes this. I think that at index of 0, it should be 0, but I don't know enough about this module to know if this will break things.

Pewtro: How are you getting negative index here

Pewtro: Also to be clear, this function cannot be used at all for anything wotlk

Pewtro: The only way to get a negative index there is if a player had below 0% of a given secondary/tertiary from rating, which shouldn't be possible

TJMoats: It was crashing on this log: 

https://dragonflight.wowanalyzer.com/report/Ntvc2rAQ3TFL4k1K/56-Normal+Sartharion+-+Kill+(5:41)/Tolario/disc/character

Looking at the code, I'm not sure exactly what the root cause was.

I'll work on making this page more WOTLK friendly when I get a chance.

Pewtro: If wotlk logs are using the stattracker they will never be accurate. That is strictly tied to retail stats and using the diminishing returns sytem in place from Shadowlands and onwards

TJMoats: I reverted this in favor of just removing the stats bit from the character page. My linter is changing some of the logs on that file though.

Pewtro: Is your formatter setup differently than the projects? even if I force format the file, I don't get any changes

Pewtro:
Other than the weird formatting in stattracker, seems fine. 
You might wanna change the date in the changelog though ;)

Pewtro (approved):

TJMoats:
> You might wanna change the date in the changelog though

Uhhh, am I missing something there? 

`change(date(2022, 10, 25), 'Defining more priest spells', Khadaj),`

PR #5252, opened 2022-10-25
improve patch version checking

ToppleTheNun:
**Shadowlands**
image

**WotLK**
image


src/interface/report/PATCHES.ts
```diff
@@ -5,67 +5,97 @@
...
+  expansion: Expansion;
 }
 
 const PATCHES: Patch[] = [
+  {
+    name: '3.4.0',
+    timestamp: 1664229600000, // GMT: Monday, 26 September 2022 22:00:00
+    urlPrefix: '',
+    isCurrent: true,
+    gameVersion: 4, // WotLK
```
danny-janse, line 25: I was thinking if we could turn this into an enum? It occours several times now through out more of the code base. Probably not something to change in the current PR but somethig to put on the backlog?

danny-janse (approved):
Lgtm! Small remark but thats beyond the scope.

src/interface/report/PATCHES.ts
kfinch: Probably a good thing for a follow on PR

PR #5254, opened 2022-10-26
add TalentCastEfficiencyRequirement component

ToppleTheNun:
add CombatLogParserContext to make CombatLogParser available
via hooks for functional components.

add TalentCastEfficiencyRequirement component
that will only render if the selected combatant
has the given talent.

TalentCastEfficiencyRequirement will also fetch
the cast efficiency of the given talent so that
we don't need to set up an AbilityRequirement in
each spec checklist.

update Vengeance checklist to use it as an
example.

screenshot of it in use:
image

### DIFF
--- a/src/interface/report/CombatLogParserContext.tsx
+++ b/src/interface/report/CombatLogParserContext.tsx
@@ -1,0 +1,26 @@
+import CombatLogParser from 'parser/core/CombatLogParser';
+import { createContext, ReactNode, useContext } from 'react';
+
+export interface CombatLogParserContext {
+  combatLogParser: CombatLogParser;
+}
+
+// This starts off undefined as we don't have an instance of CombatLogParser to work with
+// until it gets provided by the Provider.
+export const CombatLogParserCtx = createContext<CombatLogParserContext | undefined>(undefined);
+
+interface Props {
+  children: ReactNode;
+  combatLogParser: CombatLogParser;
+}
+export const CombatLogParserProvider = ({ children, combatLogParser }: Props) => (
+  <CombatLogParserCtx.Provider value={{ combatLogParser }}>{children}</CombatLogParserCtx.Provider>
+);
+
+export const useCombatLogParser = () => {
+  const context = useContext(CombatLogParserCtx);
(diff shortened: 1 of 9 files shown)


src/interface/report/CombatLogParserContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+  <CombatLogParserCtx.Provider value={{ combatLogParser }}>{children}</CombatLogParserCtx.Provider>
+);
+
+export const useCombatLogParser = () => {
+  const context = useContext(CombatLogParserCtx);
+  if (context === undefined) {
+    throw new Error('useCombatLogParser must be within CombatLogParserProvider');
+  }
+  return context;
+};
```
emallson: we may want to move the `Guide`'s `useAnalyzer` hook over here. its been quite useful.

(not for this PR tho, just a note for later)

src/interface/report/Results/index.tsx
```diff
@@ -266,9 +281,11 @@ class Results extends React.PureComponent<Props, State> {
...
 
         return (
-          <div className="container">
-            <ErrorBoundary>{tab ? tab.render() : '404 tab not found'}</ErrorBoundary>
-          </div>
+          <CombatLogParserProvider combatLogParser={parser}>
+            <div className="container">
+              <ErrorBoundary>{tab ? tab.render() : '404 tab not found'}</ErrorBoundary>
+            </div>
+          </CombatLogParserProvider>
```
emallson: rather than wrapping each return value here, we should wrap the `renderContent` call on line 463 IMO

emallson (changes requested):
this looks like a huge improvement overall, but one small change for the core `Results/index.tsx` when applying the context provider

src/interface/report/CombatLogParserContext.tsx
ToppleTheNun: I have another PR where I'm adding a lot of contexts, that would probably fit very well with that PR.

emallson (approved):

PR #5260, opened 2022-10-26
convert the report load process to FCs and hooks

ToppleTheNun:
Provide all data previously passed via functions through context instead. Enables consumption
farther down the component tree without needing to prop drill.

Can break this down into smaller PRs if needed.

Demonstration of report loading still working as expected

report_loading

### DIFF
--- a/src/interface/report/PlayerLoader.tsx
+++ b/src/interface/report/PlayerLoader.tsx
@@ -1,3 +1,4 @@
+import { useCallback, useEffect, useReducer } from 'react';
 import { Trans, t } from '@lingui/macro';
 import { captureException } from 'common/errorLogger';
 import { fetchCombatants, LogNotFoundError } from 'common/fetchWclApi';
@@ -5,11 +6,10 @@ import getFightName from 'common/getFightName';
 import getAverageItemLevel from 'game/getAverageItemLevel';
 import ROLES from 'game/ROLES';
 import SPECS from 'game/SPECS';
-import { wclGameVersionToExpansion } from 'game/VERSIONS';
+import { isUnsupportedClassicVersion, wclGameVersionToExpansion } from 'game/VERSIONS';
 import { fetchCharacter } from 'interface/actions/characters';
 import { setCombatants } from 'interface/actions/combatants';
 import ActivityIndicator from 'interface/ActivityIndicator';
-import DocumentTitle from 'interface/DocumentTitle';
 import makeAnalyzerUrl from 'interface/makeAnalyzerUrl';
 import Panel from 'interface/Panel';
 import { RootState } from 'interface/reducers';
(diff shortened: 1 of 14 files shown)


ToppleTheNun:
@emallson I can pull the guide hooks over in this PR if you'd like to get it done at once or in a separate PR. It seems a bit like mixing of concerns, but I'll leave it up to you how you'd like to review the changes. I can also break this PR up.

src/interface/report/PlayerLoader.tsx
```diff
@@ -62,148 +59,188 @@ interface State {
...
-  componentDidMount() {
-    // noinspection JSIgnoredPromiseFromCall
-    this.loadCombatants(this.props.report, this.props.fight);
-    this.scrollToTop();
-  }
+const PlayerLoaderFC = ({ children, setCombatants, history, playerName, playerId }: Props) => {
+  const [
+    { error, combatants, combatantsFightId, tanks, healers, dps, ranged, ilvl },
+    dispatchFC,
+  ] = useReducer(fcReducer, defaultStateFC);
```
emallson, line 128: we do have redux in the codebase, but i don't think that is an adventure for today :upside_down_face:

src/interface/report/PlayerLoader.tsx
```diff
@@ -237,154 +274,146 @@ class PlayerLoader extends React.PureComponent<Props, State> {
...
-      return this.renderError(error);
-    }
+  if (error) {
+    return renderError(error);
+  }
 
-    if (report.gameVersion === 2) {
-      return this.renderClassicWarning();
-    }
+  if (selectedReport.gameVersion === 2 || selectedReport.gameVersion === 3) {
```
emallson: this should use something more clear than checking if gameVersion is 2 or 3. This is actually checking `isClassic(selectedReport)` right? (I don't think that method exists)

src/interface/report/PlayerLoader.tsx
```diff
@@ -62,148 +59,188 @@ interface State {
...
+    dispatchFC,
+  ] = useReducer(fcReducer, defaultStateFC);
+  const { report: selectedReport } = useReport();
+  const { fight: selectedFight } = useFight();
+
+  const loadCombatants = useCallback(
+    async (report: Report, fight: WCLFight) => {
+      if (report.gameVersion === 2 || report.gameVersion === 3) {
+        return;
+      }
```
emallson: This should use some kind of `isClassic` check instead of checking bare gameVersion ids

emallson (changes requested):
minor change (some kind of `isClassic` call) then this looks okay. did some local testing and didn't see any freezes (which happened with the last core class -> hooks conversion :sweat_smile:)

emallson (approved):

PR #5269, opened 2022-10-29
[Shaman] Prepare restoration for DF

avilene:
- Cleanup legendaries/covenants
- Adjust spells to talents
- Fix issues with healing stream vs cloudburst as they are both talents now

Also adjusted my own contributor info as it has updated since then (it's been a while everyone!)


kfinch (approved):
Was at first scaled by the 55 file diff, but looks like just talent format changes are removing the old stuff - very nice

PR #5270, opened 2022-10-29
evang analyzer first steps

oisinod:
right now the core code is there to start looking for problems in ramps. next pr will branch off of this one, which will start to display the first set of problems


src/analysis/retail/priest/discipline/modules/guide/EvangelismAnalysis.tsx
```diff
@@ -0,0 +1,122 @@
...
+  checkForWrongCasts(ramp: CastEvent[]) {
+    return ramp
+      .map((cast, index) => {
+        if (!ALLOWED_PRE_EVANG.includes(cast.ability.guid)) {
+          return index;
+        }
+        return null;
+      })
+      .filter((cast) => !isNullOrFalse(cast));
+  }
```
kfinch: A bit confused how this works - looks like `checkForWrongCasts` builds and returns an array of the indexes of wrong casts, but then `analyzeSequence` just discards that array? Nothing is being done with the return value.

src/analysis/retail/priest/discipline/modules/guide/EvangelismAnalysis.tsx
```diff
@@ -0,0 +1,122 @@
...
+    while (!PEMITTED_RAMP_STARTERS.includes(ramp[0].ability.guid)) {
+      ramp.shift();
+    }
+    this.analyzeSequence(ramp);
+    // this.evangSequences.push({ramp: ramp, details: {nonApplicatorCasts: []}})
+  }
+
+  analyzeSequence(ramp: CastEvent[]) {
+    this.checkForWrongCasts(ramp);
+  }
```
kfinch: Why does this method exist? Its only purpose is to call the next method in the sequence..

src/analysis/retail/priest/discipline/modules/guide/EvangelismAnalysis.tsx
```diff
@@ -0,0 +1,125 @@
...
+  checkForWrongCasts(ramp: CastEvent[]) {
+    return ramp
+      .map((cast, index) => {
+        if (!ALLOWED_PRE_EVANG.includes(cast.ability.guid)) {
+          return index;
+        }
+        return null;
+      })
+      .filter(Number);
+  }
```
kfinch: Pretty sure you should be able to make this return type `number[]`, which will make downstream handling less of a pain in the ass

src/analysis/retail/priest/discipline/modules/guide/EvangelismAnalysis.tsx
```diff
@@ -0,0 +1,125 @@
...
+  TALENTS_PRIEST.RENEW_TALENT.id,
+  SPELLS.FLASH_HEAL.id,
+  TALENTS_PRIEST.RAPTURE_TALENT.id,
+  SPELLS.POWER_WORD_SHIELD.id,
+];
+
+interface Ramp {
+  timestamp: number;
+  badCastIndexes?: number[];
+}
```
kfinch: You need `rampHistory` here too right? In order for the indices to mean anything

oisinod: Yup, great spot! 

I've added it now :)

kfinch (approved):

PR #5272, opened 2022-10-30
Add Classic WotLK specs

jazminite:
Add Classic WotLK specs to create a better organization for Classic (similar to Retail)


emallson:
Is there a place that you wanted to have access to a proper `Spec` for classic? The current classic setup is just classes w/ "builds".

If we wanted to copy the specs that are present on WCL, I could get the spec IDs used there for this and we could look at transitioning to that.

jazminite:
> Is there a place that you wanted to have access to a proper `Spec` for classic? The current classic setup is just classes w/ "builds".
> 
> If we wanted to copy the specs that are present on WCL, I could get the spec IDs used there for this and we could look at transitioning to that.

Yep, I was setting this up first to do the Warlock section. I used the IDs on wowhead which seems to be what the retail specs are using.

emallson:
Gotcha, so the plan with this is to change the Classic section to behave more like the Retail one with spec-specific analyzers? I'm okay with that direction but want to make sure that I understand things correctly.

I think @TJMoats is the other major classic contributor right now. Thoughts?

jazminite:
> Gotcha, so the plan with this is to change the Classic section to behave more like the Retail one with spec-specific analyzers? I'm okay with that direction but want to make sure that I understand things correctly.
> 

Yes. For me, it makes it easier to follow retail as an example for Classic.

jazminite:
Khadaj's comment for reference:
image

PR #5276, opened 2022-10-31
[Restoration Shaman] Undercurrent

niseko:
Simple Undercurrent statistic box.
Had to update some constants and add some healing ids to make it accurate. 

image


src/analysis/retail/shaman/restoration/modules/talents/Undercurrent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.OPTIONAL(20)}
+        category={STATISTIC_CATEGORY.TALENTS}
+        size="flexible"
+      >
+        <BoringSpellValueText spellId={TALENTS_SHAMAN.UNDERCURRENT_TALENT.id}>
```
kfinch, line 54: Can use the new TalentSpellText object here instead, it shows rank but is otherwise the same: `<TalentSpellText talent={TALENTS_SHAMAN.UNDERCURRENT_TALENT}>`

Looks like this (from a Druid module): 
image

src/analysis/retail/shaman/restoration/modules/talents/Undercurrent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  talentRank = 0;
+
+  constructor(options: Options) {
+    super(options);
+
+    this.talentRank = this.selectedCombatant.getTalentRank(TALENTS_SHAMAN.UNDERCURRENT_TALENT);
+    if (!this.talentRank) {
+      this.active = false;
+      return;
+    }
```
kfinch, line 32: Could be `this.active = this.talentRank > 0;` - the return is actually not needed because event listeners aren't added for inactive modules.

src/analysis/retail/shaman/restoration/modules/talents/Undercurrent.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER | SELECTED_PLAYER_PET).spell(ABILITIES_AFFECTED_BY_HEALING_INCREASES), this.heal);
+  }
+
+  heal(event: HealEvent) {
+    const target = this.combatants.players[event.targetID];
+    if (!target) {
+      return;
+    }
```
kfinch, line 41: What's this check for? You don't use target afterwards

src/analysis/retail/shaman/restoration/modules/talents/Undercurrent.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    const undercurrentHealIncrease = this.selectedCombatant.getBuffStacks(SPELLS.UNDERCURRENT_BUFF.id) * UNDERCURRENT_HEALING_INCREASE[this.talentRank];
+    this.healing += calculateEffectiveHealing(event, undercurrentHealIncrease / 100);
```
kfinch, line 44: Why not just make `UNDERCURRENT_HEALING_INCREASE = [0, 0.005, 0.01]` rather than always dividing by 100.

kfinch (commented):
Looks good, a few minor quibbles that are non-blocking

src/analysis/retail/shaman/restoration/modules/talents/Undercurrent.tsx
avilene: I would assume its to ensure that the target healed was actually part of the fight, not something neutral outside of the fight? Although that could be included as "wasted" healing.

danny-janse:
@niseko @kfinch Merging this on request of @avilene. They'll pick up the 'issues' that kfinch pointed out.

PR #5277, opened 2022-10-31
Added Budding Leaves Statistic

kfinch:
image

### DIFF
--- a/src/analysis/retail/druid/restoration/modules/spells/BuddingLeaves.tsx
+++ b/src/analysis/retail/druid/restoration/modules/spells/BuddingLeaves.tsx
@@ -1,0 +1,101 @@
+import Analyzer, { SELECTED_PLAYER } from 'parser/core/Analyzer';
+import { Options } from 'parser/core/Module';
+import { TALENTS_DRUID } from 'common/TALENTS';
+import Events, { HealEvent, RemoveBuffEvent } from 'parser/core/Events';
+import { lifebloomSpell } from 'analysis/retail/druid/restoration/constants';
+import SPELLS from 'common/SPELLS';
+import { encodeEventTargetString } from 'parser/shared/modules/Enemies';
+import { calculateEffectiveHealing } from 'parser/core/EventCalculateLib';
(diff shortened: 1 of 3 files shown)


src/analysis/retail/druid/restoration/modules/spells/BuddingLeaves.tsx
```diff
@@ -0,0 +1,101 @@
...
+  stacksByTarget: Map<string, number> = new Map<string, number>();
+
+  /** Total healing attributable to this talent */
+  totalHealing: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+
+    this.ranks = this.selectedCombatant.getTalentRank(TALENTS_DRUID.BUDDING_LEAVES_TALENT);
+    this.active = this.ranks > 0;
```
ToppleTheNun, line 43: Can you add a return after the active check?
```suggestion
    this.active = this.ranks > 0;
    if (!this.active) {
      return;
    }
```

kfinch: Not actually needed because the event subscribe doesn't happen for inactive modules (this is why you won't see the return anywhere else)

Here is the implementation of Module's addEventListener:
```ts
addEventListener<ET extends EventType, E extends MappedEvent<ET>>(
    eventFilter: ET | EventFilter<ET>,
    listener: EventListener<ET, E>,
  ) {
    if (!this.active) {
      return;
    }
    this.owner.addEventListener(eventFilter, listener.bind(this), this);
  }
```

ToppleTheNun: I've been living a lie. Thanks!

ToppleTheNun (approved):

src/analysis/retail/druid/restoration/modules/spells/BuddingLeaves.tsx
kfinch: It's super hidden behavior, I acknowledge that 😅

PR #5285, opened 2022-11-01
[Restoration Shaman] Flash Floods update to talent ranks

avilene:


src/analysis/retail/shaman/restoration/modules/talents/FlashFlood.tsx
```diff
@@ -63,6 +63,8 @@ class FlashFlood extends Analyzer {
   constructor(options: Options) {
     super(options);
     this.active = this.selectedCombatant.hasTalent(TALENTS.FLASH_FLOOD_TALENT.id);
+    this.flashFloodHaste =
+      this.selectedCombatant.getTalentRank(TALENTS.FLASH_FLOOD_TALENT.id) * 0.1;
```
danny-janse: I would recommend turning the 0.1 into a constant instead. This way there will be no question in the future what this multiplication is for.

src/analysis/retail/shaman/restoration/modules/talents/FlashFlood.tsx
```diff
@@ -63,6 +63,8 @@ class FlashFlood extends Analyzer {
   constructor(options: Options) {
     super(options);
     this.active = this.selectedCombatant.hasTalent(TALENTS.FLASH_FLOOD_TALENT.id);
```
danny-janse, line 65: I'd recommend adding a return if the module should not be active before adding listeners and doing other logic.

danny-janse (changes requested):
Looks good! Could you make the two requested changed and add a changelog to finish it up?

src/analysis/retail/shaman/restoration/modules/talents/FlashFlood.tsx
kfinch: I've been specifically suggesting not to do this. It's hidden behavior, but listeners actually don't get added for inactive modules.

danny-janse: Oh interesting. I'm aware that listeners are not executed. The reason I've been suggesting this isbecause there can be unused code still running. Like calculating the effectiveness of an ability. Eventho its not a listener it is still code that has to be exectued in the browser - aka the end-user.

kfinch: It's such a trivial amount of extra code I doubt it matters much, but point taken. I still think the no return looks cleaner 🤷

kfinch (approved):

danny-janse (approved):

PR #5287, opened 2022-11-01
Added generic cast efficiency panel to Guide components.

kfinch:
Also used new panel to create a Brutal Slash subsection in the Feral Druid guide.
image

This PR also adds a `minimizeIcons` option to CooldownBar, and fixes an issue where charges weren't showing properly.


kfinch:
@emallson do you think I am making too much of an aesthetic decision by having `CastEfficiencyPanel` automatically wrapped by `RoundedPanel` - I could omit the RoundedPanel in CastEfficiencyPanel and just include that in BrutalSlash wrapping the CastEfficiencyPanel.

ToppleTheNun:
FWIW, I like the rounded panel being there. You could make it a boolean prop where if it's true, it gets wrapped in `RoundedPanel` and `React.Fragment` if not.

emallson:
I think its something we can revisit if it ends up being a thing we want to configure. I think its visually fine as-is (haven't read code yet)

emallson (approved):

Issue #5093, opened 2022-09-27
Figure out how unsupported / partially supported specs are to be presented


emallson:
With the release of Dragonflight, we are likely to have specs that have [literally no support](https://github.com/WoWAnalyzer/WoWAnalyzer/issues/5090). We need to figure out what we want to show in these cases. As part of that, it would be good to review what is currently shown for partially complete specs and improve clarity on what is complete & correct vs. incomplete / incorrect.

- [ ] Refine display of wholly unsupported specs
- [ ] Review existing handling of partially supported specs and see if there is room for further clarity
- [ ] (@emallson) Spin out any results of that review into implementation issue(s)

Issue #5090, opened 2022-09-27
Dragonflight Launch - Remove Code for Wholly Unsupported Specs


emallson:
[Documentation](https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Enabling-a-Spec-for-Dragonflight)

- [ ] Remove the `CombatLogParser` for each retail spec that hasn't received a compatibility update for Dragonflight (full support not required)
- [ ] Remove the `tsconfig.json` exclusions for all specs/classes.
- [ ] Check that this correctly:
  - [ ] Shows the spec on the Specs page as unsupported
  - [ ] Does not allow clicking through to an example log
  - [ ] Does not crash when visiting an unsupported spec from report character selection

ToppleTheNun:
I worked on this a bit and have a few notes:
- `parser` is currently required in `Config`. You can work around this by marking `parser` as optional in the interface. You then need to update `useParser` to conditionally invoke the function.
- Once the `tsconfig.json` exclusions are removed and you re-add the specs into `src/parser/index.ts`, you get a _bunch_ of errors from the freshly enabled specs. See below for an example: 
![image](https://user-images.githubusercontent.com/1672786/199144890-b54d82bb-d28e-435f-937e-130dcf412f00.png)
- Specs won't show up on the specializations page unless their expansion is set to `Expansion.Dragonflight` in their `Config`, which I took a swing at updating a few in #5279.

emallson:
I did a pass on this before my trip and ran into similar issues. I ended up commenting out the changelogs and combatlogparsers in particular, just leaving the config as part of the import tree. That way we can list the spec but not have to update the contents yet. Before doing that, though, I need to figure out what the updated documentation looks like.

It will probably also wait until #5091 gets some work done so we can better show totally unsupported specs.

ToppleTheNun:
Are we good to close this?

PR #5288, opened 2022-11-01
show all specs on specializations page

ToppleTheNun:
all specs are now visible on the specializations page, along with
whether or not they are supported. classes that have no supported
specs are now at the bottom of the list, subsequently ordered by
name.

addresses #5090, #5091, #5093

image

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -39,6 +39,18 @@
     "src/analysis/retail/shaman/shared",
     "src/analysis/retail/warrior/arms"
   ],
+  "files": [
+    "src/analysis/retail/deathknight/unholy/CONFIG.tsx",
+    "src/analysis/retail/druid/guardian/CONFIG.tsx",
+    "src/analysis/retail/hunter/beastmastery/CONFIG.tsx",
+    "src/analysis/retail/hunter/marksmanship/CONFIG.tsx",
+    "src/analysis/retail/hunter/survival/CONFIG.tsx",
+    "src/analysis/retail/monk/windwalker/CONFIG.tsx",
+    "src/analysis/retail/paladin/protection/CONFIG.tsx",
+    "src/analysis/retail/paladin/retribution/CONFIG.tsx",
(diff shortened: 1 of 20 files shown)


danny-janse (approved):

tsconfig.json
```diff
@@ -39,6 +39,18 @@
...
+    "src/analysis/retail/druid/guardian/CONFIG.tsx",
+    "src/analysis/retail/hunter/beastmastery/CONFIG.tsx",
+    "src/analysis/retail/hunter/marksmanship/CONFIG.tsx",
+    "src/analysis/retail/hunter/survival/CONFIG.tsx",
+    "src/analysis/retail/monk/windwalker/CONFIG.tsx",
+    "src/analysis/retail/paladin/protection/CONFIG.tsx",
+    "src/analysis/retail/paladin/retribution/CONFIG.tsx",
+    "src/analysis/retail/priest/shadow/CONFIG.tsx",
+    "src/analysis/retail/warrior/arms/CONFIG.tsx"
+  ],
```
emallson, line 53: since these are explicitly imported now, this is not necessary---typescript will check them despite the exclusion.

emallson (commented):
I'm doing some tweaking to this, in part because I also need to update the wiki for how to fix the config after this PR

tsconfig.json
emallson: actually maybe i'm wrong on this :thinking: gonna leave it alone for now

ToppleTheNun: Even if they're explicitly imported, because they're in `exclude`, TypeScript will ignore them. The only way to have TypeScript typecheck the Config file is to leave them in `files`. It was a wild afternoon finding that out.

emallson: That's really strange because the whole reason I had to comment out the imports is because of typescript typechecking everything that was imported, even if it was also excluded

emallson: TS does appear to hang if I remove them from files so I'm not touching it though lol

PR #5295, opened 2022-11-03
Holy priest two new talents

emilnormann:
Added support for Divine Word and Searing Light. 
Apart from that very minor change to ensure hymn healing is calculated correctly.


src/analysis/retail/priest/holy/modules/talents/BottomRow/DivineWord.tsx
```diff
@@ -0,0 +1,270 @@
...
+    super(options);
+    if (!this.selectedCombatant.hasTalent(TALENTS.DIVINE_WORD_TALENT.id)) {
+      this.active = false;
+      return;
+    }
+
+    // Healing spells that are affected by serenity buff
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell(SPELLS.FLASH_HEAL), this.onHeal);
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell(TALENTS.RENEW_TALENT), this.onHeal);
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell(SPELLS.GREATER_HEAL), this.onHeal);
```
kfinch: you can pass an array of spells to event listeners, so these three lines should be `    this.addEventListener(Events.heal.by(SELECTED_PLAYER).spell([SPELLS.FLASH_HEAL, TALENTS.RENEW_TALENT, SPELLS.GREATER_HEAL]), this.onHeal);`

Similar comment applies to several of the listener things below as well

src/analysis/retail/priest/holy/modules/talents/BottomRow/DivineWord.tsx
```diff
@@ -0,0 +1,270 @@
...
+        this.divineWordActive = false;
+        break;
+      case SPELLS.DIVINE_WORD_CHASTISE_TALENT_BUFF.id:
+        this.divineFavorChastiseActive = false;
+        break;
+      case SPELLS.DIVINE_WORD_SERENITY_TALENT_BUFF.id:
+        this.divineFavorSerenityActive = false;
+        break;
+    }
+  }
```
kfinch: The manual buff tracking is not needed - we have a utility that takes care of that for you. Get rid of the 'active' trackers and use `this.selectedCombatant.hasBuff(buffId)`

src/analysis/retail/priest/holy/modules/talents/BottomRow/SearingLight.tsx
```diff
@@ -0,0 +1,73 @@
...
+    this.targetsWithHolyFire.push(event.targetID);
+  }
+  onHolyFireFallOff(event: RemoveDebuffEvent) {
+    const index = this.targetsWithHolyFire.indexOf(event.targetID);
+    //Every element of the array should be unique so we don't need to delete multiple times
+    //The target should also always be in the array but still error checking
+    if (index > -1) {
+      this.targetsWithHolyFire.splice(index, 1);
+    }
+  }
```
kfinch: A few problems here, and a few ways to fix it.

Problem 1: `event.targetID` isn't unique for enemies, you also have to factor in `event.targetInstance` (different mobs with the same name will have the same targetID but different targetInstance). There is a helper method to make this easier - make the type a string and use `encodeEventTargetString(event)`.

Problem 2: A `Set` is much better suited for this than an array. You can just use the `add` method when a debuff is applied and the `delete` method when it falls, and can check for presence with `has`. For an example of this, see Feral's `SuddenAmbush` module.

Problem 3: While the Set handling works, you actually could drop this way entirely and use the `Enemies` module to check for buff presence. (I'm ok if you want to keep the manual tracking here, just want you to be aware of the other way). You'd include `Enemies` in the module dependencies
```ts
static dependencies = {
    enemies: Enemies,
  };
  protected enemies!: Enemies;
```
Then `onSmite` you can check for buff presence:
```ts
const target = this.enemies.getEntity(event);
if (target !== null && target.hasBuff(SPELLS.HOLY_FIRE)) {
   ... damage calculate
```

kfinch (commented):
Some things to look at below. Also could you post a screenshot of what these new stats look like?

emilnormann:
Look of new panels:
image

image

src/analysis/retail/priest/holy/modules/talents/BottomRow/SearingLight.tsx
emilnormann: That is definitely a lot easier, ended up just using the enemies module.

kfinch (approved):

PR #5296, opened 2022-11-03
Enabling outlaw and fixing build errors.

jsucupira:
Enabling Outlaw rogue for DF.


src/analysis/retail/rogue/assassination/modules/talents/Subterfuge.tsx
```diff
@@ -8,7 +8,7 @@ import { SuggestionFactory, ThresholdStyle, When } from 'parser/core/ParseResult
 import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
 import ItemDamageDone from 'parser/ui/ItemDamageDone';
 import Statistic from 'parser/ui/Statistic';
-import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_CATEGORY from '/parser/ui/STATISTIC_CATEGORY';
```
danny-janse: Please remove the first '/'.

src/parser/core/Combatant.ts
```diff
@@ -484,7 +484,7 @@ class Combatant extends Entity {
    * This id is the same as the spell ID on Wowhead.
    */
   hasLegendary(legendary: LegendarySpell) {
-    if (!this.scannedForLegendaries && this.legendaries.size === 0) {
+    if (!this.scannedForLegendaries && this.legendaries.size < 1) {
```
danny-janse: What issue is resolved with this change?

tsconfig.json
```diff
@@ -35,6 +35,7 @@
     "src/analysis/retail/paladin/retribution",
     "src/analysis/retail/priest/shadow",
     "src/analysis/retail/rogue/assassination",
+    "src/analysis/retail/shaman/shared",
```
danny-janse: I remember removing this line in the tsconfig. Are you running into issues with the modules in this folder?

danny-janse (changes requested):
Looks quite alright for a first pass on outlaw. Few pointers and questions below :)

tsconfig.json
jsucupira: I had to merge the origin to my local.  This was a conflict I got.

src/parser/core/Combatant.ts
jsucupira: When the toon had more than one legendary, it was not picking up one of them.

Pewtro: This does nothing, don't merge with this change in.
 < 1 is the same as === 0 except it would be true for negative size (impossible).

tsconfig.json
jsucupira: @djanse did you want me to remove it again?

src/parser/core/Combatant.ts
jsucupira: @Pewtro @djanse reverted.

tsconfig.json
danny-janse: @jsucupira That would be preferable assuming it doesnt result into any errors.

jsucupira: @djanse it shouldn't, it was just a conflict, I am guessing I didn't solve it correctly.

danny-janse: If it does, just leave it as is and i'll go over it again :)

jsucupira: @djanse it built without an issue. 
image

danny-janse (approved):
Looks good to me. Analyzing a log also seems to work.
image

PR #5298, opened 2022-11-04
[MW MONK] Improve accuracy of Upwelling and Rising Mist module due to double counting and missing chiji gusts

trevorm4:
Required knowledge: direct heals on a target proc Gust of Mists, if target has Essence font they get 2 mastery heals (1 normal, 1 duplicated). FLS EF and EF mastery buffs do not both proc and give 3 mastery events. Chiji gust of mists also get duplicated when EF is on target (# of bok * 2 chiji mastery events)

- rewrote upwelling module to use HotTracker and a set to track which hots are from extra bolts
- Added tracking for extra gust events from chiji to rm/uw modules
- Changed upwelling hot healing tracker to only look at hot ticks between 8s-12s of hot existing, as it was double counting RM extended healing with upwelling hot healing
- changed upwelling module to also count mastery events in last 4s of FLS Essence font, so we attribute extra mastery events to RM only after 12 seconds.
- Made RM module only check extra mastery events between 12-24 seconds of hot existing on target, so we attribute the extra mastery procs to upwelling only

Before
image

After
image

This PR eliminates double counting but adds chiji gom tracking which makes it go up overall


vohrr (approved):

trevorm4:
my branch got borked, gonna make a new pr

PR #5300, opened 2022-11-04
blaze of light support

oisinod:


src/common/TALENTS/priest.ts
```diff
@@ -354,6 +354,12 @@ const talents = createTalentList({
     icon: 'spell_shadow_chilltouch',
     maxRanks: 2,
   },
+  BLAZE_OF_LIGHT_TALENT: {
+    id: 215768,
+    name: 'Blaze of Light',
+    icon: 'spell_holy_searinglight',
+    maxRanks: 2,
+  },
```
emallson: just to confirm, this was added by re-running the talent generation script?

oisinod: No, how do I run that?

emallson: There's a readme in the `scripts/talents` folder, but tldr is to run this on the command line (if you get a fetch error, your node version is too old and I can just run it for you):

```bash
# Windows
npx ts-node .\scripts\talents\generate-talents.ts
# Mac/Linux
npx ts-node ./scripts/talents/generate-talents.ts
```

PR #5304, opened 2022-11-05
[Hunter] Initial transition to Dragonflight

Pewtro:
Work in progress

The initial transition to Dragonflight for the three hunter specs - this gets them all in a state where they will render without issues. 

Remaining work to do is to ensure that suggestions are 100% accurate, functionality of modules 100% accurate, and more cleanup of spells/talents etc.


kfinch (approved):
There's a lot here but it looks sane

Pewtro:
Yeah, there's a lot happening - I've nulled the spec support so they'll show as unsupported, but then I'll merge and continue in follow up PRs on the remaining to dos (of which there are many)

PR #5310, opened 2022-11-06
Add icon lookup to Classic specs

jazminite:
All classic spec IDs on WCLs are zero but icon text is unique to specs.
Added icon lookup to Classic specs.


src/parser/getConfig.ts
```diff
@@ -1,16 +1,28 @@
...
+  icon?: string,
+) {
   const expansionConfigs = AVAILABLE_CONFIGS.filter((config) => config.expansion === expansion);
   let config = specId !== 0 && expansionConfigs.find((config) => config.spec.id === specId);
+  // Classic
   if (!config) {
     config = expansionConfigs.find((config) => config.spec.id === 0 && config.spec.type === type);
+    if (!config && icon) {
+      config = expansionConfigs.find(
+        (config) => config.spec.type === type && config.spec.icon === icon,
```
emallson, line 17: not a fan of searching by icon here. we should figure out what we want to assign for spec ids (WCL ids? other ids?) and use those instead of relying on the icon matching

jazminite: I'd like to match by spec id but WCLs has zero for every Classic spec Id. Is there a way to get them to change that? They are figuring out the spec on their end because they are determining an icon. It'd be nice if they used a unique id in their schema (can't clash with retail spec ids).

emallson: Yes, we have spec IDs internally but the specID in combatantinfo is from blizzard. we could potentially override that for classic. i'll ask kihra about it.

jazminite: @emallson - per our discussion that icon is the most convenient way to do this

emallson (approved):

PR #5313, opened 2022-11-07
Update PlayerTile for Classic

jazminite:
Update PlayerTile for Classic to clear console fetch errors.

Example:
image


src/interface/report/PlayerTile.tsx
```diff
@@ -29,7 +29,12 @@ const PlayerTile = ({ player, makeUrl, config }: Props) => {
 
   useEffect(() => {
     const load = async () => {
-      if (!player.region || !player.server || !isSupportedRegion(player.region)) {
+      if (
+        player.combatant.expansion === 'wotlk' ||
```
ToppleTheNun, line 33: I think this should be turned into a utility function that checks if the given combatant's expansion is a supported retail expansion, but we can do that in a later PR.

ToppleTheNun (approved):

PR #5318, opened 2022-11-08
[Mistweaver] JSS Uptime Fix

vohrr:
Jade Serpent Statue was looking at the wrong Id for calculating healing and uptime. Fixed. Also updated the Statistic Box for potential further implementation of Unison healing contribution (however we want to do that).

Before:
image

After:
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/JadeSerpentStatue.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/JadeSerpentStatue.tsx
@@ -1,7 +1,8 @@
 import { t } from '@lingui/macro';
+import SPELLS from 'common/SPELLS';
 import { formatPercentage } from 'common/format';
 import { TALENTS_MONK } from 'common/TALENTS';
-import { SpellIcon } from 'interface';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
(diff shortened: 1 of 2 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/JadeSerpentStatue.tsx
```diff
@@ -147,15 +148,9 @@ class JadeSerpentStatue extends Analyzer {
...
-          label={
-            <>
-              <SpellIcon id={TALENTS_MONK.SUMMON_JADE_SERPENT_STATUE_TALENT.id} /> % Uptime
-            </>
-          }
-        >
-          <>{formatPercentage(this.jadeSerpentStatueUptime)}</>
-        </BoringValueText>
+        <TalentSpellText talent={TALENTS_MONK.SUMMON_JADE_SERPENT_STATUE_TALENT}>
+        <>{formatPercentage(this.jadeSerpentStatueUptime)}% Uptime</>
```
ToppleTheNun: You don't need to wrap this in a fragment, it's already being wrapped by the surrounding outside component.

ToppleTheNun (changes requested):
One small fragment comment then we're good!

ToppleTheNun (approved):

PR #5319, opened 2022-11-08
[Mistweaver] Add Unison Module

vohrr:
Added unison module with support for Jade Serpent Statue

image

image

Had to use BoringSpellValueText and the Spell Id for Unison-Heal instead of the talent due to a documented issue with certain Talents not being part of the  Spells object and also having different IDs resulting in an unknown value for the SpellLink

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
@@ -1,0 +1,74 @@
+import { TALENTS_MONK } from 'common/TALENTS';
+import SpellLink from 'interface/SpellLink';
+import SPELLS from 'common/SPELLS';
+import Analyzer, { Options, SELECTED_PLAYER, SELECTED_PLAYER_PET } from 'parser/core/Analyzer';
+import Events, { HealEvent} from 'parser/core/Events';
+import { formatNumber } from 'common/format';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+
+class Unison extends Analyzer {
+healing: number = 0;
+overhealing: number = 0;
+healingFromJss: number = 0;
+
+constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(
+        TALENTS_MONK.UNISON_TALENT.id,
+    );
+    if(!this.active) {
+        return;
+    }
+    this.addEventListener(
+        Events.heal.by(SELECTED_PLAYER).spell(SPELLS.UNISON_HEAL),
+        this.unisonHeal,
+    );
+    if(this.selectedCombatant.hasTalent(TALENTS_MONK.SUMMON_JADE_SERPENT_STATUE_TALENT.id)) {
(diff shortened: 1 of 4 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
```diff
@@ -0,0 +1,74 @@
...
+import Events, { HealEvent} from 'parser/core/Events';
+import { formatNumber } from 'common/format';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+
+class Unison extends Analyzer {
+healing: number = 0;
```
ToppleTheNun: This formatting seems off?

src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    <li>
+                        Healing from <SpellLink id={TALENTS_MONK.SUMMON_JADE_SERPENT_STATUE_TALENT.id} />: {formatNumber(this.healingFromJss)} 
+                    </li>
+                    <li>
+                        Healing from <SpellLink id={TALENTS_MONK.SOOTHING_MIST_TALENT.id} />: {formatNumber(this.healing - this.healingFromJss)}
+                    </li>
+                </ul>
+            }
+        >
+            <BoringSpellValueText spellId={SPELLS.UNISON_HEAL.id}>
```
ToppleTheNun: The spell value here also doesn't have to be the exact ID of the heal; you could use `TalentSpellText` with the talent value and it should be good enough. This is also fine though, just wanted to point it out.

vohrr: I did try that at first, but the talent and heal id values are different and `TalentSpellText` only takes the `Talent` object (I'm pretty sure) as a parameter. So unfortunately it would return the proper icon and Wowhead link but the text returned 'Unknown'.

ToppleTheNun: That might need to be something we look at in `TalentSpellText`, then. We can ignore that for this PR, I'll take a peek

src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
vohrr: I'm not sure what you mean

kfinch: Oh yeah it looks here like these fields didn't get indented - but it's weird because I thought the linter took care of that

src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
trevorm4: Can confirm, rewrote it using TalentSpellText and ran into the same issue

trevorm4 (changes requested):
Talked to Vohrr and his yarn is messed up and not linting, going to request changes.

src/analysis/retail/monk/mistweaver/modules/spells/Unison.tsx
vohrr: Formatting done

trevorm4 (approved):

ToppleTheNun (approved):

PR #5333, opened 2022-11-09
Classic Shaman Spells

jazminite:
Update Classic Shaman Spells for WotLK
image


src/analysis/classic/shaman/modules/checklist/Component.tsx
```diff
@@ -35,15 +35,15 @@ const ShamanChecklist = ({ thresholds, totemTracker, build }: ChecklistProps) =>
           <Requirement
             name={
               <Trans id="shaman.restoration.checklist.appliedPrepull">
-                <SpellLink id={SPELLS.EARTH_SHIELD} /> applied prepull
+                <SpellLink id={SPELLS.EARTH_SHIELD.id} /> applied prepull
```
ToppleTheNun, line 38: You actually might want to leave `SpellLink` in classic code pointing to the spell object, as Classic doesn't have a spells API that data can be pulled from (which passing an id tries to make use of). `SpellLink` with a spell object will just render what it is given from the spell object.

ToppleTheNun (approved):

PR #5336, opened 2022-11-10
Add bombs and cleanup Classic Engineering

jazminite:
Add bombs and cleanup Classic Engineering
image

### DIFF
--- a/src/common/SPELLS/classic/engineering.ts
+++ b/src/common/SPELLS/classic/engineering.ts
@@ -1,10 +1,40 @@
 import { spellIndexableList } from '../Spell';
 
 const spells = spellIndexableList({
+  COBALT_FRAG_BOMB: {
+    id: 67769,
+    name: 'Cobalt Frag Bomb',
+    icon: 'inv_misc_enggizmos_31',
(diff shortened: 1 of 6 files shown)


src/common/SPELLS/classic/engineering.ts
```diff
@@ -1,11 +1,41 @@
...
 
 const spells = spellIndexableList({
+  COBALT_FRAG_BOMB: {
+    id: 67769,
+    name: 'Cobalt Frag Bomb',
+    icon: 'inv_misc_enggizmos_31',
+  },
+  FRAG_BELT: {
+    id: 67890,
+    name: 'Cobalt Frag Bomb',
```
ToppleTheNun, line 11: Should this also be `Cobalt Frag Bomb` if the key name is `FRAG_BELT`?

jazminite: It's a Cobalt Frag Bomb made from the Frag Belt. There's normal Cobalt Frag Bombs too.

ToppleTheNun (approved):

PR #5351, opened 2022-11-12
ResourceTracker updates

kfinch:
* Added the ability to query ResourceTracker for resource stats within an arbitrary time window
* Fixed some stat weirdness that could occur when a rateChange happens before the first spend/gain.
* Standardized data updates


kfinch:
It appears some commits from Mistweaver and Shadow made it in without being linted, because when I pulled from Dragonflight and merged to resolve a conflict, the linter appears to have hit a bunch of other files. 

The only actual changes in this PR are in ResourceTracker and ResourceGraph

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -500,68 +530,262 @@ class ResourceTracker extends Analyzer {
...
+      if (u.timestamp < startTime) {
+        lastBeforeEnd = u;
+      } else if (u.timestamp >= startTime && u.timestamp < endTime) {
+        // we're inside the segment
+        if (!addedFirstInside && (u.rateWaste || 0) > 0 && ix > 0) {
+          // this is the first update inside the segment and there's rate waste
+          // update waste with respect to start time using previous update's rate
+          const modifiedUpdate = { ...u }; // copy because we don't want to modify the original data
+          modifiedUpdate.rateWaste =
+            ((u.timestamp - startTime) / 1000) * this.resourceUpdates[ix - 1].rate;
```
emallson, line 640: this bit i don't understand, but i may just not have thought about it enough yet :thinking:

emallson (approved):

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
kfinch: I'll add some more comments because this is somewhat 'core' code.

I'll explain here as well: the `rateWaste` field expressed the amount of resource wasted due to natural regeneration since the last update, and also is only filled in when the previous update was resource capped. Consider the following example, presume a resource that regenerates 10 per second and has a max of 100.
* Player spends on t=0, and the resource fields say it was a spend of 20 from 90 -> 70.
* Player spends on t=5000, and the resource fields say it was a spend of 20 from 100 -> 80.

This will create the following resourceUpdates:
* Update 'spend' - timestamp: 0, current: 70, rate: 10 ....
* Update 'regenCap' - timestamp: 3000, current 100, rate: 10 .... (this is fabricated from the 2nd spend, using rate / max / current)
* Update 'spend' - timestamp: 5000, current 80, rate: 10, rateWaste: 20 .... (10 res per second * 2 seconds capped = 20 wasted)

Now consider the situation where the user wants resources for the subsection starting at t=4000. We get the slice of updates within the time period, including the 'spend' update at t=5000. The problem is that its rate waste field includes waste that happened before the time period, which is not desirable. I solve this by basically redoing the cap calculation assuming the cap started at the segment start time, so in this case it would be:
`modifiedUpdate.rateWaste = ((5000 - 4000) / 1000) * 10 = 10`

kfinch: And to be clear, I have to use the rate from the previous update because rates apply to everything after them, and I could miscalculate when a rateChange happens. Another example (base rate = 10, max = 100)
* Player gain resource and caps at t=0 (rate = 10)
* Bloodlust is used at t=4000, changing rate to 13 (player is still capped)

The updates:
* Update 'gain' - timestamp: 0, current: 100, rate: 10 ....
* Update 'rateChange' - timestamp: 4000, current: 100, rate: 13, rateWaste: 40

We got the rateWaste in the rateChange update using the previous rate, because its the rate that applied while resource was being wasted.

PR #5354, opened 2022-11-13
Add Classic Warlock Affliction spec

jazminite:
- Remove combined Warlock builds to make way for specs.
- Add information specific to Classic Affliction Warlocks:
image

- Add heroism / bloodlust to throughput cooldowns:
image


src/parser/shared/modules/CooldownThroughputTracker.tsx
```diff
@@ -80,6 +80,14 @@ class CooldownThroughputTracker extends Analyzer {
...
       summary: [BUILT_IN_SUMMARY_TYPES.DAMAGE, BUILT_IN_SUMMARY_TYPES.HEALING],
     },
+    {
+      spell: SPELLS.HEROISM.id,
+      summary: [BUILT_IN_SUMMARY_TYPES.DAMAGE, BUILT_IN_SUMMARY_TYPES.HEALING],
+    },
+    {
+      spell: SPELLS.BLOODLUST.id,
+      summary: [BUILT_IN_SUMMARY_TYPES.DAMAGE, BUILT_IN_SUMMARY_TYPES.HEALING],
+    },
```
emallson: I don't think this is a change that we want for retail. can you make it conditional on the expansion? (`this.owner.config.expansion` should tell you)

emallson (changes requested):

emallson:
One small change (retail considerations). Otherwise looks good

emallson (approved):

PR #5356, opened 2022-11-13
Added new CooldownUtilBar that combines CooldownBar with a cast effic report.

kfinch:
Also:
* Added optional perf color coding to CastEfficiencyPanel.
* Fixed an issue where Feral's per-Rake eval wasn't showing.

image

image

### DIFF
--- a/src/parser/ui/CooldownUtilBar.tsx
+++ b/src/parser/ui/CooldownUtilBar.tsx
@@ -1,0 +1,85 @@
+import { CooldownBar, GapHighlight } from 'parser/ui/CooldownBar';
+import styled from '@emotion/styled';
+import { SpellIcon, SpellLink, TooltipElement } from 'interface';
+import { BadColor, GoodColor, MediocreColor, OkColor, useAnalyzer } from 'interface/guide';
+import CastEfficiency from 'parser/shared/modules/CastEfficiency';
+import { formatPercentage } from 'common/format';
+
+/**
+ * Displays a spell's cast efficiency on the left,
+ * and then that spell's {@link CooldownBar} on the right.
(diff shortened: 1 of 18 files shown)


src/parser/ui/CooldownUtilBar.tsx
```diff
@@ -0,0 +1,85 @@
...
+  gapHighlightMode,
+  minimizeIcons,
+  useThresholds,
+}: {
+  spellId: number;
+  gapHighlightMode: GapHighlight;
+  minimizeIcons?: boolean;
+  useThresholds?: boolean;
+}): JSX.Element {
+  const castEffic = useAnalyzer(CastEfficiency)!.getCastEfficiencyForSpellId(spellId);
```
ToppleTheNun: I'd recommend using `useAnalyzer(CastEfficiency)?..getCastEfficiencyForSpellId(spellId)` instead for the semantic meaning. By using `!`, you're turning off the value that we get from TypeScript of knowing that `CastEfficiency` might be null/undefined.

They both ultimately do the same thing here, but since you're also checking for definition further down with `if (castEffic`, using the Elvis instead of the bang assertion should make it more clear what's happening.

ToppleTheNun (commented):
This is a bit bike-shed-y , but I think a name like `CooldownEfficiencyBar` might communicate the intent of the component better.

ToppleTheNun (approved):

kfinch:
Not sure what 'bike-shedy' means, but yeah I like your name better

src/parser/ui/CooldownUtilBar.tsx
kfinch: Yeah good point, fixed.

PR #5361, opened 2022-11-14
Mistweaver T29 Tier set module

trevorm4:
Adds module for MW tier set, see https://www.wowhead.com/news/monk-season-1-dragonflight-tier-set-bonuses-guide-writer-first-impressions-329299

image
image

This PR also adds filtering to the Rising Mist module, so it doesn't count hot ticks and extra vivifies that were from the tier set. It's only by a small amount because the tier set extends it by 1s but ReM ticks every 2 seconds

This implementation is pretty barebones atm and is undercounting vivify/rem healing because the extension mostly causes additional partial ticks, but the rem healing count is only counting ticks that occur inside the extension window. In future PR, will try to improve the accuracy, but for now this will serve as a lower bounds.


src/analysis/retail/monk/mistweaver/modules/dragonflight/tier/T29MWTier.tsx
```diff
@@ -0,0 +1,200 @@
...
+  totalExtensionDuration: number = 0;
+  twoPieceHealing: number = 0;
+  fourPieceHealingFromBuff: number = 0;
+  extraRemHealing: number = 0;
+  extraVivHealing: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.has2Piece = this.selectedCombatant.has2PieceByTier(TIERS.T29);
+    this.has4Piece = this.selectedCombatant.has4PieceByTier(TIERS.T29);
```
vohrr: `this.has4Piece = this.selectedCombatant.has4PieceByTier(TIERS.29) && this.has2Piece;`

src/analysis/retail/monk/mistweaver/modules/dragonflight/tier/T29MWTier.tsx
```diff
@@ -0,0 +1,200 @@
...
+    this.active = this.has2Piece || this.has4Piece;
```
vohrr: `|| this.has4Piece` is unnecessary because of your if statement on the event listeners below

src/analysis/retail/monk/mistweaver/modules/dragonflight/tier/T29MWTier.tsx
```diff
@@ -0,0 +1,200 @@
...
+    ) {
+      return;
+    }
+    const hot = this.hotTracker.hots[targetId][SPELLS.RENEWING_MIST_HEAL.id];
+    const extension = this.hotTracker.getRemExtensionForTimestamp(hot, event.timestamp);
+    if (extension?.attribution.name.startsWith(ATTRIBUTION_PREFIX)) {
+      this.extraRemHealing += (event.amount || 0) + (event.absorbed || 0);
+    }
+
+    if (HasRelatedEvent(event, 'FromMistyPeaks')) {
```
vohrr: either add a debug bool or remove

vohrr (changes requested):
Should be good for a base implementation. Down the road we should look to flesh out the detail of the hps contributions by source spell in the tooltip (or we could even put this into into its own donut chart) but for now this is good

src/analysis/retail/monk/mistweaver/modules/dragonflight/tier/T29MWTier.tsx
trevorm4: Oops yea, need to remove that. did not mean to include that in the pr

vohrr (approved):

PR #5362, opened 2022-11-14
[Mistweaver] Update Suggestion Threshold for Vivify based on talent selection

vohrr:
Updated the average Renewing Mist per Vivify cast suggestion thresholds to be accurate for all talent selections available in Dragonflight.

image

Average Renewing Mist count will vary depending on selecting Rising Mist, Rapid Diffusion, and Dancing Mist.

Added logic to hide the suggestion for the edge case where the player didn't cast vivify at all in a fight

Also corrected the formatting of the Enveloping Breath suggestion


src/analysis/retail/monk/mistweaver/modules/spells/Vivify.tsx
```diff
@@ -44,16 +66,51 @@ class Vivify extends Analyzer {
...
       },
       style: ThresholdStyle.NUMBER,
     };
   }
 
+  get estimatedAverageReMs() {
+    if (this.risingMistActive) {
+      this.expectedAverageReMs = 4.44;
+    } else {
+      this.expectedAverageReMs = 2.22;
```
trevorm4: 2.22/4.44 are kind of arbitrary and we may want to change them in the future due to talent tuning, so let's put `2.22` into a global const and then use that variable (or variable * 2 if they have RM)

src/analysis/retail/monk/mistweaver/constants.ts
```diff
@@ -35,6 +35,8 @@ export const LIFECYCLES_MANA_REDUCTION_PERCENT = 0.25;
 export const MANA_TEA_DURATION = 10000;
 export const JADE_BOND_INC = 0.4;
 export const NOURISHING_CHI_INC = 0.2;
+export const DANCING_MIST = 0.05; // per rank
+export const RAPID_DIFFUSION = 3000; // per rank
```
trevorm4:
```suggestion
export const DANCING_MIST_CHANCE = 0.05; // per rank
export const RAPID_DIFFUSION_DURATION = 3000; // per rank
```
Let's make these names a bit more specific

trevorm4 (changes requested):

trevorm4 (approved):
lgtm

PR #5376, opened 2022-11-16
add SuggestionSection Guide component

ToppleTheNun:
add SuggestionSection that can render the
suggestions returned by any provided module
definitions. for instance, giving this section
the FoodChecker and WeaponEnhancementChecker
modules as props will cause it to render
suggestions for the selected player having a
food buff when the fight started.

image


kfinch (approved):
Looks very nice - thanks

PR #5380, opened 2022-11-16
[StatTracker] & [Preservation] Racial Passives & Mastery Effectiveness Update 

vohrr:
- Adding passive stat increases from a few racial passives (Dracthyr, Goblin)  to StatTracker
  - I couldn't find anywhere else in the codebase where these stat increases are implemented but noticed that blood elf's crit increase was being applied here. If this is part of a larger planned implementation feel free to reject. 

- Updated MasteryEffectiveness for Evoker to clarify that the output only shows % of effective healing affected. Shields will need to be revisited once we get all raid buffs added in (notably Mark of the Wild for druids in this case).

image


src/parser/shared/modules/StatTracker.ts
```diff
@@ -591,12 +591,16 @@ class StatTracker extends Analyzer {
...
 
   get baseMasteryPercentage() {
     const spellPoints = 8; // Spellpoint is a unit of mastery, each class has 8 base Spellpoints
-    return (spellPoints * (this.selectedCombatant.spec?.masteryCoefficient || 1)) / 100;
+    let mastery = (spellPoints * (this.selectedCombatant.spec?.masteryCoefficient || 1)) / 100;
+    if (this.selectedCombatant.race === RACES.Dracthyr) {
+      mastery += 0.018;
+    }
+    return mastery;
   }
```
emallson, line 604: honestly not a HUGE fan of this but they change so rarely that they're effectively constants so 🤷

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -266,6 +235,73 @@ class MasteryEffectiveness extends Analyzer {
     this.totalMasteryHealingDone += absorbFromMasteryBonusUsed;
   }
 
+  __masteryBoostCheck(event: ApplyBuffEvent | RefreshBuffEvent, existingAbsorb: number) {
```
emallson: rather than the __ prefix, these should just be private functions. editor should be able to mass-rename and then adding `private` to them will be easy
```suggestion
  private masteryBoostCheck(event: ApplyBuffEvent | RefreshBuffEvent, existingAbsorb: number) {
```

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -266,6 +235,73 @@ class MasteryEffectiveness extends Analyzer {
...
+      if (lowerShieldBound <= event.absorb && event.absorb <= upperShieldBound) {
+        return true;
+      }
+    }
+    return false;
+  }
+  //TODO -- THESE CALCULATIONS ARE USING A +/- 10% THRESHOLD UNTIL ALL THE NEW RAID BUFFS ARE ADDED AND INCORPORATED INTO STATTRACKER
+  __calculateShield(event: ApplyBuffEvent | RefreshBuffEvent, existingAbsorb: number) {
+    let baseShielding = 0;
+    const vers = this.statTracker.currentVersatilityPercentage + 0.03; // TODO: REMOVE WHEN MARK OF THE WILD IS ADDED TO STATTRACKER
```
emallson: imo you should just let this be 3% off until MotW is in. no risk of forgetting / missing that its gone in

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -54,14 +57,16 @@ class MasteryEffectiveness extends Analyzer {
 
   private prevokerHealthPercent: number = 0;
   private prevokerMaxHealth: number = 0;
-  private anomalyShieldApplications: Map<number, ShieldInfo | null> = new Map();
-  private twinGuardianShieldApplications: Map<number, ShieldInfo | null> = new Map();
+  private anomalyShieldApplications: Map<[number, number], ShieldInfo | null> = new Map();
+  private twinGuardianShieldApplications: Map<[number, number], ShieldInfo | null> = new Map();
```
emallson: this isn't going to work the way you want. `Map` keys are compared using `Object.is`, which is effectively reference equality. Compare the output of `Object.is(1, 1)`, `Object.is([1, 1], [1, 1])`, and `const x = [1, 1]; Object.is(x, x)`. Only the 1st and 3rd should be true. The 2nd should be false, because each time you create the array it has a different reference---even if it has identical contents.

Instead, you should use a `string` key for this, and use a function to generate the keys like:

```ts
// FIXME: not sure what the params are off-hand
function shieldKey(a: number, b: number): string {
  return `${a}-${b}`;
}
```

emallson: Reading through this, I'm actually not sure that you need the 2nd part of the key. Looks like the key is `[targetID, spellID]` but `spellID` is effectively constant---it is always the same for TA, and always the same for TG, so you can just omit (like the old code did)

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -266,6 +235,73 @@ class MasteryEffectiveness extends Analyzer {
...
+    if (this.__isEventAbilityAnomaly(event.ability.guid)) {
+      const intellect = this.statTracker.currentIntellectRating;
+      baseShielding = intellect * TEMPORAL_ANOMALY_SHIELD_COEFF * (1 + vers) + existingAbsorb;
+    }
+
+    if (this.__isEventAbilityTwinGuardian(event.ability.guid)) {
+      const initialShieldAmount = this.prevokerMaxHealth * TWIN_GUARDIAN_SHIELD_COEFF;
+      const versAddedShield = this.prevokerMaxHealth * TWIN_GUARDIAN_SHIELD_COEFF * vers;
+      baseShielding = initialShieldAmount + versAddedShield + existingAbsorb;
+    }
```
emallson: these shields stack with themselves? trying to understand what you're doing with the existing absorb value here and that's what I think is going on.

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -266,6 +235,73 @@ class MasteryEffectiveness extends Analyzer {
...
+    return Math.round(baseShielding);
+  }
+
+  __isTargetHealthierThanPlayer(playerHealth: number, targetHealth: number): boolean {
+    return playerHealth >= targetHealth;
+  }
```
emallson: this is backwards, and could just be inlined since it is short and clear on its own. (meaning: remove this method, and just do `playerHealth < targetHealth` in the method

emallson (changes requested):
okay, so bad news and good news.

bad news: Mastery Effectiveness as-is doesn't work. You're never going to find existing absorbs, so the thing its trying to do with that isn't actually going to happen in practice.

good news: there isn't *that* much to do to make it work.

this module on the whole could use some comments explaining what the different things are, and what the calculation you're doing is for.

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
vohrr: Yeah the two don't stack with each other, but Temporal Anomaly stacks on itself which is fairly common as it pulses in a small radius every 2s. The refresh value needs to be evaluated with any existing shields added back in or its way off

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
vohrr: > okay, so bad news and good news.
> 
> bad news: Mastery Effectiveness as-is doesn't work. You're never going to find existing absorbs, so the thing its trying to do with that isn't actually going to happen in practice.
> 
> good news: there isn't _that_ much to do to make it work.
> 
> this module on the whole could use some comments explaining what the different things are, and what the calculation you're doing is for.

I think I'm going to strip away most of this pre-existing stuff then. The only thing that needs to be determined is whether or not an event had mastery applied to it and what percentage of total events were affected for this module to be valuable. The only hiccup there being refresh events on the absorbs that stack (see other comment below)

vohrr:
> okay, so bad news and good news.
> 
> bad news: Mastery Effectiveness as-is doesn't work. You're never going to find existing absorbs, so the thing its trying to do with that isn't actually going to happen in practice.
> 
> good news: there isn't _that_ much to do to make it work.
> 
> this module on the whole could use some comments explaining what the different things are, and what the calculation you're doing is for.

I think I'm going to strip away most of this pre-existing stuff then. The only thing that needs to be determined is whether or not an event had mastery applied to it and what percentage of total events were affected for this module to be valuable. Commits coming.

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
vohrr: You're correct, misread my first time looking at this and had it in my head both shields were being pushed to the same map. Fixed

src/analysis/retail/evoker/preservation/modules/core/MasteryEffectiveness.tsx
```diff
@@ -115,84 +88,68 @@ class MasteryEffectiveness extends Analyzer {
...
-      healing: 0,
     });
   }
 
   onAnomalyShieldApplication(event: ApplyBuffEvent | RefreshBuffEvent) {
+    this.totalShieldEvents += 1;
+    let existingAbsorb = 0;
+    //shield refresh event's absorb value is a sum of any exising absorb and the new absorb. find the existing absorb for calc in map then  remove
     if (this.anomalyShieldApplications.get(event.targetID)) {
+      existingAbsorb = this.anomalyShieldApplications.get(event.targetID)?.event.absorb || 0;
```
emallson, line 118: note: this won't handle cases where the value is reduced by ticking damage between shield applications/refreshes. not sure if that matters for this case, though

emallson (approved):
this is much easier to follow 👍

PR #5386, opened 2022-11-18
[Mistweaver] Fix rapid diffusion tooltip and remove Life cocoon statistic 

trevorm4:
Rapid diffusion tooltip had a typo. I also removed LC statistic because it was written with the assumption that we can track heals from all players, but this is not reality. In future, I will try to rewrite the module and do an ondemand WCL query, but until then this statistic is entirely useless.


src/analysis/retail/monk/mistweaver/modules/spells/LifeCocoon.tsx
```diff
@@ -39,22 +33,6 @@ class LifeCocoon extends Analyzer {
       this.healing += calculateEffectiveHealing(event, LIFE_COCOON_HEALING_BOOST);
     }
   }
-
```
vohrr: Instead of removing/changing anything inside the module right now  just remove the entry from the Mistweaver CombatLogParser.ts file and it won't load

vohrr (changes requested):

vohrr (approved):

PR #5389, opened 2022-11-18
add Food subsection to PreparationSection

ToppleTheNun:
Example usage:
```tsx
      <PreparationSection
        recommendedFoods={[
          SHADOWLANDS_SPELLS.TENEBROUS_CROWN_ROAST_ASPIC,
          SHADOWLANDS_SPELLS.FEAST_OF_GLUTTONOUS_HEDONISM_AGI,
        ]}
      />
```

image
image


src/interface/guide/components/Preparation/ExtraDivs.tsx
```diff
@@ -0,0 +1,39 @@
...
+    case QualitativePerformance.Perfect:
+      return PerfectColor;
+    case QualitativePerformance.Good:
+      return GoodColor;
+    case QualitativePerformance.Ok:
+      return OkColor;
+    case QualitativePerformance.Fail:
+      return BadColor;
+  }
+};
```
kfinch: I thought this function already existed in the guide index... if not, you should move it there because this should be a core Guide function

src/interface/guide/components/Preparation/ExtraDivs.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface Props {
+  performance: QualitativePerformance;
+}
+export const PerformanceRoundedPanel = ({ children, performance }: PropsWithChildren<Props>) => (
+  <RoundedPanelWithColorBoxShadow color={qualitativePerformanceToColor(performance)}>
+    {children}
+  </RoundedPanelWithColorBoxShadow>
+);
```
kfinch: Maybe this one can go into `GuideDivs` too

emallson (commented):
main comment is that the 👍 case still takes up a lot of space, but that isn't the worst thing in the world.

ToppleTheNun:
> main comment is that the 👍 case still takes up a lot of space, but that isn't the worst thing in the world.

I'm not overly fond of how it's shaping up for parts of the preparation section that realistically only have one piece of data to show. I'm fine with the enchantment and weapon enhancement sections since they show multiple pieces of data.

One of my thoughts was to switch to using the `ExplanationRow` components and make Food/Potion/Flask take up less vertical space, but then it looks odd when compared to enchantments and enhancements.

Another thought was to make Food/Potion/Flask a single subsection and use the `SideBySidePanels` component to as a wrapper.

emallson:
Having a single section for consumables is probably good, but i don't think its necessarily a blocker here.

kfinch:
Yes I also think the results should be more compact, and I like the idea of consumables being side-by-side. I still think what you have is nice and didn't want to block it.

kfinch (approved):

PR #5408, opened 2022-11-21
convert StatisticBox and LazyLoad_ to TS

ToppleTheNun:
lazyloadstatisticbox-to-ts


src/analysis/retail/monk/mistweaver/modules/spells/Vivify.tsx
```diff
@@ -163,7 +163,7 @@ class Vivify extends Analyzer {
   statistic() {
     return (
       <StatisticBox
-        postion={STATISTIC_ORDER.CORE(20)}
+        position={STATISTIC_ORDER.CORE(20)}
```
emallson, line 166: i wonder how long that's been like that

emallson (approved):

PR #5409, opened 2022-11-21
migrate away from legacy context API where able

ToppleTheNun:
the legacy context API is deprecated and
prevents the conversion of some components to
functional components. the use of the hooks
context API is easier to rationalize, especially
for those newer to the codebase.


emallson (approved):
👍

you mention "where able". were there cases you couldn't convert? or only ones that you didn't convert? (or neither?)

ToppleTheNun:
There were a few other spots using legacy context (like Results) that I didn't want to migrate in this same PR as they were a bit more involved; I'll put up that work as a separate PR.

PR #5413, opened 2022-11-21
reduce usage of render props in Home page

ToppleTheNun:
homepage-no-render-props


emallson (approved):
this looks fine to me. am i correct in understanding this as a precursor to some of your work on react-router 6?

ToppleTheNun:
Yep! Intended mostly to help ease the transition as react-router v6 doesn't support render props.

PR #5415, opened 2022-11-22
Holy priest bottom row talents

foobartles:
Added support for the rest of the bottom row holy priest talents that need modules, this includes Miracle worker, Pontifex, and Rapid Recovery. Screenshots of their modules are included below. 
image
image
image


foobartles:
closed bc need to fix issue with rapid recovery implementation

PR #5420, opened 2022-11-23
[Preservation] Remove useless stats and cleanup cast link file

trevorm4:
- CastLinkNormalizer had some commented out code that shouldn't be there
- Living Flame, Dream Breath, and Spirit Bloom modules all provide statistics that can be gathered from looking at the healing tab in WCL, so there isn't much point in having them as modules


vohrr:
Living flame is a shared module with devastation so it might be better to unload it in combatlogparser in case they want to use it (I have 0 clue about anything devastation)

trevorm4:
> Living flame is a shared module with devastation so it might be better to unload it in combatlogparser in case they want to use it (I have 0 clue about anything devastation)

It's still just a dps/hps statistic for devastation, which is also useless imo. If they add talents that modify living flame in some way, then adding those as stats is good imo.

ToppleTheNun (approved):

PR #5421, opened 2022-11-23
[Mistweaver] fix crash in HotAttributor

trevorm4:
We forgot to check if the hot was on a player and then indexed into hottracker leading to bad things.

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/core/HotAttributor.ts
+++ b/src/analysis/retail/monk/mistweaver/modules/core/HotAttributor.ts
@@ -67,10 +67,9 @@ class HotAttributor extends Analyzer {
   }
 
   onApplyRem(event: ApplyBuffEvent | RefreshBuffEvent) {
-    if (
-      !this.hotTracker.hots[event.targetID] ||
-      !this.hotTracker.hots[event.targetID][SPELLS.RENEWING_MIST_HEAL.id]
-    ) {
+    const targetID = event.targetID;
+    const spellID = event.ability.guid;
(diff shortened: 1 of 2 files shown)


src/analysis/retail/monk/mistweaver/modules/core/HotAttributor.ts
```diff
@@ -95,16 +102,18 @@ class HotAttributor extends Analyzer {
...
-              event.timestamp + Number(this.hotTracker.hotInfo[event.ability.guid].procDuration),
+              event.timestamp +
+                Number(this.hotTracker.hotInfo[SPELLS.RENEWING_MIST_HEAL.id].procDuration),
               3,
             ),
         );
       this.hotTracker.addAttributionFromApply(this.rapidDiffusionAttrib, event);
-      this.hotTracker.hots[event.targetID][event.ability.guid].maxDuration = Number(
-        this.hotTracker.hotInfo[event.ability.guid].procDuration,
+      this.hotTracker.hots[event.targetID][SPELLS.RENEWING_MIST_HEAL.id].maxDuration = Number(
```
vohrr: Redundant, the event listener is only looking for rem id

vohrr (approved):

src/analysis/retail/monk/mistweaver/modules/core/HotAttributor.ts
trevorm4: Ya they're equivalent, I just think the code looks cleaner with hardcoding REM id lol. I can change it back if you want tho

PR #5423, opened 2022-11-23
[Preservation] Field of dreams module

trevorm4:
Add a module for Field of Dreams, which has a 30% chance to spawn another Emerald Blossom per Emerald Blossom creation (meaning from a hardcast or from a field of dreams proc i.e. a chain reaction).

This means that the probability of 1 Field of Dreams proc from a EB hardcast is .3 and probability of 2 procs is .3^2 = .09

image

created using https://www.warcraftlogs.com/reports/dzcBknW9xb8TKMLX#fight=20&type=healing&source=66

As you can see, they have 20 emerald blossom casts which means 20 * .3 = 6, which is what the module shows. This specific log doesn't seem to have a chain reaction, but the logic should cover both.


src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -217,52 +216,21 @@ const EVENT_LINKS: EventLink[] = [
...
-    forwardBufferMs: EB_BUFFER_MS,
-    isActive(c) {
-      return c.hasTalent(TALENTS_EVOKER.FLUTTERING_SEEDLINGS_TALENT);
-    },
+    backwardBufferMs: EB_BUFFER_MS + 500,
+    maximumLinks: 1,
     additionalCondition(linkingEvent, referencedEvent) {
-      return !HasRelatedEvent(referencedEvent, FLUTTERING_SEEDLINGS_ECHO);
+      const diff = EB_BUFFER_MS - (linkingEvent.timestamp - referencedEvent.timestamp);
+      return Math.abs(diff) < 150;
```
vohrr: Make the 150 a const if its a non-arbitrary value

vohrr (approved):
Just the small suggestion and good to go

PR #5430, opened 2022-11-23
Feature/disc dragonflight/twilight equilibrium

oisinod:


src/analysis/retail/priest/discipline/modules/spells/TwilightEquilibrium.tsx
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  onDamage(event: DamageEvent) {
+    if (
+      !this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_HOLY_BUFF.id) &&
+      !this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_SHADOW_BUFF.id)
+    ) {
+      return;
+    }
```
kfinch: What's the purpose of this early return, when all your other logic is wrapped in pretty much the same check?

src/analysis/retail/priest/discipline/modules/spells/TwilightEquilibrium.tsx
```diff
@@ -0,0 +1,106 @@
...
+      if (this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_HOLY_BUFF.id)) {
+        this.damage += calculateEffectiveDamage(event, TE_INCREASE);
+      }
+    }
+
+    if (SHADOW_DAMAGE_SPELLS.includes(event?.ability.guid ?? -1)) {
+      if (this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_SHADOW_BUFF.id)) {
+        this.damage += calculateEffectiveDamage(event, TE_INCREASE);
+      }
+    }
```
kfinch: Why are these nested ifs rather than an if with an `&&`

src/analysis/retail/priest/discipline/constants.ts
```diff
@@ -65,6 +65,26 @@ export const ATONEMENT_SOURCE_FILTER: SpellFilter = [
...
+  TALENTS_PRIEST.LIGHTS_WRATH_TALENT.id,
+];
+// Spells which are consumed by the shadow buff of Twilight Equilibrium
+export const SHADOW_DAMAGE_SPELLS = [
+  SPELLS.SHADOW_WORD_PAIN.id,
+  SPELLS.MIND_BLAST.id,
+  TALENTS_PRIEST.MINDGAMES_TALENT.id,
+  TALENTS_PRIEST.SHADOW_WORD_DEATH_TALENT.id,
+  TALENTS_PRIEST.SCHISM_TALENT.id,
+];
```
kfinch, line 86: Does TE buff DoTs for their full duration? If so, your current handling will miss most/all of the boosted periodic damage (and probably will also false positive when you have the buff while periodic sources are ticking?)

oisinod: Yup, that's right and good catch! 

I'll amend this probably tomorrow. PTW is permanently buffed once it's procced once by TE. 

I think once the cast triggers it, I will keep a boolean like buffedByTe on dot ticks. Luckily extensions and duplications of the dot all have the same behaviour, so I can just remove the boolean if the dot falls off the main target. What do you think @kfinch

kfinch: I've dealt with something similar in a few of my modules - see the handling of which targets have BT Rip in `Bloodtalons.tsx` - tracked with a set indexed by encodedTargetId called `targetsWithBtRip`. A similar handler should work with you, but triggering only on Apply and not Refresh.

src/analysis/retail/priest/discipline/modules/spells/TwilightEquilibrium.tsx
```diff
@@ -85,22 +111,28 @@ class TwilightEquilibrium extends Analyzer {
     }
   }
 
+  onPTWApply(event: CastEvent | RefreshDebuffEvent | ApplyDebuffEvent) {
+    const targetString = encodeEventTargetString(event) || '';
+    if (this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_HOLY_BUFF.id)) {
+      this.ptwTargets.add(targetString);
+    } else {
+      this.ptwTargets.delete(targetString);
+    }
```
kfinch, line 107: You implied earlier that the TE boost stays for the duration of the DoT, even through refreshes. If so, you need special case handling here that will not delete from the Set if the event is a refresh.

That would be odd behavior though - the boost really can't be downgraded?

src/analysis/retail/priest/discipline/modules/spells/TwilightEquilibrium.tsx
```diff
@@ -0,0 +1,135 @@
...
+    ) {
+      this.damage += calculateEffectiveDamage(event, TE_INCREASE);
+    }
+  }
+
+  onPTWApply(event: CastEvent | RefreshDebuffEvent | ApplyDebuffEvent) {
+    const targetString = encodeEventTargetString(event) || '';
+    if (this.selectedCombatant.hasBuff(SPELLS.TWILIGHT_EQUILIBRIUM_HOLY_BUFF.id)) {
+      this.ptwTargets.add(targetString);
+    } else if (event.type === EventType.ApplyDebuff) {
```
oisinod: @kfinch amended :)

kfinch: Since this behavior is so unusual, can you also leave a comment explaining why? Otherwise I think this PR is good to go

kfinch (approved):

PR #5431, opened 2022-11-24
Update all talent trees (except Paladin) and add helper for counting repeated talents

emallson:
- Adds spellpower.csv to repository for generating talent data. We can update it periodically using wow.export
- Updates all talents (except Paladins, see below)
- Fixes now-invalid references from deduped talents like Stormkeeper
- Adds `getRepeatedTalentCount` to count the number of times a talent has been taken (used for charge counting on spells like Stormkeeper or Empower Rune Weapon)

## Paladins

Buckle up, we're going for a ride.

Until recently, every talent had 3 associated IDs: `spellId` (the thing you cast), `nodeId` (the box in which the talent is placed), and `entryId` (the choice within the talent box---only needed for choice nodes).

That changed in early November with a UI update that lets you drag certain "combining" spells from your talent tree to the action bar by using *any* of the combining ones---including passives. Currently, exports for passives combining with actives that are within choice nodes appear to export incorrectly from raidbots and wowhead.

The reason for this? There is now a 4th ID, a "`internalSpellId`" that is only used for combining talents. It appears to be correctly read (or unneeded?) for non-choice nodes, but not choice nodes.

Paladin is the only class that has combining passives within choice nodes, and thus the only one that is broken. We can't update paladin talents until we work around this or it gets fixed upstream.

## Future Options

One option that we have is to switch to using `entryId` (which is globally unique) for distinguishing talents. Two problems:

1. That breaks all of our deduping
2. The spell name would still be wrong (e.g. instead of `SANCTIFIED_WRATH` we'd get `AVENGING_WRATH`)

Another option would be to update paladin talents, then apply a patch to add the missing talents in. There are only 2 or 3, so it would be fairly minimal maintenance.

The final option would be to do nothing and hope upstream folks fix it. This is the worst option, since it means that paladin talents are stuck in their present state---potentially indefinitely.

I think we probably end up manually adding the missing ones via a patch or import


scripts/talents/generate-talents.ts
```diff
@@ -40,7 +40,7 @@ const classes: { [classId: number]: { name: string; baseMaxResource: number } }
 
 async function generateTalents() {
   const talents: ITalentTree[] = await readJsonFromUrl(TALENT_DATA_URL);
-  const spellpower: ISpellpower[] = csvToObject(await readCsvFromUrl(SPELLPOWER_DATA_URL));
+  const spellpower: ISpellpower[] = csvToObject(await readCsvFromFile(SPELLPOWER_DATA_FILE));
```
ToppleTheNun, line 43: Since `readCsvFromFile` uses `readFileSync`, you don't need the `await` here, though it's not breaking anything.

ToppleTheNun (approved):

Pewtro (approved):

PR #5442, opened 2022-11-28
[Mistweaver] Updated Celestial Modules

vohrr:
Updated the Invoke Chi-Ji and Invoke Yu'lon Modules to include Enveloping Breath in the title, now that Enveloping Breath is its own talent in Dragonflight. Also updated formatting and tooltips to use spell links for consistency with newer modules
**After:**

- Yulon

image

- ChiJi

image


trevorm4 (approved):
lgtm

PR #5449, opened 2022-11-30
add Vault to ZONES and remove SL raids

ToppleTheNun:


src/game/ZONES.ts
```diff
@@ -33,164 +33,29 @@ interface Zone {
...
+      max: 424,
       bucket: 3,
       type: 'Item Level',
     },
     partitions: [
       {
-        name: '9.2',
-        compact: '9.2',
+        name: '10.0',
+        compact: '10.0',
```
Sharrq, line 58: Looking at castle nathria, we might need to include 10.0.5 as well to the partitions? Not positive but worth checking

emallson: The partitions should match what WCL has. We'll likely add a 10.0.5 partition due to class changes, but there's not a need to pre-emptively add them.

Sharrq (approved):
Fix the conflicts and then it should be good then

PR #5451, opened 2022-11-30
add support for Draconic Augment Rune

ToppleTheNun:


src/common/ITEMS/dragonflight/others.ts
```diff
@@ -0,0 +1,11 @@
+import { itemIndexableList } from 'common/ITEMS/Item';
+
+const potions = itemIndexableList({
```
Sharrq: Forgot to rename this from potions

ToppleTheNun: Good catch, thanks!

Sharrq (approved):
fix the conflicts and then it should be good.

PR #5458, opened 2022-12-03
[Mistweaver] Add average CDR to jade bond module

trevorm4:
image

Cleanup jade bond file a bit and add average CDR to statistic and move tooltip to the average cdr element

Also cleaned up changelog file formatting because it was bothering me


vohrr (approved):
lgtm

PR #5459, opened 2022-12-03
[Evoker] Fix empty essence burst chart

trevorm4:
If a player talented into essence burst, but never gained the buff, then the chart would be empty

Before:
image
After:
image

### DIFF
--- a/src/analysis/retail/evoker/preservation/modules/talents/EssenceBurst.tsx
+++ b/src/analysis/retail/evoker/preservation/modules/talents/EssenceBurst.tsx
@@ -112,7 +112,7 @@ class EssenceBurst extends Analyzer {
 
   statistic() {
     const donutChart = this.renderDonutChart();
-    return donutChart ? (
+    return (
       <Statistic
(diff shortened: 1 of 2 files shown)


src/analysis/retail/evoker/preservation/modules/talents/EssenceBurst.tsx
```diff
@@ -79,7 +79,7 @@ class EssenceBurst extends Analyzer {
     ].filter((item) => {
       return item.value > 0;
     });
-    return <DonutChart items={items} />;
+    return items.length > 0 ? <DonutChart items={items} /> : null;
```
vohrr, line 82: Small nit - I would rather see some indication that the talent is selected but no buffs were gained than just hiding the module entirely. Approving

vohrr (approved):

src/analysis/retail/evoker/preservation/modules/talents/EssenceBurst.tsx
trevorm4: image
New statistic, gonna merge

PR #5460, opened 2022-12-04
remove Shadowlands dungeons and raids

ToppleTheNun:


emallson (approved):
Code change looks fine. I couldn't 100% remember what this impacts, so I poked around a bit on my local and couldn't find breakage so I think we're good to merge :+1:

PR #5464, opened 2022-12-04
Add Classic Arcane Mage

jazminite:
First draft of guide for Classic Arcane Mage.

Arcane Missiles (AM) has no start cast or channeled spell event (only ticks), so a Normalizer was necessary.

## Overview
image

## Statistics
image

## Cooldowns
image


src/parser/core/CASTS_THAT_ARENT_CASTS.ts
```diff
@@ -72,6 +73,11 @@ const spells: number[] = [
   SPELLS.CARVED_IVORY_KEEPSAKE.id,
   SPELLS.ANIMA_FIELD_EMITTER_BUFF.id,
   //endregion
+
+  // CLASSIC
+  // region Mage
+  CLASSIC_SPELLS.MISSILE_BARRAGE.id,
```
Pewtro: I think we'll need a classic variation of this, so you don't accidentally set a retail spellid to not be a cast.

src/analysis/classic/mage/arcane/modules/features/Abilities.tsx
```diff
@@ -7,53 +7,24 @@ class Abilities extends CoreAbilities {
...
-      {
-        spell: [SPELLS.FROSTFIRE_BOLT.id, ...SPELLS.FROSTFIRE_BOLT.lowRanks],
-        category: SPELL_CATEGORY.ROTATIONAL,
-        gcd: { base: 1500 },
-      },
-      {
-        spell: [SPELLS.FIREBALL.id, ...SPELLS.FIREBALL.lowRanks],
-        category: SPELL_CATEGORY.ROTATIONAL,
-        gcd: { base: 1500 },
+        gcd: { base: 500, minimum: 1 }, // Tick gcd with Missile Barrage proc
```
Pewtro: I think this is a bad idea to do it this way, since it's misleading... You should keep the GCD as the GCD, and then normalize the events so that each tick isn't a cast

src/parser/core/CASTS_THAT_ARENT_CASTS.ts
jazminite: IDs don't clash between retail and classic. There are many files we use retail and classic IDs in.

src/analysis/classic/mage/arcane/modules/features/Abilities.tsx
jazminite: This is the only way I could remove most of the errors. I tried to normalize it with events but there's no trigger event.

src/parser/core/CASTS_THAT_ARENT_CASTS.ts
Pewtro: There are plenty of spell id overlaps in general between classic and retail where a given spell works fundamentally different, a split of CASTS_THAT_ARENT_CASTS would make such a possible clash, not possible.

src/analysis/classic/mage/arcane/modules/features/Abilities.tsx
Pewtro: What do you mean "trigger event"? The first cast event in a series of cast events in a short timespan should be enough, then you'll also know max amounts of ticks, so you also have information necessary to split chains of casts if that's happening in a log.

jazminite: Added normalizer

src/analysis/classic/mage/arcane/normalizers/ArcaneMissiles.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const beginChannel: ArcaneMissilesBeginChannelEvent = {
+    ...event,
+    type: EventType.BeginChannel,
+    isCancelled: true,
+    trigger: event,
+    __fabricated: true,
+    arcaneMissilesCastEvents: [],
+    arcaneMissilesDamageEvents: [],
+  };
+  event.ability.guid = SPELLS.ARCANE_MISSILES_CHANNELED.id;
```
Pewtro: This doesn't look to do anything

Pewtro (commented):
Looks good with the normalizer! :)

src/analysis/classic/mage/arcane/normalizers/ArcaneMissiles.tsx
jazminite: It changes the spell id of the 1st AM tick to the channeled spell id. This makes it possible to:
- Accurately count the Arcane Missile casts on the statistics page
- Display the channeled cast on the timeline
- Appropriately add the GCD for the cast which decreases downtime % displayed on the checklist

Pewtro: No what I mean the event.ability.guid is the last line in this particular function, so reassigning it to SPELLS.ARCANE_MISSILES_CHANNELED.id seems to not have any impact?

Pewtro: Were you instead intending to change the id before creating the beginChannel const, so that it would be inserted into the beginChannel const and returned?

jazminite: On line 40, all the event info is added from the Arcane Missiles tick cast. On line 48, the id is updated to Arcane Missiles channeled cast. It's necessary to change it after the other info is gathered.

Pewtro: But _why_ are you changing it at all? It's never used afterwards, you're immediately returning the beginChannel object and never doing anymore with the passed in event.

jazminite: It modifies the event which is used in other places. For example:

Without the change:
image

With the change:
image

It does seem a little odd at the bottom, so I moved it to the top.

Pewtro: Moving it to the top causes it to go into the fabricated event, which makes sense why it would have an impact further on since that is returned from the function, the previous thing shouldn't have had any impact since it was unused

jazminite: It did work, screenshots are from the id change at the bottom but I moved it to the top just in case.

jazminite: @Pewtro - Everything good here to merge?

Pewtro: Yeah - I added some further description in discord from when I pinged you

PR #5466, opened 2022-12-05
[Preservation] Add essence burst suggestion if player doesnt consume many buffs

trevorm4:
image

Adds a suggestion if player didn't gain any Essence Burst buffs during the encounter (which means they should be casting living flame more)

Also added a numCasts to Living Flame for future plans


vohrr (approved):
lgtm

PR #5471, opened 2022-12-05
[Preservation] Add Cycle of Life module

trevorm4:
Adds a module to show chart that attributes what percent of cycle of life healing is from what spell

image

It shows top 5 spells and groups all other healing events in "Other" to not bloat the stat


src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+class CycleOfLife extends Analyzer {
+  savedBySpell: Map<number, number> = new Map<number, number>();
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS_EVOKER.CYCLE_OF_LIFE_TALENT);
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.handleHeal);
+  }
+
+  handleHeal(event: HealEvent) {
```
vohrr, line 53: We need to handle multiple sprouts being out at once since the heals in those moments are double/triple dipping. It looks like it shows up in the log as a buff stack (id=388973) so shouldn't be too bad to do

src/parser/ui/DonutChart.tsx
```diff
@@ -7,7 +7,7 @@ import { VisualizationSpec } from 'react-vega';
 
 import 'parser/ui/DonutChart.scss';
 
-type Item = {
+export type Item = {
```
vohrr, line 10: Necessary?

vohrr (changes requested):

src/parser/ui/DonutChart.tsx
trevorm4: Yea I am taking the list of donutchart items and filtering them out in a function, I could declare it locally as well but I figure i would add it here

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
vohrr: Something funky in the log, but it looks like 1 buff stack is default; 
https://www.warcraftlogs.com/reports/Jm1PGdVr2jW6zfTq#fight=14&type=auras&source=48&ability=388973&view=events
https://www.warcraftlogs.com/reports/Jm1PGdVr2jW6zfTq#fight=14&type=auras&source=323&ability=388973&view=events
Need to account for it

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
```diff
@@ -0,0 +1,223 @@
...
+        label: 'Cycle of Life',
+        spellId: TALENTS_EVOKER.CYCLE_OF_LIFE_TALENT.id,
+        value: this.totalSavedBySpell(SPELLS.CYCLE_OF_LIFE_HEAL.id),
+        valueTooltip: formatNumber(this.totalSavedBySpell(SPELLS.CYCLE_OF_LIFE_HEAL.id)),
+      },
+      {
+        color: SPELL_COLORS.DREAM_BREATH,
+        label: 'Dream Breath',
+        spellId: TALENTS_EVOKER.DREAM_BREATH_TALENT.id,
+        value: this.totalSavedBySpell(SPELLS.DREAM_BREATH_ECHO.id + SPELLS.DREAM_BREATH.id),
```
vohrr: `this.totalSavedBySpell(SPELLS.DREAM_BREATH_ECHO.id) + this.totalSavedBySpell(SPELLS.DREAM_BREATH_ECHO.id)`

vohrr (changes requested):

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
trevorm4: Oops ya

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
trevorm4: Changed code to account for multiple seeds

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
```diff
@@ -30,30 +35,40 @@ const TRACKED_SPELL_IDS = [
...
     const amount = event.amount + (event.absorbed || 0) + (event.overheal || 0);
-    const percentSaved = this.selectedCombatant.hasBuff(SPELLS.CYCLE_OF_LIFE_BUFF.id)
-      ? CYCLE_OF_LIFE_PERCENT_SAVED
-      : CYCLE_OF_LIFE_PERCENT_SAVED *
-        this.selectedCombatant.getBuffStacks(SPELLS.CYCLE_OF_LIFE_STACK.id);
+    const percentSaved = CYCLE_OF_LIFE_PERCENT_SAVED * this.getNumActiveSeeds(event.timestamp);
     this.savedBySpell.set(event.ability.guid, curVal! + amount * percentSaved);
   }
 
+  getNumActiveSeeds(timestamp: number): number {
```
vohrr, line 62: This is probably fine but why not use the cycle's heal event as your trigger to prune expired seeds?

vohrr (approved):

src/analysis/retail/evoker/preservation/modules/talents/CycleOfLife.tsx
trevorm4: You can do that but it's gonna be less efficient because you'll have another event listener. Also, you'll have to make sure you only prune it once for each cycle of life group and then it just adds unneeded complexity imo.

PR #5472, opened 2022-12-05
[Preservation] Add checklist and dream flight module

trevorm4:
Added module to check what percent of raid is being hit by dream flight and also added checklist entries for spirit bloom, reversion, dream breath, and dream flight

image
image

Open to input on the thresholds as I kinda made them up but they seem reasonable


vohrr (approved):
lgtm

PR #5486, opened 2022-12-07
[Preservation] add checklist for Echo expirations

trevorm4:
Adding a checklist item to make sure that player is not letting echo buffs expire (i.e. not consuming them)

image
image


src/analysis/retail/evoker/preservation/modules/talents/Echo.tsx
```diff
@@ -67,6 +97,36 @@ class Echo extends Analyzer {
     return isFromTAEcho(event);
   }
 
+  get suggestionThresholds() {
+    return {
+      actual: this.percentWasted,
```
vohrr: No reason to use percentage here imo. Flat number of buffs wasted makes more sense to look at and is easier to digest for a user.

vohrr (changes requested):
Just the small suggestion on format and looks good

src/analysis/retail/evoker/preservation/modules/talents/Echo.tsx
trevorm4: I just did percentages because defining thresholds is a bit tricky based on fight length. I can change it tho and just do 1,3,5 if you think thats better.

trevorm4:
Updated to use raw numbers

image
image

vohrr (approved):

PR #5487, opened 2022-12-07
[Preservation] add dream breath targets hit suggestion and fix empowered casts

trevorm4:
Empowered cast spellid's were wrong, so I fixed them

Before: 
image
After:
image

I added a checklist item and suggestion for dream breath targets hit 
image
image

Had to link Dream Breath buff applies to cast's because stasis dream breath applies don't have a corresponding cast, so we don't want to include it in this metric. If we want to include it in this metric, then I can change to use apply grouping linking that we had setup for EB.

### DIFF
--- a/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
+++ b/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
@@ -27,6 +27,7 @@ export const DREAM_BREATH_CALL_OF_YSERA_HOT = 'DreamBreathCallOfYseraHoT'; // li
 export const LIVING_FLAME_CALL_OF_YSERA = 'LivingFlameCallOfYsera'; // link buffed living flame to buff removal
 export const FIELD_OF_DREAMS_PROC = 'FromFieldOfDreams'; // link EB heal to fluttering heal
 export const HEAL_GROUPING = 'HealGrouping'; // link EB healevents and TA pulses together to easily fetch groups of heals/absorbs
+export const BUFF_GROUPING = 'BuffGrouping'; // link ApplyBuff events together
 export const SHIELD_FROM_TA_CAST = 'ShieldFromTACast';
 
 const CAST_BUFFER_MS = 100;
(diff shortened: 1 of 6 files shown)


src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -329,6 +329,16 @@ const EVENT_LINKS: EventLink[] = [
...
     },
   },
+  // link dream breath hot application to cast
+  {
+    linkRelation: FROM_HARDCAST,
+    linkingEventId: [SPELLS.DREAM_BREATH.id],
+    linkingEventType: [EventType.ApplyBuff],
+    referencedEventId: [TALENTS_EVOKER.DREAM_BREATH_TALENT.id],
+    referencedEventType: [EventType.Cast],
+    backwardBufferMs: 2500, // there is some pretty big latency in logs, so being generous
```
vohrr: This is not latency, the cast event is timestamped at the start of the empower. ApplyBuff & RemoveBuff events are used to track the length of the empowered cast so it might make more sense to set up links for the empowered cast time to link this up. 
image
Or you could just group the buff applied events like you've done with TA/EB to track targets hit because dream breath echo is its own spellid so you don't have to worry about collision there.
The only concern  with that would be stasis potentially
Edit: Reading is hard apparently, didn't see that you mentioned that in the original post, probably want to go the grouping linking route

vohrr: This appears to be how it works with fire breath as well, but I am not seeing it work with this way for spiritbloom

vohrr (changes requested):

ToppleTheNun (commented):
I just want to add a note that an empowered cast event is coming per emallson, Kihra is working on it.

src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
trevorm4: Yeah ill just change it to grouping, it is weird that its implemented that way.

vohrr (approved):

PR #5489, opened 2022-12-07
[Priest] Fix error with Dark Evangelism Buff

DoxAshe:
If the Dark Evangelism buff exists at the start of the fight it causes an error in PrePullCooldown.  Removing the triggeredBySpellId fixes this issue.


emallson (approved):
i would rather identify/fix the root cause but if this is making the analyzer unusable / breaking major features then this is fine to merge straight away and we can fix the underlying cause later

DoxAshe:
I'm trying to fix the root cause, but I haven't figured it out yet.  It effectively causes the analyzer to break any time you filter a fight with this talent selected and it is a talent shadow is always going to be using.

PR #5493, opened 2022-12-08
add empowered cast log events

ToppleTheNun:
image
image


trevorm4:
lgtm but did we want to move the empowerstart/end to be next to cast success in the list since its technically a cast?

ToppleTheNun:
Yep, I'll do that in the morning

PR #5508, opened 2022-12-14
[MW] change vivacious vivification module to limit scope of refresh counting

trevorm4:
Changed the module to only count a wasted refresh when the amount of rems on the raid is greater than the expected number of rems

Also changed tooltip of Rapid Diffusion to use Icon for rem

Before:
image

After: 
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
@@ -47,7 +47,7 @@ class VivaciousVivification extends Analyzer {
 
   onRefresh(event: RefreshBuffEvent) {
     // every refresh is a wasted buff application and the CD restarts
-    if (this.currentRenewingMists > this.vivify.estimatedAverageReMs) {
+    if (this.currentRenewingMists >= this.vivify.estimatedAverageReMs) {
       this.wastedApplications += 1;
     }
   }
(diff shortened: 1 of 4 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
```diff
@@ -21,20 +27,38 @@ class VivaciousVivification extends Analyzer {
...
+  }
+
+  onRemoveRem(event: RemoveBuffEvent) {
+    this.currentRenewingMists -= 1;
   }
 
   onRefresh(event: RefreshBuffEvent) {
     // every refresh is a wasted buff application and the CD restarts
-    this.wastedApplications += 1;
+    if (this.currentRenewingMists > this.vivify.estimatedAverageReMs) {
```
vohrr: `>=`

src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx
```diff
@@ -21,20 +27,38 @@ class VivaciousVivification extends Analyzer {
...
   }
 
   get suggestionThresholds() {
     return {
       actual: this.wastedApplications,
       isGreaterThan: {
-        minor: 2,
-        average: 5,
-        major: 8,
+        minor: 0,
```
vohrr: I would be ok with a looser threshold here because of how fight dependent something like this is being on a static timer. It would be poor of us to count a wasted application solely because there are a lot of rems out if the raid is full health

vohrr (approved):
Few thoughts,  approving

PR #5527, opened 2022-12-16
add ability to exclusively use the Guide

ToppleTheNun:
Example of how it looks in use:
image

Example of spec not using it:
image


emallson (approved):
slightly concerned about continuing to grow new config props, but can't think of a better way to do it that doesn't involve hidden behaviors that could be potentially nasty surprises.

ToppleTheNun:
It could be put on the `CombatLogParser` instead of on the `Config`?

PR #5530, opened 2022-12-16
re-enable Beast Mastery log parser

ToppleTheNun:


emallson (approved):
assuming you tested and it works. if so, ship it

ToppleTheNun:
This was mostly to fix an issue where Beast Mastery Hunters were still clickable on the player selection screen, but would load forever when clicked.

Pewtro:
SV is the same I think 😢

ToppleTheNun:
I'll do the same thing for Survival as well, just going to go find a log.

PR #5537, opened 2022-12-19
Holy Priest Initial DF Guide

foobartles:
Initial support for holy priest dragon flight guide. Supports lightweaver build and PoH build. Still needs support for proper holy word usage, and divine word usage. The Holy Word modules need some cleanup so going to do that later and get what i have out in the meantime.

Example screenshot of what I have:
image

### DIFF
--- a/src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
+++ b/src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
@@ -1,5 +1,5 @@
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { HealEvent } from 'parser/core/Events';
+import Events, { HealEvent, CastEvent } from 'parser/core/Events';
 import SPELLS from 'common/SPELLS/';
 import TALENTS from 'common/TALENTS/priest';
 import Statistic from 'parser/ui/Statistic';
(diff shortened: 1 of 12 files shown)


src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
```diff
@@ -30,44 +44,154 @@ class Lightweaver extends Analyzer {
...
+  onHealCast(event: CastEvent) {
+    // linked heal event exists
+    const healEvent = getHeal(event);
+    if (healEvent) {
+      this.totalHealCasts += 1;
+
+      if (isCastBuffedByLightweaver(event)) {
+        // calculate effective healing from bonus
+        this.healingDoneFromTalent += calculateEffectiveHealing(healEvent, HEALING_BONUS);
+        this.overhealingDoneFromTalent = calculateOverhealing(healEvent, HEALING_BONUS);
```
emallson, line 74: this looks wrong, unless the stacks of the buff act more like charges (so a cast consumes one stack for 15% bonus healing)

src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
```diff
@@ -30,44 +44,154 @@ class Lightweaver extends Analyzer {
...
-      this.overcappedStacks += 1;
+  onFlashHealCast(event: CastEvent) {
+    // linked heal event exists
+    const healEvent = getHeal(event);
+    if (healEvent) {
+      this.totalFlashHealCasts += 1;
+      if (this.selectedCombatant.getBuffStacks(SPELLS.LIGHTWEAVER_TALENT_BUFF.id) < 2) {
+        if (this.isHighOverheal(healEvent)) {
+          this.highOverhealFlashHealCasts += 1;
+        }
```
emallson, line 95: this seems like it should track high overheal casts regardless of whether it generates a stack, but i guess if you're only showing count you need to make sure you're not double-counting :thinking:

emallson (changes requested):
Got some design comments. Only hard blocker from merging is the Lightweaver healing % thing. Once that is sorted (either by correcting me or the code), then we can merge and you can iterate on the design.

### Design Comments
Untitled-2022-09-28-1727

src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
foobartles: only one stack of lightweaver is consumed per cast of heal, so yes it is more like charges. the live implementation of this statistic is incorrectly calculating this since it is treating it like both charges are consumed for 30%. that is not how it works in game.

src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
foobartles: i didn't do this because I'm considering any flash heal that doesn't generate a stack a bad cast(red bar), regardless of its overheal. the yellow bar is for high overheal casts that generate a stack if that makes sense

foobartles:
> 

Thanks for this design feedback, it is really great! I will look at implementing this now. As far as the lightweaver comment, I looked at my code again and am confident it is reflecting how the talent works in game.

foobartles:
image
updated design for PoH

emallson (approved):

PR #5539, opened 2022-12-19
add new Defensives section behind opt-in

ToppleTheNun:
allow people to see the new mitigation section
after opt-in in order to allow feedback.


src/analysis/retail/demonhunter/vengeance/modules/core/MajorDefensives/core.tsx
```diff
@@ -0,0 +1,389 @@
+import styled from '@emotion/styled';
```
emallson, line 1: missed reviewing this but there's a lot of copying here. when possible, if you want to use something from another spec we should get it moved to the `shared` section rather than duplicating the code

ToppleTheNun: I'll move the non-changed pieces over to a shared section, I was a bit worried I might have to make some changes due to VDH weirdness and didn't want to affect Brew.

PR #5542, opened 2022-12-19
move defensive section components to guide core

ToppleTheNun:
usage in DH:
image

I tried to make it generic to make it easier to adopt; there is very basic support for debuff-based Major Defensives (VDH needs this) but it needs more attention than I can give to it atm.

I did not migrate monk over to using this as I'm less familiar with monk and didn't want to break anything by accident. VDH is working fine from my testing.


emallson:
Let me hack on this and get monk ported over before we merge 👍

ToppleTheNun:
DH is kinda weird in general and it makes the buff bar on the timeline almost unusable, not sure how to mitigate it:
image

Demon Spikes has a hasted cooldown, 2 charges, a base duration of 8s, a talent that extends its duration by up to 2s, and another talent that reduces the cooldown on it by 0.5s every time you consume a Soul Fragment.

I ended up just adding a change to allow overriding the buffer MS that gets used and setting it to an absurdly large number (45s) to try and catch it so that it's not always just taking up the full buff bar:
image

emallson:
So tbh I wouldn't include DS on this timeline at all. I'd use something separate for DS (like brew does for PB, which is kinda-sorta equivalent).

I included CB on brews, which occupies a role closer to FB for DH---its a 30ish second cooldown (depending on melee uptime) that you don't push rotationally (often you hold it for major damage spikes), vs PB or DS which are more of "if you're tanking anything, you should use this"

ToppleTheNun:
That's where I landed with Frailty; it's just so omnipresent from just doing your rotation that it's more valuable to track when you're hit without it applied than to track when you're hit with it. I'll check out how PB is implemented in more detail

src/interface/guide/components/MajorDefensives/Timeline.tsx
```diff
@@ -102,47 +113,46 @@ const BuffTooltip = ({
...
+  const analyzers = useAnalyzers(majorDefensiveAnalyzers);
 
   const tooltipData = useCallback(
     (event: ApplyBuffEvent) => {
-      const analyzer = analyzers?.find((analyzer) => analyzer.appliesToEvent(event));
-      const mit = analyzer?.mitigations.find((mit) => mit.start.timestamp === event.timestamp);
+      const analyzer = analyzers
+        ?.filter(isDefined)
+        ?.find((analyzer) =>
+          isBuffBasedMajorDefensive(analyzer) ? analyzer.appliesToEvent(event) : undefined,
```
emallson: working through this now. would *highly* discourage having behavior like this conditional on `instanceof`. Not entirely sure we need the split base class either but haven't fully looked at the debuff differences (seems currently incomplete?)

ToppleTheNun: I'm likely going to re-go through the implementation of this soon, it's been a bit since I looked at it.

emallson: okay, then maybe i hold off on porting monk code to it

ToppleTheNun:
Going back through this, the original reason for splitting buff/debuff was because one really only cares about applying the buff to self and the other really only cares about applying the debuff to others. They could probably be merged into the same class without issue since the two cases should never collide.

ToppleTheNun:
Superseded by #5995.

Issue #5543, opened 2022-12-19
LegEnhancementChecker not accurate


JeaneC:
Looking at https://www.wowanalyzer.com/report/hqF831BjCvLKM4pX/4-Heroic+Terros+-+Kill+(5:27)/Vireve/standard/overview, it claims I'm missing a leg enhancement but I definitely have it.

New here, but looking from the code, I think we just need to change `item.temporaryEnchant` to `item.permanentEnchant` for `LegEnhancementChecker.tsx`. Tested that change locally and seemed to work, but unfamiliar of other changes that might cause. I'm guessing permanent enchant is your standard enchants and temporary refers to oils or weapon runes?

emallson:
this probably got missed because SL armor kits were temporary but DF ones are permanent. the fix seems reasonable, but i can look later

JeaneC:
Happy to take as a first commit! Seems straightforward to me

emallson:
go for it 👍

ToppleTheNun:
Yep, was just a misunderstanding on my part.

PR #5544, opened 2022-12-19
Check for permanent enchants on legs

JeaneC:
Fixing small oversight from discussion in #5543. Check for `item.permanentEnchant` over `item.temporaryEnchant`


JeaneC:
Oh new to this, I also have to add a changelog entry?

ToppleTheNun:
Yep! You can add yourself as a contributor in [src/CONTRIBUTORS.ts](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/CONTRIBUTORS.ts) and then reference it in [src/CHANGELOG.tsx](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/CHANGELOG.tsx).

ToppleTheNun (approved):

PR #5546, opened 2022-12-20
[Talents] Import talentId (id) from Raidbots, and add helpers to differentiate between certain talents.

Pewtro:
Talents such as Berserk: X for Druids and X Wrath for Paladins all impact the same spellid, so the talents have the same spellid, this means the regular way of identifying whether a player has the talent won't work. Instead we can look for the talentId (simply called id in CombatantInfo from WCL), and use that to determine whether a player has the given talent.

We could roll this out to be the standard way of doing hasTalent() but that goes a bit beyond the initial scope of this PR, unless it's really desired - if so I'll do that next.

Edit: After thinking about it a bit more, maybe we just flatout don't allow talents to be checked by numbers, and force them to use a talent object to ensure we accurately check. If people agree, I'll add that refactor to a follow-up PR shortly after this one. If we allow numbers, they might be the wrong ones because people are used to passing in spellids, and those can be inaccurate on talents.


emallson (approved):
looks good

PR #5547, opened 2022-12-20
add opt-in toggle for hiding rotation explanations

ToppleTheNun:
I had a bit of extra time after raid so I took a swing at the switch to hide explanations. I wanted to allow adoption to be gradual instead of affecting everything, so the current example in VDH is using it as a session-storage-based state value per section.

An alternative to doing it this way and tracking the opt-in/opt-out across multiple sections would be to store it in the Guide context, which shouldn't take very much additional effort.

hide-explanation-toggle

### DIFF
--- a/src/analysis/retail/demonhunter/vengeance/guide/useVdhFeatureFlag.ts
+++ b/src/analysis/retail/demonhunter/vengeance/guide/useVdhFeatureFlag.ts
@@ -1,17 +1,10 @@
-import { useState } from 'react';
+import useSessionFeatureFlag from 'interface/useSessionFeatureFlag';
 
-const useVdhFeatureFlag = (featureFlag: string = '', featureFlagDefault?: boolean) => {
-  const sessionFeatureFlagSetting = window.sessionStorage?.getItem(
-    // This will filter out if featureFlag is an empty string
-    ['wowa-vdh-ff', featureFlag].filter((it) => it).join('-'),
+const useVdhFeatureFlag = (featureFlag: string, featureFlagDefault: boolean = false) => {
+  return useSessionFeatureFlag(
+    ['vdh', featureFlag].filter((it) => it).join('-'),
+    featureFlagDefault,
   );
-  const defaultFeatureFlagSetting = Boolean(featureFlagDefault);
-  const initialFeatureFlagSetting =
-    sessionFeatureFlagSetting === null
-      ? defaultFeatureFlagSetting
-      : Boolean(sessionFeatureFlagSetting);
-
-  return useState(initialFeatureFlagSetting);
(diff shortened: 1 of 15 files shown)


kfinch (approved):
Thanks for carrying me Topple

emallson:
My thinking was to put this in `Explanation` and via a context and have `Explanation` itself return null when the toggle is set. I've started using `Explanation` for other bits of text to toggle as well.

However, this looks good and covers one of the major cases where this is used.

src/analysis/retail/demonhunter/vengeance/guide/useVdhFeatureFlag.ts
```diff
@@ -1,17 +1,10 @@
-import { useState } from 'react';
+import useSessionFeatureFlag from 'interface/useSessionFeatureFlag';
 
-const useVdhFeatureFlag = (featureFlag: string = '', featureFlagDefault?: boolean) => {
-  const sessionFeatureFlagSetting = window.sessionStorage?.getItem(
-    // This will filter out if featureFlag is an empty string
-    ['wowa-vdh-ff', featureFlag].filter((it) => it).join('-'),
+const useVdhFeatureFlag = (featureFlag: string, featureFlagDefault: boolean = false) => {
```
kfinch, line 3: Does 'vdh' stand for Vengeance Demon Hunter? Is this supposed to be spec specific?

kfinch (commented):
Use appears pretty straightforward - but I am wondering if there is an easier way to pass the hideExplanation flag. Is it possible for the flag to in a state where we can get it the same way we currently `useInfo()` ?

ToppleTheNun:
We could definitely put it in either the GuideContext or a separate context that can be consumed in each explanation via a hook. I'd prefer to do it that way so that we don't have 3 different "hide explanation" toggles on the same Guide, but I wanted to get a first attempt out the door.

src/analysis/retail/demonhunter/vengeance/guide/useVdhFeatureFlag.ts
ToppleTheNun: This one was intended to be a wrapper around the usual "useSessionFeatureFlag" hook so that I could have VDH specific flags. You could use "useSessionFeatureFlag" without creating a wrapper hook

ToppleTheNun:
V2 should be available for review now, here's the new preview of it:
hide-explanation-toggle

kfinch:
Ahh this looks very nice - pretty much how I envisioned the feature working

kfinch (approved):

emallson (approved):

PR #5549, opened 2022-12-20
ignore duplicate spell IDs for talents

ToppleTheNun:


src/CHANGELOG.tsx
```diff
@@ -17,6 +17,7 @@ import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
+  change(date(2022, 12, 20), 'Ignore duplicate spell IDs for talents during development.', ToppleTheNun),
```
emallson, line 20: to confirm: the check for duplicates is skipped in prod, correct?

ToppleTheNun: It's already skipped in prod, yep!
```typescript
    if (process.env.NODE_ENV === 'development') {
      // check if there's already an existing value by the same ID
      // and throw an error if it's not exactly equal
      // the shallow equality check is added to support cases like Warrior's Sudden Death talent
      if (
        indexedByNameAndId[value.id] &&
        !indexedByNameAndId[value.id].__ignoreDuplication &&
        !value.__ignoreDuplication &&
        !shallowEqual(indexedByNameAndId[value.id], value)
      ) {
        throw new Error(`A spell with this ID already exists: ${value.id}, ${String(key)}`);
      }
    }
```

emallson (approved):

PR #5557, opened 2022-12-21
re-order Beacon handling on fight end

ToppleTheNun:
Report tested with: /report/VY38tTJkqLQ7yjnF/26-Mythic+Terros+-+Kill+(6:26)/Shirup/standard/statistics
image

### DIFF
--- a/src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
+++ b/src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
@@ -379,14 +379,14 @@ class BeaconUptime extends Analyzer {
   }
 
   _endOfFight(event: FightEndEvent) {
-    if (this.hasBoL) {
-      this._handleBoLEvents(event);
-      return;
-    }
-
     if (this.hasBoF) {
       this._handleBoLEvents(event);
       this._handleBoFEvents(event);
+      return;
+    }
+
+    if (this.hasBoL) {
+      this._handleBoLEvents(event);
       return;
     }
 
--- a/src/analysis/retail/paladin/shared/HolyPowerTracker.ts
+++ b/src/analysis/retail/paladin/shared/HolyPowerTracker.ts
@@ -1,18 +1,20 @@
 import SPELLS from 'common/SPELLS';
 import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import ResourceTracker from 'parser/shared/modules/resources/resourcetracker/ResourceTracker';
+import { Options } from 'parser/core/Analyzer';
+import { CastEvent } from 'parser/core/Events';
 
 class HolyPowerTracker extends ResourceTracker {
-  constructor(...args) {
-    super(...args);
+  constructor(options: Options) {
+    super(options);
     this.resource = RESOURCE_TYPES.HOLY_POWER;
   }
 
-  getAdjustedCost(event) {
-    if (!this.getResource(event).cost) {
+  getAdjustedCost(event: CastEvent) {
+    let cost = this.getResource(event)?.cost ?? 0;
+    if (!cost) {
       return 0;
     }
-    let cost = this.getResource(event).cost;
     if (this.selectedCombatant.hasBuff(SPELLS.FIRES_OF_JUSTICE_BUFF.id)) {
       cost = cost - 1;
     }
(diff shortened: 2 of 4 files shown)


src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
```diff
@@ -379,14 +379,14 @@ class BeaconUptime extends Analyzer {
   }
 
   _endOfFight(event: FightEndEvent) {
-    if (this.hasBoL) {
+    if (this.hasBoF) {
       this._handleBoLEvents(event);
```
Pewtro, line 383: Should this still be handling BoL events?

src/analysis/retail/paladin/shared/HolyPowerTracker.ts
```diff
@@ -1,18 +1,20 @@
...
-    super(...args);
+  constructor(options: Options) {
+    super(options);
     this.resource = RESOURCE_TYPES.HOLY_POWER;
   }
 
-  getAdjustedCost(event) {
-    if (!this.getResource(event).cost) {
+  getAdjustedCost(event: CastEvent) {
+    let cost = this.getResource(event)?.cost ?? 0;
```
Pewtro, line 14: If you do `?? 0` you don't need the !cost since it could never be nullish

ToppleTheNun: The `!cost` is still there so we don't do the cost subtraction in the FIRES_OF_JUSTICE_BUFF check, as it'll still return early due to `0` being falsy. I didn't want to go too deep with the changes here as I'm not maintaining it.

src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
ToppleTheNun: If you have Beacon of Faith, you also have Beacon of Light and should be handling both buff checks as I understand it from running through the code. Beacon of Virtue was weird but wasn't included in the initial report.

src/analysis/retail/paladin/shared/HolyPowerTracker.ts
Pewtro: then the `?? 0` is unnecessary, no?

src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
Pewtro: Alternatively - we could just not return from the two functions and they'd just fallthrough check each other - but this works too I guess

Pewtro (approved):
only stylistic things, might need a pass from actual paladin gamers eventually

mfield:
Root cause is here, this is an old assumption from SL talents where glimmer (i.e. default beacon of light) / beacon of faith / beacon of virtue were on the same row so you could only have one. now faith/virtue are an optional choice node and glimmer is a capstone you can have with any style of beacon.

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ef52354f38c0087c35eb6330810c7f1f0a9dbdfa/src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx#L129

src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
mfield: these `has*` look like they're meant to be mutually exclusive (although you technically also use beacon of light when you use beacon of faith). Since the bodies are so simple i'd probably go with something like if/elif/elif/else error to make that assumption clearer in code

mfield: or refactor the separate `has*` into one `this.beaconStyle` or something with an enum

src/analysis/retail/paladin/shared/HolyPowerTracker.ts
ToppleTheNun: Yep!

src/analysis/retail/paladin/holy/modules/beacons/BeaconUptime.tsx
ToppleTheNun: I agree that it should be refactored to be a bit more clear, the module itself is a bit arcane. I didn't want to go to that level of refactor when just trying to get the Beacon of Faith uptime to report more accurately. I'm also perfectly fine closing this PR in favor of another PR being made for the refactor, but I'd prefer to not do that myself as I don't play Holy Paladin.

mfield: this PR certainly isn't doing any harm so might as well land it and wait for a refactor in a slightly less broken state :)

PR #5559, opened 2022-12-22
Allow talents to have multiple talent IDs + re-duplicate some edge cases

emallson:
The multiple talent IDs thing lets us handle things like `Avenging Wrath: Might` correctly and with full de-duplication (where it is a spec node on all 3 paladin specs), while the typical case remains the same.

There is also some re-duplication of talents that were previously de-duplicated. The gist of it is that now if there are two talents that can be taken at once, there are two talent objects for them---even if they are otherwise identical.

For example: Stormkeeper is now represented by 3 objects total---one restoration, two elemental (because you could take both elemental talents at once). The names are `STORMKEEPER_RESTORATION_TALENT`, `STORMKEEPER_1_ELEMENTAL_TALENT`, and `STORMKEEPER_2_ELEMENTAL_TALENT`. Rogue is another case that runes into this with Shadowstep, which was split into `SHADOWSTEP_SHARED_TALENT` and `SHADOWSTEP_SPEC_TALENT` (for assa + sub).

We could rewrite `hasTalent` to key off of entry ids to check for individual talent nodes, but that has *not* been done in this PR (so `hasTalent(SHADOWSTEP_SHARED_TALENT)` will return true if you have any talent that provides that spell, not specifically that specific node. Small potential footgun there, but in line with previous behavior).

I ended up getting lost in the state manipulation of the previous version of the talent generator, so I rewrote it (and re-discovered @Pewtro's pain in the process. god bless). it does a very similar thing now, but slightly differently.


Pewtro (approved):
Seems good - end result is similiar so it looks to work, will be a bit harder if I need to update anything in there since I don't know the code (by heart at this point) anymore but oh well.

After christmas, I'll probably rewrite the hasTalent since it's bugging me that there might be incorrect assumptions from contributors that it'll be correct, but that's no hurry here.

PR #5566, opened 2022-12-24
refactor echo linking and add consumption count to echomodule

trevorm4:
add consumption count to echo module and refactor linking to ensure each echo is only linked to 1 event. There was a specific case where if a player overwrote their ta echo with a regular echo, it would essentially count the consumption twice (once for ta and once for hardcast). There was also a case where if you used verdant embrace on a person with echo, it would count both instances of the VE heal as an echo heal

Before:
image

After:
image

### DIFF
--- a/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
+++ b/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
@@ -83,6 +83,9 @@ const EVENT_LINKS: EventLink[] = [
         c.hasTalent(TALENTS_EVOKER.RESONATING_SPHERE_TALENT)
       );
     },
+    additionalCondition(linkedEvent, referencedEvent) {
+      return !HasRelatedEvent(linkedEvent, FROM_HARDCAST);
+    },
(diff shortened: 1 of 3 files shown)


src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -83,6 +83,9 @@ const EVENT_LINKS: EventLink[] = [
         c.hasTalent(TALENTS_EVOKER.RESONATING_SPHERE_TALENT)
       );
     },
+    additionalCondition(linkedEvent, referencedEvent) {
+      return !HasRelatedEvent(linkedEvent, FROM_HARDCAST);
+    },
```
emallson, line 88: is the order of link processing deterministic based on the order in which links are provided? @kfinch

trevorm4: It always links in order, so rules above the current rule will already have been linked when considering the current link.

vohrr (approved):

PR #5567, opened 2022-12-25
Added Healing Rain, Downpour and Wellspring to Unleash Life buff usage breakdown

Fassbrause:
I added 3 new colors in in the resto shaman constants for Healing Rain, Downpour and Wellspring, check whether these seem appropriate to you guys.


trevorm4 (changes requested):
Re-add the .env.local file and can you add an image to your PR to show an example of what it looks like with the new spells?

Otherwise it looks good overall

Fassbrause:
Here's how it looks:
image
My bad, I renamed env.local.example instead of copying it.

Fixed now!

trevorm4 (approved):

PR #5574, opened 2022-12-26
Evoker Devastation Guide (v1)

JeaneC:
Initialize the 1st iteration of Devastation for WoWAnalyzer.

Includes 5 sections
- Core Rotation with an APL
- Cooldowns checking CD efficiency
- Dragon Rage window analysis (Devastation's most important CD)
- Section dedicated to rounded off DPS more
- Energy Efficiency

Tested the module out a bit on the Evoker discord. The haste/gcd settings for some of these need to be fine tuned more (but I'm really new to how to figure that out)

This marks my second PR in the repo! Definitely gonna need feedback because I literally just opened the codebase a few days ago and have no idea what components or structures exists.

----
Side by side comparison with logs from Evoker Discord (wyrmrest)
[Not Good Log](https://user-images.githubusercontent.com/25225547/209577628-3f7969e4-d957-4dec-b327-ec34d690d277.png)
[Top 100](https://user-images.githubusercontent.com/25225547/209577636-3e9d6a1f-895e-47fb-a0b1-27c96f3f0e08.png)


src/analysis/retail/evoker/devastation/CONFIG.tsx
```diff
@@ -1,21 +1,33 @@
...
+  patchCompatibility: '10.0.2',
+  isPartial: false,
   // Explain the status of this spec's analysis here. Try to mention how complete it is, and perhaps show links to places users can learn more.
   // If this spec's analysis does not show a complete picture please mention this in the `<Warning>` component.
-  description: <>Hi! This is just a setup and not meant to be a functional parser.....Yet!</>,
+  description: (
+    <>
+      <p>
+        Hello, and welcome to the Devastation Analyzer! This analyzer is maintained by{' '}
+        <code>Vireve</code> , a former priest whose loving the new class!
```
kfinch: `whose` should be `who's`

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
```diff
@@ -0,0 +1,38 @@
...
+  onApplyBuff(event: ApplyBuffEvent) {
+    this.procs += 1;
+    this.lastProcTime = event.timestamp;
+  }
+
+  onRemoveBuff(event: RemoveBuffEvent) {
+    const durationHeld = event.timestamp - this.lastProcTime;
+    if (durationHeld > BURNOUT_DURATION) {
+      this.expiredProcs += 1;
+    }
```
kfinch: Buffs often don't last their exact duration to the millisecond - this way of detecting expiration is very likely to have false negatives. Unfortunately there is no easy way to do this.

The way I have been handling this is using an `EventLinkNormalizer` to pair buff removal with the proc consuming event - for this proc it would be BuffRemove and BuffStackLost of Burnout linked to the CastSuccess of Living Flame. See an example in `ClearcastingNormalizer`.

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
```diff
@@ -0,0 +1,38 @@
...
+    super(options);
+
+    this.addEventListener(
+      Events.applybuff.by(SELECTED_PLAYER).spell([TALENTS_EVOKER.BURNOUT_TALENT]),
+      this.onApplyBuff,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell([TALENTS_EVOKER.BURNOUT_TALENT]),
+      this.onRemoveBuff,
+    );
```
kfinch: Burnout stacks to two, so you're missing procs on BuffStackGained events, and usages on BuffStackLost.

src/analysis/retail/evoker/devastation/modules/abilities/Disintegrate.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell([DISINTEGRATE].map((spell) => ({ id: spell.id }))),
+      () => {
+        this.totalCasts += 1;
+        if (this.inDragonRageWindow) {
+          this.dragonRageCasts += 1;
+        }
+      },
+    );
```
kfinch: Doing this with lambdas is definitely novel, but I guess there's nothing inherently wrong with it. I'm not sure what that `map` is about when setting the listeners tho.

For readability I think I'd prefer you use method references like all the other analyzers do though.

src/analysis/retail/evoker/devastation/modules/abilities/ShatteringStar.tsx
```diff
@@ -0,0 +1,125 @@
...
+        break;
+      case LIVING_FLAME_DAMAGE.name:
+        currentWindow.fireBreath += damageEvent.amount;
+        break;
+      case AZURE_STRIKE.name:
+        currentWindow.fireBreath += damageEvent.amount;
+        break;
+      default:
+        break;
+    }
```
kfinch: We usually prefer to differentiate spells by guid, not name. Is there a particular reason you're using name here?

src/analysis/retail/evoker/devastation/modules/guide/EssenceGraph/EssenceGraph.tsx
```diff
@@ -0,0 +1,65 @@
...
+      mark: {
+        type: 'line' as const,
+        interpolate: 'step' as const,
+        color: '#33937F' as const,
+      },
+      config: {
+        view: {},
+      },
+    };
+  }
```
kfinch: What's the custom `vegaSpec` for? If it's just for custom color you can use a `lineColor` override, see RunicPowerGraph for an example.

kfinch (commented):
Overall this looks like very great work. I scanned over this with an eye towards project conventions and to see if I could catch any common mistakes, see my comments below.

I don't know anything about how Evoker actually plays so I tried to sanity check by looking at tooltips but someone else will have to evaluate if the Guide succeeds on that dimension. Have you been in contact with spec experts?

emallson:
Hey! this is a great start. I haven't gone through code review yet (looks like sref beat me to it anyway) but I showed this to the devastation evokers in my guild and got some feedback on the content (aside from typos, which need a pass):

> I think the rule on pyre is wrong in the first image, it depends on talents but 3 is too low without volatility or CB talented, it's a DPS loss over disintegrate.  Rule of thumb is without volatility/CB you can pyre at 4 target and not have it be a DPS loss, with them at 3, and you can pyre at as low as 2 target depending on CB stacks.  There is a chart - you can see 3 target pyre without is a 10k loss over just casting a disintegrate on a single target. [referring to the pyre chart on [wowhead](https://www.wowhead.com/guide/classes/evoker/devastation/rotation-cooldowns-pve-dps)]

you might have sourced this rule from simc, but worth double-checking

> I'm sort of on the fence on how useful the disintegrate clipping metric is DURING dragonrage - APL default always clips at 3 ticks during Dragonrage to make sure the extra extensions happen and turning that off with evoker.use_clipping=0 is a DPS loss.   If I saw a metric that you DID NOT clip 85% of the time during DR I'd kind of think you're doing it wrong - if it's going to be there it should be clipping before 3 ticks, not just clipping at all.

sounds like intentionally clipping disintegrate during DR is the play?

> Also no real mention of deep breath, I'm not sure on the exact use cases but probably helpful to track uses and time on cooldown similar to fire breath, ES, Shattering, and Drage

just missing deep breath.

I don't think any of these are blockers for v1 of the guide, but revisiting *specifically* the disintegrate clipping would be good to avoid giving the wrong impression [that you should never clip disintegrate]

 (the AoE rules in the APL are often a bit lax just because of technical limitations, so it is okay if they aren't exactly optimal)

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
JeaneC: Ah this is pretty smart way of guaranteeing accuracy. I did try this though and it wasn't showing any `_linkedEvents`. I think that this doesn't work with damage cast events but seems to work fine for buff events? Also needed to do this in the `EssenceBurst` section, which I now see does do proc consumption https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/analysis/retail/evoker/preservation/modules/talents/EssenceBurst.tsx. It has it for a damage spell as well, but I checked logs for preservation and never saw the disintegrate (damage spell) proc consumed either. I left a console.log() for future debugging

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
JeaneC: Thanks for the call out, added this here and in EssenceBurst

src/analysis/retail/evoker/devastation/modules/abilities/Disintegrate.tsx
JeaneC: The map was from a copy/paste 💦 , thanks for flagging

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
emallson: It does work for damage & cast events, but may not look exactly like the examples you're referencing. Brewmaster has a cast/damage link for Breath of Fire in `analysis/retail/monk/brewmaster/modules/spells/BreathOfFire/normalizer.ts`

JeaneC:
@emallson  feedback helps a lot! I have to rework the AoE APL a bit. There's a bit of logic related to stacks and talents that your guildies pointed out, but I wasn't sure how specific or laxxed I wanted the APL to be.

> sounds like intentionally clipping disintegrate during DR is the play?

It totally is the play, and I actually leave a note mentioning that, but the feedback still stands that it makes it seems like the user is wrong. 75%~85% of ticks is actually true passing and anything less or more is bad. I'll try to think of a way to demonstrate that.
image

> deep breath

Mainly omitted it because I think it's an AoE talent and don't want to flag it if the fight is not AoE. But I think I can tie this with AoE talents or it's own sub section to show the info if people want to use it.

---
I've been sharing some screenshot reports with people in #devastation discord and also dming some of the experts there. Will get more feedback from them as well! Appreciate the comments left from both of ya'll!

src/analysis/retail/evoker/devastation/modules/abilities/ShatteringStar.tsx
JeaneC: I initially did this because depending on the talent, the empowered spell ID's will change, but then I realized I can just add another case 🤦

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
JeaneC: Oh gocha! I'm so bad, I have to implement the normalizer first, I thoughtI could just check the event for it

JeaneC:
Added a performance box row for shattering star 
image

Added a checkmark pass/fail for DR section and changed the wording a bit to encourage dropping ticks for DR
image

emallson:
Small non-blocking suggestion for the bit about clipping disintegrate during Dragonrage.

Bar: have the bar be full at 85%. bar = full is a good way to visually indicate "you can't do much/any better"

Text:
> During Dragonrage, you want to to *intentionally* clip ticks of Disintegrate after your GCD ends in order to use Essence Burst procs faster. When done correctly, you will get 70-85% of Disintegrate ticks during Dragonrage. \<small\>See the Dragonrage section for additional info.\</small\>

JeaneC:
> Small non-blocking suggestion for the bit about clipping disintegrate during Dragonrage.
> 
> Bar: have the bar be full at 85%. bar = full is a good way to visually indicate "you can't do much/any better"
> 
> Text:
> 
> > During Dragonrage, you want to to _intentionally_ clip ticks of Disintegrate after your GCD ends in order to use Essence Burst procs faster. When done correctly, you will get 70-85% of Disintegrate ticks during Dragonrage. <small>See the Dragonrage section for additional info.</small>

Cool just did that! I also got my terminology corrected from the evoker discord. It's called chaining if you don't drop any ticks, but clipping if you do 💦 

image

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  onLivingFlameCast(event: CastEvent) {
+    if (isFromBurnout(event)) {
+      this.consumedProcs += 1;
+    }
+  }
+
+  get wastedProcs() {
+    return this.procs - this.consumedProcs;
+  }
```
kfinch, line 45: You're missing one special case here - buff stacks present at fight end will be counted as wasted. You can catch these by listening for FightEnd event and tallying. See basically the same case being handled in `RegrowthAndClearcasting#onFightEnd`. You'll need to count stacks though - use `getBuffStacks` instead of `hasBuff`

src/analysis/retail/evoker/devastation/modules/abilities/ShatteringStar.tsx
```diff
@@ -0,0 +1,197 @@
...
+            DISINTEGRATE,
+            FIRE_BREATH,
+            FIRE_BREATH_FONT,
+            ETERNITY_SURGE,
+            ETERNITY_SURGE_FONT,
+            LIVING_FLAME_DAMAGE,
+            LIVING_FLAME_CAST,
+            AZURE_STRIKE,
+            PYRE,
+          ].map((spell) => ({ id: spell.id })),
```
kfinch: Missed a `map` in event listeners

src/analysis/retail/evoker/devastation/modules/abilities/ShatteringStar.tsx
kfinch: Oh that's interesting, I had no idea charging up changed the spellID for Evoker. I'd still prefer it's done with guid - if you find yourself often grouping the same IDs, you can even add a const to a shared file, the way many specs do with `constants.ts`. Up to you how you want to style it though.

kfinch (commented):
Great updates - just a couple minor things left

src/analysis/retail/evoker/devastation/modules/abilities/ShatteringStar.tsx
emallson: name is also susceptible to internationalization issues. I don't remember if we force translation on WoWA, but if we don't then on non-english logs you may get non-english names.

kfinch:
One tip that can help you out / save time on future development - WoWA has a semi-hidden way of showing the full events list as WoWA sees them. Can learn a lot about ability spellIds and event types by browsing it. The following GIF illustrates roughly how I found out about how Burnout works in events:

burnout_stacks_discover

src/analysis/retail/evoker/devastation/modules/abilities/Burnout.tsx
JeaneC: Nice edge case to know about

JeaneC:
> One tip that can help you out / save time on future development - WoWA has a semi-hidden way of showing the full events list as WoWA sees them. Can learn a lot about ability spellIds and event types by browsing it. The following GIF illustrates roughly how I found out about how Burnout works in events:
> 

This would have been so helpful before HAHAHA. Literally the Spell IDs for buffs, talents, and spells were frequently not what was recorded (modified because of some talent or other factor)

----
Improved AoE details a bit, added snapfire and deep breath to cooldowns
image

As well as snapfire procs
image

kfinch (approved):
I think we've basically made it - probably good enough to go in as far as I'm concerned

PR #5578, opened 2022-12-27
allow Ctrl/Cmd-R to "force" a refresh of the fight list when used on the fight selection page

emallson:
this addresses one of the most common UX issues on WoWA: "why doesn't wowa show all of the fights?"

this may eventually be coupled with a backend change that does some more intelligent cache-busting for ongoing reports (e.g. auto-refresh if the last-known end is between 3 and 30 minutes ago and we last checked at least 30 seconds ago), but I want to see the effect this has first


ToppleTheNun (approved):
Apart from the changelog check, looks good to me.

PR #5582, opened 2022-12-28
Removing outdated Enhancement modules for IceStrike and CrashingStorms

sorrellp:
Both Ice Strike and Crashing Storms were updated in Dragonflight causing these modules to become obsolete.

Ice Strike used to refresh the cooldown of Enhancements shock spells when used and the module was tracking how much cooldown reduction was gained on average from Ice Strike. However, in Dragonflight this portion of Ice Strike was reworked and removed so we should no longer be looking for that. 

Crashing Storms was a talent that caused Crash Lightning to leave a pool on the ground that would do AOE damage to enemies standing in it and the module was tracking how much damage this effect did. In Dragonflight this talent has been entirely reworked and no longer leaves a damaging pool on the ground and so there is no longer anything to track.

This PR removes both of these modules to avoid confusion for anyone analyzing their enhance logs.


emallson (approved):
this seems fine, but you have some merge conflicts to resolve. since the change is just "delete the file" it probably won't be too bad

sorrellp:
Should be good now

emallson (approved):

PR #5587, opened 2022-12-29
Add Classic Templates

jazminite:
### Description
- Add Classic healing spec template
- Add Classic shared folder template
- Update Druid spell (`NATURES_SWIFTNESS`) because there's a Shaman spell with the same name

### Motivation
- Makes it easier to add new specs

### Testing
- N/A - No functionality added in this PR


emallson (approved):
tbh i'm not totally sure i like the idea of having a template like this---seems like it will bitrot and require additional maintenance.

that said: i'm down to give it a shot and see how it pans out

jazminite:
> tbh i'm not totally sure i like the idea of having a template like this---seems like it will bitrot and require additional maintenance.
> 
> that said: i'm down to give it a shot and see how it pans out

Thanks - This will save me a ton of time. If it becomes a problem, I'll remove it.

PR #5591, opened 2022-12-30
Stasis module for prevoker

trevorm4:
Adds spell breakdown for stasis to easily figure out what people are spending their stasis on

image


trevorm4:
Some notes:
There were some weird discrepancies in the log where the Stasis buff would be removed and reapplied (happened after hover), which I think is a bug but need to account for it here in order for the module to not crash

There was up to 300ms delay between empowerend and stasis removebuffstack so used 500 as a reasonable buffer.

vohrr (approved):

PR #5603, opened 2023-01-02
Fix 2 piece for rdruid

trevorm4:
Our healing boost was too large and was getting counted as overhealing, let's just take the crit healing and count a percent of it as 2 piece healing


kfinch (approved):
A subtle difference but you're right, this is more accurate

PR #5609, opened 2023-01-04
[👹 Demonology] Add support for `Shadow's Bite` and `Dread Calling` talents

darya-talanina:
[Logs](https://www.warcraftlogs.com/reports/bPRACG3nj7gF1V8c#fight=5) used to test, character: Tidaldrainer

Screenshot 2023-01-05 at 00 34 11

## Shadows Bite ([wowhead](https://www.wowhead.com/spell=387322/shadows-bite))

* 19 buffed casts
* 51 total cast
* 19/51 = 0.3725

totalCasts
withBuff

## Dread Calling ([wowhead](https://www.wowhead.com/spell=387391/dread-calling))

Counted manually

dreadStalkers


src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    this.dreadCallingStacksConsumed += this.selectedCombatant.getBuffStacks(
+      SPELLS.DREAD_CALLING_BUFF.id,
+      event.timestamp,
+    );
+  }
+
+  statistic() {
+    const averageNumberOfStacks = this.dreadCallingStacksConsumed / this.callDreadstalkersCasts;
+    const averageDamageBuff = this.DAMAGE_BUFF_PER_STACK * averageNumberOfStacks;
```
trevorm4: Generally we like to use getters for defining these variables so that other files can easily use them in the future if needed, see https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/analysis/retail/evoker/preservation/modules/talents/Reversion.tsx if you need an example

src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (
+      <Statistic category={STATISTIC_CATEGORY.TALENTS} size="flexible">
+        <BoringSpellValueText spellId={TALENTS_WARLOCK.DREAD_CALLING_TALENT.id}>
+          {formatPercentage(averageDamageBuff, 0)}%{' '}
+          <TooltipElement
+            content={`${averageNumberOfStacks.toFixed(2)} stacks per cast on average`}
+          >
+            <small>average buff on Call Dreadstalkers</small>
+          </TooltipElement>
+        </BoringSpellValueText>
```
trevorm4: Use `TalentValueText` instead of `BoringSpellValueText`

src/analysis/retail/warlock/demonology/modules/talents/ShadowsBite.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    if (this.selectedCombatant.hasBuff(SPELLS.SHADOWS_BITE_BUFF.id, event.timestamp)) {
+      this.buffedDemonboltCasts += 1;
+    }
+  }
+
+  statistic() {
+    return (
+      <Statistic category={STATISTIC_CATEGORY.TALENTS} size="flexible">
+        <BoringSpellValueText spellId={TALENTS_WARLOCK.SHADOWS_BITE_TALENT.id}>
```
trevorm4: Ditto, use TalentValueText

src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Each Soul Shard spent on Hand of Gul'dan increases the damage of your next Call Dreadstalkers by 4%.
+ */
+
+class DreadCalling extends Analyzer {
+  callDreadstalkersCasts: number = 0;
+  dreadCallingStacksConsumed: number = 0;
+
+  DAMAGE_BUFF_PER_STACK = 0.04;
```
trevorm4: Define this as a const variable outside of the class since its not really needed as a member variable

trevorm4 (changes requested):

darya-talanina:
Thanks for the review @trevorm4 . I took all your feedback into account. Could you please take another look?

src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
```diff
@@ -0,0 +1,68 @@
...
+import TalentSpellText from 'parser/ui/TalentSpellText';
+
+const DAMAGE_BUFF_PER_STACK = 0.04;
+
+/**
+ * Each Soul Shard spent on Hand of Gul'dan increases the damage of your next Call Dreadstalkers by 4%.
+ */
+
+class DreadCalling extends Analyzer {
+  _callDreadstalkersCasts: number = 0;
```
ToppleTheNun: More of a note than anything, but prefer declaring these as `private` instead of prefixing them with `_` if they're not intended to be used outside of the class.

ToppleTheNun (approved):

src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
darya-talanina: Thanks for the tip @ToppleTheNun, I didn’t know about `private` keyword.
However The _ here serves to avoid duplicate name between class attribute and getter method. Would you have an idea on how I can avoid it otherwise?

ToppleTheNun: No really strong opinions here, but maybe something like "consumingCasts" and "buffStacksConsumed". The references to which buff and which spell consumed it should be able to be inferred by reading the analyzer.

trevorm4 (approved):

trevorm4:
Looks good overall, lets change to private and then we should be good to go

src/analysis/retail/warlock/demonology/modules/talents/DreadCalling.tsx
darya-talanina: I renamed the attributes to remove "_". To avoid duplication with getter methods names, I made attribute names more concise, while keeping the getters names more explicit, as they are part of class' API and what they do should be clear from their name without having to implementation.

invfo: Could you let me know if you have other feedback or merge the PR otherwise?

darya-talanina:
@trevorm4 should be good to merge now 🚀

PR #5616, opened 2023-01-05
index contributors by nickname

### LINKED ISSUE #5615: Invalid contributors due to nickname mismatch
image

Caused by the nickname field on a `Contributor` not matching the name of the export.

ToppleTheNun:
index contributors by nickname instead of by export name.

closes #5615

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -17,6 +17,7 @@ import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
+  change(date(2023, 1, 5), 'Fix changelogs showing some contributors as invalid.', ToppleTheNun),
   change(date(2023, 1, 1), 'Add Pull Request template.', jazminite),
   change(date(2022, 12, 28), 'Allow fetching Vault of the Incarnates parses.', ToppleTheNun),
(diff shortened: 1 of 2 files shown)


emallson (commented):
is this because when looking up they're looked up by nickname, so there was an implicit assumption that export name was nickname?

ToppleTheNun:
Yep!

```tsx
// src/interface/ContributorDetails.tsx
const { contributorId } = this.props;
const contributor: Contributor | undefined = contributors[contributorId];
```
```tsx
// src/interface/ContributorButton.tsx
const ContributorButton = ({ nickname, avatar, link = true }: Props) => {
 // ...
<ContributorDetails contributorId={nickname} />
 // ...
};
```
```tsx
// src/interface/NewsList
<ReadableListing>
    {item.contributors.map((contributor) => (
        <Contributor key={contributor.nickname} {...contributor} />
    ))}
</ReadableListing>
```

ToppleTheNun:
The first thing I tried to do with this was make an ESLint rule to highlight and fix the issue, but properly fixing it would require a yarn.lock change (change `"eslint-plugin-wowanalyzer": "file:./packages/eslint-plugin-wowanalyzer",` to `"eslint-plugin-wowanalyzer": "link:./packages/eslint-plugin-wowanalyzer",` in package.json) and there are multiple contributors whose nickname can't replicated as an export name (e.g., `4ply`).

emallson (approved):

PR #5617, opened 2023-01-05
healing subtraction for upwelling

trevorm4:
### Description

Subtract out base ef healing (i.e. not from extra bolts or extra hot duration) from upwelling number. 

### Motivation

Upwelling is overcounting and is not nearly as good as wowa says.

### Testing

Before
image

After

image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
@@ -81,12 +81,17 @@ class Upwelling extends Analyzer {
     );
   }
 
-  get averageExtraBolts() {
+  get averageStacks() {
     return this.extraBolts / this.castEF;
   }
 
   get lostEfHealing() {
-    return (this.baseEfHealing * this.averageExtraBolts) / (BASE_BOLTS + 1);
+    if (this.averageStacks < 12) {
+      // number of bolts is 1:1 with normal ef if less than 12 stacks
+      return 0;
+    }
+    const wastedBolts = this.averageStacks - 12;
+    return this.baseEfHealing * (1 - wastedBolts / 6);
   }
 
   get totalHealingAll() {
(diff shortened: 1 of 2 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
```diff
@@ -156,6 +168,8 @@ class Upwelling extends Analyzer {
       this.totalAbsorbs += event.absorbed || 0;
       this.extraBolts += 1;
       this.fromExtraBolts.add(targetID);
+    } else {
+      this.baseEfHealing += event.amount || 0;
```
vohrr: `this.baseEfHealing += (event.amount || 0) + (event.absorbed || 0);`

src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
```diff
@@ -203,8 +217,12 @@ class Upwelling extends Analyzer {
         this.masteryOverhealing += event.overheal || 0;
         this.masteryAbsorbed += event.absorbed || 0;
       }
-      this.masteryTickTock = !this.masteryTickTock;
+    } else {
+      if (!this.masteryTickTock) {
```
vohrr: This does not need to be accounted for. You were never accounting for base gusts procs in any original EF calculation so you do not need to subtract it from additional mastery you would get from extra duration hots

src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
```diff
@@ -80,13 +81,22 @@ class Upwelling extends Analyzer {
     );
   }
 
+  get averageExtraBolts() {
+    return this.extraBolts / this.castEF;
+  }
+
+  get lostEfHealing() {
+    return (this.baseEfHealing * this.averageExtraBolts) / (BASE_BOLTS + 1);
```
vohrr: For percentage of casts missed it would be 1-(avgbolts/18) since your average extra bolts per cast is effectively the percentage toward a full additional cast, not percent missed

vohrr (changes requested):
Made a few comments. This will need a refresh buff event handler as well. An average of 13-15 unique targets hit means pandemic hots are happening virtually every cast

vohrr (approved):

PR #5619, opened 2023-01-06
[Balance] Update boomkin config and description

hartra344:
Just minor updates to the Boomkin config and description for Dragonflight


ToppleTheNun (changes requested):
Add a CHANGELOG entry and this should be good to go.

ToppleTheNun (approved):

PR #5621, opened 2023-01-07
fix math for upwelling

trevorm4:
Fixing formula

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
@@ -13,7 +13,7 @@ import TalentSpellText from 'parser/ui/TalentSpellText';
 import { SpellLink } from 'interface';
 import StatisticListBoxItem from 'parser/ui/StatisticListBoxItem';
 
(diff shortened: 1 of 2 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/Upwelling.tsx
```diff
@@ -86,12 +86,8 @@ class Upwelling extends Analyzer {
...
 
   get lostEfHealing() {
-    if (this.averageStacks < 12) {
-      // number of bolts is 1:1 with normal ef if less than 12 stacks
-      return 0;
-    }
-    const wastedBolts = this.averageStacks - 12;
-    return this.baseEfHealing * (1 - wastedBolts / 6);
+    const lostBolts = this.averageStacks / 2; // you lose .5 bolts a second
+    return this.baseEfHealing * (lostBolts / BASE_BOLTS);
```
emallson, line 90: does this make the module count bolts that aren't added by upwelling for the lost healing? not sure if thats relevant

trevorm4: Yes it does, because upwelling stacks equate to lost casts so just counting extra healing from it and displaying it as a statistic is very misleading

trevorm4:
Merging, me and vohrr discussed this offline

PR #5623, opened 2023-01-07
[hpal] some dragonflight talent fixes and beacon refactoring

mfield:
### Description
Ability configuration:
- dedupe hasted cd logic
- dedupe Unbreakable talent cdr for affected spells
- sanctified wrath is 40% off not 50%
- consecrate is 9s cd for holy
- avenging crusader increases cooldown rate rather than reducing cooldown
  - main difference is if you gain/lose the AC buff during a CD you gain/lose partial benefit
  - whereas with e.g. sanctified wrath if holy shock is on cd and you press wings, the current cd is not affected
  - i'll track the cooldown acceleration in a separate analyzer later with SpellUsable#applyCooldownRateChange

Beacon refactoring:
- beacon types aren't as different as the code made them seem
- reduce duplication, simplify BeaconUptime

### Motivation
Working toward real 10.0.2 support. Virtue is common on a few fights in raid and should be supported properly.

### Testing
used my own logs (don't look ew i played so bad)
beacon of virtue: https://www.warcraftlogs.com/reports/KnHDVW1T7XfAY9gj#type=auras&fight=last&target=30
image
image

beacon of faith:  https://www.warcraftlogs.com/reports/9VKqLN7bTZvPWDXd#fight=1&type=healing
beacon of faith with <100% uptime: https://www.warcraftlogs.com/reports/9VKqLN7bTZvPWDXd#type=healing&fight=29
image
image

Some cooldown related errors in console, no more beacon transfer errors though.


src/common/SPELLS/paladin.ts
```diff
@@ -123,6 +123,11 @@ const spells = spellIndexableList({
     icon: 'spell_holy_flashheal',
     manaCost: 2200,
   },
+  HOLY_SHOCK_CAST: {
+    id: 20473,
+    name: 'Holy Shock',
+    icon: 'spell_holy_searinglight',
+  },
```
mfield: This is a talent (tho perma selected). Warcraft logs sends over Infusion of Light events (buff gained probably?) with this spell ID which leads to <SpellLink> showing the wrong icon/spell name. Seems like the easiest fix for now, even though it duplicates from the TALENTS configuration.

Long term I suspect <SpellLink> (and related) should look up spell ids in TALENTS if it doesn't find anything in SPELLS. Would be happy to attempt if that sounds like a good change.

mfield: on live, all these Infusion of Light should be Holy Shock instead
image

src/analysis/retail/paladin/holy/modules/beacons/BeaconAnalyzer.tsx
```diff
@@ -0,0 +1,23 @@
...
+    if (this.selectedCombatant.hasTalent(TALENTS.BEACON_OF_FAITH_TALENT)) {
+      this.beaconType = 'BEACON_OF_FATH';
+    } else if (this.selectedCombatant.hasTalent(TALENTS.BEACON_OF_VIRTUE_TALENT)) {
+      this.beaconType = 'BEACON_OF_VIRTUE';
+    } else {
+      this.beaconType = 'BEACON_OF_LIGHT';
+    }
+    this.beaconBuffIds = BEACON_TYPES[this.beaconType];
+  }
+}
```
trevorm4: If you need to access beaconType in other files, you should export const variables that store `BEACON_OF_<type>` that way other files don't have to compare to a string and can use the const variable instead.

src/analysis/retail/paladin/holy/modules/beacons/BeaconTargets.ts
```diff
@@ -20,11 +19,15 @@ class BeaconTargets extends Analyzer {
...
   currentBeaconTargets: number[] = [];
+  currentBeaconTargetsByBeaconId: { [beaconId: number]: number[] } = {};
   maxBeacons = 1;
 
   hasBeacon(playerId: number) {
     return this.currentBeaconTargets.includes(playerId);
   }
+  specificBeaconTargets(beaconId: number) {
+    return (this.currentBeaconTargetsByBeaconId[beaconId] ?? []).length;
+  }
```
trevorm4: You may want to rename this to something like `getBeaconTargetForId` or something similar

trevorm4 (changes requested):
small stuff but lgtm overall

src/analysis/retail/paladin/holy/modules/beacons/BeaconAnalyzer.tsx
mfield: Certainly a traditional ask... I get where this is coming from. But I'm not sure this is actually better? The literal string union type checks just as strictly, offers the same exhaustiveness checking, and vscode and jetbrains autocomplete them just as well. So the enum just seems redundant. If blizz added another beacon type, having the enum would increase the surface area of the related wowa change.

emallson: If blizzard adds another beacon type, wouldn't we want to have the uses that need to be exhaustive flagged automatically by the expanded enum?

I do think it is cleaner to use an enum for this regardless. Autocompletion is much more reliable for enum types than string unions

mfield: (i've already put in the enum, not trying to get out of this or anything)
exhaustiveness checking is identical for a const enum or a literal type union [ts playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBApmArgWxgIQKIEEDCB5AOQH0AVATQAUMYBvAKBnW32LwDEiAZASQHEAJEgBoGTXISLsibLNxL8RjTONYcAatwBKJAKoYRAXwDcdOqEixlLIgGUqnTkW4ARGzAC8tUYwDaVieRUAHT+qlx8ggC6AFwwPgCMkYqMcaGklBghzBJSMnL8MXHxQgBMSd6p2cSBmWlSGtp6hT4AzOUGMACGEDDm0CZ0AGaIYMBQAJbg8AAeABadiNDjAG5wGEjIABQARnCd5iQAngAOcLFpNQCUsdsgIAA2e2BeKTAQAO7jUMCzMDt7BxOcEuL1eKWA3TgYmsNSyKkkHB4AhIsQATnAoIhUc8oKjEHAKq8IRAoRcMnDrLlZPI0RisTi8QSweDIdCAuS6uotLoMLTMdiYINOvcSRUDHRxUMRmNJs9tqiQABrBDrFD-fbgI6nc5VdJUa4wW4PJ6gxgfL4-P67DVgLXA01g4mk3WwznhZF8+kwXH4wkskls6oc3VU-KegU+pkpcWSvqWXV2DAOJyucIkDCaLCONQZmzcQgeB1upGCWIJcqvN15GlFUoVlJuho8sttQxdHpxgbDUYTKZwOYLJarThfOCo4XqwHamDKw4gQbeoHzwO2eyOFw2NMZrNEHOaPOEA1Gx6dZ70V7m76-SeaoEg8-M3qsgDkxYiJGf4YZvsfT4Dr5DDhq34T8YHRflvyjZknRgAD4Xqbk9FA8CvSFEUoJgGNTG7GUpnlJUEBHKAxwna0pzOGc4DnBcoCXBc0kTZMNy3TNs1zfMCCPO4TzPCpL0tG9bTvB0iRfN9kWQukI0ZP1GBguDKSA6kQK-b0ZNeGMgA)

i'd argue autocompletion is easier because it should work if you import a typed function without also having to import the enum type. i get that it looks like a magic number and we're conditioned not to like those. but it's really not, only the fixed set of literal strings will be accepted by the compiler, same as using enum values.

src/common/SPELLS/paladin.ts
mfield: solved with help from discord by passing spell objects instead of ids.

src/analysis/retail/paladin/holy/modules/beacons/BeaconAnalyzer.tsx
emallson: the autocompletion thing is more about poor editor support than any fundamental issue with typed string unions. Typed string unions have some other issues with inference (ask me about `as const` on the vega specs sometime :skull:), but otherwise the same as enums. I'm aware.

trevorm4 (approved):
lgtm, just fix the merge conflicts and then we're good to go

mfield:
Sorry for the delay. Merged in upstream and looks fine in dev mode.

PR #5624, opened 2023-01-07
Added statistic for Feral Druid T29 set

kfinch:
Added statistic calculating damage from Feral Druid T29 set.
Also includes barebones implementation of `ItemSetLink` (I don't expect this to be used often and so don't want to build something to complicated yet)
- Test report URL: `/report/LV74Fga2D3NGZYKv/40-Mythic+Eranog+-+Kill+(3:12)/Maymays/standard/statistics`
- Screenshot(s):
image
feral_t29_tooltip


src/interface/BaseTooltipProvider.ts
```diff
@@ -20,6 +20,12 @@ class BaseTooltipProvider {
   static itemRelative(id: number, details: any) {
     throw new Error('NotImplemented');
   }
+  static itemSet(...args: [number, any]) {
```
ToppleTheNun: This probably needs to be moved into "useTooltips" after the PR was merged earlier today

kfinch: Yeah I saw I conflicted with something as soon as I put up the PR - will do a fix at some point today when I have time

PR #5630, opened 2023-01-09
[Shadow] Initial Guide version of shadow priest analysis

DoxAshe:
### Description

This creates a prototype guide view for shadow priest.  It adds sections in this view for:
- Shadow's resource, Insanity
- Damage over time effects
- Active Time and Cancelled Casts
- Cooldowns
- Using Procs
- Preparation
This also removes some analysis of void bolt, which has not been applicable since the rework at the start of shadowlands.

It is not finished, but it covers everything from the normal view and it works for any configuration of shadow's talents without causing an error.

### Motivation

The guide has enough in it to be useful, and there are already a lot of changes.  Before I add additional analysis, I wanted the current analysis to be added into the new view.
I'm hoping to get feedback from the community on adjustments and additions.

### Testing

Most of this guide is based on other classes' guides.
I wanted to add an Insanity graph to this pull request, but I could not get it to be created without crashing.  I do not know if I'm missing something obvious, or if Insanity works differently than other resources, like Astral Power, Fury and Pain.

### Screenshot(s):

Insanity and DoTs
Core and Cooldowns
Proc Usage


kfinch:
This is a quick review based just on the screenshot, haven't had a time to dive into the code yet:
* I like the overall breakdown of information, but I'm a bit confused by the 'Core' section - not much in it yet so I'm not sure what your vision is for it. With DoTs and Proc Usage already pulled out, what is left?
* If Dark Evangelism uptime doesn't change rotation at all, I'm of the opinion it shouldn't be in the Guide section at all. It's still a great candidate to go in the Statistics tab.
* A subjective visual thing, but I think the DoT uptime bars should have different colors so they scan easier. I like pulling colors from the icon, so maybe blueish for Vamp Touch and orangey for SWP
* The breakdown bars in Proc Usage need tooltips - I'm especially confused by the two bars for Mind Flay - what do they mean?

DoxAshe:
> This is a quick review based just on the screenshot, haven't had a time to dive into the code yet:
> 
> * I like the overall breakdown of information, but I'm a bit confused by the 'Core' section - not much in it yet so I'm not sure what your vision is for it. With DoTs and Proc Usage already pulled out, what is left?
> * If Dark Evangelism uptime doesn't change rotation at all, I'm of the opinion it shouldn't be in the Guide section at all. It's still a great candidate to go in the Statistics tab.
> * A subjective visual thing, but I think the DoT uptime bars should have different colors so they scan easier. I like pulling colors from the icon, so maybe blueish for Vamp Touch and orangey for SWP
> * The breakdown bars in Proc Usage need tooltips - I'm especially confused by the two bars for Mind Flay - what do they mean?

Thank you for the suggestions, I've made changes for all of them.
I have removed Dark Evangelism from the core section, and left it as a statistic.  I have changed the colors of the DoT uptime bars.  I've removed one of the bars from Mind Flay: Insanity and moved it to a statistic, since the information in that bar does not effect shadow's rotation and it makes the section clearer.  
The core section is a little empty right now, but it will eventually have casts of mind blast and shadow word: death, as well as the active time graph, but is still very small. I've moved Insanity and DoTs into Core as Sections.  Let me know if you think this looks better:

NewCore

Improved MFI

src/analysis/retail/priest/shadow/modules/guide/ExtraComponents.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const ColoredStrong = styled.strong`
+  color: ${(props) => props.color};
+`;
+
+interface PerformanceStrongProps extends ComponentProps<typeof ColoredStrong> {
+  children: ReactNode;
+  performance: QualitativePerformance;
+}
+export const PerformanceStrong = ({ performance, ...others }: PerformanceStrongProps) => (
```
ToppleTheNun, line 14: I want to move this to a shared place instead of copying it between specs. It doesn't need to be fixed in this PR.

src/analysis/retail/priest/shadow/modules/resources/InsanityUsage.tsx
```diff
@@ -2,10 +2,14 @@ import { t } from '@lingui/macro';
 import { formatPercentage } from 'common/format';
 import SPELLS from 'common/SPELLS';
 import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
+import { Panel } from 'interface';
+import ResourceBreakdown from 'parser/shared/modules/resources/resourcetracker/ResourceBreakdown';
 import { SpellLink } from 'interface';
 import { ResourceLink } from 'interface';
 import Analyzer from 'parser/core/Analyzer';
 import { When, ThresholdStyle } from 'parser/core/ParseResults';
+//import ResourceBreakdown from 'parser/shared/modules/resources/resourcetracker/ResourceBreakdown';
```
ToppleTheNun: These can be removed as they were added above.

ToppleTheNun (changes requested):
One very small thing that should be fixed and then it's good to go imo.

src/analysis/retail/priest/shadow/modules/guide/ExtraComponents.tsx
DoxAshe: That makes sense.  Whenever it is moved to a shared place I will update shadow to use that.

src/analysis/retail/priest/shadow/modules/resources/InsanityUsage.tsx
DoxAshe: I have removed both of those.

ToppleTheNun (approved):

PR #5631, opened 2023-01-09
Add essence chart for Evoker

trevorm4:
### Description

Adding chart to breakdown essence spent by ability

### Testing
Devastation 
image
Preservation 
image


tyndivelspaz:
Doesn't Pyre cost essence for Devastation? Would it be helpful to put that into the array to capture as well? And possibly putting Disintegrate into Preservation?

Reason for Disintegrate for Preservation would be to display all usages of essence. On the argumentative side of that reasoning would be to only display what is relevant and ideal. I'm okay either way just a suggestion on if we should give all the data vs ideal/suggested spell data.

vohrr:
> Doesn't Pyre cost essence for Devastation? Would it be helpful to put that into the array to capture as well? And possibly putting Disintegrate into Preservation?
> 
> Reason for Disintegrate for Preservation would be to display all usages of essence. On the argumentative side of that reasoning would be to only display what is relevant and ideal. I'm okay either way just a suggestion on if we should give all the data vs ideal/suggested spell data.

Those are all included, the donut chart is filtering out any items that were not cast in a given log

@trevorm4 this is redundant with the Essence Usage tab 
image

tyndivelspaz:
That is the last time I do a PR on mobile. Brain decided to just skip right over them.

trevorm4:
> > Doesn't Pyre cost essence for Devastation? Would it be helpful to put that into the array to capture as well? And possibly putting Disintegrate into Preservation?
> > Reason for Disintegrate for Preservation would be to display all usages of essence. On the argumentative side of that reasoning would be to only display what is relevant and ideal. I'm okay either way just a suggestion on if we should give all the data vs ideal/suggested spell data.
> 
> Those are all included, the donut chart is filtering out any items that were not cast in a given log
> 
> @trevorm4 this is redundant with the Essence Usage tab image

Yeah ur right lolz

PR #5642, opened 2023-01-13
Add Classic Caster template

jazminite:
### Description
- Add Classic Caster template for WotLK

### Motivation
- Makes it easier for new users (and myself) to add new specs

### Testing
- There's no live functionality yet but I took this screenshot after testing it on local:
image


ToppleTheNun (approved):
As emallson previously approved the other template and the screenshot looks good, 👍

PR #5645, opened 2023-01-14
add PTR talents

ToppleTheNun:
add ability to generate talents for PTR builds.

update existing spellpower.

move devastation evoker buff that was added to the
talents file instead of the spells file.

I checked the Druid change as well and it matches what's in-game.


emallson:
My main question here is: how do you use this for development?

If you manually import the PTR file, then you have to update everywhere that you imported it when transitioning to live. This could potentially be `codemod`-ed, especially if we delete the PTR files after the patch goes live.

Merging the PTR talents into the main talents seems like a pretty universally bad idea, so that is probably the only way to practically use this.

I don't think that we want the PTR files to be everpresent when the PTR spends most of its time with no talent changes.

ToppleTheNun:
> If you manually import the PTR file, then you have to update everywhere that you imported it when transitioning to live. This could potentially be codemod-ed, especially if we delete the PTR files after the patch goes live.

This was where my thought process was heading with the relatively large changes happening to talents in the 10.0.5 PTR builds (Guardian Druid whole tree rework, for instance). I mostly wanted to enable people who wanted to start working on it ahead of time to do so, similarly to how we enabled people to work on Dragonflight ahead of time.

I'm trying to think through options for how to flag the use of PTR talents for cases where we'd subsequently delete them after a patch.

> Merging the PTR talents into the main talents seems like a pretty universally bad idea, so that is probably the only way to practically use this.

Agreed here.

> I don't think that we want the PTR files to be everpresent when the PTR spends most of its time with no talent changes.

I don't necessarily agree or disagree here, simc has the spell data everpresent and it doesn't seem to cause issues. I could be persuaded either way on it.

Pewtro:
I think that PTR data has much more of a reason to exist in simc than in wowa, because there's quite often _some_ things different on PTR than on live, even if it isn't talents - and there's plenty of simc implementation going on ahead of time, much less so for log analysis (and that there's plenty of sims done with ptr=1 vs virtually 0 log analysis done from PTR logs) 

I, personally, think it'd be better that if people want to do PTR talent stuff they could temporarily add them elsewhere (for the specific spec/class) vs us having PTR talents (for everyone) lying around always.

ToppleTheNun:
Fair enough. I've been persuaded. For the temporarily added talents, it'd probably be worth adding some documentation around how to do it.

emallson:
They can also run the talent generation script with the updated build number on their PTR branch, which we can add instructions for.

Pewtro:
I mean, I guess it depends if they want to just do a quick and dirty into SPELLS or want to generate a fully fledged talents setup. 
Alternatively they just use the readme in scripts/talents and generate for PTR in a branch and work on that and then PR when PTR is going live (can add some comments in that readme for that purpose)

ToppleTheNun:
I'll make a new branch and add documentation in the readme in the `scripts/talents` directory on how to generate talents based on PTR, including getting the talent tree info and the spellpower.

emallson:
(might be worth backporting your devastation evoker fixes though :slightly_smiling_face:)

ToppleTheNun:
In that case, I'll just re-open this and mark it as a draft, then ping you when it's ready to go again.

ToppleTheNun:
@emallson should be good for re-review.

emallson (approved):

PR #5646, opened 2023-01-14
add playwright tests for validation

ToppleTheNun:
add playwright tests to help with automatic validation of large scale changes.

image

image


emallson:
Looks like the tests timed out. I'm a big fan of this conceptually. The test you wrote for vengeance can probably be made generic over all specs using the CONFIG's example report.

ToppleTheNun:
I need to figure out what's causing it to timeout, it times out in Webkit on Windows for me as well.

e2e/report-selection.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+test('loading Vengeance DH report', async ({ page }) => {
+  await page.goto('/');
+
+  await page.getByPlaceholder('https://www.warcraftlogs.com/reports/<report code>').click();
+  await page
+    .getByPlaceholder('https://www.warcraftlogs.com/reports/<report code>')
+    .fill('https://www.warcraftlogs.com/reports/hg34bJNkWAfXCmFn');
+  await page.getByRole('heading', { name: 'Fight selection' }).waitFor();
+  await expect(page).toHaveURL('/report/hg34bJNkWAfXCmFn');
```
MartijnHols: When making e2e tests, I usually like to stop my test here so a single test has a single origin -> action -> result as much as possible. I am well aware this isn't ideal for most e2e tests, but in this case you're navigating over multiple completely different pages, and when that's the case I usually prefer them as separate test suites (you could have something like _player selection_, _fight selection_ and _report_ test suites). Since this test ensures the URL has to be `/report/hg34bJNkWAfXCmFn` if it succeeds, the next test can safely use `await page.goto('/report/hg34bJNkWAfXCmFn')`. The benefit is smaller, focussed tests that only test a single thing. A nice bonus is that this would also test whether opening a direct link (/report/hg34bJNkWAfXCmFn) works correctly and doesn't rely on additional state.

MartijnHols:
I was wondering why you went with Playwright instead of Cypress. Looking at your tests, I like the syntax a lot more, and that alone might be a good reason. I need to give it a try some day.

ToppleTheNun:
> I was wondering why you went with Playwright instead of Cypress. Looking at your tests, I like the syntax a lot more, and that alone might be a good reason. I need to give it a try some day.

I've had some significant issues with Cypress in the past, so wanted to give Playwright a try instead. Two additional reasons I found are:
- It doesn't automatically install the browser dependencies. I've previously had issues where Cypress tried downloading the browsers during the dependency install and it bricked the installation due to a download failure. Not requiring every user to have the browsers installed felt good as well.
- The test generation/recording mechanism is incredibly helpful for writing tests. You can start a generation session and just click through your application and it will generate a basic test file for you. You can then change it however you wish.

e2e/report-selection.spec.ts
ToppleTheNun: I agree that there should be broken down tests as well, but I don't see any harm in having one long test that goes through the entire user journey. The only thing that might be less palatable about having both would have some level of duplication, imo.

ToppleTheNun:
> The test you wrote for vengeance can probably be made generic over all specs using the CONFIG's example report.

It looks like there are some limitations here (that shouldn't be, per Microsoft):
```
src\analysis\retail\deathknight\blood\CONFIG.tsx:24
  description: <>
               ^

SyntaxError: Unexpected token '<'
```

Sample test file:
```
import { expect, test } from '@playwright/test';

import CONFIGS from '../src/parser';
import { isCurrentExpansion } from '../src/game/Expansion';

const supportedConfigs = CONFIGS.filter(it => it.exampleReport && isCurrentExpansion(it.expansion));

test("", async ({page}) => {
});
```

I'll open an issue over on the Playwright repo about this if I can get a reproduction together.

ToppleTheNun:
I got the tests working for loading all specs example reports. I did have to trim the list to a few that I know work, as a few specs have something breaking during the tests:
- Mistweaver Monk example report is for Skolex from last expac and it fails to load in development mode; it would work in production though.
- Unholy DK is erroring on switching to the `About` tab due to something with runes.

emallson:
Okay, i did some work on the partial support modal in particular. Between that and a little work on the test itself, partially supported specs are no longer flaky.

I changed it to wait for the `networkidle` load state to help improve reliability. Its not quite realistic, but it seemed to help?

ToppleTheNun:
Will pull locally and test it out a bit

ToppleTheNun:
There's still some occasional flakiness with the About tab tests that I can spend a bit more time trying to pin down, but it mostly seems to happen when loading the report takes a long time.

emallson:
FWIW most of the flakiness i experienced after the fixes I put in went away when i ran with a single worker (`-j1`), which is also how CI runs.

ToppleTheNun:
That's probably a good enough indication that trying to run this particular one down isn't worth the squeeze.

PR #5647, opened 2023-01-14
Fixed an issue where the Events tab would 404

kfinch:
The Events tab link wasn't being called right after the changes. The below fix works, but I'm not sure if it's the "right" way to do it.


ToppleTheNun (approved):
A relative link should have worked, I'll look into why it didn't later

PR #5657, opened 2023-01-15
[Monk] Minor stuff

Hurstilol:
Clean up Changelog, adding Innerpeace max energy to the energytracker


trevorm4 (changes requested):
Remove old contributors from changelog and then we should be good to go

PR #5660, opened 2023-01-16
Added support for Unleash Life target increases for Chain Heal, Downpour and Healing Rain statistics.

Fassbrause:
Added support for Unleash Life target increases for Chain Heal, Downpour and Healing Rain statistics.

image
image

Recommendation for Chain Heal will now check whether Ancestral Reach is talented and increase suggested target count based on this.

I left target increases from Unleash Life out of the max target recommendations as this changes depending on how you use Unleash Life and could confuse people. I am not quite sure on how to best communicate this yet, maybe we want to replace the hard numbers (out of 6 targets possible for Healing Rain for example) to a (out of the maximum amount of targets).


src/analysis/retail/shaman/restoration/modules/spells/HealingRain.tsx
```diff
@@ -98,7 +120,42 @@ class HealingRain extends Analyzer {
...
       // dirty fix for partial ticks happening at the same time as a real tick
       healingRainTick.hits =
-        healingRainTick.hits + 1 > 6 ? (healingRainTick.hits = 6) : healingRainTick.hits + 1;
+        healingRainTick.hits + 1 > this.maxTargets
+          ? (healingRainTick.hits = this.maxTargets)
+          : healingRainTick.hits + 1;
+    }
+  }
+
+  _onCast(event: CastEvent) {
```
ToppleTheNun, line 129: Prefer `private` functions instead of prefixing the name of the function with `_` if you don't want other people calling them

ToppleTheNun (approved):

PR #5661, opened 2023-01-16
Updated cooldown durations to match new DF talents (Unholy DK)

Bicepspump:
### Description

Updated outdated Cooldown durations to match new Dragonflight version.

Talents covered:

- Dark Transformation
- Apocalypse
- Unholy Assault

### Motivation

Spec is currently not indicated as supported, I intend to change that through a set of PRs.

### Testing

Load up https://www.warcraftlogs.com/reports/jXn6BbyPh97JV1Fm#fight=24&type=damage-done, cooldowns now correctly state 95% or so, was previously more than 100% (used old CDR logic for cooldowns).

- Test report URL: https://www.warcraftlogs.com/reports/jXn6BbyPh97JV1Fm#fight=24&type=damage-done

### DIFF
--- a/src/analysis/retail/deathknight/unholy/modules/features/Abilities.tsx
+++ b/src/analysis/retail/deathknight/unholy/modules/features/Abilities.tsx
@@ -8,6 +8,7 @@ import SPELL_CATEGORY from 'parser/core/SPELL_CATEGORY';
 class Abilities extends CoreAbilities {
   spellbook(): SpellbookAbility[] {
     const combatant = this.selectedCombatant;
+    const UNHOLY_COMMAND_SCALING = [0, 8, 15];
     return [
       // roational
       {
@@ -55,7 +56,8 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.DARK_TRANSFORMATION.id,
         category: SPELL_CATEGORY.COOLDOWNS,
-        cooldown: 60 - combatant.getTalentRank(talents.UNHOLY_COMMAND_TALENT) * 7.5,
+        cooldown:
+          60 - UNHOLY_COMMAND_SCALING[combatant.getTalentRank(talents.UNHOLY_COMMAND_TALENT)],
         gcd: {
           base: 1500,
         },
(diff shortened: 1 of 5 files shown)


src/analysis/retail/deathknight/unholy/modules/features/Abilities.tsx
```diff
@@ -55,7 +55,7 @@ class Abilities extends CoreAbilities {
       {
         spell: SPELLS.DARK_TRANSFORMATION.id,
         category: SPELL_CATEGORY.COOLDOWNS,
-        cooldown: 60,
+        cooldown: 60 - combatant.getTalentRank(talents.UNHOLY_COMMAND_TALENT) * 7.5,
```
Khazakdk: Should be `* 8` right?

`Effects          :
#1 (id=801793)   : Apply Aura (6) | Add Flat Modifier (107): Spell Cooldown (11)
                   Base Value: -8000 | Scaled Value: -8000 | PvP Coefficient: 1.00000 | Misc Value: 11 | Target: Self (1)
                   Affected Spells: Dark Transformation (63560)
                   Family Flags: 69
Description      : The cooldown of Dark Transformation is reduced by ${$s1/-1000} sec.`

Khazakdk (changes requested):
One minor note to make sure a value is correct and then its good to go!

src/analysis/retail/deathknight/unholy/modules/features/Abilities.tsx
Bicepspump: This was actually a shortcut as the talent is 8 and then 15 seconds reduction. Do you want me to change it to have an if else statement instead to correctly account for both ranks (not sure how valuable this is since I believe it's very rare anyone takes one out of two of these talent points).

ToppleTheNun: I would suggest a constant declared somewhere that you can reference like:
```typescript
export const UNHOLY_COMMAND_SCALING = [0, 8, 15];

// ...

cooldown: 60 - UNHOLY_COMMAND_SCALING[combatant.getTalentRank(talents.UNHOLY_COMMAND_TALENT)]
```

Bicepspump: Sounds good!

ToppleTheNun (approved):

PR #5680, opened 2023-01-18
[Balance] Implement Statistics Tracking for Rattle The Stars

hartra344:
### Description
A set of statistics to track the usefulness of Rattle The Stars boomkin talent.

### Motivation
One of the primary buffs that Moonkins need to manage, this lets Boomkin analyze how well they managed it and what they were able to get from it.

### Testing
image

### DIFF
--- a/src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
+++ b/src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
@@ -1,0 +1,188 @@
+import { formatDuration, formatNumber, formatPercentage } from 'common/format';
+import SPELLS from 'common/SPELLS';
+import HasteIcon from 'interface/icons/Haste';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, {
+  EventType,
+  ApplyBuffEvent,
+  ApplyBuffStackEvent,
+  RemoveBuffEvent,
+  RemoveBuffStackEvent,
+  FightEndEvent,
+  CastEvent,
+  DamageEvent,
+} from 'parser/core/Events';
+import { currentStacks } from 'parser/shared/modules/helpers/Stacks';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import { TALENTS_DRUID } from 'common/TALENTS';
+import { calculateEffectiveDamage } from 'parser/core/EventCalculateLib';
+import {
+  STARFALL_BASE_COST,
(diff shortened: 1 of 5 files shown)


src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
```diff
@@ -0,0 +1,189 @@
...
+              </tbody>
+            </table>
+          </>
+        }
+      >
+        <BoringSpellValueText spellId={TALENTS_DRUID.RATTLE_THE_STARS_TALENT.id}>
+          <>
+            {formatNumber(this.totalAPSaved)} <small>Astral Power Saved</small>
+            <br />
+            {formatNumber(dpsIncrease)} <small>approx. DPS added</small>
```
emallson: since you're not accounting for the (potential) increased cast count, i would put a `>` on this. like `>3,804 approx. DPS added`

src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
```diff
@@ -0,0 +1,189 @@
...
+                {this.buffStacks.map((e, i) => (
+                  <tr key={i}>
+                    <th>{i} Stacks</th>
+                    <td>{this.buffedStarsurges[i]}</td>
+                    <td>{this.buffedStarfalls[i]}</td>
+                    <td>
+                      {formatPercentage(e.reduce((a, b) => a + b, 0) / this.owner.fightDuration)}%
+                    </td>
+                  </tr>
+                ))}
```
emallson, line 172: i feel like we have something that covers this use-case already but idk what it is 🤔

hartra344: I think I stole this calculation from the Starlord module, would be happy to try something else if we have something better.

src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
```diff
@@ -0,0 +1,190 @@
...
+            </table>
+          </>
+        }
+      >
+        <BoringSpellValueText spellId={TALENTS_DRUID.RATTLE_THE_STARS_TALENT.id}>
+          <>
+            {formatNumber(this.totalAPSaved)} <small>Astral Power Saved</small>
+            <br />
+            {'>'}
+            {formatNumber(dpsIncrease)} <small>approx. DPS added</small>
```
kfinch: There actually is an existing module that helps format all of this and does the DPS calculation as well - you can use `<ItemPercentDamageDone greaterThan amount={this.totalAddedDamage} />`

src/analysis/retail/druid/balance/modules/spells/RattleTheStars.tsx
kfinch: This should be fine for now - I can get this in

kfinch (approved):

Issue #5561, opened 2022-12-23
[Hunter] Dragonflight Updates


Pewtro:
This list will be filled out as I get reports in, and as I register things myself over the coming weeks. 

## Features 
- [ ] Update APL Checker for Beast Mastery
- [ ] Update APL Checker for Marksmanship
- [ ] Update APL Checker for Survival
- [ ] Start work on transitioning to some guide formats for Beast Mastery 
- [ ] Start work on transitioning to some guide formats for Marksmanship
- [ ] Start work on transitioning to some guide formats for Survival

## Talents
- [ ] 

## Known Bugs / Issues 
- [ ] Serpentstalker's Trickery talent is causing Serpent Sting suggestion to pop up because it casts Serpent Sting too often.

Pewtro:
Outdated and I didn't get around to it

PR #5686, opened 2023-01-20
[Hunter] Update Barbed Shot and Kill Command

Pewtro:
### Description

Barbed Shot and Kill Command have a lot of hidden resets and interactions with each other that has a large impact on their cast efficiency. This commit aims to improve that by handling some of them as best as we can by assuming the latest possible reset trigger, is what reset a given spell

### Motivation

Touching up cast efficiency as part of updating to be 10.0 accurate. 
Towards #5561 

### Testing

- Test report URL: `/report/NBMbz2Dcq43k17tJ/90-Mythic+Eranog+-+Kill+(2:41)/Povish/standard/statistics`
- Screenshot(s):

Before
image

After 
image

### DIFF
--- a/src/analysis/retail/hunter/beastmastery/modules/core/SpellUsable.tsx
+++ b/src/analysis/retail/hunter/beastmastery/modules/core/SpellUsable.tsx
@@ -58,22 +58,12 @@ class SpellUsable extends CoreSpellUsable {
     const spellId = triggerEvent.ability.guid;
     if (spellId === TALENTS.BARBED_SHOT_TALENT.id) {
       if (this.isOnCooldown(spellId) && this.chargesAvailable(spellId) === 0) {
-        this.endCooldown(
-          spellId,
-          this.lastPotentialTriggerForBarbedShotReset
-            ? this.lastPotentialTriggerForBarbedShotReset.timestamp
-            : undefined,
-        );
+        this.endCooldown(spellId, this.lastPotentialTriggerForBarbedShotReset?.timestamp);
       }
     }
     if (spellId === TALENTS.KILL_COMMAND_SHARED_TALENT.id && this._has4pc) {
       if (this.isOnCooldown(spellId) && this.chargesAvailable(spellId) === 0) {
-        this.endCooldown(
-          spellId,
-          this.lastPotentialTriggerForKillCommandReset
-            ? this.lastPotentialTriggerForKillCommandReset.timestamp
-            : undefined,
-        );
+        this.endCooldown(spellId, this.lastPotentialTriggerForKillCommandReset?.timestamp);
       }
     }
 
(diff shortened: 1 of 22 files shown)


src/analysis/retail/hunter/beastmastery/modules/core/SpellUsable.tsx
```diff
@@ -1,31 +1,63 @@
...
 
 class SpellUsable extends CoreSpellUsable {
   static dependencies = {
     ...CoreSpellUsable.dependencies,
   };
 
-  lastPotentialTriggerForBarbedShotReset: CastEvent | null = null;
+  lastPotentialTriggerForBarbedShotReset: AnyEvent | null = null;
+  lastPotentialTriggerForKillCommandReset: AnyEvent | null = null;
+  _has4pc: boolean = false;
```
emallson:
```suggestion
  private _has4pc: boolean = false;
```

src/analysis/retail/hunter/beastmastery/modules/core/SpellUsable.tsx
```diff
@@ -34,6 +66,17 @@ class SpellUsable extends CoreSpellUsable {
         );
       }
     }
+    if (spellId === TALENTS.KILL_COMMAND_SHARED_TALENT.id && this._has4pc) {
+      if (this.isOnCooldown(spellId) && this.chargesAvailable(spellId) === 0) {
+        this.endCooldown(
+          spellId,
+          this.lastPotentialTriggerForKillCommandReset
+            ? this.lastPotentialTriggerForKillCommandReset.timestamp
+            : undefined,
```
emallson: this can be simplified to:
```suggestion
          this.lastPotentialTriggerForKillCommandReset?.timestamp,
```

src/parser/core/EventLinkNormalizer.ts
```diff
@@ -35,15 +35,15 @@ export type EventLink = {
...
   anySource?: boolean;
-  /** Iff true, the two events may be linked even with different targets.
+  /** If true, the two events may be linked even with different targets.
    * In most cases this should be false, and will default to false when omitted */
   anyTarget?: boolean;
-  /** Iff defined, links will also be added from the referenced event to the linking event using this relation. */
+  /** If defined, links will also be added from the referenced event to the linking event using this relation. */
   reverseLinkRelation?: string;
-  /** Iff defined, this spec will create at most the given number of links to matching reference events.
+  /** If defined, this spec will create at most the given number of links to matching reference events.
```
emallson, line 46: just as a note, the `iff`s here are "if and only if", not just typos. however in pretty much every case they're equivalent to the regular usage of "if" so idc if we change them.

emallson (changes requested):
looks pretty fine to me. couple of nitpicks.

one bigger comment is: could the DireBeast / DireCommand modules be written more simply using the `freecast` events from the normalizer (since it is already doing the work of distinguishing them)?

i am not sure the answer is yes and i don't necessarily think that ought block the PR

Pewtro:
The cast event unfortunately doesn't carry the target so it can't help distinguish where the Dire Beast originated from, but I might be able to simplify some of the logic using the FreeCastEvent, I'll take a stab at it tomorrow with fresh eyes

emallson:
> The cast event unfortunately doesn't carry the target

oh that makes sense 👍

src/parser/core/EventLinkNormalizer.ts
Pewtro: I'll leave them as is, they all seem like regular ifs the ones I found

PR #5695, opened 2023-01-23
Add font of magic module

trevorm4:
### Description

Adding font of magic module which adds an extra target to spiritbloom at max empower, dream breath and fire breath do the same amount of damage/healing total regardless of empower so not gonna include it

### Testing
image
image
https://www.warcraftlogs.com/reports/Jmp4nYZkwLHh1K7q#fight=4&type=healing&source=10&translate=true


oisinod (approved):
👍

PR #5701, opened 2023-01-24
[Balance] T29 4P

ap2355:
### Description

Module for Touch the Cosmos (Balance Druid T29 4P)

### Motivation

New module

### Testing

- Test report URL: /report/z3BJ1DkRrqp6y4j8/30-Mythic+Sennarth,+The+Cold+Breath+-+Kill+(6:53)/Idraly/standard/statistics
- Screenshot(s): 
- 
image
image


hartra344:
I think I would consder removing the damage component on this one since there's really no way to correctly measure the damage amp for starfall. I think the > here hand waves it a little too much and isn't really a great approximation on any fight other than terros.

kfinch:
I'd be fine with the damage increase number if Starfall could be included, but I presume overlapping Starfalls make this not doable?

hartra344:
> I'd be fine with the damage increase number if Starfall could be included, but I presume overlapping Starfalls make this not doable?



> I'd be fine with the damage increase number if Starfall could be included, but I presume overlapping Starfalls make this not doable?

yeah I don't think this is feasible since it's assigned to starfall on cast rather than damage and there's no way to descern which starfall ticks has the damage amp.

src/analysis/retail/druid/balance/modules/spells/TouchTheCosmos.tsx
```diff
@@ -0,0 +1,155 @@
...
+    //Starfall
+    if (event.ability.guid === SPELLS.STARFALL.id) {
+      return;
+    }
+  }
+
+  statistic() {
+    const dpsAdded = this.totalDamage / (this.owner.fightDuration / 1000);
+    return (
+      <Statistic
```
hartra344, line 46: one more nit I realized looking at the 2P code, cant you add it to the items category?

ap2355:
I "removed" all the damage part but left it commented in the file if we want to reconsider it in a near future. This module will disappear next patch anyway.

hartra344 (approved):

emallson (changes requested):
rather than doing a lot of commenting out, just delete the old code. we can retrieve it from git's history (and if you have questions / aren't sure how to do this, you can always ping me in discord for help)

PR #5715, opened 2023-01-25
add veil of pride module

trevorm4:
Add veil of pride module, add sub stats for all SG modules, add suggestion for SG overcapping, and mark MW as fully supported

image


vohrr (approved):
lgtm

PR #5717, opened 2023-01-26
[BALANCE] Slight improvement on Eclipse and filler spell suggestions …

hartra344:
### Description

Loosening of filler suggestions since wrath is used in far more scenarios and changing the description of how eclipse to should be used to better align with playstyle in Dragonflight.

### Testing

- Test report URL: /report/Y1F4wq2cxpgDbMBT/31-Mythic+Kurog+Grimtotem+-+Kill+(8:47)/Fullmoonphil/standard/overview
- Screenshot(s): 
image


ap2355:
The whole comment at the beginning of FillerUsage is outdated, either remove it or correct it.
I don't if the guide section is more about filler usage or eclipse in itself but you could track if the player enters Solar which should never happen at the moment.
Would you prefer display Umbral Embrace usage in this guide section or in one of its own ?

hartra344:
> The whole comment at the beginning of FillerUsage is outdated, either remove it or correct it. I don't if the guide section is more about filler usage or eclipse in itself but you could track if the player enters Solar which should never happen at the moment. Would you prefer display Umbral Embrace usage in this guide section or in one of its own ?

Yeah theres a lot still wrong that needs to be updated. The primary thing I was looking to patch up here was the checklist description that was misleading and the timeline view where pretty much every filler was red and saying it was wrong. This whole thing needs reworked before being made into a guide.

Good call on the comment cleanup though, done.

emallson (approved):

PR #5722, opened 2023-01-26
[hpal] implement/update some paladin cdr and cooldown acceleration sources

mfield:
### Description
Paladin CDR and cooldown acceleration changed a bit from shadowlands to dragonflight.

### Motivation
More accurate SpellUsable stats

### Testing
Some spot checking on my own logs e.g. `/report/nAzLjQWmZTDcv93g/20-Mythic+Terros+-+Kill+(6:04)/Glimmerer/standard/statistics`


src/parser/core/EventFilter.ts
```diff
@@ -7,7 +7,7 @@ const VALID_BY_FLAGS = SELECTED_PLAYER | SELECTED_PLAYER_PET;
 export type SpellInfo = {
   id: number;
 };
-export type SpellFilter<T extends SpellInfo = SpellInfo> = T | T[];
+export type SpellFilter<T extends SpellInfo = SpellInfo> = T | readonly T[];
```
ToppleTheNun, line 10: I don't think I have any issues with this, but curious about the change.

mfield: i'm a fan of `as const` so this just lets me pass in something like this while setting up an event listener https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5722/commits/04c5418bda8e7bdf92868d301b42420b6114812e#diff-772fdc75020e2577a67c4499412077cb9a6706b8780828867e8426f2ce2047b8R13-R23

src/parser/ui/BoringSpellValueText.tsx
```diff
@@ -2,12 +2,13 @@
...
  * Use this only as the last resort, prefer a custom layout instead.
  */
+import Spell from 'common/SPELLS/Spell';
 import { SpellIcon } from 'interface';
 import { SpellLink } from 'interface';
 import { ReactNode } from 'react';
 
 interface Props {
-  spellId: number;
+  spellId: number | Spell;
```
ToppleTheNun, line 11: I think I'd rather we change this to a different name but that's not for this PR. We're a bit inconsistent in how spells vs spell IDs are handled.

ToppleTheNun (approved):

PR #5723, opened 2023-01-26
move PerformanceBoxRow to grid instead of flexbox

ToppleTheNun:
### Description

Move PerformanceBoxRow to CSS grid instead of flexbox.

### Motivation

Flexbox-based PerformanceBoxRow from raid:
image

Flexbox-based PerformanceBoxRow from dungeon:
image

Flexbox-based PerformanceBoxRow doesn't wrap and individual boxes have min-widths, causing them to expand outside of
their designated container. I solved for this in VDH by adding `flex-wrap: wrap`, but it causes other problems:
image

A flexbox container with `flex-wrap: wrap` that has wrapped contents will see each wrapped line have individually sized items;
it isn't possible to force multiple lines to have the same child item size using flexbox.

Enter CSS grid.

The same VDH block from before:
image

### Testing

- Test report URL: `/report/Fdfk6LPncg17BHAr/13-Heroic+Raszageth+the+Storm-Eater+-+Kill+(9:24)/2-Sigilweaver/standard/overview`, `/report/PdZjrmG4qaK8y7WD/44-Heroic+Terros+-+Kill+(6:16)/Mauly/standard/overview`
- Screenshot(s): Linked multiple above :)


ToppleTheNun:
Merging this on Monday morning if no concerns arise by then.

PR #5739, opened 2023-01-28
Update Holy Priest for 10.0.5

foobartles:
### Description

Update the spec with the few changes from 10.0.5 by updating the statistics for Burning Vehemence and Searing Light.

### Motivation

Keep spec up to date.

### Testing

https://www.warcraftlogs.com/reports/2ngQFxHh7ftNR4V6#fight=42&type=damage-done&source=112
Log using burning vehemence and searing light

### DIFF
--- a/src/analysis/retail/priest/holy/modules/talents/BottomRow/EmpoweredRenew.tsx
+++ b/src/analysis/retail/priest/holy/modules/talents/BottomRow/EmpoweredRenew.tsx
@@ -7,7 +7,6 @@ import AbilityTracker from '../../core/AbilityTracker';
 import SPELLS from 'common/SPELLS';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
 import TalentSpellText from 'parser/ui/TalentSpellText';
-import { SpellLink } from 'interface';
 
 /**
  * Renew instantly heals your target for 10% of its total periodic effect.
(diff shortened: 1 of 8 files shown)


src/analysis/retail/priest/holy/modules/talents/BottomRow/EmpoweredRenew.tsx
```diff
@@ -31,25 +30,13 @@ class EmpoweredRenew extends Analyzer {
   get effectiveHealing() {
     return (
       this.abilityTracker.getAbility(SPELLS.EMPOWERED_RENEW_TALENT_HEAL.id).healingEffective +
-      this.abilityTracker.getAbility(SPELLS.EMPOWERED_RENEW_TALENT_HEAL.id).healingAbsorbed +
-      this.echoOfLightMastery.masteryHealingBySpell[SPELLS.EMPOWERED_RENEW_TALENT_HEAL.id]
```
foobartles, line 35: removing mastery from this since we are not including mastery in any other has modules. also was causing bug if crash if player never casted renew with this talent

src/common/TALENTS/priest.ts
```diff
@@ -142,6 +142,13 @@ const talents = createTalentList({
     maxRanks: 2,
     entryIds: [103738],
   },
+  BURNING_VEHEMENCE_TALENT_DAMAGE: {
```
ToppleTheNun: This should go in `src/common/SPELLS/priest.ts` instead of here so that it doesn't get lost the next time we do a talent regeneration.

ToppleTheNun (changes requested):

trevorm4 (approved):

PR #5742, opened 2023-01-28
add Spark of Insight module

trevorm4:
adding spark module as requested by Sups

image


oisinod:
Does this account for the case where you already have one (or two with essence attunement) stacks of essence burst and are just overwriting the stack you already had?

trevorm4:
Nope, let me add refreshes.

oisinod (approved):

PR #5753, opened 2023-01-30
[Hunter] Early work on Beast Mastery Guide 

Pewtro:
#### Hunter 

Resource Usage: 
image

Rotation: 
image

#### Others

This PR turns off some console.logs() we have in prod, in armorMitigation and in NavigationBar. 

This PR also fixes an issue with the buff stack tracker that's present on production, if we look at https://wowanalyzer.com/report/JDZh8Yz3gmWA2dKR/2-Mythic+Eranog+-+Kill+(3:29)/Artydh/standard/overview we will see that the graph sometimes reports there are 7 stacks of soul fragments, but if we look into the events that never happens. 
image

This PR fixes that by using the payloads from the apply/removebuffstack events, which results in this: 
image

That result matches what we see in the events tab, the only user of BuffStackTracker and BuffStackGraph is @ToppleTheNun so requesting a review on that change here. 
### Testing

- Screenshot(s):


src/analysis/retail/hunter/beastmastery/normalizers/BarbedShotNormalizer.ts
```diff
@@ -0,0 +1,47 @@
...
+    const fixedEvents: AnyEvent[] = [];
+    const possibleBarbedShotBuffs = [
+      SPELLS.BARBED_SHOT_BUFF.id,
+      SPELLS.BARBED_SHOT_BUFF_2.id,
+      SPELLS.BARBED_SHOT_BUFF_3.id,
+      SPELLS.BARBED_SHOT_BUFF_4.id,
+      SPELLS.BARBED_SHOT_BUFF_5.id,
+      SPELLS.BARBED_SHOT_BUFF_6.id,
+      SPELLS.BARBED_SHOT_BUFF_7.id,
+      SPELLS.BARBED_SHOT_BUFF_8.id,
```
emallson, line 18: NotLikeDuck

emallson (commented):
looks fine to me. assuming you'll want to wait on topple's review.

src/analysis/retail/hunter/beastmastery/normalizers/BarbedShotNormalizer.ts
Pewtro: I wanted to use the EventOrderNormalizer,  but this has to move the event multiple events back so here we are 😢
1 cast = 7 events on the timestamp with the actual cast in the middle NotLikeDuck 
image

src/parser/retail/armorMitigation.ts
```diff
@@ -51,7 +51,7 @@ export const getArmorCoefficient = (fight: WCLFight) =>
 
 const clamp = (actual: number, min: number, max: number) => Math.min(max, Math.max(min, actual));
 
-const debug = true;
+const debug = false;
```
ToppleTheNun, line 54: Thanks for fixing this!

ToppleTheNun (approved):
Looks good to me; DH Soul Fragment stacks _can_ get up to 7, albeit briefly, and not in the log used here. This doesn't prevent tracking up to 7, so it should be fine.

PR #5754, opened 2023-01-30
Fix chilled clarity in potionchecker

trevorm4:
We were missing spell ids for r2/r3 leading to bad suggestions


Pewtro:
Having 4 arrays of the same potions seems a bit overkill, we should probably do something about that at some point (at least combat potion in the two files are the same so could be 1 value), I know a mana pot is in one of them but not in the others etc, but otherwise no reason for us to have that many representations of same info.

Pewtro (approved):

ToppleTheNun (approved):

trevorm4:
One of the arrays is for strong potions (aka max rank) but certain specs don't need max rank chilled clarity, which is why I had to include all of them in that array.

ToppleTheNun:
This was a bigger thing in Shadowlands due to Shadowlands having multiple distinct spell IDs for each tier of potion (Cosmic vs Spiritual). Chilled Clarity is the only potion that has distinct IDs atm, at least until next tier when we might get new healing potions.

Pewtro:
Sure but these two look to be exactly same as an example
```typescript
static extraAbilityInfo = {
    name: 'Combat Potion',
    buffSpellId: [
      SPELLS.ELEMENTAL_POTION_OF_ULTIMATE_POWER.id,
      SPELLS.ELEMENTAL_POTION_OF_POWER.id,
      SPELLS.BOTTLED_PUTRESCENCE.id,
      SPELLS.POTION_OF_FROZEN_FOCUS.id,
      SPELLS.RESIDUAL_NEURAL_CHANNELING_AGENT.id,
      SPELLS.DELICATE_SUSPENSION_OF_SPORES.id,
      SPELLS.POTION_OF_CHILLED_CLARITY_R1.id,
      SPELLS.POTION_OF_CHILLED_CLARITY_R2.id,
      SPELLS.POTION_OF_CHILLED_CLARITY_R3.id,
    ],
```
and 
```typescript
export const COMBAT_POTIONS: number[] = [
  SPELLS.ELEMENTAL_POTION_OF_ULTIMATE_POWER.id,
  SPELLS.ELEMENTAL_POTION_OF_POWER.id,
  SPELLS.BOTTLED_PUTRESCENCE.id,
  SPELLS.POTION_OF_FROZEN_FOCUS.id,
  SPELLS.RESIDUAL_NEURAL_CHANNELING_AGENT.id,
  SPELLS.DELICATE_SUSPENSION_OF_SPORES.id,
  SPELLS.POTION_OF_CHILLED_CLARITY_R1.id,
  SPELLS.POTION_OF_CHILLED_CLARITY_R2.id,
  SPELLS.POTION_OF_CHILLED_CLARITY_R3.id,
];
```

trevorm4:
Yeah those seem unnecessary for sure.

ToppleTheNun:
Those can probably be merged, yeah.

PR #5763, opened 2023-02-01
rewrite test utilities in TypeScript

ToppleTheNun:
### Description

Integration tests can't be written easily atm due to tsconfig exclusions. This is one step towards making it easy to write tests again.


src/parser/core/tests/TestCombatLogParser.ts
```diff
@@ -0,0 +1,64 @@
...
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    config: Config = {} as Config,
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    report: Report = { friendlies: [] as PlayerInfo[], friendlyPets: [] as PetInfo[] } as Report,
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    selectedPlayer: PlayerInfo = { id: 1 } as PlayerInfo,
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    selectedFight: Fight = { start_time: 0, offset_time: 0, filtered: false } as Fight,
+    combatantInfoEvents: CombatantInfoEvent[] = [],
+  ) {
```
emallson: can we avoid all of these lint rules by defining these as constants outside the class with the `as X` assertions on them?

ToppleTheNun: Yep, went ahead and added all of the currently unused values on the objects as well.

emallson (approved):

PR #5800, opened 2023-02-15
Warrior

Ac1dBomb:
### Testing

- Screenshot(s):


jazminite (changes requested):
CI isn't passing because you need to add a changelog entry. You can add it here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/CHANGELOG.tsx

If you haven't contributed before, you will need to add your info here:
https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/CONTRIBUTORS.ts

PR #5806, opened 2023-02-15
[Shadow] Add additional information about cooldowns to guide view

DoxAshe:
### Description

This adds additional analysis and descriptions to Shadow Crash, Void Eruption, Void Torrent, and Twins of the Sun Priestess to shadow's guide. 

### Motivation

Some of shadow's cooldowns needed better descriptions and analysis in the guide view.

### Testing

- Screenshot(s):
Shadow Cooldowns


emallson (approved):
Looks good :+1: 

A general note on design: you currently have this organized as "for each metric, show all cooldown usages." I think it is generally more usable if you have it as "for each cooldown usage, show all metrics".

**However,** (a) we don't have great re-usable stuff built for that (and the examples that I'd had in the Brew stuff kinda aren't used any more cuz talents changed), and (b) I'm not entirely sure how that translates to Shadow since you have multiple small cooldowns that you overlap. In the case where people *correctly* overlap them its easy enough, but the point of wowa is helping with the cases don't do it correctly :upside_down_face: 

**This is not a blocker** and I'm good to merge this rn.

PR #5809, opened 2023-02-16
[OUTLAW] Audacity first implementation

Zac-oihfdwrrtuinvbcp:
### Description

 Add Audacity buff, set the buff to be shown in the timeline for easier tracking, Add necessary code so that ambush is considered higher prio than sinisters/pistol with audacity proc up and will suggest it to the user both in the checklist and the suggestions sections

### Motivation

 Outlaw spec is currently unsupported, I will try working on it bit by bit if possible.

### Testing

- Test report URL: `[/report/...](https://www.warcraftlogs.com/reports/9T2xP3Hyw1GJ48Dd#fight=21&type=damage-done&source=2)`
- Screenshot(s):
image


src/analysis/retail/rogue/outlaw/modules/features/Checklist/Component.tsx
```diff
@@ -108,6 +108,26 @@ const OutlawRogueChecklist = ({
           thresholds={thresholds.opportunity}
         />
       </Rule>
+      <Rule
+        name="Make sure to use your audacity procs"
```
trevorm4:
```suggestion
        name=<>Make sure to use your <SpellLink id={TALENTS.AUDACITY_TALENT.id}/> procs</>
```

Also, change usages of SPELLS.AUDACITY to TALENTS.AUDACITY_TALENT

src/common/SPELLS/rogue.ts
```diff
@@ -573,6 +573,11 @@ const spells = spellIndexableList({
     name: 'Opportunity',
     icon: 'ability_rogue_pistolshot',
   },
+  AUDACITY: {
```
trevorm4:
```suggestion
  AUDACITY_BUFF: {
```

trevorm4 (changes requested):
Prefer using TALENTS.AUDACITY_TALENT when possible and use jsx blocks with SpellLink for readability

src/common/SPELLS/rogue.ts
Zac-oihfdwrrtuinvbcp: I replaced this by `AUDACITY_TALENT_BUFF:` instead because I noticed other talents buffs being named this way, I hope that is correct

trevorm4 (changes requested):
Can you add a changelog entry? After that, we should be good to go

trevorm4 (approved):

PR #5814, opened 2023-02-17
fix essence font miscalcs for mw

trevorm4:
Adds event linking for essence font to be more accurate

Before:
image
(keep in mind I had a buffer of 1 and every single one of these is 17/18 instead of 18/18)
After:
image


src/analysis/retail/monk/mistweaver/modules/spells/EssenceFont.tsx
```diff
@@ -189,30 +198,44 @@ class EssenceFont extends Analyzer {
...
 
-    let tooltip = null;
+  onCast(event: CastEvent) {
+    const totalHit = getNumberOfBolts(event);
+    const expected = this.getExpectedApplies(event);
+    const percentHit = totalHit / expected;
     let value = QualitativePerformance.Perfect;
-    if (cancelled) {
+    if (percentHit < 0.85) {
+      this.numCancelled += 1;
```
vohrr, line 225: Now that this is much more accurate, it would be considered a cancelled cast if percentHit is anything other than 1, correct? So it should only be QP.Good/Perfect and QP.Fail being pushed to castEntries

vohrr (changes requested):

src/analysis/retail/monk/mistweaver/modules/spells/EssenceFont.tsx
trevorm4: Previously, 16 and above were not considered cancels. I think it just depends on how strict we want to be, I don't thinking missing 1 or 2 bolts is a huge deal.

vohrr (approved):

PR #5819, opened 2023-02-18
First pass at Frost Death Knight Guide

Khazakdk:
This also includes a change to ResourceBreakdown to give it a scale factor parameter similar to what was done with the resource graph.  Without this parameter, resources like Runic Power and Astral Power were being shown at 10x their actual values.  Tagging @kfinch since he's familiar with the similar change made to ResourceGraph

### Testing

- Test report URL: `/report/RVqnbxYaCvN1crPH/18-Mythic+Kurog+Grimtotem+-+Kill+(9:05)/Khazak/standard/overview#prototype
- Screenshot(s):
image
image


src/analysis/retail/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
```diff
@@ -49,12 +61,26 @@ class BreathOfSindragosa extends Analyzer {
...
       this.casts -= 1;
     }
+    this.totalDuration += event.timestamp - this.beginTimestamp;
+    const duration = event.timestamp - this.beginTimestamp;
+    this.castTracker.push({
+      timestamp: this.beginTimestamp,
+      startingRunicPower: this.startingRunicPower / 10,
+      duration: duration / 1000,
+      fightEnded: true,
+    });
```
kfinch, line 83: Overall looks good and a nice improvement to ResourceBreakdown. Only suggestion would be to add `RUNIC_POWER_SCALE_FACTOR` to a constants file somewhere, and use that value both in `RunicPowerDetails` and for startingRunicPower like `startingRunicPower: this.startingRunicPower * RUNIC_POWER_SCALE_FACTOR`

kfinch (approved):

src/analysis/retail/deathknight/frost/modules/talents/BreathOfSindragosa.tsx
Khazakdk: Good idea.  I'll do that for AP as well

PR #5824, opened 2023-02-19
handle RP "fights" for classic

emallson:
there is a small chance that on rare occasions this could load out-of-date combatantinfo. we could whitelist specific fights (Algalon, Yogg-Saron, and Thorim all have this problem. Previously several fights in SWP/BT did as well), but that requires maintaining a whitelist and I'd rather not.

the consequences of using the wrong combatantinfo basically means loading the wrong gear/talents (and thus potentially picking the wrong spec)

you can see the issue in action here: https://classic.warcraftlogs.com/reports/1TNz8XaFcb7KhVAP#fight=43&type=summary&view=events

if you switch to fight 42, you can see the combatantinfo and encounterstart events. this is due to the way that WCL splits (buggy) RP from fights.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -19,6 +19,7 @@ import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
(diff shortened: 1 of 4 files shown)


jazminite (approved):
lgtm - loaded your branch and tested several pages

PR #5826, opened 2023-02-20
[Classic Restoration Druid] Correctly attribute revitalize when multiple rdruids in raid

Arbixal:
### Description

Attributes the Revitalize proc to the correct resto druid. Previously both resto druids would get credit for all revitalize procs.

### Motivation

Reviewing a guild log with the rare situation of two resto druids, and needed more accurate information.

### Testing

Compare current functionality to the new one for logs with two resto druids (finding such logs will be the hardest part).

- Test report URL: `/report/FLh4VKfxZWgnPRw9/14-Normal+General+Vezax+-+Kill+(7:07)/Bixshift/standard/statistics`
- Screenshot(s):
### Before
image

### After
image
plus
image

### DIFF
--- a/src/analysis/classic/druid/shared/talents/Revitalize.tsx
+++ b/src/analysis/classic/druid/shared/talents/Revitalize.tsx
@@ -5,7 +5,7 @@ import { SpellIcon } from 'interface';
 import { SpecIcon } from 'interface';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Combatant from 'parser/core/Combatant';
-import Events, { EventType, AnyEvent, HealEvent } from 'parser/core/Events';
+import Events, { EventType, AnyEvent, HealEvent, ResourceChangeEvent } from 'parser/core/Events';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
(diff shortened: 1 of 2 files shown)


src/analysis/classic/druid/shared/talents/Revitalize.tsx
```diff
@@ -73,7 +91,11 @@ class Revitalize extends Analyzer {
   }
 
   resourceEventReduce = (totals: ResourcesByPlayer, evt: AnyEvent) => {
-    if (evt.type !== EventType.ResourceChange) {
+    if (
+      evt.type !== EventType.ResourceChange || // Not a resource event
+      !this.eligibleCasts[truncTimestamp(evt)] || // truncated timestamp does not match a rejuv/wild growth heal
+      !this.eligibleCasts[truncTimestamp(evt)][evt.targetID] // no heal for truncated timestamp for the given target
+    ) {
```
emallson: this can bin two adjacent events into different bins. for example: cast at timestamp 100 and revitalize at 101. those now have different bins and don't match.

your best bet might be storing the eligible casts sorted by timestamp, then doing a binary search for one that is within your tolerance

Arbixal: Good idea. If there is a better/existing function let me know, otherwise I have implemented my own.

emallson (approved):

PR #5830, opened 2023-02-22
[OUTLAW] Fan the hammer normalizer

Zac-oihfdwrrtuinvbcp:
### Description

This normalizer remove the subsequent pistol shots cast events generated by "fan the hammer" talent that are not diferenciated from player casted pistol shots in log events, this makes procs analyzer more accurate and will be useful for future analyzers that won't want to take these events into account.

For now this normalizer is just removing the unwanted events, this doesn't seem to cause any issues but replacing them with another type of event (damage events?) would probably be better? If needed I will update it when I figure out how to.

### Testing

- Test report URL: `[/report/...](https://www.warcraftlogs.com/reports/xkjMa7yvKdDwHWRz#fight=21&type=damage-done&source=240)`
- Screenshot(s):
old:
image
new:
image


ToppleTheNun:
Would checking if stacks of Opportunity were consumed within some millisecond window allow you to detect if a given Pistol Shot cast was affected by Fan the Hammer?

Zac-oihfdwrrtuinvbcp:
> Would checking if stacks of Opportunity were consumed within some millisecond window allow you to detect if a given Pistol Shot cast was affected by Fan the Hammer?

No I don't think so, a stack is being consumed for every hard casted ps and every fth shot without distinction

ToppleTheNun:
This is out of curiosity, but would checking the energy cost of the Pistol Shot "cast" be viable? I checked one of the events and it had an energy cost of `0`.

Another thing I'm thinking that you might be able to do is to provide a spec `GlobalCooldown` analyzer that overrides it similarly to how Survival does it for Barrage:
```tsx
import SPELLS from 'common/SPELLS/rogue';
import { CastEvent } from 'parser/core/Events';
import CoreGlobalCooldown from 'parser/shared/modules/GlobalCooldown';
import getResourceSpent from 'parser/core/getResourceSpent';
import RESOURCE_TYPES from 'game/RESOURCE_TYPES';

class GlobalCooldown extends CoreGlobalCooldown {
  static dependencies = {
    ...CoreGlobalCooldown.dependencies,
  };

  /*
   * Pistol Shots triggered by Fan the Hammer show up as casts of the same spell ID, but aren't actually casts.
   * Because they aren't casts, they don't have GCD events.
   */
  onCast(event: CastEvent) {
    const energyCost = getResourceSpent(event, RESOURCE_TYPES.ENERGY);
    if (spellId === SPELLS.PISTOL_SHOT.id && energyCost === 0) {
      return;
    }
    const isOnGCD = this.isOnGlobalCooldown(spellId);
    if (!isOnGCD) {
      return;
    }
    super.onCast(event);
  }
}

export default GlobalCooldown;
```
This ends up looking like this instead and has the benefit of not deleting the cast event:
image

Zac-oihfdwrrtuinvbcp:
Checking the energy cost would definitely be a better idea yea you are right, 
About the spec GlobalCooldown analyzer, I'm not certain it would exactly achieve all my goals? 2 examples coming to my mind: in the audacity analyzer I'm currently checking with a FilteredDamageTracker if the player casted pistol shot with an audacity buff which would be wrong, this wouldn't work for this particular use as the cast event would still get triggered if I understand it correctly?
Another example: I'm currently trying to rework the cp overcap module: with how outlaw currently works (and it was similar in SL), it doesn't make much sense checking cp overcap as it's not really a relevant metric, however I do want to check if the player casted certain spells at a specific cp threshold depending on rules, casting pistol shot at 6cp for example would be incorrect, without deleting the fake casts event they would trigger this as well.

Although I'm thinking I could just check if the casts are "fake" in the different relevant modules which might just be better.

ToppleTheNun:
I would advocate for checking if they have a global cooldown on the CastEvent being processed by the observer, though that means using `subscribeToCastEvent` instead of `subscribeInefficientCast`.

Zac-oihfdwrrtuinvbcp:
If that's alright with you I would like to stay with the idea of the FtH normalizer for now as it makes some of my incoming changes easier to deal with. I will probably move on to your GlobalCooldown analyzer idea in the future tho, will just have to figure some stuff out that I can't quite do right now with my currently limited knowledge (Never developed in anything other than C/C++/C# before, and never worked on a project of this type/size)

ToppleTheNun (approved):
I'm okay with the current solution in the short-term; long-term this should be rewritten so that the events aren't deleted.

PR #5842, opened 2023-02-24
[OUTLAW] Guide resource section

Zac-oihfdwrrtuinvbcp:
### Description

First pass at a guide with a resource section for now, mostly copied from the assassination one although there are some slight differences.

Added a small section to track Blade rush usage as it's a spell mostly used for energy efficiency (currently would have small issues in aoe fights), will also add killing spree(and maybe thistle tea) here in the future since its used for the same purpose.

The builderUse section is tracking builders casted above the recommended finisher points threshold (with some exceptions) instead of checking if cp were wasted, this seems more relevant for outlaw as we have been overcaping a decent amount of combo point in DF (and even with some SL builds) making raw cp waste tracking almost irrelevant.

### Motivation

Working toward outlaw 10.0.5+ support.

### Testing

- Test report URL: `https://www.warcraftlogs.com/reports/t27zV8NLgPvJq3Kk#fight=17&type=damage-done&source=18`
- Screenshot(s):
image

### DIFF
--- a/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
+++ b/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
@@ -7,13 +7,19 @@ import Statistic from 'parser/ui/Statistic';
 import { STATISTIC_ORDER } from 'parser/ui/StatisticBox';
 import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import getResourceSpent from 'parser/core/getResourceSpent';
-
 import {
   FINISHERS,
   getMaxComboPoints,
 } from '../../constants';
+import Finishers from '../features/Finishers';
 
 export default class FinisherUse extends Analyzer {
+  static dependencies = {
+    finishers: Finishers,
+  };
+
+  protected finishers!: Finishers;
+
   totalFinisherCasts = 0;
   lowCpFinisherCasts = 0;
 
@@ -69,7 +75,7 @@ export default class FinisherUse extends Analyzer {
     }
 
     this.totalFinisherCasts += 1;
-    if (cpsSpent < getMaxComboPoints(this.selectedCombatant) - 1) {
+    if (cpsSpent < this.finishers.recommendedFinisherPoints()) {
         this.lowCpFinisherCasts += 1;
     }
   }
--- a/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
+++ b/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
@@ -1,0 +1,76 @@
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, { CastEvent } from 'parser/core/Events';
+import { BadColor, GoodColor } from 'interface/guide';
+import { ResourceLink} from 'interface';
+import DonutChart from 'parser/ui/DonutChart';
+import Statistic from 'parser/ui/Statistic';
+import { STATISTIC_ORDER } from 'parser/ui/StatisticBox';
+import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
+import getResourceSpent from 'parser/core/getResourceSpent';
+
+import {
+  FINISHERS,
+  getMaxComboPoints,
+} from '../../constants';
+
+export default class FinisherUse extends Analyzer {
+  totalFinisherCasts = 0;
+  lowCpFinisherCasts = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(FINISHERS), this.onCast);
+  }
+
+  get maxCpFinishers() {
+    return (
+      this.totalFinisherCasts -
+      this.lowCpFinisherCasts
+    );
+  }
+
+  get chart() {
+    const items = [
+      {
+        color: GoodColor,
+        label: 'Max CP Finishers',
+        value: this.maxCpFinishers,
+        tooltip: (
+          <>This includes finishers cast at {getMaxComboPoints(this.selectedCombatant) - 1} CPs.</>
+        ),
+      },
+      {
+        color: BadColor,
+        label: 'Low CP Finishers',
+        value: this.lowCpFinisherCasts,
+      },
+    ];
+
+    return <DonutChart items={items} />;
+  }
+
+  statistic() {
+    return (
+      <Statistic position={STATISTIC_ORDER.CORE(6)}>
+        <div className="pad">
+          <label>
+            <ResourceLink id={RESOURCE_TYPES.COMBO_POINTS.id} /> spender usage
+          </label>
+          {this.chart}
+        </div>
+      </Statistic>
+    );
+  }
+
+  private onCast(event: CastEvent) {
+    const cpsSpent = getResourceSpent(event, RESOURCE_TYPES.COMBO_POINTS);
+    if (cpsSpent === 0) {
+      return;
+    }
+
+    this.totalFinisherCasts += 1;
+    if (cpsSpent < getMaxComboPoints(this.selectedCombatant) - 1) {
+        this.lowCpFinisherCasts += 1;
+    }
+  }
+}
--- a/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
+++ b/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
@@ -1,16 +1,13 @@
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Events, { CastEvent } from 'parser/core/Events';
 import { BadColor, GoodColor } from 'interface/guide';
-import { ResourceLink} from 'interface';
+import { ResourceLink } from 'interface';
 import DonutChart from 'parser/ui/DonutChart';
 import Statistic from 'parser/ui/Statistic';
 import { STATISTIC_ORDER } from 'parser/ui/StatisticBox';
 import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import getResourceSpent from 'parser/core/getResourceSpent';
-import {
-  FINISHERS,
-  getMaxComboPoints,
-} from '../../constants';
+import { FINISHERS, getMaxComboPoints } from '../../constants';
 import Finishers from '../features/Finishers';
 
 export default class FinisherUse extends Analyzer {
@@ -29,10 +26,7 @@ export default class FinisherUse extends Analyzer {
   }
 
   get maxCpFinishers() {
-    return (
-      this.totalFinisherCasts -
-      this.lowCpFinisherCasts
-    );
+    return this.totalFinisherCasts - this.lowCpFinisherCasts;
   }
 
   get chart() {
@@ -76,7 +70,7 @@ export default class FinisherUse extends Analyzer {
 
     this.totalFinisherCasts += 1;
     if (cpsSpent < this.finishers.recommendedFinisherPoints()) {
-        this.lowCpFinisherCasts += 1;
+      this.lowCpFinisherCasts += 1;
     }
   }
 }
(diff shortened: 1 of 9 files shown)


Zac-oihfdwrrtuinvbcp:
I would welcome any input on alternative ways to display the blade rush stuffs as I'm unsure the DonutChart are really appropriate here and I'm not yet familiar with everything that can be done.

trevorm4:
So for these types of buff uptimes during conditons, I would recommend using a uptimeBarSubStatistic, see https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/analysis/retail/monk/mistweaver/modules/spells/VivaciousVivify.tsx

And for good casts vs bad casts of finishers, I would recommend using a PerformanceBoxRow see https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/analysis/retail/monk/mistweaver/modules/spells/EssenceFont.tsx

trevorm4 (changes requested):
sending back to your queue to use better visuals

Zac-oihfdwrrtuinvbcp:
This is the updated version for the blade rush part, I think its indeed way better this way, added a performance box row aswell to track "bad" casts
image


For the finisher stuffs I will be keeping it that way as this was meant to be a quick overview to show if the user was often using early finishers or not, I plan to add a more detailled finisher subsection with a PerformanceBoxRow in the core rotation section that will be more in depth by tracking early finishers and between the eyes vs slice and dice vs dispatch finishers choice

ToppleTheNun:
> For the finisher stuffs I will be keeping it that way as this was meant to be a quick overview to show if the user was often using early finishers or not, I plan to add a more detailled finisher subsection with a PerformanceBoxRow in the core rotation section that will be more in depth by tracking early finishers and between the eyes vs slice and dice vs dispatch finishers choice

This is how both Assassination Rogue and Feral Druid handle it, fwiw.

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -15,44 +33,143 @@ class BladeRush extends Analyzer {
...
+  }
+
+  get guide(): JSX.Element {
+    this.unusableUptimes.at(-1)!.end = this.owner.fight.end_time;
+    const explanation = (
+      <p>
+          <strong>
+            <SpellLink id={TALENTS.BLADE_RUSH_TALENT} />
+          </strong>{' '}
+          should be used whenever your energy drop under 70-80, this energy treshold increase with
```
ToppleTheNun:
```suggestion
          should be used whenever your energy drop under 70-80, this energy threshold increase with
```

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -15,44 +33,143 @@ class BladeRush extends Analyzer {
...
+
+  // Count blade rush time spent off cd
+  onBladeRushUsable(event: UpdateSpellUsableEvent) {
+    if(event.updateType === UpdateSpellUsableType.BeginCooldown) {
+        this.endUsablePeriod(event.timestamp);
+      }
+  }
+
+  get guide(): JSX.Element {
+    this.unusableUptimes.at(-1)!.end = this.owner.fight.end_time;
```
ToppleTheNun: I would recommend not using `!` here, it will cause an error if there aren't any values in `this.unusableUptimes`. It's also manipulating state in a function that's intended to be returning a value for display. I'd suggest adding an event listener for `fightend` and doing this there:

```typescript
    // other stuff here
    this.addEventListener(Events.fightend, this.finalize);
  }

  // other stuff here

  private finalize() {
    const uptime = this.unusableUptimes.at(-1);
    if (!uptime) {
      return;
    }

    uptime.end = this.owner.fight.end_time;
  }
```

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -1,10 +1,28 @@
...
 import TALENTS from 'common/TALENTS/rogue';
+import { SpellLink } from 'interface';
+import { RoundedPanel } from 'interface/guide/components/GuideDivs';
+import { Uptime } from 'parser/ui/UptimeBar';
+import uptimeBarSubStatistic from 'parser/ui/UptimeBarSubStatistic';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
+import { BoxRowEntry, PerformanceBoxRow } from 'interface/guide/components/PerformanceBoxRow';
+import { QualitativePerformance } from 'parser/ui/QualitativePerformance';
+
+const MIN_ENERGY_TRESHOLD = 70;
```
ToppleTheNun:
```suggestion
const MIN_ENERGY_THRESHOLD = 70;
```

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -1,10 +1,28 @@
...
+const MAX_ENERGY_TRESHOLD = 90;
```
ToppleTheNun:
```suggestion
const MAX_ENERGY_THRESHOLD = 90;
```

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -1,10 +1,28 @@
...
+const ACCEPTABLE_ENERGY_TRESHOLD = 80;
```
ToppleTheNun:
```suggestion
const ACCEPTABLE_ENERGY_THRESHOLD = 80;
```

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -1,10 +1,28 @@
...
+
+
+//-- TODO: MAX_ENERGY_TRESHOLD should scale with targets hit to support aoe fights
```
ToppleTheNun: It's somewhat difficult to get a good handle on whether or not a fight is a good candidate for AoE solely based on logs. For instance, if you're assigned to be boss group on Broodkeeper Diurna, you likely won't be doing any AoE damage despite the fight's potential. On Eranog, you have very short periods of high burst AoE.

It's not to say that it's impossible, but there are likely lower hanging fruit that will provide more value.

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
```diff
@@ -15,44 +33,143 @@ class BladeRush extends Analyzer {
...
+      this.onBladeRushUsable,
     );
+    
+    this.unusableUptimes.push({
+      start: this.owner.fight.start_time,
+      end: -1,
+    });
+  }
+
+  endUsablePeriod(timestamp: number) {
```
ToppleTheNun: It's not really an issue, but consider marking functions not intended to be consumed from outside of this module as `private`.

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { STATISTIC_ORDER } from 'parser/ui/StatisticBox';
+import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
+import getResourceSpent from 'parser/core/getResourceSpent';
+import {
+  FINISHERS,
+  getMaxComboPoints,
+} from '../../constants';
+import Finishers from '../features/Finishers';
+
+export default class FinisherUse extends Analyzer {
```
ToppleTheNun, line 13: This isn't a thing for this PR, but this indicates to me that we should probably have a shared version of this class somewhere.

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  totalFinisherCasts = 0;
+  lowCpFinisherCasts = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(FINISHERS), this.onCast);
+  }
+
+  get maxCpFinishers() {
```
ToppleTheNun, line 28: I'm unaware of Outlaw's use of Animacharged; is that something that's worth representing here? You can check the Assassination module for an implementation if it is.

This also isn't an issue for this PR.

ToppleTheNun (changes requested):
Sorry for the delay on this. There are a few comments that need to be fixed (mostly spelling things) and a couple notes for later PRs or for myself.

There are also some odd formatting issues that I've noticed; are the Git hooks running?

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
Zac-oihfdwrrtuinvbcp: I chose to not really bother with the keep it rolling - resounding clarity build for now as it has fallen quite a bit behind now, just checked the top 100 of every boss and couldnt find a single log playing RC.
It's something I want to support in the future of course but for now it seemed irrelevant.

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
Zac-oihfdwrrtuinvbcp: Yep, a lot of my first classes were mostly copy paste with slight changes of your assassination stuff, so having some of that shared would probably be a good idea.

src/analysis/retail/rogue/outlaw/modules/talents/BladeRush.tsx
Zac-oihfdwrrtuinvbcp: Maybe this comment was missunderstood so let me explain it a bit better, blade rush basically gain value with target count, so let say the threshold in single target would be 70 energy, with 5 targets it would increase to be much higher to the point you would almost want to press the spell on cd.
So the idea behind this was to check the numbers of targets around the user to implement that as a factor to the Max_energy_treshold.

I'm not entirely sure if we can check the numbers of targets around us tho? But at the very least outlaw has blade flurry constantly running in aoe so I could check the number of target hit.

Zac-oihfdwrrtuinvbcp:
> Sorry for the delay on this. There are a few comments that need to be fixed (mostly spelling things) and a couple notes for later PRs or for myself.
> 
> There are also some odd formatting issues that I've noticed; are the Git hooks running?

Still new to this so I'm not even sure what git hooks are supposed to be, but I will have a look.

ToppleTheNun:
> Still new to this so I'm not even sure what git hooks are supposed to be, but I will have a look.

Git hooks are scripts installed your Git repository that "hook" into the Git lifecycle. An example of this (that we use) is a pre-commit hooks that formats code before it's committed. In our case, we take every file that is "staged" and run a linter/formatter over those files to ensure that all files are consistent and similar.

What tool are you using to do your commits?

Zac-oihfdwrrtuinvbcp:
I'm just using GitHub desktop, I did notice code reformatting by itself so I guess I have at least something going on

ToppleTheNun:
I pulled the branch and run the formatter and it's making some changes. Can you try running `git commit --amend` in a terminal instance and see if it's outputting something like the below?
image

Zac-oihfdwrrtuinvbcp:
image
Just this

ToppleTheNun:
If you make a change in a file (just add a newline or something) and add it with `git add thepathtothefile` followed by `git commit --amend`, does the lint-staged message from before pop up as well?

Zac-oihfdwrrtuinvbcp:
Yeah that worked, cant see all the lint-staged messages anymore my console thing is probably not great but it did reformate like half the file

Zac-oihfdwrrtuinvbcp:
Oh yeah so the issue is just on my school computer that is a bit scuffed and apparently doesnt have git linked with github desktop, the reformatting seems to work fine when I commit from my personal computer so this shouldn't be an issue anymore.

ToppleTheNun:
Can you run the files in this commit through the format process with prettier as well?

Zac-oihfdwrrtuinvbcp:
Should have fixed everything

ToppleTheNun (approved):

src/common/SPELLS/rogue.ts
```diff
@@ -578,6 +578,14 @@ const spells = spellIndexableList({
     name: 'Audacity',
     icon: 'ability_rogue_ambush',
   },
+
+  //Tiers
+
+  OUTLAW_ROGUE_TIER_28_2P_SET_BONUS: {
```
ToppleTheNun, line 584: I would recommend renaming this to `OUTLAW_ROGUE_T28_2P_PROC` or something similar rather than `_BONUS`, but that's nitpicky and not for this PR.

PR #5851, opened 2023-02-25
Fix Cast Efficiency key generation

emallson:
Fixes the large error dumped in the console by React for the CastEfficiency component.

This still occurs in a few other components (notably `ShuffleOverview` for brewmaster...which then immediately disappears, so likely related to the initial state with nothing present but I'm not fixing that one right now)


ToppleTheNun (approved):

jazminite (approved):
👍

PR #5864, opened 2023-03-02
[OUTLAW] Guide APL Checker

Zac-oihfdwrrtuinvbcp:
### Description

This is a first pass at an APL checker for outlaw, seems to be working fine for most of the basic stuffs but I do have some interrogations regarding some conditions that I cannot comprehend and get working.

1 - Every 'optional' condition seems to break everything, very simple example:
image
This is returning true whenever snc buff is down, this seems to be working fine in this sample example:
image
We can see the user have no snc between 11 and 17s and is prompted to reroll, now if I un comment the `optional(buffMissing(SPELLS.BURIED_TREASURE)),` condition, as I understood it it should prompt the player to roll when snc and bt buffs are down but also allow him to keep bt alone same as it is doing with snc, however:
image
Suddenly even with both buffs down the user casts are not flagged as mistakes.

2 - Second thing I cannot figure out is how to use the pandemic option with the 'buffMissing' condition:
Here is my very simple BTE condition 
image
I want the user to have to refresh bte if the debuff is down but also be allowed to refresh up to 4s before (to not drop the debuff), bte is not a skill with actual pandemic and its duration is variable since its a finisher
image
Here the user cast bte with max cp at 2s, he will apply the debuff for 21s, however he is prompted to refresh it as soon as the spell cd is back up

I couldn't understand exactly how the `debuffMissing` function was working especially since it doesn't have a 'RefreshDebuff' in it which would make me think it wouldn't handle refreshes? I also couldn't figure out what the `buffDuration` calculation was really doing.

Thanks in advance for the help.

### Testing

- Test report URL: `[/report/...](https://www.warcraftlogs.com/reports/t27zV8NLgPvJq3Kk#fight=17&type=casts&source=18&start=2483411&end=2526210)`
- Screenshot(s):
image


emallson:
- The `optional` thing looks like a bug
- I don't think `debuffMissing` has support for variable duration debuffs but could look at adding that :thinking: not exactly sure how to tie that in for the general case, might be better to have it be a custom condition for rogue stuff. Feral druid and evoker also have variable duration DoTs afaik. will have to look at it

Zac-oihfdwrrtuinvbcp:
> * I don't think `debuffMissing` has support for variable duration debuffs but could look at adding that 🤔 not exactly sure how to tie that in for the general case, might be better to have it be a custom condition for rogue stuff. Feral druid and evoker also have variable duration DoTs afaik. will have to look at it

Shouldn't the function at least work without a pandemic option, with simply `debuffMissing(SPELLS.BETWEEN_THE_EYES),`, glancing at the function I would think it would since it seems to just be checking if the debuff was present with apply and removeDebuff events
image

but I still have the same issue

Zac-oihfdwrrtuinvbcp:
I looked a bit more into it and correct me if I'm wrong but I think there is a mistake in the debuffMissing function, 
image
When the `encodedTargetString` get created it currently set the second part of the string as 1 if no `targetInstance` is found, after checking what the `encodeEventTargetString` was returning it appears that the default `instance` is set to 0 and not 1, 
image

emallson:
That does look like a bug with it, yup

Zac-oihfdwrrtuinvbcp:
After looking a bit more into it I think there is more issues with this function:

-  it doesn't have a `RefreshDebuff` case contrary to the `buffMissing` function which seems to be built the same way,
- the `timeRemaining` which is being sent to the `buffDuration` function seems to also be incorrect to me

image

The `buffDuration` function also seems to have some issues? Unless I am not properly understanding its purpose the calculation seems off to me, I am unsure what the pandemicCap original intended way of use was, the only very few examples of it being used seemed to be unsure aswell (from the brewmaster apl):
image

image

If issues there is they would also be present in the `BuffMissing` function altough this only impact a few modules (only hunters and brewmaster modules are using it)

emallson:
Okay, I'm going to work on this this weekend. The brew thing is probably a
typo because I was in a bit of a hurry doing the last round of rotation
updates but as a whole this is not a great state for such a core condition
to be in

emallson:
@Zac-oihfdwrrtuinvbcp I went through this afternoon and fixed *several* bugs with the buffMissing/debuffMissing implementation. Only one of them impacted brewmaster (`buffDuration` was not calculating partial pandemic extensions at all, always full extensions), which is part of why they went uncaught.

To remedy that, there are tests for both conditions now that should help ensure they behave going forward.

emallson:
I also looked into the behavior of `optional`. It isn't technically a bug, but the interaction with `and` is pretty confusing.

tl;dr - `optional` functions by checking if the spell being cast matches the one for the rule. if so, it checks the condition. if not, it says "this condition is not met". when combined with `and`, this means that any branch containing optional makes the whole thing optional.

I have done a couple of things to make this less of a footgun:

1. renamed `optional` -> `optionalRule` to indicate that it is for the *whole rule*
2. added a dev-time warning to `and` that triggers if any of the nested conditions use `optionalRule`.

To get the behavior you describe, you'll want to combine `and` and `or` with `optionalRule` (`or` doesn't have the same infectious behavior with nested `optionalRule`---though it still isn't super clear behavior):

```ts
or(
  // allow rerolling if you're missing either SnC or BT, but don't require it
  optionalRule(
    or(
      buffMissing(SPELLS.SKULL_AND_CROSSBONES),
      buffMissing(SPELLS.BURIED_TREASURE),
    )
  ),
  // require rerolling if you're missing both SnC and BT
  and(
    buffMissing(SPELLS.SKULL_AND_CROSSBONES),
    buffMissing(SPELLS.BURIED_TREASURE),
  )
)
```

this is the same *logically* as having *two rules*: one that is optional for rerolling with either but not both buffs present, and one that is not optional for rerolling with neither present. the only real difference is how UI components will treat it.


```ts
{
  spell: rtb,
  // allow rerolling if you're missing either SnC or BT, but don't require it
  condition: optionalRule(
    or(
      buffMissing(SPELLS.SKULL_AND_CROSSBONES),
      buffMissing(SPELLS.BURIED_TREASURE),
    )
  ),
},
{
  spell: rtb,
  // require rerolling if you're missing both SnC and BT
  and(
    buffMissing(SPELLS.SKULL_AND_CROSSBONES),
    buffMissing(SPELLS.BURIED_TREASURE),
  )
}
```

I think the UI is *likely* to be more what you want if you put both in a containing `or`, and as long as it is commented appropriately I don't think it will cause further confusion.

Zac-oihfdwrrtuinvbcp:
Thanks a lot for all of this, I thought I would have to do something like that for the `optional` conditions after looking a bit more into the `and` but my brain couldn't really figure it out, your explanations and examples will greatly help me though, thanks again

Zac-oihfdwrrtuinvbcp:
Should have updated most of it, some things are still a bit inaccurate but It should be a decent first pass, would really appreciate a sanity check on the `optionalRule` conditions logic just in case my brain failed me. I went over casts from the timeline and the flagged casts seemed to be mostly accurate.

src/analysis/retail/rogue/outlaw/modules/apl/buffsCount.tsx
```diff
@@ -0,0 +1,60 @@
...
+  latencyOffset: number = 0,
+): Condition<number | null> {
+  return {
+    key: `buffsCount-${spells.map((spell) => spell.id).join('-')}-${count}`,
+    init: () => 0,
+    update: (state, event) => {
+      switch (event.type) {
+        case EventType.ApplyBuff:
+          if (spells.some((spell) => spell.id === event.ability.guid)) {
+            return state !== null ? state + 1 : 0;
```
emallson: small bug, this is going to count the first buff application as 0 buffs.

simpler way:

```suggestion
            return (state ?? 0) + 1;
```

src/analysis/retail/rogue/outlaw/modules/apl/buffsCount.tsx
```diff
@@ -0,0 +1,60 @@
...
+          }
+          break;
+        case EventType.RemoveBuff:
+          if (spells.some((spell) => spell.id === event.ability.guid)) {
+            return state !== null ? state - 1 : 0;
```
emallson: this one shouldn't need to change since you don't want to end up at -1 buffs.

src/analysis/retail/rogue/outlaw/CONFIG.tsx
```diff
@@ -31,8 +31,8 @@ export default {
   ),
   // A recent example report to see interesting parts of the spec. Will be shown on the homepage.
   exampleReport:
-    '/report/wptPT3mfWavbj9KY/33-Heroic+Conclave+of+the+Chosen+-+Kill+(6:22)/13-Kuracz',
-
+    'https://www.warcraftlogs.com/reports/PVf9yQjhGDzZaBmv#fight=18&type=damage-done&source=11',
```
emallson: this should be a link to a wowa report, not a WCL report.

you can take the bit starting at `/report` from your local when testing

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,268 @@
...
+const rtbCondition = () => {
+  const rtbBuffsToCheck = ROLL_THE_BONES_BUFFS.filter((spell) => spell !== SPELLS.GRAND_MELEE);
+  return and(
+    describe(
+      or(
+        // allow rerolling if you're missing either SnC or BT, but don't require it
+        optionalRule(or(buffMissing(SPELLS.SKULL_AND_CROSSBONES), buffMissing(SPELLS.BROADSIDE))),
+        // require rerolling if you're missing both SnC and BT
+        and(buffMissing(SPELLS.SKULL_AND_CROSSBONES), buffMissing(SPELLS.BROADSIDE)),
+      ),
```
emallson, line 65: i *think* this should work because the `or` will prevent the infectious interaction of `optionalRule` with `and`? something to keep an eye on.

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,268 @@
...
+  {
+    spell: TALENTS.SHADOW_DANCE_TALENT,
+    condition: describe(
+      and(
+        buffMissing(SPELLS.AUDACITY_TALENT_BUFF),
+        buffMissing(SPELLS.OPPORTUNITY),
+        //This is a given no point displaying it
+        notInStealthCondition(),
+        //We want to allow the user to press dance at max cp, but that is not a requirement
+        or(not(hasFinisherCondition()), optionalRule(hasFinisherCondition())),
```
emallson, line 142: probably also works, i think?

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,268 @@
...
+  {
+    spell: SPELLS.SLICE_AND_DICE,
+    condition: and(
+      hasFinisherCondition(),
+      buffMissing(SPELLS.SLICE_AND_DICE, {
+        timeRemaining: 18000,
+        //Since SnD as a variable duration depending on cp spent this is inacurate for now
+        duration: 45000,
+        pandemicCap: 1.3,
+      }),
```
emallson, line 186: this probably needs to have a custom condition written for variable duration buffs, but i don't think that is a blocker.

emallson (changes requested):
some minor changes then we should be good to merge

emallson (approved):

Zac-oihfdwrrtuinvbcp:
Just wanted to add a small disclaimer about spells/talents missing to not potentially mislead users.

PR #5872, opened 2023-03-05
[Restoration Shaman] Redo Unleash Life

vohrr:
### Description

Rewrote Unleash Life using event linking for the underlying interactions between the buff and the affected spells

### Motivation

The current iteration of this module was inaccurate and only partially implemented. It is now much more accurate and analyzes all spells that the talent affects
### Testing
`/report/H9VDyfZ4FKABNtGk/36-Mythic+Kurog+Grimtotem+-+Kill+(8:49)/Fakeblonde/standard/statistics` - most common
`/report/PdXj1MctQk63fAWZ/16-Mythic+Terros+-+Kill+(5:34)/Dalfaëran/standard/statistics`  - all but healing surge
`report/c8BjmPGdf2gHQAVM/27-Mythic+Terros+-+Kill+(6:41)/17-Bigdaddyjeff/standard/statistics` - healing surge

### Screenshot(s):
Before
image

After
image

PR #5874, opened 2023-03-05
Destro DF & 10.0.5 updates

dsnam:
### Description

Hopefully this doesn't count as too large a PR, if it does lemme know and I will split it up into individual spell/talent change PRs.

This PR updates some Destro talent/spell modules to take into account new and updated mechanics in DF:
1. Havoc module also works with Mayhem now and shows Havoc uptime when Mayhem is talented.
2. Added module to track Burn to Ashes-buffed Incinerates
3. Updated Madness of the Azj'aqir to also track buffed Rains of Fire and Shadowburns, which is new in DF compared to old legendary version which only buffed Chaos Bolt

I also made some assorted changes/cleanup:
1. Removed old Roaring Blaze DoT component code (hasn't been a DoT since Legion I think?)
2. Updated a few modules' statistics to use `TalentSpellText` instead of `BoringSpellValueText` since they are now talents

### Motivation

Noticed destro (and all warlock) shows an unmaintained/out of date warning when you use it so figured I would see what needs updating.

### Testing

- Has madness buffing multiple spells: `report/F2BYTjRKA7dDgLxr/4-Mythic++Halls+of+Valor+-+Kill+(33:35)/Shodam/standard/statistics`
- Player has Mayhem talented: `report/3wBraYxZ6Ty8RzJG/46-Mythic+Kurog+Grimtotem+-+Kill+(5:24)/141-Fisor/standard`
- Screenshot(s):
Mayhem (with uptime) and Burn to Ashes statistic visible here:
image
Madness tracking buffs other than Chaos Bolt:
image


src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
```diff
@@ -19,21 +22,48 @@ class Havoc extends Analyzer {
   protected enemies!: Enemies;
 
   damage = 0;
+  talent: Talent = TALENTS.HAVOC_TALENT;
+
+  HAVOCABLE_ABILITY_IDS = [
```
trevorm4: define this at a const global var, no need for it to be a member var

src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
```diff
@@ -54,12 +84,19 @@ class Havoc extends Analyzer {
           </>
         }
       >
-        <BoringSpellValueText spellId={SPELLS.HAVOC.id}>
+        <TalentSpellText talent={this.talent}>
           {formatNumber(this.dps)} DPS{' '}
```
trevorm4: Use `<ItemDamageDone>`

trevorm4 (changes requested):
Fairly certain you didn't run the formatter on this code either, make sure to do yarn install and it should run es-lint prior to git comitting (if this doesnt happen, try deleting node_modules and doing a fresh `yarn install`)

src/analysis/retail/warlock/destruction/modules/talents/Eradication.tsx
```diff
@@ -10,9 +10,10 @@ import { calculateEffectiveDamage } from 'parser/core/EventCalculateLib';
...
 import { When, NumberThreshold, ThresholdStyle } from 'parser/core/ParseResults';
 import Enemies from 'parser/shared/modules/Enemies';
-import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
 import Statistic from 'parser/ui/Statistic';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import TalentSpellText from 'parser/ui/TalentSpellText';
+
 
 const MAX_TRAVEL_TIME = 3000; // Chaos Bolt being the slowest, takes around 2 seconds to land from max range, added a second to account for maybe target movement?
 const ERADICATION_DAMAGE_BONUS = 0.1;
```
vohrr: This talent is a 2-point talent for 5% and 10%, this constant should be defined as .05 and then have a var defined as `ERADICATION_DAMAGE_BONUS * this.selectedCombatant.getTalentRank(TALENTS.ERADICATION_TALENT)`

src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
```diff
@@ -19,21 +22,48 @@ class Havoc extends Analyzer {
...
 
   constructor(options: Options) {
     super(options);
-    this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDamage);
+    if (this.selectedCombatant.hasTalent(TALENTS.MAYHEM_TALENT)) {
+      this.talent = TALENTS.MAYHEM_TALENT;
+    }
+    this.active = this.selectedCombatant.hasTalent(TALENTS.HAVOC_TALENT) || this.selectedCombatant.hasTalent(TALENTS.MAYHEM_TALENT);
+
+    if (this.active) {
```
vohrr: Unnecessary if statement

src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
```diff
@@ -19,21 +22,48 @@ class Havoc extends Analyzer {
...
+      this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDamage);
+    }
   }
 
   onDamage(event: DamageEvent) {
     const enemy = this.enemies.getEntity(event);
```
vohrr, line 49: If this function is only supposed to look at your havocable abilities, then consider using your HAVOCABLE_ABILITY_IDS const differently. Define it as a spell array that you can add to the event listener (essentially just remove the .map()): `this.addEventListener(Events.damage.by(SELECTED_PLAYER).spell([HAVOCABLE_ABILITIES]), this.onDamage);`
map your spellIds later on if you still need them

Havoc would probably be better handled using event linking so that you could easily filter out the hardcasted damage  vs the havoc damage to a target, probably out of scope for this PR though

vohrr (changes requested):

src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
dsnam: ah thanks for this tip, I didn't check if I could pass a list of spells for damage events. What can I look at to learn about event linking? I would like to revisit the wl spec modules after I do a pass on each one to make sure at least what already exists is updated for DF mechanics and add new things or tighten up calculations like this, so yeah I wouldn't edit it in this PR but I'd queue it up for a later one.

vohrr: There are some good examples in the Mistweaver Monk and Preservation Evoker files, filenames - CastLinkNormalizer.ts
Pres might be better to look at since most of the monk ones are used for hot tracking attribution.
My current PR for RSham https://github.com/WoWAnalyzer/WoWAnalyzer/pull/5872 also has a lot of event linking implementation in it as well

dsnam:
> Fairly certain you didn't run the formatter on this code either, make sure to do yarn install and it should run es-lint prior to git comitting (if this doesnt happen, try deleting node_modules and doing a fresh `yarn install`)

@trevorm4 You are right, I jumped straight into testing with the docker compose so I missed that there was an autoformatter to set up. I also set things up on my host machine so that the git hook works now and I manually ran prettier on what I had already edited, hopefully that makes them consistent again.

src/analysis/retail/warlock/destruction/modules/talents/Havoc.tsx
dsnam: Thanks, I will have a look. In the meantime I have swapped to passing the ability list into the listener as described.

vohrr (approved):

PR #5889, opened 2023-03-10
Affliction 10.0.5

dsnam:
### Description

- updated Soul Conduit for all specs to take into account talent rank scaling, and updated Affliction's module to use rapture as the example spender instead of UA.
- added some hasTalent checks for Darkglare so certain components don't try to render when you don't have it talented
- fixed Vile Taint not checking for the right debuff id
- updated Shadow Embrace and Nightfall to take into account stacking and talent scaling
- Added modules for Wrath of Consumption, Inevitable Demise, Pandemic Invocation, and Tormented Crescendo

### Testing

- Test report URL: 
Log showing Wrath and Vile Taint: `report/4TZ7hgKVbw2jk8Pz/30-Mythic+Sennarth,+The+Cold+Breath+-+Kill+(6:41)/Omenity/standard/statistics`
Log showing Inevitable Demise, Pandemic Invocation, Tormented Crescendo: `report/z3qaHpWNC4JKVYjf/18-Mythic+Terros+-+Kill+(5:34)/Dodse/standard/statistics`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/warlock/affliction/modules/spells/InevitableDemise.tsx
+++ b/src/analysis/retail/warlock/affliction/modules/spells/InevitableDemise.tsx
@@ -1,0 +1,75 @@
+import SPELLS from 'common/SPELLS';
+import TALENTS from 'common/TALENTS/warlock';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, {
+  ApplyBuffEvent,
+  ApplyBuffStackEvent,
+  RefreshBuffEvent,
+  RemoveBuffEvent,
+} from 'parser/core/Events';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import TalentSpellText from 'parser/ui/TalentSpellText';
+
+const BUFF_DURATION = 20000;
+const BUFFER = 100;
+const MAX_STACKS = 50;
+
+class InevitableDemise extends Analyzer {
+  wastedStacks = 0;
+  stacks = 0;
+  private buffApplyTimestamp: number | null = null;
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS.INEVITABLE_DEMISE_TALENT);
+    this.addEventListener(
+      Events.applybuff.by(SELECTED_PLAYER).spell(SPELLS.INEVITABLE_DEMISE_BUFF),
+      this.onInevitableDemiseApplyRefresh,
+    );
+    this.addEventListener(
+      Events.refreshbuff.by(SELECTED_PLAYER).spell(SPELLS.INEVITABLE_DEMISE_BUFF),
+      this.onInevitableDemiseApplyRefresh,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.INEVITABLE_DEMISE_BUFF),
+      this.onInevitableDemiseRemove,
+    );
+    this.addEventListener(
+      Events.applybuffstack.by(SELECTED_PLAYER).spell(SPELLS.INEVITABLE_DEMISE_BUFF),
+      this.onInevitableDemiseApplyStack,
+    );
+  }
+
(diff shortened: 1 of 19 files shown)


src/analysis/retail/warlock/affliction/modules/spells/SoulConduit.tsx
```diff
@@ -4,14 +4,13 @@ import TALENTS from 'common/TALENTS/warlock';
...
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import TalentSpellText from 'parser/ui/TalentSpellText';
 
 import SoulShardTracker from '../resources/SoulShardTracker';
 
-const TICKS_PER_UA = 4;
-const SC_PROC_CHANCE = 0.15;
+const SC_PROC_CHANCE_BASE = 0.05;
 
 class SoulConduit extends Analyzer {
```
dsnam, line 15: now that this in the class tree it seems like these should be merged and I could just set what "you were refunded enough shards for x more casts of y spender" message shows based on the combatant spec, but then it'd just be one file with a lot going on. not sure which way of organizing it is preferable.

src/analysis/retail/warlock/affliction/modules/spells/InevitableDemise.tsx
```diff
@@ -0,0 +1,75 @@
...
+      this.onInevitableDemiseApplyStack,
+    );
+  }
+
+  onInevitableDemiseApplyRefresh(event: ApplyBuffEvent | RefreshBuffEvent) {
+    if (this.stacks === MAX_STACKS) {
+      this.wastedStacks += 1;
+    }
+    this.buffApplyTimestamp = event.timestamp;
+  }
```
dsnam, line 49: unlike some of the other buffs with stacks I dealt with in this PR like Tormented Crescendo I noticed the event log for Inevitable Demise has a ton of refresh buff events, basically one for every apply/applystack as well. do some buffs just tend to spam the combat log or is that a relic of this one being older?

emallson: some buffs do that, yea

src/analysis/retail/warlock/affliction/modules/spells/SoulConduit.tsx
emallson: if its easy, generalizing is good. however, i'd rather have 3 simple modules (which can import shared constants) than 1 complicated module that tries to do everything.

src/analysis/retail/warlock/affliction/modules/spells/ShadowEmbrace.tsx
```diff
@@ -30,6 +30,9 @@ class ShadowEmbrace extends Analyzer {
   };
   protected enemies!: Enemies;
 
+  BONUS_PER_STACK =
+    BONUS_PER_STACK_BASE * this.selectedCombatant.getTalentRank(TALENTS.SHADOW_EMBRACE_TALENT);
```
emallson, line 34: did not know that you could use `this` in property initializers outside the constructor. neat

emallson (approved):
this all looks pretty reasonable, assuming you're not double-counting stacks due to the `refreshbuff` thing you noted.

dsnam:
@emallson I didn't notice that last comment after I saw it got approved so in case you were waiting on an answer to that before merging: yeah I am pretty sure it's handling the stacks accurately, or at least as accurately as the stack count on the applystack event lets it be. I spot checked it with some logs and the overcapped stack metric was the same as the number of refresh events that came through while at max stacks.

On that note though, is there an existing framework to add tests that are like "run this log or fake event sequence and check that x module's metric has y value," or would something like that be too brittle? I see some modules appear to have something like that but the files are quite old and I can't tell if the code's dead. If it's not I'd be happy to add some for this.

emallson:
> On that note though, is there an existing framework to add tests that are like "run this log or fake event sequence and check that x module's metric has y value," or would something like that be too brittle? I see some modules appear to have something like that but the files are quite old and I can't tell if the code's dead. If it's not I'd be happy to add some for this.

we have some test tools, but with the dragonflight revamp a lot of the old test code got killed. it is possible (but a bit of a pain) to spin up analyzers to run events through, and we have *some* infra to download log data to run against, but it frankly needs a bit of a refresh

PR #5892, opened 2023-03-12
[Prot Warrior] First once over

abelito75:
### Description

Just a quick once over of everything. Make sure its roughly working, file locations, etc etc


ToppleTheNun (approved):
Everything looks good. Is there a specific reason for moving talent analyzers into the spells directory? Most specs have adopted a pattern of splitting core spells into the spells directory and talents into the talents directory. You don't necessarily need to adopt the same pattern; mostly curious.

abelito75:
To me, personally, the idea of talents has kind of faded away with this talent tree rework
So I don't see a reason of separating them

I'm debating an idea of making folders for "spell groups" where a set of talents/spells are all working on the same ideas 
IE things that interact with shield slam

PR #5897, opened 2023-03-12
[Outlaw] Core rotation section with finisher subsection

Zac-oihfdwrrtuinvbcp:
### Description

First subsection of the core rotation section that tracks finisher usage.

Getting this error since I added this new subsection, not exactly sure what causes this
image
Also very rarely after making a change? (not exactly sure when it actually happens, haven't been able to reproduce the issue) the boxes aren't all getting rendered, not sure if that would be linked to this error

### Testing

**Without Greenskin wickers talent:** `https://www.warcraftlogs.com/reports/kGZzydbhQrpMDLvw#fight=3&type=damage-done&source=16`
image

**With Greenskin wickers talent:** `https://www.warcraftlogs.com/reports/ADQxzVR9NvLt8rc4#fight=13&type=damage-done&source=111`
image

### DIFF
--- a/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
+++ b/src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
@@ -127,12 +127,6 @@ export default class FinisherUse extends Analyzer {
       event as TargettedEvent<any>,
     );
     const bteCDRemainingTime = this.spellUsable.cooldownRemaining(SPELLS.BETWEEN_THE_EYES.id);
-    const hasSnD = this.selectedCombatant.getBuff(
-      SPELLS.SLICE_AND_DICE.id,
-      event.timestamp,
-      0,
-      200,
-    );
 
     if (this.hasGreenskinTalent) {
       checklistFinisherChoice.summary = (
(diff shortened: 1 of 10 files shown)


emallson:
the `:nth-child` error is just a warning for SSR, which we don't use. it can be safely ignored.

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -49,28 +78,318 @@ export default class FinisherUse extends Analyzer {
...
+                <SpellLink id={SPELLS.BETWEEN_THE_EYES.id} /> debuff.
+              </div>
+            );
+          }
+        }
+        break;
+      default:
+        checklistFinisherChoice.details = <div>{checklistFinisherChoice.details}.</div>;
+        break;
+    }
```
emallson: this should probably be split out into several methods. really big switch bodies are a source of pretty sneaky bugs related to missing `break`s

Zac-oihfdwrrtuinvbcp:
Hey @emallson. Sorry to bother, I was just wondering if you had a chance to take a look at this PR? No pressure, I understand if you didn't have the time. I just wanted to check in and make sure it didn't get lost or forgotten. Also, no rush on this, I don't have any other PRs coming up just yet anyway. Thanks

emallson:
i have an in-progress review that i forgot to finish. i'll wrap that up :)

emallson (changes requested):

emallson:
I am not 100% sold on using the box grids for these kinds of very frequent spells. It works (and there is something nice about seeing a sea of green) but there may be better options.

but, we don't need perfection to put things online. it just needs to be good, which i think is the case here

Zac-oihfdwrrtuinvbcp:
> this should probably be split out into several methods. really big switch bodies are a source of pretty sneaky bugs related to missing `break`s

So I can do it like that, just splitting what's in each switch block into a method, it feels a bit silly tho, not sure exactly if that would be the correct approach, I especially don't like that I'm copy pasting some variables like `bteCDRemainingTime` and `bteDebuffRemainingTime` into the methods but I'm not sure if sending them as parameters would make much sense either

To be entirely fair I have rewritten this part of the code a bunch of times trying to make it easy to understand (at first it was just a huge if else statement), it still is kind of a mess but I don't really have a solution so feel free to suggest.

emallson (approved):
i think this is good, the main issue with large `switch` statements is accidentally falling through, so a big if/else is also a choice there. small switches are easier to work with and avoid the sneaky accidental fallthrough

PR #5899, opened 2023-03-12
Havoc Guide updates

ToppleTheNun:
### Description

Add Rotation section to Havoc Guide, include Furious Gaze on Eye Beam breakdown.

### Motivation

A few changes were requested to help make the Havoc Guide more useful.

### Testing

- Test report URL: `/report/6g4TqjXz9BH2ywDk/1-Mythic+Eranog+-+Kill+(2:15)/Hotbees/standard`
- Screenshot(s):
image
image


emallson:
IMHO the sidebars here have too much text.

- Throw Glaive has a lot of repeating what the talents you're taking do. This can be trimmed down to something like "Throw Glaive is buffed by many talents in the spec tree, and is only worth casting if you take at least Soulrend and Furious Throws." Could optionally add a list of talents as spell links that buffed Throw Glaive to drive that point home.
- Essence Break's issue is mostly spelling out the cast sequence for the EB window. Would like to have a better pattern for solving this (Brew needs it now for EK, 2/3 mage specs want it, etc). For now I would probably make them collapsed by default that you can expand to view the ideal sequence.

ToppleTheNun:
Agreed, especially seeing how few checks there actually are for Essence Break and how little data there really is to show for Throw Glaive. I'll revisit how they're displayed.

ToppleTheNun:
Updated Throw Glaive section:
image

Updated Essence Break section:
image

Updated Essence Break with both collapsed views open:
image

emallson (approved):

PR #5901, opened 2023-03-14
[Core] Anal typing on cooldown tracking

abelito75:
### Description

I made cooldown tracker more anal when it comes to typing so its harder for a developer to shoot themselves in the foot
### Motivation

It was broken
### Testing
I opened every single spec that had a default log and made sure the cooldown tracker was actually working
Note: some default logs are broken and actually older than me. We need to update that next.


emallson (commented):
this is going to fix the type error. i wasn't totally following the cd tracker conversation, but i guess the question i have is: does this api make sense?

i was the one that initially asked for the expansion flag in order to prevent certain classic spells (which were being added to the shared tracker to avoid repeating them in every classic spec) from showing for retail. e.g. bloodlust.

for a single spec implementation, we already know the expansion---why do we need to repeat it?

basically: i think this API is something that is specifically for the core code and rather than infecting *every* class module with the manual expansion spec, we should change how the core code handles these so that it correctly treats "expansion unset" as the default case

abelito75:
I talked about it with someone in [discord](https://discord.com/channels/316864121536512000/316864385052049420/1084315457051426850) and the suggestion there was to make the property required.
I plan to change a lot of this class in the near™ future which might include breaking classic and retail into children for this very reason 
IE
AbstractCooldownThroughtputTracker
RetrailCooldownThroughtputTracker extends AbstractCooldownThroughtputTracker
ClassicCooldownThroughtputTracker extends AbstractCooldownThroughtputTracker

This way if we need custom logic specific to one or the other we don't add new variables and possible bugs to the other one
Which also solves the whole expansion issue

TLDR: 
I think its annoying right now but a quick stop gap to get everything working while I rework `CooldownThroughputTracker`

emallson:
> AbstractCooldownThroughtputTracker
RetrailCooldownThroughtputTracker extends AbstractCooldownThroughtputTracker
ClassicCooldownThroughtputTracker extends AbstractCooldownThroughtputTracker

I don't think this kind of structure is correct *because* so little changes from retail to classic. The `CooldownThroughputTracker` should fundamentally be doing the same thing for both expansions, the difference is primarily the spell list. Splitting like this risks a different class of bugs (i.e. someone fixes a bug in the Retail copy but not the Classic one).

It is perhaps best to make the spell list a simple getter? and then its:

```ts
if (retail) {
  return retail spells;
} else {
  return classic spells;
}
```

and inheritors override it as:

```ts
return [
   ...super.spells,
   custom spells,
];
```

abelito75:
In all reality I didn't expect there to be many (if any) changes between retail and classic but thought it was a clean way to divide them... but that said a simple if is FAR more effective and readable

abelito75:
With that being said... Do we want to merge the stop gap in so all specs at least work
Then I go bang my head against the change to clean it up to a more logical format?

emallson:
If this fixes breakage then yes we go ahead and merge rn

PR #5925, opened 2023-03-19
add "Hide Good Casts" toggle

ToppleTheNun:
### Description

Add a "Hide Good Casts" toggle.

### Motivation

emallson commented the below on #5897, which got me thinking about ways to still give the dopamine hit of seeing a sea of green when you do well, but also allow focus on just things that you can improve.
> I am not 100% sold on using the box grids for these kinds of very frequent spells. It works (and there is something nice about seeing a sea of green) but there may be better options.

### Testing

- Test report URL: `/report/r4JtW9hxdzwc3G6q/8-Mythic+The+Primal+Council+-+Kill+(3:00)/22-Toppledh/standard`, `/report/Fdfk6LPncg17BHAr/13-Heroic+Raszageth+the+Storm-Eater+-+Kill+(9:24)/Xanq/standard/overview`
- Screenshot(s):
hide-good-casts

### DIFF
--- a/src/analysis/retail/demonhunter/havoc/Guide.tsx
+++ b/src/analysis/retail/demonhunter/havoc/Guide.tsx
@@ -16,6 +16,7 @@ import {
   PERFECT_TIME_AT_FURY_CAP,
 } from './modules/resourcetracker/FuryTracker';
 import FuryCapWaste from './guide/FuryCapWaste';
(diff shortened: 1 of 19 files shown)


ToppleTheNun:
Will be coming back to enhance this further at another point in time.

PR #5932, opened 2023-03-19
Radiant spark analysis

cameronstubber:
### Description

Updates the analysis of Radiant Spark to be more accurate towards Dragonflight. How the talent and its follow on Harmonic Echo are used seem to be different from in Shadowlands.

Arcane follows a very rigid spell cast cycle for burn phases which is more what is being checked now.

### Motivation

To improve the spec.

### Testing

Look at some Arcane Logs and see suggestions being made for Radiant Spark, about using the 4 spells and standing in the rune of power.


cameronstubber:
If you want me to do it again with a fixed branch so the conflict isnt here just reject this and I will do it again.

src/analysis/retail/mage/arcane/talents/RadiantSpark.tsx
```diff
@@ -1,79 +1,136 @@
...
-          'Not enough (or too many) Arcane Blast casts during Radiant Spark - Casted ' +
+          'You didnt cast 4 Arcane Blasts during the Radiant Spark Ramp. Casted ' +
             this.arcaneBlastCasts +
-            ', expected ' +
-            spellCountThreshold,
+            ', expected 4',
         );
+
+      wasBadRadiantSpark = true;
+      this.badRadiantSparkBecauseDidntUseFourSpells += 1;
```
trevorm4, line 126: Ignoring logging, this could be combined into 1 if statement by saying `if(this.arcaneSurgeCasts + this.arcaneBlastCasts < 4) ... `

trevorm4 (approved):

src/analysis/retail/mage/arcane/talents/RadiantSpark.tsx
trevorm4: Resolve the merge conflicts and potentially change this and then I will merge

cameronstubber: Will fix the conflicts and come back with a new commit.

Regarding the combine the if statement into surge + blast < 4 - I had taken the decision to do this in 2 steps because I wanted the code to be readable to a future developer in a way that it should clearly match the rules I have described and am trying to meet. So it would be like

- Rule 1 - 1 Surge and 3 Blasts - ok this code block is for this rule.
- Rule 2 - 4 Blasts - the next block is for this rule.

If you still want me to change it, I will. Maybe in future modules I can come up with a better 'rule checker' structure instead of chained if/elses.

PR #5957, opened 2023-03-25
Arcane Mage - Arcane power rename to Arcane Surge

### LINKED ISSUE #5930: [Arcane Mage] - Arcane Surge is called Arcane Power in many places
I didnt play a mage in Shadowlands but my assumption is this is an after-effect of migrating the spec to Dragonflight.

To begin with we should simply ensure that where-ever we are talking about Arcane Surge in the analysis, we rename everything away from Arcane Power.

We need to rebuild specific analysis for Arcane Surge windows but that is a separate development task as opposed to an issue.

image

image

cameronstubber:
### Description

fixes #5930

Renamed files in the codebase from Arcane Power to Arcane Surge, where their internal references to all talent and spell id's were swapped to Arcane Surge as part of Dragonflight support.

This Issue is not to fix the recommendations made, simply to move whats already implemented in-line with the correct wording. I will fix the recommendations as future PRs.

### Motivation

To prepare for building new analysis tools using Arcane Surge and Burn Phase tracking.

### Testing

I have done generic tasks for testing as this is just a codebase change:

- Checked dev server console for errors.
- Checked Chrome Console for errors
- Loaded Arcane Mage log reports for Retail and clicked through all tabs/pages. Reading text which used to say Arcane Power and looking that it now says Arcane Surge
- Loaded Arcane Mage classic logs and clicked through all tabs (looking to see if I had crashed anything in the core code not realising).


cameronstubber:
I don't want to touch the translation files as I can see some references in Frost Mage which I know nothing about. I also don't want to break anything in the core files if I can avoid it. I think this is sufficient for moving forward to Arcane Surge analysis module development.

emallson (approved):
this is *prooobably* fine. would just poke @Sharrq if you haven't already in discord to see if they have any arcane stuff in the works. iirc they've done more frost/fire tho

Sharrq:
I have nothing in the works for arcane. It's the one spec I don't play so it's historically been the hardest for me to maintain. These all look like pretty simple name changes, and aren't expected to fix the analysis so I'm cool with it.

I've been trying to keep an eye on the arcane changes here too make sure they are good, but forgot to reply/approve

PR #5964, opened 2023-03-30
Improve accuracy of veil of pride module

trevorm4:
Before we blindly took half of all Sheilun's healing but in reality a majority of the overhealing is going to come from the veil of pride stacks. This PR tries to attribute overhealing/healing to non-pride vs pride stacks.

Before
image
After
image
image

In this log, roughly 71% of the overhealing was from extra veil stacks so it makes sense that veil is not directly half of sheilun's healing

Also updated visuals of legacy to have precision and icon for alignment
image

report/1qcwTk7dBmz8LQF6/14-Mythic+Broodkeeper+Diurna+-+Kill+(8:17)/Livingmist/standard/statistics

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/VeilOfPride.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/VeilOfPride.tsx
@@ -12,8 +12,6 @@ import SheilunsGift from './SheilunsGift';
 import Events, { CastEvent, HealEvent } from 'parser/core/Events';
 import { calculateEffectiveHealing, calculateOverhealing } from 'parser/core/EventCalculateLib';
 
-const MAX_STACKS = 10;
-
 class VeilOfPride extends Analyzer {
   static dependencies = {
     sheilunsGift: SheilunsGift,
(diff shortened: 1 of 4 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/VeilOfPride.tsx
```diff
@@ -1,27 +1,88 @@
...
+    }
+    const timeDiff = timestamp - this.lastCast;
+    const nonVeilStacks = Math.floor(timeDiff / SECONDS_PER_CLOUD);
+    const rawVeilStacks = Math.floor(timeDiff / VEIL_SECONDS_PER_CLOUD);
+    return this.getExtraStacks(nonVeilStacks, rawVeilStacks);
+  }
+
+  onHeal(event: HealEvent) {
+    // double clouds = 100% increase -> 2x / x - 1 = 1
+    const baseStacks = this.stacksOnCast - this.extraStacksOnCast;
```
vohrr: The logic in getExtraClouds and getExtraStacks might be overcomplicating this. Veil increases the rate of cloud generation by 100%, so its fair to say that its a constant 100% effective increase for our valid range of clouds regardless of what the exact timestamp is, since this can evaluate to an overly high effective increase. 
image

vohrr: The effective increase call itself should be fine for correcting the overvalued doubling of hps

vohrr (approved):

PR #5966, opened 2023-03-30
fix Blessing of Seasons icons

ToppleTheNun:
### Description

Update Blessing of Seasons icons to match in-game icons.

### Motivation

Blessing of Seasons icons are blue on the WCL icon CDN because they're pulled from Wowhead which has the old icons. The Blizz CDN is giving 403s for these icons.

### Testing

- Screenshot(s):
image


kfinch (approved):
lol @ `BAD_ICONS.ts` didn't even know we had that

PR #5968, opened 2023-03-31
refactor spellId to spell on SpellLink

ToppleTheNun:
### Description

Update SpellIcon and SpellLink's `spellId`/`id` prop to just `spell` to indicate that it takes more than just a spell ID.

### Motivation

It being `spellId`/`id` was confusing as to what values it could take as a property.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -23,6 +23,7 @@ import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
+  change(date(2023, 3, 31), 'Refactor SpellLink props to use "spell" instead of "id".', ToppleTheNun),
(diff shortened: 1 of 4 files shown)


emallson (commented):
assuming that the LSP/IDE that generated this change didn't break the world. merge conflicts now because other stuff got merged, but in principle this is fine.

one thing i'm concerned about is breaking other contributors' local work-in-progress stuff. i think the process here should likely be:

1. *add* the `spell` prop as you have, but with the `id` prop still in place (but marked deprecated so they get IDE warnings)
2. ping in the `#codebase-changelog` channel, informing people they should swap and we're going to remove the `id` prop entirely
3. followup PR in a couple (few?) weeks that replaces any remaining uses and hard-removes the id prop.

ToppleTheNun:
Agreed, I'll clean this up a bit.

emallson (approved):

PR #5969, opened 2023-03-31
add ability to use PTR tooltips

ToppleTheNun:
### Description

Add ability to use PTR tooltips.

### Motivation

WCL has support for 10.1 PTR Aberrus logs in zone ID 33 and 10.1 PTR M+ keystone logs in zone ID 34. This enables people working on things that use PTR logs (like T30 tier set support) to see accurate tooltips for the log that they are working on.

### Testing

- Test report URL: `/report/CA3DaQvTdJ6x1PqL/9-Mythic+Kazzara,+the+Hellforged+-+Wipe+9+(4:18)/Tjuanbeam/standard/character`
- Screenshot(s):

**Before**
image

**After**
image
image

### DIFF
--- a/src/game/ZONES.ts
+++ b/src/game/ZONES.ts
@@ -29,6 +29,7 @@ interface Zone {
   encounters: Encounter[];
   brackets: Bracket;
   partitions?: Partition[];
+  usePtrTooltips?: boolean;
 }
 
 const ZONES: Zone[] = [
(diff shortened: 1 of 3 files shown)


src/game/ZONES.ts
```diff
@@ -64,6 +65,26 @@ const ZONES: Zone[] = [
...
+    id: 33,
+    name: 'Aberrus',
+    frozen: false,
+    encounters: [],
+    brackets: {
+      min: 402,
+      max: 450,
+      bucket: 3,
+      type: 'Item Level',
+    },
```
emallson, line 78: this might have to change to match the wcl brackets. not sure what this does exactly tbh :thinking:

emallson (approved):

src/game/ZONES.ts
ToppleTheNun: It's actually not used anywhere and neither are partitions.

ToppleTheNun: We could update the query that's used to use partitions if we so chose - this would only impact the Character Parses page and I don't know how often that's actually used.

PR #5979, opened 2023-04-04
switch to pnpm

ToppleTheNun:
### Description

Switch to pnpm as the package manager for the project.

### Motivation

Relevant Discord conversation starts here: https://discord.com/channels/316864121536512000/316864385052049420/1092484091011674122

Exported log of convo: https://gist.github.com/ToppleTheNun/d5406a53897b88b6c1ceb6e83c557702

### Notes

The change to the ESLint config is due to `@martijnhols/eslint-config` having dependencies that brought in old versions of ESLint, which broke the ability to build the application.

This takes advantage of [pnpm workspaces](https://pnpm.io/workspaces) to make using the eslint plugin easier; we could also include any other packages we want to develop here as well.


ToppleTheNun:
Currently going through a "out of memory" issue when trying to do a `lingui extract` when yarn does it just fine. Can't migrate while this is occurring.

ToppleTheNun:
It's alive!

ToppleTheNun:
This is included in the Vite branch so I'm dropping this PR for now.

PR #5980, opened 2023-04-05
Update shaohaos suggestion for revival

trevorm4:
Updates logic to only mark it as mistake if they had doubt available and they didnt pre-cast SG

image
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/Revival.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/Revival.tsx
@@ -2,7 +2,7 @@ import { formatThousands } from 'common/format';
 import SPELLS from 'common/SPELLS';
 import { Talent } from 'common/TALENTS/types';
 import { TALENTS_MONK } from 'common/TALENTS';
-import { SpellLink } from 'interface';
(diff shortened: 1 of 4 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/Revival.tsx
```diff
@@ -180,18 +184,36 @@ class Revival extends Analyzer {
           });
           const allPerfs = [efPerf];
           if (this.selectedCombatant.hasTalent(TALENTS_MONK.SHAOHAOS_LESSONS_TALENT)) {
-            let lessonPerf = QualitativePerformance.Good;
-            if (!cast.lessonsBuffActive) {
-              lessonPerf = QualitativePerformance.Fail;
+            let lessonPerf = QualitativePerformance.Fail;
+            if (
+              cast.lessonsBuffActive ||
+              this.shaohaos.getNextBuff() !== SPELLS.LESSON_OF_DOUBT_BUFF
```
vohrr: I would prefer to change this to any buff except Fear since its the only one that doesn't provide substantial value to revival

vohrr (approved):
Just the one comment otherwise lgtm

PR #5990, opened 2023-04-15
T30 prevoker tier set module

trevorm4:
### Description

T30 tier set module

### Testing

- Test report URL: `report/6DqVHhbZCtLWpj4X/21-Mythic+Zskarn+-+Wipe+2+(1:52)/Lerrastrasza/standard/statistics`
- Screenshot(s):
image
image
image

### DIFF
--- a/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
+++ b/src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
@@ -584,6 +584,15 @@ const EVENT_LINKS: EventLink[] = [
       return c.hasTalent(TALENTS_EVOKER.SPARK_OF_INSIGHT_TALENT);
     },
   },
+  {
+    linkRelation: FROM_HARDCAST,
+    reverseLinkRelation: FROM_HARDCAST,
+    linkingEventId: SPELLS.ESSENCE_BURST_BUFF.id,
+    linkingEventType: [EventType.ApplyBuff, EventType.ApplyBuffStack, EventType.RefreshBuff],
+    referencedEventId: [SPELLS.LIVING_FLAME_HEAL.id, SPELLS.LIVING_FLAME_DAMAGE.id],
+    referencedEventType: EventType.Cast,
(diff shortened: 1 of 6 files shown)


src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -584,6 +584,15 @@ const EVENT_LINKS: EventLink[] = [
       return c.hasTalent(TALENTS_EVOKER.SPARK_OF_INSIGHT_TALENT);
     },
   },
+  {
```
vohrr, line 587: Might be helpful/easier to event link the tier essence burst procs to the apply and remove buff events of Essence Rush, ID: 409899  and then look for that instead of having to do negative logic checks in your isEBFromTier30 function 
image

src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -719,4 +728,21 @@ export function wasEbConsumed(event: ApplyBuffEvent | ApplyBuffStackEvent) {
   return HasRelatedEvent(event, ESSENCE_BURST_LINK);
 }
 
+export function isEbFromT30Tier(
```
vohrr, line 731: Small nit on the naming - Eb is easily confused with emerald blossom

vohrr (approved):
Just the one suggestion otherwise looks good

src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
trevorm4: I think this is actually trickier because theoretically someone could cast living flame inside essence rush and proc essence burst but it won't be from the tier set.

vohrr: Oh just saw this, I meant specifically linking to the apply and remove buff events since they occur on the exact time stamp of the procs, not just checking if the buff is active. This works as well so doesn't matter

PR #5993, opened 2023-04-16
Remove motivation from pr template

trevorm4:
Pretty much nobody uses the motivation part and its pretty useless atm, so removing it.


jazminite (commented):
I use it but if you feel that strongly about it, we can remove it.

emallson:
@jazminite I think for a lot of PRs the motivation part is hard to fill out beyond "I want to add this feature" or "the talent was updated", which are also covered by the description

jazminite (approved):

PR #5995, opened 2023-04-16
Port MajorDefensive code to core (again) and clean up / simplify usage

emallson:
### Description

This is a second stab at porting the `MajorDefensive` code to the core guide section of the codebase. It includes a couple of additional changes based on discussions in discord and lessons learned from the previous PR:

- The default `statistic` has been removed. There is now a `MajorDefensiveStatistic` that can be used as a convenient pre-built option.
- The buff/debuff handling logic has been unified and Vengeance code unified with Brewmaster code in a way that works well for both and lacks footguns during setup.
  - As part of this, it is now possible to use this code for defensives that don't fall neatly into either buff or debuff. I don't know *exactly* where this would come up, but if/when it does we won't need to rewrite it.
- All core components take either an analyzer, list of analyzers, or list of mitigations (depending on what they need). Most components should technically function outside the guide context now. This means that the caller is responsible for calling `useAnalyzer` / `useAnalyzers`.
- Reliance on link normalizers has been removed. These are still used elsewhere in the brew code and I am not sure whether they're still used for VDH so i left them in place.
- You can't do `type A = Class<B>` so there are technically sub-classes (`MajorDefensiveBuff`, `MajorDefensiveDebuff`) that set type parameters for the two common cases.
- Added a good bit of documentation + pointers to simple & complex implementations to look at for examples.

### Motivation

Both @kfinch and @abelito75 were asking about for Guardian Druid and Prot Warrior. @ToppleTheNun was using this but duplicated the code because my implementation was non-portable.

### Testing

Conveniently, my raid team has both a brewmaster and a vdh so I tested primarily on them (+ a couple of rando low parse logs):

- Brewmaster: `/report/dFhzw48DMj3AX1g6/48-5+Dathea,+Ascended/Eisenpelz/standard`
- Vengeance: `/report/A14qTMW3YkJGtzQn/8-Mythic+Raszageth+the+Storm-Eater+-+Kill+(12:16)/Guccidh/standard/overview`

There should be very few/no visible changes, other than these:

- The Vengeance mitigation segments are now normalized by max hp instead of by largest seen mitigation across all mitigation analyzers.
- 2 of the Vengeance statistics are now in the General category instead of Talents (they aren't talents, they just were put into the Talents category by the old default implementation).


ToppleTheNun:
An interesting bug not related to the changes here, just surfaced by the change:
image

ToppleTheNun (approved):
I'm happy that we'll be able to reduce the copied code!

PR #5998, opened 2023-04-17
[Hunter] Tier 29 BM tier, general housekeeping and bugfixes

Pewtro:
### Description

Adds support for T29 sets for Beast Mastery
Fixes a few bugs around modules
Using ts-prune found all unused exports in retail hunter folders and cleaned them up.


emallson:
i am gonna need to try this `ts-prune` thing :eyes:

Pewtro:
It's very neat! 

Quick and dirty get a list of unused for a class/spec on linux/macos: 
`npx ts-prune | grep **/retail/monk`

On windows, I don't have a good equivalent, but I think it should be doable using findstr somehow

ToppleTheNun:
iirc it's technically built into the TypeScript server as well. You can setup VS Code to do it automatically: https://www.techiediaries.com/vscode-automatically-organize-typescript-imports/

WebStorm can do it as well.

Pewtro:
That doesn't look like what ts-prune does

ToppleTheNun:
I think I misread "exports" as "imports"

B0uss:
> It's very neat!
> 
> Quick and dirty get a list of unused for a class/spec on linux/macos: `npx ts-prune | grep **/retail/monk`
> 
> On windows, I don't have a good equivalent, but I think it should be doable using findstr somehow

Hi @Pewtro, on windows, I think we could use `npx ts-prune | select-string "retail\\monk"`

emallson (approved):

PR #6003, opened 2023-04-21
[Shadow] Improve Void Bolt spellusable tracking and Max Cast calculations

DoxAshe:
### Description

This change improves the spell usable tracking of Void Bolt's cooldown.  It also improves the guide view to properly show when void bolt is available.  
It also calculates the max casts of Void Bolt in a new way, that should be significantly more accurate.

### Testing

Void Bolt is a weird spell that doesn't work well with the usual cast efficiency, since there are large periods outside of voidform where the spell cannot be cast.  There is probably a more robust way of calculating this using spellhistory, but I was unable to get that to work for only one voidform at a time.


src/analysis/retail/priest/shadow/modules/Abilities.tsx
```diff
@@ -16,13 +15,12 @@ class Abilities extends CoreAbilities {
...
-          suggestion: true,
+          suggestion: false,
           recommendedEfficiency: 0.85,
-          maxCasts: (cooldown: number) =>
-            calculateMaxCasts(cooldown, combatant.getBuffUptime(SPELLS.VOIDFORM_BUFF.id)),
+          //maxCasts: calculateMaxCasts(cooldown, combatant.getBuffUptime(SPELLS.VOIDFORM_BUFF.id)),
         },
         damageSpellIds: [SPELLS.VOID_BOLT.id, SPELLS.VOID_BOLT_DISSONANT_ECHOES.id],
-        enabled: combatant.hasTalent(TALENTS.VOID_ERUPTION_TALENT),
+        enabled: false, //combatant.hasTalent(TALENTS.VOID_ERUPTION_TALENT),
```
emallson: should this be false? it indicates that the spell can't be cast at all to other systems.

EDIT: oh i see, this is added by the VB module. Probably want to entirely remove this entry.

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
```diff
@@ -1,39 +1,178 @@
...
+
+  VB = [0]; //timestamps of voidbolt spellusable updates
 
   constructor(options: Options) {
     super(options);
 
+    this.active = this.selectedCombatant.hasTalent(TALENTS.VOID_ERUPTION_TALENT);
+
+    //I don't know how to add an event listener at the start of the fight,
+    this.addEventListener(Events.damage, this.onFightstart);
```
emallson: you can use a listener for `combatantinfo` events i believe

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
```diff
@@ -1,39 +1,178 @@
...
+  }
+
+  onFightstart(event: DamageEvent) {
+    if (this.oncestart === 0) {
+      const startevent = event;
+      startevent.timestamp = this.owner.fight.start_time;
+      //At the start of the fight, VB is not available.
+      //Set VB to be on cooldown, and slow the cooldown rate so it remains unusable.
+      this.spellUsable.applyCooldownRateChange(SPELLS.VOID_BOLT.id, rate);
+      this.spellUsable.beginCooldown(startevent, SPELLS.VOID_BOLT.id);
```
emallson: not sure how i feel about this

emallson (commented):
Would this be more cleanly implemented using `ExecuteHelper`? Can turn off the regular execute part with `lowerThreshold = -1` and use the `executeOutsideRangeEnablers` / `countCooldownsAsExecuteTime` options to enable VB while in Voidform.

You could look at hunter's `KillShot` module for an example that isn't *too* dissimilar (they have an actual execute window, and use `singleExecuteEnablers` instead of `executeOutsideRangeEnablers` because their buffs are single-use procs).

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
DoxAshe: I'm still a little stuck, I tried to addEventListner for Events.combatantinfo but maybe I'm missing something I needed to import?

src/analysis/retail/priest/shadow/modules/Abilities.tsx
DoxAshe: I've removed it from abilities.

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
DoxAshe: It isn't great, but there seems to be no other way to make a spell not usable.  Without doing this, voidbolt shows as available for durations where it is not.
Incorrect:
VBwrong

Correct:
VBcorrect

DoxAshe:
I've tried using Executehelper, and have it track when Void Form is active.  It calculates the time in voidform correctly, but the cooldown it uses for voidbolt is off and I was unable to fix it.  It ended up giving a number that was very far off.
In addition, it rolls over partial missed casts from previous Void Forms.  The way I've done it, it floors the missed potential casts after each use of the cooldown.  Its overall minor, but on very long fights it can add up to be several casts.

The way I've done is weird, and if its too unusual I can keep working on it to try to find another way.

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
emallson: try using `EventType.CombatantInfo` instead of `Events.combatantinfo`, but now that you mention it i'm not even sure that combatantinfo events get dispatched to analyzers tbh. If not, you can just leave it as you have it.

emallson:
Hmm, I think I'm okay with this as-is as long as we add some warning comments not to spread this to other specs, but VB is very much an oddball so i don't even know if there's another spec that'd crib from this.

The issue with the guide component feels more like we need to make ExecuteHelper into something first-class that the guide supports for these? not sure.

Pewtro:
ExecuteHelper is definitely made to be all encompassing, so if there's something it isn't covering - should be touched up, but I feel I went around all the corner cases for it 🤔

emallson:
I think its that the guide components don't respect it. @DoxAshe how do you feel about converting this to `ExecuteHelper` and then I PR a change this weekend to make the guide uptime component respect its info?

EDIT: although if there are CD issues from VB via ExecuteHelper then that may be a wash anyway. I'm not sure if overall VB efficiency is very useful as a stat vs showing it per VF cast but I know fuckall about spriest so I could be off base there

Pewtro:
I'd have to run through the math but I think the way it's calculating is accurate enough because even if you get 50% of the CD at the tail-end of VF, that tracks well with you being able to cast it instantly on entering next VF, but I don't have the code in front of me - but VF doesn't sound that different to other execute spells, so not sure why it wouldn't be accurate for that in particular.

DoxAshe:
If execute helper could show some effect on the guide view I think that would be great, and would solve the issues with the graph in the guide view.  I'd also use that for Shadow Word: Death, and I'm sure other specs would like it for their executes. But for void bolt, I also wanted to track its availability in spell usable, though that is something I don't need right now and can work on later or find a work around.
I could change the void bolt to show how many missed cast there were in each voidform, but even if I was to do that I would still want an overall efficiency to show.

I'll try using executehelper with it again and see if I can get it to work with Voidbolt.

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
DoxAshe: I was using Event instead of EventType, thank you.
I gave this a try, but it didn't work. I'll keep it as is.

DoxAshe:
I've used execute helper for void bolt, and it calculates the amount of time I'm in Void Form correctly, but I don't think it has a way of calculating void bolt's cooldown. 
The way I've seen other execute spells work is by using abilities.getAbility().cooldown.  But for void bolt this number is off.
For the example report, I should have 21 casts of voidbolt, but instead I get 19.

DoxAshe:
Sorry for the delay, I've removed the excess code from voidform since it is now handled by the new CastEfficiencyBar and Execute Helper.
It now looks like this, using only the Cast Efficiency Bars and Execute Helper:
Voidbolt
Thank you for all your help getting this to work.

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
```diff
@@ -0,0 +1,75 @@
...
+        maxCasts: () => this.maxCasts,
+      },
+    });
+  }
+
+  adjustMaxCasts() {
+    const cooldown = this.abilities.getAbility(SPELLS.VOID_BOLT.id)!.cooldown * 1000;
+    this.maxCasts += Math.ceil(this.totalExecuteDuration / cooldown);
+
+    console.log('CD of VB', cooldown, this.maxCasts);
```
emallson:
```suggestion
```
stray console.log

emallson (approved):
one stray `console.log` and then we merge + go

(i think github will let me just apply that so ima try that now)

PR #6006, opened 2023-04-22
[Druid Guardian] - Updated for Dragonflight, implemented Guide

kfinch:
### Description

This pull request updates Guardian Druid for Dragonflight and implements the Guide. There are a number of legacy statistic modules which have been disabled for now until I can find the time to update / overhaul them - the Guide alone should be sufficient to make this a usable tool.

I've also made a few minor changes / additions to core components:
* Damage Mitigated dot chart now takes `QualitativePerformance` for each dot value instead of boolean (allows Ironfur to show 0 stacks vs 1 stack vs many stacks, probably uses for other tanks too)
* Added optional yScale to MajorDefensives chart, allowing yAxis size to be settable to customized percentage of player's max health. This is useful because some mitigation is coded as absorbs, making raw incoming damage magnitude very different for different tanking mechanics.
* Added helper function for deciding if incoming damage was mitigated by armor.

### Testing

- Test report URL: `/report/DVa7RLYBdZcwKXQv/11-Mythic+The+Primal+Council+-+Kill+(2:34)/Sol/standard/overview`
- Screenshot(s): 
image


emallson (approved):
looks good! It seems like the mitigation section stuff mostly worked out, which I'm glad to see :)

> Added optional yScale to MajorDefensives chart, allowing yAxis size to be settable to customized percentage of player's max health. This is useful because some mitigation is coded as absorbs, making raw incoming damage magnitude very different for different tanking mechanics.

So under the hood the mitigation chart adds the damage mitigated back onto damage taken, so it *should* be constant regardless of how much you mitigate.

However, I wrote it initially for Brewmaster which is the lowest HP tank, and then Guardian is the highest HP (I think?) aside from like BDK during Vamp Blood so the scale proooobably needed to change for that reason. (e.g. my monk has around 515k hp in raid with buffs, while a random similar ilvl bear is 733k---a gap of around 30% and even then the brew one is probably a tad high. i played around with setting it to 80% of brew hp before just settling on 100%)

kfinch:
Ahhh that makes sense, but still I found that the damage scaling of the graph was way too low - I assume you're fine with the yScale because you approved, so gonna merge this. We can fiddle with the scaling more later if needed.

PR #6008, opened 2023-04-23
tier 30 support for holy priest

foobartles:
### Description

Add support for the Tier 30 4 set for holy priest. I don't think it is possible to track 2 piece because of the way that PoM is logged to the combat log. It would not be clear if a PoM application is tied to 2 piece proc or is just from an existing PoM bouncing.

### Testing
Test report from PTR:
https://www.warcraftlogs.com/reports/AkdM1wpqKVTrLYFG#fight=17
image

### DIFF
--- a/src/analysis/retail/priest/holy/modules/dragonflight/tier/tier30/Tier30HolyPriest4Set.tsx
+++ b/src/analysis/retail/priest/holy/modules/dragonflight/tier/tier30/Tier30HolyPriest4Set.tsx
@@ -1,0 +1,147 @@
+import SPELLS from 'common/SPELLS';
+import TALENTS from 'common/TALENTS/priest';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import Events, { CastEvent, HealEvent } from 'parser/core/Events';
+import {
+  calculateEffectiveHealing,
+  calculateOverhealing,
+  calculateEffectiveDamage,
+} from 'parser/core/EventCalculateLib';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
+import {
+  getSerenityHealEvent,
+  getChastiseDamageEvent,
+  getSalvationHealEvents,
+  getSanctifyHealEvents,
+} from 'analysis/retail/priest/holy/normalizers/CastLinkNormalizer';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import { formatNumber, formatPercentage } from 'common/format';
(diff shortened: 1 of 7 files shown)


trevorm4:
It's possible to track the 2 piece but it would require a decent amount of coding, you would need to do event linking and apply similar logic to Dancing Mists to track which are from casts and which are from procs (similar idea to dancing mist tracking) since procs will generally apply after the original bounce.

src/analysis/retail/priest/holy/modules/dragonflight/tier/tier30/Tier30HolyPriest4Set.tsx
```diff
@@ -0,0 +1,147 @@
...
+    return this.selectedCombatant.getBuffStacks(
+      SPELLS.HOLY_PRIEST_TIER_30_4_SET_BUFF.id,
+      null,
+      FALL_BUFFER,
+    );
+  }
+
+  onSerenityCast(event: CastEvent) {
+    const stacks = this.tierStacks;
+    console.log(stacks);
```
trevorm4: Get rid of this log

src/analysis/retail/priest/holy/modules/dragonflight/tier/tier30/Tier30HolyPriest4Set.tsx
```diff
@@ -0,0 +1,147 @@
...
+      });
+    }
+  }
+
+  onSalvationCast(event: CastEvent) {
+    const stacks = this.tierStacks;
+    if (stacks > 0) {
+      this.stacksOnCast.push(stacks);
+      const healEvents = getSalvationHealEvents(event);
+      console.log(healEvents);
```
trevorm4: Remove

src/common/SPELLS/priest.ts
```diff
@@ -517,6 +517,11 @@ const spells = spellIndexableList({
     name: 'Seize the moment',
     icon: 'inv_mace_1h_artifactheartofkure_d_03',
   },
+  HOLY_PRIEST_TIER_30_4_SET_BUFF: {
+    id: 409479,
+    name: 'Inspired Word',
+    icon: 'ability_paladin_sacredcleansing',
+  },
```
trevorm4, line 524: Put this in ITEMS/priest.ts instead

trevorm4 (changes requested):

foobartles:
> It's possible to track the 2 piece but it would require a decent amount of coding, you would need to do event linking and apply similar logic to Dancing Mists to track which are from casts and which are from procs (similar idea to dancing mist tracking) since procs will generally apply after the original bounce.

I can look into doing this if I have time, haven't really been playing my priest recently so my motivation for stuff like that has gone down the drain haha

trevorm4:
I saw you requested review but don't see a new commit, did you forget to push?

foobartles:
> I saw you requested review but don't see a new commit, did you forget to push?

🤦 My bad, I hit push but closed my laptop before it finished and the push timed out lol. Sorry about that, should be pushed now

trevorm4 (approved):

PR #6009, opened 2023-04-24
[Classic] Potions updates

jazminite:
### Description

- Add potion spells and items
- Update Potion Panel to work for both Retail and Classic
- Update Classic's Combat Potion Checker
- Add Classic Potion Checker
- Update Retail Potion Checker to use dynamic messaging

### Testing

Test report URLs
- `/report/VQW4jb2GrkNF6Kaz/45-Hardmode+Thorim+-+Kill+(2:25)/Farmpapi`
- `/report/v3TJFzZNCdM7a8Hx/11-Normal+Patchwerk+-+Kill+(2:01)/Melancailey`

Screenshot(s)
- Fail
image
image
- Average
image
image
- Good
image


src/interface/guide/components/Preparation/ConsumablesSubSection/PotionPanel/index.tsx
```diff
@@ -1,12 +1,19 @@
...
+import Expansion from 'game/Expansion';
 
-const PotionPanel = () => {
-  const potionChecker = useAnalyzer(PotionChecker);
+interface Props {
+  expansion?: Expansion;
+}
+const PotionPanel = ({ expansion }: Props) => {
+  const UsePotionChecker =
+    expansion === Expansion.WrathOfTheLichKing ? ClassicPotionChecker : PotionChecker;
```
ToppleTheNun: I would suggest using `isClassicExpansion` rather than `expansion === Expansion.WrathOfTheLichKing` so that we don't have to update this check if/when Cata classic comes out.

ToppleTheNun (changes requested):
One comment, otherwise it looks good to me.

src/interface/guide/components/Preparation/ConsumablesSubSection/PotionPanel/index.tsx
jazminite: Updated :)

ToppleTheNun (approved):

PR #6018, opened 2023-04-29
Change load conditions for reversion cast eff bar & Holy Paladin t30 tier set

trevorm4:
### Description

Forgot to split up the commits and cba so combined PR here

### Testing

image
tooltip
image

### DIFF
--- a/src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
+++ b/src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
@@ -2,7 +2,7 @@ import SPELLS from 'common/SPELLS';
 import { TIERS } from 'game/TIERS';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import { calculateEffectiveHealing, calculateOverhealing } from 'parser/core/EventCalculateLib';
-import Events, { HealEvent, ResourceChangeEvent } from 'parser/core/Events';
+import Events, { CastEvent, HealEvent, ResourceChangeEvent } from 'parser/core/Events';
 import BoringValueText from 'parser/ui/BoringValueText';
 import ItemHealingDone from 'parser/ui/ItemHealingDone';
 import Statistic from 'parser/ui/Statistic';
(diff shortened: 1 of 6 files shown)


src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import { TALENTS_PALADIN } from 'common/TALENTS';
+import { formatNumber } from 'common/format';
+import HIT_TYPES from 'game/HIT_TYPES';
+import { getLightsHammerHeals } from '../../../normalizers/CastLinkNormalizer';
+import { SpellLink } from 'interface';
+
+const CRIT_HEAL_AMOUNT = 1.65;
```
vohrr: This just got hotfixed to 60% more didn't it

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+    this.active = this.selectedCombatant.has2PieceByTier(TIERS.T30);
+    if (!this.active) {
+      return;
+    }
+    this.has4Piece = this.selectedCombatant.has4PieceByTier(TIERS.T30);
+    this.addEventListener(
+      Events.heal.by(SELECTED_PLAYER).spell(SPELLS.HOLY_SHOCK_HEAL),
+      this.onHolyShockHeal,
+    );
+    this.addEventListener(
```
mfield, line 62: maybe needs a `if (!this.has4Piece) { return }` here, or checks in the event handlers?

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+        this.fourPcHealing += ev.amount;
+        this.fourPcOverhealing += ev.overheal || 0;
+        this.extraLhHits += 1;
+      });
+    }
+    // mark events as counted
+    healEvents.forEach((ev) => {
+      this.countedHeals.add(ev);
+    });
+    this.lhTiktok = !this.lhTiktok;
```
mfield: i do like how simple this approach is. but i am a bit worried how it behaves if the raid has to move in and out of the area and misses some ticks. i think we should reset `this.lhTiktok = false` when we see a new cast event for light's hammer.

an alternative approach could be based on the timing of the ticks. since the tick rate doesn't look hasted maybe this isn't too complicated?

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import { TALENTS_PALADIN } from 'common/TALENTS';
+import { formatNumber } from 'common/format';
+import HIT_TYPES from 'game/HIT_TYPES';
+import { getLightsHammerHeals } from '../../../normalizers/CastLinkNormalizer';
+import { SpellLink } from 'interface';
+
+const CRIT_HEAL_AMOUNT = 1.65;
+const TWO_PIECE_HEAL_INC = CRIT_HEAL_AMOUNT / 1.5; // 210% heal instead of 150% on crit
```
mfield: crit heals are 200% in pve.
and if i understand these numbers correctly, the comment includes the most recent buff to 60% extra crit healing but `CRIT_HEAL_AMOUNT` does not?

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
vohrr: Correct, LH ticks are not hasted so you could just add a timestamp check with some ms threshold to protect the integrity of your clock

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  onLHHeal(event: HealEvent) {
+    if (this.countedHeals.has(event)) {
+      return;
+    }
+    const healEvents = getLightsHammerHeals(event);
+    if (this.lhTiktok) {
+      healEvents.forEach((ev) => {
+        this.fourPcHealing += ev.amount;
```
vohrr: `this.fourPcHealing += ev.amount + (ev.absorbed || 0);`

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  <SpellLink spell={TALENTS_PALADIN.LIGHTS_HAMMER_TALENT} /> hits
+                </li>
+              )}
+            </ul>
+          </>
+        }
+      >
+        <BoringValueText label="Heartfire Sentinel's Authority (T30 Set Bonus)">
+          <h4>2 Piece</h4>
+          <ItemHealingDone amount={this.hsHealing} /> <br />
```
vohrr, line 167: CDR tally would be nice to see here but not something to hold the pr up over

vohrr (changes requested):

src/analysis/retail/paladin/holy/modules/dragonflight/tier/T30TierSet.tsx
trevorm4: Yea it did, was planning to put a pr in on tuesday so that ptr log analysis was accurate.

trevorm4:
image
Updated the logic some and result is in the same ballpark which is good. Went with the approach where even second intervals are considered base ticks [0,1) U [2,3) ... and we count odd second intervals as healing from the tier set.

vohrr (approved):

PR #6039, opened 2023-05-05
WW: New Serenity/Non-Serenity APLs, cooldown/ability tracking fixes

awiss:
### Description
- Updates to APLs. There are now two APLs, one for during Serenity, one for any other time. These now have their own sections in the checklist (the Serenity one only appears if you had it talented)
- Refactored BlackoutKick.tsx, and fixed a bug I introduced due to my last modRate fix.
- Added a custom SpellUsable, which attributes early RSKs to Teachings of the Monastery cd reset procs (if talented)
- Stopped counting total fists of fury hit stats during serenity, since you are supposed to cancel it then.

APL bugfixes:
- Added a normalizer to make sure Dance of Chi Ji buff is removed after the SCK cast event
- **Removes SpendResource from hasResource APL condition, which was causing incorrect resource states.**
- **Requires `sourceID === info.playerId` in aplProcessesEvent so that SEF image casts aren't counted**

FYI @emallson these are the two APL changes here which affect other specs.

Happy to hear whatever people think is still left to get WW to an acceptable release state.

### Testing

- Test report URL:
  -  With serenity: `/report/vkb8nPmxa1BNKq6t/3-Heroic+Terros+-+Kill+(3:34)/Tenooki/standard/overview`
  -  Without serenity: `/report/vkb8nPmxa1BNKq6t/24-Heroic+Dathea,+Ascended+-+Kill+(4:07)/Tenooki/standard/overview`
- Screenshot(s):


src/analysis/retail/monk/windwalker/modules/core/SpellUsable.tsx
```diff
@@ -0,0 +1,52 @@
+import SPELLS from 'common/SPELLS/monk';
```
vohrr, line 1: Teachings of the Monastery is a talent on the mistweaver tree as well, should move this into monk/shared folder and/or put into the Teachings of the Monastery module file

src/analysis/retail/monk/windwalker/modules/core/SpellUsable.tsx
```diff
@@ -0,0 +1,52 @@
...
+class SpellUsable extends CoreSpellUsable {
+  static dependencies = {
+    ...CoreSpellUsable.dependencies,
+  };
+
+  lastPotentialTriggerForRskReset: AnyEvent | null = null;
+  constructor(options: Options) {
+    super(options);
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(SPELLS.BLACKOUT_KICK),
```
vohrr, line 20: There is a different spellId for blackout kicks that come from Teachings of the Monastery make sure to add that to the listener.
Regular BoK: `https://www.wowhead.com/spell=100784`
Totm BoK: `https://www.wowhead.com/spell=228649`

src/analysis/retail/monk/windwalker/modules/spells/FistsofFury.tsx
```diff
@@ -41,16 +48,23 @@ class FistsofFury extends Analyzer {
...
   }
 
-  get averageTicks() {
-    return this.fistsTicks / this.casts;
+  onFistsCast(event: CastEvent) {
+    if (!this.selectedCombatant.hasBuff(TALENTS_MONK.SERENITY_TALENT.id)) {
+      this.inSerenity = false;
+      this.nonSerenityCasts += 1;
+    } else {
+      this.inSerenity = true;
```
vohrr: This seems like an odd way to be tracking whether serenity is active or not, I think I would rather see event listeners tied to the buff  apply/remove events of serenity itself or inject the serenity module if there is tracking for its active state there

src/analysis/retail/monk/windwalker/modules/talents/XuensBattlegear.tsx
```diff
@@ -6,7 +6,7 @@ import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Events, { DamageEvent } from 'parser/core/Events';
 import { ThresholdStyle, When } from 'parser/core/ParseResults';
 import AbilityTracker from 'parser/shared/modules/AbilityTracker';
-import SpellUsable from 'parser/shared/modules/SpellUsable';
+import SpellUsable from 'analysis/retail/monk/windwalker/modules/core/SpellUsable';
```
vohrr, line 9: I know you only imported this into this module but while you're here you should update this statistic to be in the TALENTS category since its no longer a SL Leggo
line 105:
`category={STATISTIC_CATEGORY.ITEMS}` => `category={STATISTIC_CATEGORY.TALENTS}`

src/common/SPELLS/monk.ts
```diff
@@ -163,6 +163,11 @@ const spells = spellIndexableList({
     name: "Invoker's Delight",
     icon: 'inv_inscription_80_warscroll_battleshout',
   },
+  FURY_OF_XUEN_BUFF: {
```
vohrr, line 166: May as well add the stacking buff SPELL for this while you're in here
`https://www.wowhead.com/spell=396167`

vohrr (changes requested):

src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
```diff
@@ -1,63 +1,110 @@
...
+    spell: TALENTS.FISTS_OF_FURY_TALENT,
+    condition: and(inSerenity, serenityDurationRemainingLT(1500)),
+  },
+  {
+    spell: TALENTS.FAELINE_STOMP_TALENT,
+    condition: and(
+      hasTalent(TALENTS.FAELINE_HARMONY_TALENT),
+      buffMissing(SPELLS.FAELINE_HARMONY_BUFF),
+    ),
+  },
```
emallson: this is missing `inSerenity`

src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
```diff
@@ -1,63 +1,110 @@
...
+    spell: TALENTS.CHI_WAVE_TALENT,
+    condition: inSerenity,
+  },
+  {
+    spell: SPELLS.TIGER_PALM,
+    condition: inSerenity,
+  },
+]);
+
+export const nonSerenityApl = build([
```
emallson: the way you've structured this, this apl will also apply to serenity actions. you could add a `and(not(inSerenity))` to them all if you want to explicitly exclude serenity.

emallson (commented):
this implementation will probably *work* but it obviously is less than ideal. i probably need to look seriously at adding sub-APLs like simc supports to be able to break out logically separate pieces.

you could alternately mix the serenity / non-serenity APLs but I think that would just be an illegible mess from a presentation perspective.

src/analysis/retail/monk/windwalker/modules/core/SpellUsable.tsx
awiss: Hmm, I have to investigate whether the extra kicks can actually proc the rsk reset, but if so will do.

src/common/SPELLS/monk.ts
awiss: Good idea.

Random question: do you have and tips for tracking down spell ids for different buffs / spells? I often don't have much confidence just from searching through wowhead, instead resorting to just going in game and getting the buff the proc myself.

ToppleTheNun: image

If you open a report in WoWA, you can click "About" and then "View all events" to see a nice view of all of the events in a log - you can filter from there as well.

image

src/analysis/retail/monk/windwalker/modules/core/SpellUsable.tsx
trevorm4: They can proc it.

src/analysis/retail/monk/windwalker/modules/spells/FistsofFury.tsx
awiss: Ok, I'll inject the serenity module. Is there a way I could use this module in an apl condition as well?

awiss: I renamed this to lastCastInSerenity, which is a better name.
The idea is I don't want to count any damage ticks from a FoF cast made during serenity (even if the serenity buff expires), since we we aren't counting that cast in the denominator of the final calculation.

src/parser/shared/metrics/apl/index.ts
```diff
@@ -391,12 +391,14 @@ export function aplProcessesEvent(
...
   applicableSpells: Set<number>,
+  playerId: number,
 ): event is BeginChannelEvent | CastEvent {
   return (
     (event.type === EventType.BeginChannel ||
       (event.type === EventType.Cast &&
         event.ability.guid !== result.mostRecentBeginCast?.ability.guid)) &&
-    applicableSpells.has(event.ability.guid)
+    applicableSpells.has(event.ability.guid) &&
+    event.sourceID === playerId
```
emallson, line 401: i guess this is needed because of SEF clones copying your spell casts?

src/parser/core/EventsNormalizer.ts
```diff
@@ -14,6 +14,13 @@ abstract class EventsNormalizer extends Module {
    */
   abstract normalize(events: AnyEvent[]): AnyEvent[];
 
+  /**
+   * Override this value to true to designate an event normalizer to not be run during the
+   * CombatLogParser's normalize function. These normalizers will instead to used on the final
+   * event history (including fabricated events) which is provided to the APL checker.
+   */
+  isAplNormalizer: boolean = false;
+
```
emallson: i am not a fan of special casing this. let me look at what the underlying issue is and see what i can figure out.

emallson (commented):
let me look at the fof normalization thing. other changes to apl code look fine.

src/parser/shared/metrics/apl/index.ts
awiss: Yes otherwise they are all counted as casts

src/parser/core/EventsNormalizer.ts
awiss: Ok, I'll get rid of this. I can fix it by just not applying Channeling to Fists of Fury (the BeginChannel/EndChannel events it adds aren't used anyways).

emallson (approved):

vohrr (approved):

awiss:
Ok, ready to merge now.

PR #6040, opened 2023-05-05
[Shadow] Update for 10.1

DoxAshe:
### Description

This update uses the new Priest talents and abilities.  I've changed the old Priest Talent file to the new one that was generated for this patch (that was later reverted).  I'm not sure if the other priest specs need the old Talent File, but this gets shadow to work on the new one.

- Updated ability cooldowns to their changed values.
- Updated insanity generation to the new values.
- Removed the modules for talents that were removed from the talent tree.  
- Added the new talents (Mind's Eye and Distorted Reality) that adjust the cost of devouring plague.
- Added the new Surge of Insanity Talent for both Mind Flay and Mind Spike
- Updated suggestions/thresholds for several procs.

Shadow has a lot of new talents, and while this doesn't have statistics for most of them, none of them cause any errors in the guide view.  In addition, all of the changes to base spells or old talents have been made, so they should all be working properly.
For now, I have left the compatibility as partial, just in case some of the suggestions/thresholds end up being incorrect.

### Testing
The only issue I have found is that the old view doesn't load when selected but just reloads the guide view.  Until I update the guide view I wasn't going to try to update the old view.

- Screenshot(s):


emallson:
> The only issue I have found is that the old view doesn't load when selected but just reloads the guide view. Until I update the guide view I wasn't going to try to update the old view.

this is probably something in old JS code throwing an error.

src/analysis/retail/priest/shadow/modules/Abilities.tsx
```diff
@@ -118,6 +93,10 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
+        insanityCost:
+          5000 -
+          combatant.getTalentRank(TALENTS.MINDS_EYE_TALENT) * 5 +
+          combatant.getTalentRank(TALENTS.DISTORTED_REALITY_TALENT) * 5,
```
emallson: is the scale on these correct? that would be 0.1% cost reduction per talent point. looks like you actually want `* 500` to reduce the cost by 5 insanity per.

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
```diff
@@ -80,6 +82,38 @@ const CoreCooldownsGraph = () => {
     );
   }
 
+  /* There are some issues with SW:D highlight sections that need to be worked out furtuer before implementation.
```
emallson: i will investigate this

src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
```diff
@@ -10,7 +12,10 @@ import { SHADOW_WORD_DEATH_EXECUTE_RANGE } from '../../constants';
 class ShadowWordDeath extends ExecuteHelper {
   static executeSources = SELECTED_PLAYER;
   static lowerThreshold = SHADOW_WORD_DEATH_EXECUTE_RANGE;
+  static singleExecuteEnablers: Spell[] = [SPELLS.DEATHSPEAKER_TALENT_BUFF];
+  //static executeOutsideRangeEnablers: Spell[] = [TALENTS.INESCAPABLE_TORMENT_TALENT]; //Need to fabricate a buff for when Inescapable Torment(mindbender) is active.
```
emallson, line 16: we could modify `ExecuteHelper` a bit to make it support using other start/stop events. you wouldn't get the convenience of setting it with a prop like this, but you could add your own event listeners for mindbender and call `applyExecuteEnabler` and `removeExecuteEnabler` when it spawns/dies rather than fabricating a buff

src/analysis/retail/priest/shadow/modules/Abilities.tsx
DoxAshe: You are correct, I have updated this

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
DoxAshe: It seems that when there are multiple targets, it creates a highlight for each one that is in execute range, causing it to look like this:
SWD bad
instead of something like this:
SWD better

I ended up running into a different problem where some of the highlights last the entire fight, but I haven't looked into it very much yet.

src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
DoxAshe: That would work great for this.

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
emallson: Yea that makes sense. I saw you added some code to merge overlapping ranges. The crash is caused when *no* enemies have an execute range for SW:D. I have a fix for it inbound.

src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
emallson: I looked into this a little bit. Looks like either way you need fabrication because the mindbender doesn't have a death event and the duration is variable https://www.warcraftlogs.com/reports/gPq7Kyv46RCmkzfn#fight=6&type=summary&pins=0%24Separate%24%23244F4B%24casts%240%240.0.0.Any%24214110952.0.0.Priest%24true%240.0.0.Any%24false%24200174&hostility=&source=30.1&view=events

so i'm not gonna worry about doing that as part of this PR. i can do that separately and then you can do the mindbender fabrication (either way) and then either use a buff or manually call apply/remove.

DoxAshe: Sounds good.  Its something I will work on in the future.

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
DoxAshe: Thank you! Sorry I didn't finish it myself, when I saw that the new shadow priest talents were causing trouble I tried to get something up and running as soon as I could.

emallson: all good! shadow got a ton of changes, i'm impressed you got this out so fast!

DoxAshe:
I'm not sure why one of the checks was not successful.  I can't tell if this is something I did wrong in shadow priest or something else.

emallson:
The e2e check has been timing out recently. its not related to your changes. I'm not sure exactly what is going on

emallson:
If you're happy with where this PR is, we can merge this. that check is optional

ToppleTheNun:
> The e2e check has been timing out recently. its not related to your changes. I'm not sure exactly what is going on

I have a fix for this that I'm working on, should have it up soon.

DoxAshe:
This is a great spot for shadow to be in.  Once when I have some logs of the new raid I will be able to update it some more to make it complete.

PR #6043, opened 2023-05-05
fix playwright tests broken by patch 10.1

ToppleTheNun:
### Description

Fix Playwright tests failing due to patch 10.1.


emallson (approved):
long term might want a way to do this that doesn't rely on hard-coded values, but for now this is good :+1:

PR #6050, opened 2023-05-08
Fix Flow of the Tides

Elitesparkle:
### Description

Flow of the Tides gives 1 extra bounce regardless of Riptide being consumed.

### Testing

No testing.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -17,12 +17,14 @@ import {
   Vireve,
   Pilsung,
   HerzBlutRaffy,
-  Abelito75
+  Abelito75,
+  Elitesparkle
 } from 'CONTRIBUTORS';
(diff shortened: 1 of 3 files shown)


vohrr (changes requested):
Couple things:
You are a missing a few requirements for a PR to be merged, check them out here 
 https://github.com/WoWAnalyzer/WoWAnalyzer/wiki
There is a bit more to this fix than this, just removing the return statement will cause the code to evaluate every chain heal healevent as though it is buffed by 30%.
You should be fine to simply define a new parameter `healingIncrease` and set its value (0 or 30%) based on whether  or not the riptide was consumed in the handler

vohrr (commented):
Going to close this as I have the fix in que

vohrr:
Goi g

PR #6051, opened 2023-05-08
Fix Mana Cost for Shaman Spells and Talents

Elitesparkle:
### Description

This is an update for Mana Cost values of Abilities and Talents for Restoration Shaman.

### Testing

Compare the values with in-game ones, with Dragonflight 10.1 version.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -17,14 +17,12 @@ import {
   Vireve,
   Pilsung,
   HerzBlutRaffy,
-  Abelito75,
-  Elitesparkle
+  Abelito75
 } from 'CONTRIBUTORS';
 import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
-  change(date(2023, 5, 8), 'Fix Mana Costs for Shaman Spells and Talents.', Elitesparkle),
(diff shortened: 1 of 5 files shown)


src/CHANGELOG.tsx
```diff
@@ -17,12 +17,14 @@ import {
...
   HerzBlutRaffy,
-  Abelito75
+  Abelito75,
+  Elitesparkle
 } from 'CONTRIBUTORS';
 import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 export default [
+  change(date(2023, 5, 8), 'Fix Mana Costs for Shaman Spells and Talents.', Elitesparkle),
```
vohrr: This is the wrong file for spec-specific changes, you want to add your changelog entry at `src\analysis\retail\shaman\restoration\CHANGELOG.tsx`

vohrr (changes requested):

src/CHANGELOG.tsx
Elitesparkle: I went for that because I've changed the Mana Cost for Generic Spells and Talents too. Still wrong?

vohrr: yep

trevorm4 (approved):

PR #6061, opened 2023-05-10
shard playwright tests

ToppleTheNun:
### Description

trying to speed up playwright tests from taking 30 minutes.

Dropped full CI by 15+ minutes.
image


.github/workflows/build.yml
```diff
@@ -232,14 +232,44 @@ jobs:
...
+          name: e2e-build
+          path: build.tar.gz
+
+  e2e-test:
+    needs: [e2e-build]
+    name: e2e-test-shard-${{ matrix.shard }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        shard: [1, 2, 3, 4]
```
emallson, line 247: my only concern is whether we're going to run up against some github action limits, otherwise this looks like a pretty straightforward win to me 👍

ToppleTheNun: simc is currently doing 113 actions on every PR, so this should be okay. I think it gets a bit more worrying depending on how long these may run.

emallson: 👍 we'll find out, worst case we turn the e2e stuff off if it causes problems

PR #6062, opened 2023-05-10
[Shadow] Updates for more 10.1 talent changes

DoxAshe:
### Description

Many of shadow's talent's behavior changed even if the talent wasn't touched by the rework.  I've tested every talent to make sure I've updated each one to its new effect.  The major ones are:
- Death Speaker gives a full charge of SW:D instead of just restoring one on cooldown
- Shadowy Insight restores one cooldown of mind blast instead of giving a full charge.
- Auspicious Spirts generates Insanity differently
- Death and Madness no longer gives a buff that grants insanity, and instead grants it directly
- Idol of YoggSarron and Cthun have different spell ids
- Dark Evangelism is trigged by more spells
- Surge of Insanity does not have a refresh buff event, so I had to change it to track refreshes using devouring plague and its stacks.

There were also some abilities in SPELLS that were the same as they were in TALENTS.  I've removed them and replaced them with the talents, which should make things easier if shadow gets updated again in the future.

The explanations in the guide now match the new rotation for 10.1, so I've removed the spec from only being partially supported, as well as updated the example report to the new raid.

### Testing

There is still some strange behavior when Void Torrent is canceled early.  I think it just works differently now, so that still needs to be updated.


src/analysis/retail/priest/shadow/modules/talents/Deathspeaker.tsx
```diff
@@ -44,17 +47,41 @@ class Deathspeaker extends Analyzer {
...
+      false,
+    );
     this.lastProcTime = event.timestamp;
   }
 
   onBuffRemoved(event: RemoveBuffEvent) {
+    this.abilities.decreaseMaxCharges(event, TALENTS.SHADOW_WORD_DEATH_TALENT.id, 1);
+    if (this.spellUsable.chargesAvailable(TALENTS.SHADOW_WORD_DEATH_TALENT.id) === 1) {
+      // In certain circumstances, if you have 1 charge available after the buff, you can end up having negative charges spellUsable.onCooldown
+      // Resting the cooldown entirely fixes the issue.
```
emallson, line 77: poking @kfinch about this. this is a very unique case but still maybe something he knows about

src/common/SPELLS/priest.ts
```diff
@@ -836,14 +809,14 @@ const spells = spellIndexableList({
...
-    icon: 'spell_priest_mindspike',
+  IDOL_OF_CTHUN_RESOURCE: {
+    id: 377358,
+    name: 'Idol of CThun',
+    icon: 'achievement_boss_cthun',
   },
 
-  IDOL_OF_YOGG_SARON_VOID_SPIKE_CLEAVE_DAMAGE: {
-    id: 396895,
+  IDOL_OF_YOGG_SARON_DAMAGE: {
```
emallson, line 787: big 👍 for making sure to remove old stuff from SPELLS!

src/analysis/retail/priest/shadow/modules/talents/IdolOfCthun.tsx
```diff
@@ -22,17 +25,27 @@ class IdolOfCthun extends Analyzer {
...
+  onResource(event: ResourceChangeEvent) {
+    this.insanity += event.resourceChange;
+  }
+
   statistic() {
     return (
       <Statistic category={STATISTIC_CATEGORY.TALENTS} size="flexible">
         <BoringSpellValueText spellId={TALENTS.IDOL_OF_CTHUN_TALENT.id}>
           <ItemDamageDone amount={this.damage} />
+          <br />
```
emallson: please wrap in `<div />` if you want a piece of content to occupy its own line instead of inserting brs, which is typically more brittle.

(i don't always catch these in review, and it may never cause problems in this case, but it also makes things a bit easier in the long run if explicit breaks are avoided)

emallson (changes requested):
only one very minor change, but if you don't have time to get to it right now its small enough that we can merge just to get further spriest support in.

src/analysis/retail/priest/shadow/modules/talents/Deathspeaker.tsx
DoxAshe: This was happening with the Shadowy Insight before it was changed as well.  I don't know why it happens, but if you are at max charges when your charges are reduced it sometimes gives you a negative amount onCooldown.  I think having a check to make sure spellusable.OnCooldown doesn't allow for negative numbers would fix this, but I haven't looked at it.

src/analysis/retail/priest/shadow/modules/talents/IdolOfCthun.tsx
DoxAshe: I've updated all shadow's statistics to use `<div>` instead of `<br />`.
Should I switch to using `<div>` in other places as well?

src/common/SPELLS/priest.ts
DoxAshe: I've removed a few more unused spells and some that could just be done with TALENTS

src/analysis/retail/priest/shadow/modules/talents/IdolOfCthun.tsx
emallson: it is just that generally using explicit `<br/>`s can introduce unintended behavior when dealing with different screen sizes. telling the browser what content should be part of a single block (aka using a `<div />` or another element with `display: block;`) and then letting it handle line breaks typically works better across screen sizes.

emallson (approved):
looks good!

PR #6067, opened 2023-05-12
fix multiple rank talent tooltip links

ToppleTheNun:
### Description

Fix multiple rank talent tooltip links. I tried to account for specs that have the same talent ID but different definition IDs (e.g., Avenging Wrath).

### Testing

- Screenshot(s):
Before:
image

After:
image


emallson (approved):
this looks pretty reasonable

PR #6085, opened 2023-05-20
[Shadow] Update Void Torrent

DoxAshe:
### Description

Void Torrent insanity generation has been inaccurate since 10.1.  This fixes the issue, however, it can't distinguish between over capped insanity and the cast getting canceled early.  This is because Void Torrent doesn't generate any insanity waste, and doesn't have a resource change event a max insanity.

### Testing
I'm not sure if I've overlooked someway of getting the Insanity over capping of Void Torrent.


emallson (approved):
i think the approach of combining "wasted" with "lost due to cancellation" is pretty reasonable given the limitations of the log here

PR #6095, opened 2023-05-24
Add Insidious Ire Analysis to Shadow Priest

### LINKED ISSUE #6094: [SPriest] Insidious Ire Tracker
Would love to see an addition in the Statistics -> Talent section that highlights the DPS gain and percent of casts that are affected by Insidious Ire talent. Specifically: number of casts of Mind Blast and Void Torrent while the target is affected by all three dots.

I may take a look at if I could implement this, but it would be great if anyone who's familiar with the code already is able to just toss this in :)

comp615:
Closes #6094 

### Description

Insidious Ire plays a big part in the Shadow priest rotation. It can account for 5% or more of total damage, but requires that all 3 dots be applied when damage is done by Mind Blast or Void Torrent. This change aims to expose that piece of the rotation by showing the percent of hits that benefit from Insidious Ire.

Changes:
- Add a statistics block showing efficiency and DPS separately and combined
- Accounts for people (somehow) not talented into Void Torrent
- Adds an old-style "suggestions" block based on some random log perusing (I'm not a great SP, feel free to update but I think it's reasonable)

Open questions / TODOs:
- Was not sure exactly how internationalization is done. Do I need to lingui extract? Someone else?

### Testing

- Test report URL: `/report/qkDwh2TB3xyNrJ6P/44-Mythic+Magmorax+-+Kill+(5:29)/Retylx/standard/statistics`
- Screenshot(s):
image

cc @DoxAshe


DoxAshe:
This looks great!  I had been working on this myself, but this way of doing it is much better than my approach.  I can add this info into the guide view once this is merged if you would like me to.

I do think that the calculations of the damage gain is off.  You have a percent of the total damage, instead of a percent of the unincreased damage.  You can use 'calculateEffectiveDamage' to do this, or just change it from:

`damageGained: ired.reduce((sum, damage) => sum + damage.amount * this.insidiousIrePct, 0),`
to
`damageGained: ired.reduce((sum, damage) => sum + (damage.amount - (damage.amount / (1 + this.insidiousIrePct))), 0),`

comp615:
@DoxAshe Ahhh, smart. Thanks for the feedback! I've updated to use `calculateEffectiveDamage` since I love me a good reuse. I also added what I think is a decent "guide" entry. Feel free to commandeer and update the wording; I assume most of the other entries were written by you so it'd be good to have a consistent style and vision :)

DoxAshe:
That looks great!  I like what you wrote for the guide, I would leave it as.

src/analysis/retail/priest/shadow/modules/talents/InsidiousIre.tsx
```diff
@@ -0,0 +1,239 @@
...
+      );
+    }
+  }
+
+  ireDataForSpell(spell: SpellInfo) {
+    const damageInstances = this.eventHistory.getEvents(EventType.Damage, {
+      spell,
+    });
+
+    // Parition casts by if the target was affected by all three dots and thus would have been ired
```
trevorm4:
```suggestion
    // Partition casts by if the target was affected by all three dots and thus would have been ired
```

trevorm4 (approved):

PR #6109, opened 2023-06-07
[Shadow] Improve VoidBolt CD tracking

DoxAshe:
### Description

Void Bolt's cast efficiency has been incorrect.  Instead of using the average cooldown from getAbility, I track every update to its spell usable to get a more accurate calculation during each voidform.
This still uses Execute helper to make the graph hide the sections outside voidform.

### Testing

In the test report void bolt was cast 25 times, but the old method gives a maximum of 23.  With the new calculation, it correctly shows a maximum of 27.

- Test report URL: `/report/xy3RCn7w1FLpqmvV/4/Solaqt`

### DIFF
--- a/src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
+++ b/src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
@@ -2,7 +2,11 @@ import SPELLS from 'common/SPELLS';
 import Spell from 'common/SPELLS/Spell';
 import TALENTS from 'common/TALENTS/priest';
 import { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events from 'parser/core/Events';
+import Events, {
+  ApplyBuffEvent,
+  RemoveBuffEvent,
+  UpdateSpellUsableEvent,
+} from 'parser/core/Events';
 import Abilities from 'parser/core/modules/Abilities';
 import SPELL_CATEGORY from 'parser/core/SPELL_CATEGORY';
 import ExecuteHelper from 'parser/shared/modules/helpers/ExecuteHelper';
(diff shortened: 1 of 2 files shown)


src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
```diff
@@ -53,9 +77,73 @@ class Voidbolt extends ExecuteHelper {
...
+    //this adds timestamps of voidbolt spellusable updates
+    this.VB.push(event.timestamp);
+  }
+
+  onVBCast() {
+    this.castVB += 1;
+  }
+
+  //VB is an unusal spell. It is likely that using ExecuteHelper would be better than this for most spells.
+  getVB() {
```
emallson: would suggest a clearer title for this. `getX` methods usually just return `X`, but this builds the values of `miss`. maybe `calculateMissedVBs`?

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
```diff
@@ -53,9 +77,73 @@ class Voidbolt extends ExecuteHelper {
...
+      }
+    }
+
+    //find the average CD during this voidform
+    for (let i = 0; i < cd.length; i += 1) {
+      total = total + cd[i];
+    }
+    const averagecd = total / cd.length;
+
+    this.miss = this.miss + Math.floor(waste / averagecd); //Any remainder is not a possible cast, so it is floored.
```
emallson: I think this is wrong if you happen to call it twice. It will then process `this.VB` twice and produce (up to) double the missed value.

EDIT: it isn't but it relies on `enterVoidform` to reset the state. i would probably reset the `this.VB` field to an empty list at the end of this method as well as a precaution.

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
```diff
@@ -53,9 +77,73 @@ class Voidbolt extends ExecuteHelper {
...
+    for (let i = 0; i < diff.length; i += 1) {
+      if (i % 2 === 0) {
+        //even
+        waste = waste + diff[i];
+      }
+      if (i % 2 !== 0 && i !== diff.length - 1) {
+        //odd and not include final if it is odd.
+        cd.push(diff[i]);
+      }
+    }
```
emallson: this might be more cleanly implemented by directly building `waste` and `cd` than by building `diff`, then processing that.

```tsx
let waste = 0;
let castCount = 0;
let total = 0;

// build the waste amounts from even values, which represent the time between coming off cd and casting
for (let i = 2; i < this.VB.length; i += 2) {
  waste += this.VB[i] - this.VB[i - 1];
}

// build the cd duration from the odd values, which are the time between cast and coming off cd
for (let i = 3; i < this.VB.length; i += 2) {
  total += this.VB[i] - this.VB[i - 1];
  castCount += 1;
}

const averageCd = total / castCount;
```

emallson (changes requested):
the overall implementation looks functional, some code structure comments

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
DoxAshe: I have changed the name to make it clearer.

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
DoxAshe: I have done so to make sure it does not double count if its called twice.

src/analysis/retail/priest/shadow/modules/spells/Voidbolt.tsx
DoxAshe: I have changed it to this to make the calculation much simpler.

emallson (approved):

PR #6127, opened 2023-06-20
Update the Balance Druid module

Jundarer:
### Description

Gets rid of most inaccuracies and updates various modules to give better recommendations. Biggest updates are in the filler module. Still requires some more work to give fully correct recommendations and be able to spot all mistakes.

### Testing
Any Balance Druid log /report/twMgN3Zhkc9dPCnv/8-Normal+Echo+of+Neltharion+-+Kill+(3:07)/Tunderer/standard


kfinch (approved):
Looks like good cleanup to me

PR #6141, opened 2023-06-25
[Items] Correctly hide Voice of the Silent Star if not equipped

Pewtro:
On live it shows up like this when not equipped: 
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -25,6 +25,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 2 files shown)


emallson (approved):

emallson:
looks like you also need an early return because it tries to read a value from the cloak item that doesn't exist

Pewtro:
It's just grabbing the itemlevel and then calculation a stat value no? I can add it, just didn't look necessary on my local version

emallson:
oh hmm its crashing in a test, but that may actually just be because the test doesn't have a cloak mocked (not something that would actually occur in production)

PR #6142, opened 2023-06-25
[Project] Deduplicate yarn.lock

Pewtro:
Deduped through https://github.com/scinos/yarn-deduplicate


ToppleTheNun (approved):
Is it worth adding this to the "dependency version update" checklist?

PR #6152, opened 2023-06-29
[Enhancement Shaman] Guide updates, bug fixes

Seriousnes:
### Description

- I have created an index for all of the icons to a) add code-completion when importing icons, and b) make the imports look cleaner
- Improved enhancement APL
- Fixed some issues with the event order and link normalizers for handling Thorim's Invocation
- Refactored shaman files to replace `SpellLink` id with spell
- Removed lingu references from enhacement module

### Testing
- Test report URL: `report/Akj4mWNp162zrqRf/41-Mythic+Rashok,+the+Elder+-+Wipe+16+(6:32)/4-Seriousnes/standard`

### Screenshots
Cooldowns section added to guide
image


src/analysis/retail/shaman/enhancement/modules/talents/LegacyOfTheFrostWitch.tsx
```diff
@@ -119,9 +125,11 @@ class LegacyOfTheFrostWitch extends Analyzer {
   }
 
   get extraDamage() {
-    return Object.keys(this.buffedSpells)
-      .map((guid) => this.buffedSpells[Number(guid)])
-      .reduce((current, total) => (total += current));
+    const spellList = Object.keys(this.buffedSpells).map((guid) => this.buffedSpells[Number(guid)]);
+    if (spellList?.length > 0) {
+      return spellList.reduce((current, total) => (total += current));
```
ToppleTheNun: `reduce` calls should provide a default argument to help avoid some fun bugs.

```suggestion
      return spellList.reduce((current, total) => (total += current), 0);
```

src/analysis/retail/shaman/enhancement/modules/talents/Ascendance.tsx
```diff
@@ -0,0 +1,165 @@
...
+  onApplyBuff(event: ApplyBuffEvent | RefreshBuffEvent) {
+    if (event.type === EventType.ApplyBuff) {
+      this.current = {
+        event: {
+          ...event,
+          type: EventType.Cast,
+          sourceID: event.sourceID!,
+          __fabricated: true,
+          __modified: true,
+        },
```
emallson: to make sure i understood this right: you basically are mapping the `ApplyBuff` of a proc ascendance into the same type as a `Cast` of ascendance, right?

depending on what you need, it *might* be simpler to just apply a type of `CastEvent | ApplyBuffEvent` to the `current.event` field.

src/analysis/retail/shaman/enhancement/modules/core/SpellUsable.tsx
```diff
@@ -13,6 +15,17 @@ class SpellUsable extends CoreSpellUsable {
...
+  isAvailable(spellId: number): boolean {
+    const hasAscendance = this.selectedCombatant.hasBuff(TALENTS.ASCENDANCE_ENHANCEMENT_TALENT.id);
+    if (spellId === SPELLS.WINDSTRIKE_CAST.id && !hasAscendance) {
+      return false;
+    }
+    if (spellId === TALENTS.STORMSTRIKE_TALENT.id && hasAscendance) {
+      return false;
+    }
+    return super.isAvailable(spellId);
+  }
```
emallson: i *don't think* this impacts any of the spell availability metrics since they're checked at the end of the fight, but i'm not 100% on that.

it is technically correct and will produce possibly-useful debug logging so its fine to have.

emallson (approved):
couple of notes, but i think this looks fine

src/analysis/retail/shaman/enhancement/modules/talents/Ascendance.tsx
Seriousnes: I might revert/delete this entire file, it's not actually doing what I wanted it to do. The initial idea here was to treat a Deeply Rooted Elements proc as a cast of Ascendance when the module was extending `MajorCooldown`, but that had it's own set of issues.

emallson: gotcha

src/analysis/retail/shaman/enhancement/modules/core/SpellUsable.tsx
Seriousnes: This shouldn't have made it's way in either, I'll take it out.

PR #6154, opened 2023-06-30
[Evoker] Implemented handler for Empower Channeling

Krealle:
### Description

- Implemented Channeling handler for Empowers so they properly show their channel time on timeline.
- Added Disintegrate to channel list.
- Changed base gcd for empowers(Eternity Surge & Fire Breath) to better reflect actual gcd (should help with downtime tracking).
- Added PI to Devastation timeline.

 Screenshot(s):
image
image
image


src/parser/shared/normalizers/Channeling.ts
```diff
@@ -279,6 +304,73 @@ export function buffChannelSpec(spellId: number): ChannelSpec {
...
+          laterEvent.type === EventType.EmpowerEnd
+        ) {
+          endCurrentChannel(laterEvent, state);
+          break;
+        } else if (
+          laterEvent.type === EventType.BeginCast ||
+          (laterEvent.type === EventType.Cast &&
+            isRealCast(laterEvent) &&
+            laterEvent.timestamp > event.timestamp &&
+            event.sourceID === playerInfo.combatant.id)
```
ToppleTheNun: I don't think this is actually what you're after, `playerInfo.combatant.id` _should_ return `undefined` due to the implementation.

Krealle: You were correct, changed to `playerInfo.playerId` - now properly returns playerID.

emallson (approved):

PR #6162, opened 2023-07-03
[Core] Adjust Cast Efficiency calculations

Pewtro:
This change should alleviate some of the >100% cast efficiencies, since the last EndCooldown event was scheduled at the cooldowns expected end time (which could be past the fights duration) and that time from fight end till EndCooldown was inadvertently added to the efficiency calculation

I'm looking through some more logs at the moment to make sure this doesn't negatively impact anything since this has a pretty broad impact on spell efficiency across the board. 

There should be no impact on super short fights and long CDs: 
https://wowanalyzer.com/report/CjDqYaWp91PNvAT3/20-Normal+Kazzara,+the+Hellforged+-+Kill+(1:17)/Rouran/standard/statistics
Before:
image
After:
image

But it should be fixing >100% cast efficiency metrics: 
https://wowanalyzer.com/report/RafxBcAjqzmXnHF9/1-Heroic+Kazzara,+the+Hellforged+-+Kill+(2:14)/Munwell/standard/statistics
Before:
image
After:
image

Additionally to fixing >100% efficiencies, it also fixes some weird cast > maxCast instances: 
https://wowanalyzer.com/report/Cq4YzhMyXZWFmw6K/15-Mythic+Kazzara,+the+Hellforged+-+Kill+(3:31)/Pohner/standard/statistics
Before:
image
After: 
image

It will have minor impact on near 100% as well (Keg Smash) - but have them closer to their cast/maxCast counterpart. 
45/47 but 99% efficiency doesn't seem quite right but 45/47 and 97% efficiency seems much closer to accurate. 
https://wowanalyzer.com/report/zfZL42MA8Br1FKTQ/17-Mythic+Magmorax+-+Kill+(5:31)/Eisenpelz/standard/statistics
Before: 
image
After:
image


emallson (approved):
i think this looks reasonable. it is certainly simpler and easier to explain why the numbers are what they are

PR #6204, opened 2023-07-12
[hpal] add judgment to iol usage graph + mana gain from divine revelations

mfield:
test logs:
- https://www.warcraftlogs.com/reports/qZrTGNd8C4DPbzKm#fight=last&type=healing&source=19
image
image
- https://www.warcraftlogs.com/reports/m3v6Z8DrdK7MtVzb#fight=23&type=healing&source=15
image
image

PR #6205, opened 2023-07-13
[hpal] add Daybreak talent

mfield:
### Description
Adds Daybreak talent. Removes some misattributed healing from  Glistening Radiance.
Maybe conflicts with #6204 i can resolve once one is merged.

### Testing

live
image

this pr
image


abelito75:
I think my glimmer re-write kind of invalidates most of this
There are things that I don't fully cover such as mana but everything else should be covered in my other PR

emallson:
might need to PR yours @abelito75 and then compare notes?

abelito75:
Its this guy! #6206

It tracks more glimmer proc sources and has a great UI
It does miss out on the MP5 and CastLinkNormalizer but I think outside of that it does everything and fixes some bugs
We talked about it a bit in discord but I don't think we came to a decision

mfield:
let's figure out what we want in #6206 first. i left some notes there. but i'm not attached if this gets dropped nbd.

mfield:
taking this opportunity to break down per cast since most of the overview is in the Glimmer box
image

src/analysis/retail/paladin/holy/modules/talents/Daybreak.tsx
```diff
@@ -0,0 +1,124 @@
...
+      Events.cast.spell(TALENTS.DAYBREAK_TALENT).by(SELECTED_PLAYER),
+      this.onCast,
+    );
+  }
+
+  onCast(event: CastEvent) {
+    this.castEvents.push(event);
+
+    const manaEvents = GetRelatedEvents(event, DAYBREAK_MANA);
+    manaEvents.forEach((e: AnyEvent) => {
```
abelito75: IDK if its possible but is there any way we can not do `AnyEvent`... maybe `.filter()` before hand?
Maybe an ugly cast on the whole array?

src/analysis/retail/paladin/holy/modules/talents/Daybreak.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <th>Overhealing</th>
+              </tr>
+            </thead>
+            <tbody>
+              {Object.values(this.castEvents).map((cast, i) => (
+                <tr key={i}>
+                  <td>{this.owner.formatTimestamp(cast.timestamp)}</td>
+                  <td>{GetRelatedEvents(cast, DAYBREAK_MANA).length}</td>
+                  <td>
+                    {formatNumber(
```
abelito75: should probably do what you did for mana here and break it into functions. A lot easier to read IMO

abelito75 (changes requested):

src/analysis/retail/paladin/holy/modules/talents/Daybreak.tsx
mfield: harder to change the type here, but easy to change the next line to `if (e.type === EventTypes.ResourceChange)` which i think is nicer than the explicit cast. decent resolution?

abelito75: That works but also not the end of the world if you just wanna send it

abelito75 (approved):

PR #6206, opened 2023-07-13
[Holy Paladin] Glimmer re-write

abelito75:
### Description
A lot of the refreshing metrics are outdated
HoW said I could remove them
This is what we look like now

image


mfield:
I don't see anything concerning. Maybe I'm just seeing everything as a nail now that I know more about cast link normalizer but that's what I'd use to implement this rather than the lastCast tracking.

As an anecdote -- one aspect of my own play I've been looking at is why my mana felt sketchy but others are playing mercy + barrier. I think the answer is my daybreaks aren't reaping enough glimmers. So now I'm wondering why I'm not spreading glimmer effectively enough. So I'm a bit sad to lose the triggers/cast metrics on the old glimmer and glistening stats. This metric is also something that the cast link normalizer makes real, real easy to track (and to point out particularly bad instances).

abelito75:
The data for Procs/Trigger is in the tooltip... which isn't great I'll admit
I just found this UI so I don't know if we can just append that stat to the main visual of it or not

Last cast tracking is kind of janky


Personally, I think we should figure out which of the two PRs seems to be a better base, merge that one in, then reapply changes from the old one we want

AKA what is the least effort to re-implement lmao

The tooltips can be decently complex but in the end tooltips are never viewed so maybe adding another talent box for that info is smart 
Since at least in my keeping everything grouped is smart
image

PR #6210, opened 2023-07-14
[hpal] add Tyr's Deliverance

mfield:
Open to suggestions on what/how to present this info. e.g. was unsure between aggregate or split out direct/buffed healing in the main statistic.

image
image
https://www.warcraftlogs.com/reports/L743dxBZTW1nXY6w#fight=11&type=healing&source=20


src/analysis/retail/paladin/holy/modules/talents/TyrsDeliverance.tsx
```diff
@@ -0,0 +1,149 @@
...
+        ])
+        .by(SELECTED_PLAYER),
+      this.onBuffedHeal,
+    );
+  }
+
+  onApply(event: ApplyBuffEvent) {
+    this.casts += 1;
+    this.lastApply = event.timestamp;
+    console.log('GLIMDEBUG apply buff', this.owner.formatTimestamp(event.timestamp));
```
abelito75: No console.logs in prod

src/analysis/retail/paladin/holy/modules/talents/TyrsDeliverance.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  onRemove(event: RemoveBuffEvent | FightEndEvent) {
+    if (!this.selectedCombatant.hasBuff(TALENTS.TYRS_DELIVERANCE_TALENT.id)) {
+      return;
+    }
+    console.log(
```
abelito75: No console.logs in prod

src/analysis/retail/paladin/holy/modules/talents/TyrsDeliverance.tsx
```diff
@@ -0,0 +1,149 @@
...
+      Events.removebuff.spell([TALENTS.TYRS_DELIVERANCE_TALENT]).by(SELECTED_PLAYER),
+      this.onRemove,
+    );
+    this.addEventListener(Events.fightend, this.onRemove);
+
+    this.addEventListener(
+      Events.heal.spell([SPELLS.TYRS_DELIVERANCE_HEALING_INCREASE]).by(SELECTED_PLAYER),
+      this.onDirectHeal,
+    );
+    this.addEventListener(
```
abelito75: Would be nice to see this per spell

abelito75 (changes requested):
This seems to not divide out Boundless Salvation from the base Tyrs cast... I thought we wanted to be able to see both

src/analysis/retail/paladin/holy/modules/talents/TyrsDeliverance.tsx
mfield: whoops :(

src/analysis/retail/paladin/holy/modules/talents/TyrsDeliverance.tsx
mfield: image

mfield:
> This seems to not divide out Boundless Salvation from the base Tyrs cast... I thought we wanted to be able to see both

I'm not sure what you had in mind here.  For me, seeing how much I extended my Tyr's buff on average is enough for now. Can we call this good enough for a first pass and extend it later if we want to try and dive into something like "why'd you heal Alice when Bob had the buff and could have given you more extension?"

abelito75:
My brain was just 
Base 20 seconds of Tyrs goes to the Tyrs box so from 0 - 20
Extended 40 seconds goes to the Boundless box so from 20 - 60

No like `why didn't you extend it longer` or anything

Since you can't see that in logs with a ton of pain

But I'm also fine with merging this in and splitting it up later

mfield:
attempt 1 at splitting: i do kind of like how simple the separate boxes are. i don't like how physically separated they are. is there a way to force them next to or on top of each other?
image

abelito75:
I don't think so 
But if that is okay we cna merge it in

mfield:
yep let's send it then

PR #6212, opened 2023-07-14
Swap the names AnyEvent and MappedEvent

mfield:
### Description
Currently `AnyEvent` refers only to the subset of Events with subclasses listed in `MappedEventTypes`, while `MappedEvent` can refer to any `Event<T>` (but using the subclasses from `MappedEventTypes` when possible). This PR swaps those names to be less confusing.

Does not address any semantic errors in the code, this is only an automated refactor.

### Testing
typescript is happy


emallson (commented):
this is a very interesting PR. looking at it, there are cases i see that should clearly be using `AnyEvent` instead of `MappedEvent` (after swapping them).

there aren't any that I see that really *need* to be `MappedEvent` outside of dealing with type issues (i.e. needing to know that we *can* discriminate the union to produce an event type).

I wonder if it would be better to just replace this with `AnyEvent` everywhere? I feel like 99% of these uses would be more accurately reflected by that.

I'll probably try doing that refactor and see how angry `tsc` gets :sweat_smile:

mfield:
dropping in favor of #6223

PR #6216, opened 2023-07-16
[elemental] Add SK section

awildfivreld:
### Description

Add a section on proper usage of Stormkeeper.

Proper usage of Stormkeeper involves 3 criteria that should be met before casting the spell, then executing a correct cast order after pressing it. There is some wiggle room to when in the cast order you cast SK, which has been accounted for. I realize this PR (more specifically the SK module itself) might be on the larger side, but I figure getting SK right requires quite a bit of data and presentation.

Outside of the SK module itself I've fixed what I suspect to be a bug in the 'getRemainingBuffTimeAtApplication' function in core/Entities. Previously to this fix, it always returned "0" as the buff time at application/refresh. Now it returns a map with the refresh time as key, and buff duration (including pandemic window) as value. The previous implementation also did not seem to account for buff refreshes, so I've fixed that as well. If the old behavior is indeed 100% correct, then I need to go back to the drawing board on my usage of this function.

#### Mostly well executed window. Missing minor damage detail like Flameshock not having enough recommended duration left
/report/tnLg84dqxX1ZcTYC/15-Heroic+Rashok,+the+Elder+-+Kill+(3:33)/Fivreld/standard/overview
image

#### Cast lava burst before SK. This is valid. Same applies for casting the Elemental Blast before SK. 
/report/tnLg84dqxX1ZcTYC/15-Heroic+Rashok,+the+Elder+-+Kill+(3:33)/Fivreld/standard
image

#### Major error in the cast order
/report/tnLg84dqxX1ZcTYC/24-Heroic+Magmorax+-+Kill+(3:13)/Fivreld/standard
image


awildfivreld:
cc: @periodic

src/analysis/retail/shaman/elemental/constants.tsx
```diff
@@ -21,3 +21,7 @@ export const ON_CAST_BUFF_REMOVAL_GRACE_MS = 50;
  * in the guide section.  For use with `ExplanationAndDataSubSection`.
  */
 export const GUIDE_EXPLANATION_PERCENT_WIDTH = 40;
+
+export const ELECTRIFIED_SHOCKS_DURATION = 9000;
+export const ELEMENTAL_BLAST_COST = 75;
```
awildfivreld: Feedback from HawkCorrigan:

This should prolly be done correctly with EotS (talent that reduces the cost of these spells).

awildfivreld: Agree.

Did some codebase searching and found this analyzer called "SpellResourceCost". Using this is a much better solution than what I have here.

src/parser/core/Entity.ts
```diff
@@ -211,27 +211,32 @@ class Entity {
     sourceId: number | null = null,
   ): Map<number, number> {
     const buffTimesAtApplication: Map<number, number> = new Map<number, number>();
-    const sortedApplicationTimestamps: number[] = this.getBuffHistory(spellId, sourceId)
-      .map((buffEvent) => buffEvent.timestamp)
-      .sort((a, b) => a - b);
+    const sortedApplicationTimestamps: number[] = [];
+
+    this.getBuffHistory(spellId, sourceId).forEach((e) => {
+      sortedApplicationTimestamps.push(e.timestamp, ...e.refreshHistory);
```
periodic: This should probably be a map instead of using `push`.  Similar to the way it was written above.  The `forEach`/`push` pattern can cause subtle bugs and makes it less clear what the intent is.  `map` makes it clear we're just mapping elements one-to-one.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+  firstRotationCast: CastEvent;
+}
+
+// Spells that are considered to start the damage part of the SK window.
+const WINDOW_START_SPELLS = [
+  SPELLS.LIGHTNING_BOLT.id,
+  TALENTS.CHAIN_LIGHTNING_TALENT.id,
+  TALENTS.LAVA_BURST_TALENT.id,
+  TALENTS.EARTH_SHOCK_TALENT.id,
+  TALENTS.ELEMENTAL_BLAST_TALENT.id,
```
periodic, line 72: Probably should add Earthquake here too, in case someone is running with the talent where they alternate.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+
+    this.recordCooldown(this.activeWindow);
+    this.lastSKHardcast = null;
+    this.activeWindow = null;
+  }
+
+  onEventDuringSK(event: CastEvent) {
+    /* Don't include events that did not happen between hardcasts and the
+    SK buff falling off. */
+    if (!this.lastSKHardcast) {
```
periodic, line 176: [Try] Would this be simpler if we just checked for the buff?  You can use the grace-period argument to `hasBuff` to check even if the removal event was processed before the cast event.  That's what we do for Surge of Power (and probably should do for Master of the Elements since it does weird custom tracking)

```
this.selectedCombatant.hasBuff(
  TALENTS.STORMKEEPER_1_ELEMENTAL_TALENT.id,
  event.timestamp,
  ON_CAST_BUFF_REMOVAL_GRACE_MS
)
```

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+      return;
+    }
+
+    if (!this.activeWindow && WINDOW_START_SPELLS.includes(event.ability.guid)) {
+      this.activeWindow = {
+        event: this.lastSKHardcast,
+        maelstromOnCast: this.maelstromTracker.current,
+        flameshockDurationOnCast: this.enemies.getLongestDurationRemaining(
```
periodic: [Try] Shouldn't this be checking for the buff on the current target?  We could skip the whole 'longest duration' logic. Really it's that Flameshock should be on the LvB target and then all the other casts should be on the same target so they get EleShocks and LightningRod (if talented).

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+          end: -1,
+          events: [this.lastSKHardcast],
+          performance: QualitativePerformance.Perfect,
+        },
+        firstRotationCast: event,
+      };
+
+      /* If the user started the window with a spender, then the maelstrom
+      has already been consumed */
+      if (event.ability.guid === this.stSpender.id) {
```
periodic: [Do] I think this would read a lot better as a conditional in the object above.  Being down here makes it confusing because figuring out what the value of that field will be requires looking in two different places.  Additionally, can't we just look up the resource cost for any of them?  It'll just be missing if it didn't use maelstrom.

```
maelstromOnCast: this.maelstromTracker.current + (event.resourceCost[RESOURCE_TYPES.MAELSTROM.id] ?? 0),
```

If that doesn't work you can use a ternary expression like
```
maelstromOnCast: this.maelstromTracker.current +
  (event.ability.guid === this.stSpender.id && event.resourceCost && event.resourceCost[RESOURCE_TYPES.MAELSTROM.id])
    ? event.resourceCost[RESOURCE_TYPES.MAELSTROM.id]
    : 0,
```

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+        const resourceCost = event.resourceCost ? event.resourceCost : {};
+        this.activeWindow.maelstromOnCast += resourceCost[RESOURCE_TYPES.MAELSTROM.id] || 0;
+      }
+    }
+
+    if (this.activeWindow) {
```
periodic, line 221: [Do] if you switch the order of these two if-statements I think you can get a better flow of "if this is an active window do this, otherwise if it is the start of the window then start the window".  It just means you can get rid of the `!this.activeWindow` above.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -1,37 +1,264 @@
...
+      const spenderNotAlreadyCast = !this.activeWindow.timeline.events
+        .map((e) => e.ability.guid)
+        .includes(this.stSpender.id);
+
+      if (
+        event.ability.guid === this.stSpender.id &&
+        !this.selectedCombatant.hasBuff(
+          SPELLS.MASTER_OF_THE_ELEMENTS_BUFF.id,
+          event.timestamp,
+          100,
```
periodic: [Do] I added a constant for this so we can be consistent about how we use it across modules.  It's `ON_CAST_BUFF_REMOVAL_GRACE_MS`.  Also, that way it links back to a comment about why it's there.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -46,6 +273,239 @@ class Stormkeeper extends Analyzer {
...
+    const showPrecastAny = showPrecastSop || showPrecastMote;
+
+    const extraDetails = (
+      <div
+        style={{
+          overflowX: 'scroll',
+        }}
+      >
+        {showPrecastAny && (
+          <>
```
periodic: [Consider] Why do the other elements have `p` wrappers but this one doesn't?  Maybe it's a formatting thing, but it was just wondering if maybe you want it to be a paragraph for consistency.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -58,19 +518,48 @@ class Stormkeeper extends Analyzer {
...
+          <SpellLink spell={TALENTS.CHAIN_LIGHTNING_TALENT} />
+          s. Therefore, you want to combine this effect with as many of the other damage amplifying
+          effects you have at your disposal.
+        </p>
+        <p>
+          In short, the ideal cast order is the following: <br />
+          <small>For more information, see the written guides.</small>
+        </p>
+        <p>
+          (<SpellIcon spell={SPELLS.FLAME_SHOCK} /> -&gt;) (
```
periodic: [Try] You probably just want `&rarr;` here instead of `->` because it'll look better. :D

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -58,19 +518,48 @@ class Stormkeeper extends Analyzer {
...
-      />
+      <>
+        <p>
+          <strong>
+            <SpellLink spell={TALENTS.STORMKEEPER_1_ELEMENTAL_TALENT} />
+          </strong>{' '}
+          massively amplifies the strength of your next two
+          <SpellLink spell={SPELLS.LIGHTNING_BOLT} /> or
+          <SpellLink spell={TALENTS.CHAIN_LIGHTNING_TALENT} />
+          s. Therefore, you want to combine this effect with as many of the other damage amplifying
```
periodic: [Do] For consistency of pluralization you want either "Lightning Bolts or Chain Lightnings" or "Lightning Bolt or Chain Lightning Casts".

periodic (commented):
Wow that module gets long.  I'll admit to getting a little lost in all the functions.  I wonder if we could spend a little more time to trim it up a bit.

Otherwise I just left a bunch of smaller comments, mostly just style and structure.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: Good point. This logic originates from when I stored these values on SK hardcast, but since I'm now storing on a damaging cast event, I have a target.

src/parser/core/Entity.ts
awildfivreld: 'e.refreshHistory' here is an array of timestamps when the buff was refreshed. Since it seems that the intention of the method is to list duration times at all refreshes (I read "re-application" as refresh, but that might be wrong), then the loop is no longer a one-to-one, but one-to-potentially-many. 

Is there is a way to use .map, but potentially insert 1-N elements into the list?

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: I need to track the hardcast somewhere, as I need some way to distinguish if the SK buff is from the hardcast or T30 proc. It turns out that accounting for the T30 proc in this module is more work than I first anticipated, and therefore is planned for an upcoming PR.

Perhaps leave this check in for now, then see how it looks once we account for T30?

src/parser/core/Entity.ts
periodic: Oh, that's `Array.prototype.flatMap<T, U>((t: T) => U[]): U[]`.  It'll map each element to an array of elements and then flatten them all into one array.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
periodic: Ooh, right, the proc.  You are right.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: Went with the first suggestion.

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: Added "casts"

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: I originally did as you suggested, but then I discovered that it needs to be in the original order. 

In your suggestion, the first spell of the window won't be processed by the code that determines if the player cast the spells in the correct order. For example if the first ability is an EB, then the code won't check and correctly mark if MotE was present before window start.

Do you feel I should refactor the code to account for this, or is the original order fine considering this use-case?

src/parser/core/Entity.ts
```diff
@@ -211,27 +211,31 @@ class Entity {
...
         0,
         buffTimeAtLastApplication - timeDiffBetweenApplications,
       );
-      buffTimesAtApplication.set(sortedApplicationTimestamps[i], buffAmountAtCurrentApplication);
       lastApplicationTimestamp = sortedApplicationTimestamps[i];
       buffTimeAtLastApplication = Math.min(
         buffAmountAtCurrentApplication + baseBuffLength,
         maxBuffLength,
       );
+      buffTimesAtApplication.set(sortedApplicationTimestamps[i], buffTimeAtLastApplication);
```
emallson, line 238: my understanding of the changes to this file is:

- you added refreshes to `sortedApplicationTimestamps`, so responses from this method are now conceptually "times a buff would have been applied" and not just "times a buff was actually freshly applied"
- the change to the first apply to give it the full `baseBuffLength` remaining duration seems fine
- the change here handles pandemic-ing a buff.

is that correct?

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
```diff
@@ -46,6 +287,241 @@ class Stormkeeper extends Analyzer {
...
+      </div>
+    );
+
+    return { extraDetails, checklistItem };
+  }
+
+  /**
+   * Determine the performance of how much maelstrom the user had when casting SK
+   */
+  _determineMaelstromPerformance(maelstromRequired: number, cast: SKCast): QualitativePerformance {
```
emallson: these underscore-prefixed methods should probably be marked `private`

emallson (commented):
if my understanding of the `Entity` changes is correct, this is good to merge. I'm not going to be a stickler about getting `private` on the underscore methods, its a pretty minor detail.

src/parser/core/Entity.ts
awildfivreld: Yes. That is correct. From what I can tell, the one usage of this function (getRemainingBuffTimeAtTimestamp below) seems to be implemented with this logic in mind.

emallson (approved):

src/analysis/retail/shaman/elemental/modules/talents/Stormkeeper.tsx
awildfivreld: Ah. I come from python where we don't have such luxury. 

Fixed as suggested.

PR #6223, opened 2023-07-17
Rename MappedEvent to AnyEvent, clobber old AnyEvent

mfield:
Similar intent to #6212, but following @emallson's suggestion that probably all the old uses of `MappedEvent` actually wanted `AnyEvent`.


emallson (approved):
this appears to have typechecked, and I think probably generates less surprise and confusion (`AnyEvent` is easy to grok, `MappedEvent` requires understanding the internal event type mapping)

:+1:

PR #6224, opened 2023-07-17
use satisfies keyword instead of asRestrictedTable

ToppleTheNun:
### Description

`asRestrictedTable` is a [Constrained Identity Function](https://kentcdodds.com/blog/how-to-write-a-constrained-identity-function-in-typescript). `satisfies` basically makes it unnecessary.

babel and typescript-eslint updates are to support the new keyword.


mfield:
what a cool feature!

src/common/SPELLS/index.ts
```diff
@@ -35,31 +38,29 @@ import SHAMAN from './shaman';
...
-  ...MONK,
-  ...PALADIN,
-  ...PRIEST,
-  ...ROGUE,
-  ...SHAMAN,
-  ...WARLOCK,
-  ...WARRIOR,
-  ...DRAGONFLIGHT,
-} as const;
+const ABILITIES = safeMerge(
```
emallson, line 42: i assume this change is what pointed out the issue with evoker vs. dh glide?

src/common/maybeGetTalentOrSpell.ts
```diff
@@ -0,0 +1,14 @@
...
+export const maybeGetTalentOrSpell = (
+  key: string | number | undefined,
+  expansion = RETAIL_EXPANSION,
+) => {
+  const talent = maybeGetTalent(key);
+  if (talent) {
+    return talent;
+  }
+  return maybeGetSpell(key, expansion);
+};
```
emallson: this is going to be quietly broken if a talent exists that uses the same spell id as a classic spell. should be pretty rare but i would not be surprised to see some old priest/rogue/dk spells fall into this category.

should probably just skip the talent check if `isClassicExpansion(expansion)`

src/common/TALENTS/maybeGetTalent.ts
```diff
@@ -0,0 +1,38 @@
...
+  TALENTS_WARLOCK,
+  TALENTS_MONK,
+  TALENTS_DRUID,
+  TALENTS_DEMON_HUNTER,
+  TALENTS_EVOKER,
+);
+const InternalTalentTable = indexById<Talent, typeof TALENTS>(TALENTS);
+
+export const maybeGetTalent = (key: string | number | undefined) => {
+  return key ? InternalTalentTable[key as any] : undefined;
```
emallson, line 37: suprised this doesn't trigger the `no-explicit-any` lint (or do we have that turned off? :thinking:)

emallson (commented):
this looks good :+1: removes a good bit of the complexity from our spell table type machinery. downside is that it'll require a `yarn install` for people to get on the new version

src/common/SPELLS/index.ts
ToppleTheNun: Yep, it was just quietly broken before.

src/common/TALENTS/maybeGetTalent.ts
ToppleTheNun: It's probably buried somewhere in the current config, but I think it's disabled due to the amount of `: any` I see in the app.

emallson (approved):

PR #6238, opened 2023-07-20
[Augmentation] Implement support for prepull buffing

Krealle:
### Description

Augmentation when played optimally will precast their buffs (Prescience, and at times Ebon Might). I've added fabricated prepull events for Prescience and made sure Ebon Might goes on CD when it is casted pre-pull.
We need the pre-pull events from Prescience for proper analysis.

Also added Prescience to Cooldown graph, along with changing `CastEfficiencyBar` from `spellId` to `spell`.

- Test report URL: `/report/k4x6TM7K1XLG3p9V/16-Mythic+Echo+of+Neltharion+-+Kill+(7:01)/Vollmer/standard`
- Screenshot(s):
image
image
image


ToppleTheNun (approved):

Krealle:
@ToppleTheNun It seemed to run out of memory during the build, any idea why this is?

ToppleTheNun:
> @ToppleTheNun It seemed to run out of memory during the build, any idea why this is?

I've seen this happening more and more and I suspect it's due to the size of the project plus library versions. Updating lingui seems to help but has its own complications.

PR #6241, opened 2023-07-22
[Augmentation] Performance check & Normalization update

Krealle:
### Description

1. By demand from Saeldur the performance check for Buffs (Prescience, Ebon Might & Shifting Sands) will now be more ruthless.  
   - Added role/spec performance checks to enable checking if another Augmentation was buffed. Buffing another Augmentation is pretty much on par if not worse than buffing a healer.
   - Along with this is also a small refactor to Ebon Might & Shifting Sands module to make it more maintainable.
2. Updated Prescience pre-pull normalizer to also look for RefreshBuffEvents.
3. Added Ebon Might normalizer to fix an inconsistency much like what happens with Prescience, where extra BuffApply events are being pushed right before BuffRemove events, which results in us having 2x the amount of buffs out than we actually do.

### Testing

- Test report URL: `/report/k4x6TM7K1XLG3p9V/27-Mythic+Scalecommander+Sarkareth+-+Kill+(7:28)/Vollmer/standard`
- Screenshot(s):
image
image
image

### DIFF
--- a/src/analysis/retail/evoker/augmentation/modules/abilities/ShiftingSands.tsx
+++ b/src/analysis/retail/evoker/augmentation/modules/abilities/ShiftingSands.tsx
@@ -11,6 +11,7 @@ import { logSpellUseEvent } from 'parser/core/SpellUsage/SpellUsageSubSection';
 import Combatants from 'parser/shared/modules/Combatants';
 import ROLES from 'game/ROLES';
 import Combatant from 'parser/core/Combatant';
+import SPECS from 'game/SPECS';
 
 interface ShiftingSandsApplications {
   event: ApplyBuffEvent;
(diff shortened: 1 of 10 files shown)


src/analysis/retail/evoker/augmentation/modules/abilities/ShiftingSands.tsx
```diff
@@ -134,8 +161,13 @@ class ShiftingSands extends Analyzer {
         ),
       };
     }
+    return presciencePerformance;
+  }
 
+  private getRolePerformance(application: ShiftingSandsApplications) {
+    const className = application.combatant.spec?.className.replace(/\s/g, '') ?? '';
```
emallson, line 168: not a huge fan of generating a css class name on the fly like this, but it does work sooooo :thinking:

emallson (approved):
looks fine

PR #6246, opened 2023-07-23
update to lingui 3.17.2

ToppleTheNun:
### Description

Update to lingui 3.17.2 in preparation for an eventual move to lingui 4. This involved replacing any usages of the `t` macro
that weren't in the React context with `defineMessage`: https://lingui.dev/releases/migration-4#translation-outside-react-components-migration


emallson (approved):
i have not exhuastively checked every file but I am pretty confident in this PR nonetheless

PR #6249, opened 2023-07-24
[Core] Timeline Tooltips Updates

abelito75:
### Description

Currently our tooltip on the timeline is an `override` based system where a inefficient cast will override everything from an enhanced cast. When it comes to iconography this is fine but when it comes to the tooltip, this is an issue as you could be losing critical data such as buff consumption without realizing it.

The system also was very barebones in terms of supporting multiple modules feeding into the same fields on the events. The developer would have to manually set/update/combine tooltips every time they wished to add a tooltip. Most of the time this was avoided by just `not placing two on the same event` but this feels rather limiting for a simple fix.

Code Examples:
| Tooltip | Before | After |
|:-----:|:-----|:-----:|
| Enhanced Cast | event.meta = event.meta \|\| {}; <br /> event.meta.isEnhancedCast = true; <br /> event.meta.enhancedCastReason = "something"  | addEnhancedCastReason(event, "something") |
| Inefficient Cast | event.meta = event.meta \|\| {}; <br /> event.meta.isEnhancedCast = true; <br /> event.meta.enhancedCastReason = "something"  | addInefficenetCastReason(event, "something") |

The `Before` way, doesn't even go into combining events while the new does
You could easily write that code but doing it every single time is rather cumbersome

| Tooltip | Before | After |
|:-----:|:-----:|:-----:|
|None| image | image|
|Enhanced Cast|image|image|
|Inefficient Cast|image|image|
|Inefficient but Enhanced Cast| image|image|

I didn't update other parts of the code base to use this just use as I thought it would make the PR too cumbersome


src/parser/core/EventMetaLib.tsx
```diff
@@ -0,0 +1,39 @@
+import { CastEvent } from './Events';
+
+export const addInefficenetCastReason = (event: CastEvent, reason?: React.ReactNode) => {
```
ToppleTheNun:
```suggestion
export const addInefficientCastReason = (event: CastEvent, reason?: React.ReactNode) => {
```

ToppleTheNun (changes requested):

src/parser/core/EventMetaLib.tsx
abelito75: Good catch!

ToppleTheNun (approved):

PR #6260, opened 2023-07-26
[Devastation] Refactor of Causality module

Krealle:
### Description

Old implementation was kinda unmaintainable, and didn't give completely accurate CDR results.

Refactored the implementation, along with changing the way we calculate Pyre CDR, now gives a lot more accurate results.

Also updated statistics output to use Donut charts.

### Testing

- Test report URL: `/report/KnADgVdGWcakZmzr/11-Mythic+Assault+of+the+Zaqali+-+Kill+(4:31)/Daudormi/standard/statistics`
- Screenshot(s):
image

### DIFF
--- a/src/analysis/retail/evoker/devastation/modules/talents/Causality.tsx
+++ b/src/analysis/retail/evoker/devastation/modules/talents/Causality.tsx
@@ -25,9 +25,9 @@ class Causality extends Analyzer {
 
   enemyHitCounter: number[][] = [];
 
-  pyreCounters!: Map<number, number>;
   maxPyreCount: number = 5;
   previousPyreDamageEvent: number = 0;
+  pyreCounter: number = 0;
 
   eternitySurgeSpell = this.combatant.hasTalent(TALENTS.FONT_OF_MAGIC_DEVASTATION_TALENT)
     ? SPELLS.ETERNITY_SURGE_FONT
@@ -83,9 +83,9 @@ class Causality extends Analyzer {
    *
    * Due to all of this, it is fairly annoying to attribute damage events to any singular
    * Pyre.
-   * Solution used is to compare each damage in a given tick and see how much each individual
-   * enemy is hit, if they are hit multiple times we can safely assume it was from
-   * another unique Pyre cast.
+   * Solution used is to count amount of targets hit on a given tick, if the same target is
+   * hit twice, we assume it's a new pyre event and start to recount.
+   * We will also reset counter once a new hit later than the buffer amount happens.
    * From my testing this provides very accurate results.
    * There are potentially some very specific edgecases that would break this, but
    * so far I haven't experienced this. */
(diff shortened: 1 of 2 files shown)


src/analysis/retail/evoker/devastation/modules/talents/Causality.tsx
```diff
@@ -53,165 +64,164 @@ class Causality extends Analyzer {
...
-            (CDRAmount - this.spellUsable.cooldownRemaining(SPELLS.FIRE_BREATH_FONT.id)) / 1000;
+          const effectiveCDR = this.spellUsable.cooldownRemaining(spellId) / 1000;
+          const wastedCDR = (CDRAmount - this.spellUsable.cooldownRemaining(spellId)) / 1000;
+
+          source.CDR += effectiveCDR;
+          source.wastedCDR += wastedCDR;
         }
-        // Reduce the CD
-        this.spellUsable.reduceCooldown(SPELLS.FIRE_BREATH_FONT.id, CDRAmount);
+        this.spellUsable.reduceCooldown(spellId, CDRAmount);
```
mfield: i would let `reduceCooldown` do all the effective/wasted math.
```
const effectiveCDR = this.spellUsable.reduceCooldown(spellId, CDRAmount);
const wastedCDR = CDRAmount - effectiveCDR;
```

src/analysis/retail/evoker/devastation/modules/talents/Causality.tsx
```diff
@@ -53,165 +64,164 @@ class Causality extends Analyzer {
...
+   * In this example we now expect to have 4 seperate sets of damage events in perfect order.
+   * but with differing traveltime and Volatility to take into account we will have a
+   * mix match of events along with a, potentially, infinite extra sets of
+   * damage events. Each individual pyre *should* have all it's damage events in the same tick.
+   *
+   * Due to all of this, it is fairly annoying to attribute damage events to any singular
+   * Pyre.
+   * Solution used is to compare each damage in a given tick and see how much each individual
+   * enemy is hit, if they are hit multiple times we can safely assume it was from
+   * another unique Pyre cast.
```
mfield: one assumption i might try to add to this is to see if damage events are grouped by pyre instance (rather than being allowed to interleave). e.g. even if every event happens at the same millisecond would the logs look like this?
pyre A hits enemy 1
pyre A hits enemy 2
pyre A hits enemy 3
pyre B hits enemy 2  -- first repeated target, must be a new pyre
pyre B hits enemy 3
pyre B hits enemy 4

if so could simplify this logic a bit. i looked at a few p2 sark pulls and this looks like it would create a decent partitioning.

Krealle: Just so I'm sure I understand this correctly, you are suggesting to simply reset the counter if a repeat target is hit?

mfield: yeah, or if more than `buffer` time has passed. you'd know better than me if you've seen any data that contradicts this assumption though. i only looked at a very small number of logs to see if it was possible.

Krealle: Tried it out with a couple m+ logs, gives the same results as current iteration, thanks for the input! :)

mfield (approved):
LGTM

ToppleTheNun (approved):

PR #6262, opened 2023-07-26
[elemental] Add a section on the spender window

awildfivreld:
### Description

This adds a section on properly using the single target spender. I looked into expanding it for 2-3-4 targets, but I chose to delay that as each increase introduces subtle changes in the rotation.

### Testing

- Test report URL: `/report/tnLg84dqxX1ZcTYC/15-Heroic+Rashok,+the+Elder+-+Kill+(3:33)/Fivreld/standard`
- Screenshot(s):
image

Hovering on the @ 's reveal the times when the spender was cast.

### DIFF
--- a/src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
+++ b/src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
@@ -101,6 +101,9 @@ class SpenderWindow extends Analyzer {
     }
   }
 
+  /**
+   * Start a new spender window if one is not already active.
+   */
   onMSSpender(event: CastEvent) {
     if (this.activeSpenderWindow) {
       return;
@@ -118,6 +121,9 @@ class SpenderWindow extends Analyzer {
     };
   }
 
+  /**
+   * End a spender window if there is an active one.
+   */
   onSopConsumer(event: CastEvent) {
     if (
       !this.activeSpenderWindow ||
@@ -143,6 +149,14 @@ class SpenderWindow extends Analyzer {
     this.activeSpenderWindow = null;
   }
 
+  /**
+   * Create a single row for the spender window section.
+   * @param subWindow The windows that applies for this row.
+   * @param header The text preceeding the performance mark
+   * @param performanceThresholds Which thresholds to use to calculate performance
+   * @param windowTimestampCallable The fucntion to render the timestamps list for the window. Defaults to show '@ 0:01, 0:34' etc.
+   * @returns
+   */
   private spenderWindowsRow(
     subWindow: FinishedSpenderWindow[],
     header: JSX.Element,
--- a/src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
+++ b/src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
@@ -1,0 +1,295 @@
+import ThresholdPerformancePercentage, {
+  LTEThreshold,
+  GTEThreshold,
+} from './shared/ThresholdPerformancePercentage';
+import SPELLS from 'common/SPELLS';
+import TALENTS from 'common/TALENTS/shaman';
+import { Talent } from 'common/TALENTS/types';
+import { formatDuration } from 'common/format';
+import { SpellIcon, SpellLink, TooltipElement } from 'interface';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, { CastEvent } from 'parser/core/Events';
+import Enemies from 'parser/shared/modules/Enemies';
+import { ON_CAST_BUFF_REMOVAL_GRACE_MS } from '../../constants';
+
+interface ActiveSpenderWindow {
+  timestamp: number;
+  motePresent: boolean;
+  flameshockPresent: boolean;
+}
+
+interface FinishedSpenderWindow extends ActiveSpenderWindow {
+  elshocksPresent: boolean;
+  sopUse: CastEvent;
+}
+
+const SOP_SPENDERS = [
+  SPELLS.LIGHTNING_BOLT,
+  TALENTS.CHAIN_LIGHTNING_TALENT,
+  SPELLS.FLAME_SHOCK,
+  TALENTS.FROST_SHOCK_TALENT,
+  TALENTS.LAVA_BURST_TALENT,
+];
+
+const GOOD_SOP_SPENDERS = [SPELLS.LIGHTNING_BOLT.id, TALENTS.CHAIN_LIGHTNING_TALENT.id];
+
+const PERFECT_WINDOWS_THRESHOLD: GTEThreshold = {
+  type: 'gte',
+  perfect: 0.9,
+  good: 0.8,
+  ok: 0.7,
+};
+const WRONG_SOP_THRESHOLD: LTEThreshold = {
+  type: 'lte',
+  perfect: 0,
+  good: 0.1,
+  ok: 0.15,
+};
+const MISSING_ELSHOCKS_THRESHOLD: LTEThreshold = {
+  type: 'lte',
+  perfect: 0,
+  good: 0.2,
+  ok: 0.3,
+};
+const MISSING_MOTE_THRESHOLD: LTEThreshold = {
+  type: 'lte',
+  perfect: 0,
+  good: 0.2,
+  ok: 0.3,
+};
+const MISSING_FLAMESHOCK_THRESHOLD: LTEThreshold = {
+  type: 'lte',
+  perfect: 0,
+  good: 0.2,
+  ok: 0.3,
+};
+
+class SpenderWindow extends Analyzer {
+  static dependencies = {
+    enemies: Enemies,
+  };
+
+  enabled: boolean;
+
+  protected enemies!: Enemies;
+
+  activeSpenderWindow: ActiveSpenderWindow | null;
+  spenderWindows: FinishedSpenderWindow[];
+
+  private stSpender: Talent = TALENTS.EARTH_SHOCK_TALENT;
+
+  constructor(options: Options) {
+    super(options);
+
+    this.activeSpenderWindow = null;
+    this.spenderWindows = [];
+
+    if (this.selectedCombatant.hasTalent(TALENTS.ELEMENTAL_BLAST_TALENT)) {
+      this.stSpender = TALENTS.ELEMENTAL_BLAST_TALENT;
+    }
+
+    this.enabled = this.selectedCombatant.hasTalent(TALENTS.SURGE_OF_POWER_TALENT);
+
+    /* There is no point in tracking this at all if the player doesn't have SoP */
+    if (!this.enabled) {
+      return;
+    }
+
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(this.stSpender), this.onSpender);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SOP_SPENDERS), this.onSopSpender);
+  }
+
+  onSpender(event: CastEvent) {
+    if (this.activeSpenderWindow) {
+      return;
+    }
+
+    this.activeSpenderWindow = {
+      timestamp: event.timestamp,
+      motePresent: this.selectedCombatant.hasBuff(
+        SPELLS.MASTER_OF_THE_ELEMENTS_BUFF.id,
+        null,
+        ON_CAST_BUFF_REMOVAL_GRACE_MS,
+      ),
+      flameshockPresent:
+        this.enemies.getEntity(event)?.hasBuff(SPELLS.FLAME_SHOCK.id, event.timestamp) || false,
+    };
+  }
+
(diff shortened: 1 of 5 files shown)


awildfivreld:
cc: @periodic

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+  activeSpenderWindow: ActiveSpenderWindow | null;
+  spenderWindows: FinishedSpenderWindow[];
+
+  private stSpender: Talent = TALENTS.EARTH_SHOCK_TALENT;
+
+  constructor(options: Options) {
+    super(options);
+
+    this.activeSpenderWindow = null;
+    this.spenderWindows = [];
```
periodic: [Nit] You might just have these initializers happen on the fields themselves.  I think it's generally idiomatic directly initialize fields that can be.  Not entirely sure what the style in this codebase is though.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+    if (this.selectedCombatant.hasTalent(TALENTS.ELEMENTAL_BLAST_TALENT)) {
+      this.stSpender = TALENTS.ELEMENTAL_BLAST_TALENT;
+    }
+
+    this.enabled = this.selectedCombatant.hasTalent(TALENTS.SURGE_OF_POWER_TALENT);
+
+    /* There is no point in tracking this at all if the player doesn't have SoP */
+    if (!this.enabled) {
```
periodic: [Nit] It might read better if you swapped this conditional and put the `addEventListener` calls inside, instead of having an early return.  The early return could add bugs if someone added more initialization and didn't notice this here.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+  motePresent: boolean;
+  flameshockPresent: boolean;
+}
+
+interface FinishedSpenderWindow extends ActiveSpenderWindow {
+  elshocksPresent: boolean;
+  sopUse: CastEvent;
+}
+
+const SOP_SPENDERS = [
```
periodic: [Nit] I got a little confused that there are two different types of spenders in this file, maelstrom spenders and SOP spenders.  You might consider changing the the name of one.  At a glance `onSpender` / `onSopSpender` and `sopSpender` / `stSpender` look pretty similar.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(this.stSpender), this.onSpender);
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SOP_SPENDERS), this.onSopSpender);
+  }
+
+  onSpender(event: CastEvent) {
+    if (this.activeSpenderWindow) {
+      return;
+    }
+
+    this.activeSpenderWindow = {
```
periodic, line 108: [Nit] There's another way to write this with the null-coalescing operator.  Not sure if you like that more or less.

```typescript
this.activeSpenderWindow = this.activeSpenderWindow ?? {
  ...
};
```

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+      flameshockPresent:
+        this.enemies.getEntity(event)?.hasBuff(SPELLS.FLAME_SHOCK.id, event.timestamp) || false,
+    };
+  }
+
+  onSopSpender(event: CastEvent) {
+    if (!this.activeSpenderWindow) {
+      return;
+    }
+    if (
```
periodic, line 124: [Do] Probably just combine these two if-statements since the body is the same.  Makes it clear they are doing the same thing and puts all the guards in one expression.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { ON_CAST_BUFF_REMOVAL_GRACE_MS } from '../../constants';
+
+interface ActiveSpenderWindow {
+  timestamp: number;
+  motePresent: boolean;
+  flameshockPresent: boolean;
+}
+
+interface FinishedSpenderWindow extends ActiveSpenderWindow {
+  elshocksPresent: boolean;
```
periodic, line 23: [Consider] Might be at least worth adding a TODO to note that we should also add Lightning Rod here later.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+      sopUse: event,
+    });
+    this.activeSpenderWindow = null;
+  }
+
+  private spenderWindowsRow(
+    subWindow: FinishedSpenderWindow[],
+    header: JSX.Element,
+    performanceThresholds: LTEThreshold | GTEThreshold,
+    windowTimestampCallable: (windows: FinishedSpenderWindow[]) => JSX.Element[] = (windows) =>
```
periodic, line 160: [Consider] If this is never used, why have it here as a default argument?  I guessing this was left over for something.  Might be worth a comment.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,295 @@
...
+            to account for this yourself if relevant.
+          </small>
+        </p>
+      </>
+    );
+
+    const hasMote = this.selectedCombatant.hasTalent(TALENTS.MASTER_OF_THE_ELEMENTS_TALENT);
+    const hasElshocks = this.selectedCombatant.hasTalent(TALENTS.ELECTRIFIED_SHOCKS_TALENT);
+
+    type WindowBreakdown = {
```
periodic, line 204: TIL that types don't have to be top-level.

periodic (commented):
Overall it looks decent. I had a bunch of little nits you can take if you like them.  I would love to see more doc comments though.  I think they would help make it clear what's going on (and it's nice that they show up in my ID on hover)

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: Lightning rod will always be present, as it is applied by the EB.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: It value is overridden in the inefficient SoP row to add the spell that was actually used to the tooltip.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: Moved

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: Moved as suggested

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: I found that in another PR of this repo, but I decided against using it as AFAIK it isn't really a common operator outside TS. I tend to like to be explicit, so the code reads "If there is an active window, don't do anything in this function". I feel '??=' requires some interpretation to reach my intent.

Can of course use it if it is an established pattern in TS.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: Applied as suggested

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
awildfivreld: Renamed MS related spenders to include MS in the name, and renamed SoP spenders to SoP consumers.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,308 @@
...
+  private stMSSpender: Talent = TALENTS.EARTH_SHOCK_TALENT;
+
+  constructor(options: Options) {
+    super(options);
+
+    if (this.selectedCombatant.hasTalent(TALENTS.ELEMENTAL_BLAST_TALENT)) {
+      this.stMSSpender = TALENTS.ELEMENTAL_BLAST_TALENT;
+    }
+
+    this.enabled = this.selectedCombatant.hasTalent(TALENTS.SURGE_OF_POWER_TALENT);
```
emallson: this should be `this.active`. if you use that, `addEventListener` is a no-op so you don't need the conditional (but there may be other initialization that you need to skip)

```suggestion
    this.active = this.selectedCombatant.hasTalent(TALENTS.SURGE_OF_POWER_TALENT);
```

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,308 @@
...
+  good: 0.2,
+  ok: 0.3,
+};
+
+class SpenderWindow extends Analyzer {
+  static dependencies = {
+    enemies: Enemies,
+  };
+
+  enabled: boolean;
```
emallson:
```suggestion
```

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,308 @@
...
+    super(options);
+
+    if (this.selectedCombatant.hasTalent(TALENTS.ELEMENTAL_BLAST_TALENT)) {
+      this.stMSSpender = TALENTS.ELEMENTAL_BLAST_TALENT;
+    }
+
+    this.enabled = this.selectedCombatant.hasTalent(TALENTS.SURGE_OF_POWER_TALENT);
+
+    /* There is no point in tracking this at all if the player doesn't have SoP */
+    if (!this.enabled) {
```
emallson:
```suggestion
    if (!this.active) {
```

emallson (changes requested):
minor change to use the existing `active` flag instead of `enabled` for this module

you might want to look at using either the cast breakdown accordions (seen on rdruid [here](https://wowanalyzer.com/report/ja4D7nXkh6bJLPyg/16-Mythic+Rashok,+the+Elder+-+Kill+(5:59)/Flemingo/standard)) or cast breakdown boxes (seen on brew [here](https://wowanalyzer.com/report/ja4D7nXkh6bJLPyg/16-Mythic+Rashok,+the+Elder+-+Kill+(5:59)/Eisenpelz/standard)) in order to show specific spender windows, but that doesn't need to be in this PR

(also sorry this lingered so long without a followup review!)

awildfivreld:
> minor change to use the existing `active` flag instead of `enabled` for this module
> 
> you might want to look at using either the cast breakdown accordions (seen on rdruid [here](https://wowanalyzer.com/report/ja4D7nXkh6bJLPyg/16-Mythic+Rashok,+the+Elder+-+Kill+(5:59)/Flemingo/standard)) or cast breakdown boxes (seen on brew [here](https://wowanalyzer.com/report/ja4D7nXkh6bJLPyg/16-Mythic+Rashok,+the+Elder+-+Kill+(5:59)/Eisenpelz/standard)) in order to show specific spender windows, but that doesn't need to be in this PR
> 
> (also sorry this lingered so long without a followup review!)

I did use the cast breakdown one before, but then I realized there are potentially **many** (0-40) instances, where a substantial amount of them might have an inefficiency. I may experiment with individual breakdowns of these in the future.

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -0,0 +1,306 @@
...
+  good: 0.2,
+  ok: 0.3,
+};
+
+class SpenderWindow extends Analyzer {
+  static dependencies = {
+    enemies: Enemies,
+  };
+
+  active: boolean;
```
emallson: this is now shadowing the parent class' `active` field. i don't *think* it causes issues since JS is late bound, but you should be able to just delete this declaration.

```suggestion
```

emallson:
> I did use the cast breakdown one before, but then I realized there are potentially many (0-40) instances, where a substantial amount of them might have an inefficiency. I may experiment with individual breakdowns of these in the future.

That makes sense :+1:

emallson (approved):

PR #6265, opened 2023-07-27
[core/haste] Add astral flare tracking

awildfivreld:
### Description

Adds the Sarkareth haste buff to the Haste module, so that it is accounted for, and stop a flood of console errors when looking at the fight.


awildfivreld:
Not sure if its my editor or pre-commit that doesn't like the spacing in the changelog, but I can try to remove the unecessary changes in it if you want.

emallson:
that looks like prettier running, should be fine

emallson (approved):

PR #6270, opened 2023-07-29
use wclClass/SpecName for filtering on character parses

emallson:
Followup to previous one. Rather than trying to match on the label, just bluntly supply keys to use.


Pewtro:
Remember demo locks and 👍🏼

Pewtro (approved):

PR #6278, opened 2023-08-02
Implement buffRemaining APL condition

Pewtro:
### Description

Adds support for a `buffRemaining` APL check condition. 

It can be used like so: 

```ts
buffRemaining(SPELLS.STEADY_FOCUS_BUFF, 15000, { atMost: 5000, atLeast: 2000 })
```

By default it will not accept a buff that is missing as valid, even if you pass in only atMost is passed in as the range argument even if technically no buff is less than an arbitrary value. 
If that is however something you want to return true, you can send in a 4th argument of `true`, like so: 
```ts
buffRemaining(SPELLS.STEADY_FOCUS_BUFF, 15000, { atMost: 5000 }, true)
```
If you were to pass in any (positive) atLeast value, it would fail regardless of the 4th argument being true or not. 
However, generally, I would suggest you use buffMissing in combination with this rule if you find yourself needing to pass in true.

This intentionally doesn't look at RemoveBuffStackEvent, as that _generally_ doesn't refresh the duration of the buff nor does it remove the buff completely, that is handled by RemoveBuffEvent.  

It does **NOT** handle pandemic'ing buffs.


src/parser/shared/metrics/apl/conditions/buffRemaining.tsx
```diff
@@ -0,0 +1,53 @@
...
+    },
+    validate: (state, event) => {
+      if (state === null) {
+        //Buff is not applied so buff has no remaining time so only atMost can be true
+        return Boolean(missingCanBeValid && range.atMost && !range.atLeast);
+      }
+      const timeRemaining = state.timestamp + buffDuration - event.timestamp;
+      return (
+        timeRemaining >= (range.atLeast || 0) && (!range.atMost || timeRemaining <= range.atMost)
+      );
```
emallson, line 44: this doesn't handle buffs that pandemic correctly. i don't think that is an immediate goal for the current use-case, but you could swap the `buffDuration` to be a `PandemicData` object instead of a number, and then use the `buffDuration` helper in `./util.tsx`

emallson (commented):
this seems fine on a technical level. will leave it up to you whether to merge w/o dealing with pandemic stuff.

PR #6279, opened 2023-08-03
rapture analysis now ignores off gcd spells

Tapir42:
"Rapture analysis" module now ignores off gcd spells like Fade, Pots, or enchants (healing darts).


oisinod (approved):
Nice spot! 👍

oisinod:
The pipelines won't pass until you make a changelog entry though, @Tapir42 😄

PR #6284, opened 2023-08-05
[Core] EventLinknormalizer Re-added events Fix

abelito75:
### Description
The current EventLinkNormalizer would from time to time on initial load randomly fire twice and re-add the same events to the _linkedEventsArray

This causes an issue where statistics that depend on this just are 2x the value as expected

You can test this with applying a Time Filter as it will retrigger event normalization (for some reason) causing the bug

Personally, I don't like this approach as it just adds a lot of un-needed looping of events but I couldn't think of another way to do this without making events themselves more complex

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -27,6 +27,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
+  change(date(2023, 8, 4), <>Added additional checks to make sure EventLinkedNormalizer doens't re-add the same events.</>,  Abelito75),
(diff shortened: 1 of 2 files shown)


trevorm4:
Are they the actual same event objects? If so, could we add a field to the event object to mark it as linked instead of doing this calculation maybe?

abelito75:
They are the same object references so 100% yes

But where would you add the field? Since we want to make sure we can still link any given Event to any other given Event even if its linked to another Event, we just want to prevent duplicate links

IE
Event A, B, C

// case 1
A is linked to B
A is linked to C
// valid

// case 2
A is linked to B
B is linked to C
// valid

// case 3
A is linked to B
A is linked to B
// invalid

abelito75:
Closing in favor of https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6292

PR #6288, opened 2023-08-07
begone, SpellLink#id

ToppleTheNun:
### Description

Remove `SpellLink#id`, `SpellIcon#id`, `BoringSpellValue#spellId`, and `BoringSpellValueText#spellId`.


emallson (approved):
core changes look fine. only skimmed the resulting changes but if it typechecks that should be good

PR #6292, opened 2023-08-08
[Core] Fix double processing of event links

trevorm4:
We have an issue where we process an event multiple times in the eventlinknormalizer. This PR adds a field to the event to mark it as processed so we don't accidentally process it multiple times. 

Before:

image

After

image


abelito75:
LGTM
But since its core should probably get Topple/Emallson ✅ on it

ToppleTheNun (approved):
I'm in favor of this approach due to the difficulty in pinning down _why_ exactly we get double links sometimes.

PR #6296, opened 2023-08-08
fix eventlink bug and dont break everything this time

trevorm4:
### Description

Caches which eventlinks we've processed for a specific event to avoid double prcoessing an event link.

### Testing

Tested fix on MW and verified that disc/brew did not break manually

- Screenshot(s):
- Before
image

After
image

### DIFF
--- a/src/parser/core/Events.ts
+++ b/src/parser/core/Events.ts
@@ -5,6 +5,7 @@ import * as React from 'react';
 import EventFilter from './EventFilter';
 import { PetInfo } from './Pet';
 import { PlayerInfo } from './Player';
+import { EventLink } from './EventLinkNormalizer';
 
 export enum EventType {
   Destroy = 'destroy', // super rare, apparently happens on dausegne, no idea what it is?
@@ -330,6 +331,8 @@ export interface Event<T extends string> {
   prepull?: boolean;
   /** Other events associated with this event. Added by WoWA normalizers. Meaning is context sensitive */
   _linkedEvents?: LinkedEvent[];
+  // true if links have been processed for this event to prevent duplicate linking
+  _processedLinks?: Set<EventLink>;
   /** True iff the event was created by WoWA */
   __fabricated?: boolean;
   /** True iff the event's content was modified by WoWA */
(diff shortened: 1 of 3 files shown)


src/parser/core/Events.ts
```diff
@@ -330,6 +331,8 @@ export interface Event<T extends string> {
   prepull?: boolean;
   /** Other events associated with this event. Added by WoWA normalizers. Meaning is context sensitive */
   _linkedEvents?: LinkedEvent[];
+  // true if links have been processed for this event to prevent duplicate linking
```
emallson: [very tiny grammar nit]
```suggestion
  // contains the links that have been processed for this event to prevent duplicate linking
```

emallson: oh you already fixed it, damn

trevorm4: Yea didn't update comment when i first linked but it should be good now

src/parser/core/Events.ts
```diff
@@ -330,6 +331,8 @@ export interface Event<T extends string> {
   prepull?: boolean;
   /** Other events associated with this event. Added by WoWA normalizers. Meaning is context sensitive */
   _linkedEvents?: LinkedEvent[];
+  /** Set of eventLinks that have been processed already to avoid duplicated linking */
+  _processedLinks?: Set<EventLink>;
```
emallson, line 335: i am a little torn on this being a `Set<EventLink>` vs a `Set<string>`. relying on object identity has some awkwardness around object destructuring / spreading, but i don't think people generally use that here? and it would require some very weird call sequence to have the same exact normalizer (not two unique files, but one file) run twice in the first place

`string` doesn't have those issues but does have potential conflict issues with the relation keys. however, i think those issues already exist regardless of using `string` for this so it might be cleaner to treat `EventLink.linkRelation` as the "primary key".

trevorm4: I think using EventLink is likely more precise because you can have 2 different event links with the same `linkRelation` but different `reverseLinkRelation` which means you would need 2 strings in order to uniquely identify it. 

WRT to it running twice, I'm not sure that it runs twice but I think that its receiving the same event multiple times. In the case of MW, you can apply a time filter and then it will feed in the same event multiple times and mess up the vivify count breakdown significantly.

emallson: that is reasonable

PR #6297, opened 2023-08-08
[Classic] Character page

jazminite:
### Description

- Add Classic Character page
- Add `size` parameter to `interface WCLParse `
- Add `CLASSIC_DIFFICULTIES` to game difficulties
- Add Classic Zones file for raids

### Testing

**NOTE**: Search on the main page is still hidden because it's also linked to guild search.

Functionality is there if you navigate directly.

- Test URL: `http://localhost:3000/character/US/Benediction/Jazminites?game=classic`
- Screenshot(s):
image


emallson (commented):
am i correct in understanding that the new `CharacterParses` and `CharacterParsesList` components are *mostly* duplicated from the retail version? It looks like there is a reliance on `const`s (e.g. the `ZONES` import, the default zone) that make the current one not quite usable, but there is so much duplication that it seems we *should* be able to simply set a `game="classic"` prop on the base `CharacterParses` and get the desired behavior

it would likely involve changing the access from consts to instance methods that check the prop, but it seems worth not to duplicate ~900 lines of code.

jazminite:
> am i correct in understanding that the new `CharacterParses` and `CharacterParsesList` components are _mostly_ duplicated from the retail version? It looks like there is a reliance on `const`s (e.g. the `ZONES` import, the default zone) that make the current one not quite usable, but there is so much duplication that it seems we _should_ be able to simply set a `game="classic"` prop on the base `CharacterParses` and get the desired behavior
> 
> it would likely involve changing the access from consts to instance methods that check the prop, but it seems worth not to duplicate ~900 lines of code.

Yep, I wasn't quite sure how to use the same pages for both Retail and Classic, so I made duplicates.
If you have any guidance, please let me know.

emallson:
I can take a look, without digging into it I'm not sure my suggestions would be super useful :sweat_smile:

emallson (commented):
@jazminite i merged the classic components into the retail ones, didn't end up being too bad. in the process i found a bug with how we handled encounter selection (it operated off of name instead of id), which i fixed since it impacted ToGC

emallson:
(if the changes look OK to you, you or I can merge it)

jazminite:
> (if the changes look OK to you, you or I can merge it)

Your changes look good. I added a fix for loadRealms to avoid the console warning. Let me know if that's good.

emallson (approved):

PR #6299, opened 2023-08-08
[Shadow] Improve Inescapable Torment Talent

DoxAshe:
### Description

With the most recent balance patch, shadow's best build has changed to include the Inescapable Torment Talent.  This adds additional analysis to the talent in the guide view and breaks down the amount of mindbender extension the talent gives per cast instead of only the total extension.
This also fixes the talent showing more extension than it should when hitting multiple targets.

### Testing

It is difficult to track the duration of mindbender since it is a pet, so I directly calculated the amount of extension based upon each damage instance of inescapable torment.  I think using the next cast or end of fight is the only way to tell which damage events go with what cast of mindbender, but if there is a better way to do this I can change it.

- Test report URL: `/report/T9cKJWvgy7V2fD4n/1-Heroic+Kazzara,+the+Hellforged+-+Kill+(2:19)/Mollify/standard/overview`
- Screenshot(s):

InescapleTorment

### DIFF
--- a/src/analysis/retail/priest/shadow/modules/talents/InescapableTorment.tsx
+++ b/src/analysis/retail/priest/shadow/modules/talents/InescapableTorment.tsx
@@ -36,9 +36,7 @@ class InescapableTorment extends Analyzer {
     this.addEventListener(Events.fightend, this.onEnd);
   }
 
-  onCast(event: CastEvent) {
-    //Since there is no way to tell when mindbender ends, we resolve the previous Mindbender when a new one is cast.
-    //If this is not the first mindbender cast, then we generate the extension.
+  private finalizeMindBenderCast() {
     if (this.castTime !== 0) {
       const tooltip = (
         <>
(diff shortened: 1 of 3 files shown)


src/analysis/retail/priest/shadow/modules/talents/InescapableTorment.tsx
```diff
@@ -21,13 +29,74 @@ class InescapableTorment extends Analyzer {
...
+        value = QualitativePerformance.Ok;
+      }
+      if (this.extension <= 3) {
+        value = QualitativePerformance.Fail;
+      }
+
+      this.MBExtension.push({ value, tooltip });
+    }
+
+    this.totalTime += this.extension; // add previous extension to total time.
```
emallson: instead of a copy/paste between this and `onCast`, you could put them in a shared method (e.g. `protected finalizeMindBenderCast`). `onCast` needs a little more because it then sets up a new mind bender cast, but they're mostly shared.

emallson (approved):
looks fine, one minor style comment, but if you don't want to / don't have time to touch that right now we can just merge

src/analysis/retail/priest/shadow/modules/talents/InescapableTorment.tsx
DoxAshe: I've changed it to use a protected method.

emallson (approved):

PR #6300, opened 2023-08-09
disable augvoker M+ analysis

ToppleTheNun:
### Description

Disable Augmentation Evoker analysis in M+ due to issues with retrieving events in a timely manner.

### Testing

- Test report URL: `/report/Hd38L4fCYXjZ7yVR/10-Mythic++Neltharion's+Lair+-+Wipe+1+(41:28)`
- Screenshot(s):
image

### DIFF
--- a/src/interface/report/PlayerTile.tsx
+++ b/src/interface/report/PlayerTile.tsx
@@ -16,6 +16,9 @@ import { useWaDispatch } from 'interface/utils/useWaDispatch';
 import { useWaSelector } from 'interface/utils/useWaSelector';
 import { makeThumbnailUrl } from 'interface/makeAnalyzerUrl';
 import { useLingui } from '@lingui/react';
+import SPECS from 'game/SPECS';
+import { isFightDungeon } from 'common/isFightDungeon';
+import { useFight } from 'interface/report/context/FightContext';
 
 interface Props {
   player: Player;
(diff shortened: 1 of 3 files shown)


src/interface/report/PlayerTile.tsx
```diff
@@ -103,6 +107,41 @@ const PlayerTile = ({ player, makeUrl, config }: Props) => {
...
+    return (
+      <span
+        className="player"
+        onClick={() => {
+          alert(
+            `Augmentation Evoker is not currently supported for M+ logs due to issues with retrieving the appropriate data for analysis. Augmentation is still supported for raid and we hope to re-enable it for M+ soon.`,
+          );
+        }}
+      >
+        <div className="role" />
```
emallson: rather than duplicating the contents, this could be done by defining a parent component, then picking a parent based on the spec:

```tsx
const BlockLoading = (message) => ({ children }) => <span className="player" onClick={() => alert(message)}>{children}</span>;

let Container = ({ children}) => <Link {...}>{children}</Link>;

if (spec.id === SPECS.AUGMENTATION_EVOKER.id && isFightDungeon(fight)) {
  Container = BlockLoading(`...`);
}

return (
  <Container>
    <div className = "role" />
     <!-- ... -->
   </Container>
 );
 ```

emallson (commented):
only qualm is in duplicating the contents of the element. otherwise seems fine. (could also do this without extra components by putting the content in a var and then wrapping, e.g. `return condition ? <span>{children}</span> : <Link>{children}</Link>;`)

ToppleTheNun:
Should be much less duplication now.

PR #6303, opened 2023-08-10
[Augmentation] Implement Breath of Eons pre-pull normalizer

Krealle:
### Description
Breath of Eons was causing issues when being cast pre-pull, due to lack of Cast event. I've added a normalizer that fabricates the missing Cast event.

This should fix the issues being cause with the Breath of Eons module.

### Testing

- Test report URL: `/report/TBmZYxjR1p2kgKtJ/1-Heroic+Echo+of+Neltharion+-+Kill+(5:05)/20-Findabattu/standard`


emallson (commented):
technically seems fine, but could this use the core `PrePullNormalizer` instead? (maybe issues with it being aoe, applying buffs and debuffs?) if not, should add a comment explaining why

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsRotational.tsx
```diff
@@ -227,14 +227,20 @@ class BreathOfEonsRotational extends Analyzer {
      * Due to other Augmentations apparently also wanting to use Breath of Eons alongside you :) */
     const flightEvents = getBreathOfEonsBuffEvents(event);
     const flightData: SpellTracker[] = [];
+    let debuffHits = getBreathOfEonsDebuffApplyEvents(event).length;
+    // if breath was from pre-cast we won't neccesarily have the proper links.
+    // So set this to 1 to make graph look proper
+    if (debuffHits === 0) {
```
ToppleTheNun: Will this also be true if you just straight up miss?

Krealle: This is used to decide the size of the area being drawn on the graph for the time spent flying, so doesn't really affect anything besides making sure it actually has a "size" beyond 0.

Krealle:
@emallson Think I missed `PrePullNormalizer ` being a thing. Will check it out after my keys :)

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsRotational.tsx
Krealle: To add to this `temporalWoundsCounter` in `breathPerformance` array is used to determine whether or not a Breath hit.

Krealle:
Hmm I can get it to create the prepull cast event using `PrePullNormalizer`, but it doesn't include any `CastLinks`. Is there a way to generate those after the fact?

emallson:
Probably? you could add a rule to your link normalizer for it. however, if it is more complicated than that you might just want to use your custom normalizer rather than pre-pull normalizer *and* a custom one.

Krealle:
And how would you go about adding such a rule? Not sure I've seen it yet.

emallson:
hmm, this might just ("just") be a normalizer ordering issue. you could `console.log` both normalizers and see if they run in the order you want. i think you want the `PrePullCooldownsNormalizer` to run before your `EventLinkNormalizer`, right? if so, you can adjust the load order by setting the `priority` property (lowest goes first)

ToppleTheNun:
All modules also default to `0` priority so you may need to make the `EventLinkNormalizer` a `1`.

Krealle:
So if I understand it correctly, In my `CastLinkNormalizer` I add this to the constructor? I have a feeling I got it wrong, because it isn't changing the ordering around.
```
class CastLinkNormalizer extends EventLinkNormalizer {
    super(options, EVENT_LINKS);
    this.priority = 1;
  }
}
```

ToppleTheNun:
does this work?
```tsx
export default class SpiritBombEventLinkNormalizer extends EventLinkNormalizer {
  priority = 1;
  
    super(options, EVENT_LINKS);
  }
}
```

Krealle:
No, the ordering is still the same with that :(

emallson:
hmm, if you push your changes I can poke around a bit and see whats up

ToppleTheNun:
I just did some poking in VDH land and here's something interesting I found:
```
normalizer running: fightEndNormalizer (0)
normalizer running: energizeCompat (6)
normalizer running: applyBuffNormalizer (7)
normalizer running: cancelledCastsNormalizer (8)
normalizer running: phaseChangesNormalizer (9)
normalizer running: missingCastsNormalize (10)
normalizer running: channeling (11)
normalizer running: shearFractureNormalizer (45)
normalizer running: immolationAuraLinker (46)
normalizer running: soulCleaveEventLinkNormalizer (47)
normalizer running: spiritBombEventLinkNormalizer (48)
normalizer running: defensiveBuffLinkNormalizer (49)
normalizer running: theHuntNormalizer (50)
normalizer running: felDevastationNormalizer (51)
normalizer running: sigilOfFlamesNormalizer (52)
normalizer running: prepullNormalizer (113)
```

Priority is assigned based on where it's loaded in the order: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/parser/core/CombatLogParser.tsx#L413

The options passed to `loadModule` always override the priority value we set in the class: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/parser/core/CombatLogParser.tsx#L371-L379

ToppleTheNun:
Those are Shadowlands links but the same code exists in the Dragonflight branch.

ToppleTheNun:
So we might need to make it _exclude_ priority in that override if we want people to be able to set their own priority values.

emallson:
oh, that is interesting

ToppleTheNun:
So the order that you specify modules in `specModules` affects priority. :)

emallson:
it feels like an antipattern to set `priority` at all as part of the options, but not sure what we'd break changing it. at the very least, we can have the base class not overwrite priority. lemme poke around with the commit that just got pushed and see if that produces the expected behavior

emallson:
image
ok, after some core changes and further bumping the priority, it is running the `CastLinkNormalizer` after the `PrePullCooldownsNormalizer`

Krealle:
Ah that is awesome! I'll push in the changes I need, then you can just merge whenever you feel it's ready.

Krealle:
Good to go on my end.

PR #6306, opened 2023-08-12
Large update to abilities file for blood dk

yajinni:
Added many missing abilities and talents, including checks for the talents before they are displayed.
First step in getting bdk updated for DF.
Will be reviewing the suggestions next.


ToppleTheNun (changes requested):
Changelog and this is good to go.

PR #6308, opened 2023-08-12
Fix/classic shadow priest Vampiric touch uptime

lavjamanxd:
### Description
Vampiric Touch had the same uptime percent as Devouring Plague, because a wrong spellID was used in the analyzer.

### Testing
- Test report URL: any report which has a classic shadow priest in


emallson (approved):
good catch! thanks for the fix :)

PR #6312, opened 2023-08-13
[elemental] Add abc graph

awildfivreld:
### Description

Add a section and graph on always be casting.

Also wraps the resource graph in a dropdown. 

### Testing

- Screenshot(s):
image


awildfivreld:
cc: @periodic

emallson (approved):
this is a step in a good direction. we don't currently have the data to support "active time in last N seconds" but that would be a good progression i think? (or maybe the chart would be illegible, hard to say)

PR #6315, opened 2023-08-16
[core] update resourcetracker include tooltip and optional wasted line

awildfivreld:
### Description

Improves the core ResourceTracker to include a tooltip with information at the timestamp under the cursor, and an optional boolean flag to enable a line for wasted resources. You can enable the wasted line by setting `includeWasted = true` on the class. This defaults to `False`, and must be manually enabled in each graph.

You can change the color of the wasted line by overriding `get wastedColor()`. 

### Testing

There are currently 15 specs that subclass the ResourceTracker module at least once.

* Frost DK
* Havoc DH
* Vengeance DH
* Balance druid
* Feral druid
* Guardian druid
* Devestation evoker
* BM hunter
* Shadow priest
* All rouge specs (in your `shared/`)
* Elemental shaman
* Enhancement shaman
* Demo warlock

I'm including one visual example in this, but the same functionality is currently live in the elemental shaman analyzer if you want see it for yourself. https://wowanalyzer.com/report/L1R2zVKY8ZDxCTNb/15-Mythic+Scalecommander+Sarkareth+-+Kill+(7:21)/Nerdshockz/standard/overview#prototype

Before:
image

After, without wasted line:
image

After, with wasted line:
image

### DIFF
--- a/src/parser/shared/modules/ResourceGraph.tsx
+++ b/src/parser/shared/modules/ResourceGraph.tsx
@@ -3,6 +3,7 @@ import Analyzer from 'parser/core/Analyzer';
 import BaseChart, { formatTime } from 'parser/ui/BaseChart';
 import { AutoSizer } from 'react-virtualized';
 import { VisualizationSpec } from 'react-vega';
+import { Color } from 'vega';
 
 abstract class ResourceGraph extends Analyzer {
   /** Implementer must override this to return the ResourceTracker for the resource to graph */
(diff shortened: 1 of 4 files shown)


ToppleTheNun (approved):
LGTM for Havoc/Vengeance/Assassination.

emallson (approved):
can't speak for the spec maintainers, but in terms of code this looks fine.

Seriousnes (approved):
Enhancement Shaman ✅

src/parser/shared/modules/ResourceGraph.tsx
```diff
@@ -8,12 +8,26 @@ abstract class ResourceGraph extends Analyzer {
...
+  includeWasted: boolean = false;
+
   /** Implementer may override this to give the graph line a custom color.
    *  Color must be in format '#rrggbb', where rr gg and bb are hex values. */
   lineColor(): string | undefined {
     return undefined;
   }
 
+  /** Implementer may override this to give the wasted line a custom color. */
+  wastedColor(): string {
```
Pewtro: Should this be a template literal instead of simply a string to aid implementers? 
Something like 
```
`rgba(${number},${number},${number},${number})`
```

Pewtro (commented):
Minor thing but otherwise :+1:

src/parser/shared/modules/ResourceGraph.tsx
ToppleTheNun: I could see wanting to support hex colors as an option too.

Pewtro: If the lib supports it just add it as a union? Dunno what is supported by whatever takes in the opt

awildfivreld: I see that vega has a "Color" type, but which basically just are strings with aliases. I did some digging, but it seems like a proper hex type completely blows up the TS type checker. 

I've extended the docstring and changed `string` to `Color` instead, which hopefully should inform users of what is expected.

awildfivreld:
I'd say this can be merged now?

emallson:
i think so

PR #6322, opened 2023-08-21
[WIP][Frost Mage] 10.1.5 Update

Sharrq:
Work in Progress. Do Not Merge.

Updating Frost Mage for 10.1.5

### DIFF
--- a/src/analysis/retail/mage/frost/talents/GlacialSpike.tsx
+++ b/src/analysis/retail/mage/frost/talents/GlacialSpike.tsx
@@ -5,11 +5,13 @@ import SPELLS from 'common/SPELLS';
 import TALENTS from 'common/TALENTS/mage';
 import { SpellLink } from 'interface';
 import { TooltipElement } from 'interface';
-import Analyzer, { SELECTED_PLAYER, Options } from 'parser/core/Analyzer';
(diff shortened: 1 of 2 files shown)


src/analysis/retail/mage/frost/talents/GlacialSpike.tsx
```diff
@@ -18,92 +20,52 @@ class GlacialSpike extends Analyzer {
...
-    }
-  }
+    //If Glacial Spike was shattered on the main target, disregard it
+    badCasts = badCasts.filter((gs) => {
+      const spikeDamage = this.eventHistory.getEvents(EventType.Damage, {
+        searchBackwards: false,
+        spell: SPELLS.GLACIAL_SPIKE_DAMAGE,
+        count: 1,
+        startTimestamp: gs.timestamp,
+        targetID: gs.targetID,
```
emallson, line 51: this should also pass `targetInstance` or if you glacial spike something that has multiple instances, you'll find a bunch of possibly-unrelated matches. also, if you're GSing something that has multiple instances and you don't provide an instance, your current implementation won't match (for example: using GS on the void mobs in P2 of Neltharion. each has a defined instance ID, but they all share the same target ID. the current code (probably) won't find those events)

Sharrq:
Closing this out, i have a different branch with a full rewrite of most of the core frost modules, will PR that one instead.

PR #6323, opened 2023-08-21
Improved Garrote + Sepsis interaction implementation

juemrami:
### Description
- Changed the proper stealth spell to use while combatant has the Sepsis buff active.
- Added Sepsis buff to list of buffs for the Improved Garrote snapshot specification.
- Added Sepsis  to `Abilities.tsx`.
- removed Garrote cooldown value from `Abilities.tsx` to hide the cast efficiency bar since it's not necessary to use Garrote on cooldown

### DIFF
--- a/src/CONTRIBUTORS.ts
+++ b/src/CONTRIBUTORS.ts
@@ -77,7 +77,7 @@ export const Fashathus: Contributor = {
     },
   ],
 };
-export const juemrami: Contributor = {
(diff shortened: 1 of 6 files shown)


ToppleTheNun (changes requested):
Add a changelog and this is good to go. I have a couple of other Sepsis changes from Whispyr to add in once this is in.

ToppleTheNun (approved):

PR #6334, opened 2023-08-27
[Elemental] Add electrified shocks subsection

awildfivreld:
### Description

Adds subsection analyzing electrified shocks usage. 

Also does some minor visual fixups in a couple other modules. 

### Testing

image

### DIFF
--- a/src/analysis/retail/shaman/elemental/modules/talents/Icefury.tsx
+++ b/src/analysis/retail/shaman/elemental/modules/talents/Icefury.tsx
@@ -1,27 +1,110 @@
+import { formatDuration } from 'common/format';
 import TALENTS from 'common/TALENTS/shaman';
-import { SpellLink } from 'interface';
+import { Expandable, SpellLink } from 'interface';
+import { PerformanceMark, SectionHeader } from 'interface/guide';
+import CooldownExpandable, {
+  CooldownExpandableItem,
+} from 'interface/guide/components/CooldownExpandable';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events from 'parser/core/Events';
+import Events, { ApplyBuffEvent, RefreshBuffEvent, RemoveBuffEvent } from 'parser/core/Events';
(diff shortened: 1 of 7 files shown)


src/analysis/retail/shaman/elemental/modules/talents/Icefury.tsx
```diff
@@ -38,19 +111,13 @@ class Icefury extends Analyzer {
...
-  }
-
   suggestions(when: When) {
     when(this.suggestionThresholds).addSuggestion((suggest, actual) =>
       suggest(
         <>
           You should fully utilize your <SpellLink spell={TALENTS.ICEFURY_TALENT} /> casts by
           casting 4 <SpellLink spell={TALENTS.FROST_SHOCK_TALENT} />s before the{' '}
-          <SpellLink spell={TALENTS.ICEFURY_TALENT} /> buff expires. Pay attention to the remaining
+          <SpellLink spell={TALENTS.ICEFURY_TALENT} /> buff expires.Pay attention to the remaining
```
emallson, line 120: this change looks like a typo (should have a space after the period?)

emallson (commented):
on a technical level this seems fine. i have some design comments:

- I think this is probably the wrong component to use for a spell that is cast every 30s. The boxes style is likely more suited to this frequency. You can combine this with a detailed breakdown on click, similar to expanding these (see [brew](https://wowanalyzer.com/report/dQK73G29RgpcVJZz/7-Mythic+Rashok,+the+Elder+-+Kill+(6:06)/Eisenpelz/standard) for example; though I have a CSS issue to fix I see because something blew out the scale :upside_down_face:)

- If you do want to stick to this style, I would not separate perfect usages out like this because it makes it harder to follow the information. Perfect casts are just casts that you managed exceptionally well, they are not a separate category of casts (IMHO)

awildfivreld:
Yeah when I took a more critical look at it I realized that I approached this quite wrong, hence the draft status of it. 

I have a much revamped version, coming probably tomorrow (giving theorycrafters some time to comment on the new version), that incidentally have addressed both your comments.

awildfivreld:
Will re-create. Feel that is cleaner than re-using this one.

PR #6336, opened 2023-08-29
disable ESLint during build step

ToppleTheNun:
### Description

we already run ESLint during a previous step, this should speed the build step up.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,14 +27,14 @@ jobs:
       - name: Get yarn cache directory path
         if: steps.cache.outputs.cache-hit != 'true'
         id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -27,14 +27,14 @@ jobs:
       - name: Get yarn cache directory path
         if: steps.cache.outputs.cache-hit != 'true'
         id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+        run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
```
ToppleTheNun, line 30: This change is due to the old method being deprecated by GitHub: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

emallson (approved):

PR #6339, opened 2023-08-30
[elemental] Add electrified shocks subsection

awildfivreld:
### Description

Expands the Icefury analyzer, to support an electrified shocks analyzer.

Also other smaller touchups.

### Testing

image


emallson (approved):
there are some interesting elements here that I might explore too :thinking: 

also a minor note about anonymous JSX functions: they tend to not do good things to React's ability to skip re-renders since each render call will see new functions. however, the stuff here should basically never re-render so it doesn't need touching.

PR #6341, opened 2023-08-31
[Augmentation] Fix undefined issue in Shifting Sands module

Krealle:
### Description

Some guy found a target with undefined buffs, now we default to `false` if that happens again.

### Testing

- Test report URL: `/report/qymghkvXH93fat4x/18-Mythic+Scalecommander+Sarkareth+-+Wipe+18+(7:30)/Dragoproject/standard/overview`

### DIFF
--- a/src/analysis/retail/evoker/augmentation/CHANGELOG.tsx
+++ b/src/analysis/retail/evoker/augmentation/CHANGELOG.tsx
@@ -5,6 +5,7 @@ import TALENTS from 'common/TALENTS/evoker';
 import SPELLS from 'common/SPELLS/evoker';
 
 export default [
+    change(date(2023, 8, 31), <>Fix issue with <SpellLink spell={SPELLS.SHIFTING_SANDS_BUFF} /> module when targets buffs wasn't defined.</>, Vollmer),
     change(date(2023, 8, 30), <>Fix issue with <SpellLink spell={TALENTS.EBON_MIGHT_TALENT} /> module when buff target spec wasn't defined.</>, Vollmer),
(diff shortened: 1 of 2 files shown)


emallson (approved):
there's probably an underlying issue to fix there, but this is fine to just make it stop erroring

Krealle:
@emallson guess you were right :)) any chance you could send a link to the log that is giving the error?

ToppleTheNun:
@Krealle https://wowanalyzer.com/report/t6Nv7xz9XJy8DAdq/22-Heroic+Balakar+Khan+-+Kill+(0:53)/Modria/standard

ToppleTheNun:
Simplest "fix" is to just put a `?` after target:
```
  private onCast(event: ApplyBuffEvent) {
    const target = this.combatants.players[event.targetID];

    const shiftingSandsApplication: ShiftingSandsApplications = {
      event: event,
      ebonMightOn: target?.hasBuff(SPELLS.EBON_MIGHT_BUFF_EXTERNAL.id) ?? false,
      prescienceOn: target?.hasBuff(SPELLS.PRESCIENCE_BUFF.id) ?? false,
      combatant: target,
    };

    this.shiftingSandsApplications.push(shiftingSandsApplication);
  }
```

Doesn't really handle the possible error cases from `combatant: target` being undefined, though.

Krealle:
hmm, one of the players in the log is not in the combatant list, when I use `console.log(this.combatants.players);` I only get 4/5 players returned. The missing one being the one that is causing issues.

Krealle:
Seems like it's completely broken for that player actually.
image

ToppleTheNun:
This is included in that player's combatantinfo event in both the V1 API and the V2 API responses:
```json
{
  "error": "expected int but found [P]"
}
```

Not sure what that error means exactly, but it's why the character is blocked.

PR #6348, opened 2023-09-05
[Shadow] Add Active Time Graph

DoxAshe:
### Description

This copies the active time graph from elemental shaman (https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6312#issue-1848706987) and adds it to the shadow priest guide.

### Testing

- Screenshot(s):
ATG


src/analysis/retail/priest/shadow/modules/features/getUptimeGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+interface GraphData {
+  timestamp: number;
+  activeTimePercentage: number;
+}
+
+const GRAPH_SAMPLE_COUNT = 2; // Show value for every Nth GCD in the graph.
+const GRAPH_MAX_Y = 1; // The maximum value on the Y axis. Used to prevent the graph skyrocketing.
+
+export default function getUptimeGraph(
```
emallson: my only comment is that if this is a straight copy of the ele graph, we should prob share it. other than that, this is good

DoxAshe: I think I've moved getUptimeGraph to a shared location for both Ele and Shadow priest.

emallson (approved):

PR #6353, opened 2023-09-06
Fix Elemental Blast for Enhancement shamans

Pewtro:
Blizz only half-way split the talents, the talent itself is the enhancement one but all casts and damage events are elemental shaman talent ids. weeee

Patches getRepeatedTalentCount to match the method of detection to be similiar to hasTalent so we don't get different results between the two.

Fixes a few Stormkeeper issues that I noticed while double checking new implementation of getRepeatedTalentCount


src/analysis/retail/shaman/enhancement/modules/Abilities.tsx
```diff
@@ -413,7 +413,7 @@ class Abilities extends CoreAbilities {
       },
       {
         // This is no error. We actually use the elemental shaman elemental blast spell id.
-        spell: TALENTS_SHAMAN.ELEMENTAL_BLAST_ENHANCEMENT_TALENT.id,
+        spell: TALENTS_SHAMAN.ELEMENTAL_BLAST_ELEMENTAL_TALENT.id,
         enabled: combatant.hasTalent(TALENTS_SHAMAN.ELEMENTAL_BLAST_ENHANCEMENT_TALENT),
         charges:
           combatant.getRepeatedTalentCount(TALENTS_SHAMAN.ELEMENTAL_BLAST_ENHANCEMENT_TALENT) +
```
ghost: This line should also use the elemental version

ghost (changes requested):

src/analysis/retail/shaman/enhancement/modules/Abilities.tsx
Pewtro: Why? The talent id is correct since it's used in hasTalent - so why would it not be accurate in getRepeatedTalentCount?

Pewtro: getRepeatedTalentCount behaviour has been patched, should be accurate now.

src/parser/core/Combatant.ts
```diff
@@ -132,7 +132,9 @@ class Combatant extends Entity {
    * talents (like Empower Rune Weapon or Stormkeeper).
    */
   getRepeatedTalentCount(talent: Talent): number {
-    return this._combatantInfo.talentTree.filter((entry) => entry.spellID === talent.id).length;
+    return this._combatantInfo.talentTree.filter((entry) =>
+      talent.entryIds.some((talentEntryId) => talentEntryId === entry.id),
+    ).length;
   }
```
emallson: unless i'm reading this wrong, this doesn't work. `ELEMENTAL_BLAST_ELEMENTAL_TALENT` has entry ids `[101866]` while `ELEMENTAL_BLAST_ENHANCEMENT_TALENT` has entry ids `[101831]` so it will always be 0 if you e.g. use the `ELEMENTAL` version when you're looking at `ENHANCEMENT`.

however i think this method is just...dead...unless we want to match on names. each Stormkeeper node now has distinct spell and entry ids, same for Empower Rune Weapon.

Pewtro: This makes it so you don't have to use the "wrong" class talent when counting your own talent. It correctly sets the charges for elemental now when they pass in the elemental version, as opposed to if we leave it unchanged, they have to pass in the enhancement variant (because that spellid matches). 

But it might be dead, in that case a bit of refactoring is probably needed in the usage places

emallson: oh i see, that makes sense. this isn't restoring the old behavior fully, this is making it not completely broken. a `@deprecated` flag in the docstring wouldn't hurt if we wanna move away from it.

Pewtro: Would the recommendation be to use getTalentRank in something similiar to this:
```ts
charges: getTalentRank(TALENT_A) + getTalentRank(TALENT_B)
```

emallson: yea that is what people were using before we added this. kinda icky but could maybe make it more palatable by changing it to something like:

```ts
getRepeatedTalentCount(...talents: Talent[]) {
     return talents.reduce((sum, talent) => sum + this.getTalentRank(talent), 0);
}
```

then you call with:

```ts
charges: getRepeatedTalentCount(TALENT_A, TALENT_B)
```

or we can just go back to ye olde +, i'm fine either way

Pewtro: getRepeatedTalentCount is no more and instead replaced by 
```ts
  /**
   * Return the number of ranks that the provided talents have in total.
   * Useful for repeated talents (like Empower Rune Weapon or Stormkeeper).
   */
  getMultipleTalentRanks(...talents: Talent[]): number {
    return talents.reduce((count, talent) => count + this.getTalentRank(talent), 0);
  }
  ```

emallson (approved):

PR #6354, opened 2023-09-07
Add mana tea statistic

trevorm4:
### Description

Adds statistic and scaffolding for guide part (but not guide section quite yet). Also disables lifecycles for now

### Testing

- Test report URL: `report/ghkWqFVA8tBP92c1/50-Mythic+Magmorax+-+Kill+(4:54)/Livingmist/standard/statistics`
- Screenshot(s): 
image
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/ManaTea.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/ManaTea.tsx
@@ -12,6 +12,7 @@ import Events, {
   ApplyBuffEvent,
   CastEvent,
   HealEvent,
+  RefreshBuffEvent,
   ResourceChangeEvent,
 } from 'parser/core/Events';
 import { ThresholdStyle, When } from 'parser/core/ParseResults';
@@ -58,6 +59,7 @@ class ManaTea extends Analyzer {
   effectiveHealing: number = 0;
   manaPerManaTeaGoal: number = 0;
   overhealing: number = 0;
+  stacksWasted: number = 0;
   manaRestoredSinceLastApply: number = 0;
   protected abilityTracker!: AbilityTracker;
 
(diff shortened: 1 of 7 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/ManaTea.tsx
```diff
@@ -100,7 +118,7 @@ class ManaTea extends Analyzer {
     }
 
     if (
-      this.selectedCombatant.hasBuff(TALENTS_MONK.MANA_TEA_TALENT.id) &&
+      this.selectedCombatant.hasBuff(SPELLS.MANA_TEA_BUFF.id) &&
```
vohrr, line 132: I know it didn't have this before but this rework is the perfect time to make sure we account for other mana reductions in the kit that currently skew this metric - chiji stacks, yulon, CF, TFT vivs etc

vohrr (changes requested):
Just the one request to filter out other mana reductions otherwise the code lgtm, Would really like to see:

- Average time spent channeling
- Average stacks consumed per cast
- Stack gain from crit 

Won't hold the PR up over these though they can be added later

trevorm4:
> Just the one request to filter out other mana reductions otherwise the code lgtm, Would really like to see:
> 
> * Average time spent channeling
> * Average stacks consumed per cast
> * Stack gain from crit
> 
> Won't hold the PR up over these though they can be added later

Average stacks is in the tooltip already but I can add it to the main stat if you'd like.  I'll add the crit stack gaining and lifecycles in the next pr since they are both related to stacks.     

I think once I implement lifecycles I will add a generic chart for mana tea stack source to try to avoid making this statistic overly verbose.

src/analysis/retail/monk/mistweaver/modules/spells/ManaTea.tsx
abelito75: This might be a good for a normalizer as then it would actually show up in other modules like Cooldowns

abelito75:
I want to say I heard there is an upper limit to mana tea stacks you could have

Might want to add that if it is true

vohrr:
> I want to say I heard there is an upper limit to mana tea stacks you could have
> Might want to add that if it is true

Good point, 20 stacks is the max and would be good to show wasted stacks. Though I don't think it is required to push this through

src/analysis/retail/monk/mistweaver/modules/spells/ManaTea.tsx
trevorm4: I agree that it should be a normalizer, I think that this could be applied to every healing spec and you can create a generic extendable ManaCostNormalizer. I am going to limit the scope of this PR and implement a generic version in another PR.

vohrr (approved):

trevorm4:
Added stack wasting and average channel time to the tooltip

PR #6356, opened 2023-09-08
[Devastation] More in-depth Disintegrate analysis

Krealle:
### Description

Added a graph for deepdiving into Disintegrate casts over a fight.
This gives more in-depth feedback on overall cast effeciency.

This should help make it more accessible for people to figure out how they should be properly playing around clipping/chaining, and at what point in their rotation they are making mistakes.
This also makes it easier for people to compare what other players are doing, since it can be akward to figure out easily using WCL. (Commonly asked q in Evoker discord)

@ljosberinn 

### Testing

- Test report URL: `/report/6RgwY1MV3CcJv792/25-Mythic+Scalecommander+Sarkareth+-+Kill+(7:06)/Nimeji/standard/overview`
- Screenshot(s):
image
image
image


ljosberinn:
LGTM from my end, this will be great but I'd also love to see this available for dungeons; could maybe filter dungeonPulls for boss encounter id and isKill and render a graph for each boss?

Krealle:
The problem with dungeon support rn is that the graph completely breaks when it gets that long, haven't really looked into why since I felt it was kinda w/e for dungeons. But could take another look at it.
Making graphs on a boss by boss basis might be the play yea.

Krealle:
@ljosberinn dungeon support now included 😄 

https://github.com/WoWAnalyzer/WoWAnalyzer/assets/3404958/28ec25f6-bf5c-4610-9fbb-4b06ac116285

ljosberinn:
actually pogging!

emallson (approved):

PR #6358, opened 2023-09-09
[MW] Add energizing brew module

trevorm4:
### Description

Adds relatively simple module to show mana/time saved on mana tea casts

### Testing

- Test report URL: `report/ghkWqFVA8tBP92c1/50-Mythic+Magmorax+-+Kill+(4:54)/Livingmist/standard/statistics`
- Screenshot(s): 
Screenshot 2023-09-08 at 10 39 11 PM
Screenshot 2023-09-08 at 10 38 24 PM

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/EnergizingBrew.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/EnergizingBrew.tsx
@@ -45,7 +45,8 @@ class EnergizingBrew extends Analyzer {
     }
     // channel duration is equal to time saved because its 50% faster
     const duration = getManaTeaChannelDuration(event);
-    const gcdTime = this.globalCooldown.getGlobalCooldownDuration(TALENTS_MONK.MANA_TEA_TALENT.id);
+    const gcdTime = this.globalCooldown.getGlobalCooldownDuration(SPELLS.VIVIFY.id); // dummy full gcd spell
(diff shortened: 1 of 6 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/EnergizingBrew.tsx
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }
+
+  onManaRestore(event: ResourceChangeEvent) {
+    this.totalRestored += calculateEffectiveResourceRestored(event, MANA_INCREASE);
+  }
+
+  onApplyBuff(event: ApplyBuffEvent) {
+    // channel duration is equal to time saved because its 50% faster
+    this.timeSaved += getManaTeaChannelDuration(event) || 0;
```
vohrr: Could probably do the same thing here that was just done in ManaTea for prepull channels. If we're going to record a 0 duration channel to time saved we shouldn't add 1 to total teas it should probably just return

vohrr: I also don't know how I feel about this for channels <= the gcd. Since you're locked into the gcd regardless you're not really 'saving time' and more so consuming additional stacks. But I guess that's semantics

trevorm4:
Added gcd and prepull logic

vohrr (approved):

PR #6359, opened 2023-09-09
[Mistweaver] add mana tea stack breakdown module

trevorm4:
### Description

Deletes lifecycles module and adds mana tea stacks breadown module which is only applicable when lifecycles is talented

### Testing

- Test report URL: `report/wanWm4qVMhtydfRg/44-Mythic+Scalecommander+Sarkareth+-+Kill+(7:42)/Trevormonk/standard/statistics`
- Screenshot(s):
image
image


src/analysis/retail/monk/mistweaver/constants.ts
```diff
@@ -34,7 +34,6 @@ export const TEACHINGS_OF_THE_MONASTERY_DURATION = 20000;
 // Talent Constants
 export const CF_BUFF_PER_STACK = 0.2;
 export const MAX_CHIJI_STACKS = 3;
-export const LIFECYCLES_MANA_REDUCTION_PERCENT = 0.25;
 export const MANA_TEA_MAX_STACKS = 20;
```
vohrr, line 37: Delete the one on line 48 as well

src/analysis/retail/monk/mistweaver/modules/spells/ManaTeaSources.tsx
```diff
@@ -0,0 +1,140 @@
...
+      (this.naturalStacks.usedStacks + this.naturalStacks.wastedStacks)
+    );
+  }
+
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE(10)}
+        size="flexible"
+        category={STATISTIC_CATEGORY.TALENTS}
```
vohrr: Move to core or theorycraft as mentioned  in the other comment depending on which route we go

src/analysis/retail/monk/mistweaver/CombatLogParser.ts
```diff
@@ -52,7 +52,6 @@ import AverageTimeBetweenRSKSs from './modules/spells/AverageTimeBetweenRSKs';
 import ChiBurst from './modules/spells/ChiBurst';
 import InvokeChiJi from './modules/spells/InvokeChiJi';
 import JadeSerpentStatue from './modules/spells/JadeSerpentStatue';
-//import Lifecycles from './modules/spells/Lifecycles';
 import ManaTea from './modules/spells/ManaTea';
```
vohrr, line 55: 2 ways to go about this:

1.  We have more info that we should detail out when talenting lifecycles - We can either keep the lifecycles stub out and rework it shortly here or
2. We can add the additional LC details to this module. 

I personally like the idea of leaving this as is and keeping a new LC module. This one works well as a supplemental module for both talents  in a non-talent category like Core or Threorycraft.  
LMK what you think

trevorm4: I think new Lifecycles module is probably cleanest since this is already 2 pie charts.

PR #6365, opened 2023-09-15
Remove duplicate "Primary Stats"

oBusk:
### Description

While working on another PR I wanted to "correctly" implement a "Primary Stat" buff. I realised that the spec has a `primaryStat`, but it's not really used anywhere. I did a small refactoring so that `primaryStat` returns the same "STAT" that is used in other places. This way it can be passed to `getName()`, `getIcon()` etc.

I also did a change to StatTracker so that you can implement "dynamic" buffs without items. This can be useful for some buffs who depend on something with the Player.

I _wanted_ to change the "StatBuff" so that you create a _single_ method that returns all stats, insead of the current situation where each stat has it's own method, so in cases where some buffs gives either one or another stat, or multiple stats, you have to keep rechecking the same things. Like if a buff gives both strength and haste depending on the itemlevel, you will have to do the same thing for both strength and haste. 
However, I'm not sure if would be annoying to have to pass in `{ stat: { strength: 0 } }` for static values. So coudln't see a natural API for it.

### Testing

Test that stats seem to work as expected.


src/parser/core/Combatant.ts
```diff
@@ -46,7 +46,7 @@ class Combatant extends Entity {
     return this._combatantInfo.player;
   }
 
-  get spec(): Spec | undefined {
```
emallson: this change is not correct. there are combatants with missing spec info, mostly cases with missing `combatantinfo` events. it happens much more often on Classic than Retail

emallson (changes requested):
this PR looks good overall, only qualm is that `spec` may actually be undefined due to data issues. if this breaks the destructuring used for the `primaryStat` helper, then we can add a helper type to handle that.

src/parser/shared/modules/features/STAT.tsx
```diff
@@ -13,26 +13,41 @@ import StaminaIcon from 'interface/icons/Stamina';
...
+  HASTE_HPCT = 'hastehpct',
+  HASTE_HPM = 'hastehpm',
   VERSATILITY_DR = 'versatilitydr',
   LEECH = 'leech',
   AVOIDANCE = 'avoidance',
   SPEED = 'speed',
   UNKNOWN = 'unknown',
 }
 
+type STAT = PRIMARY_STAT | SECONDARY_STAT | OTHER_STAT;
```
ToppleTheNun, line 42: I'd recommend declaring this type as `Stat` and then removing the `eslint-disable-next-line` below it.

oBusk: I figured it was to remain as a enum functionally, but maybe it's just better to be clear between type and const

oBusk:
> this PR looks good overall, only qualm is that `spec` may actually be undefined due to data issues. if this breaks the destructuring used for the `primaryStat` helper, then we can add a helper type to handle that.

What is the expected behavior when spec is null though? Can we just throw an error?

src/parser/shared/modules/features/STAT.tsx
ToppleTheNun: Supposing that you do the below, you'll get an error about `STAT` being a type but is being used as a value:
```ts
const something = STAT.HEALTH;
```

Unfortunately, can't union enums and have the union work as a directly addressable enum.

emallson:
> What is the expected behavior when spec is null though? Can we just throw an error?

you may be looking at player A who *does* have a spec, while player B *does not* have a spec. We don't want to disable every player because one player is missing data (this case *does* happen sometimes in retail, especially with sales)

you can only view a player that has a spec, so anything where you know where the `Combatant` is the selected player, then you can assume the spec is present and let the other branch crash & burn.

src/parser/shared/modules/features/STAT.tsx
oBusk: You can't union enums, but on the rows below this I declare a const as well. It's possible to have a variable which will be used when used as a value, and a type which will be used when used as a type. So pretty much like an enum. 

```
// type used as type and object used as value
const myStat: STAT = STAT.STRENGTH;
```

But I can understand that it's clearer to have two different names for them.

oBusk:
But could selectedcombatant not be of a type where spec is not optional then? So `Combatant` has `spec?` but `SelectedCombatant` has it as defined?

emallson:
yea that was the "helper type" i referenced above. `selectedCombatant` always has a spec (outside of cases where you've modified the URL)

src/parser/shared/modules/features/STAT.tsx
Pewtro: There's a reason the no-redeclare rule exists, imo it's bad design if one cannot discern which is which easily, so I'd vote for different name. Personally, I'm not a big fan of going for a `Stat` though as then it's just same word but different casing changes the meaning.
I'd probably personally defer to something like `STAT_TYPES`, but it might even be that the const should be the with the different name rather than the type, might need to see how Stat is used in this module as is

oBusk:
> yea that was the "helper type" i referenced above. `selectedCombatant` always has a spec (outside of cases where you've modified the URL)

Ah sorry! Ok but then I'll make a helper type and throw an error if it is somehow undefined (like if you have modified url)

ToppleTheNun:
We could probably make a `SelectedCombatant` type that we assign to `selectedCombatant`, as `selectedCombatant` does still require you to use `.spec?` everywhere.

ToppleTheNun:
Though I will also note that doing the above has risks too, since people can manually construct the URL to access the report for their character even if we disable it in the UI.

oBusk:
> Though I will also note that doing the above has risks too, since people can manually construct the URL to access the report for their character even if we disable it in the UI.

But as I mentioned, I figure we can just throw error if the selectedcombatant does not have a spec?

Would a user even gain anything currently if trying to view a scuffed log where their character does not even have a spec?

src/parser/shared/modules/features/STAT.tsx
emallson: i totally missed this when reviewing. I agree that we don't want to redeclare the type as a const.

it may not be the prettiest, but since `enum`s are objects at runtime, perhaps just...duplicate the values? these things don't change very often.

that is: go back to `enum STAT` and add an `enum PRIMARY_STAT` that includes the 3 primary stats. the `SECONDARY_STAT` and `OTHER_STAT` values are new and not currently used.

emallson:
They won't even get a `CombatLogParser` class properly constructed in that case, so it'll error far before it reaches any real analysis code

src/parser/shared/modules/features/STAT.tsx
oBusk: > i totally missed this when reviewing. I agree that we don't want to redeclare the type as a const.
> 
> it may not be the prettiest, but since `enum`s are objects at runtime, perhaps just...duplicate the values? these things don't change very often.
> 
> that is: go back to `enum STAT` and add an `enum PRIMARY_STAT` that includes the 3 primary stats. the `SECONDARY_STAT` and `OTHER_STAT` values are new and not currently used.

That's fair 😅

oBusk: There we go, kept the original `enum`, then create two new enums for subset of main `STAT` enum. 

There's no overlapping type/value, still works.

oBusk: Actually I take it back.. the solution with multiple enums isn't great either because the types `STAT !== PRIMARY_STAT`... So if you have a function that expect `STAT` you cannot pass `PRIMARY_STAT.STRENGTH` eventhough the string is `strength`.

The `type STAT; const STAT` let's us build an object with enum values and a type which is a union of those enums. That way `PRIMARY_STAT.STRENGTH` satisfies `PRIMARY_STAT` _and_ `STAT`. `PRIMARY_STAT.STRENGTH === STAT.STRENGTH` will also be true.

Having a value and type with the same name is [perfectly legal](https://www.typescriptlang.org/docs/handbook/declaration-files/deep-dive.html#simple-combinations-one-name-multiple-meanings), an `enum` is actually both a value and type to begin with. Also looking at the eslint-rule in particular, it even has a [recommendation](https://typescript-eslint.io/rules/no-redeclare/) to not be enabled since Typescript itself will catch redeclarations.

If we still don't like this approach, I can only see two ways. 

1. Separate the value and type by name, as suggested originally. Either separate sub-values `enum STAT; const PrimaryStat; type PRIMARY_STAT;` _or_ separate main Stat `const STAT; type Stat; enum PRIMARY_STAT`. 
2. Have a `enum STAT` and then `const PRIMARY_STAT = { STRENGTH: STAT.STRENGTH }` but _no_ type for the primary/secondary subsets.

oBusk:
@emallson I did a Helper type and enforced that "Selected Combat" is always that type, and we can access `spec` without having to check null. 

I did the assertion all the way in `Combatant.ts` so we could assert it once and not have to recheck in `get`-ers

I'm sure there's more properties we could assert in this helper type.

Have a look at the commit please! 64c387422ad92ae8bb44ed81a05b91146260878e

src/parser/core/Combatant.ts
oBusk: @emallson I did a Helper type and enforced that "Selected Combat" is always that type, and we can access `spec` without having to check null. 

I did the assertion all the way in `Combatant.ts` so we could assert it once and not have to recheck in `get`-ers

I'm sure there's more properties we could assert in this helper type.

Have a look at the commit please! 64c387422ad92ae8bb44ed81a05b91146260878e

oBusk: Ooof this [breaks](https://github.com/WoWAnalyzer/WoWAnalyzer/actions/runs/6255497848/job/16984935621?pr=6365#step:5:1467) all the integration tests because eventhough [`TestCombatLogParser` overrides selectedCombatant](https://github.com/oBusk/WoWAnalyzer/blob/31fb77a36c62cdd0713818fd8b7d53019e05cdb7/src/parser/core/tests/TestCombatLogParser.ts#L33), the `Combatants` module runs in whole, meaning it will fail... Not sure what would be the correct approach.. Any thoughts?

emallson: Looks like the [Combatants](https://github.com/oBusk/WoWAnalyzer/blob/primary-stat/src/parser/shared/modules/Combatants.ts#L57) module does not use `selectedCombatant` to get the selected combatant

oBusk: No exactly, normally it's vice versa `selectedCombatant()` uses `Combatants.selected`. In the dummy combatlogparser we try to ignore what the `Combatants` module is doing by just overriding `selectedCombatant`, but `Combatants` will still run.

emallson: yea in this case it looks like it the `this.players` array is not initialized in a way that functions with the `TestCombatLogParser`.

for the sake of making it function with current tests, you could make it just skip the check if the combatant isn't defined (`this.players[...] === undefined`). Not ideal but retrofitting a mock combatant onto every test would also be unpleasant.

oBusk: I guess more "correct" would either be to mock the combatantinfo or mock the combatant module itself? But I'm not deep enough into the module system to know how to do that. 

It seems to work with this little check if the combatant is defined, do you feel this is correct? Or should we avoid messing with Combatant module and just doing some type assertion in CombatLogparser instead.

src/parser/shared/modules/Combatants.ts
```diff
@@ -54,7 +54,11 @@ class Combatants extends Entities<Combatant> {
         this.players[combatantInfo.sourceID] = new Combatant(this.owner, combatantInfo);
       }
     });
-    this._selected = this.players[this.owner.playerId];
+    const selected = this.players[this.owner.playerId];
+    if (selected != null && !validSelectedCombatant(selected)) {
```
emallson:
```suggestion
    if (selected && !validSelectedCombatant(selected)) {
```

can be simplified to this to avoid the double-equals check (immediately drew my eye because those are *usually* a bad thing but it does actually do useful things here by passing on null *and* undefined)

src/parser/core/Combatant.ts
emallson: Since this case comes up for the tests only, I think its ok if we cheat a little. Our test infra as a whole needs love following a bunch of the work from this expansion but hasn't gotten it yet

src/parser/shared/modules/Combatants.ts
oBusk: I prefer to explicitly check for `undefined` and `null` rather than just check for truthy. `a == null` is true for `undefined` and `null` and it's clear to the reader that we're checking that it's defined rather than checking for a truthy value. But saying that, this is no hill I plan to expire upon

emallson:
I'm gonna debug the e2e tests locally. They're failing even after retrying which makes me think something actually broke but the error message isn't helpful.

oBusk:
I'm gonna take another stab at a different method syntax for the stattracker. I'm running into another scenario where it would be nice to be able to calculate values once instead of repeating operations for each stat.

oBusk:
> I'm gonna debug the e2e tests locally. They're failing even after retrying which makes me think something actually broke but the error message isn't helpful.

I'm gonna debug it locally! Managed to set it up and get going. Started doing #6376 to make e2e tests a little more helpful

src/parser/core/Combatant.ts
oBusk: The e2e tests that was failing was because of the changes to the Combatants module, so I figured the `SelectedCombatant` change was more hassle than it was worth and rolled it back, and added a helper getter which assures that `primaryStat` is available (or die trying)

oBusk:
Reworked and e2e tests going through lets gooooooooo

src/parser/shared/modules/StatTracker.ts
```diff
@@ -956,8 +974,10 @@ class StatTracker extends Analyzer {
             ' ...unable to handle stats buff, making no stat change.',
           );
         }
+        return 0;
       }
-      return 0;
+
+      return buffVal(selectedCombatant, null as any);
```
emallson: `null as any` is a spicy one. this also makes the typing of the function unsound. `BuffVal` is a number or `(Combatant, Item) => number` and as a result can assume that the 2nd param is non-null since TS values are not inhabited by `null` by default.

this change should either be reverted or the type of `BuffVal` should be changed.

src/parser/shared/modules/features/STAT.tsx
emallson: given the documentation and the other quirks of TS in this area, I'm okay with using the double-namespace to have both exist with the same name.

emallson (changes requested):

src/parser/shared/modules/StatTracker.ts
emallson: fixed by allowing `null` for the `item` param and refining the behavior of the relevant functions

emallson (approved):

PR #6376, opened 2023-09-23
Fail e2e tests on unexpected errors

oBusk:
### Description

As is becoming clear in #6365, the e2e tests can fail but doesn't really give any helpful error message. The problem is that the Error boundaries catches the errors, and displays the "Something went wrong" page for the enduser. 

The playwright tests just stares at this screen until the test times out. This obviously means CI is running for a lot longer than necessary, and also that the error is just "timed out" rather than the actual error.

This PR aims to get better e2e tests by looking for console errors during the test, checking if it's an error we "expect", and if it is not, fail the test. I tried a different approach where we skip the error boundaries so that only uncaught errors would fail the tests, but react router seems to bounder all errors.

I've opted to look at _all_ error messages (not only uncaught ones) with a strict list of ignored errors. This means that we have to be very explicit in ignoring an error message.

To make the tests go through I'm going through and fixing _a lot_ of react rendering errors.


src/interface/guide/components/CooldownExpandable.tsx
```diff
@@ -36,8 +36,8 @@ const CooldownExpandable = ({ header, checklistItems, detailItems, perf }: Props
         {checklistItems && checklistItems.length !== 0 && (
           <section>
             <header style={{ fontWeight: 'bold' }}>Checklist</header>
-            <tbody>
-              <table>
+            <table>
+              <tbody>
```
emallson, line 40: this is a fun one

src/analysis/retail/monk/mistweaver/modules/spells/CloudedFocus.tsx
```diff
@@ -264,7 +264,7 @@ class CloudedFocus extends Analyzer {
         {keys.map((key) => {
           const map = this.stackMap.get(key);
           return map ? (
-            <>
+            <div key={key}>
```
emallson, line 267: note that it is possible to add a key to fragments using `import { Fragment } from 'react';` and then:

```tsx
<Fragment key={key}>
```

i don't think it needs to change for most of these, but noting for the future.

src/analysis/retail/evoker/devastation/modules/guide/CoreRotation.tsx
```diff
@@ -25,7 +25,7 @@ export function CoreRotation({ modules, info }: GuideProps<typeof CombatLogParse
       </p>
       <AplSectionData checker={AplCheck.check} apl={AplCheck.apl()} />
       <hr />
-      <p>
+      <div>
```
emallson, line 28: also note: `<p>` and `<div>` do not have the same styles. I checked several specs to confirm it didn't make them look (super) bad but this is generally not a 1-1 swap (and the browser *may* allow invalid nesting in some cases)

src/analysis/retail/shaman/elemental/modules/features/SpenderWindow.tsx
```diff
@@ -158,18 +158,20 @@ class SpenderWindow extends Analyzer {
...
     performanceThresholds: LTEThreshold | GTEThreshold,
     windowTimestampCallable: (windows: FinishedSpenderWindow[]) => JSX.Element[] = (windows) =>
-      windows.map((w) => <>{formatDuration(w.timestamp - this.owner.fight.start_time)}, </>),
+      windows.map((w, i) => (
+        <span key={i}>{formatDuration(w.timestamp - this.owner.fight.start_time)}, </span>
+      )),
   ) {
     return (
-      <p style={{ padding: 0, margin: 0 }}>
+      <>
```
emallson: this is a breaking change. ill push a fix for it since im staring at it rn.

emallson (approved):
approved with the fix to ele sham

src/analysis/retail/monk/mistweaver/modules/spells/CloudedFocus.tsx
oBusk: Did not know that!

src/analysis/retail/evoker/devastation/modules/guide/CoreRotation.tsx
oBusk: Yes. But understood it needed to be a block element. Inspecting the guides looks like there was no default `<p>` `margin-top`/`margin-bottom` to account for.

It's unfortunate that all guides are written with `<p>` tags as default when people who are not seasoned frontend/react developers know about the limitiations/rules surrounding `<p>` tags. Guides could probable exposed a React component to wrap all blocks in with some padding and whatever, and not use any default tags tbh.

emallson: You can blame me for a lot of the use of `<p>`. I used it a lot in some of the early stuff that got used as a reference

PR #6380, opened 2023-09-24
replace most usages of maybeGetSpell

ToppleTheNun:
### Description

Replace most usages of `maybeGetSpell` with `maybeGetTalentOrSpell`.


emallson (approved):
not sure why the CI failure, but this looks fine

ToppleTheNun:
I'll figure out why it's failing when I'm back from vacation

abelito75:
Figure it out now ╰（‵□′）╯

ToppleTheNun:
Issue was that ultimately, we want spells that aren't in the `SPELLS` object to be put in there, even if they're talents. We just want the talent version to be the name/icon that's used instead of whatever's logged first.

My original version of updating `addSpellInfo` just made talents not go in the spellbook, which broke Fire Mage, Brewmaster Monk, and Vengeance Demon Hunter analysis:
```ts
  addSpellInfo(ability: Omit<Ability, 'type'>) {
    if (maybeGetTalentOrSpell(ability.guid) || !ability.name || !ability.abilityIcon) {
      return;
    }

    registerSpell(ability.guid, ability.name, ability.abilityIcon.replace(/\.jpg$/, ''));
  }
```

The version that's ultimately in this PR is "more correct" in that if the spell is already in the spellbook, we exit early; if a talent exists by the same spell ID, we use the name from the talent and otherwise just use what's logged:
```ts
  addSpellInfo(ability: Omit<Ability, 'type'>) {
    // If the spell is already in the spellbook, we ignore it
    if (maybeGetSpell(ability.guid) || !ability.name || !ability.abilityIcon) {
      return;
    }

    // If the spell is a talent, we want to use the talent version instead of whatever
    // is logged by the game, because it can be misleading
    const talent = maybeGetTalent(ability.guid);
    registerSpell(
      ability.guid,
      talent?.name ?? ability.name,
      (talent?.icon ?? ability.abilityIcon).replace(/\.jpg$/, ''),
    );
  }
```

src/interface/PremiumLoginPanel.tsx
```diff
@@ -98,7 +98,7 @@ const PremiumLoginPanel = () => {
   return (
     <div className="panel">
       <div className="panel-body" style={{ padding: '0 15px', position: 'relative' }}>
-        <TransitionGroup classNames="logged-in" timeout={1000}>
+        <TransitionGroup className="logged-in" timeout={1000}>
```
emallson: pretty sure this is not a correct change: https://github.com/WoWAnalyzer/WoWAnalyzer/commit/ba5c2dd07204082ffbe74d491a0055b42a67aa77

relevant PR: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6381

emallson (commented):
everything except the `TransitionGroup` looks ok to me. consider this a 👍 after that

src/interface/PremiumLoginPanel.tsx
ToppleTheNun: I was getting this error in Chrome, which lead to the change:
image

PR #6384, opened 2023-09-26
Add tracking and analyzing of all Dragonflight weapon enchants

oBusk:
> This PR replicates code from #6365 and #6378 so remains a draft until they are resolved.

While hunting haste to get cooldowns to match up, I realised that the specified value for Wafting Devotion was incorrect. For one, it did not look at the quality of the enchant, but more importantly, it did not account for players potentially dualwielding weapons with the same enchant, which actually causes the stat values to stack. So I set out to implement this, and quickly decided that having code for checking dualwield and ranks as part of the `StatTracker` would become unwieldy and annoying to maintain, so I decided to implement the stat tracking as a separate module calling `StatTracker.add()`. 
Once I had a module for the tracking of the stat, I felt like I could show some statistic for the procs and the value of it.
Once that was done for Wafting Devotion and Wafting Writ, being the obvious haste buffs, the completionist in me wanted to implement similiar tracking and analyzing of all weapon enchants, and so I did. 

I wrote and used https://wcl.nulldozzer.io/ to find logs for every enchant and rank, so each module has a comment with those logs. Below I will link 🔗 to a single log for each enchant.

# Stat Proc Enchants

For each enchant that procs a stat buff we inform the StatTracker of the values of the stat, based on the quality and potential mh/oh. We also track the number of procs, uptime and the calculated stat value of the proc throughout the fight.

image image

- Writs
  - Burning Writ (Critical Strike) [🔗](http://localhost:3000/report/4DZNxRHMbjdnW6QC/21-LFR+Kazzara,+the+Hellforged+-+Kill+(3:17)/Favelaninja/standard/statistics)
  - Earthen Writ (Mastery) [🔗](http://localhost:3000/report/2gBbXWMnP3wkKy87/2-LFR+Kazzara,+the+Hellforged+-+Kill+(2:54)/Yvrndrsv/standard/statistics)
  - Frozen Writ (Versatility) [🔗](http://localhost:3000/report/2gBbXWMnP3wkKy87/2-LFR+Kazzara,+the+Hellforged+-+Kill+(2:54)/Yvrndrsv/standard/statistics)
  - Sophic Writ (Strenght) [🔗](http://localhost:3000/report/Z7P4bRcY2DJMHzgC/25-LFR+Kazzara,+the+Hellforged+-+Kill+(3:49)/Mithandour/standard/statistics)
  - Wafting Writ (Haste) [🔗](http://localhost:3000/report/rxR13mGzY4WkH2QN/7-LFR+Kazzara,+the+Hellforged+-+Kill+(4:01)/V%C3%A2ny%C3%A4/standard/statistics)
- Earthen Devotion (Armor) [🔗](http://localhost:3000/report/ACXnrk1TFN6ZLWx9/1-LFR+Kazzara,+the+Hellforged+-+Kill+(3:27)/Oghenim/standard/statistics)
- Sophic Devotion (Strength) [🔗](http://localhost:3000/report/cvgJNhdDMjna1prx/5-LFR+Kazzara,+the+Hellforged+-+Kill+(3:24)/Odinslaw/standard/statistics)
- Wafting Devotion (Haste) [🔗](http://localhost:3000/report/HBZCMDav3dTyXmgn/3-LFR+Kazzara,+the+Hellforged+-+Kill+(2:53)/Alanwatt/standard/statistics)

# Burning Devotion [🔗](http://localhost:3000/report/YjCz1Gw2JqmWc3Ng/4-LFR+Kazzara,+the+Hellforged+-+Kill+(3:01)/Gwenfair/standard/statistics)

We track the procs, and outgoing healing.

image image

# Frozen Devotion [🔗](http://localhost:3000/report/FJQ28G6LgynCK9tj/3-LFR+Assault+of+the+Zaqali+-+Kill+(3:55)/Aequit%C3%A2s/standard/statistics)

We track the procs, outgoing damage, as well as statistic over how many targets we hit with each proc, since the damage scales after number of targets.

image 
image

# Shadowflame Wreathe [🔗](http://localhost:3000/report/fJrQVzPpACd6FnTM/1-LFR+Kazzara,+the+Hellforged+-+Kill+(3:42)/Kasp%C3%ACan/standard/statistics)

We track the number of procs, outgoing damage, as well as how much damage to self.

image image

# Spore Tender

Spore tender is an interesting Enchant in terms of design and tracking. I have created two modules to handle the effects of this enchant.

## Spore Tender [🔗](https://www.warcraftlogs.com/reports/bGafrjQ8Mp6VcLDt#fight=3&type=summary&source=1)

So we track how many times it procs, as well as statistics as to which targets recieved the buff. We don't worry about which particular stat that the targets receive as we dan't know the targets ratings at the time of applications.

image
image
image

## Invigorating Spore Cloud [🔗](http://localhost:3000/report/CkDv213xNw8Lj4Xa/11-LFR+Kazzara,+the+Hellforged+-+Kill+(3:13)/Sayyadenia/standard/statistics)

A separate module to track buffs you recieve from Spore Tender enchant. This module also checks which secondary stat to give at the time of application, and correctly removes the same stat when it expires.

image image image


src/parser/retail/modules/items/dragonflight/enchants/StatProcEnchantAnalyzer.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  constructor(
+    stat: STAT,
+    enchantSpell: Spell,
+    buff: Spell,
+    ranks: StatProcEnchantRank[],
+    options: Options,
+  ) {
+    super(enchantSpell, ranks, options);
+    this.statTracker = (options as Options & { statTracker: StatTracker }).statTracker;
```
oBusk: If someone can tell me what's wrong when the DI system does not set `this.statTracker` but I can access it in `options`. Feels like this happens most of the time, but sometimes it works. I've tried to look in the debugger, but I cannot really follow

emallson: its a quirk of the JS class system. superclasses cannot asssign properties to fields that are not in the superclass (even if they're in the subclass), which means that the `super` call cannot initialize most of the dependency fields. In general, this means that you can't use the DI modules via `this` until after the constructor call, at which point the `CombatLogParser` class actually assigns the values.

there's a new(ish) helper for this if you create a class by extending `Analyzer.withDependencies(deps)`, which will make typed access to the dependencies available on `this.deps`

emallson: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/dragonflight/src/analysis/retail/druid/restoration/modules/spells/GroveGuardians.tsx#L33 can see `withDependencies` used here, and notably `this.deps` works within the constructor (as long as you've called `super(options)`)

oBusk: That's very nifty! It doesn't seem to workout if I want to extend the classes I've created. I fiddled around with a mixin that is not a static method on the `Analyzer` class, and thus can add dependencies even if it's a second level descendant of `Analyzer`.

https://github.com/oBusk/WoWAnalyzer/commit/995c0b5ca89f7799713fd02646d582ebfc88a258

Do you think this could be something? It seems to work even if the middle generation adds their own `deps`.

emallson: I am fine with `Analyzer.withDependencies` being a shorthand for this. seems reasonable.

ex:

```ts
class Analyzer extends Module {
  // ...
  static withDependencies = withDependencies.bind(null, Analyzer);
  // ...
}
```

oBusk: I've adopted this, please have a look @emallson

src/parser/retail/modules/items/dragonflight/enchants/StatProcEnchantAnalyzer.tsx
```diff
@@ -0,0 +1,102 @@
...
+  amount: number;
+}
+
+const deps = { statTracker: StatTracker };
+
+/**
+ * Abstraction to reuse the same code for tracking stats from proccing enchants such as
+ * writs and Sophic Devotion.
+ */
+abstract class StatProccEnchantAnalyzer extends withDependencies(
```
emallson:
```suggestion
abstract class StatProcEnchantAnalyzer extends withDependencies(
```

small typo

src/parser/retail/modules/items/dragonflight/enchants/InvigoratingSporeCloud.tsx
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  private statisticTooltip() {
+    return (
+      <>
+        Friendly healers with the <SpellLink spell={SPELLS.SPORE_TENDER_ENCHANT} /> enchant has a
+        chance to grant you a buff that increases your highest secondary stat. This is is a
+        breakdown of benefit you have gained from this buff.
+      </>
+    );
```
emallson: i am not sure that this should be a tooltip---its pretty important for contextualizing the number shown.

emallson: i guess maybe it is okay because there is a separate stat for Spore Tender itself, but a title for the stat like `Personal <SpellLink>Invigorating Spore Cloud</SpellLink> Gain`

src/parser/retail/modules/items/dragonflight/enchants/FrozenDevotion.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <>
+        {this.procCount(numberProcs)}, and hit a total of {totalHits} targets (
+        {averageTargets.toFixed(1)} per proc) for an average of {formatNumber(averageDamage)}{' '}
+        damage, resulting in a total damage of {formatNumber(totalDamage)}.
+      </>
+    );
+
+    const content = (
+      <>
+        <DamageIcon /> {formatNumber(dps)} DPS <small>{formatPercentage(percentage)}%</small>
```
emallson: i think this can be replaced with `<ItemDamageDone />`

src/parser/retail/modules/items/dragonflight/enchants/ShadowflameWreathe.tsx
```diff
@@ -0,0 +1,97 @@
...
+          uptime, doing a total of <strong>{formatNumber(this.outgoingDamage)}</strong> damage.
+          <br />
+          Each time triggered, did an average of{' '}
+          {formatNumber(this.incomingDamage / this.buffApplications)} damage to self.
+        </>
+      ),
+      content: (
+        <>
+          <div className="value">
+            <DamageIcon /> {formatNumber(dps)} DPS <small>{formatPercentage(percentage)}%</small>
```
emallson: i think this can be replaced with `<ItemDamageDone />`

emallson (commented):
several minor things, this overall looks good

apologies for the delay in reviewing it, had a pretty hectic week-ish IRL so i fell behind on PRs that require real, in-depth reading

src/parser/retail/modules/items/dragonflight/enchants/InvigoratingSporeCloud.tsx
avilene: Just two tiny grammar notes

- has -> have 
- breakdown of benefit -> breakdown of the benefit

oBusk: > Just two tiny grammar notes
> 
> * has -> have
> * breakdown of benefit -> breakdown of the benefit

Oh thanks a lot! Fixed!

> i guess maybe it is okay because there is a separate stat for Spore Tender itself, but a title for the stat like `Personal <SpellLink>Invigorating Spore Cloud</SpellLink> Gain`

image

Linebreak becomes a bit of a nightmare.

What do you think of this? 

image

src/parser/retail/modules/items/dragonflight/enchants/FrozenDevotion.tsx
oBusk: I had some recollection of it looking terrible. In my mind it was all the large font size (% too), but I can see now that it is `small` and has been for a long time so I don't know what I'm remembering.. Maybe just the added "of total damage" caused line breaks in some case... Anyway I'll use ItemDamageDone instead!

oBusk:
Thanks a lot for the pointers! 👍 I've amended them and re-merged main branch!

emallson (approved):

PR #6389, opened 2023-10-03
Refactor `GetRelatedEvents` and usage

Seriousnes:
### Description

I have updated the `GetRelatedEvents` helper with a number of changes.

1. The function now has includes a generic type parameter of type `AnyEvent`, and the results are cast back to the generic type parameter. This removes nearly all the typecasting currently used.
2. There is an additional overload where a filter callback function can be provided, which consolidates the `filter` appended to most calls to `GetRelatedEvents` into a single function call.
3. A new `GetRelatedEvent` function (with the same signature as `GetRelatedEvents`) to assist in retrieving a single linked event (typically either the first or only event).

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -23,12 +23,14 @@ import {
   Awildfivreld,
   nullDozzer,
   Bigsxy,
+  Seriousnes,
 } from 'CONTRIBUTORS';
 import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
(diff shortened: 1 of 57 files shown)


emallson (approved):
my initial reaction was that I don't think the `filter` parameter is necessary.

however, it does make the `GetRelatedEvent` implementation clean and *is* used (although this usage pattern might be an antipattern).

in the meantime, this is a clear improvement to the API, even if i think that simply calling `GetRelatedEvents(...).filter(...)` is fine for the multiple-events case.

PR #6391, opened 2023-10-06
[Augmentation] Implement Breath of Eons helper module

Krealle:
### Description
TL;DR: More indepth analysis on Breath of Eons uses, based on what buffed actors are doing :-)
<sub><sub>sorry for wall of text</sub></sub>

### Intro
So Breath of Eons, your big CD as an augmentation, accumulates a %% of your buffed targets damage and replicates it after a short period.
Now we already have rotational analysis on uses, ie. track different casts within the window / visualize buff management etc.

But the main crux of Breath of Eons actually comes in your timing of the cast! Due to it mainly ramping off your buff targets damage, you optimally need to "snipe" them at the height of their burst. There is a golden timing around the ~14s mark when buffing classes such as Demo and Unholy.

Now its all good and well to have the theory in place, but how does it actually hold up in practice? Well we could go to WCL and try to manually figure out every proper timings for our Breaths, but due to how Breat of Eons damage output component functions, this can, however, be a bit tricky.

### Problem points
Breath of Eons works by you flying over your targets and applying a debuff to them, mobs that have this debuff will individually accumulate damage done to them by your buffed players and after the debuff runs out, explode and replicate a %% of the damage.

BUT, if the mob dies before the debuff runs out, you will lose out on *all* the accumulated damage in said mob. 
This is problem 1.

Problem 2 is that if you drop Ebon Might in your Breath of Eons window, you will no longer be accumulating damage, essentially halting your burst.

Problem 3 is that you need to filter out certain damage events to make sure you are getting a proper representation of the actualy damage you will have properly attributed to your Breath of Eons.

### Solution

So to help with streamlining to process of analysis this rather large chunk of your damage I've made a module that will look through each individual Breath of Eons window and figure out the following:

- Deal with filtering damage events
- Display lost damage to dropping Ebon Might
- Display lost damage to mobs dying early
- Display how much damage individual buffed players contributed
- Figure out if there were more optimal timings to use Breath of Eons (Using a buffer to look around the current cast)
- Display the best timing possible along with the potential damage increase.

Above analysis should help people figure out and learn the "proper" timings for when to use Breath of Eons in the future.

A thought I had for the analysis was to also compare against non-buffed players, to figure out if there were better buff targets available, it would be fairly simple to add in, but I already had problems with fetching events just for the 5 players needed for each window.
I'm using `fetchEvents` with the `maxPages` kinda cranked up to collect the events needed for the analysis. I'm not quite sure if this is something that's going to cause issues or if it's fine to keep it high (potentially higher even if we want to implement this feature). 

Where it gets even wilder is if we wanted to look across the entire fight to *really* deepdive into finding the optimal Breath timings but, erm, yea, collecting events for the entire fight is prolly not really something that is feasible, and as such, this can be left as an exercise for the reader x)

### Bonus

Along with the added module I've also done a bit of housekeeping:

1. Made a new guide section specifically for Helper modules.
   - The buff helper module rapidly became both the most popular, and most missed module, so these are moved to the top.
2. Updated blacklist.
3. added `subtractsFromSupportedActor` property to `DamageEvent`
4. Expanded on the graph module I made for `Disintegrate` analysis and added support for explanations, behaves much like other generic guide modules.
Has explanations on the left and a graph on the right. Can also be used to only show a graph without explanations, see `Disintegrate` module for Devastation.
I'm not sure if this should be made into a core component? It's mainly built for my specific usecases, but maybe others could make use of it.

### Testing

- Test report URL: `/report/gfjptA4HWKQ3DrPJ/15-Mythic+Echo+of+Neltharion+-+Kill+(6:12)/Kullingfly/standard/overview`
- Screenshot(s):
image
image
image
image


src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsHelper.tsx
```diff
@@ -0,0 +1,538 @@
...
+    start: number;
+    end: number;
+  }[] = [];
+
+  /** Generate filter based on black list and whitelist
+   * For now we only look at the players who were buffed
+   * during breath */
+  function getFilter(window: BreathOfEonsWindows) {
+    const playerNames = Array.from(window.breathPerformance.buffedPlayers.keys());
+    const nameFilter = playerNames.map((name) => `"${name}"`).join(', ');
```
ljosberinn, line 58: edge case but in case theres 2 players with the same name, you may want to use the actor.id over actor.name but ... yea

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsHelper.tsx
```diff
@@ -0,0 +1,538 @@
...
+      breathStart,
+      breathEnd,
+      damageToDisplay,
+      topWindow,
+      sourceInRange,
+    };
+  }
+
+  function generateGraphDataForWindow(
+    topWindow: any,
```
ljosberinn: can be typed I'd assume?

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsHelper.tsx
```diff
@@ -0,0 +1,538 @@
...
+      'Breath Window',
+      !topWindow ? <>You didn't hit anything.</> : undefined,
+    );
+
+    return newGraphData;
+  }
+
+  function generateExplanationContent(
+    topWindow: any,
+    inRangeSum: any,
```
ljosberinn: same

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsHelper.tsx
```diff
@@ -0,0 +1,538 @@
...
+    const content: JSX.Element = (
+      <div className="flex-container">
+        <div className="table">
+          <div className="flex-row">
+            <div className="flex-cell">
+              <TooltipElement
+                content="Because of the way Blizzard handles damage attribution, the values 
+                    displayed here may have a small margin of error. Additionally, if an enemy 
+                    becomes immune or takes reduced damage when your Breath of Eons explodes, this 
+                    value may also be overestimated, e.g. Neltharion going immune mid-Breath."
```
ljosberinn: as useful as it currently is, id remove the example as soon enough people wont be familiar with the encounter anymore

ljosberinn:
LGTM after those couple amendmends

src/analysis/retail/evoker/augmentation/modules/breahtofeons/BreathOfEonsHelper.tsx
Krealle: swapping to using IDs breaks everything 🙃
https://www.warcraftlogs.com/reports/gfjptA4HWKQ3DrPJ/#fight=15&type=damage-done&pins=2%24Off%24%23244F4B%24expression%24source.id%20in%20(7,%2014,%2017,%2020,%202)%20OR%20source.owner.id%20in%20(7,%2014,%2017,%2020,%202)

emallson:
looks like all the comments have been addressed, im resolving conflicts and merging

PR #6393, opened 2023-10-07
Add ICC Zone resources

emallson:
This adds the boss images. The icons + splashes are taken from WCL. The icons are a bit low-res (the icons on WCL are MUCH smaller) and I may follow up on that, but there aren't actually boss achievs for most bosses. (This may be a waifu2x angle)


emallson:
image

image
image


some previews of the images

PR #6394, opened 2023-10-07
make ICC the default classic zone

emallson:
This builds on #6393 and makes ICC the default zone for character parses. It is not needed for direct analysis of logs.


jazminite (approved):
lgtm - ty for doing this, I was going to do it then saw this :)

PR #6405, opened 2023-10-09
[Shadow] Fix Broken Icons

DoxAshe:
### Description

Shadow has several icons that are not displaying properly.  This updates these talents to the correct icons.
I know that this will be overwritten in the future, but this will at least fix the issue for now until I can figure out how to properly change these icons.
This also adds additional information to the guide view for Mind Games.

### Testing

Is there a way to fix these icons without needing to upload them to Bad Icons?

- Screenshot(s):
VEBad


src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
```diff
@@ -138,7 +138,8 @@ const ShortCooldownsGraph = () => {
         <SpellLink spell={TALENTS.MINDGAMES_TALENT} />
       </strong>{' '}
       is a lower priority spell, but when it is cast, make sure{' '}
-      <SpellLink spell={TALENTS.DEVOURING_PLAGUE_TALENT} /> is on its target.
+      <SpellLink spell={TALENTS.DEVOURING_PLAGUE_TALENT} /> is on its target so it benifits fully
```
ToppleTheNun: I think you want this to be `benefits fully` instead of `benifits fully`.

ToppleTheNun:
> Is there a way to fix these icons without needing to upload them to Bad Icons?

There is not currently a way, unfortunately. It's fairly painless to add to bad icons though.

DoxAshe:
I've added the missing Icons to Bad Icons and reverted the changes I made to Talents.

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
DoxAshe: Thank you. I have fixed this.

ToppleTheNun (approved):

PR #6415, opened 2023-10-14
Paginated loading

emallson:
### Description

Use pagination to load data for M+ in chunks to avoid network time-outs. This allows *most* specs to function even in very long / very spammy dungeons.

The one exception that I've seen so far is Fire Mage, which has performance issues in the analysis stage when there are too many events. It remains disabled at this time and we can disable other specs as needed. I checked the other Mage specs, Aug, Brew, and a handful of other specs to confirm that they load successfully.

Pagination technically doesn't care whether you're looking at a raid or dungeon, it just checks fight length: any fight >15m long switches to pagination automatically and loads data in 5-minute chunks. Technically, this means that you could uber speedclear a key in 14 minutes and hit it unpaginated, but if that happens and causes problems then I guess we'll revisit this.

### Testing

Load any M+ log in local dev. Here's one with fire to illustrate the reason it remains disabled: `/report/YzFX4RdMALwt98nh/2-Mythic++Brackenhide+Hollow+-+Kill+(31:15)`


src/interface/report/hooks/useEvents.ts
```diff
@@ -15,19 +23,54 @@ const useEvents = ({
...
   const [events, setEvents] = useState<AnyEvent[] | null>(null);
+  // we manually paginate here. WCL could turn pagination back on in the WCL API but the amount loaded per page is too low
+  // for M+ to be usable without deeper changes on the WCL side.
+  const pageCount = fightPageCount(fight);
+  const [pagesLoaded, setPagesLoaded] = useState(0);
 
   useEffect(() => {
+    let cancelled = false;
+    const loadPage = async (startTime: number, endTime: number) => {
+      return await fetchEvents(report.code, startTime, endTime, player.id);
```
ToppleTheNun: Since this function is already `async`, don't need an `await` here. Technically don't need to make it `async` either, as it'll be a promise either way.

ToppleTheNun (approved):
Happy with this and excited to see it working!

PR #6421, opened 2023-10-17
Refactor Voice of the Silent Star to give/steal the correct stats

oBusk:
### Description

The implementation of the Voice of the Silent Star module looked at the stats on first load and then just applied those stats throughout the fight, disregarding situations where another buff would change which stat is the highest. The case where I'm noticing this is for Fury warrios where it's common to enchant Wafting Devotion, but gear for mastery, meaning Mastery is the highest, until Wafting Procs, at which point Haste becomes the highest stat and if Voice of the Silent Star would proc at that point, it would instead give haste.

I considered if there was any value in trying to track how the "call of the voice" climbs to the 10th when it procs, but figured it's not really relevant.

Also added a separate module, which primary goal is to ensure that any stolen stats are tracked, but also tracks which players has stole from you and shows a statistic for it.

### Testing

- Test report URL: http://localhost:3000/report/p4bgq8k6QjBV7ztT/15-Mythic+Kazzara,+the+Hellforged+-+Kill+(3:29)/Wimmywonwon/standard/statistics
- Screenshot(s):

  image image


src/parser/retail/modules/items/dragonflight/UsurpedFromBeyond.tsx
```diff
@@ -0,0 +1,273 @@
...
+            );
+          })}
+        </tbody>
+      </table>
+    );
+  }
+
+  statistic() {
+    if (this.thefts.length === 0) {
+      // This module is always active, but if never robbed, should not show any statistic
```
emallson, line 202: we do get all players' combatantinfo and cloaks cannot be swapped in combat so you could do `this.active = this.combatantInfoEvents.some(hasCloak);`

emallson (commented):
one minor comment, otherwise seems good. can merge without if you'd prefer, since its not a big deal

src/parser/retail/modules/items/dragonflight/UsurpedFromBeyond.tsx
oBusk: Aah that's a great idea! I've added a check that _another_ player has cloak and only activate in that case!

emallson (approved):

PR #6428, opened 2023-10-20
Move all `parser/ui` components to typescript

emallson:
just some mostly-mindless internal work that i did while waiting on other stuff


src/parser/ui/DistanceRadarRing.tsx
```diff
@@ -0,0 +1,29 @@
+import { CSSProperties } from '@emotion/serialize';
```
ToppleTheNun: Might want the `CSSProperties` from React instead of emotion

emallson: Oops, yup that's a bad auto import

PR #6429, opened 2023-10-21
[Augmentation] Implement 10.2 support

Krealle:
### Description

Preparing Augmentation for 10.2.

Implemented features:

1. Ebon Might module now accounts for extra Prescience duration with T31 2pc.
2. Stats for T31 4pc.

Along with these I've disabled the Helpers (Specifically with S3 in mind), since hooks are pretty borked rn, along with needing to update blacklist, and verifying how various new abilities/trinkets/etc actually interact with Augmentation buffs :)))))).

So ideally this gets merged before S3 starts, but no need to merge it before 10.2 hits, since the Helpers can still provide some value till then :)

Might also have to start bonking people for pre-pull buffing since those are now getting stripped on pull, but thats tbd.

### Testing

- Test report URL: `/report/9GQbkmT1X4VpPx8H/4-Mythic+Igira+the+Cruel+-+Wipe+4+(6:32)/Andract/standard/overview`
- Screenshot(s):
image
image
image
image


kfinch:
This needs to re-target `dragonflight` branch, and some manner of merging needs to happen because the changelog includes a lot of changes from other specs that have already gone in.

Krealle:
hmm, not quite sure why it's failing the test, looks to be failing on a guardian test?

Pewtro:
dont worry about that test

emallson:
Looks like a simple load error, rerunning the job to check. If it fails again might be worth checking the live site to make sure the log wasn't deleted or something

kfinch (approved):
I see you commented out helpers rather than deleting - I presume that means they can be fixed, but you don't want to deal with it right now?

Krealle:
It can't really be fixed for now, the support hooks are in a really bad spot - and this is without even knowing how bad the new tier set hooks will be.

PR #6445, opened 2023-10-27
windwalker: fix FoF channeling and APL

smvoss:
### Description

FoF was breaking the APL due to both the cast and `beginchannel` being checked, causing the second one to raise a failure. Add property to track the previous cast in the APL as well as previous `beginchannel` so that we can watch for duplicates that happen in either direction.

### Testing

- Test report URL: `report/cLWkGhMRyXKN1BdT/6-Mythic+Magmorax+-+Kill+(5:37)/Durpn/standard`
- Screenshot(s):

Channel now tracked:
image


emallson (approved):
this could *probably* be folded into the same field but i don't feel the need to push for that right now, especially because a major refactor of the apl internals is on my (long) todo list

PR #6446, opened 2023-10-27
Move reducers and actions into slices

avilene:
### Description

Replaced all previous reducers / actions with slices resulting into less hardcoded codes for f.e. actions and tracking everything in one file.

### Testing

Test by doing anything... really.


emallson (approved):
code looked mostly fine (just some `console.log`s) but i pulled it locally and noticed some breakage:

- selecting a fight didn't work
- jumping right to a player within a fight did work, but the navbar didn't have any navigation

the issues were caused by some `state = X` that were missed in moving things around during the refactor. I updated several other places that I saw it as well and now things appear functional. User stuff (premium, user state) is always a PITA to test locally since we don't have accounts, so I'm happy enough to merge and then fix if premium/login is broken.

PR #6451, opened 2023-10-29
update Sentry setup to use @sentry/react

ToppleTheNun:
The sourcemaps upload is kinda janky because we're using create-react-app atm; it's a lot simpler with vite. Might wait to do it until we move to Vite, pending review.

### Testing

Can't really test it locally, need to deploy to a "real" environment and force an error to test it.


emallson (commented):
code for this looks fine. when you say "kinda janky" is it just the extra workflow step with environment stuff that is a little jank? or something flaky with the upload itself?

ToppleTheNun:
It's the extra step that has to integrate after the build instead of during it; it injects extra information into the sourcemaps after the build and then uploads them to Sentry.

It feels a bit weird in comparison to it just occurring during the build process like you'd get with the Sentry webpack and vite plugins, but that's the price we pay for using create-react-app.

ToppleTheNun:
Also need to update the Sentry SDK in this, there's a new version that fixes some issues.

ToppleTheNun:
This should be good to go now.

emallson:
i forgot about this :notlikeduck: i'll get it re-reviewed (and hopefully merged) tomorrow

emallson (approved):

emallson (approved):

PR #6477, opened 2023-11-09
[Shadow] Add Tier 31 Four Piece.

DoxAshe:
### Description

Add statistics for shadow priest's tier 31 four piece.  This set gives a stacking buff every time Shadow Word: Death deals damage, that then can be consumed to increase the initial damage of Shadow Word: Pain, or the damage from Shadow Crash.

Shadow's tier 29 statistic was also updated to not appear if the bonus is not active.

I also noticed a strange issue with Shadow Word: Death's spell usable from the death speaker talent, that I have been unable to fix but will continue to work on for a future update.  With the tier set revolving around Shadow Word: Death, this issue is important to fix.

### Testing

Death Speaker is a buff that increases Shadow Word: Death max charges by one and restores one charge.  However, it restores both charges when it should only restore one.  This happens in both ptr and 10.1 logs.  I don't know if there is something I am missing to cause both charges to be restored instead of one.

- Test report URL: `report/61fWqPBYTDx3M8rX/6-Normal+Gnarlroot+-+Kill+(2:54)/Publik/standard`
- Screenshot(s):

Shadow Priest T31 4set

SWD issue


src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
```diff
@@ -75,7 +82,7 @@ const CoreCooldownsGraph = () => {
       VoidboltAnalyzer?.executeRanges.map(fromExecuteRange);
 
     message = (
-      <Trans id="guide.priest.shadow.sections.corecooldowns.graphVB2">
+      <Trans id="guide.priest.shadow.sections.corecooldowns.graphVB">
```
trevorm4: We're abandoning translations inside specs, feel free to remove this wrapper

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  castSC = false; //this tells if a cast of Shadow Crash has happened
+  recentSCTimestamp = 0; //This is the time of the first instance of shadowcrash damage to a target from one cast.
+  bonusSWP = 2.25; // multiplier per stack for Shadow Word Pain
+  bonusSC = 0.5; // multiplier per stack for Shadow Crash
+
+  recentApplied = false; //the order of the events requires this for edge cases.
+  timestampOfLast = 0; //Used to calculate unused buffs
+  maxDuration = 60000; //max duration of the buff
+  maxStacks = 12; //maximum stacks
```
trevorm4: Since these are constants you should move them to a global variable instead of an instance variable

src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
DoxAshe: I'll start going through and removing them from shadow's guide and statistics.

src/analysis/retail/priest/shadow/modules/talents/DeathAndMadness.tsx
```diff
@@ -51,10 +53,20 @@ class DeathAndMadness extends Analyzer {
...
-        this.spellUsable.endCooldown(TALENTS.SHADOW_WORD_DEATH_TALENT.id);
+        this.spellUsable.endCooldown(
+          TALENTS.SHADOW_WORD_DEATH_TALENT.id,
+          event.timestamp,
+          false,
+          false,
+        );
+        DEBUG &&
+          console.log('Shadow Word: Death Reset', this.owner.formatTimestamp(event.timestamp, 1));
         this.resets += 1;
```
kfinch: You have a misconception on how Death and Madness works. There is no "waiting 10 seconds to miss the reset" - the reset happens immediately upon using SWD when the condition is met (either you have a Deathspeaker proc or the target is in execute range - AND SWD fails to kill the target). The '10 seconds' is how often a Death and Madness proc can happen, where this proc IS the instant reset.

Your activation logic looks mostly correct, but I think you need to track 10 seconds from last reset, not from last cast. It's possible this accounts for some of the funniness.

src/analysis/retail/priest/shadow/modules/tier/Tier29ShadowPriest4Set.tsx
```diff
@@ -22,13 +23,15 @@ class ShadowPriestTierSet extends Analyzer {
...
   protected haste!: Haste;
+  has4Piece: boolean = true;
 
   uptime4Set = 0; //Buff uptime in milliseconds
   is4SetActive = false;
   timestampOfLast4SetActivation = 0; //Used to calculate uptime in case the fight ends with the buff active
 
   constructor(options: Options) {
     super(options);
+    this.has4Piece = this.selectedCombatant.has4PieceByTier(TIERS.T29);
```
kfinch, line 34: It's impossible to have both the T31 4set and T29 4set at the same time - why bother with this?

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
```diff
@@ -0,0 +1,191 @@
...
+  onShadowWordPain(Event: DamageEvent) {
+    //Since the damage event of the inital hit occurs right after the cast, this will only allow that damage event to be added and increased and not further ticks of the dot.
+    //In addition, the recentStacks is not reset until a new stack is gained. This way, if two shadow word pains are cast (or shadow crashes) without gaining the buff since the last cast, the damage will not be counted.
+    if (this.castSWP && this.recentApplied) {
+      this.recentApplied = false;
+      this.stacksSWP += this.recentStacks;
+      this.damageSWP += calculateEffectiveDamage(Event, this.recentStacks * this.bonusSWP);
+      this.castSWP = false;
+    }
+  }
```
kfinch: It's likely you could detect this instead by looking for SWP damage that doesn't have the `tick = true` flag, this indicating it was the direct initial hit. I can't confirm this unfortunately because all the PTR logs I could find were consuming the proc with Shadow Crash instead of a SWP hardcast.

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
```diff
@@ -0,0 +1,191 @@
...
+import SpellUsable from 'parser/shared/modules/SpellUsable';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import { calculateEffectiveDamage } from 'parser/core/EventCalculateLib';
+import ItemDamageDone from 'parser/ui/ItemDamageDone';
+
+const MAX_DURATION = 60000; //max duration of the buff
+
+class Tier31ShadowPriest4Set extends Analyzer {
```
kfinch, line 19: All this logic is a PRIME candidate for using an `EventLinkNormalizer` - link T31 buffremove to Shadow Crash cast and SWP cast. You can then listen only for those casts and detect if they were buffed by checking the link.

You could then also you a link normalizer to link SWP cast to SWP direct damage, and SC cast to SC damage (with a large forward buffer to account for travel time, this should be fine because SC has a cooldown). Follow the link from cast to damage tally right there.

kfinch (commented):
This looks largely correct, but I commented on some possible improvements.

Also I messed around on my priest to see if I could figure out the charge thing - it shows on my character bar that when I get a Deathspeaker proc, I suddenly have 2 charges of SWD, which is very odd because it says nothing about giving more charges. As soon as I cast I immediately lose the 2nd charge.  My **guess** is that the dev programming Deathspeaker wanted it to give the Death and Madness reset, and did it in a weird / lazy way. It's extra weird because I tested and you get that extra charge even when talented out of Death and Madness. I genuinely think this is a WoW dev being lazy.

src/analysis/retail/priest/shadow/modules/talents/DeathAndMadness.tsx
DoxAshe: That did help catch one reset that I was missing, thank you.

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
DoxAshe: That is a much better way to do it.  
I've been using this log /report/61fWqPBYTDx3M8rX/6-Normal+Gnarlroot+-+Kill+(2:54)/Publik/standard/statistics
It uses SW:P twice to spend stacks, and using tick gave the same results as checking when it was cast.

src/analysis/retail/priest/shadow/modules/tier/Tier29ShadowPriest4Set.tsx
DoxAshe: The Tier29 was always showing up in statistics even when it wasn't equipped.  I could take out the entire T29 from analysis if that would be better.

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
kfinch: Oh you even referenced the log in the PR lol, I should have used my brain

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
DoxAshe: I'll start working on using that

src/analysis/retail/priest/shadow/modules/tier/Tier29ShadowPriest4Set.tsx
kfinch: I think you should either add a `this.active` flag to Tier29 or just remove the module outright - doesn't seem right that it will be running when almost no one will have the set equipped

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
kfinch: I can mock up what I'm thinking if you'd like

kfinch: Here's what I have - the link specs in your normalizer would be something like this:

```ts
{
    linkRelation: CONSUMES_T31_BUFF,
    linkingEventId: [TALENTS.SHADOW_CRASH_TALENT.id, SPELLS.SHADOW_WORD_PAIN.id],
    linkingEventType: EventType.Cast,
    referencedEventId: SPELLS.SHADOW_PRIEST_TIER_31_4_SET_BUFF.id,
    referencedEventType: EventType.RemoveBuff,
    forwardBufferMs: BUFFER_MS,
    backwardBufferMs: BUFFER_MS,
    anyTarget: true,
  },
  {
    linkRelation: HITS_TARGET,
    linkingEventId: SPELLS.SHADOW_WORD_PAIN.id,
    linkingEventType: EventType.Cast,
    referencedEventId: SPELLS.SHADOW_WORD_PAIN.id,
    referencedEventType: EventType.Damage,
    forwardBufferMs: BUFFER_MS,
    backwardBufferMs: BUFFER_MS,
  },
  {
    linkRelation: HITS_TARGET,
    linkingEventId: TALENTS.SHADOW_CRASH_TALENT.id,
    linkingEventType: EventType.Cast,
    referencedEventId: TALENTS.SHADOW_CRASH_TALENT.id,
    referencedEventType: EventType.Damage,
    forwardBufferMs: SC_TRAVEL_BUFFER_MS,
    backwardBufferMs: BUFFER_MS,
    anyTarget: true,
  },
  ```
with constants
```ts
// doesn't really matter what these say, it's just for readability when inspecting events
const HITS_TARGET = 'HitsTarget';
const CONSUMES_T31_BUFF = 'ConsumesT31';

const BUFFER_MS = 100;
const SC_TRAVEL_BUFFER_MS = 3000; // long buffer to look for damage events to be safe - SC has much longer cooldown so this should be fine
```
and lookup helpers
```ts
export function consumedT31Buff(event: CastEvent): boolean {
  return HasRelatedEvent(event, CONSUMES_T31_BUFF);
}

export function getHits(event: CastEvent): DamageEvent[] {
  return GetRelatedEvents<DamageEvent>(event, HITS_TARGET);
}
```

DoxAshe: Sure, I haven't used a Normalizer before so a starting point would be great.

kfinch: Then in your actual analyzer you'd watch stack gains of the buff (because unfortunately you can't get prior stack count from the RemoveBuff) and also would watch casts of SWP and SC
```ts
  onBuffApplied(Event: ApplyBuffEvent) {
    this.stacks = 1;
  }

  onBuffStack(Event: ApplyBuffStackEvent) {
    this.stacks = Event.stack;
  }

onSwpCast(event: CastEvent) {
  if (consumedT31Buff(event)) {
    getHits(event).forEach(e -> {
        this.damageSWP += calculateEffectiveDamage(e, this.stacks * this.bonusSWP);
    }
  }
}

// similar to onSwpCast for onScCast
```

DoxAshe: Thank you!  I'll use all this as a starting point and get working on it

kfinch: It's possible the SWP hit linker also needs an `additionalCondition` to avoid accidentally tallying nearby ticks - not sure if this would practically happen tho

src/analysis/retail/priest/shadow/modules/tier/Tier29ShadowPriest4Set.tsx
DoxAshe: I've added `this.active` to both tier sets

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
DoxAshe: I've got the normalizer working.  Thank you for your help on getting started with it, it ended up working great.
I did need to add an `additionalCondition` but it was only 100ms of ticks so it didn't make much of a difference.

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
```diff
@@ -0,0 +1,164 @@
...
+    this.timestampOfLast = Event.timestamp;
+  }
+
+  onBuffRemoved(Event: RemoveBuffEvent) {
+    this.recentStacks = this.stacks;
+    if (MAX_DURATION <= Event.timestamp - this.timestampOfLast) {
+      this.recentStacks = 0;
+    }
+    this.stacks = 0;
+  }
```
kfinch: Unless I'm missing something, all this is superfluous now and can be removed

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4SetNormalizer.tsx
```diff
@@ -0,0 +1,77 @@
...
+  DamageEvent,
+} from 'parser/core/Events';
+import { Options } from 'parser/core/Module';
+import TALENTS from 'common/TALENTS/priest';
+import SPELLS from 'common/SPELLS';
+import { TIERS } from 'game/TIERS';
+const HITS_TARGET = 'HitsTarget';
+const CONSUMES_T31_BUFF = 'ConsumesT31';
+
+const BUFFER_MS = 1000;
```
kfinch: On my example I used a buffer of 100. This is a much longer buffer than we'd normally use for insta-applied things, and this is longer than the minimum GCD so false positives are possible. Is there a reason you made it this large?

kfinch (commented):
Looks very nice, just a couple more things I noticed

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4SetNormalizer.tsx
DoxAshe: It is a mistake, it should be shorter

src/analysis/retail/priest/shadow/modules/tier/Tier31ShadowPriest4Set.tsx
DoxAshe: I think I still need to calculate the number of stacks using `OnBuffApplied `and `OnBuffStack`, but otherwise you are right.  I thought I needed to make sure I had the right number of stacks in edge cases, but the normalizer handles all of them.

kfinch: Oh yeah you do still need the stack count. The 'recent applied' stuff is handled by the normalizer because we do the tally at the timestamp the cast happens instead of waiting for the damage.

kfinch (approved):

PR #6483, opened 2023-11-12
Refactored Rolling Average Active Time Graph, and updated which specs use it.

kfinch:
### Description

* Refactored `getUptimeGraph` to be a React component, and to do rolling average instead of cumulative average.
* Moved the rolling average calculation out of `AlwaysBeCasting` so that the computation will only happen for specs that actually want to display this.
* Updated the data points to be on a fixed time schedule rather than 'every other spellcast'
* Updated Shadow Priest and Elemental Shaman guides to use the new rolling average
* Added Active Time Graph to Balance Druid

### Testing

- Test report URL: `/report/MQzVBb2aWxYy7KD1/1-Mythic+Igira+the+Cruel+-+Wipe+1+(4:34)/Wallaa/standard`
- Screenshot(s):
image


src/parser/ui/ActiveTimeGraph.tsx
```diff
@@ -2,24 +2,111 @@ import BaseChart, { formatTime } from 'parser/ui/BaseChart';
...
+    }
+    const windowDuration = Math.min(ROLLING_AVERAGE_WINDOW_DURATION, timestamp - fightStart);
+    const activeTimePercentage = getRollingAverage(
+      activeTimeSegments,
+      currIdx - 1,
+      timestamp,
+      windowDuration,
+    );
+    graphData.push({ timestamp, activeTimePercentage });
+  }
```
emallson: you should use `useMemo` to keep re-renders from re-running this code. (not a correctness issue, just uses less CPU). something like this:

```tsx
  // Generate active time rolling average from the ABC module's active time segments
  const graphData = useMemo(() => {
    const graphData: GraphData[] = [];
    // existing code (github suggestion diff is terrible)
    return graphData;
  }, [fightStart, fightEnd, activeTimeSegments]);
```

emallson (approved):
looks good. one comment about (what i think will be) a pretty easy optimization to avoid recomputing chart data on rerenders.

src/parser/ui/ActiveTimeGraph.tsx
kfinch: Yeah good idea I'll add that

PR #6486, opened 2023-11-13
Druid Balance - tweaked cast effic requirements, added per-Incarn active time breakout, added Wild Mushroom section

kfinch:
### Testing

- Test report URL: `/report/VdQcjPHGwXtgDpKL/4-Heroic+Volcoross+-+Kill+(4:53)/Peenguu/standard/overview`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/druid/balance/modules/spells/CelestialAlignment.tsx
+++ b/src/analysis/retail/druid/balance/modules/spells/CelestialAlignment.tsx
@@ -1,0 +1,128 @@
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import { TALENTS_DRUID } from 'common/TALENTS';
+import { cdDuration, cdSpell } from 'analysis/retail/druid/balance/constants';
+import SpellLink from 'interface/SpellLink';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
+import Spell from 'common/SPELLS/Spell';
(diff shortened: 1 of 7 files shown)


src/analysis/retail/druid/balance/modules/spells/CelestialAlignment.tsx
```diff
@@ -0,0 +1,128 @@
...
+            percentActivePerf = QualitativePerformance.Fail;
+          }
+
+          const checklistItems: CooldownExpandableItem[] = [];
+          checklistItems.push({
+            label: <>Stay Active!</>,
+            result: <PerformanceMark perf={percentActivePerf} />,
+            details: <>({formatPercentage(cast.activeTimePercentage ?? 0, 0)}% active time)</>,
+          });
+          // TODO more checks?
```
emallson, line 105: i do like this check! especially for specs with really important CD windows, active time within the CD window seems very useful

i know a lot of the rules from SL incarn were venthyr-specific, but might be something about resources spent?

emallson (approved):

src/analysis/retail/druid/balance/modules/spells/CelestialAlignment.tsx
kfinch: I don't think raw spent AsP is a helpful metric, but I could also do a specific "don't overcap" rule within Incarn. Gonna check more with the Balance TC community about other things that could be added here

PR #6489, opened 2023-11-14
[Augmentation] Added functionality to find optimal buff targets for Breath Helper

Krealle:
### Description
I'll be honest with y'all, I completely forgot I wrote this a month ago x)

This adds another function to the Breath Helper module that will attempt to find better buff targets than the currently buffed ones, specifically for Breath windows.

Since Breath of Eons is based on actual damage dealt this is very accurate, non-buffed players will obviously be devalued, but as can be seen here:
image
Even when you only have a singular Aug in your raid it is still possible to find a massive gain. (eg. the non buffed players still outperformed the buffed players)
So we won't ever show harmful examples, but we will also not catch *all* scenarios where a better buff target could have been found; but for now that isn't really a concern, since at worst case we still show better options. 

And can't really be fixed until the combatlog is fixed up a bit by blizzard.

- Test report URL: `/report/fbP61CKGwhrmnHMq/14-Mythic+The+Amalgamation+Chamber+-+Kill+(4:33)/Gerryk/standard`


ljosberinn:
send it

ToppleTheNun (approved):

PR #6497, opened 2023-11-16
refactor: refatoring code smell missing union type

LucasLevyOB:
### Description

Refactoring Code Smell missing union types(or Type aliases) helps make the code more readable and reusable. Thus, the implementation of future refactorings and features can become easier and faster.

The modifications consisted of the creation of some types that join other types, in other cases the union was made through existing types


emallson (changes requested):
In general, I think all of these changes *except* the evoker one are simply less readable. They remove a clear inline union and replace it with a single-use type alias that lives at a different location in the file.

I'm happy enough to merge the Vega one in the evoker code but would prefer not to have the others. I'm open to other opinions on the matter, though

ToppleTheNun:
Agreed with emallson.

LucasLevyOB:
I'm afraid I agree with this too.

I made the requested change, @emallson.
Can you take a look again?

emallson (approved):
this looks good, thanks 👍

PR #6499, opened 2023-11-17
Rsham T31

BrocoliCosmique:
I reactivated the guide for restoration shaman, and tentatively implemented a module for T31.
2pc is pretty solid, 4pc is untested (no logs yet) and shaky. The 4pc detection is based on the same EventNormalizer as Primal Tide Core, so there will definitely be interferences if both 4pcT31 and the talent are active, to be refined when data is available.

This is my first commit, please teach me everything I did wrong :D


src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+  tidalReservoirHealing: number = 0;
+  tidalReservoirOverhealing: number = 0;
+  bonus4pcRiptideHealing: number = 0;
+  percent4pcRiptideHealing: number = 0;
+  bonus4pcRiptideAmount: number = 0;
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.has2PieceByTier(TIERS.T31);
+    this.has4pc = this.selectedCombatant.has4PieceByTier(TIERS.T31);
```
emallson, line 46: you don't need this field or to check it in your methods---they won't be called if `this.active` is false.

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+      this.percent4pcRiptideHealing += (event.amount + (event.absorbed || 0)) * RIPTIDE_4P_POWER;
+
+      if (event.tick) {
+        if (!this.riptideTracker.hots[targetId] || !this.riptideTracker.hots[targetId][spellId]) {
+          return;
+        }
+      } else if (isFrom4pcT31(event)) {
+        this.bonus4pcRiptideHealing +=
+          (event.amount + (event.absorbed || 0) + (event.overheal || 0)) * (1 - RIPTIDE_4P_POWER);
```
emallson: is it intentional to count overhealing here? you are including it in the HPS numbers shown in the statistic, which is unconventional for healing.

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+    if (this.has4pc) {
+      const targetId = event.targetID;
+      const spellId = event.ability.guid;
+
+      this.percent4pcRiptideHealing += (event.amount + (event.absorbed || 0)) * RIPTIDE_4P_POWER;
+
+      if (event.tick) {
+        if (!this.riptideTracker.hots[targetId] || !this.riptideTracker.hots[targetId][spellId]) {
+          return;
+        }
```
emallson: i'm not sure what this is doing. if the condition is true, it immediately returns. if the condition is false, it exits the nested `if`s and then immediately returns. it feels like this is maybe some WIP code that was not totally cleaned up (or forgotten and not finished?)

emallson (changes requested):
hey! this looks pretty good overall. i added a couple of comments for you to look at.

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
vohrr: Think this is fine since the stat has two separate line items for 2pc and 4pc - hides the 4pc block when only 2pc is equipped

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+    return this.bonus4pcRiptideHealing + this.percent4pcRiptideHealing;
+  }
+
+  on2pcHeal(event: HealEvent) {
+    this.tidalReservoirHealing += event.amount + (event.absorbed || 0);
+    this.tidalReservoirOverhealing += event.overheal || 0;
+  }
+
+  handleRiptideHeal(event: HealEvent) {
+    if (this.has4pc) {
```
vohrr: move this if statement to line 48 and wrap the event listener in it

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+import ItemSetLink from 'interface/ItemSetLink';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import Statistic from 'parser/ui/Statistic';
+import { formatNumber, formatPercentage } from 'common/format';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
+import RiptideTracker from '../core/RiptideTracker';
+import Combatants from 'parser/shared/modules/Combatants';
+
+const RIPTIDE_4P_POWER = 0.2; // Riptide bonus power = 25% => Tier set value = 25%/125% = 20% of total amount
```
vohrr: 25%

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  on2pcHeal(event: HealEvent) {
+    this.tidalReservoirHealing += event.amount + (event.absorbed || 0);
+    this.tidalReservoirOverhealing += event.overheal || 0;
+  }
+
+  handleRiptideHeal(event: HealEvent) {
+    if (this.has4pc) {
+      const targetId = event.targetID;
```
vohrr: This method should be rewritten imo, use your attribution check to determine if the entire hot is from 4pc and if it is tally all of the healing, otherwise just tally the effective 25% increase,. something like:

`if(isFrom4pcT31(event)) {
  this.fourPieceHealing += event.amount + (event.absorbed || 0); 
}
else {
 this.fourPieceHealing += calculateEffectiveHealing(event, RIPTIDE_4P_POWER);
}`

You don't need to check for ticks if the tooltip is accurate - the bonus applies to all riptide healing 
You also don't need separate variables for the 4pc healing this way, just use one `this.fourPieceHealing`

vohrr (changes requested):

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
BrocoliCosmique: You are correct, this is probably a copy/paste error

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
BrocoliCosmique: 25% is 20% of 125% (total rate of riptide healing with the bonus from the set applied).
I stored it this way to avoid having to constantly multiply the logged values by RIPTIDE_BONUS/(1+RIPTIDE_BONUS) to get the tier set effective bonus.
I can change it if you find it confusing.

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
BrocoliCosmique: I think the nesting is wrong, the else_if should be linked to the nested if and not the top level one. I'll look into it.

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
BrocoliCosmique: You're correct, this is the way it was done in the method I took inspiration from (module PrimalTideCore) and I just messed up the order. Thanks for the review and the suggestions !

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
vohrr: It is not confusing but you are writing unnecessary code instead of using the helper method shown in the example code that does all this for you

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
emallson: oh, you're right

src/analysis/retail/shaman/restoration/modules/dragonflight/Tier31.tsx
emallson: note that `addEventListener` is a no-op if `this.active === false`

emallson: i can't read, this is the 4pc check. carry on

emallson (approved):
this looks better to me. since @vohrr is already involved i'd like them to check the riptide maths

vohrr (approved):
LGTM

PR #6513, opened 2023-11-23
[Devastation] Update Disintegrate guide section

Krealle:
### Description

New tier set leaves you with a lot more resourcess so clipping is bit more relaxed outside of DR.


ljosberinn:
send it

ToppleTheNun (approved):

PR #6520, opened 2023-11-26
hotfix: MaelstromWeaponSpenders unhandled abilities error

danny-janse:
As reported by Hysteriix, MaelstromWeaponSpenders is disabled under specific conditions.

I identified this to certain abilities are not handles by the MaelstromWeaponTracker causing errors. This hotfix focuses on just getting the module to stay loaded by allowing MaelstromWeaponSpender to ignore these exceptions. Any of the exceptions will show 0 casts and spend instead so this still requires some love after it has been applied.

This can be tested with the following log;
 `/report/3d9RAc8qYfmL72KV/41-Normal+Tindral+Sageswift,+Seer+of+the+Flame+-+Kill+(5:31)/Hysteriix/standard/statistics`

image


Seriousnes:
Looks good to me

emallson (approved):

PR #6521, opened 2023-11-26
refactor: refactoring large file and too many props smells

LucasLevyOB:
### Description

Refactoring of Code Smells Large File and Too Many Props

Code Smell Large File, as the name suggests, is related to the size of React files, which may or may not be related to the size of the component. This type of Code Smell can cause difficulty in understanding the React component, as it has many functionalities integrated into the file, thus affecting possible tasks in them.

Code Smell Too Many Props is related to the number of props that are used in the component, it also relates components created within another component. Components with many props tend to have additional features and become more difficult to understand.

### Testing

Still in the testing phase, as there is a very large volume of modifications, I am still testing


LucasLevyOB:
Still in the testing phase, I created the PR just to have better control.

emallson (commented):
in its current form, this PR is effectively un-reviewable due to size and un-mergeable due to scope (aka there will basically always be merge conflicts).

if you want to work on this, please tackle one at a time instead of putting everything in one PR. additionally, you should *probably* check with the maintainer(s) of the specs on Discord to make sure that your changes aren't going to conflict with their ongoing work (e.g. i know there has been some ongoing work on Breath of Eons-related code)

PR #6522, opened 2023-11-26
refactor: refactoring code smell large component

LucasLevyOB:
### Description

Refactoring of Code Smell Large Component.

This Code Smell is related to the size of the React component, often components with a large number of lines have functionalities that can be separated into new components or new files. This Code Smell can improve code understanding and facilitate future development

### Testing

Still in the testing phase, as there is a very large volume of modifications, I am still testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "sass-text-stroke": "1.2.0",
     "swr": "^2.2.0",
     "universal-cookie": "^4.0.4",
+    "use-error-boundary": "^2.0.6",
     "vega": "~5.23.0",
     "vega-lite": "^5.6.0"
   },
--- a/src/analysis/retail/demonhunter/vengeance/modules/talents/CycleOfBinding.tsx
+++ b/src/analysis/retail/demonhunter/vengeance/modules/talents/CycleOfBinding.tsx
@@ -1,14 +1,8 @@
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import TALENTS from 'common/TALENTS/demonhunter';
 import SpellUsable from 'parser/shared/modules/SpellUsable';
-import Events, { CastEvent } from 'parser/core/Events';
-import {
-  getTargetsAffectedByElysianDecree,
-  getTargetsAffectedBySigilOfChains,
-  getTargetsAffectedBySigilOfFlame,
-  getTargetsAffectedBySigilOfMisery,
-  getTargetsAffectedBySigilOfSilence,
-} from 'analysis/retail/demonhunter/vengeance/normalizers/CycleOfBindingNormalizer';
+import Events, { ApplyDebuffEvent, CastEvent, DamageEvent } from 'parser/core/Events';
+import cycleOfBindingNormalizer from './cycleOfBindingImports';
 import {
   getElysianDecreeSpell,
   getSigilOfChainsSpell,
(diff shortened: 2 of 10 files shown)


LucasLevyOB:
Still in the testing phase, I created the PR just to have better control.

package.json
```diff
@@ -36,6 +36,7 @@
     "sass-text-stroke": "1.2.0",
     "swr": "^2.2.0",
     "universal-cookie": "^4.0.4",
+    "use-error-boundary": "^2.0.6",
```
emallson, line 39: not a fan of adding a dependency for this when we already have a working solution (and one that isn't even particularly large, this is a net-100 loc change *at best*)

src/analysis/retail/demonhunter/vengeance/modules/talents/cycleOfBindingImports.ts
```diff
@@ -0,0 +1,15 @@
+import {
```
emallson, line 1: this file doesn't need to exist, you can use `import * as cycleOfBindingNormalizer` if you want them namespaced instead of individually imported

src/interface/utils/ignoreErrorBoundary.ts
```diff
@@ -0,0 +1,16 @@
+import isTriggeredByExternalScript from './isTriggeredByExternalScript';
```
emallson, line 1: looks like you forgot to commit a file

src/interface/utils/handleErrorEventBoundary.ts
```diff
@@ -0,0 +1,13 @@
...
+  // SEE: https://github.com/facebook/react/issues/10474
+  if (error && error.stack && error.stack.includes('invokeGuardedCallbackDev')) {
+    return;
+  }
+  console.log('Caught a global error');
+
+  return error;
+};
+
+export default handleErrorEventBoundary;
```
emallson, line 13: this function should not be split out. it is small and as-is will pollute the global namespace (aka auto-import completions will start listing what is effectively an internal method for the top-level error boundary)

src/interface/utils/ignoreErrorBoundary.ts
```diff
@@ -0,0 +1,16 @@
...
+  // NOTE: These filters only prevent the error state from being triggered. Sentry automatically logs them regardless.
+  if (isTriggeredByExternalScript(error)) {
+    console.log('Ignored because it looks like a third party error.');
+    return true;
+  }
+
+  return false;
+};
+
+export default ignoreErrorBoundary;
```
emallson, line 16: similar comment here: this is a 12-line function that doesn't really belong in the global namespace. it should remain a local `const` within the root error boundary file.

src/analysis/retail/demonhunter/vengeance/modules/talents/CycleOfBinding.tsx
```diff
@@ -67,103 +40,37 @@ export default class CycleOfBinding extends Analyzer.withDependencies(deps) {
...
-      Events.cast.by(SELECTED_PLAYER).spell(sigilOfSilenceSpell),
-      this.onSigilOfSilenceCast,
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(sigilOfSilenceSpell), (event) =>
+      this.onEvent(event, cycleOfBindingNormalizer.getTargetsAffectedBySigilOfSilence),
     );
-    this.addEventListener(
-      Events.cast.by(SELECTED_PLAYER).spell(sigilOfChainsSpell),
-      this.onSigilOfChainsCast,
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(sigilOfChainsSpell), (event) =>
+      this.onEvent(event, cycleOfBindingNormalizer.getTargetsAffectedBySigilOfChains),
```
emallson, line 56: this is a pretty good refactor. could be cleaner if the list of getters were indexed by spell so you could just loop over them, but that doesn't exist as-is so not too fussed.

emallson (commented):
similar comment as for the PR about large files, but less extreme since this one is *far* smaller: if you're going to add more to this, please use a separate PR for different specs to keep it reviewable

some of these refactorings are pretty questionable. why split up `RootErrorBoundary`? 170 lines is perfectly reasonable. similarly: why use this `cycleOfBindingImports` file? 

in general, these blind "component too large" refactorings don't really respect the cohesion and natural boundaries of a component. *it is okay to have a 200 line file!* it is even okay to have a 500 line file! code organization & project architecture is not about lines of code.

while I appreciate the refactor to a functional error boundary, due to [limitations of react](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) all functional error boundaries are secretly class error boundaries, so refactoring it in-tree to a functional component is unfortunately not possible (and refactoring it to use a FC imported from an external dep only moves the problem to where we can't see it)

src/analysis/retail/demonhunter/vengeance/modules/talents/CycleOfBindingStatistic.tsx
ToppleTheNun, line 1: If you're going to split components out like this, please make the name of the talent a folder and put the analyzer in `index.tsx` and any child components in that folder.

src/analysis/retail/demonhunter/vengeance/modules/talents/CycleOfBinding.tsx
```diff
@@ -67,103 +40,37 @@ export default class CycleOfBinding extends Analyzer.withDependencies(deps) {
...
-    const affectedBy = getTargetsAffectedBySigilOfSilence(event);
-    if (affectedBy.length === 0) {
-      return;
-    }
-    this.reduceSigilCooldowns();
-  }
-
-  private onSigilOfChainsCast(event: CastEvent) {
-    const affectedBy = getTargetsAffectedBySigilOfChains(event);
+  private onEvent(
```
ToppleTheNun, line 69: I am a fan of this refactor.

PR #6540, opened 2023-12-01
[Shadow]Fix Shadow Word: Death Max Casts, and Fix Deathspeaker's Cooldown Rest

DoxAshe:
### Description

Shadow Word Death's maximum casts have been incorrect, and even off by double the correct amount with Deathspeaker talented.  I have updated Shadow Word Death's maxCast calculation, and fixed issues with Deathspeaker.
Deathspeaker has been resetting all charges of shadow word: death instead of only giving one charge back.  The reason this is happening is because `abilities.increaseMaxCharges` does not update the maxCharges with `spellUsable.endCooldown`.  This causes both charges to be reset, instead of only one.

To work around this issue, I have used `spellUsable.OnMaxChargesIncrease` to increase the charges, which fixes the issue.  I did it this way because this fix should not impact other specs.

### Testing

I have tested this with combinations of deathspeaker and the four piece, and in all situations this now calculates the maximum casts correctly.  However, when a player without the four piece is facing multiple targets, the guide graph does not always properly show the casts or available segments.  This was occuring before this change, and is something that I plan on fixing later.

Test report URL: 

- With Deathspeaker and 4 piece: '/report/32hnqfXN6ZVkQPLG/4-Normal+Volcoross+-+Kill+(2:22)/Doxashe/standard'
- With Deathspeaker and no 4 piece: '/report/X6kmRCdHvNWAQzZF/15-Normal+Volcoross+-+Kill+(1:25)/Anàtéma/standard/overview'
- Without either: 'report/MQphqtvxW13wj4Yf/64-Normal+Volcoross+-+Kill+(2:00)/Mentalistwo/standard/overview'

### DIFF
--- a/src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
+++ b/src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
@@ -8,6 +8,8 @@ import SPELL_CATEGORY from 'parser/core/SPELL_CATEGORY';
 import ExecuteHelper from 'parser/shared/modules/helpers/ExecuteHelper';
 import { TIERS } from 'game/TIERS';
 import { SHADOW_WORD_DEATH_EXECUTE_RANGE } from '../../constants';
+import DeathAndMadness from '../talents/DeathAndMadness';
+import Deathspeaker from '../talents/Deathspeaker';
 
 class ShadowWordDeath extends ExecuteHelper {
   static executeSources = SELECTED_PLAYER;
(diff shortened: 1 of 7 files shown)


src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
```diff
@@ -53,15 +57,21 @@ class ShadowWordDeath extends ExecuteHelper {
...
-      //The Death and Madness talent lets you cast twice each cast in execute.
-      this.maxCasts += ExecuteCasts;
-    }
-    this.maxCasts += ExecuteCasts + this.totalNonExecuteCasts;
+    const DeathAndMadnessCasts = this.deathandmadness.getResets();
+    const DeathSpeakerCasts = this.deathspeaker.getProcsUsed();
+
+    //console.log("SWD Totals:",ExecuteCasts,"NE",this.totalNonExecuteCasts,"DM",DeathAndMadnessCasts,"DS",DeathSpeakerCasts);
+
+    this.maxCasts +=
```
emallson: should this be `this.maxCasts =` instead of `+=`? i don't think this method *should be* called twice but not sure its guaranteed not to be

emallson:
> The reason this is happening is because abilities.increaseMaxCharges does not update the maxCharges with spellUsable.endCooldown. This causes both charges to be reset, instead of only one.

this sounds like a bug that should be fixed instead of worked around. let me do some looking

src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeath.tsx
DoxAshe: It should be.  I've fixed this.

src/parser/shared/modules/SpellUsable.ts
DoxAshe, line 1: I think the issue comes on line 302, because `cdInfo.maxCharges` remains at one instead of increase to two.  This causes the code to delete the current Cooldowns, leading to the spell going to its maximum number of charges, which is two.
This should change when from fabricated IncreaseMaxCharges Event that should be occurring with `abilities.increaseMaxCharge`, but I cannot figure out why it isn't.

emallson (approved):

PR #6547, opened 2023-12-03
[FROST DK] KM out-of-gcd Fix

Bicepspump:
### Description

The code currently counts Killing Machine generations that are out-of-GCD, categorising them as minor mistakes due to the player not being able to prevent them. However, this is also currently catching manually generated KM procs that do not occur outside the GCD. The reason for this is that the KM buff event occurs just after the actual GCD (0 or 1 MS) and is therefore caught as inside a GCD.

### Testing

- Test report URL: https://www.warcraftlogs.com/reports/ZqFVTL8Q9AwBgf7M#fight=17&type=damage-done&source=430
- Screenshot(s):
New -> image
Old -> image

### DIFF
--- a/src/analysis/retail/deathknight/frost/modules/features/KillingMachine.tsx
+++ b/src/analysis/retail/deathknight/frost/modules/features/KillingMachine.tsx
@@ -97,7 +97,8 @@ class KillingMachineEfficiency extends Analyzer {
     // 3/24/23, trying out disabling lag tolerance for km refreshes if the player has fatal fixation talented, may need more work
     if (
       (!this.fatalFixation || (this.fatalFixation && this.currentStacks === 2)) &&
-      timeSinceGCD < this.lastGCDDuration + LAG_BUFFER_MS
+      timeSinceGCD < this.lastGCDDuration + LAG_BUFFER_MS &&
+      timeSinceGCD > 1
     ) {
       this.procsWithinGcd += 1;
       return;


Khazakdk (changes requested):
You’ve changed the behavior of the module without updating the UI to clearly communicate what’s going on.  In addition, like I said on Discord, you need to use something more like an apl module for tracking manual km generation.  Red errors are for undisputed mistakes.  There are times when manual km generation may lead to proc wastes due to resource constraints.

Bicepspump:
> You’ve changed the behavior of the module without updating the UI to clearly communicate what’s going on. In addition, like I said on Discord, you need to use something more like an apl module for tracking manual km generation. Red errors are for undisputed mistakes. There are times when manual km generation may lead to proc wastes due to resource constraints.

The current UI clearly states that the red KMs should be "Killing Machines refreshed during an available GCD" which the functionality doesn't do. If you are in your pillar and generate a KM manually on an available GCD, it counts it as being in a GCD and adds it to the yellow bar.

My change is simply to fix this incorrect behavior, not change the functionality.

Khazakdk:
The wording is accurate.  It might need additional explanation when oblit is talented but remember the module needs to work for all builds.  Again, what really needs to be done is a specific module for obliteration that considers the current state of the player

Khazakdk:
We only flag things to the red bar when it’s caused by a player mistake.  KMs refreshed inside pillar are not necessarily player mistakes.

ToppleTheNun:
With feedback given and not addressed for a month, I'm closing the PR.

If the comments are resolved, feel free to reopen this PR.

PR #6572, opened 2023-12-19
upscale our headshot images for icc/amirdrassil to match what is actu…

emallson:
this greatly improves the visual quality (IMHO). they're less pixely/blurry. there are limits on what can be done given that they're sourced from pretty small images, but this at least is better

for ref: these images show at 91x91 on my screen, so this is *slightly* oversized but I was upscaling with waifu2x and didn't want to give a long float for the scale factor to try to get exactly that :upside_down_face:


emallson:
for ref, these are the before/after side-by-side at the same display size:

## Amirdrassil

<div>


</div>

## ICC
<div>


</div>

PR #6574, opened 2023-12-21
bring in eslint config from vite branch

ToppleTheNun:
### Description

pnpm causes a few issues with the dependency resolution with the old eslint config and inlining it all into our own .eslintrc.js fixes it.

All of the dependencies used here should theoretically already exist on developer machines and thus not need a full reinstall - verified.


.eslintrc.js
```diff
@@ -1,18 +1,186 @@
...
+    // See https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore for
+    // reasons why.
+    // https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore/blob/master/lib/rules/all.js
+    'plugin:you-dont-need-lodash-underscore/compatible',
+    'plugin:react/jsx-runtime',
+  ],
+  plugins: ['wowanalyzer', 'progress', 'no-only-tests'],
   rules: {
-    'progress/activate': process.stdout.isTTY ? 1 : 0,
+    'progress/activate': 1,
```
emallson: don't want this I don't think? it means any tool calling `eslint` for JSON output breaks since progress data gets dumped

.eslintrc.js
```diff
@@ -1,18 +1,186 @@
+require('dotenv-flow').config({
+  silent: true,
+});
```
emallson, line 3: what's the story here? were we not respecting .env files properly before this?

.eslintrc.js
ToppleTheNun: yep, will get that fixed

.eslintrc.js
ToppleTheNun: This was included in the eslint config we were using previously; most of the contents in the .eslintrc.js here are just copied from the existing config.

emallson (approved):

PR #6583, opened 2023-12-30
[WIP] Forced Downtime events

kfinch:
### Description

Adds a Normalizer that inserts fabricated "forced downtime" events into the event stream. Specific fights can be configured to add forced downtime events to times when the selected player doesn't have targets to damage / heal due to way the fight flows (for example Tindral intermissions). Further modules can consume the downtime events to give better analysis - for example not yelling at the player for failing to use a cooldown during an intermission.

This pull request includes the normalizer, and a scheme to easily specify downtime windows for fights. This pull request does NOT include modules that consume the events. The normalizer on its own does not do anything visible to the user.

Planned follow-on work to use these events:
* Show forced downtime windows in most 'timeline' views / graphs
* Don't count forced downtime against the player for 'active time' calculation
* Don't count forced downtime against the player for 'cooldown util' calculation

Areas for discussion:
* What do people think should count as 'forced downtime'? Tindral intermissions are pretty unambiguous, but there are many more gray areas.
  * Should the short period with no target between Fyrakk P1 and I1 count?
  * Should we try to detect when all adds are dead in Gnarlroot intermission and count that?
  * Should we count some segments as downtime for damage dealers but not healers? (ex. Painsmith Intermission) 
  * Should any player specific mechanics count as 'Forced Downtime'? I'd say we should be as conservative as possible with this, and the only mechanic that should count for this in DF is Sark 'downstairs' phase.
* What modules should use these events?
* Is everyone happy with the field definitions for the `DowntimeSpec` and the `ForcedDowntimeEvent` objects?

### Testing

Check fights with configured downtime and make sure the events are inserted as expected

- Test report URL: `/report/9zxZVkYqb3aWdCLf/22-Heroic+Tindral+Sageswift,+Seer+of+the+Flame+-+Kill+(7:13)/Azlann/standard/events`


kfinch:
I'm planning to revisit this for TWW S1. Looks like there will be a few fights with downtime where having this would be useful.

emallson:
I spent some time reading through this yesterday and I think you're basically going in the same direction that I was considering for dealing with Arms (and other downtime specs) in Cata: generate "gaps" where downtime is acceptable, feed that to ABC, ~~profit~~

(i have written *no code* for the arms thing yet soooooo no conflicts no worries)

PR #6585, opened 2023-12-31
[Fire Mage] Living Bomb, Meteor, Combustion Active Time

Sharrq:
- Adjusted Living Bomb to count casts that hit 5+ targets as ok, and adjust wording accordingly
- Adjusted Combustion Active Time so it actually counts properly and in seconds instead of milliseconds
- Adjusted wording of Meteor to indicate that it needs to land during Combustion.

Also adjusted AlwaysBeCasting in Core to be able to get the Active Time in milliseconds as well as percent, and adjusted the percent function to call the milliseconds one to avoid duplicating the entire function.

### DIFF
--- a/src/analysis/retail/mage/fire/CHANGELOG.tsx
+++ b/src/analysis/retail/mage/fire/CHANGELOG.tsx
@@ -6,6 +6,9 @@ import { Sharrq, ToppleTheNun } from 'CONTRIBUTORS';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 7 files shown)


Sharrq:
@kfinch @emallson Can one (or both) of you check the changes i made to AlwaysBeCasting in Core? I adjusted the new Active Time Percent function so you can also get Milliseconds instead if you want.

kfinch:
Yes the changes in core look correct to me

kfinch (approved):

PR #6586, opened 2023-12-31
[Evoker] Improved accuracy of EssenceTracker module

Krealle:
### Description

Previously essence tracking looked like this:
image

With the changes I've implemented it now looks like this: 
image

The current implementation of resource tracking for essence had a few flaws in it that made it wildly inaccurate, with these changes it is now able to accurately keep track of Essence.

Tagging @trevorm4 since it technically touches Preservation too.

### SpellEssenceCost (Shared)
- Implemented module for Essence reduction talents `Dense Energy` & `Volcanism`.

### EssenceTracker (Preservation -> Shared)
- EssenceTracker now properly changes `regenRates`, it was previously changing the `currentRegenRate`, instead of changing the `baseRegenRate`, resulting in unwanted interactions.
- Fixed Preservation getting `+1 maxEssence` based on `Font of Magic` talent instead of `Power Nexus` talent.
- Added `Power Swell` buff for `regenRate`.
- Added `SpellEssenceCost` dependency.

### Guides (Devastation & Augmentation)
- Updated Devastations `EssenceGraphSection`
- Implemented Augmentation `EssenceGraphSection`

### ResourceTracker (Core)
This module had to have a few changes to properly support Essence tracking. Currently `ResourceTracker` assumes whole numbers for resources, which works great for most, but Essence is a bit funky. Essence is a 0-5(6 with talent) ressource that isn't defined by `int` but `float` eg. you wont have a linear generation like this: `1 -> 2 -> 3 -> 4` but rather `1 -> 1.2 -> 1.3 -> 1.5 -> etc..`. 
- Therefore I've implemented `useGranularity` override that will enable tracking resources using `float` instead of `int`.
- Implemented `initialResources` override such that a starting amount can be defined.
- Implemented `adjustResourceMismatch` override that will make sure an update at the `beforeAmount` is pushed alongside the new `current` - otherwise the resource gain from `prevUpdate` and `current` won't be plotted, resulting in "impossible" plots.

Essentially, we only update resources based on certain events. So when we `spend` there will be some form of discrepancy as to what `current` values we have in our `fightData` array. 
Our `prevUpdate` might show `2.5` (marked in blue) whilst our `beforeAmount` (shown as green) says `3.1`. Without this change we won't ever push that we were at `3.1` ressouces, and therefor get inaccurate/impossible plots.

image

A practical example can be seen here:
The above graph is with `adjustResourceMismatch=true` and the lower graph is with `adjustResourceMismatch=false` All spenders here cost `3`.
image

Happy new years 🥳 

### Testing
- Test report URL: 
Devastation: `/report/JwKkvLhd3McHfjnm/21-Mythic+Volcoross+-+Kill+(3:33)/Vollmer/standard`
Augmentation: `/report/JwKkvLhd3McHfjnm/21-Mythic+Volcoross+-+Kill+(3:33)/Evozerg/standard`
Preservation: `/report/a:ZrzCf14FGAgpWPB9/26-Mythic+Smolderon+-+Wipe+26+(7:06)/Player+(49)/standard/statistics`


src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -157,6 +159,15 @@ export default class ResourceTracker extends Analyzer {
    *  timestamp as a valid gain.
    */
   allowMultipleGainsInSameTimestamp = false;
+  /** Instead of calculating resource as rounded numbers, use decimal precision for granularity.
+   * This is needed for specs that have resource generation that isn't provided in whole numbers
+   * eg. Evoker's Essence */
+  useGranularity = false;
+  /** amount of decimals to use for granularity */
+  granularity = 2;
```
emallson, line 167: the way this is usually handled is by multiplying by the number of places that you want. i.e. for a granularity of 2 as specified here, you multiply by 100. WCL does some of this internally

it looks like there are only 2 spots that `useGranularity` really comes into play, and one of those i'm not clear on (see below)

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
```diff
@@ -405,11 +416,43 @@ export default class ResourceTracker extends Analyzer {
       prevUpdate &&
       timestamp <= prevUpdate.timestamp + MULTI_UPDATE_BUFFER_MS;
     const beforeAmount =
-      reportedBeforeAmount !== undefined && !withinMultiUpdateBuffer
+      reportedBeforeAmount !== undefined && !withinMultiUpdateBuffer && !this.useGranularity
```
emallson, line 421: whats the role of the `!this.useGranularity` here?

Krealle: `reportedBeforeAmount` is a rounded number, whilst the `calculatedBeforeAmount` have been changed to return non-rounded numbers with `this.useGranularity=true`, so we need to use the `calculatedBeforeAmount` if we want to have decimals/granularity.

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
Krealle: Is this what you have you in mind?
`granularity = 100;`
```
const naturalGain = this.useGranularity
      ? Math.round(timePassedSeconds * lastUpdate.rate * granularity ) / granularity )
      : Math.round(timePassedSeconds * lastUpdate.rate); // whole number amount of resources pls

```

emallson: rereading my comment, i'm not sure what im suggesting tbh D:

this is probably fine as-is

src/parser/shared/modules/resources/resourcetracker/ResourceTracker.ts
emallson: this could use a comment just to make sure that nobody removes the condition since the line *looks* valid without it

Krealle: Added a comment 👍

PR #6591, opened 2024-01-04
Add statistics for Echoing Tyrstone trinket.

Arbixal:
### Description

Added info statistics box for Echoing Tyrstone trinket.

### Testing

Go to a report where a healer is using Echoing Tyrstone. Check values against the values in the report on warcraft logs (tooltip has more details about calculations).

- Test report URL: `/report/b64983ymw7FZJ1dH/19-Heroic+Tindral+Sageswift,+Seer+of+the+Flame+-+Kill+(7:46)/Bixshift/standard/statistics`
- Screenshot(s):
image


src/parser/retail/modules/items/dragonflight/EchoingTyrstone.tsx
```diff
@@ -0,0 +1,219 @@
...
+import Events, { ApplyBuffEvent, HealEvent, Item } from 'parser/core/Events';
+import BoringSpellValueText from 'parser/ui/BoringItemValueText';
+import ItemPercentHealingDone from 'parser/ui/ItemPercentHealingDone';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import Statistic from 'parser/ui/Statistic';
+
+const ECHOING_TYRSTONE_COOLDOWN = 120000; // 120sec
+
+// Need a better way to do this. The calculateSecondaryStat functions don't seem to line up
```
emallson: looks like our stat scaling multipliers are out of date

Arbixal: Even updating the values (which are very minor fractional changes) the values don't line up. Is on-use scaling different?
image
image

emallson: it may be using a different curve. i don't recall offhand how to find the scaling curve that a spell uses

emallson: i am digging into this, we only have a couple of the CR scaling multiplier tables copied over, it may be on a non-trinket/jewelry one. trying to source an in-game screenshot of the trinket to confirm that wowhead has it right

emallson: okay so part of the issue is that wowhead scaling is wrong. im checking the calculated value against what we compute with the trinket multiplier table and will see what i find

## Wowhead

image
https://www.wowhead.com/item=207552/echoing-tyrstone?ilvl=483

## In-Game
image

emallson: image

this is the 483 version. the 273 (in-game, previous comment) vs 272 shown here is just down to rounding (wow always rounds up in tooltips iirc)

emallson: for the heal value, we need this table, ugh https://github.com/simulationcraft/simc/blob/dragonflight/engine/dbc/generated/rand_prop_points.inc

the formula is just `base_value * (damage_secondary(actual ilvl) / damage_secondary(base ilvl))` where `damage_secondary` is the 3rd column (its `ilvl, damage_replace_stat, damage_secondary`)

we *almost never* need this formula because we can read the healing/damage events but tyrstone's cap obviously matters a bit

i can copy it in, but the constants (and stat scaling) will change again next expansion (the formula probably won't though)

emallson (approved):

PR #6604, opened 2024-01-12
Guide mode for frost mage

earosselot:
### Description

I'm working on adding the Guide mode to frost mage.
I've only added a ABC graph to check if I'm on the right track because this is my first PR.

I think that the default view is still the Checklist. I tested several times, but I couldn't find that configuration. That is the reason behind the WIP tag...

image

### DIFF
--- a/src/CONTRIBUTORS.ts
+++ b/src/CONTRIBUTORS.ts
@@ -2340,3 +2340,8 @@ export const dub: Contributor = {
   github: 'wtodom',
   discord: 'its_me_dub',
 };
+
+export const Raistlinn: Contributor = {
+  nickname: 'Raistlinn',
+  github: 'earosselot',
+};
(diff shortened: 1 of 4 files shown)


emallson:
> I think that the default view is still the Checklist. I tested several times, but I couldn't find that configuration. That is the reason behind the WIP tag...

if you don't set `guideDefault: true` then the UI defaults to Checklist. it doesn't look like you have 👍

emallson (commented):
code-wise this looks fine. gonna tag @Sharrq since they're the current mage person and I don't know what plans they have for guide/checklist

Sharrq:
I do not have any current plans for the checklist or the guide. My focus is on getting all 3 mage specs up to par first before I looked into replacing the checklist with the guide.

That said, I dont like the checklist, so I am fine with someone setting up the guide. But the screenshot in the PR only shows active time, and I would rather not replace the checklist with the guide if it is only active time. I havent checked out the branch or anything to look at it so maybe you have more in there than what is in the screenshot

earosselot:
I only added the active time to see if its that was the way to do it, and to keep PRs little. I'll expand the section with more spells soon. The checklist is still the default view until we are happy with the guide :)

I could add more spells to this PR, but it will get bigger...

Sharrq:
Yea I am fine with you adding the guide a little at a time as long as you dont change the default view from the checklist until it is done

PR #6616, opened 2024-01-17
Frost mage Guide - Icy Veins - Ray of Frost - Comet Storm

earosselot:
### Description

Added guide panels for Icy Veins - Ray of Frost - Comet Storm.

Comet storm has the explaination for AoE, but the cast efficiency is not taking it into account. I will work on AoE once single target is done for most of the spells.

- Screenshot(s):
image
image


src/analysis/retail/mage/frost/talents/RayOfFrost.tsx
```diff
@@ -1,13 +1,52 @@
...
+const SpellSeq = ({ spells }: { spells: Spell[] }) => (
+  <SequenceContainer>
+    {spells.map((spell, index, array) => (
+      <Fragment key={index}>
+        <SpellIcon spell={spell} key={index} />
+        {index < array.length - 1 && <ChevronIcon />}
+      </Fragment>
+    ))}
+  </SequenceContainer>
+);
```
kfinch: This is a useful UI snippet, please move it to a new standalone file in `src/parser/ui/...` so it can be used by everyone.

src/analysis/retail/mage/frost/core/IcyVeins.tsx
```diff
@@ -28,17 +35,25 @@ class IcyVeins extends Analyzer {
...
+      Events.removebuff.by(SELECTED_PLAYER).spell(TALENTS_MAGE.ICY_VEINS_TALENT),
       this.onIcyVeinsEnd,
     );
     this.addEventListener(Events.fightend, this.onFightEnd);
+    // Guide
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(TALENTS_MAGE.ICY_VEINS_TALENT),
+      this.generateCastEntryOnBuffEnd,
+    );
+    this.addEventListener(Events.fightend, this.generateCastEntryOnFightEnd);
```
kfinch: Pretty sure this will also pickup Time Anomaly procs, which may not be desired.

My first thought on how to handle this is to add a `SPELL_CAST` link to `CastLinkNormalizer` for Icy Veins, allowing you to differentiate casts vs procs on the applybuff, and then you can check that link and disregard procs. A possible issue with that could be a bad linkage if player gets TA proc and then activates Icy Veins hardcast (I simply don't know how that will look in logs and it might mess with things)

kfinch (commented):
A few comments but overall looks good.

A general comment about guide text: I think you should do an editing pass to cut down on 'wordiness' of explanations. Multi paragraph sections can be intimidating to look at - make an explanation as short as possible while still conveying needed information

Sharrq:
> A few comments but overall looks good.
> 
> A general comment about guide text: I think you should do an editing pass to cut down on 'wordiness' of explanations. Multi paragraph sections can be intimidating to look at - make an explanation as short as possible while still conveying needed information

I agree with this. Also quite a few typos and grammatical issues

src/analysis/retail/mage/frost/core/IcyVeins.tsx
earosselot: Thanks! I will look on what you said. I think I have some TA logs to test this situation. I'll let you know what I found....

earosselot:
I'll work on the explanations to make them more concise. I'll let you know when is done. Thanks for the reviews !

src/analysis/retail/mage/frost/talents/RayOfFrost.tsx
earosselot: Done, extracted from Mistweaver Monk too (where I tooked from).

src/analysis/retail/mage/frost/core/IcyVeins.tsx
earosselot: Do as you said, and test it on raid. Waited until proc pops and after casted IV. It worked.
Buff (TA proc) started at 0:05, and I casted at 0:08. The analysis only watch for the "casted" portion.
image
image

earosselot:
Well, I made the explanations more concise, without loosing information. I also improved tooltips a little bit. For Comet Storm I separated explanations for whether you're talented with Coldest Snap or not. I think is a lot better now :)
image
CS without CS
image
CS with CS
image

Sharrq:
The explanations are better, but there are still some grammatical things and some stuff that is worded a little weirdly. That said, I can adjust those grammatical things after you finish to clean it up a little bit. 

Just for consistency to avoid straying too much from what I have been doing for all the mage specs, can you change all the Talent references back to TALENTS.SPELL_NAME instead of TALENTS_MAGE.SPELL_NAME. I know some people in core modules specify the class the talent comes from, but I dont see the need to do that in mage specific modules, and I would rather everything in mage be using the same style instead of some using TALENTS.* and some using TALENTS_MAGE.* .... unless there is some reason it has to be that way for the guide, but i dont think that is the case.

earosselot:
@Sharrq done the TALENTS thing back to mage original. I'm not a js expert and tought it might be better in some way given that all the other classes have it that way. But I agree with the consistency across all the mage specs is valuable. It's ready now to fix the grammatical issues you found.

kfinch (approved):
I think this is good enough to go in - easy to put further updates in later PRs. Some general comments:
* I like the bullet point formatting in the guide text, improves readability
* You're correct that `TALENTS` vs `TALENTS_MAGE` doesn't matter, I think Sharrq just wants his Mage stuff to use TALENTS
* I noticed on your 'cast eval' boxes, you used all the colors - you shouldn't feel obligated to do this and in fact you usually shouldn't. A lot of spells are just green/red for me. I'm still fine with how you have it, but I think it deserves a second look on follow on work.

kfinch:
In particular regarding the cast eval colors, I don't like that Ray of Frost can have a clipped tick and still be evaluated 'Good'. Our design up to this point is that 'Good' means 'did nothing wrong', and 'Perfect' is only sometimes used when the player gets something extra.

Sharrq:
Yea, you can proceed without the grammatical stuff, ill do a pass once the entire guide is done and do it all at once.

earosselot:
> In particular regarding the cast eval colors, I don't like that Ray of Frost can have a clipped tick and still be evaluated 'Good'. Our design up to this point is that 'Good' means 'did nothing wrong', and 'Perfect' is only sometimes used when the player gets something extra.

Nice to know. I'll have this in mind for the next modules, and I'll correct Ray of Frost later as well.

PR #6626, opened 2024-01-20
Attribute Lifebind healing in grace period module

trevorm4:
### Description

When the source heal for a lifebind heal event is done to a target with grace period it increases the healing done by the lifebind event on the linked target. This PR adds functionality to calculate the healing increase for this case and attribute it in the grace period module


emallson (approved):
this looks reasonable, even if i don't fully follow the evoker word salad

PR #6630, opened 2024-01-23
Winter's Chill guide Frost mage

earosselot:
### Description

Added Flurry's Winter's Chill to guide view, building on top of Winter's Chill module.

I also found and fixed an issue with the filtering inside the method WintersChill.missedPrecasts(). It was not working if the precast didn't shatter since it did not appear in Winter's Chill damage events. This was due to the cast ability GUID being different from the damage ability GUID, and they were being compared for equality.
glacial spike: damage.guid = 228600 / cast.guid = 199786
frostbolt: damage.guid = 228597 / cast.guid = 116

- Screenshot(s):
image
image
image
image


Sharrq:
The guide suggestions and such look fine to me, but would need someone else to look at the code to make sure that is good.

earosselot:
@trevorm4 can you look at the code please ?

emallson:
i can go through the code in a bit

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
```diff
@@ -0,0 +1,157 @@
...
+    let performance = QualitativePerformance.Fail;
+    if (this.precast && this.precastInDamageEvents()) {
+      performance = QualitativePerformance.Good;
+    } else if (!this.precast && this.precastIcicles === 4) {
+      performance = QualitativePerformance.Good;
+    } else if (this.precast && !this.precastInDamageEvents()) {
+      performance = QualitativePerformance.Fail;
+    } else if (!this.precast) {
+      performance = QualitativePerformance.Fail;
+    }
```
emallson, line 86: you can technically skip these two branches since the default is fail. however, leaving them as a sort of documentation of the intended behavior is fine imo

src/analysis/retail/mage/frost/normalizers/CastLinkNormalizer.ts
```diff
@@ -29,6 +29,18 @@ export const SPELL_DAMAGE = 'SpellDamage';
...
 
 const EVENT_LINKS: EventLink[] = [
+  {
+    reverseLinkRelation: SPELL_CAST,
+    linkingEventId: SPELLS.FROSTBOLT.id,
+    linkingEventType: EventType.Cast,
+    linkRelation: SPELL_DAMAGE,
+    referencedEventId: SPELLS.FROSTBOLT_DAMAGE.id,
+    referencedEventType: EventType.Damage,
+    anyTarget: true,
```
emallson: is `anyTarget: true` enabled because of splitting ice? but `maximumLinks: 1` is set. the `anyTarget` setting can probably be removed

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  getPerformanceDetails(): JSX.Element {
+    const spellsArray: Spell[] = Object.values(SPELLS);
+    const precast: Spell | undefined = spellsArray.find(
+      (spell) => spell.id === this.precast?.ability.guid,
+    );
+    const flurry: Spell | undefined = spellsArray.find(
+      (spell) => spell.id === this.flurry?.ability.guid,
+    );
```
emallson: the `SPELLS` table is also indexed by spell id, so you can replace these with:

```ts
const precast = this.precast ? SPELLS[this.precast.ability.guid] : undefined;
const flurry = this.precast ? SPELLS[this.flurry.ability.guid] : undefined;
```

a couple notes on that:

1. if the Flurry spell is actually constant, you should just use `SPELLS.FLURRY_NAME` instead of looking it up
2. if you expect that the precast/flurry spells may sometimes not exist (e.g. precasting a non-mage spell?), you should use `maybeGetSpell` instead of directly indexing `SPELLS`

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
```diff
@@ -0,0 +1,157 @@
...
+    if (precast) {
+      spells.push(precast);
+    }
+    if (flurry) {
+      spells.push(flurry);
+    }
+    this.damageEvents.forEach((damage) => {
+      const spell: Spell | undefined = spellsArray.find(
+        (spell) => spell.id === damage.ability.guid,
+      );
```
emallson: same note about using `SPELLS[ability.guid]` / `maybeGetSpell(ability.guid)`

src/analysis/retail/mage/frost/core/WintersChill.tsx
```diff
@@ -236,6 +259,91 @@ class WintersChill extends Analyzer {
...
+            {iceLance}
+          </small>
+        </ul>
+        Your rotations should look like this: <br />
+        <SpellSeq spells={[SPELLS.FROSTBOLT, TALENTS.FLURRY_TALENT, cooldown, cooldown]} /> <br />
+        <SpellSeq
+          spells={[TALENTS.GLACIAL_SPIKE_TALENT, TALENTS.FLURRY_TALENT, cooldown, cooldown]}
+        />{' '}
+        <br />
+        filling the {cooldownIcon}s with {this.hasGlacialSpike && <>{glacialSpikeIcon} </>},{' '}
```
emallson: the leading comma here is outside the JSX for glacial spike, but should be inside it for cases where glacial spike is not taken.

```suggestion
        filling the {cooldownIcon}s with {this.hasGlacialSpike && <>{glacialSpikeIcon},{' '}</>}
```

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
```diff
@@ -0,0 +1,157 @@
...
+        spells.push(spell);
+      }
+    });
+    const tooltip = (
+      <>
+        <PerformanceMark perf={this.getPrecastPerformance()} />
+        {this.getPrecastPerformanceMessage()} <br />
+        <PerformanceMark perf={this.getShatterPerformance()} />
+        {this.getShatterPerformanceMessage()} <br />
+        <SpellSeq spells={spells} />
```
emallson: ideally use block elements instead of `<br/>` for layout. also missing spaces between the checkmarks and the text.

(prettier may not like this suggestion as-is)

```suggestion
        <div><PerformanceMark perf={this.getPrecastPerformance()} />{' '}
        {this.getPrecastPerformanceMessage()}</div>
        <div><PerformanceMark perf={this.getShatterPerformance()} />{' '}
        {this.getShatterPerformanceMessage()} </div>
        <SpellSeq spells={spells} />
```

emallson (changes requested):
a few minor changes. the only truly important one (IMO) is about the use of the `SPELLS` table, but the others are probably worth looking at

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
earosselot: yes, I decided to leave it this way for documentation and consistency with the performance message.

src/analysis/retail/mage/frost/normalizers/CastLinkNormalizer.ts
earosselot: My bad, you're right, should be left to default. Splitting ice does not apply to frostbolt anyways.

src/analysis/retail/mage/frost/core/WintersChillEvent.tsx
earosselot: 1. as I'm working right now on single target only, flurry will be constant, but other spells could apply winters chill too. I'm changing the name to applier for a wider support and better semantics.
2. precast could not exist because the normalizer only looks for frostbolt and glacial spike in a certain amount of time before the buff apply. As for flurry (applier now) I think that the cast not existing is a defensive programming thing that was already there.

I was looking for something like `maybeGetSpell` but didn't found it, thanks for pointing it out 👍

emallson (approved):

PR #6639, opened 2024-02-03
[Evoker] Implement Source of Magic & Potent Mana modules

Krealle:
### Description
Implemented statistics modules for the shared Evoker talents `Source of Magic` and `Potent Mana`.

Percentage for `Source of Magic` mana gain is just the amount of a mana bar it has provided.

Enabled for all specs, tagging @trevorm4 since it includes Preservation.

- Test report URL: `/report/ZdMWQn9DvKr8HVJB/9-Mythic+Fyrakk+the+Blazing+-+Kill+(9:38)/Vollmer/standard/statistics`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/evoker/shared/modules/talents/PotentMana.tsx
+++ b/src/analysis/retail/evoker/shared/modules/talents/PotentMana.tsx
@@ -1,0 +1,146 @@
+import { Options } from 'parser/core/Analyzer';
+import TALENTS from 'common/TALENTS/evoker';
+import fetchWcl from 'common/fetchWclApi';
+import { WCLDamageDoneTableResponse, WCLHealing, WCLHealingTableResponse } from 'common/WCL_TYPES';
+import SourceOfMagic from './SourceOfMagic';
+import { POTENT_MANA_MULTIPLIER } from '../../constants';
+import LazyLoadStatisticBox from 'parser/ui/LazyLoadStatisticBox';
+import { SpellIcon, SpellLink } from 'interface';
+import ItemHealingDone from 'parser/ui/ItemHealingDone';
+import ItemDamageDone from 'parser/ui/ItemDamageDone';
+import { formatNumber } from 'common/format';
+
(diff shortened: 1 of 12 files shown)


trevorm4:
non wcl-query code looks good, I'll let topple or emallson approve that part. I know that we use LazyLoadStatistics for ones that rely on WCL queries in general since they're relatively expensive and not everybody cares about that stat all the time.

src/analysis/retail/evoker/shared/modules/talents/PotentMana.tsx
```diff
@@ -0,0 +1,146 @@
...
+        const filter = `
+        ((source.name='${window.player}' OR source.owner.name='${window.player}') AND
+        timestamp>=${window.start - this.owner.fight.start_time} AND 
+        timestamp<=${window.end - this.owner.fight.start_time})
+        `;
+        return filter;
+      })
+      .join(' OR ');
+
+    filter = String(filter) + 'AND effectiveHealing > 0';
```
ToppleTheNun, line 49: I think you may want an extra space before the `AND` here?

ToppleTheNun (approved):
Looks good as a whole, one very small note on the query itself.

src/analysis/retail/evoker/shared/modules/talents/PotentMana.tsx
Krealle: filter output looks like this
```
((source.name='Vìdel' OR source.owner.name='Vìdel') AND
timestamp>=0 AND 
timestamp<=578671)
AND effectiveHealing > 0
```
So shouldn't need an extra space.

emallson (approved):
queries look sane to me

emallson:
we do sometimes fire off extra queries in normal operation for a spec. the brew code does that for Purifying Brew. as long as it is relatively cheap (like constant number of queries done) and used sparingly it is fine.

trevorm4:
fixed in #6648

PR #6640, opened 2024-02-03
Adding Glacial Spike, Brain Freeze and Fingers of Frost to Frost Mage Guide

earosselot:
### Description

Adding these new modules to the frost mage guide: Glacial Spike, Brain Freeze and Fingers of Frost.
Fixed a bug where Glacial Spike where not analyzed as shattered even tough they were shattered.

With this PR I think the Guide is complete (imo), but let me know if you find anything that should be included.
If nothing else is needed I will send a new PR setting guide view to default. There we can correct the typos or grammatical issues of the complete guide as we agreed with @Sharrq in the first guide PR. 

- Screenshot(s):

image
image

### DIFF
--- a/src/analysis/retail/mage/frost/CHANGELOG.tsx
+++ b/src/analysis/retail/mage/frost/CHANGELOG.tsx
@@ -2,10 +2,11 @@ import { change, date } from 'common/changelog';
 import SPELLS from 'common/SPELLS';
 import TALENTS from 'common/TALENTS/mage';
 import { SpellLink } from 'interface';
-import { Raistlinn, Sharrq, Sref, ToppleTheNun } from 'CONTRIBUTORS';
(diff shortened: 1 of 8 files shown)


Sharrq:
The spec stuff looks fine at a glance. will need someone else to look at the code to make sure it is ok since I am unfamiliar with the guide.

emallson (approved):
code-wise this looks fine. design-wise, showing only summary statistics for brain freeze / fof is not ideal but it is better than having no info for them.

im not familiar enough with the current state of frost mage to know what the correct way to visualize them is. with the guide in general, a core idea is that we want to show individual mistakes. rather than "4% expired" there should be a way to see info about the expiration, such as the timestamp.

these procs are so frequent that i am not sure that our existing UI elements can really support showing both all the correct and all the incorrect uses of these procs, though

PR #6652, opened 2024-02-15
[Hunter] Workaround for Volcoross having weird events

Pewtro:
Volcoross is notoriously bad at having events that make no sense, e.g. https://www.warcraftlogs.com/reports/qpJ4tV8aYZ6CRGrL#fight=10&type=auras&source=45&ability=272790&view=events where Apply Buff events are going out without the buff expiring. This adds a new flag to the buff tracker labelled as experimental because it might have other adverse side effects and uses that to try and work around the weird events.

I'm not at all personally  convinced this is worth having in because it's one boss and the events are super scuffed for this boss in particular, so maybe this doesn't even make it analyzable on it's own... 

# Before:
image
image

# After 
image
image

### DIFF
--- a/src/analysis/retail/evoker/augmentation/modules/talents/BlisteringScalesStackTracker.tsx
+++ b/src/analysis/retail/evoker/augmentation/modules/talents/BlisteringScalesStackTracker.tsx
@@ -39,6 +39,11 @@ export default class BlisteringScalesStackTracker extends Analyzer {
     ? 20
     : 15;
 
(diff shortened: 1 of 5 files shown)


emallson (approved):
the code for this seems sane. i'm torn between not wanting boss-specific hacks to live forever, and also wanting our stuff to work as much as is possible. we work around a ton of weirdness already, is another edge case for blatantly busted logs that bad, really?

PR #6661, opened 2024-02-26
switch to assets.rpglogs.com for WoW Icons

emallson:
the WoW CDN is notoriously flaky with icon names. this also nukes the BAD_ICONS list, which should (hopefully) be fine.

I went through a few different specs that had exceptions in the list and checked them:

- DK (Blood)
- DH (Havoc)
- Druid (Guardian, Resto)

things generally look fine. Worth loading up on your local and double-checking, since this impacts all specs.

The one I know that is messed up is the Chi Surge icon for Brewmaster, which is also messed up on WCL (sigh)

### DIFF
--- a/scripts/talents/generate-talents.ts
+++ b/scripts/talents/generate-talents.ts
@@ -107,7 +107,7 @@ const entryToSpell = (
 ): GenericTalentInterface => ({
   id: entry.spellId!,
   name: entry.name!,
-  icon: entry.icon,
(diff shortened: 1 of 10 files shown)


ToppleTheNun (approved):
Works for all of the specs I tested with:
- Vengeance Demon Hunter
- Havoc Demon Hunter
- Retribution Paladin
- Assassination Rogue

PR #6664, opened 2024-03-02
Fix broken icon for pala/priest

avilene:
### Testing

- Screenshot(s):


emallson (changes requested):
These manual changes to the talent files will be lost next time they're regenerated. I can look at what is causing the icon names to be mismatched

PR #6669, opened 2024-03-08
fix TS2612 findings

ToppleTheNun:
### Description

Fix TS2612 findings caused by bumping the `target` field in the tsconfig.json to `ES2022` or later.

More or less, when `target` is `ES2022` or later, this setting defaults to true: https://www.typescriptlang.org/tsconfig#useDefineForClassFields

This is the behavior it causes: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier

The fix is to not redeclare the fields that are defined in the parent OR to use the `declare` modifier if the type you're using actually matters.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -34,6 +34,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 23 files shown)


ToppleTheNun:
Doesn't work while using create-react-app, closing and merging the commit into `minimal-vite`.

PR #6671, opened 2024-03-12
Update disc priest modules to be compatible with the current patch

oisinod:
### Description
Config file for disc priest updated as no major changes were added to the spec in 10.2.5


Sharrq:
👍

PR #6672, opened 2024-03-13
[Preservation] Implemented Time of Need module

Harreks:
### Description

Implemented a module to track the performance of the Time of Need talent

Lazyload because it pulls damage taken events while the Time of Need clone is up to check if it's healing saved the person that procced it or if it was irrelevant

### Testing

- Test report URL: `/report/cf8p1kZtmxvDCXqN/1-Mythic++Atal'Dazar+-+Kill+(25:42)/Harreks/standard/statistics`
- Screenshot(s):
image


src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -1062,4 +1074,19 @@ export function isEbFromReversion(
...
+  const allHealing = GetRelatedEvents<HealEvent>(event, TIME_OF_NEED_HEALING) ?? null;
+  const tonHealing: HealEvent[] = [];
+  if (allHealing) {
+    allHealing.forEach((heal) => {
+      //Only grab heals that come from the spawned entity.
+      //This is to prevent Player VE's from getting pulled cause they have the same id
+      if (heal.sourceID === event.targetID) {
+        tonHealing.push(heal);
+      }
+    });
```
emallson: rather than doing this filtering here, you can do it during normalization by specifying an `additionalCondition` above like:

```tsx
additionalCondition: (linkingEvent, referencedEvent) => (linkingEvent as SummonEvent).targetID === (referencedEvent as HealEvent).sourceID
```

(the `as <subtype>` clauses are unfortunately needed because of the `AnyEvent` parameter types on the interface).

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,241 @@
...
+          //If ToN did no healing, it bugged out
+          this.spawns[index].result = Result.Bug;
+        }
+      }
+    });
+  }
+
+  async load() {
+    //Run every ToN spawn through parseDamageTaken()
+    for (const spawn of this.spawns) {
```
emallson, line 130: how many ToN spawns would you expect in a typical M+ run? potentially worried about hitting burst limits if there are hundreds of possible spawns in a long key.

looking at [this log](https://www.warcraftlogs.com/reports/KGaV46B9LrFHhk8W#fight=1&type=healing&source=18&translate=true) i see an avg of like 1 every 2m, which is probably fine?

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,241 @@
...
+      await fetchWcl(`report/events/damage-taken/${this.owner.report.code}`, {
+        start: spawn.summon.timestamp,
+        end: spawn.summon.timestamp + 8000,
+      }).then((json) => {
```
emallson: rather than `await`ing a `then`ed promise, just await the `fetchWcl` directly and use the type parameter:

```ts
const json = await fetchWcl<WCLEventsResponse>(...);
```

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,241 @@
...
+      }
+    });
+
+    this.spawns.push(currentEvent);
+  }
+
+  parseDamageTaken(tonSummon: SummonEvent, eventList: WCLEventsResponse) {
+    //Iterate through all Time of Need spawns to find the one that matches
+    this.spawns.forEach((spawn, index) => {
+      if (spawn.summon.timestamp === tonSummon.timestamp) {
```
emallson: rather than iterating through all spawns to find the one matching the summon event, directly pass the spawn from the caller (there is only one caller and it passes `spawn.summon` so that gives you direct access).

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,241 @@
...
+                } else {
+                  //Initialize the healing from Time of Need with the VE
+                  let tonHealingAtThisPoint = spawn.verdantEmbrace?.amount || 0;
+                  //If there is no living flames casted, or this hit was before the first living flame landed
+                  if (
+                    spawn.livingFlames.length === 0 ||
+                    hit.timestamp < spawn.livingFlames[0].timestamp
+                  ) {
+                    //Check if the player would've died from this hit without the Verdant Embrace healing
+                    if (
```
emallson: rather than nesting this 7 `if`s deep, use early returns on the negated conditions to make things easier to follow. your logic is basically:

1. check that there is healing done, if not return `Bug`
2. check that there is damage taken, if not return `Not` (i don't think you currently do this?)
3. go through the damage taken events and see if the healing from ToN prevented a death, either from an individual heal (`Proc`) or from the total healing (`Long`).
    - if they died despite the healing, early return `Dead`
    - if none of the above, `Not`.

that doesn't need to be nested so deeply.

i don't know enough about the ToN talent to know if it makes sense to consider some large amount of healing that didn't directly prevent a death as relevant, i'll leave that up to you + the healer folks.

emallson (changes requested):
got a few changes, mostly code style and not correctness. will defer to @trevorm4 and @vohrr as to whether the analysis is appropriate for the talent

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
Harreks: The talent itself has a 1min cd so theoretical maximun is as many minutes as the key had, realistically i've seen between 9 and 16 spawns per key

vohrr:
@trevorm4  defer to you on this one

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    this.addEventListener(
+      Events.summon.by(SELECTED_PLAYER).spell(TALENTS_EVOKER.TIME_OF_NEED_SUMMON),
+      this.onSpawn,
+    );
+  }
+
+  onSpawn(event: SummonEvent) {
+    const HealEvents = getTimeOfNeedHealing(event);
```
trevorm4:
```suggestion
    const healEvents = getTimeOfNeedHealing(event);
```
casing

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+enum Result {
+  Proc = 'Saved by Verdant Embrace',
+  Long = 'Saved by all the healing',
+  Not = 'Was not in danger of death',
+  Dead = 'Died anyway',
+  Bug = 'Time of Need fizzled out',
+  Unparsed = "This event wasn't reviewed",
+}
+
+interface tonEvent {
```
trevorm4:
```suggestion
interface TonEvent {
```
nit: casing

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+  spawns: tonEvent[] = [];
+
+  constructor(options: Options) {
+    super(options);
+
+    this.active = this.selectedCombatant.hasTalent(TALENTS_EVOKER.TIME_OF_NEED_TALENT);
+
+    if (!this.active) {
+      return;
+    }
```
trevorm4: technically not necessary

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+    this.spawns.push(currentEvent);
+  }
+
+  parseDamageTaken(tonSpawn: tonEvent, eventList: WCLEventsResponse) {
+    //Check that ToN actually did any healing or if it bugged out
+    if (tonSpawn.verdantEmbrace === null && tonSpawn.livingFlames.length === 0) {
+      return Result.Bug;
+    }
+    //If there is no damage events, then Time of Need didn't save the player from anything
+    if (eventList.events.length === 0) {
```
trevorm4:
```suggestion
    if (!eventList.events.length) {
```

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+      return Result.Not;
+    }
+    const hits = eventList.events as DamageEvent[];
+    //Iterate all the damage taken events during the time window
+    for (const hit of hits) {
+      //Check that the info from the damage event is valid
+      if (hit.hitPoints !== undefined && hit.targetID === tonSpawn.target) {
+        //If the hitpoints after the hit are equal to 0, the player died anyway
+        if (hit.hitPoints === 0) {
```
trevorm4:
```suggestion
        if (!hit.hitPoints) {
```
not gonna do it for every case, but prefer doing `!a.property` instead of `a.property === 0`

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+import { SpellIcon, SpellLink } from 'interface';
+import { formatNumber } from 'common/format';
+import fetchWcl from 'common/fetchWclApi';
+import { WCLEventsResponse } from 'common/WCL_TYPES';
+import { Trans } from '@lingui/macro';
+import Combatants from 'parser/shared/modules/Combatants';
+
+//All possible results from a Time of Need proc
+enum Result {
+  Proc = 'Saved by Verdant Embrace',
```
trevorm4: Not sure, but if it lets you do it you should use `<>Saved by<SpellLink spell={TALENTS_EVOKER.VERDANT_EMBRACE_TALENT} /></>`

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+  statistic() {
+    return (
+      <LazyLoadStatisticBox
+        category={STATISTIC_CATEGORY.TALENTS}
+        position={STATISTIC_ORDER.OPTIONAL(60)}
+        loader={this.load.bind(this)}
+        label={
+          <Trans>
+            <SpellLink spell={TALENTS_EVOKER.TIME_OF_NEED_TALENT} /> events
+          </Trans>
```
trevorm4: I think we stopped using these for individuals modules, should be fine to remove
```suggestion
            <SpellLink spell={TALENTS_EVOKER.TIME_OF_NEED_TALENT} /> events
```

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+          <div>
+            <SpellLink spell={TALENTS_EVOKER.TIME_OF_NEED_TALENT} /> will cast one{' '}
+            <SpellLink spell={TALENTS_EVOKER.VERDANT_EMBRACE_TALENT} /> and several
+            <SpellLink spell={SPELLS.LIVING_FLAME_HEAL} />s on the player that caused the proc.
+            Depending on the amount of damage that player took during the 8 seconds
+            <SpellLink spell={TALENTS_EVOKER.TIME_OF_NEED_TALENT} /> is active the event is
+            classified as one of the following:
+            <ul>
+              <li>
+                <b>Saved by Verdant Embrace:</b> If the player would've died from damage after the{' '}
```
trevorm4:
```suggestion
                <b>Saved by <SpellLink spell={TALENTS_EVOKER.VERDANT_EMBRACE_TALENT} />:</b> If the player would've died from damage after the{' '}
```

src/analysis/retail/evoker/preservation/normalizers/CastLinkNormalizer.ts
```diff
@@ -823,6 +825,19 @@ const EVENT_LINKS: EventLink[] = [
       );
     },
   },
+  {
+    linkRelation: TIME_OF_NEED_HEALING,
+    linkingEventId: TALENTS_EVOKER.TIME_OF_NEED_SUMMON.id,
+    linkingEventType: EventType.Summon,
+    referencedEventId: [SPELLS.VERDANT_EMBRACE_HEAL.id, SPELLS.TIME_OF_NEED_LIVING_FLAME.id],
+    referencedEventType: EventType.Heal,
+    forwardBufferMs: 8000, //8 seconds, the duration of Time of Need
```
trevorm4: Put this into a constant at the top of the file similar to other links, makes it a lot easier to refactor stuff on patches.

src/common/TALENTS/evoker.ts
```diff
@@ -1324,6 +1324,14 @@ const talents = {
     entryIds: [115564],
     definitionIds: [{ id: 120576, specId: 1468 }],
   },
+  TIME_OF_NEED_SUMMON: {
```
trevorm4: Keep this in SPELLS, this file gets re-generated on patches so it will get removed automatically.

trevorm4 (changes requested):

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
Harreks: Doesn't seem to work, it just sends the whole thing as a string instead of making the SpellLink, maybe theres another way to handle the possible results?

Krealle: You could use an object instead of an enum to make it work eg.
```tsx
const Result = {
  Proc: (
    <>
      Saved by <SpellLink spell={TALENTS_EVOKER.VERDANT_EMBRACE_TALENT} />
    </>
  ),
  Long: 'Saved by all the healing',
  Not: 'Was not in danger of death',
  Dead: 'Died anyway',
  Bug: 'Time of Need fizzled out',
  Unparsed: "This event wasn't reviewed",
} as const;
//if you want a dynamic type; alternatively just "JSX.Element | string" it
type Results = typeof Result[keyof typeof Result];
```

src/analysis/retail/evoker/preservation/modules/talents/TimeOfNeed.tsx
Krealle:
```tsx
    for (const spawn of this.spawns) {
      const target = this.combatants?.players[spawn.target];
      const filter = target && target?.name ? `target.name = "${target.name}"` : '';
      const damageTakenEvents = await fetchWcl<WCLEventsResponse>(
        `report/events/damage-taken/${this.owner.report.code}`,
        {
          start: spawn.summon.timestamp,
          end: spawn.summon.timestamp + 8000,
          undefined,
          filter,
        },
      );
```
You could potentially add this filter to reduce the amount of events you query

emallson: what is the `undefined` for?

emallson: (filter is a good addition)

Krealle: the undefined param is optional `actorId` iirc, but suppose you could just do `filter: filter` to avoid the undefined

emallson: since this is an object literal, `filter` is shorthand for `filter: filter`. I don't believe you need to explicitly handle the actor id param, so omitting the undefined should be fine. If other comments are addressed I can handle that and get it merged

Harreks: All the other comments got addressed (and also the filter with the 'undefined') in my last 2 commits

PR #6674, opened 2024-03-14
Fix getBuffStacks method

earosselot:
### Description

Fix a bug where a flurry cast without precast was marked as fail when casted at 4 icicles (altough is a correct cast). 

Flurry event at 4 icicles before:
image

Flurry event at 4 icicles now:
image

[Link to log](https://www.warcraftlogs.com/reports/gLfHRVjd7ByKkxqF#fight=last&type=casts&source=2&view=events&pins=2%24Off%24%23244F4B%24expression%24ability.name%20in%20(%22Glacial%20Spike%22,%20%22Ice%20Lance%22,%20%22Flurry%22)%5E0%24Separate%24%23909049%24auras-gained%24-1%24183490188.0.0.Mage%24183490188.0.0.Mage%24true%24183490188.0.0.Mage%24true%24205473)

[Link to video](https://www.youtube.com/watch?v=qsXpG2Oe5TM&ab_channel=EduardoRosselot) showing the flurry cast on 4 icicles at @0:16

The problem was caused by the method getBuffStacks returning the last stack update of the active buff and not the stacks at the requested timestamp. I made a correction, might not be the JS way to do it :).

For example: 
image
For this buff object you will always get 5 stacks if all the stacks were processed, even if you ask for the 13seconds timestamp.


src/parser/core/Entity.ts
```diff
@@ -115,9 +115,31 @@ class Entity {
...
+    const currentTimestamp = forTimestamp !== null ? forTimestamp : this.owner.currentTimestamp;
+    let stacks = 0;
+    let stacksTimestamp = 0;
+    buff.stackHistory.forEach((stack) => {
+      if (stacksTimestamp < stack.timestamp && stack.timestamp < currentTimestamp) {
+        stacks = stack.stacks;
+        stacksTimestamp = stack.timestamp;
+      }
+    });
+    return stacks;
```
emallson: what edge case does this address beyond just calling `getBuff` with the `forTimestamp` value? is `buff.stacks` always the last-seen value of stacks, not necessarily the one at `forTimestamp`?

earosselot: > what edge case does this address beyond just calling `getBuff `with the `forTimestamp `value?

This method looks into the stackHistory array and gathers the lastest update for the given timestamp.

> is `buff.stacks` always the last-seen value of stacks, not necessarily the one at `forTimestamp`?

As far as I know, yes. I show you this debug case to make my point:
image

As you can see, the calls to `getBuff` are both returning 5 stack when they (as I understand the intended behavior of the method) should return 4 and 2 for `forTimestamp` and `forTimestamp-2000` respectively.

the code now does return these values (same debug position as above):
image

I don't know if I'm explaining good enough. Also, I think this should be tested in some clases to watch the behavior of the change.

emallson: that makes sense :+1:

src/parser/core/Entity.ts
```diff
@@ -111,13 +111,27 @@ class Entity {
...
+      forTimestamp,
+      bufferTime,
+      minimalActiveTime,
+      sourceID,
     );
+    const currentTimestamp = forTimestamp !== null ? forTimestamp : this.owner.currentTimestamp;
+
+    let maxStackForTimestamp: StackHistoryElement = { timestamp: 0, stacks: 0 };
+    buff?.stackHistory.forEach((stack) => {
+      if (maxStackForTimestamp.stacks < stack.timestamp && stack.timestamp < currentTimestamp) {
```
emallson: looks like a typo here. this should be:

```suggestion
      if (maxStackForTimestamp.timestamp < stack.timestamp && stack.timestamp < currentTimestamp) {
```

emallson (changes requested):

src/parser/core/Entity.ts
earosselot: ups

src/parser/core/Entity.ts
earosselot: I've checked on some clases that use the method and everything looks the same.

PR #6678, opened 2024-03-18
[Devastation] Update APL check module and some timeline cleanup

Krealle:
### Description

Recent changes to Deva APL had me looking over the APL check module again and came to the conclusion that it was fairly unwieldy to look at, so have tried to simplify it and write more condense descriptions for the different rules. Also put me down a little rabbit hole when trying to clean up the timeline a bit 😅 

As I was writing up this PR I found a bug/weird interaction with how `aplProcessesEvent` handles back-to-back channeled casts of the same spell, eg. Living flame cast back to back will only look at the first cast and ignore the subsequent casts, so I'll leave this as draft until I solve that.

image
The last two Living Flame casts here are ignored and therefor not rule checked, if I do the quick and dirty and just remove fabricated beginchannel events for Living Flame it counts them all, but a more general solution would be much better, since this interaction most likely extends to other spells as well.

#### Solution 

So above issue has now been resolved, turned out to be the logic in `aplProcessesEvent` didn't work too well with non-instant cast abilities that were cast back-to-back. I've now added some extra checks to ensure we properly detect chain casts.
An example of how it looks can be seen here:
image
image
image

For demonstration purposes I've set all the casts there are currently detected and rulechecked as inefficient(red border) and the ones the new conditions catch(previously ignored) as highligthed (white border).

I've tested this change on a few different logs for all specs that have AplCheck module implemented and the new logic doesn't catch anything funky there (which makes sense since only melee specs have it implemented, and everything is caught properly already there).

This issue might be rooted in inconsistent normalization of different spells, eg. `Disintegrate` when chain cast never actually generates an `endchannel` event. I'll look a bit more into this later to see if I can figure out a general solution to this.

### Deva / Evo
Reworked how rules are being made, now we generate them in a different file where we can pass talent information along to more easily account for talent conditionals in rules.

Added `range` values to `Evoker` rotational spells to fix issues with `rangeCheck` for APL check.
This was actually causing a lot of issues, think the distance calculation might be a bit off. On Volcoross logs it would output a distance to boss of 80 yards when the player was pewpewing the boss (Evoker has 25 yard range).
Can be observed on this log: `/report/awxFTf9PdDRGVArY/85-LFR+Volcoross+-+Kill+(3:07)/Enchulao/standard`

Cleaned up `Buffs`.
Added `Living Flame` to `EssenceBurstNormalizer` to fix event ordering issue, caused issues with resource checks when it was cast with Burnout active.

**Moving this to another pr**
~~Implemented `LivingFlamePrePullNormalizer` to fix an edge case that `PrePullCooldowns` doesn't cover.~~

### Core
Implemented a new `buffPresentSoon` conditional, to help figure out when a buff will next be present.
Added `fallback` param to `debuffMissing` conditional so you can override the fallback return.
Added `initial` param to `hasResources` conditional so you can define expected intial resources.
Fixed `APLCheck` skipping chain casted abilities, as mentioned earlier.

**Moving these to another pr**
~~Fixed the `damageSpellIds` for `Nymue`, so `PrePullCooldowns` can properly create prepull cast event.~~
~~Added `Nymue` to `Channeling` normalizer and added `PrePullCooldowns` as a dependency so it will normalize the fabricated prepull events too.~~

### Testing

- Test report URL: `/report/BnvQVMajTym1FhgX/12-Mythic+Smolderon+-+Kill+(5:43)/Blueprínt/standard/overview`
- Screenshot(s):

### example 1: (Gnarlroot)
before:
image
after:
image

### example 2: (Smoldy Boi)
before:
image
after:
image


Krealle:
Moving the pre-pull `Living Flame` normalization and `Nymue` / `Channeling` changes to another PR to reduce the scope of this one.

emallson:
i haven't gotten around to a full review of this yet (sorry!) but wanted to chime in:

> This issue might be rooted in inconsistent normalization of different spells, eg. Disintegrate when chain cast never actually generates an endchannel event. I'll look a bit more into this later to see if I can figure out a general solution to this.

some of this stems from being in functional-analyzer-land where we can't call other class-based analyzers to do things like check cooldown status or buff presence. some of the quirkiness of the apl checker will go away (i fucking hope) when it gets moved to an event-driven, class-based analyzer

Krealle:
No worries with the wait, the PR did get a big bigger than I initally thought it would be 😅 
Yea, making it a class-based analyzer would indeed be sweet!

src/parser/shared/metrics/apl/conditions/buffSoonPresent.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+type BuffInformation = {
+  buffs: TrackedBuffEvent[];
+  nextBuff: number | undefined;
+  currentIndex: number;
+};
+
+type SourceOptions = 'player' | 'external' | 'any';
+
+export function buffSoonPresent(
```
emallson, line 22: this is a bit of a scary APL condition---not because it is inherently harmful, but because it lets people write APL rules that are not actually reflective of in-game play. i'd recommend a docblock for this explaining that:

```suggestion
/**
 * Check if a buff will be present soon after the present time.
 *
 * This should ONLY be used for buffs that the player has complete control over, such as buffs tied to cooldown use.
 * DO NOT use this to "look ahead" for random procs to implement impossibly optimal APLs.
 */
export function buffSoonPresent(
```

src/parser/shared/metrics/apl/conditions/debuffMissing.tsx
```diff
@@ -73,7 +76,7 @@ export function debuffMissing(
       const targets = getTargets(event, targetOptions?.targetLinkRelation);
       if (targets.length === 0) {
         //No target so we can't check for a debuff
-        return false;
+        return fallback;
```
emallson, line 79: hmm, i wonder if `fallback = true` should be the default. that would be a breaking change so *not now* but...maybe

emallson (approved):
approved, but i would like that docblock in (which will probably require a local prettier run to fix formatting)

src/parser/shared/metrics/apl/conditions/buffSoonPresent.tsx
Krealle: Added the docblock - didn't even consider that usecase but makes sense to discourage it.

PR #6680, opened 2024-03-19
Bring BDK to 10.2.5

avilene:
### Description

- Fixed Death and Decay issue
- Updated guide to have other sections
Still missing APL but that is a WIP as well :D 

### Testing

- Test report URL: `/report/y3D2KWrZtCd1AkVY/29-Heroic+Volcoross+-+Kill+(2:47)/Arlie/standard`
- Screenshot(s):


emallson:
:eyes: i have some in-progress bdk stuff that has a bit more ambitious scope that i'll have to merge with this

src/analysis/retail/deathknight/blood/modules/Abilities.ts
```diff
@@ -280,6 +281,7 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
+        enabled: !combatant.hasTalent(TALENTS.RAPID_DECOMPOSITION_TALENT),
```
emallson: i believe this is incorrect. D&D is always available for DK afaik, and rapid decomp doesn't remove it.

src/analysis/retail/deathknight/blood/modules/core/RuneForgeChecker.tsx
```diff
@@ -35,11 +35,11 @@ class BloodRuneForgeChecker extends RuneForgeChecker {
...
+        importance: SUGGESTION_IMPORTANCE.MINOR,
         suggestion: (
           <Trans id="deathknight.blood.runeforgeSuggestion.sanguination">
-            Don't use <SpellLink spell={SPELLS.RUNE_OF_SANGUINATION} /> as Blood Death Knight, use{' '}
-            <SpellLink spell={SPELLS.RUNE_OF_THE_FALLEN_CRUSADER} /> instead.
+            Only use <SpellLink spell={SPELLS.RUNE_OF_SANGUINATION} /> as Blood Death Knight on a
+            pure ST fight, use <SpellLink spell={SPELLS.RUNE_OF_THE_FALLEN_CRUSADER} /> instead.
           </Trans>
         ),
       },
```
emallson: this is probably worth outright removing, or somehow gating behind a M+ check. sanguination is currently worth using on every raid fight when played well. it isn't recommended by default on wowhead in part because the execution requirement is much higher, but even on a fight with heavy AoE like Tindral most of the top logs use Sanguination.

we don't need to recommend using Sanguination (again: much higher execution requirement than Fallen Crusader), but shouldn't recommend removing it except in M+

src/analysis/retail/deathknight/blood/modules/guide/index.tsx
```diff
@@ -1,11 +1,59 @@
...
+      isActive: (c) => c.hasTalent(TALENTS_DEATH_KNIGHT.ICEBOUND_FORTITUDE_TALENT),
+    },
+    {
+      spell: TALENTS_DEATH_KNIGHT.VAMPIRIC_BLOOD_TALENT,
+      isActive: (c) => c.hasTalent(TALENTS_DEATH_KNIGHT.VAMPIRIC_BLOOD_TALENT),
+    },
+    {
+      spell: TALENTS_DEATH_KNIGHT.ANTI_MAGIC_SHELL_TALENT,
+      isActive: (c) => c.hasTalent(TALENTS_DEATH_KNIGHT.ANTI_MAGIC_SHELL_TALENT),
+    },
```
emallson, line 42: these should *probably* go in a major defensive setup instead, but that is a lot more code so i don't think it needs to happen before merging this. the cooldown section is a good first step.

emallson (changes requested):
the only thing that really, *really* needs to change for a merge is the incorrect D&D disable. however, we should probably change the suggestion to no longer recommend against Rune of Sanguination

src/analysis/retail/deathknight/blood/modules/Abilities.ts
avilene: This is because of the object in line 261 that has a different cast efficiency.

src/analysis/retail/deathknight/blood/modules/core/RuneForgeChecker.tsx
avilene: True, how do we check m+ vs raid in these things though?

src/analysis/retail/deathknight/blood/modules/Abilities.ts
emallson: ahhh, I missed that. Rather than duplicate the object, you could make the cast efficiency conditional on the talent (I think). Something like:


```
castEfficiency: hasTalent(...) ? { ... } : undefined;
```

src/analysis/retail/deathknight/blood/modules/core/RuneForgeChecker.tsx
emallson: We have a helper for this, but I can't remember it offhand. I'll dig it up

emallson: I think this would be calling https://github.com/WoWAnalyzer/WoWAnalyzer/blob/14e371d4bb3650cbd6f333760f2f12b1762f74b5/src/common/isMythicPlus.ts#L4 on `this.owner.fight`

EDIT: or `options.owner.fight` since i think this is in the constructor

avilene: Aight ill fix :3

src/analysis/retail/deathknight/blood/modules/Abilities.ts
avilene: will fix!

emallson (approved):

PR #6681, opened 2024-03-21
[Core] Potential solution to casts and channels on timeline overlapping eachother

Krealle:
### Description

Currently the timeline looks like this:
image
image

Currently `BeginChannel` events are being rendered, and then the `Cast` events are rendered ontop of them, leaving to lots of bloat and an overabundance of ablility inserts which can make it difficult to discern their importance/meaning. 

When I took a look at how the rendering is being handled in the `Casts` module, I noticed that for the `renderCast` function there is an immediate check for whether or not the `CastEvent` has a `channel` prop, cancelling the render if it does.

Looking at what this `channel` prop is actually used for I found out that it's not actually being used for anything currently (besides the one check in `renderCast`.
And when looking at the `Channeling` normalizer, we already have `BeginChannelEvent` and `EndChannelEvent` events being fabricated.

### Proposal 

So my proposal to this, would be to change the `channel` prop type in `BaseCastEvent` to `EndChannelEvent` and attach the `EndChannelEvent` to the associated `CastEvent` when we fabricate it, then the `Casts` module can properly ignore overlapping renders, and the `channel` prop could be used for potential analysis.

Example of how it would look with above implementation:
image
image

With this implementation, channeled casts should now only ever have their channel events rendered and their cast events ignored. Leaving only off-GCD casts being insterted.

### Empowers are dumb
One note to this, it won't currently work with Empowers, atleast not in their current state (their behaviour out of the box is kinda too rip bozo and should be normalized). 
But have an idea for a normalizer that could adress this and make Empower events more streamlined, such that the `empowerChannelSpec` could be reworked to be less hacky and could properly attach itself to the cast events from Empowers.

Untitled

### Testing

- Test report URL: `/report/BNxf9VWHzjtyrDYQ/12-Mythic+Igira+the+Cruel+-+Kill+(4:10)/Mcbaguette/standard/timeline`


Sharrq:
Looks good to me, Fire Mage looks a lot cleaner

emallson (approved):
sharrq approved it and the code looks sane so I'm happy enough merging this to improve the timeline (and we'll see what fallout comes of it)

PR #6687, opened 2024-03-26
add Dragonflight season 4 dungeons

ToppleTheNun:
### Description

Add Dragonflight season 4 dungeons.


emallson:
only comment is that rather than duplicating the images we could import the images from their original seasons. otherwise seems fine

ToppleTheNun:
That's less duplicative, yeah, will change that.

PR #6693, opened 2024-03-30
[Evoker] Implement Living Flame prepull normalizer

Krealle:
### Description

Implementing a prepull normalizer for `Living Flame`, to deal with an edgecase that prevents `PrePullCooldowns` normalizer from fabricating prepull `BeginCast` event.
Tagging @trevorm4 since I implemented it for Preservation.

Also noticed that the export for `LivingFlameNormalizer` was being exported from `LeapingFlamesNormalizer` so have fixed that, and enabled that normalizer for Devastation and Augmentation.

### Testing

- Test report URL: `/report/BNxf9VWHzjtyrDYQ/8-Mythic+Gnarlroot+-+Kill+(2:43)/Vollmer/standard/timeline`
- Screenshot(s):
before:
image
after:
image


src/analysis/retail/evoker/shared/modules/normalizers/LivingFlamePrePullNormalizer.ts
```diff
@@ -0,0 +1,86 @@
...
+class LivingFlamePrePullNormalizer extends EventsNormalizer {
+  static dependencies = {
+    ...EventsNormalizer.dependencies,
+    haste: Haste,
+  };
+  protected haste!: Haste;
+  normalize(events: AnyEvent[]) {
+    const fixedEvents: AnyEvent[] = events;
+
+    fixedEvents.some((event, eventIdx) => {
```
emallson: use of `some` here is awkward. i understand why but :thinking:

emallson (approved):
tentative :+1:. not sure how others feel about the use of `.some(...)` for this (or if it is an established pattern that I'm just not familiar with)

src/analysis/retail/evoker/shared/modules/normalizers/LivingFlamePrePullNormalizer.ts
Krealle: yea looking at it again this should just have been a `for const`, I'll change it 👍

PR #6666, opened 2024-03-04
Adding defensive usage guide to demo warlock

ZyerTCoder:
### Description

Added a defensive guide section to the shared warlock modules
Added Dark Pact
Added Unending Resolve
Added damage taken timeline and casts breakdown
Dark Pact analyzer remakes the AllCooldownUsageList component

Added extra exports to AllCooldownUsagesList
Added subsection title to AllCooldownUsagesList (still shows up if you hide the explanation)

### Testing
[smolderon](http://localhost:3000/report/myndpGBw3k7rjVDf/4-Mythic+Smolderon+-+Kill+(7:01)/Yumiblood/standard/overview)
[m+](http://localhost:3000/report/g6Zm8BWhjkGA7MVQ/7-Mythic++Atal'Dazar+-+Kill+(25:28)/Fuitclaudit/standard/overview)
[fyrakk](http://localhost:3000/report/L8c23DzkWHvgtbKB/33-Mythic+Fyrakk+the+Blazing+-+Wipe+31+(8:31)/Yumiblood/standard)
[destro](http://localhost:3000/report/kQbaNq9XDZLFyTCj/31-Mythic+Smolderon+-+Kill+(5:50)/Splatlack/standard/statistics)
[aff](http://localhost:3000/report/TkAwcG2dF6hmBvCx/27-Mythic+Smolderon+-+Kill+(6:39)/Slayroth/standard/overview)

#### Other AllCooldownUsagesList users
[prot pala](http://localhost:3000/report/kKVw18ZnNxCTpDtR/33-Mythic+Fyrakk+the+Blazing+-+Wipe+33+(7:00)/Thatscr%C3%A4zy/standard)
[vdh](http://localhost:3000/report/dFzCLcjyqMbX43KP/20-Mythic+Volcoross+-+Kill+(3:11)/Toppledh/standard)
[guardian](http://localhost:3000/report/VWt9ZTmxfhQyLw14/9/11)
[brewmaster](http://localhost:3000/report/mLcvBXz6YVwTZ2QG/24-Mythic+Volcoross+-+Kill+(4:43)/Cobaro/standard)

image

PR #6694, opened 2024-04-01
[Evoker] Adding MajorDef analysis

Krealle:
### Description

Got inspired by #6666, so added a shared MajorDef guide section for Evoker.

- Added MajorDef guide section
- Added `Obsidian Scales`
- Added `Renewing Blaze`
- Added `Twin Guardian`
- Implemented a `DefensiveNormalizer` to fix an issue with defensives randomly getting removed/re-applied during their duration
- Enabled these for `Augmentation` and `Devastation`

I've held off on making an analyzer for `Zephyr` due to it only reducing AoE events which would need to be manually whitelisted, *and* it relies on external events to show true values (you could ofcourse just extrapolate the external mitigation based on your own, but that's not ideal).

The `Renewing Blaze` performance check is a bit in the air about whether or not the values are properly set, been working with @ljosberinn to try and figure them out; ~He'll be using/testing it a good amount in his quest for R1 heal parses so I've made it show up at the top for him so he can more easily test it out in practice and provide feedback.~

### Testing

- Test report URL: `/report/wXGcj8gNYyfQrvVT/34-Mythic+Fyrakk+the+Blazing+-+Wipe+25+(9:07)/Xephyris/standard/overview`
- Screenshot(s):
image
image
image


src/analysis/retail/evoker/shared/modules/MajorDefensives/TwinGuardian.tsx
```diff
@@ -0,0 +1,275 @@
...
+import { QualitativePerformance } from 'parser/ui/QualitativePerformance';
+import { ReactNode } from 'react';
+import Statistic from 'parser/ui/Statistic';
+import BoringValue from 'parser/ui/BoringValueText';
+import { formatNumber } from 'common/format';
+import { getTwinGuardianPartner } from '../normalizers/DefensiveCastLinkNormalizer';
+
+/**
+ * Twin Guardian is a talent that is tied to Rescue.
+ * When you use Rescue it will apply an absorb shield to both players for 30% of the Evokers current HP.
```
ljosberinn: max* hp. current would be sad

ljosberinn:
looks good to me overall, probably can remove the special casing for me however 🤣

PR #6702, opened 2024-04-05
[Shadow] Fix Error in Mind Spike & Mind Flay Insanity

DoxAshe:
### Description

In 10.2, Mind Spike & Mind Flay Insanity changed to gain procs from every other cast of Devouring Plague instead of every cast.  Since this buff does not have a refreshBuffEvent, this caused mistakes in the way overwriting procs were calculated. This update fixes the error in counting the wasted procs.


emallson (approved):
what a weird talent. assuming the e2e test issue is an intermittent infrastructure issue

PR #6703, opened 2024-04-06
[Evoker] Empower Normalization

Krealle:
### Description

As previously mentioned in  #6681 Empower behaviour has some inconsistencies:

- When you start channeling an empower you get a `EmpowerStart` & `Cast` event. Then if you finish the channel you will get a `EmpowerEnd` event. This last event isn't triggered when you cancel the channel. 
 If you instant cast the Empower with `Tip the Scales`, you only get a `Cast` event.
This provides a few challenges for analysis: 
  - Determining which event to base your analyzers on
  - Trying to determine whether a cast actually finished
  - Or when trying to determine the empowerment level
- Empowers produce different GCDs:
  - Base `1500ms` GCD when the spell gets channeled to completetion.
  - `500ms` GCD when instant cast with `Tip the Scales`
  - And a short lockout when the spell is cancelled.
- Empowers start their cooldown as soon as you start channeling, but if you cancel the channel this cooldown is reset.

**Implemented the following:**
- `EmpowerNormalizer` that produces the missing `EmpowerEnd` events from instant cast Empowers.
  - We also create `CastLinks` between the `Cast` & `EmpowerEnd` (as well as `Tip the Scales`) events, which we can use to determine if a cast was succesfully finished, or instant cast.
- `SpellUsable` normalizer to prevent starting CD on cancelled channels.
- `GlobalCooldown` normalizer to fix GCD handling.

**Along with the following changes:**
- `Channel` normalizer:
  - `empowerChannelSpec`  has been updated to work with the updated empower behaviour.
  - Removed `beginEmpowerCurrentChannel`.
  - Removed empower conditional from `attachChannelToCast`.
  - Added `Dream Breath` & `Spiritbloom` to `empowerChannelSpec`.
- Updated the `gcd` props for Empowers in `Abilities` (All specs+shared)
- Preservation: Removed `EMPOWERED_CAST` `CastLink` is now in new `EmpowerNormalizer`
- Augmentation:
  - Removed `EmpowerNormalizer`
  - Removed `TIP_THE_SCALES_CONSUME`  `CastLink` from Augmentation, is now in new `EmpowerNormalizer`

In general these changes should help with making analysis a bit more consistent, along with cleaning up the timeline.

@trevorm4 Based on a quick look through Preservations modules this shouldn't cause any issues with the current analysis. But let me know if you spot anything. 😄 

### Testing

- Test report URL: 
  - Deva: `/report/BNxf9VWHzjtyrDYQ/12-Mythic+Igira+the+Cruel+-+Kill+(4:10)/Vollmer/standard/timeline`
  - Pres: `/report/HXwny3JLV6rMYbDf/9-Mythic+Igira+the+Cruel+-+Kill+(3:44)/Zyldra/standard/timeline`
  - Aug: `/report/wXGcj8gNYyfQrvVT/34-Mythic+Fyrakk+the+Blazing+-+Wipe+25+(9:07)/Xephyris/standard/timeline`
- Screenshot(s):

### Current
image
image
image
image
### New
image
image
image
image


emallson (approved):
this looks like a nice improvement. my only real nit is that i'm not sure we need to fall back to scanning for the `empowerend` if the link normalizer failed to find it. however, i don't feel strongly either way.

i'll let the other evoker maintainers check this before hitting the merge button

Krealle:
> my only real nit is that i'm not sure we need to fall back to scanning for the `empowerend` if the link normalizer failed to find it. however, i don't feel strongly either way.

It could most likely be removed yea, during testing i didnt manage find a log where it ever actually triggered. (The only example that comes to mind is you somehow having low enough haste to have a 6s+ channel - going beyond our castlink buffer😅)

Krealle:
Okay so after playing with it some more, I think it's safe to completely remove the lookahead. I can reduce it down to this and it still behaves the same across all the logs I'm looking at.

The lookahead might be needed if `cancelCurrentChannel` begins to push a cancel channel event, but that seems unlikely?
```ts
export function empowerChannelSpec(spellId: number): ChannelSpec {
  const guids = [spellId];
  const handler: ChannelHandler = (
    event: AnyEvent,
    _events: AnyEvent[],
    _eventIndex: number,
    state: ChannelState,
  ) => {
    // Tipped cast is instant cast so don't start a channel
    if (event.type === EventType.Cast && !isFromTipTheScales(event)) {
      // do standard start channel stuff
      cancelCurrentChannel(event, state);
      beginCurrentChannel(event, state);

      const potentiallyEmpowerEnd = getEmpowerEndEvent(event);
      if (potentiallyEmpowerEnd) {
        // Empower finished channeling so we end the channel
        endCurrentChannel(potentiallyEmpowerEnd, state);
      } else {
        // Empower didn't finish channeling so we cancel the channel
        cancelCurrentChannel(event, state);
      }
    }
  };
  return {
    handler,
    guids,
  };
}
```

emallson (approved):

PR #6716, opened 2024-04-10
update paths used by Sentry for error reporting

ToppleTheNun:
### Description

Fix paths used by Sentry as our JS files now live at URLs like https://wowanalyzer.com/assets/index-6D59v4AH.js.


MartijnHols (approved):
LGTM. Just thought I'd mention vite also has https://www.npmjs.com/package/@sentry/vite-plugin, which might be nice to further remove custom solutions

ToppleTheNun:
Yep, which we are using, but probably not to its full potential atm

PR #6719, opened 2024-04-11
[Evoker] Leaping Flames module rework and Essence Burst normalization

Krealle:
### Preface

So just to preface, this got a bit bigger than intended due to needing to improve the way `Essence Burst` generation sources are dealt with, without it the `Leaping Flames` analyzer wouldn't really have gotten much more accurate. This does however have the added benefit of setting up a nice framework for dealing with `Essence Burst` attribution across multiple analyzers.
There are a few more details about the specifics of this implementation [here](https://github.com/Krealle/WoWAnalyzer/pull/3).

### Description

So the changes made to the `Leaping Flames` module/normalizer aims to increase the accuracy of the both the EB generation tracking, as well as the damage/heal hits tracking.
The previous way that the `cast links` were dealth with were essentially: "overattribute and fix in post". Which meant we had an overreliance for extrapolation. This didn't play overly nice with other analyzers such as `Pupil of Alexstraza` for Augmentation, since both would claim eachothers hits as their own.
The previous way EB generation attribution was dealth with was along the same line not as deterministic and you'd like it to be.

So the following have been changed:
- Rework the links in `LeapingFlamesNormalizer` to prevent overattribution, and to increase accuracy of the links being placed.
  - Mainly in this is the implementation of target checks and buff stack count for Leaping Flames hits.
- Rework `LeapingFlames` analyzer to function in a more deterministic manner, and use a better probability calculation method to produce more accurate distribution of EB generation.
  - Added some more stats for consumed/wasted `Leaping Flames` buffs.
- Gone through the dependencies of various normalizers to ensure proper run order. eg. Preservation was relying on links being made in `LeapingFlamesNormalizer`, but due to unfortunate run ordering, Preservations normalizer was actually running before `LeapingFlamesNormalizer`.

And just to include the `Essence Burst` part:
- Implementation of `EssenceBurstCastLinkNormalizer` which purpose is to help streamline the process of attributing EB generation sources, across all analyzers/normalizers.
- Moved and renamed `EBRefreshNormalizer` from Preservation to shared `EssenceBurstRefreshNormalizer`.
  - This also get a small rework due to it removing unwanted events when the ordering of events wasn't as expected. eg. It removed one of my `Leaping Flames` events 😅 

### Testing

- Test report URL: 
[Devastation](/report/naqCD18rgjK6GRcP/27-Mythic++Suramar+-+Kill+(31:55)/Vollmer/standard/statistics)
before
image
after
image
[Preservation](/report/p2ZQyKVf4jPDYbt7/27-Mythic+Fyrakk+the+Blazing+-+Kill+(9:22)/Antiflexx/standard/statistics)
before
image
after
image
[Augmentation](/report/wXGcj8gNYyfQrvVT/34-Mythic+Fyrakk+the+Blazing+-+Wipe+25+(9:07)/Xephyris/standard/statistics)
before
image
after
image


ljosberinn:
I'm a Krealle Yes-Man. I see a PR from Krealle, I say yes.

Jokes asides, can't say much about the implementation due to it being deep into wowa internals/specifics but I was blessed with updates during development and those made sense so I'm assuming so does the PR.

emallson (commented):
the code largely looks sane but frankly i don't know that i know enough about what is going on here to judge this. @trevorm4 and other evoker maintainers may need to weigh in

Krealle:
that is fair, I should probably mention specifically what has been changed for Preservation. I have kept it as minimal as possible. 

The only changes that affect them is: 
- `EssenceBurstRefreshNormalizer` normalizer being reworked
- Some tweaking to the linking of `Living Flame`/`Leaping Flames` cast/hits
  - `LeapingFlamesNormalizer` links were previously made after their own `CastLinkNormalizer` links (due to run/priority order), which meant that their `T31_2PC` links (the only links dependent on the links changed) wasn't able to execute the `additionalCondition` properly without `LeapingFlamesNormalizer` dependency - but since they are getting T30 back, it is a bit of a moot change really
- Dependencies to `EssenceBurstRefreshNormalizer` & `LeapingFlamesNormalizer` added for their `CastLinkNormalizer`

I haven't touched any of their specific EB links. (They weren't relying on the EB links that were changed)

Otherwise it's just the `Leaping Flames` module itself that is changed.

emallson:
ill do a merge today and get this in. no complaints from other maintainers so we'll ship it and see what happens 😅

trevorm4:
overall lgtm, i think there might be some duplicated logic for pres and this normalizer but i dont feel like fixing it so will probably keep as is for a while

PR #6723, opened 2024-04-12
Chore/knip cleanup

Pewtro:
### Description

Use https://github.com/webpro/knip to detect dead/duplicated code. 
Removing unnecessary exports should generally improve typescript performance.


emallson:
skimmed this, seems good :+1: 

there's a small conflict with a recent devastation pr that will be a bit of a pain to handle but tbh the answer might just be to use `--theirs` for that one file and take everything else.

PR #6729, opened 2024-04-18
[Shadow] Update ShadowyApparitions and add statistics for the talents Tormented Spirits and Phantasmal Pathogen

DoxAshe:
### Description

This adds statistics for shadow's talents of Tormented Spirits and Phantasmal Pathogen.  This also organizes all of shadow's talents that affect Shadowy Apparitions into a single Statistics Box so it is easier to see them all together.  The season 4 tier set interacts with the Apparitions, so having accurate numbers for these talents will be very useful.  

### Testing

I believe that the damage calculations for Phantasmal Pathogen is correct, as it is a similar effect to shadow's other talent Insidious Ire, and the value it gives is about what is expected.  

- Test report URL: `/ report/FPvzaLGAjqJ8W1M7/2-Heroic+Kazzara,+the+Hellforged+-+Kill+(3:02)/Agonyhand/standard/statistics'

- Screenshot(s):

Capture


src/analysis/retail/priest/shadow/modules/spells/ShadowyApparitions.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <BoringSpellValueText spell={TALENTS.SHADOWY_APPARITIONS_TALENT}>
+          <div>
+            <ItemDamageDone amount={this.damage} />
+          </div>
+        </BoringSpellValueText>
+
+        {this.selectedCombatant.hasTalent(TALENTS.AUSPICIOUS_SPIRITS_TALENT) ? (
+          this.auspiciousSpirits.subStatistic()
+        ) : (
+          <div />
```
emallson: if you want to render nothing for this case, then `null` works rather than `<div/>`. (same for the other two)

src/analysis/retail/priest/shadow/modules/talents/PhantasmalPathogen.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const [pathogen, unpathogen] = damageInstances.reduce<[DamageEvent[], DamageEvent[]]>(
+      (result, damage) => {
+        const enemy = this.enemies.getEntity(damage);
+        const wasDP = enemy && enemy.hasBuff(TALENTS.DEVOURING_PLAGUE_TALENT.id, damage.timestamp);
+        result[wasDP ? 0 : 1].push(damage);
+        return result;
+      },
+      [[], []],
+    );
```
emallson: this is going to be quadratic complexity and is very likely to perform super bad in M+.

rather than using `getEvents` + reducing w/ `hasBuff(x, timestamp)`, use event listeners to collect the damage + buff status (using `hasBuff` without a specific timestamp is plenty fast if you just want to call it inside a shadowy apparition damage event handler).

emallson (changes requested):
code mostly looks fine. comment on Phantasmal Pathogen to address.

have to be careful using `EventHistory`, since it can explode your runtime if you're not careful. this case is simple enough to move to event handlers.

src/analysis/retail/priest/shadow/modules/spells/ShadowyApparitions.tsx
DoxAshe: That does work better, thanks

src/analysis/retail/priest/shadow/modules/talents/PhantasmalPathogen.tsx
DoxAshe: That is good to know.  I have changed it to use event listeners instead of eventHistory.

emallson (approved):

PR #6734, opened 2024-04-22
[Chore] Remove unused exports detected by Knip

Pewtro:
### Description

v2 of https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6723, this time removing a lot of unused exports so they're only available in the file where they're used. If not used at all they got removed afterwards.


emallson:
my only concern with this is the removal of ability whitelists, e.g. the WW mastery whitelist and Ret Wings whitelist. however, if these are currently unused they may well have bitrotted to the point of being wrong

Pewtro:
I had similar concerns, so I left some unused exports that I deemed most.. preserve-worthy? example, certain events/event types that are currently unused.

emallson:
yea. removing 2.2k lines of code is certainly attractive, though. maybe we try this and see what the fallout is 🤔

PR #6735, opened 2024-04-22
Improve performance box row CSS

emallson:
### Description

This improves the performance box row CSS colors, getting rid of the clashing yellow in favor of fading out inactive elements.

(as part of this i toyed around with making the colors more accessible for rg-colorblind folks but that requires much larger changes so I did not include any of it. this is not great in that regard, but it isn't really worse than what we currently have)

### Production

image
image

### This PR

image
image


vohrr (approved):
lgtm

PR #6736, opened 2024-04-23
[Frost] Refresh for S4

Khazakdk:
### Description

Fixes a couple minor issues and improves the Killing Machine tracking. Removed some Ability cooldowns so they don't show potentially confusing cast efficiency metrics. Fixed a spell icon that had `_` where `-` should be. If it happens again may need to look into a way to persist it across regens.
### Testing

- Test report URL: `/report/ydXzMVT8H3cQfGh2/13-Mythic+Volcoross+-+Kill+(3:22)/Khazak/standard`
- Screenshot(s):


src/common/TALENTS/deathknight.ts
```diff
@@ -245,7 +245,7 @@ const talents = {
   CHILL_STREAK_TALENT: {
     id: 305392,
     name: 'Chill Streak',
-    icon: 'spell_frost_piercing_chill',
+    icon: 'spell_frost_piercing-chill',
```
emallson, line 248: can add this to the `BAD_ICONS` object to make it persist across regens: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/b67a4cbb5d48d98450accd2dbf6e9e34a3c214c4/src/interface/BAD_ICONS.ts#L11-L16

PR #6738, opened 2024-04-24
Support DF S3 and S4 tier sets for enhancement

Seriousnes:
### Description

* Adding support for S4 tier set
* Earth Shield mana cost was wrong
* Frost Mage changelog was using wrong format

### Testing

- Test report URL: `/report/mjypMY7QtC6TLP1w/17-Mythic+Fyrakk+the+Blazing+-+Kill+(9:26)/2-Seriousnes/standard`

PR #6739, opened 2024-04-24
bump for season 4 start

ToppleTheNun:
### Description

Includes:
- bump patch compatibility for all specs that had 10.2.6 as their current spec
- add missing headshots for VotI
- marks 10.2.6 as no longer current
- marks current patch as 10.2.7
- updates e2e test log

### Testing

- Test report URL: `/report/akZLCTYbN2XpQFmg`
- Screenshot(s):
image


emallson:
note that current patch is *not* 10.2.7. that releases in a couple weeks (i made the same mistake in SL)

ToppleTheNun:
Oops you are right, I will fix this in the morning

e2e/report-selection.spec.ts
Seriousnes, line 1: If it's just the `reportCode` and `fightCode` being updated throughout, maybe make add constants for it that only need to be changed in one spot.

E.g. use these and reference them in tests
```
const reportCode = 'bjqrZRnNdvKPXt13';
const fightCode = '34-Mythic+Fyrakk+the+Blazing+-+Kill+(8:47)';
```

src/analysis/retail/shaman/enhancement/CONFIG.tsx
Seriousnes, line 1: I'm handling this one in #6738

PR #6744, opened 2024-04-26
[Shadow] Add Statistics for Psychic Link

DoxAshe:
### Description

This adds a statistics for Shadow's Psychic Link Talent.  Psychic Link deals damage to all targets with vampiric touch when shadow casts certain single target abilities.  In the combat log, the cause of the psychic link damage isn't shown, so this statistic is needed to show the contribution of each single target spell to that Psychic Link damage.

### Testing

I tried to do this using a Normalizer, as this seems like a good use for one.  However, I wasn't able to get it to work correctly.  If there is something I missed with the normalizer I can try to switch back to using one.

- Test report URL: `/report/mjQ1rAaN79bvRHgk/1-Mythic++Uldaman+-+Kill+(24:59)/Doxashe/standard/statistics`
- Screenshot(s):

PsychicLink


emallson (approved):
`EventLinkNormalizer` does seem good for this but also...Psychic Link is just tied to every spell so it might be easier to just do the `Analyzer` like you did rather in practice. If you don't need the links outside of this module, I'd say this is perfectly reasonable

PR #6747, opened 2024-04-30
[Shadow]  Add statistics for several Shadow Talents.

DoxAshe:
### Description

This adds statistics for shadow's talents of Void Touched, Mind's Eye, Distorted Reality, Maddening Touch, Mind Melt, Mastermind, and Dark Evangelism.  These are simple talents that add extra damage or extra critical chance to shadow's other abilities.  Even though they are simple, most of their value isn't clear by only looking at logs, so these should all be useful.
I also updated Psychic Link to correctly work for all of Devouring Plague's damage.

### Testing

Sorry there are several talents all at once, if it would be better I can break up this up into several smaller pull requests.
 
- Screenshot(s):
VoidTouched
Mind's Eye
DistortedReality
Maddening Touch
MindMelt
mastermind
Dark Evangelism


src/analysis/retail/priest/shadow/modules/talents/DistortedReality.tsx
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  onDevouringPlagueDamage(event: DamageEvent) {
+    this.damage += calculateEffectiveDamage(event, this.multiplierDistortedReality);
+  }
+
+  onDevouringPlagueCast(event: CastEvent) {
+    const resource = event.classResources?.at(0)?.cost; //Some buffs grant free Devouring Plagues, which do not have a resource cost
+    if (resource != null) {
```
emallson: so this technically works because it is using loose equality (aka `==` instead of `===`). generally, loose equality should be avoided because of how complex its rules are.

this should be either

```suggestion
    if (resource !== undefined) {
```

or

```suggestion
    if (!resource) {
```
(ironic that I recommend this, since it is loose equality still. however, the behavior around `!` is much simpler and harder to mess up)

src/analysis/retail/priest/shadow/modules/talents/MindsEye.tsx
```diff
@@ -0,0 +1,48 @@
...
+    this.active = this.selectedCombatant.hasTalent(TALENTS.MINDS_EYE_TALENT);
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS.DEVOURING_PLAGUE_TALENT),
+      this.onDevouringPlague,
+    );
+  }
+
+  onDevouringPlague(event: CastEvent) {
+    const resource = event.classResources?.at(0)?.cost; //Some buffs grant free Devouring Plagues, which do not have a resource cost
+    if (resource != null) {
```
emallson: same comment as above

emallson (changes requested):
small changes, otherwise looks good

src/analysis/retail/priest/shadow/modules/talents/DistortedReality.tsx
DoxAshe: I've updated it to `!== undefined` from `!=null`, and I'll make sure to avoid loose equality in the future.  Also, I also tried `!resource`, but it didn't seem to work the same.

emallson (approved):

PR #6754, opened 2024-05-03
[Devastation] Volatility module rework

Krealle:
### Description

Blizzard changed how Pyre events functioned at some point leading to this module breaking and showing wildly inaccurate results.

Have now completely reworked the logic behind the module to function based on `CastLinks`. Essentially we just add links between `Pyre` damage events and their associated cast event; either `Pyre` or `Dragonrage`, which throws out 3 `Pyre`s. 

To keep complexity low we don't add links to `Volatility` events, and rather treat non-linked events as procs.

This leads to much more accurate results and a nice reduction in complexity, compared to the old solution.

### Issue with spawned mobs
There is an issue with this implementation, which effectively can't be resolved due to how events are logged for spawned mobs. eg. in [this](https://www.warcraftlogs.com/reports/1R4pNJh8Kfz9PtAC/#fight=3&source=1&type=damage-done&phase=1&pins=0%24Separate%24%23244F4B%24casts%240%240.0.0.Any%24186387984.0.0.Evoker%24true%240.0.0.Any%24false%24357211%5E0%24Separate%24%23909049%24casts%240%240.0.0.Any%24186387984.0.0.Evoker%24true%240.0.0.Any%24false%24375087&ability=357212&view=events&start=2848191&end=2851192) log, the same unique mob is hit multiple times by the same `Pyre` cast, which is impossible. This being a problem due to us relying on Unique mob hits to differentiate between `Cast` and `Proc` hits.

This *will* lead to over counting for these situations, but since there isn't really anything we can do about it, we'll just accept this inherent inaccuracy. 

I'm not sure if this is a WCL issue or if the mobs are actually logged with the same GUID.

image

The effect this has is dependent on which dungeon you are running, `Ruby Life Pools` being the biggest offender due to first boss pumping out adds. From the logs I've looked at this leads to a ~2-5% inaccuracy, with `Ruby` going up to ~10%. 🤷 

### Other changes
- Implemented a small `Pyre` module that will show a breakdown of the damage sources.
- Updated example log
- Fixed guide links in `IntroSection`

### Testing

 `/report/Zz9qAWptC8XnywD3/1-Mythic++The+Nokhud+Offensive+-+Kill+(33:19)/Vollmer/standard/statistics`
image

 `/report/bRHWXLc14CYrjhfa/2-Mythic++The+Azure+Vault+-+Kill+(20:54)/Vollmer/standard/statistics`
image

 `/report/1R4pNJh8Kfz9PtAC/3-Mythic++Ruby+Life+Pools+-+Kill+(26:07)/Vollmer/standard/statistics`
image

image


emallson:
god i hate the GUID re-use bug. how many whelplings does she actually spawn in each wave?

emallson (approved):

emallson:
https://www.warcraftlogs.com/reports/1R4pNJh8Kfz9PtAC/#fight=3&type=resources&phase=1&pins=0%24Separate%24%23244F4B%24casts%240%240.0.0.Any%24186387984.0.0.Evoker%24true%240.0.0.Any%24false%24357211%5E0%24Separate%24%23909049%24casts%240%240.0.0.Any%24186387984.0.0.Evoker%24true%240.0.0.Any%24false%24375087&view=events&start=2848191&end=2851192&hostility=1&source=25.1 this looks like guid re-use. you can see the hp jumping up and down, indicating that the events are mixed from 2 npcs.

PR #6756, opened 2024-05-03
Improve cache busting via Refresh button

emallson:
Part of the v2 server migration greatly improved the use of caching and cache-control headers on the server. This had the side-effect of revealing some issues with our cache-busting method. Specifically: if you bust the cache by setting a query param, the new response is stored under a new URL. But when you refresh the page *the old response under the old URL is read.*

This PR changes the cache-busting method to use the `cache-control` header via the [`cache` parameter of `fetch`](https://developer.mozilla.org/en-US/docs/Web/API/fetch#options). When set to `reload`, the behavior is:

> The browser fetches the resource from the remote server without first looking in the cache, but then will update the cache with the downloaded resource.

per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache).

While I was working on this, I realized that we can't retrieve Awakened parses on the character parses page due to how the partitions are set up on WCL. I worked around it since we don't really support old seasons' logs anyway.

This requires [this PR](https://github.com/WoWAnalyzer/server/pull/61) on the backend.

### DIFF
--- a/src/game/ZONES.ts
+++ b/src/game/ZONES.ts
@@ -21,6 +21,7 @@ export interface Zone {
   frozen?: boolean;
   encounters: Encounter[];
   usePtrTooltips?: boolean;
+  partition?: number;
 }
 
 const ZONES: Zone[] = [
@@ -28,6 +29,7 @@ const ZONES: Zone[] = [
     id: 31,
     name: 'Vault of the Incarnates',
     frozen: false,
+    partition: 4,
     encounters: [
       vaultOfTheIncarnates.bosses.Eranog,
       vaultOfTheIncarnates.bosses.Terros,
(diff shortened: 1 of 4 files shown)


src/game/ZONES.ts
```diff
@@ -21,13 +21,15 @@ export interface Zone {
...
   usePtrTooltips?: boolean;
+  partition?: number;
 }
 
 const ZONES: Zone[] = [
   {
     id: 31,
     name: 'Vault of the Incarnates',
     frozen: false,
+    partition: 4,
```
ToppleTheNun, line 32: Back to partitions mattering! Woo!

ToppleTheNun (approved):

PR #6759, opened 2024-05-06
[Devastation] Scintillation module rework

Krealle:
### Description

Previous `Scintillation` impl was prone to overcounting in AoE situation and when proc hits were close to cast hits. This PR solves that.

- Implemented `EternitySurgeNormalizer` to add `CastLinks` between `Eternity Surge` casts and hits, so we can better differentiate between `Scintillation` hits and casted hits.
- Reworked `Scintillation` module to take advantage of new normalizer.
- Implemented `calculateEffectiveDamageFromCritDamageIncrease` fn for `EventCalculateLib`, to generically calculate crit damage increases. Eg. `Your damaging critical strikes deal 230% damage instead of the usual 200%.`
- Implement this new fn for `Spellweavers Dominance` module.
  - Update multiplier for `Spellweavers Dominance`.

The reason for making a custom normalizer for the `Eternity Surge` `Castlinks`, is that due the nature of how `Eternity Surge` functions it can be difficult to easily attribute hits to casts, due to it having differing travel time based on the distance to your target, and that `Scintillation` procs have no inherent event logged, and the damage events shares the same spellId.

We have to compare amount of hits in certain buffered windows, along with comparing damage (`Scintillation` procs deal 50% of base damage), to ascertain which window of events are from cast/procs, which is non-ideal in the generic `CastLinkNormalizer`. Since we only get reference points as we apply links.

### Testing

- Test report URL: 

1.  `/report/wT279gkJZ3NbYvyV/1-Mythic+Echo+of+Neltharion+-+Kill+(5:43)/Zørix/standard/statistics`
Before:
image
After:
image

2. `/report/Zz9qAWptC8XnywD3/1-Mythic++The+Nokhud+Offensive+-+Kill+(33:19)/Vollmer/standard/statistics`
Before:
image
After:
image


src/analysis/retail/evoker/devastation/modules/normalizers/EternitySurgeNormalizer.ts
```diff
@@ -0,0 +1,278 @@
...
+              rawAmount -= calculateEffectiveDamageFromCritDamageIncrease(
+                initialEvent,
+                SPELLWEAVERS_DOMINANCE_CRIT_MULTIPLIER,
+              );
+            }
+
+            rawAmount /= 2;
+          }
+
+          normalizedAmount = rawAmount;
```
emallson, line 149: minor nit: `rawAmount` doesn't need to be a separate variable, right?

emallson (approved):

src/analysis/retail/evoker/devastation/modules/normalizers/EternitySurgeNormalizer.ts
Krealle: Yea that is a leftover var from previous version of that calc 😅

PR #6771, opened 2024-05-22
update GitHub Actions versions

ToppleTheNun:
### Description

Update GitHub Actions versions so it stops yelling at me.

Makes all of this go away:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -151,7 +151,7 @@ jobs:
           VITE_VERSION: ${{ github.sha }}
           VITE_FORCE_PREMIUM: true
       - run: tar -czf dist.tar.gz dist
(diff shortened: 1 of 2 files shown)


emallson (commented):
assuming CI runs fine, this lgtm

ToppleTheNun:
the tags we create on the Docker image in production today are:
```
wowanalyzer/wowanalyzer:dragonflight
wowanalyzer/wowanalyzer:sha-bb423d0
```

with this setup, it should be:
```
wowanalyzer/wowanalyzer:dragonflight
wowanalyzer/wowanalyzer:sha-bb423d0
```
`type=sha`: https://github.com/docker/metadata-action?tab=readme-ov-file#typesha
`type=ref`: https://github.com/docker/metadata-action?tab=readme-ov-file#typeref

PR #6783, opened 2024-06-04
use addInefficientCastReason everywhere possible

ToppleTheNun:
### Description

Use `addInefficientCastReason` everywhere instead of `event.meta.isInefficientCast` so that logic can be centralized.

Adds an ESLint rule for enforcing this.

### DIFF
--- a/src/analysis/retail/deathknight/blood/components/ResourceWasteProblemRenderer.tsx
+++ b/src/analysis/retail/deathknight/blood/components/ResourceWasteProblemRenderer.tsx
@@ -13,6 +13,7 @@ import EmbeddedTimelineContainer, {
 } from 'interface/report/Results/Timeline/EmbeddedTimeline';
 import { AnyEvent, CastEvent, EventMeta, EventType } from 'parser/core/Events';
 import { useMemo } from 'react';
+import { replace } from 'parser/core/EventMetaLib';
 
 interface WastedMeta extends EventMeta {
   _wastedRp: number;
@@ -45,9 +46,7 @@ export function ResourceWasteProblemRenderer({
               return <>This cast wasted {this._wastedRp} RP.</>;
             },
           };
-          // TODO: Come up with a better way of handling this so that it doesn't break any existing meta
-          // eslint-disable-next-line wowanalyzer/event-meta-inefficient-cast
-          lastCast.meta = meta;
+          replace(lastCast, meta);
         } else {
           const meta = lastCast.meta as WastedMeta;
           meta._wastedRp += event.waste;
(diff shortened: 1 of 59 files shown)


ToppleTheNun:
@emallson need your eyes on this section in Blood: https://github.com/ToppleTheNun/WoWAnalyzer/blob/4797b518ccc62ae6a9ed4dacecbc3876193b19b6/src/analysis/retail/deathknight/blood/components/ResourceWasteProblemRenderer.tsx#L48

src/analysis/retail/deathknight/blood/modules/features/Ossuary.tsx
```diff
@@ -40,12 +46,7 @@ class Ossuary extends Analyzer {
     } else {
       this.dsWithoutOS += 1;
 
-      event.meta = event.meta || {};
-      event.meta.isInefficientCast = true;
-      event.meta.inefficientCastReason = defineMessage({
-        id: 'deathknight.blood.ossuary.ineffectiveCast',
-        message: `This Death Strike cast was without Ossuary.`,
-      });
+      addInefficientCastReason(event, ineffectiveCastMessage);
```
emallson, line 49: this one will conflict with my blood apl pr because its actually broken and displays `[Object object]` right now :see_no_evil: will be an easy conflict to fix though. will revert the change in my pr because this solution is better.

src/analysis/retail/deathknight/blood/components/ResourceWasteProblemRenderer.tsx
```diff
@@ -45,6 +45,8 @@ export function ResourceWasteProblemRenderer({
               return <>This cast wasted {this._wastedRp} RP.</>;
             },
           };
+          // TODO: Come up with a better way of handling this so that it doesn't break any existing meta
+          // eslint-disable-next-line wowanalyzer/event-meta-inefficient-cast
```
emallson: so this is operating on a copy of events and intentionally overwriting the meta of the copy (see [this branch](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6783/files#diff-710f818e171abb0cb5e6daff3b65c2986efefa10f915e06f49dd55e04dd36854R57-R62)). we could potentially add a `replace` option, or just allow this to slide with an `eslint-disable` like you have.

emallson:
the overall approach looks good IMO. just a question of what to do with my blood shenanigans :sweat_smile:

src/analysis/retail/deathknight/blood/components/ResourceWasteProblemRenderer.tsx
ToppleTheNun: I'll write a `replace` function in EventMetaLib. The other thing I've been thinking about is getting the type of the EventMeta off of the event, which could have a function for it here as well.

emallson (approved):

PR #6789, opened 2024-06-08
Improve support for Fyr'alath by tracking channel

oBusk:
### Description

Was picking my sisyphus project #6481 back up, and realised Fyr'alath cast shows up as an instant cast and affects uptime negatively.

Built my own normalizer to look for damage events and removal of the buff which triggers on the final blast.

Also added the Fyr'alath cast to the GCD

### Testing

- Test report URL: `/report/DPcp61VFAQhT2KNC/42-Heroic+Kazzara,+the+Hellforged+-+Kill+(1:18)/Rampagebonk/standard/timeline`
- Screenshot(s):
   image


src/parser/shared/normalizers/FyralathNormalizer.ts
```diff
@@ -0,0 +1,93 @@
+import SPELLS from 'common/SPELLS';
+import { AnyEvent, BeginChannelEvent, EndChannelEvent, EventType } from 'parser/core/Events';
+import EventsNormalizer from 'parser/core/EventsNormalizer';
+import InsertableEventsWrapper from 'parser/core/InsertableEventsWrapper';
+
+/**
+ * Normalizes Rage of Fyr'alath casts to include BeginChannel and EndChannel events.
+ */
+export class FyralathNormalizer extends EventsNormalizer {
```
emallson, line 9: is there a reason that this is not simply a `buffChannelSpec` in the core `Channeling` normalizer?

oBusk: I tried around a little with it, but it didn't seem to work. The buff is a completely separate id than the cast? There's also cases where the buff would never be applied as well

emallson: hmmm okay. this is only gonna be around for the 2 months with the release announcement, so happy enough to get it working now and delete in tww

PR #6817, opened 2024-06-17
[Mistweaver] Fix vivify cast entries bug caused by vivacious vivification change.

vohrr:
### Description

Vivacious Vivification can proc during a hardcasted vivify and is not consumed until the following vivify cast. This is often causing two vivify casts and their subsequent eventlinked healing to occur simultaneously. 
This PR fixes that and does some Guide section verbiage updates

### Screenshots
Log Events Example (
image

Before:
image

After: 
image
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/Vivify.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/Vivify.tsx
@@ -25,6 +25,8 @@ const RAPID_DIFFUSION_SPELLS = [
 const BASE_AVERAGE_REMS = 2.22;
 const RM_AVG_REM_DIFF = 3;
 
+type InvigoratingMistHealPerPlayer = { [key: number]: Set<string> };
+
 class Vivify extends Analyzer {
   static dependencies = {
     spellUsable: SpellUsable,
@@ -34,7 +36,7 @@ class Vivify extends Analyzer {
   protected spellUsable!: SpellUsable;
   protected upliftedSpirits!: UpliftedSpirits;
   casts: number = 0;
-
+  healsPerPlayer: InvigoratingMistHealPerPlayer = {};
   mainTargetHealing: number = 0;
   mainTargetOverhealing: number = 0;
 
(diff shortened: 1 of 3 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/Vivify.tsx
```diff
@@ -280,57 +275,80 @@ class Vivify extends Analyzer {
...
+        const effective = invigHeal.amount + (invigHeal.absorbed || 0);
+
+        fullOverhealHits += !effective ? 1 : 0;
+        healingPerCast += effective;
+        overhealPerCast += invigHeal.overheal || 0;
+
+        this._tallyUpliftedSpiritsCDR(invigHeal);
+        this.healsPerPlayer[targetId]
+          ? this.healsPerPlayer[targetId].add(heal_id)
+          : (this.healsPerPlayer[targetId] = new Set<string>().add(heal_id));
```
emallson, line 321: i'm not sure I follow what this `healsPerPlayer` tracking is doing. it seems like the keys will often not be duplicated because the invigorating mists heals will have different amounts/overheals even if the timestamp is identical.

actually i think i get it---this happens because the invig heal is processed twice. `targetsInGrouping` limits each vivify to processing at most 1 invig per player. then `healsPerPlayer` limits each invig to being processed at most once even if linked to multiple vivifies. is that right?

vohrr: Yep! Its a bit of a bandaid I know, if there is a cleaner way to do this let me know. The identical timestamp threw all my attempts and solving it via EventLinkNormalizer

emallson: yea i think this would *technically* be solvable with some `additionalCondition` magic but its messy enough that I'm not sure its worth it, and it would ultimately be very similar to what you're doing here. this could use a brief explanation of what the two sets are doing but otherwise seems okay.

minor nit would be to use a `Set` instead of an array for `targetsInGrouping` but the list is small enough that I'm not sure if the overhead of using a Set is outweighed by the faster lookups

vohrr: updated

PR #6826, opened 2024-06-18
bump tww talents + re-enable resource inclusion

emallson:
RAIDBOTS data has been improved so this no longer needs the hacks to determine class from spec.

as a result of this import, i had to disable several additional specs:

- Fury Warrior
- BM Hunter
- Survival Hunter
- MM Hunter
- Holy Priest
- Disc Priest

none of the specs that has been re-enabled for TWW was broken by this change (though I think @trevorm4 added some mana values that might be overwritten?)


src/common/TALENTS/priest.ts
```diff
@@ -2,14 +2,6 @@
...
 
 const talents = {
-  ABOMINATION_LIMB_TALENT: {
-    id: 383269,
-    name: 'Abomination Limb',
-    icon: 'ability_maldraxxus_deathknight',
-    maxRanks: 1,
-    entryIds: [96177],
-    definitionIds: [{ id: 101179, specId: 258 }],
-  },
```
emallson, line 12: if anyone is curious why the priest file is -4,102 lines, its stuff like this from the previous bad raidbots data

ToppleTheNun: HUH

emallson: there was a bunch of duplicated data showing the shadow priest spec id. @ljosberinn knows all about it

ljosberinn: shadow is so meta, even rogue and dk got a new shadow spec (and others)

trevorm4:
I didnt add any mana values because i dont have access to beta so couldnt check mana costs :P so nothing to worry about there

PR #6842, opened 2024-06-26
[RFC] Update spec support level description

emallson:
### Description

The motivator for this change is to better represent the maintainership of specs that get basic support added, either by one of us or by a drive-by contributor. I want to accomplish two things:

1. I want to be able to mark a spec as supported without having a dedicated maintainer shown.
2. I want to be able to differentiate between the "Foundation / Core" support level and "actually has a maintainer" support level.

While I was working on this, I also did some cleanup (either because of conflicts or just cuz i was there):
- removed most `builds` support
- removed some unused test tooling

**Current**

We had three main things that determined support level: `contributors`, `isPartial`, and `patchCompatibility`.

- `contributors` being empty would trigger some "No maintainers" / "Unmaintained" messages
- `isPartial` triggered the partial confirmation dialog
- `patchCompatibility` being outdated or null resulted in patch warnings or marking the spec as unsupported.

**Proposal**

Adding specs with basic support puts the current system in an awkward place. The analysis is going to be correct, but isn't going to have spec-specific stuff applied. Should it get the modal warning? But then if it doesn't, should a spec with basic support *and* partial spec support get it?

What I'm proposing is to replace `isPartial` key and most use of `contributors` in determining support. Instead, having a `supportLevel` key with the following behaviors:

- Unmaintained: spec is unmaintained. trying to view a spec like this will prompt you to confirm, as the current `isPartial` does
- Foundation: basically the core/foundational support level. `Abilities` is right, `SpellUsable` is right, but there isn't spec analysis. Contributors **are not** listed as maintainers.
- MaintainedPartial: the spec has a dedicated maintainer and considers spec analysis partially complete
- MaintainedFull: the spec has a dedicated maintainer and considers spec analysis fully complete.

Notably: there is only a modal confirmation if a spec is fully unmaintained *or* behind on `patchCompatibility`. A current spec with even foundational support does not get a warning---it is correct, even if not complete.

This also adds some defaults to the `About` page that don't require configuring the description at Unmaintained/Foundation levels.

### Spec List
image

### Report Footer
image

### About Page

image


ToppleTheNun:
As a whole, I really like the changes here. Explaining what `partial` support meant was kinda awkward to do and this should make it more clear.

I wonder if we should include a similar alert banner for any guide/checklist for a spec that isn't `MaintainedFull` just to caveat the analysis. Putting it just on the about page doesn't surface that information very clearly for most people.

emallson:
Yea, i would like to come up with something for the `Foundation` level in particular that conveys "this does not have a dedicated maintainer" that does not also sound like "don't use this" or "don't trust this" but i haven't sat down again to really think through that

emallson:
I went back through this and did some text updates today. I added a basic badge for the support level and think this might be the direction to go for explaining other support levels. However, I did not use this on the spec page because I did not add badges for the other support levels.

I also spent some time looking at adding some kind of banner/message about the support level but ultimately punted on it. The reason, ultimately, is [that our header is super tall already](https://github.com/WoWAnalyzer/WoWAnalyzer/assets/4909458/aedb8c9e-a719-4709-8646-4098a1ecdc11) and I didn't want to add more to it.

Also also: I made the contributor badges line up with text when they're used inline, which means that they're no longer slightly below or above the baseline of other text. (I am actually a bad example because all-lowercase now means that my username appears below-center relative to the profile picture :see_no_evil:)

## About

image

## Footer

image

## Contributors

### Before
image

### After
image

Sharrq:
In the footer, are you able to hover over foundational support to get a tooltip that explains what that means? 

I like the changes overall, but I would assume the about page is probably the least viewed page of all the analyzers, the footer is probably pretty close to that... But there is also only so much you can do without bloating the header even more.

Maybe you could put a small blurb in the header under the player name or something? Something super short like "foundational support provided by x". Granted that could get to be too much for specs with multiple maintainers, so idk. Mostly spit balling.

But I do think having access to what foundational support means (and the others) without going to the about screen would be helpful.

emallson:
> Maybe you could put a small blurb in the header under the player name or something? Something super short like "foundational support provided by x". Granted that could get to be too much for specs with multiple maintainers, so idk. Mostly spit balling.

I think that doing a bit of layout work is gonna be next on my TODO besides furthering spec support / core TWW work. so i think maybe this gets punted for now and looked at again after the header is made smaller.

> But I do think having access to what foundational support means (and the others) without going to the about screen would be helpful.

I considered adding a tooltip to it but wasn't sure. Can definitely add one

smvoss:
I love this idea foundationally, but my one hesitation really comes down to the verbiage of level of support.

As much as I generally hate "stoplight" representation around these kinds of things, I feel like "maintained" is a word that is really obvious in intention to other programmers / us, but I think wouldn't draw necessarily the correct conclusion to somebody who doesn't have this context. Since this is a wow based tool, I wonder if the verbiage should line up with the standard common -> legendary coloring scheme, but at the same time I don't know if that REALLY solves the problem.

Maybe it doesn't matter that much, but I think it may not be super clear to the end user what the real difference between "Foundation" and "MaintainedPartial" are.

emallson:
that is a very interesting point about "Maintainer" / "Maintained" meaning something specific to people in development that maybe doesn't translate to the general population :thinking: I don't have a real answer for that but I am definitely open to using different terminology for it. I don't have better ideas (yet?) but this PR *also* doesn't add (much*) use of it so I don't see it as an *immediate* problem.

\* there is a usage in the default "About" section for "Foundational" specs

------

I think this has gotten overall positive support, and not much negative. The "negative" (in my reading, correct me if wrong) is more pointing out that this maybe doesn't go *far enough* in solving the "regular people don't understand this" problem. Either due to not surfacing it very well (nothing on the main page, About page being not viewed much), or due to terminology (does "Maintainer" mean the same thing to people outside dev circles?)

I will probably make a final pass on it this weekend (probably adding a descriptive tooltip like Sharrq suggested so at least when you *do* see the footer it explains itself), and then merge. My 2c is that I don't think this needs to solve every problem with communicating these ideas, but general consensus seems like this is an improvement

Sharrq:
Yea I agree that it is fine as is, with room to iterate on further down the road.

emallson:
(posting a reminder to myself to update the docs when this is merged)

smvoss:
>this maybe doesn't go far enough in solving the "regular people don't understand this" problem

I agree, and I think this is still a very good step in the right direction. Incremental updates are better than no updates.

PR #6853, opened 2024-07-05
Add support for Classic Unholy DK

emallson:
### Description

This adds foundation support for Unholy DK, replacing the old template-based Checklist.

This also has two relatively important Core changes for Classic:

1. Adding separate stat ratings for Cata Classic (if we have these anywhere I couldn't find them) so that haste percentage from rating is calculated correctly
2. Adjusting the minimum GCD for Classic to be 1s. The 0.75s min apparently is a [change made for Legion.](https://warcraft.wiki.gg/wiki/Cooldown#Global_cooldown) 

image

image


src/analysis/classic/deathknight/blood/CombatLogParser.ts
```diff
@@ -6,6 +6,7 @@ import AlwaysBeCasting from './modules/features/AlwaysBeCasting';
 import Buffs from './modules/features/Buffs';
 import CooldownThroughputTracker from './modules/features/CooldownThroughputTracker';
 import FoundationGuide from 'interface/guide/foundation/FoundationGuide';
+import ArmyOfTheDead from '../shared/ArmyOfTheDead';
 // Spells
 // import SpellName from './modules/spells';
```
jazminite: Remove import comment now that there is a spell. Also, the `// Spells` comment can be moved above line 9

src/analysis/classic/deathknight/blood/CombatLogParser.ts
```diff
@@ -18,6 +19,7 @@ class CombatLogParser extends BaseCombatLogParser {
     cooldownThroughputTracker: CooldownThroughputTracker,
     // Spells
     // spellName: SpellName,
```
jazminite: Remove import comment now that there is a spell.

jazminite (changes requested):

PR #6855, opened 2024-07-06
[Classic] Foundations Updates

jazminite:
### Description

- Update Foundation Guides to use `div` instead of `p` (DOM warnings)
- Add back Always Be Casting module to Caster template without suggestions.
- Add Always Be Casting module in Demonology Warlock (fixes Active time color)

### Testing

- Test report URL: `/report/8MkqJXZaB13pwLFT/5-Heroic+Magmaw+-+Kill+(3:54)/Voluptuouss`
- Screenshot(s):
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -36,6 +36,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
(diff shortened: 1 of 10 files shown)


emallson (approved):
Looks like this shrank/removed padding but I'll take a peek locally later and see if it actually did and if so do some css tweaks

PR #6857, opened 2024-07-07
adds npc ablities on the timeline module.

JordanKlaers:
### Description

This changes adds npc abilities on the timeline module. A useEffect calls for the data only if the use changes the display toggle to true for the first time. The intention is to only call the data once (unless the other effect dependencies change) to prevent additional page load delay. There are toggles to render npc abilities that were instant cast, channeled casts, silenced/interrupted abliities, and boss abilities. Only abilities that did damage to a player or were interrupted/silenced show up. Interrupted abilities render slightly above the timeline with an arrow that points to when the cast first started.

### Testing

Open the timeline module and toggle the npc casts to render. This works for both raid and mythic plus

- Test report URL: `/report/MQn1XAvPgpcfjhd3/14-Mythic++Ruby+Life+Pools+-+Kill+(27:45)/Eonworm/standard/timeline`
- Screenshot(s):
npcAbilities


src/interface/report/Results/Timeline/Component.jsx
```diff
@@ -197,6 +244,54 @@ class Timeline extends PureComponent {
...
+                {EnemySpellTypeToggle(
+                  'Silenced/Interrupted Abilities',
+                  () => this.handleToggle('silencedInterrupted'),
+                  silencedInterrupted,
+                )}
+                {EnemySpellTypeToggle(
+                  'Boss Abilities',
+                  () => this.handleToggle('bossAbilities'),
+                  bossAbilities,
+                )}
```
emallson: these should all be react components. to do this, you need to change the parameters to `EnemySpellTypeToggle` to be in an object:

```jsx
const EnemySpellTypeToggle = ({ label, toggleCallBack, checked }) => {
  // ...
};
```

and then replace these with:

```suggestion
<EnemySpellTypeToggle
  label={'Instant Cast Abilities'}
  toggleCallBack={() => this.handleToggle('instantCast')}
  checked={instantCast} />
<EnemySpellTypeToggle
  label={'Channeled Abilities'}
  toggleCallBack={() => this.handleToggle('channeledAbilities')}
  checked={channeledAbilities} />
<EnemySpellTypeToggle
  label={'Silenced/Interrupted Abilities'}
  toggleCallBack={() => this.handleToggle('silencedInterrupted')}
  checked={silencedInterrupted} />
<EnemySpellTypeToggle
  label={'Boss Abilities'}
  toggleCallBack={() => this.handleToggle('bossAbilities')}
  checked={bossAbilities} />
```

while this might seem like a style thing, there are subtle differences in how function call syntax vs component syntax is run and its better to use the more common style.

src/interface/report/Results/Timeline/Component.jsx
```diff
@@ -197,6 +244,54 @@ class Timeline extends PureComponent {
               '--cast-bars': castEvents.length,
             }}
           >
+            {ShowNPCSpellsToggle(
+              'Render NPC Spells on the timeline',
+              () => setRenderNPCSpells(!shouldRenderNPCSpells),
+              shouldRenderNPCSpells,
+            )}
```
emallson: same as the other comment. note that creating a synonym via `const X = Y` also works with react components, so this is just converting to component syntax

src/interface/report/Results/Timeline/EnemyCasts.tsx
```diff
@@ -0,0 +1,170 @@
...
+  showInstantCasts,
+  showChanneledCasts,
+  showInterruptedAbilities,
+  showBossAbilities,
+  ...others
+}: Props) => {
+  const renderCast = (event: NpcCastEvent | NpcBeginCastEvent) => {
+    const className = '';
+    const level = 0;
+    return RenderIcon(event, {
```
emallson, line 119: same here, make this a react component.

to expand a little: when you call a function producing JSX like this, it is run immediately and produces JSX which then becomes part of the render tree.

OTOH, if you return this as JSX: `<RenderIcon event={event} ... />` then `RenderIcon` is not immediately called. instead, it (1) counts as a separate component for hooks/state, and (2) becomes a part of the render tree that can be expanded later.

this matters in like two cases, but those two cases cause bugs that are very hard to debug :sweat_smile:

src/interface/report/Results/Timeline/EnemyCasts.tsx
```diff
@@ -0,0 +1,170 @@
...
+      start,
+      windowStart,
+      secondWidth,
+      reportCode,
+      actorId,
+      style: {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        '--level': level > 0 ? level : undefined,
+      },
```
emallson: rather than using `@ts-ignore`, add the custom CSS props to the `style` type:

```
style?: CSSProperties & { '--level'?: number; };
```

(same applies to other `@ts-ignore` uses)

src/interface/report/Results/TimelineTab.tsx
```diff
@@ -3,17 +3,210 @@ import CombatLogParser from 'parser/core/CombatLogParser';
...
+                acc[cur.sourceID] = cur.player;
+              }
+              return acc;
+            },
+            {},
+          );
+
+          //This groups damage events together. Helpful for aoe spells from the enemy that hit multiple players at the same time
+          const nonMeleeDamageEvents = damageStuff
+            .filter((val: any) => val.ability.name !== 'Melee')
```
emallson: best not to rely on ability name for this. however, you can also add this to your original query:
```
"type = 'damage' AND source.type in ('NPC', 'Boss') AND ability.id > 1"
```
(and similar for the cast query since melees have cast events)

Melee having an ability ID is a WCL construct iirc (in-game it is a separate kind of event entirely). the IDs for melees are either 1 (Physical) or `-X` where `X` is a school ID (so `-4` is a Fire-school melee)

there are some "Melee" abilities that don't qualify for this, but often they don't have the name `Melee` either. Garothi Worldbreaker comes to mind. Usually stationary bosses.

src/interface/report/Results/TimelineTab.tsx
```diff
@@ -3,17 +3,210 @@ import CombatLogParser from 'parser/core/CombatLogParser';
...
+                dmgEvent.timestamp <= event.timestamp + 10000 && //Assumes a damage event from an npc ability happens within 10 seconds
+                dmgEvent.sourceID === event.sourceID &&
+                dmgEvent.ability.name === event.ability.name
+              );
+            });
+            event['matchingDmgEvent'] = matchingDmgEvent;
+            return (
+              //remove melee events, and events that do not damage allies.
+              //keep events that were silenced/interrupted
+              event.ability.name !== 'Melee' &&
```
emallson: see above about removing melees via query

src/interface/report/Results/TimelineTab.tsx
```diff
@@ -3,17 +3,210 @@ import CombatLogParser from 'parser/core/CombatLogParser';
...
+  useEffect(() => {
+    const fetchData = async () => {
+      const userRequestFirstTime =
+        !hasUserRequestsNPCSpells.current &&
+        prevShouldRenderNPCSpells.current === false &&
+        shouldRenderNPCSpells === true;
+      //This condition ensures npc abilities are only fetched once for the firs time a use requests them
+      if (userRequestFirstTime) {
+        hasUserRequestsNPCSpells.current = true;
+      }
```
emallson: gonna just say that I don't like this ref-based approach to making the effect run once. seems like just having the correct dependencies array should make this work, but there may be something that I'm missing.

emallson (changes requested):
i have not run this locally yet but will do so this evening and get more feedback in. I have a few comments just from the code itself, though

emallson:
okay, so before reading this feedback I want to say: this works and is really nice in practice, even in raid. I definitely underestimated how useful this would be even as a non-healer. 

what this needs now is polish and handling of some edge cases. my notes from playing with this on my local:

- [x] works in classic raids (classic dungeons are hard unsupported)
  - [x] loading damage taken *from* enemies is 6MB on this Magmaw encounter---over 3x the original query for player events
    - okay this is actually Magmaw-specific. but we should also make sure that the target is a player to prevent bad cases like this!
    - rashok also has this problem. not sure why. environmental damage spam?
- [x] works in brackenhide
  - need to exclude damage taken from Environment (source.id = -1). some things like SW:D appear there
- [x] timelines shown in guides still work and don't have extra UI elements added
- [x] toggling the rendering of npc spells repeatedly causes the events to be re-fetched each time
- toggling the filters does not appear to do anything right now? at least on this classic Chimaeron log: http://localhost:3000/report/whg9yZrGNjzDWLX7/37-Heroic+Chimaeron+-+Kill+(4:12)/Cimeqz/standard/timeline
  - [x] okay some of them work, but boss/instant don't appear to behave as expected
  - [x] silenced/interrupted is very narrow. probably better to expand to "stopped" and just exclude everything that wasn't completed.
- [ ] toggling the filters is very slow in large M+ (Brackenhide: http://localhost:3000/report/JBka7NCwgvR9VGWL/63-Mythic++Brackenhide+Hollow+-+Kill+(32:19)/%D0%9F%D0%BE%D0%BB%D0%B8%D0%BD%D1%81%D0%BA%D0%B8/standard/timeline) to update, even on my work PC which is pretty beefy
  - this is an optimization thing and not a fundamental blocker (I think)
- [ ] the background for the interrupted spells is kinda off. there's a visible gradient that I don't think is intended to be there? maybe it is. my first impression was that it was rendering multiple things stagged and generating a gradient artifact as a result
- [x] there is a noticeable delay when loading the abilities between toggling the feature and having the abilities appear on the timeline. during this time, the timeline is visible but empty. on slower connections this will be even more noticeable
  - this just needs a loading state
- [ ] the timestamp from the npc ability timeline can be overlapped by buffs from the player timeline
  - i am actually questioning whether the npc timeline needs separate timestamps. with tall buff-stacking classes maybe it does :thinking:

JordanKlaers:
I updated the useEffect to use a state variable instead of the ref which more effectivly ensures the effect is not called multiple times.
I made an update to use css variables to manage the toggling of the npc abilities to hopefully speed that up.
I added a "loading" when the data is being fetched. I was not sure how you want it to look visually, so if you are unhappy with how it looks please let me know and ill be happy to update it.
Along the same lines, the red gradient you called out, that was my approach, along with raising the spell icon above the timeline, to indicate that it was a "stopped" cast. if you want that changed (should I just remove the red gradient and leave the red outline?) please let me know.

Also, I define a "stopped" cast as an event that has a begingCast event with no matching cast event. I connect those events in the useEffect when they are initially fetched. I then use the lack of a 'matchedCast' key on entries that have a "beginCast" to be considered a "stopped" cast. Is that an acceptable appraoch or do you have a different way I should detect a stopped cast?

I also add a unique class on the individual npc spells, either "boss", "stopped", "channeled" or  "instant" (not the exact classNames used) and I then toggle their visibility with the css variable matching the class. So that means a boss ability will still show even if its a channeled spell and the channeled spells toggle is off, since its considered a boss abilitiy. I noticed this is somewhat odd with raid bosses.
One of the example urls had one non boss type spell cast, and the 7 others fit the boss category. I thought to update the filters available/how they function if there were only boss type spell casts, but in that example there was more than just boss.
Would you like me to leave set something up to only include filters if there are abilities that match the filter? or perhaps some other way to filter ablities, or maybe not at all?

src/interface/report/Results/Timeline/EnemyCasts.tsx
```diff
@@ -119,50 +118,48 @@ const EnemyCasts = ({
...
   actorId,
-  showInstantCasts,
-  showChanneledCasts,
-  showInterruptedAbilities,
-  showBossAbilities,
   ...others
 }: Props) => {
-  const renderCast = (event: NpcCastEvent | NpcBeginCastEvent) => {
-    const className = '';
+  console.log('is enemy casts rerendering');
```
emallson: as a heads up: the react devtools plugin can show you when something re-renders and explain why it re-rendered. very useful tool!

src/interface/report/Results/TimelineTab.tsx
```diff
@@ -79,10 +86,9 @@ const TimelineTab = ({ parser }: Props) => {
             parser.fight.start_time,
             parser.fight.end_time,
             undefined,
-            "type = 'damage' AND source.type in ('NPC', 'Boss')",
+            "type = 'damage' AND source.type in ('NPC', 'Boss') AND ability.id > 1 AND source.id > 0 AND target.type = 'Player'",
```
emallson: making a note for myself that this *could* fail if a boss/npc casts an ability which then has damage sourced to the environment (which does happen, unfortunately). but this should be fine in 99% of cases and deals with the outliers downloading several MB of JSON

emallson:
This is some really good progress.

- The damage event issue has been basically entirely resolved (down to ~300k from >6mb on [this magmaw](http://localhost:3000/report/whg9yZrGNjzDWLX7/30-Heroic+Magmaw+-+Kill+(3:59)/Cimeqz/standard/timeline)
- All the toggles react instantly on regular boss fights, though there are still some issues on brackenhide. (Brackenhide is a tough but useful test case for "how is this gonna work in the worst M+ logs")
  - Using CSS rather than JS to do the toggle is a good thought, but ultimately isn't enough---partially because the styles themselves take time to compute. 
image
    Purple = CSS computation time, Yellow = JS time

The timeline as a whole doesn't perform *great* on these huge M+ so that isn't the end of the line, just something to keep in mind. 

So, next steps: I will probably do a pass on the UI to tweak the CSS (both for perf (ask me about talent trees :)) and to reduce the space taken up by the toggles) and then ping others to give input on it.

This is a big feature so you've signed yourself up for more process than a PR would normally get :P

JordanKlaers:
do you want toggles lined horrizontally? or do you have a preference for a wayy to move them to take of less space?
ask me about talent trees - what is this in regards to?

Ill try to find a way to make the toggling faster again!

Who do I ping?

emallson:
(i forgot to merge this last week. i am going to merge it early tomorrow when i have time to baby-sit in case of query issues)

PR #6861, opened 2024-07-15
[Classic] Demo Warlock Fabricate Prepull cast events

Bhahlou:
### Description

Metamorphosis and Demon Soul: Felguard are used prepull in Demonology to snapshot mastery. This fabricates cast event at fight start time when aura is found to make the cooldown section correct

### Testing

- Test report URL: /report/QcJmMP6yXCY9wqdK/5-Heroic+Theralion+and+Valiona+-+Kill+(4:00)/5-Bhahlou/standard
- Screenshot(s):
image


src/analysis/classic/warlock/demonology/CombatLogParser.ts
```diff
@@ -11,6 +11,8 @@ import Buffs from './modules/Buffs';
 import CooldownThroughputTracker from './modules/features/CooldownThroughputTracker';
 import FoundationGuide from 'interface/guide/foundation/FoundationGuide';
 import CancelledCasts from 'parser/shared/modules/CancelledCasts';
+import MetamorphosisNormalizer from './modules/normalizers/MetamorphosisNormalizer';
+import DemonSoulNormalizer from './modules/normalizers/DemonSoulNormalizer';
```
jazminite: I could be wrong but could these 2 files be combined into 1 file called `PrepullNormalizer` to analyze both spells?

If the files are combined, I'd move the file to the shared folder so all Warlock specs can use.
If the files stay separate, I'd move the DemonSoulNormalizer to the shared folder so all Warlock specs can use.

jazminite (changes requested):

src/analysis/classic/warlock/demonology/CombatLogParser.ts
Bhahlou: I will try to merge both files. However I see no reason to share with others specs, only demono cast Demon Soul pre-pull :)

jazminite: > I will try to merge both files. However I see no reason to share with others specs, only demono cast Demon Soul pre-pull :)

It might not be best practice (for the Warlock casting), but it's still something they could do, so it's best to share the file between the specs.

Bhahlou: > It might not be best practice (for the Warlock casting), but it's still something they could do, so it's best to share the file between the specs.

You are right. At the end, while merging files, I found that 'PrepullCooldowns' normalizer was already existing and did exactly what I wanted to do

jazminite (approved):
lgtm

PR #6869, opened 2024-07-16
Fixed issue showing ResourceGraph natural regeneration

kfinch:
### Description

An update to ResourceGraph last year accidentally removed parts that make naturally regenerating resources render correctly (with a smooth upward line rather than a step). This change reverts specifically the part that was broken.

### Testing
I tested by looking at several resource graphs, including the one the initial change was for. All non-regenerating resource graphs look identical, but the regenerating ones are fixed.
- Test report URL: `/report/NCgvk7mBrHa9QTAZ/6/19`

Before:
image

After:
image


emallson (approved):
i noticed a *very slight* change to the maelstrom graph (look at around 6:30), but it mostly looks identical

but its so close that im game to fix the regen resources and do another pass if there end up being actual problems with non-regen

## Production

image

## This PR

image

PR #6872, opened 2024-07-18
re-add padding to foundation guide

emallson:
this fixes some padding that was removed in [this PR](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6855/files) that fixed invalid DOM nesting.

the margin value here was load-bearing for making the text readable and the paragraphs get smooshed together without it.

## Production
image

## After PR
image


jazminite (approved):
lgtm

PR #6880, opened 2024-07-22
add foundation support for Cata Survival Hunter

emallson:
### Description

this spec has two bits of cooldown shenanigans:

- Glyph of Kill Shot: if a kill shot does not kill its target, reset its   cooldown. 6s icd. the cooldown reset happens before damage (sometimes  well before damage).
- Explosive Shot: Lock 'n Load provides a cooldown reset + 2 additional  uses. I am not *100%* sure that this is how LnL functions but it fixed all the cooldown issues to implement it this way. I can see cases of 4x ES in a row with only one LnL proc so I'm pretty sure it does

beyond that, just a lot of cleanup. Hunter has a lot of TBC-era stuff lying around still.

i did not add all of the call pet / dismiss pet ids because they are *so rare* in practice but you will see occasional top logs doing pet swaps during downtime.

### Testing
image


jazminite (approved):
lgtm

PR #6893, opened 2024-07-24
use WCL's built-in pagination rather than our 5m-page system

emallson:
this fixes issues with m+ mass pulls like in Algathar Academy that front-load a huge number of events (the first 5m of a high algethars can be 80+ mb of json)

this requires both [server](https://github.com/WoWAnalyzer/server/pull/62) and WCL changes so unfortunately the ability of people to test this out is going to be limited til at least the WCL change is in

### DIFF
--- a/src/interface/report/hooks/useEvents.ts
+++ b/src/interface/report/hooks/useEvents.ts
@@ -1,17 +1,10 @@
-import { fetchEvents } from 'common/fetchWclApi';
+import { WCLEventsResponse } from 'common/WCL_TYPES';
+import fetchWcl from 'common/fetchWclApi';
 import { AnyEvent } from 'parser/core/Events';
 import { WCLFight } from 'parser/core/Fight';
 import { PlayerInfo } from 'parser/core/Player';
 import Report from 'parser/core/Report';
 import { useEffect, useState } from 'react';
-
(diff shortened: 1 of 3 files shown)


src/interface/report/hooks/useEvents.ts
```diff
@@ -23,53 +16,47 @@ const useEvents = ({
   player: PlayerInfo;
 }) => {
   const [events, setEvents] = useState<AnyEvent[] | null>(null);
-  // we manually paginate here. WCL could turn pagination back on in the WCL API but the amount loaded per page is too low
-  // for M+ to be usable without deeper changes on the WCL side.
-  const pageCount = fightPageCount(fight);
-  const [pagesLoaded, setPagesLoaded] = useState(0);
+  const [currentTime, setCurrentTime] = useState<number>(fight.start_time);
 
   useEffect(() => {
```
ToppleTheNun, line 21: this reminds me that I want to refactor our data loading to not use `useEffect` directly at some point. I'll probably lump that into my to-do list.

ToppleTheNun: Made an issue for it here: https://github.com/WoWAnalyzer/WoWAnalyzer/issues/6894

PR #6897, opened 2024-07-26
fix partial support on spec page

ToppleTheNun:
### Description

Fix partial support indication on spec list.

### Testing

- Test report URL: `/specs`
- Screenshot(s):
image


emallson (approved):
whoops

PR #6900, opened 2024-07-26
Update Enhancement and Elemental to 11.0.0

Seriousnes:
### Description

Updating to 11.0.0

### Testing

- Test report URL: `/report/6gXhTvrFYd8zHcBt/21-Mythic+Fyrakk+the+Blazing+-+Kill+(7:20)/Seriousnes/standard/`
- Screenshot(s):


src/analysis/retail/shaman/enhancement/modules/Abilities.tsx
```diff
@@ -414,12 +47,9 @@ class Abilities extends CoreAbilities {
...
         // This is no error. We actually use the elemental shaman elemental blast spell id.
-        spell: TALENTS_SHAMAN.ELEMENTAL_BLAST_ELEMENTAL_TALENT.id,
-        enabled: combatant.hasTalent(TALENTS_SHAMAN.ELEMENTAL_BLAST_ENHANCEMENT_TALENT),
-        charges: combatant.getMultipleTalentRanks(
-          TALENTS_SHAMAN.ELEMENTAL_BLAST_ENHANCEMENT_TALENT,
-          TALENTS_SHAMAN.LAVA_BURST_TALENT,
-        ),
+        spell: TALENTS.ELEMENTAL_BLAST_ELEMENTAL_TALENT.id,
+        enabled: combatant.hasTalent(TALENTS.ELEMENTAL_BLAST_ENHANCEMENT_TALENT),
+        charges: 1, //combatant.getMultipleTalentRanks(TALENTS.ELEMENTAL_BLAST_ENHANCEMENT_TALENT, TALENTS.LAVA_BURST_TALENT),
```
emallson: stray comment?

src/analysis/retail/shaman/enhancement/modules/talents/ElementalBlast.tsx
```diff
@@ -67,10 +67,11 @@ class ElementalBlast extends BaseElementalBlast {
...
 
-    this.maxCharges = this.selectedCombatant.getMultipleTalentRanks(
-      TALENTS.ELEMENTAL_BLAST_ENHANCEMENT_TALENT,
-      TALENTS.LAVA_BURST_TALENT,
-    );
+    this.maxCharges = 1;
+    // this.maxCharges = this.selectedCombatant.getMultipleTalentRanks(
+    //   TALENTS.ELEMENTAL_BLAST_ENHANCEMENT_TALENT,
+    //   TALENTS.LAVA_BURST_TALENT,
+    // );
```
emallson: if this is now truly 1 charge, should just remove the commented section

src/analysis/retail/shaman/shared/Abilities.tsx
```diff
@@ -0,0 +1,384 @@
...
+            default:
+              return c.hasTalent(TALENTS.LAVA_BURST_TALENT);
+          }
+        },
+        cooldown: (_) => (combatant.hasBuff(TALENTS.ASCENDANCE_ELEMENTAL_TALENT.id) ? 0 : 8),
+        category: SPELL_CATEGORY.ROTATIONAL,
+        gcd: {
+          base: 1500,
+        },
+      },
```
emallson: rather than having spec-specific logic for this spell here, this should just be added to the individual specs. it is a little bit of duplication, but makes it easier to keep track of. it also means that we don't need to add extra complexity to the core `Ability` type.

if you really want to keep this here, I will note that the spec and selected combatant are accessible in the shared analyzer so you can just use `combatant` here instead of needing to defer computation with the `(c) => { ... }` stuff.

src/analysis/retail/shaman/enhancement/modules/talents/ElementalBlast.tsx
Seriousnes: It is currently bugged in prepatch and has 1 charge, so this was only meant to be temporary. I'll revert this

emallson: ooooh that makes sense

src/analysis/retail/shaman/shared/Abilities.tsx
Seriousnes: > rather than having spec-specific logic for this spell here, this should just be added to the individual specs. it is a little bit of duplication, but makes it easier to keep track of. it also means that we don't need to add extra complexity to the core Ability type.

My personal opinion is it's easier to maintain _overall_ if shared abilities are in one spot, even if the spec specific logic adds some complexity.

> if you really want to keep this here, I will note that the spec and selected combatant are accessible in the shared analyzer so you can just use combatant here instead of needing to defer computation with the (c) => { ... } stuff

Good idea, done. Also reverted the change to `Ability.ts` `enabled` as a result.

src/analysis/retail/shaman/enhancement/modules/Abilities.tsx
Seriousnes: Yep, was also related to the other charge change. Reverted.

PR #6908, opened 2024-07-30
Enable Frost Mage for TWW

earosselot:
Initial support for Frost Mage in TWW.

- Cleared changelog
- Added Icy Veins Guide to About
- Updated config
- Solved bug when not taking RoF (most likely in TWW right now)
- Added Hide Explanation Toggle to the Guide
- Corrected Thermal Void time (25s -> 30s)

### DIFF
--- a/src/analysis/retail/mage/frost/Guide.tsx
+++ b/src/analysis/retail/mage/frost/Guide.tsx
@@ -7,6 +7,7 @@ import ActiveTimeGraph from 'parser/ui/ActiveTimeGraph';
 import { SpellLink } from 'interface';
 import TALENTS from 'common/TALENTS/mage';
 import PreparationSection from 'interface/guide/components/Preparation/PreparationSection';
+import { HideExplanationsToggle } from 'interface/guide/components/HideExplanationsToggle';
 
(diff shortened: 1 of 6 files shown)


src/analysis/retail/mage/frost/Guide.tsx
```diff
@@ -50,19 +51,22 @@ export default function Guide({ modules, events, info }: GuideProps<typeof Comba
   return (
     <>
       <Section title="Core">
+        <HideExplanationsToggle id="hide-explanations-cooldowns" />
```
emallson: these should technically have different ids since this is actually a DOM id. browsers let it slide but generally should be distinct.

emallson (commented):
i don't see any updates to abilities. are Frost abilities really the same in TWW?

src/analysis/retail/mage/frost/Guide.tsx
earosselot: they haven't changed much, @Sharrq already added Winter's Chill and Permafrost Lances, the only two new talents, on the Fire and Arcane update.

earosselot: ids changed

emallson (approved):

PR #6917, opened 2024-08-03
clean up lingering Shadowlands and Dragonflight stuff

ToppleTheNun:
### Description

Started cleaning up some Shadowlands and Dragonflight stuff and got carried away.


emallson (approved):
looks like a CI failure in e2e. otherwise seems good.

PR #6928, opened 2024-08-09
Fix Classic Arcane Mage active time estimation

emallson:
### Description

This fixes several issues with Arcane Mage in specific and Cata Classic in general. The full list is:

- Built the talent reduction of the Arcane Explosion GCD into the `Abilities` entry
- Updated the Arcane Missiles normalizer and removed the old spell entry
- Add the Shard of Woe Celerity haste buff to `StatTracker`
- Update the Berserking haste % for Cata Classic.

as part of the last bit, I refactored how we add/override spells in `Haste` so that the base module can properly handle common spells across different game versions without subclasses needing to be aware of it.

### Testing

tested on https://wowanalyzer.com/report/ADmhtCJz74qHabfX/23-Heroic+Maloriak+-+Kill+(3:46)/Thunder/standard/overview

**Before**

image

**After**

image

### Other Notes

To find the Arcane Missiles issue, I went through in 30 second increments and checked active time. That was able to narrow it down to a few Arcane Missiles casts bumping active time well above 100% after the other fixes were in place. Might be useful tech to have for future debugging

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -35,6 +35,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
+  change(date(2024, 8, 9), <>Fix the haste bonus of <SpellLink spell={SPELLS.BERSERKING.id} /> in Classic Cataclysm.</>, emallson),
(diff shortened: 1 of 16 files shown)


jazminite (approved):
lgtm

PR #6915, opened 2024-08-01
[Shadow] Update for War Within Prepatch

DoxAshe:
### Description

Not too much has changed for shadow going into the prepatch.  Mindgames has been removed, and Shadow Crash now has two versions.  This update gets everything working again, and I've moved some constants to better be able to change them as tuning continues over the next few weeks.  

- Test report URL: `report/WpZ6XjMmxGLVHzJn/17-Mythic+Ulgrax+the+Devourer+-+Kill+(4:09)/Dreadtesties/standard)`
- Screenshot(s):

PR #6930, opened 2024-08-09
[TWW] Initial support for Resto Shaman

BrocoliCosmique:
Basic reactivation of the spec for 11.0 and minor addition to the Prototype guide : added Reactive Warding text when applicable, healing display on the right side yet to be implemented

### DIFF
--- a/src/analysis/retail/shaman/restoration/CHANGELOG.tsx
+++ b/src/analysis/retail/shaman/restoration/CHANGELOG.tsx
@@ -1,48 +1,14 @@
 import { change, date } from 'common/changelog';
-import SPELLS from 'common/SPELLS';
-import TALENTS from 'common/TALENTS/shaman';
+//import SPELLS from 'common/SPELLS';
+//import TALENTS from 'common/TALENTS/shaman';
 import {
-  Arlie,
-  Elitesparkle,
-  Fassbrause,
-  niseko, Seriousnes,
-  ToppleTheNun,
-  Vetyst,
-  Vohrr,
+  Seriousnes,
   Ypp,
 } from 'CONTRIBUTORS';
-import { SpellLink } from 'interface';
+//import { SpellLink } from 'interface';
 
 // prettier-ignore
 export default [
+  change(date(2024, 8, 9), <>Initial support for Restoration Shaman in The War Within</>, Ypp),
   change(date(2024, 7, 27), <>Partial update for 11.0.0</>, Seriousnes),
-  change(date(2024, 6, 29), <>Added Tier 32 module</>, Ypp),
-  change(date(2023, 12, 2), <>Update suggestions for T31 set bonuses</>, Ypp),
-  change(date(2023, 11, 28), <>Fix Resurgence mana returns</>, Ypp),
-  change(date(2023, 11, 16), <>Reactivate Restoration, implement T31 Tier Set for Restoration</>, Ypp),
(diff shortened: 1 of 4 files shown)


src/analysis/retail/shaman/restoration/CHANGELOG.tsx
```diff
@@ -1,48 +1,14 @@
...
-  Fassbrause,
-  niseko, Seriousnes,
-  ToppleTheNun,
-  Vetyst,
-  Vohrr,
+  Seriousnes,
   Ypp,
 } from 'CONTRIBUTORS';
-import { SpellLink } from 'interface';
+//import { SpellLink } from 'interface';
```
kfinch: Please just remove unneeded imports, they can be added back in when they're needed

tsconfig.json
```diff
@@ -59,9 +59,7 @@
     "src/analysis/retail/hunter/beastmastery",
     "src/analysis/retail/hunter/marksmanship",
     "src/analysis/retail/hunter/survival",
-    "src/analysis/retail/hunter/shared",
-    "src/analysis/retail/shaman/restoration",
-    "src/analysis/retail/shaman/shared"
+    "src/analysis/retail/hunter/shared"
```
kfinch: Mistaken removal of shaman from tsconfig?

tsconfig.json
```diff
@@ -80,7 +78,6 @@
     "./src/analysis/retail/warrior/fury/CONFIG.tsx",
     "./src/analysis/retail/hunter/beastmastery/CONFIG.tsx",
     "./src/analysis/retail/hunter/marksmanship/CONFIG.tsx",
-    "./src/analysis/retail/hunter/survival/CONFIG.tsx",
-    "./src/analysis/retail/shaman/restoration/CONFIG.tsx"
+    "./src/analysis/retail/hunter/survival/CONFIG.tsx"
```
kfinch: Mistaken removal of shaman from tsconfig?

kfinch (changes requested):

tsconfig.json
BrocoliCosmique: Saw that on another commit ( https://github.com/WoWAnalyzer/WoWAnalyzer/pull/6915/commits/a1588eea958555d7803f2aeb4ee9ff05f66cd676 ) and thought it was required as part of the reactivation of the spec. Doesn't look like it makes any difference however, so now I'm confused as to what exactly is "excluded" about these specs.

BrocoliCosmique:
Thanks for the review @kfinch, the code has been updated according to your remarks.

tsconfig.json
kfinch: You know that's a fair point I'm not sure what the deal is with this partial list of specs here. Still, it is not needed for reactivation and we shouldn't change it until we know why we're changing it.

kfinch (approved):

tsconfig.json
emallson: If you're re-enabling a spec, it should be removed from tsconfig. The exclusion only applies to un-imported files in dev mode but it seems to do weird things with lazy loading (?) in production if you enable a spec and don't remove from tsconfig. We'll see if there are any unusual errors with resto shaman 😅

kfinch: Oh shit my bad I think I got mixed up on what this was - I guess we'll see what happens lmao

PR #6934, opened 2024-08-10
[Devastation] Add Always Be Casting guide section

Krealle:
### Description

Added a `Always Be Casting` guide section.

Also implemented `buffAndNextCastChannelSpec` for `channeling` normalizer. 

Using `buffChannelSpec`, chained casts of `Disintegrate` on the same target were both getting their `endChannelevent` put at the same `removeBuff` event, since `refreshBuff` events aren't checked. `buffChannelSpec`  ignores cast events to allow instants that are usable during a channel, this doesn't apply to spells such as `Disintegrate`.

I've therefore added a new channel spec that will end a channel on either next `removeBuff`, `cast` or `beginCast` event.

### Testing

- Test report URL: `/report/76abxPF2BjrK83Yg/6-Mythic+Volcoross+-+Kill+(2:23)/Dwx/standard/overview`
- Screenshot(s):
image


src/parser/shared/normalizers/Channeling.ts
```diff
@@ -355,6 +355,53 @@ function buffChannelSpec(spellId: number): ChannelSpec {
...
+ * when it starts, and then a RemoveBuff event when it finishes. Some instead put a Debuff on the target, but the
+ * principle is the same.
+ *
+ * This handler works by handling cast events with the given guid, and then scanning forward for the
+ * matched removebuff (or removedebuff) event, as well as another cast or begincast event
+ * and then making the pair of beginchannel and endchannel events based on them.
+ *
+ * @param spellId the guid for the tracked Cast and RemoveBuff/RemoveDebuff events.
+ */
+function buffAndNextCastChannelSpec(spellId: number): ChannelSpec {
```
emallson, line 370: i feel like in order to prevent bad worst-case behavior, this should have a max duration (if possible without breaking chained casts). but i can't come up with a scenario where it would break because we also reset on the next cast of the same ability. so the worst case scenario is one where you never cast it again and we never see a removebuff---in which case we just look through the remaining events once. so maybe it is fine as-is :thinking:

emallson (approved):

src/parser/shared/normalizers/Channeling.ts
Krealle: It's only for the buff/debuff part where it checks for same spell, for casts it will reset on any cast. So it should be pretty safe.

emallson: Is it going to reset on any cast? the handler is only called for the guids in the list. otherwise the default handler is called (which might actually end the channel. it does at least for `begincast`. not familiar enough to fully parse the `cast` branch)

Krealle: The normalizer works by starting a channel on a given event, then it will do a forward search throughout the event array for the specified end event.

emallson: oh, yup, i see what you're saying now. :+1: i agree this seems safe

PR #6968, opened 2024-08-20
Mage Arcane - guide cleanup stuff + Arcane Tempo improvements

kfinch:
### Description

This PR contains some Guide cleanup stuff and also improvements to the Arcane Tempo section.
* Removed Supernova from Cooldown bars section, as it is not rotational. I considered removing the Guide section entirely but didn't. What's there does not seem to mesh with what's in any of the guides. I think it will be used rotationally with Unerring Proficiency only - the module should probably be reworked to be Unerring Proficiency specific.
* Removed Arcane Orb from Cooldown bars section - it's short cooldown and used reactively, doesn't belong in the section about visually longer cooldown use.
* Moved Evocation and PoM cooldown bars behind a talent conditional. (Most players won't take PoM, we shouldn't be showing the bar for them).
* Removed Siphon Storm module entirely - Evocation usage is already covered by the Cooldown bars, and the only actual Siphon Storm check is if it lines up with Arcane Surge, which is already being checked in the Arcane Surge section.
* Improvements to the Arcane Tempo display - we now calculate and display "average stacks", and I added a statistic as well modeled after the Starlord one in Balance Druid.

### Testing

- Test report URL: `/report/r4bwPWDqV1BYgpad/32-Mythic+Ulgrax+the+Devourer+-+Kill+(6:46)/凛冽寒风/standard/overview`
- Screenshot(s):
image
image
image


Sharrq (approved):
Only thing I would say is I think the arcane orb cast efficiency might be important enough to keep, but maybe not in the cool downs section. Maybe that can go in the arcane orb section so it is at least visible. 

But everything looks good to me

kfinch:
Updated Arcane Orb text and added the suggested cooldown bar

image

PR #6973, opened 2024-08-21
[Holy Paladin] Talent updates

Texleretour:
### Description

- Add Overflowing Light effective absorb stat

### Testing

- Test report URL: `/report/fHk1awXpvV3yMZg7/59-Mythic+Rashok,+the+Elder+-+Kill+(2:50)/Priiyaa/standard/statistics`
- Screenshot(s):
image


src/analysis/retail/paladin/holy/modules/talents/LightsProtection.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Statistic
+        size="flexible"
+        category={STATISTIC_CATEGORY.TALENTS}
+        tooltip={<p>Estimated damage reduced: {formatNumber(this.totalDamagePrevented)}</p>}
+      >
+        <TalentSpellText talent={TALENTS_PALADIN.LIGHTS_PROTECTION_TALENT}>
+          <div>
+            {formatNumber(this.DRPS)} <small>estimated DRPS</small>
+          </div>
+        </TalentSpellText>
```
emallson: this is only including damage reduction on yourself, but the talent reads "Allies with Beacon of Light receive 5% less damage". is it actually player-only or should this be adjusted (either text adjustment to clarify, or code adjustment to include DR on other players)

Texleretour: mb, i have to fix it it's all players w/ beacon indeed

Texleretour: is fetching WCL for additional data the only way to go ? Just checking out before I do something extremely inefficient @emallson

emallson: we probably don't have helpers that you can use to determine who has the buff (though you might be able to iterate over friendlies?), but:

- we have `combatantinfo` events for every player, which *should* include pre-pull buffs. (if beacon is missing lmk and i'll pass it along)
- you can see `applybuff`/`removebuff` to other players by the paladin during combat, so if they move it you can track it
- worst case: you can see who is being healed by it and use that to determine who had the buff on pull

emallson: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/c3be75fbf6adc6cd6b30a19486847c5f8ee41645/src/parser/shared/modules/Combatants.ts#L12 this might be useful to iterate over players

Texleretour: mmh maybe I'm just not seeing it but that seems a little expensive if the player is running Beacon of Virtue (the aoe beacon on a 15 sec CD), doesn't it ? @emallson 

image
it looks like it's the same w/ Events.damage

EDIT: Is the Discipline's Lenience implementation outdated?

emallson: oh i see the issue---you can tell when someone has beacon but not what damage they take.

yea this should be a load-on-demand stat. the preservation `TimeOfNeed` module does something similar to fetch the data, so you could start there (though that talent is a wee bit more complicated in how it works so everything *after* data loading is going to be different)

it is also okay for some talents to not have stats, even if interesting.

Texleretour: Thanks for the answer, it looks cleaner than Lenience. There isn't any non-deprecated load-on-demand component ? 
Either way your last point might conclude this. I may go back to this later but there are more important talents to implement rn (hello hero talents)

emallson (approved):

PR #6979, opened 2024-08-23
I have once again rewritten ABC

kfinch:
### Description
The edge cases in the current version were driving me crazy so I rewrote it again.
This new version tracks activity "edges" that can arbitrarily overlap, removing the need for overlap detection that caused all the bugs.
To calculate actual active time, we regenerate active time segments treating the 0->1 edge as activity start and the 1->0 edge as activity end.

### Testing
- Test report URL: `/report/XJytz49PfbYL6mvR/14-Mythic+Smolderon+-+Kill+(3:45)/Zulyd/standard/overview`

NEED TO TEST WITH A BUNCH MORE REPORTS BEFORE THIS GOES IN PROBABLY. POSTING FOR REVIEW.


kfinch:
This might have a bad interaction with AlwaysBeCastingHealing, checking on that

src/parser/shared/modules/AlwaysBeCasting.tsx
```diff
@@ -147,10 +84,139 @@ class AlwaysBeCasting extends Analyzer {
...
+   * Generates working active time segements if needed. Most likely no one queries the
+   * activeTimeSegments until after processing is complete, in which case this function will be
+   * called only once.
+   *
+   * Algorithm for turning edges into segments:
+   * activeTimeEdges represent when arbitrarily overlapping individual active time segments start
+   * or stop. If we sort them in chronological order and then step through while incrementing the
+   * counter an activity start edge and decrementing the counter on an activity end edge, we can
+   * detect inactivity time (when counter is 0) and activity time (when counter is greater than 0).
+   * Use this to generate the segment's union, which will be non-overlapping and in order.
```
emallson, line 141: this is smart, i like this. the code is easy to understand too

src/parser/shared/modules/AlwaysBeCasting.tsx
```diff
@@ -147,10 +84,139 @@ class AlwaysBeCasting extends Analyzer {
...
+  /** The active time (in ms) recorded */
+  get activeTime() {
+    if (this.activeTimeMemo === undefined) {
+      this.activeTimeMemo = this.getActiveTimeMillisecondsInWindow(
+        this.owner.fight.start_time,
+        this.owner.fight.end_time,
+      );
+    }
+    return this.activeTimeMemo;
+  }
```
emallson: does this break if you change the window being considered? i think probably not because the CLP gets rebuilt? i think?

emallson (approved):
this looks so much cleaner. and, on top of that: i think it will be really easy to fix the Cata Arms Warrior issue by adding activity edges based on where they have no spells available to cast. (or, at least it will be easy to adjust active time once the detection of those periods is written!)

question about what happens with the memoized number when you switch phases of a fight, but i'm pretty sure it works as-is (and if not, it is not going to be hard to change---the memoization can just be keyed by the current start time)

emallson:
(i also have not run this locally yet so ymmv. im happy to jam some reports through this weekend)

src/parser/shared/modules/AlwaysBeCasting.tsx
kfinch: I think the play here is to also check the memo against start and end time, and if either of those change we clear the memo.

kfinch:
Will make updates to the memoization and test some more now that I'm back from vacation. Hopefully get this in soon

kfinch:
Hooking in support for HABC made this slightly uglier, but I think it's still good and readable

emallson (approved):

PR #6986, opened 2024-08-25
remove report Redux reducer

ToppleTheNun:
### Description

Removes the report Redux reducer, which was storing the entire report in Redux state, while we're also storing the entire report in the ReportContext.

### Testing

- Test report URL: `/report/6YPW2m1bMBLAgvQG/12-Mythic+Tindral+Sageswift,+Seer+of+the+Flame+-+Kill+(5:17)/148-Toppledh/standard`


emallson (commented):
not necessarily opposed to this but: what prompted this?

also: do we stash state in browser storage (cookie or `localstorage`)? if so do we need to rev a cache key?

ToppleTheNun:
I noticed that with very large reports, I was getting a console warn about a Redux reducer taking longer than 250ms to execute and tracked it down to the report reducer. It made sense to me because we were storing the report twice. So I did a little digging and found that the report in Redux was only being used in the Navigation Bar, so that led me to this refactor. It does feel a bit snappier when I'm navigating locally, at least.

As far as I can tell, the only part of the Redux store that's actually cached on the client side is the report history portion that we show on the right side of the home page, so this shouldn't be a problem.

emallson (approved):
that makes complete sense 👍

PR #6989, opened 2024-08-27
[Resto Shaman] Downpour targets & lives saved statistic

BrocoliCosmique:
### Talent updates for TWW :
- Downpour :
  - Target cap 6 -> 5
  - No more debuff based on targets hit
  - New buff +10% max hp
- Ancestral vigor
  - 2 ranks -> 1 rank for the talent for 10% max hp

### Lives saved statistic
Tried to adapt the module to account for both Ancestral vigor and Downpour, while making it not *that much* more intensive. ~~As a result the estimated number of lives saved is very wrong.~~ Should be correct now.

- Test report URL: `report/yPWVrMvA9x7FZ6bK/1-Mythic++Algeth'ar+Academy+-+Kill+(23:59)/Ypp/standard/about`, `/report/Q1F9LCcDzp3gvNaH/1-Mythic++Siege+of+Boralus+-+Wipe+1+(31:24)/Hdershaman/standard/statistics`
- Screenshot(s):
image
image


src/analysis/retail/shaman/restoration/modules/talents/AncestralVigor.tsx
```diff
@@ -127,11 +138,65 @@ class AncestralVigor extends Analyzer {
...
+          event.timestamp,
+          100,
+          50,
+        )
+          ? ANCESTRAL_VIGOR_INCREASED_MAX_HEALTH
+          : 0;
+
+        // compute health ratio relevant for the statistic, with the current active buffs
+        const currentBonusHealthRatio =
+          1 - 1 / (1 + currentAncestralVigorIncrease + currentDownpourIncrease);
```
Seriousnes: Buffs that provide a percentage increase multiplicative, not additive. For example, with a base HP of `4016122` having both buffs will give a maximum hp of `4859508` not `4819346` (difference of `40161`).

`(1 + 0.1) * (1 + 0.1) = 1.21`, vs `1 + (0.1 + 0.1) = 1.20` which is exactly the difference (1% max hp)

Seriousnes (changes requested):

src/analysis/retail/shaman/restoration/modules/talents/AncestralVigor.tsx
BrocoliCosmique: Good to know, I was so convinced in stacked additively that I didn't even test it. it is fixed now.

Seriousnes (approved):

PR #6996, opened 2024-09-01
Wrote EchoOfLight Attributor

Liavre:
### Description

Wrote a module that is intended to be used by almost every other Holy Priest talent/spell to get a reasonably accurate approximate attribution from the resulting mastery HoT.

image

In an effort to make the PR not unbearably long, the attributor is being pushed separately and implementing it into the statistics of all the talents will be done later in its own PR.


src/analysis/retail/priest/holy/normalizers/CastLinkNormalizer.ts
```diff
@@ -241,7 +244,25 @@ const EVENT_LINKS: EventLink[] = [
       return !HasRelatedEvent(linkingEvent, HARDCAST_POWER_WORD_SHIELD);
     },
   },
-
+  {
+    linkRelation: ECHO_OF_LIGHT_ATTRIB_EVENT,
+    linkingEventId: null,
```
emallson, line 248: if you have `ABILITIES_THAT_TRIGGER_MASTERY` (seen below), why not use it here as well?

src/analysis/retail/priest/holy/modules/core/EchoOfLightAttributor.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * if ticks exist, get the amount each tick overhealed and apply that against the
+     * initial heal (raw * mastery %) divided per tick number, and apply getEffectiveHealing math per tick
+     */
+    if (eolHealEvents.length > 0) {
+      const initialHealPerTick = initialHeal / eolHealEvents.length;
+
+      const empty: HealEvent[] = []; //eslint....
+      Object.values(
```
emallson: why is this wrapped in `Object.values`?

src/analysis/retail/priest/holy/modules/core/EchoOfLightAttributor.tsx
```diff
@@ -0,0 +1,128 @@
...
+        eolHealEvents.reduce((n, eolHealEvent) => {
```
emallson: it doesn't look like `n` is used, so this can be `eolHealEvents.forEach` instead, which then also means that you don't need the `empty` value.

src/analysis/retail/priest/holy/modules/core/EchoOfLightAttributor.tsx
```diff
@@ -0,0 +1,128 @@
...
+    // blacklist non EOL heals
+    if (ABILITIES_THAT_DONT_TRIGGER_MASTERY.includes(eolEvent.ability.guid)) {
+      return 0;
+    }
+
+    // get average healing/overhealing over the next 6s of EOL
+    const eolHealEvents = GetRelatedEvents(eolEvent, ECHO_OF_LIGHT_ATTRIB_EVENT);
+    if (eolHealEvents.length > 0) {
+      const empty: HealEvent[] = []; //eslint....
+      Object.values(
```
emallson: same notes here about unneeded `Object.values` / using `forEach` instead of `reduce`

src/analysis/retail/priest/holy/constants.ts
```diff
@@ -103,6 +103,16 @@ export const ABILITIES_AFFECTED_BY_APOTHEOSIS_TALENT = [
...
 
+export const ABILITIES_THAT_DONT_TRIGGER_MASTERY = [
+  TALENTS.RENEW_TALENT.id,
+  SPELLS.PREMONITION_OF_PIETY.id,
+  SPELLS.DIVINE_IMAGE_BLESSED_LIGHT_HEAL.id,
+  SPELLS.DIVINE_IMAGE_DAZZLING_LIGHT_HEAL.id,
+  SPELLS.DIVINE_IMAGE_HEALING_LIGHT_HEAL.id,
+  SPELLS.DIVINE_IMAGE_TRANQUIL_LIGHT_HEAL.id,
+  TALENTS.ESSENCE_DEVOURER_TALENT.id,
+];
```
emallson: rather than listing this out, it might be better to keep the whitelist so that you don't accidentally attribute mastery healing to non-spec heal sources like embellishments, trinkets, or potions.

emallson (changes requested):

src/analysis/retail/priest/holy/normalizers/CastLinkNormalizer.ts
Liavre: Originally I had it for some math, I will delete ABILITIES_THAT_TRIGGER_MASTERY link below since I never ended up using it. It is easier to use a blacklist than a white list

src/analysis/retail/priest/holy/constants.ts
Liavre: Embellishments/trinkets/potions/etc do contribute to EoL. It is triggered from pretty much anything that isnt a HoT or from a pet, regardless of if it is class specific.

emallson: oh, interesting. i did not know that. that makes your other use of a null/blacklist rather than whitelist make a ton of sense

src/analysis/retail/priest/holy/modules/core/EchoOfLightAttributor.tsx
Liavre: That is how the reference I used did it, now I changed it to how the swiftmend module uses its forEach implementation

Liavre:
Implemented the changes posted. Echo of light is a particularily messy ability since it can trigger off non spec/class things, it will have be be maintained with a blacklist (or whitelist but that could be more work).

the approximate ends up being very close, and its still possible this is an error from something not being blacklisted that should be (units are in percentage for a temporary module I'm only using locally to validate).
image

emallson (approved):

PR #6999, opened 2024-09-05
[Arms/Fury] Ambitiously track all rage generation

oBusk:
### Description

This PR is a product of my long term work to improve Fury and Arms Warrior support, which I'm trying to separate out into multiple PRs to be reasonable.

One thing that was clear was that rage tracking is spotty and haphazard. I'm a believer in Abelitos mantra that a lot of things should be normalizers, and have adopted this approach. Normalize events by ensuring ResourceChange events exist for all missing sources, and the separately, normalize by attributing all resourcechanges to different spells to provide more information to the end user.

Melee hits, while lacking the "resourche change" events, actually track current rage, so if we track what rage is before meleecast happens, we can figure out how much was generated. It get's trickier when you are already capped, because then we cannot use this, and need to calculate how much you _should_ have generated. This is relatively straight forward for Arms, but get's annoying for fury since offhand generates half the rage. I use a few tricks to try to figure out if any meleecast is offhand or not, but in the end, it's going to be a guess. 

I created one globally shared module, a WindfuryLinkNormalizer, that connects Auto Attack with the "ExtraAttach" event, and the "ExtraAttack" event with the subsequent new meleeswing. This can be helpful for any melee attack that tries to figure out things like swing timers.

### Testing

- Test report URL: 
  - Fury `/report/TLGtVJ2Mz8cxKwYk/1-Mythic+Gnarlroot+-+Kill+(2:15)/Ezlx/standard/rage-usage`
  - Arms `/report/zRKf7jqag2J384tx/36-Mythic+Volcoross+-+Kill+(3:37)/Vigoo/standard/rage-usage`
- Screenshot(s):
image
image


oBusk:
So tracking is pretty good now, it stays in sync with logged rage for the most part. It _does_ mismatch sometimes because of timing issues. Some rage gains triggering slower than we'd hope for, meaning other events can sneak in between. 

I'll do a lot more verifying of talents once S1 begins, but the code is pretty stable right now, so I'll mark as ready for review!

The larger normalizers like the RageEventGenration and RageAttribute are a bit large and hard to read, tried looking into ways to make them more structured, but without turning them into massive state machines I couldn't really find any way... Would love suggestions!

src/analysis/retail/warrior/arms/normalizers/FinishingBlowsResourceChange.ts
```diff
@@ -0,0 +1,28 @@
...
+import TALENTS from 'common/TALENTS/warrior';
+import { AnyEvent, EventType } from 'parser/core/Events';
+import EventsNormalizer from 'parser/core/EventsNormalizer';
+
+/**
+ * With the change in War Within pre-patch where Strength of Arms no longer gives 8 rage
+ * when casting Overpower on a target below 35% health, the log still logs with the Strength of Arms
+ * ability, so we just replace it with the Finishing Blows to get a better attribution.
+ */
+export default class FinishingBlowsResourceChange extends EventsNormalizer {
```
emallson, line 11: i don't necessarily agree with this being a normalizer. normally this is something that you can group later without too much difficulty. does this using the "wrong" ability have other bad downstream effects?

src/analysis/retail/warrior/arms/normalizers/ImprovedExecuteNormalizer.ts
```diff
@@ -0,0 +1,85 @@
...
+        (event.ability.guid === SPELLS.EXECUTE.id ||
+          event.ability.guid === SPELLS.EXECUTE_GLYPHED.id)
+      ) {
+        const rage = getRage(event, this.selectedCombatant);
+
+        if (rage?.cost != null && rage.cost > 0) {
+          const damageEvent = executeDamageEvent(event);
+          if (damageEvent?.overkill == null || damageEvent?.overkill < 0) {
+            // Possible this should be .floor() instead of .round()
+            const refund = Math.round(rage.cost * 0.1);
```
emallson, line 36: round is probably close enough absent any direct proof one way or the other.

src/analysis/retail/warrior/shared/modules/debuggers/RageCountDebugger.tsx
```diff
@@ -0,0 +1,137 @@
...
+        }
+      }
+
+      if (expectedRage != null && rage?.amount != null) {
+        if (rage.amount === expectedRage) {
+          // This doesn't really work in the debug UI.
+          // this.addDebugAnnotation(event, {
+          //   summary: `Rage correct: ${rage.amount}`,
+          //   color: 'green',
+          // });
```
emallson, line 63: are there just too many events when you enable this?

emallson (approved):
tentative approve---not a fan of rewriting what spell an event logs as when there are other solutions (afaik---correct me if i am wrong) that are less invasive. rewriting events like that has potential for sneaky errors later, like if they *do* fix it and then re-use the spell (as part of a new talent, for example) for something else.

src/analysis/retail/warrior/shared/modules/debuggers/RageCountDebugger.tsx
oBusk: Yeah it's for every single event that logs the current rage, which is all spenders and gains. The debug UI renders one row of dots per minutes, so if you get 100s of them, they squeeze and overflow. Not sure if the debugui should handle that many dots either. I've only just now discovered the debug ui

src/analysis/retail/warrior/arms/normalizers/FinishingBlowsResourceChange.ts
oBusk: Do you disagree with the entire "fixing" of rage happening in normalizers, or specifically of this rewriting one spell to another?

My starting point was looking at the [protection RageGenerationNormalizer](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/ab7815a8394297bb51b76e225692fdc97f43d6c5/src/analysis/retail/warrior/protection/normalizers/RageGenerationNormalizer.ts), which uses the same principle, but also trys to attribute various bonuses as a single normalizer, which becomes complicated, so I've separated the addition of new events with splitting/attribution into separate normalizers. Following the same model, attributing the rage gain from [Finishing Blows](https://www.wowhead.com/spell=400205/finishing-blows) to the talent Finishing Blows, rather than the talent [Strength of Arms](https://www.wowhead.com/spell=400803/strength-of-arms). Strength of Arms still exists, but does not have any rage gain anymore, so any logging of rage gains to that talent is just incorrect. 

If they were to adjust the logging event to correctly point to the Finishing Blows talent, it should work as expected, this normalizer will just be useless. If they were to change strength of arms to have some other rage gain in the future, it will be annoying and I'll have to figure out which rage is which, but it will require changes eitehr way.

I'd love to know if you think there's ways that are cleaner or more robust!

src/analysis/retail/warrior/arms/normalizers/ImprovedExecuteNormalizer.ts
oBusk: I kept having rounding mismatchews every now and then however I rounded it :'( 

Actual rage is probably a float, but they log it as rounded values. I thought the whole 100 rage = 1000 rage was to not have decimals :( But yeah rounded is probably as good as it gets. Could remove this comment

src/analysis/retail/warrior/arms/normalizers/FinishingBlowsResourceChange.ts
emallson: i think that the splitting and reattributing in normalization is a pretty reasonable thing to do, especially when you have a lot of analyzers that rely on those numbers being correct. basically: that is doing hard, complicated things to make everything else easier down the line.

switching out a spell isn't really doing something hard. it makes some things a little bit easier, but isn't doing the same kind of heavy lifting.

however, there's a lot of churn happening right now and i also think it is important to let y'all get code online to support things now that the raid is out so i'm not gonna block the PR over this, just be mindful of it and if it ends up being easy to do this *without* rewriting the spell then maybe come back and adjust that

src/analysis/retail/warrior/arms/normalizers/ImprovedExecuteNormalizer.ts
emallson: yea, sounds similar to other resources then. we just do the best we can with the logged info

src/analysis/retail/warrior/shared/modules/debuggers/RageCountDebugger.tsx
emallson: that makes sense

src/analysis/retail/warrior/arms/normalizers/FinishingBlowsResourceChange.ts
oBusk: @emallson I think that is a great way to view things. So would the other way to adjust the tracking of this spell in the ResourceTracker?

PR #7004, opened 2024-09-06
[Evoker] Update MajorDefensive module for Lifecinders & Hardened Scales

Krealle:
### Description

- Added the +10% multiplier for Hardened Scales.
- Updated Normalizer for cleaning up buffs for Renewing Blaze (it was removing a bit too many now that you have external buff in the mix)
- Updated CastLinks for Renewing Blaze
- Updated Renewing Blaze def module to account for partner buff.

### Testing

- Test report URL: `/report/92fBTX4tjxZN7gMp/19-Mythic++The+Necrotic+Wake+-+Kill+(38:19)/Salfevion/standard/overview`
- Screenshot(s):


Krealle:
@trevorm4 Don't think you are using this module, but tagging because it is shared change.

trevorm4 (approved):
lgtm

PR #7006, opened 2024-09-07
[core] switch to ESLint 9

ToppleTheNun:
### Description

Switch to ESLint 9 and flat config since ESLint 8 is EOL in October. Pulls some heavy inspiration and a couple of snippets from @putstack/eslint-config-typescript (thanks, Putput).

Also bumps the TypeScript version.


emallson:
superceded by #7309

PR #7008, opened 2024-09-08
[core] add S1 raid/dungeons, clean out old raids

ToppleTheNun:
### Description

Add TWW S1 raid and dungeons, clean out old raids.

### Testing

- Test report URL: `/report/WVCHLMcNzPyt1mQ9`, `/report/TwA2FKjpqb1k3RDW`
- Screenshot(s):
image
image


src/game/raids/builders.ts
```diff
@@ -0,0 +1,10 @@
+import type { Boss } from 'game/raids';
+
+export function buildBoss(params: { id: number; name: string; background?: string }): Boss {
+  return {
+    ...params,
+    headshot: `https://assets.rpglogs.com/img/warcraft/bosses/${params.id}-icon.jpg`,
```
ToppleTheNun, line 6: Requesting review from @emallson mostly because of the change to use this rather than embed our own copy of the JPGs.

emallson: ill double check but that is probably fine

PR #7026, opened 2024-09-13
Druid Feral - updated guide text to indicate finishers should always be used with 5 CPs. Fixed a typo in Rip breakdown text

kfinch:
During Dragonflight, we sometimes wanted to use finishers at 4 CPs, but in TWW it's always 5. Updated some checks and also Guide text to account for this.

### DIFF
--- a/src/analysis/retail/druid/feral/modules/spells/RipUptimeAndSnapshots.tsx
+++ b/src/analysis/retail/druid/feral/modules/spells/RipUptimeAndSnapshots.tsx
@@ -244,12 +244,6 @@ class RipUptimeAndSnapshots extends Snapshots {
         <CastSummaryAndBreakdown
           spell={SPELLS.RIP}
           castEntries={this.castEntries}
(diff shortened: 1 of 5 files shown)


src/analysis/retail/druid/feral/modules/spells/RipUptimeAndSnapshots.tsx
```diff
@@ -244,12 +244,6 @@ class RipUptimeAndSnapshots extends Snapshots {
         <CastSummaryAndBreakdown
           spell={SPELLS.RIP}
           castEntries={this.castEntries}
-          goodExtraExplanation={
-            <>
-              or a cast with problems that procced{' '}
-              <SpellLink spell={TALENTS_DRUID.BLOODTALONS_TALENT} />
-            </>
-          }
```
kfinch, line 252: Rip can't proc BT - I think this was an accidental copy paste from Rake.

PR #7035, opened 2024-09-15
[Core] Update spellAvailable to properly adjust behaviour to inverse

Pewtro:
### Description

The inverse option only changed the text, not the validation behaviour which made it require an optional not() to achieve that effect. 
I've come to prefer option objects over raw booleans that aren't descriptive of what they actually do

### DIFF
--- a/src/parser/shared/metrics/apl/conditions/spellAvailable.tsx
+++ b/src/parser/shared/metrics/apl/conditions/spellAvailable.tsx
@@ -12,7 +12,7 @@ export default function spellAvailable(
   spell: Spell,
   options: SpellAvailableOptions = {},
 ): Condition<UpdateSpellUsableEvent | null> {
-  const inverse = !!options.inverse;
+  const inverse = Boolean(options.inverse);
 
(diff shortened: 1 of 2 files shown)


src/parser/shared/metrics/apl/conditions/spellAvailable.tsx
```diff
@@ -18,7 +24,8 @@ export default function spellAvailable(
         return state;
       }
     },
-    validate: (state, _event) => state === null || state.isAvailable,
+    validate: (state, _event) =>
+      state === null || state.isAvailable || (inverse && !state.isAvailable),
```
emallson: shouldn't this be:

```suggestion
    validate: (state, _event) =>
      state === null || (!inverse && state.isAvailable) || (inverse && !state.isAvailable),
```

otherwise you match when the spell is available even when inverted

Pewtro: Yes, good catch - fixed

PR #7036, opened 2024-09-16
Fix Yulon cd duration

trevorm4:
### Description

Yulon duration in cooldowns tab was completely borked, so just removing this custom logic as debugging rems is very costly :3 

before:
image

after:
image


vohrr (changes requested):
Removing this makes the fromYulon method in HotTracker and the attribution in HotAttributor unused, lets remove them as well.

vohrr (approved):

vohrr:
Lgtm just need to resolve the changelog conflict

PR #7043, opened 2024-09-17
Fix broken link for Wowhead Class Guides

JoeParrinello:
### Description

Updated Wowhead link for outdated specs to point to working link.

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -22,6 +22,7 @@ import {
   Putro,
   Seriousnes,
   Sref,
+  Taevis,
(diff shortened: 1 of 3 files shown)


ToppleTheNun:
Add yourself as a contributor in https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/CONTRIBUTORS.ts and a changelog entry in https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/CHANGELOG.tsx and this should be good to go.

ToppleTheNun (approved):

PR #7053, opened 2024-09-20
[hpal] new spells + beacon config

mfield:
### Description

Configured some new TWW spells and hero talents.
Added a debugging feature to the beacon module to guess what unconfigured heal might have caused the beacon target and whether it's a full transfer or half.

console is mostly quiet now

### Testing

- Test report URL: `https://www.warcraftlogs.com/reports/CKhDxdX9c8Mg23FB#fight=27&type=healing&source=13`
- Screenshot(s):


src/analysis/retail/paladin/holy/modules/beacons/BeaconHealSource.ts
```diff
@@ -199,6 +212,54 @@ class BeaconHealSource extends Analyzer {
...
+          Math.abs(expectedBeaconTransfer - beaconTransferRaw) <= 2 // allow for rounding errors on Blizzard's end
+        );
+      });
+
+      if (match) {
+        this.log('Likely match:', { ability: fmtAbility(match.ability), healEvent: match });
+        return;
+      }
+
+      const matchHalf = this.unconfiguredHeals.findLast((healEvent) => {
```
emallson, line 244: since you're only looking at unconfigured heals this list is probably not going to be too long but noting that this could have pretty bad performance if that ever changes

emallson (approved):

PR #7080, opened 2024-09-25
fix: TWW Fix in the premium login component

danny-janse:
There is an error being throw on the login premium tab on the local environment only.
You simply head to this page; http://localhost:3000/premium and the console will display the following error;

> React does not recognize the `classNames` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `classnames` instead.


emallson (commented):
this is not a correct fix (unless the underlying library changed its prop names *again*, sigh). the warning comes from the `TransitionGroup` component not properly sanitizing its properties before passing on to underlying html elements. changing this to `classnames` will almost certainly make the transitions break

danny-janse:
I'm not sure how to properly replicate the feature in itself. From what I can find you're probably right and react is throwing out errors/warnings for no reason. Happy to close the MR if no proper testing method exists.

emallson:
i just checked on production (finally, sorry) and the transition looks broken anyway

danny-janse:
Seems to work now?

EDIT:
I see that you changes some code around yourself. Nice!

PR #7082, opened 2024-09-26
feature: TWW Surging Totem PrePull Normalizer

danny-janse:
As reported by Unreality/Killashaman over on discord. Surging Totem is not properly loaded. Upon investigation this seems to be caused because the ability was used just before the pull and was still active once the pull started. To solve this i've added a normalizer that based on the totem's death event will add a prepull cast event. The death event allowes  for correct calculation of when the totem was placed.

You can use the following report to test this yourselves;
https://wowanalyzer.com/report/Cj4AFaK7Y6DQwXmN/51-Heroic+The+Bloodbound+Horror+-+Kill+(4:45)/Killashaman/standard/character

### DIFF
--- a/src/analysis/retail/shaman/restoration/modules/Abilities.ts
+++ b/src/analysis/retail/shaman/restoration/modules/Abilities.ts
@@ -79,7 +79,9 @@ class Abilities extends CoreAbilities {
       },
       {
         spell: TALENTS.HEALING_RAIN_TALENT.id,
-        enabled: combatant.hasTalent(TALENTS.HEALING_RAIN_TALENT),
+        enabled:
+          combatant.hasTalent(TALENTS.HEALING_RAIN_TALENT) &&
+          !combatant.hasTalent(TALENTS.SURGING_TOTEM_TALENT),
         category: SPELL_CATEGORY.ROTATIONAL,
         cooldown: 10,
         timelineSortIndex: 17,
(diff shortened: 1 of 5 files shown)


src/analysis/retail/shaman/restoration/modules/Abilities.ts
```diff
@@ -100,9 +102,7 @@ class Abilities extends CoreAbilities {
         category: SPELL_CATEGORY.ROTATIONAL,
         cooldown: combatant.hasTalent(TALENTS.TOTEMIC_SURGE_TALENT) ? 24 : 30,
         timelineSortIndex: 17,
-        gcd: {
-          base: 1500,
-        },
+        gcd: null,
```
emallson, line 105: is this really off-gcd? or is it cast "by the totem" (and only the totem) or something? if it is only cast by the totem, it should go in `CASTS_THAT_ARENT_CASTS` instead

emallson (approved):
im going to merge this because it fixes rsham analysis, but making a note that if the totem healing rain is not an actual ability used by rshams, it should be `CASTS_THAT_ARENT_CASTS` instead of an off-gcd ability entry with separate cooldown tracking

src/analysis/retail/shaman/restoration/modules/Abilities.ts
danny-janse: It is cast by totem, so doesn't really have a CD. I see you've already merged this, does it need any change?

PR #7084, opened 2024-09-26
fix: Disc priest changelog

danny-janse:
I noticed that the retail disc priest ChangeLog had a wrongly formatted date, Which caused the date to appear in the far future: 2025. This PR corrects this mistake.

image


Pewtro:
Thanks :)

PR #7107, opened 2024-10-04
[Holy Paladin] Uptime issues fix

Texleretour:
### Description

Added GCD to Holy Armaments and Shield of the Righteous
Removed GCD to Blessing of the seasons

Increased Holy Prism cooldown to 30 seconds

### DIFF
--- a/src/common/SPELLS/paladin.ts
+++ b/src/common/SPELLS/paladin.ts
@@ -285,6 +285,11 @@ const spells = {
     name: 'Blessing of Spring',
     icon: 'ability_ardenweald_paladin_spring',
   },
+  SACRED_WEAPON_TALENT: {
+    id: 432472,
+    name: 'Sacred Weapon',
+    icon: 'inv_ability_lightsmithpaladin_sacredweapon',
+  },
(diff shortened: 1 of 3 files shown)


src/common/SPELLS/paladin.ts
```diff
@@ -285,6 +285,11 @@ const spells = {
     name: 'Blessing of Spring',
     icon: 'ability_ardenweald_paladin_spring',
   },
+  SACRED_WEAPON_TALENT: {
```
danny-janse, line 288: I'd personally not suffix this with `_TALENT` since its not inside the TALENTS folder, but that's just a matter of taste

danny-janse (approved):
lgtm

src/common/SPELLS/paladin.ts
Texleretour: What would you suffix it with ? I'm not so used to the conventions going around here :)

emallson: its kind of an odd case because its an ability that you cast that is granted by a talent...but it isn't in the talent data dump. i don't think we really have a convention for that.

Issue #7090, opened 2024-09-28
Arcane Mage Analysis Breaking


Acratin:
Getting this basic type of bug with every attempt to use on Arcane Mage

Sentry event ID: 85b4e690a2954b459569777504cd2abf

Cannot read properties of undefined (reading 'borderColor')

TypeError: Cannot read properties of undefined (reading 'borderColor')
    at https://wowanalyzer.com/assets/CombatLogParser-kJiKV-As.js:1:16683
    at Array.map (<anonymous>)
    at Te.render (https://wowanalyzer.com/assets/CombatLogParser-kJiKV-As.js:1:16568)
    at Gf (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:65371)
    at Rg (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:65169)
    at P_ (https://wowanalyzer.com/assets/index-1xYkTXSt.js:47:9613)
    at C_ (https://wowanalyzer.com/assets/index-1xYkTXSt.js:47:946)
    at uk (https://wowanalyzer.com/assets/index-1xYkTXSt.js:47:874)
    at Yo (https://wowanalyzer.com/assets/index-1xYkTXSt.js:47:728)
    at ad (https://wowanalyzer.com/assets/index-1xYkTXSt.js:45:10801)
    at https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:38677
    at e.unstable_runWithPriority (https://wowanalyzer.com/assets/index-1xYkTXSt.js:32:3836)
    at Nr (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:38455)
    at Z1 (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:38623)
    at un (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:38558)
    at er (https://wowanalyzer.com/assets/index-1xYkTXSt.js:45:8411)
    at Object.enqueueSetState (https://wowanalyzer.com/assets/index-1xYkTXSt.js:43:42540)
    at ro.setState (https://wowanalyzer.com/assets/index-1xYkTXSt.js:18:1371)
    at https://wowanalyzer.com/assets/CombatLogParser-kJiKV-As.js:1:16093
The above error occurred in the component:
    at Te (https://wowanalyzer.com/assets/CombatLogParser-kJiKV-As.js:1:15728)
    at div
    at section
[...]

Issue #7066, opened 2024-09-23
Mage Arcane Error While rendering


ViniciusTafuri:
This is usually caused by a bug in our code. If you're handy with computers please consider sending us a Pull Request with a fix on [GitHub](https://github.com/WoWAnalyzer/WoWAnalyzer). Otherwise please let us know in an issue on [GitHub](https://github.com/WoWAnalyzer/WoWAnalyzer) or leave us a message on [Discord](https://wowanalyzer.com/discord) so we can fix it for you.

The error
Technical information to help you fix it. Or us if not you. Technical information to help whoever is inclined to fix the issue.

Sentry event ID: 45373e4110fd43d7bd2fc18a3739af68

Cannot read properties of undefined (reading 'borderColor')

TypeError: Cannot read properties of undefined (reading 'borderColor')
    at https://wowanalyzer.com/assets/CombatLogParser-CfpuSIA2.js:1:16730
    at Array.map (<anonymous>)
    at Te.render (https://wowanalyzer.com/assets/CombatLogParser-CfpuSIA2.js:1:16615)
    at Gf (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:65371)
    at Rg (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:65169)
    at P_ (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:47:9613)
    at C_ (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:47:946)
    at uk (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:47:874)
    at Yo (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:47:728)
    at ad (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:45:10801)
    at https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:38677
    at e.unstable_runWithPriority (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:32:3836)
    at Nr (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:38455)
    at Z1 (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:38623)
    at un (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:43:38558)
    at er (https://wowanalyzer.com/assets/index-1Xr_gzdK.js:45:8411)
[...]

PR #7113, opened 2024-10-05
[Arcane Mage] Update for The War Within (Again)

Sharrq:
This PR updates several modules that required changes after the onslaught of blizzard hot fixes and changes that Arcane Mage received in the first few weeks of the expansion.

- Fixed Arcane Charge Tracking (Manually)
- Rewrote Arcane Barrage to use the new requirements
- Rewrote Arcane Missiles to use the new requirements
- Added an indicator if the player did not cast Arcane Orb at all in the fight, instead of showing NaN and empty bars
- Adjusted Nether Precision and Clearcasting to use CastSummaryAndBreakdown
- Fixed a bug that would crash logs for dungeons like Siege of Boralis and Mists which have more than 5 bosses (including boss summons)

image


Sharrq:
This resolves #7090 and #7066

Sharrq:
@emallson not sure if there was a better way to fix the ManaLevelGraph crash other than adding more border colors for bosses with more than 5 "bosses"
@kfinch I updated everything, I think, for the latest arcane changes. Ill probably merge this later today or tomorrow and can tweak it as people provide feedback.

emallson:
> not sure if there was a better way to fix the ManaLevelGraph crash other than adding more border colors for bosses with more than 5 "bosses"

i think adding extras to prevent crashes is fine. we can look at specific boss fights / dungeons to fix cases where non-boss enemies are incorrectly included

PR #7115, opened 2024-10-05
Remove BDK rotational analysis and handle Deathbringer effects

emallson:
I have not had the time to fix up the rotational stuff properly. It changes a lot with Deathbringer in the mix. So we're removing it rather than leaving in wrong info.

This also includes fixes to several issues related to Deathbringer:

- Added Reaper's Mark
- Normalize Marrowrend's Rune cost when Exterminate is present
- Handle Exterminate in RuneWaste and MarrowrendUsage
- Handle Deathbringer-related CDR

image


src/analysis/retail/deathknight/blood/CombatLogParser.ts
```diff
@@ -104,13 +104,13 @@ class CombatLogParser extends CoreCombatLogParser {
...
     deathStrike: DeathStrike,
     bloodShield: BloodShield,
-    aplCheck: AplCheck,
 
     // normalizers
     deathStrikeNormalizer: DeathStrikeLinkNormalizer,
     bloodShieldNormalizer: BloodShieldNormalizer,
     resourceOrderNormalizer: ResourceOrderNormalizer,
     boneShieldOrderNormalizer: BoneShieldOrderNormalizer,
+    ExterminateCostNormalizer,
```
danny-janse, line 113: This seems to be differnt from the standard we're using.
```suggestion
    exterminateCostNormalizer: ExterminateCostNormalizer,
```

danny-janse (commented):
Noticed a small thing, lgtm besides that.

src/analysis/retail/deathknight/blood/CombatLogParser.ts
emallson: the `dependencies` array needs to be unique by name except that you also have to match the name of superclasses you're overriding. ive started using the shorthand a bit. if people don't like it i can switch back to using `lowerCamelCase: CamelCase` instead

src/analysis/retail/deathknight/shared/ExterminateCostNormalizer.ts
```diff
@@ -0,0 +1,73 @@
...
+export default class ExterminateCostNormalizer extends EventsNormalizer {
+  normalize(events: AnyEvent[]): AnyEvent[] {
+    const newEvents = [];
+    let activeBuff: number | undefined = undefined;
+    for (let event of events) {
+      if (this.isExterminateBuff(event) && event.type === 'applybuff') {
+        activeBuff = event.ability.guid;
+      } else if (
+        this.isExterminateBuff(event) &&
+        event.type === 'removebuff' &&
```
danny-janse: Wouldn't this nicer? Keep in mind that you cannot just apply this suggestion without importing the EventType enum aswell.
```suggestion
        event.type === EventType.RemoveBuff &&
```

src/analysis/retail/deathknight/shared/ExterminateCostNormalizer.ts
```diff
@@ -0,0 +1,73 @@
+import SPELLS from 'common/SPELLS';
+import talents from 'common/TALENTS/deathknight';
+import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
+import { AnyEvent, CastEvent, HasAbility } from 'parser/core/Events';
```
danny-janse: In case the EventType suggestions are applied.
```suggestion
import { AnyEvent, CastEvent, EventType, HasAbility } from 'parser/core/Events';
```

src/analysis/retail/deathknight/shared/ExterminateCostNormalizer.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * The "Exterminate" buff reduces Obliterate/Marrowrend rune costs by 100% or 50% (depending on which buff). This is not reflected in the `classResources` field of the event, which breaks rune tracking.
+ */
+export default class ExterminateCostNormalizer extends EventsNormalizer {
+  normalize(events: AnyEvent[]): AnyEvent[] {
+    const newEvents = [];
+    let activeBuff: number | undefined = undefined;
+    for (let event of events) {
+      if (this.isExterminateBuff(event) && event.type === 'applybuff') {
```
danny-janse: Here too

```suggestion
      if (this.isExterminateBuff(event) && event.type === EventType.ApplyBuff) {
```

src/analysis/retail/deathknight/shared/ExterminateCostNormalizer.ts
```diff
@@ -0,0 +1,73 @@
...
+      HasAbility(event) &&
+      (event.ability.guid === SPELLS.EXTERMINATE_BUFF_1.id ||
+        event.ability.guid === SPELLS.EXTERMINATE_BUFF_2.id)
+    );
+  }
+
+  isExterminateConsumer(event: AnyEvent): event is CastEvent {
+    return (
+      HasAbility(event) &&
+      event.type === 'cast' &&
```
danny-janse: And here
```suggestion
      event.type === EventType.Cast &&
```

src/analysis/retail/deathknight/shared/ExterminateCostNormalizer.ts
emallson: good catch

PR #7117, opened 2024-10-05
attempt to remove user (ip) information before sending to sentry

emallson:
vetyst pointed out that sentry currently includes IP information. we don't have any need for this info, so this is an attempt to disable it.

ref:
- https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/identify-user/
- https://docs.sentry.io/platforms/javascript/data-management/sensitive-data/#before-send-before-send-transaction

i need to test this with an actual sentry token to confirm nothing breaks, but also would not mind comment (esp from @Pewtro @kfinch)

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -40,6 +40,7 @@ if (import.meta.env.VITE_SENTRY_DSN) {
     allowUrls: ['wowanalyzer.com/assets/'],
 
     beforeSend(event) {
+      // this is *an attempt* to keep sentry from sending up user info that we don't want & don't need
       if (event.user) {
         delete event.user;
       }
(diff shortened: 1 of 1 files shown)


emallson:
sentry is being a pain filtering this 💀 

i can confirm that it doesn't cause a crash at least, but hard to confirm whether it will also halt all errors being reported (which it shouldn't, but never know).

i can put this in and then revert immediately if it does shut down error reporting.

danny-janse:
@emallson Checking if the errors still come through can be easiliy done with issue #7066

src/index.tsx
```diff
@@ -39,6 +39,14 @@ if (import.meta.env.VITE_SENTRY_DSN) {
     environment: import.meta.env.VITE_ENVIRONMENT_NAME,
     allowUrls: ['wowanalyzer.com/assets/'],
 
+    beforeSend(event) {
+      if (event.user) {
```
Pewtro, line 44: Maybe we can add a comment _why_ we're doing this, but otherwise seems fine

PR #7120, opened 2024-10-06
fix(TTW): Resolve rare instance of duplicated events.

danny-janse:
Noticed an error being throw for Void Torrent talent of the shadow priest through sentry. Upon investigation it seemed that all of the events were duplicated as you can see in the log here;

* /report/hKXnp4AcNJjzC7VP/9-Heroic+The+Bloodbound+Horror+-+Kill+(4:09)/Popispectre/standard/events

I notived that the events were being downloaded from the WCL Api. Investigating this it seemed that the fight was logged/uploaded twice within with the same timestamps. To fix this issue i decided to filter down to the events of the selected fight only.

The effects are quite noticable for the Statistics overview;

Before:
image
image

After:
image
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -38,6 +38,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
+  change(date(2024, 10, 6), 'Resolved a rare instance of duplicated events', Vetyst),
   change(date(2024, 10, 5), 'Fix viewing the character tab for certain regions.', Vetyst),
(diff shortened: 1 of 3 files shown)


danny-janse:
I would've filtered the events by fight through the API if there was an option for this but it only accepts start_time and end_time

emallson:
After thinking about this a bit, I don't believe that it is worth imposing the extra runtime cost on *every* fight view in order to fix a bug with corrupted logs.

On the WCL side, the standard practice is "corrupted logs don't have to work, we just have to be able to clean them up". I think that is ultimately the most sensible approach here.

PR #7123, opened 2024-10-06
[Core] Updating item level scaling data

Seriousnes:
### Description

* Updated stat multiplier scaling tables from simc data

### Testing

- Screenshot(s):


src/common/ITEMS/thewarwithin/trinkets.ts
```diff
@@ -16,6 +16,11 @@ const trinkets = {
     name: "Spymaster's Web",
     icon: 'inv_11_0_raid_spymastersweb_purple',
   },
+  QUICKWICK_CANDLESTICK: {
+    id: 225649,
+    name: 'Quickwick Candlestick',
+    icon: 'trade_archaeology_candlestub',
+  },
```
danny-janse, line 23: Would you be able to add a very simple analyzer for the trinket? One that just adds the ability to the spellbook via the core combatLogParser This will allow all specs to stop throwing warnings from the spellUsable when this specific trinket is used.

 See example: `src/parser/retail/modules/items/thewarwithin/trinkets/SignetOfThePriory.ts`

src/analysis/retail/shaman/shared/Abilities.tsx
```diff
@@ -22,6 +23,12 @@ class Abilities extends CoreAbilities {
         gcd: null,
         enabled: faction === 'Alliance',
       },
+      {
+        spell: SPELLS.BERSERKING.id,
+        category: SPELL_CATEGORY.COOLDOWNS,
+        gcd: null,
+        enabled: combatant.race === RACES.Troll,
+      },
```
danny-janse: Are you sure Berserking isn't already in the spellbook? This should be dynamically added via the core combatLogParser if the character is detected to be a troll.

danny-janse (commented):
Some good small changes. A few things caught my attention, see comments below.

ps: I haven't really looked into the statsMultiplierTable data as im not sure how to properly test this.

src/analysis/retail/shaman/shared/Abilities.tsx
Pewtro: Berserking has it's [own module](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/parser/shared/modules/racials/troll/Berserking.jsx) already so there should be no need to add it :)

Seriousnes: This fixed the `spellUsable` debug annotation of ability missing from the spellbook, so I'm guessing the berserking module isn't working as we'd hope

Seriousnes: I'll revert this. The issue was caused by a M+ report where the combatant's race is `null`.

emallson:
@Pewtro had opinions on this. would appreciate a note from him before merging

Pewtro:
Is this PR meant to be changing 93 files with a bunch of non item level scaling, or can we separate things out?

Seriousnes:
> Is this PR meant to be changing 93 files with a bunch of non item level scaling, or can we separate things out?

Definitely not, I've rebased the branch and now it has the correct no. of modified files

src/parser/core/stats.test.jsx
```diff
@@ -64,12 +64,12 @@ describe('stats', () => {
...
+    const spoilsOfNeltharus = (itemLevel) => calculateSecondaryStatDefault(250, 482, itemLevel);
 
-    expect(spoilsOfNeltharus(382)).toBeWithin(2253.943, 1); // Normal
-    expect(spoilsOfNeltharus(402)).toBeWithin(2528.523, 1); // Mythic+
-    expect(spoilsOfNeltharus(441)).toBeWithin(3063.954, 1); // Max upgrade Mythic+
-    expect(spoilsOfNeltharus(447)).toBeWithin(3146.328, 1); // Vault
+    expect(spoilsOfNeltharus(382)).toBeWithin(2209, 1); // Normal
+    expect(spoilsOfNeltharus(402)).toBeWithin(2514, 1); // Mythic+
+    expect(spoilsOfNeltharus(441)).toBeWithin(3108, 1); // Max upgrade Mythic+
+    expect(spoilsOfNeltharus(447)).toBeWithin(3199, 1); // Vault
```
Pewtro:
```suggestion
    expect(spoilsOfNeltharus(382)).toBeWithin(2209.46, 1); // Normal
    expect(spoilsOfNeltharus(402)).toBeWithin(2513.98, 1); // Mythic+
    expect(spoilsOfNeltharus(441)).toBeWithin(3107.79, 1); // Max upgrade Mythic+
    expect(spoilsOfNeltharus(447)).toBeWithin(3199.14, 1); // Vault
```

image
image
image
image

Pewtro (commented):
My initial opposition to this was that it was suggested to 

1. have a different set of stat scaling tables for each expansion
2. remove stat scaling tests for DF items
3. both

which are in my opinion all wrong options, but this now looks to be updating the tests to correct values which is the only correct option in my mind. 

I think for future tests, we should add some in with the appropriate level range, and not go "out of bounds" of what is possible in current expansion/our current stat tables (this might mean updating the stat tables mid expansion) - that way it shouldn't be changing when we get future scaling patterns (until a level/ilevel squish inevitably comes around)

Pewtro:
and well, the tests NEED to pass on old trinkets too because we have all the scaling values so there is no reason the function shouldn't be calculating correctly for a lower ilevel trinket 🤔

Pewtro:
Vessel of Searing Shadow and Spoils of Neltharus are both on scaling -7 which is what we have as default - so they should both be passing tests

image
image

Effect 1 from spoils being seen used in the stat buffs (not pictured) and effect 4 from vessel being reference in the tooltip as increasing haste by $s4

Pewtro:
Basically I am worried that we are calculating incorrectly from a lower item ilevel to a much higher item level (250 -> 382), this is the only test in stats.jsx where we calculate such a large jump, and it also happens to be the only one failing now

Edit: I didn't read far enough - they all fail right now, but from a quick test with updating them to now accurate values, the jewelry tests all pass and the vessel of searing shadow test also passes. It's only neltharus failing by a wide margin that isn't fixed by just updating the values - which still brings me to the original part of this comment where I am worried about our calculation being very inaccurate over wide ilevel spans

Commit that updates the remaining values to showcase that: https://github.com/Pewtro/WoWAnalyzer/commit/abd35c4ede25edd1417e69af21c80a885694d33b

emallson:
I will pull updated values for the neltharus trinket in game and compare. Assuming it hasn't changed, we can probably just be more precise with scaling

emallson:
okay, i tracked down the issue. ultimately, the exponential approximation only works within limited ilvl bands (basically within the max level of the same expansion) because it looks like the raw item budget can shift in arbitrary ways when crossing those boundaries. the ilvl 250 version of the neltharus trinket is a bit of a headscratcher because I don't think you could ever get one of those? maybe as part of level up? not sure.

using this neltharus trinket as an example, you can get exact stat matches by applying the formula `rand prop point * trinket combat rating mult * spell coeff`:

| ilvl | rand prop point | trinket cr mult | coeff | **stat gain** |
| --- | ---- | --- | --- | --- |
|250|127.3841171|1.5|2.521|481.7030389|
|460|1361.187378|0.989953698|2.521|3397.078958|

but note that the rand prop point (aka "raw, base item budget") is not exponential based on ilvl based on either the "15% per 15 ilvl" version or the "1% per 1 ilvl" version. it is more than both

it seems like our current method would continue to be mostly fine within that context. single digit rating errors, etc, but nothing major.

but now i've gone through the effort of tracking all of the tables down to do this correctly...so i may just pr that

Pewtro (approved):
I'm fine with this, we just need to do `this.selectedCombatant.getTrinket(ITEMS.SPYMASTERS_WEB.id)!.itemLevel,` in the spymaster implementation, since it's already guarded by `this.selectedCombatant.hasTrinket(ITEMS.SPYMASTERS_WEB.id);`, so we know that getTrinket can't return undefined. 

Then need to get lingui to ignore stats.ts since it doesn't support import attributes

emallson:
we might be able to get lingui to behave by switching babel plugin order so the import attributes get stripped out before it sees them 🤔

PR #7127, opened 2024-10-08
Hide Circle of Healing analysis if not talented

xizbow1:
### Description

Circle of Healing analysis in the Overview section was showing regardless of whether or not it was talented.

### Testing

- Test report URL: `https://www.warcraftlogs.com/reports/FPh2BbcpaCYfKLXw#fight=12&type=summary&source=7`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/priest/holy/modules/spells/CircleOfHealing.tsx
+++ b/src/analysis/retail/priest/holy/modules/spells/CircleOfHealing.tsx
@@ -25,10 +25,7 @@ class CircleOfHealing extends Analyzer {
   constructor(options: Options) {
     super(options);
 
-    if (!this.selectedCombatant.hasTalent(TALENTS.CIRCLE_OF_HEALING_TALENT)) {
-      this.active = false;
-      return;
-    }
+    this.active = this.selectedCombatant.hasTalent(TALENTS.CIRCLE_OF_HEALING_TALENT);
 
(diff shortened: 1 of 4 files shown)


src/analysis/retail/priest/holy/modules/spells/CircleOfHealing.tsx
```diff
@@ -82,6 +82,9 @@ class CircleOfHealing extends Analyzer {
   }
 
   get guideSubsection(): JSX.Element {
+    if (!this.selectedCombatant.hasTalent(TALENTS.CIRCLE_OF_HEALING_TALENT)) {
```
trevorm4: So this is a way of doing it but I don't think its the correct way, you should just prevent calling this function by adding a check here https://github.com/WoWAnalyzer/WoWAnalyzer/blob/87fee3fc4ce93725cb89e8aefdfddc79b247c8bd/src/analysis/retail/priest/holy/Guide.tsx#L25

Can follow example from https://github.com/WoWAnalyzer/WoWAnalyzer/blob/87fee3fc4ce93725cb89e8aefdfddc79b247c8bd/src/analysis/retail/monk/mistweaver/Guide.tsx#L29

trevorm4 (changes requested):

trevorm4 (approved):

PR #7128, opened 2024-10-08
Resonant Words guide subsection

xizbow1:
…n. Does not work.

### Description

Added analysis of Resonant Words usage on Holy Priest (in essence, don't use it without a different buff). Currently doesn't work. I am posting this PR now so I can get help with my stupid mistake.

### Testing

- Test report URL: `https://www.warcraftlogs.com/reports/pPv2TdnbJtYCW749#fight=15&type=healing&source=59`
- Screenshot(s):


Liavre:
While heal is the best cast, flash heal is an okay cast. You should add logic for PoH (bad) and circle of healing (okay but sometimes the best cast in other tiers) or if the player overcapped RW charges (REALLY bad, can check for HW casts if player has resonant word stacks)
image

PR #7132, opened 2024-10-12
correct PoH and a few other things in the guide

Liavre:
Changed things in the guide to reflect current gameplay better, as I often get DM'd about the guide being wrong.


Liavre:
This is what the rendered output looks like:
image

emallson (approved):

PR #7136, opened 2024-10-12
Fix resto shaman ability cooldown times

frankyi-gh:
### Description
* apply Totemic Surge CDR to all totem spells
* ascendance cooldown reduced to 2m if first ascendant is talented
* HTT cooldown reduced to 2m15s if current control is talented
* totemic recall cooldown reduced to 2m if call of the elements is talented
* fix graceful spirit granting 60s CDR instead of 30s

### Testing

HTT and ascendance CD bars should not overlap anymore. IDK how to test the other changes but they are very simple
- Test report URL: `/report/m6zYGgcaNCT1PtRD/12-Heroic+Broodtwister+Ovi'nax+-+Kill+(7:18)/Pandagoesbaa/standard/overview`
- Screenshot(s): 
image

### DIFF
--- a/src/CONTRIBUTORS.ts
+++ b/src/CONTRIBUTORS.ts
@@ -2517,3 +2517,15 @@ export const SebShady: Contributor = {
   github: 'SebShady42',
   discord: 'SebShady',
 };
+
+export const PandaGoesBaa: Contributor = {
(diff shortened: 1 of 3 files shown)


emallson (approved):
thanks!

PR #7139, opened 2024-10-14
Add Aspect of Harmony guide for MW

trevorm4:
### Description

- Adds base analyzer class that can be used for both mw/brm
- Adds guide section for Aspect of Harmony for mw

Overhealing check might be a bit strict atm, open to changing thresholds or removing it

### Testing
image
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
@@ -36,7 +36,8 @@ class AspectOfHarmony extends AspectOfHarmonyBaseAnalyzer {
       actual: pctRaid,
       isGreaterThanOrEqual: targetPerfRange,
     });
-    const pctOverheal = info.totalHealing / (info.totalHealing + info.overhealing);
+    const pctOverheal =
+      info.overhealing > 0 ? info.overhealing / (info.totalHealing + info.overhealing) : 0;
     const overhealPerf = evaluateQualitativePerformanceByThreshold({
       actual: pctOverheal,
       isLessThanOrEqual: overhealingPerfRange,
--- a/src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
@@ -1,0 +1,98 @@
+import AspectOfHarmonyBaseAnalyzer, {
+  CastInfo,
+} from 'analysis/retail/monk/shared/hero/MasterOfHarmony/talents/AspectOfHarmonyBase';
+import { formatPercentage } from 'common/format';
+import { TALENTS_MONK } from 'common/TALENTS';
+import { PerformanceMark } from 'interface/guide';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
+import { RoundedPanel } from 'interface/guide/components/GuideDivs';
+import { BoxRowEntry, PerformanceBoxRow } from 'interface/guide/components/PerformanceBoxRow';
+import SpellLink from 'interface/SpellLink';
+import {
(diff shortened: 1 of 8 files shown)


src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
```diff
@@ -0,0 +1,99 @@
...
+const targetPerfRange: QualitativePerformanceThresholdRange = {
+  perfect: 1,
+  good: 0.9,
+  ok: 0.8,
+  fail: 0.7,
+};
+
+// maybe be more strict in the future once better logs emerge
+const overhealingPerfRange: QualitativePerformanceThresholdRange = {
+  good: 0.6,
```
vohrr, line 27: These are good for now

src/analysis/retail/monk/shared/hero/MasterOfHarmony/talents/AspectOfHarmonyBase.tsx
```diff
@@ -0,0 +1,129 @@
...
+          entry.numBuffs += 1;
+        }
+      } else {
+        entry.numDebuffs += 1;
+      }
+    }
+  }
+
+  onHeal(event: HealEvent) {
+    // coalesence doesn't buff Aspect HOT
```
vohrr: I don't see a reason why we would want to ignore the direct healing/damage from the hot/dot when totaling a cast entry. Is this a copy/paste mishap from Coalesce.tsx?

src/analysis/retail/monk/shared/hero/MasterOfHarmony/talents/AspectOfHarmonyBase.tsx
```diff
@@ -0,0 +1,129 @@
...
+    );
+    this.addEventListener(
+      Events.applydebuff.by(SELECTED_PLAYER).spell(SPELLS.ASPECT_OF_HARMONY_DOT),
+      this.onPeriodicApply,
+    );
+    this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.onHeal);
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDmg);
+  }
+
+  initEntry(
```
vohrr: small nit but the name initEntry was a bit confusing at first since it is acting as both a 'get latest' and 'set'

vohrr (changes requested):

src/analysis/retail/monk/shared/hero/MasterOfHarmony/talents/AspectOfHarmonyBase.tsx
trevorm4: It was essentially to measure the overheal of coalesence to make sure that you weren't wasting the buff and the player was forcing overhealing onto a target solely for coalesence, but open to removing it if you think excluding it is unnecessary.

vohrr: I think the direct healing is significant enough that this should either account for both or we should account for both in some way. High overheal on the hot can be an indicator that you're not spreading AoH well

src/analysis/retail/monk/mistweaver/modules/heroTalents/AspectOfHarmony.tsx
```diff
@@ -0,0 +1,99 @@
...
+      isGreaterThanOrEqual: targetPerfRange,
+    });
+    const pctOverheal =
+      info.overhealing > 0 ? info.overhealing / (info.totalHealing + info.overhealing) : 0;
+    const overhealPerf = evaluateQualitativePerformanceByThreshold({
+      actual: pctOverheal,
+      isLessThanOrEqual: overhealingPerfRange,
+    });
+    const value = getAveragePerf([targetPerf, overhealPerf]);
+    const tooltip = (
```
vohrr, line 46: Can we throw the time stamp of the cast event in here as well

PR #7142, opened 2024-10-17
fix: Lightweaver heals not consuming buff showing as bad cast.

fel1n3:
### Description

There are often cases when heals buffed by Lightweaver do not consume the buff, but still buff the spell (i.e. cast speed and healing) with the current approach via the normalizer these heals are not caught and are shown to the player as a bad cast. I'm not sure if simply checking for the buff is a bad approach versus how it's currently being done.

example of bug

### Testing

In-game bug can be recreated by spamming Flash Heal > Heal until the buff simply isn't consumed.
- Test report URL: `reports/Fb1P32yvHGnramjK#fight=23`
- Screenshot(s):

before
after


src/analysis/retail/priest/holy/modules/talents/BottomRow/Lightweaver.tsx
```diff
@@ -9,7 +9,7 @@ import ItemHealingDone from 'parser/ui/ItemHealingDone';
 import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
 import { SpellLink } from 'interface';
 import { calculateEffectiveHealing, calculateOverhealing } from 'parser/core/EventCalculateLib';
-import { isCastBuffedByLightweaver, getHeal } from '../../../normalizers/CastLinkNormalizer';
```
trevorm4, line 12: Can you clean this function up and the associated linking logic since its no longer needed

trevorm4 (changes requested):

PR #7162, opened 2024-10-28
feat(warrior/protection): Added TWW1 2P Season 1 tier set effect.

Rzial:
### Description

Shield Slam skill events are processed way better if we use the 2P set buff as complementary element for resets.

### Testing

- Test report URL: /report/9qDCZjkbJc3PGABV/49-Mythic+Ulgrax+the+Devourer+-+Kill+(6:40)/Сарлет/standard/debug
- Screenshot(s):
image


emallson (approved):
nice and simple :)

PR #7163, opened 2024-10-28
feat(warrior/protection): Added Burst of Power talent as a reset trigger for Shield Slam

Rzial:
### Description

Shield Slam skill events are processed better in Mountain Thane using the Burst of Power buff as complementary element for resets.

### Testing

- Test report URL: /report/XrA3BxZH2FdWRQm9/15-Mythic+Sikran,+Captain+of+the+Sureki+-+Kill+(4:24)/Adian/standard/debug.`
- Screenshot(s):
image


src/analysis/retail/warrior/protection/modules/talents/BurstOfPower.tsx
```diff
@@ -0,0 +1,25 @@
...
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS.BURST_OF_POWER_TALENT);
+
+    this.addEventListener(Events.applybuff.spell(SPELLS.BURST_OF_POWER_BUFF), this.resetShieldSlam);
+
+    this.addEventListener(
+      Events.removebuffstack.spell(SPELLS.BURST_OF_POWER_BUFF),
+      this.resetShieldSlam,
+    );
```
emallson, line 19: should this include `removebuff`? aka is it "Buff application resets cooldown, then the next cast has no cooldown." or "each cast with the buff active does not trigger the cooldown, but there is no reset on buff apply"? (or some hybrid?)

this is one of those cases where the talent tooltip doesn't tell the full story and it is worth documenting the expected behavior here in brief in case it changes later

PR #7168, opened 2024-10-30
[Shadow]  Add Support and Statistics for Shadow Archon Hero Talent Tree

DoxAshe:
### Description

This adds statistics for Shadow's Archon Hero Talent Tree.  That statistics added are for: 
-Perfected Form
-Resonant Energy
-Manifested Power
-Empowered Surges
-Energy Compression
-Sustained Potency
-Concentrated Infusion
-Energy Cycle

Most of these are simple statistics, but there are a few of them that were more complicated.  Manifested Power required changing how Surges of Insanity are tracked, Sustained Potency required changes to Void Form extensions, and Concentrated Infusion grants additional haste during Power Infusion.

### Testing
I've added a lot, and I would appreciate a second set of eyes just to make sure I haven't made any glaring mistakes for the statistics.  In addition, there were a couple talents that are a little tricky.

Concentrated Infusion gives an additional 10% haste from only your own Power Infusion.  I think `Events.applybuff.by(SELECTED_PLAYER).spell(TALENTS.POWER_INFUSION_TALENT)` is only the player's power infusion, but if I am mistaken on that then any PI would grant the additional haste.

Sustained Potency extends Void Form, but it shouldn't be included in the Void Form extension metric.  When doing this, I found a strange problem where the buffer in `this.selectedCombatant.getBuffStacks()` would not work.  I was able to work around this, but couldn't figure out why.

Manifested Power grants Surge of Insanity for every Halo cast.  Since Surge of Insanity does not create a refresh buff event, checking to see if there was any wasted procs is difficult.  I am pretty confident I have it working for any conditions, but it feels like there should have been a more straightforward way to do this.  

- Test report URL: `report/gdBvxajDfACLJNPT/1-Normal+Ulgrax+the+Devourer+-+Kill+(3:17)/Doxashe/standard`
- Screenshot(s):
Archon Hero Talents


src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
```diff
@@ -60,10 +96,10 @@ class Voidform extends Analyzer {
       </>
     );
     let value = QualitativePerformance.Good;
-    if (extension <= 15) {
+    if (extension <= 15.1) {
       value = QualitativePerformance.Ok;
     }
-    if (extension <= 10) {
+    if (extension <= 10.1) {
```
emallson: whats the story here?

src/analysis/retail/priest/shadow/modules/talents/Archon/Concentrated Infusion.tsx
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+  protected haste!: Haste;
+
+  constructor(options: Options) {
+    super(options);
+
+    this.active = this.selectedCombatant.hasTalent(TALENTS.CONCENTRATED_INFUSION_TALENT);
+
+    this.addEventListener(
```
emallson, line 21: making a note that pre-TWW this would not have worked right because refreshes of PI didn't log a new PI buff. however, they have fixed that so this should be correct for all TWW logs (unless they un-fix it :upside_down_face:)

src/analysis/retail/priest/shadow/modules/talents/Archon/EnergyCompression.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onHalo(event: DamageEvent) {
+    this.damageHalo += event.amount;
+
+    //Damage from Energy Compression
+    if (this.selectedCombatant.hasTalent(TALENTS.ENERGY_COMPRESSION_TALENT)) {
+      this.damageEnergyCompression += calculateEffectiveDamage(
+        event,
+        ARCHON_ENERGY_COMPRESSION_MULTIPLIER,
+      );
+    }
```
emallson, line 47: this event handler only runs if `this.active` is true, so you don't need to check `hasTalent` again here.
```suggestion
    //Damage from Energy Compression
    this.damageEnergyCompression += calculateEffectiveDamage(
      event,
      ARCHON_ENERGY_COMPRESSION_MULTIPLIER,
    );
```

src/analysis/retail/priest/shadow/modules/talents/Archon/EnergyCompression.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <Statistic size="flexible" category={STATISTIC_CATEGORY.HERO_TALENTS}>
+        <BoringSpellValueText spell={TALENTS.HALO_SHADOW_TALENT}>
+          <ItemDamageDone amount={this.damageHalo} />
+        </BoringSpellValueText>
+        {this.selectedCombatant.hasTalent(TALENTS.ENERGY_COMPRESSION_TALENT) && (
+          <BoringSpellValueText spell={TALENTS.ENERGY_COMPRESSION_TALENT}>
+            <ItemDamageDone amount={this.damageEnergyCompression} />
+          </BoringSpellValueText>
+        )}
```
emallson, line 60: similar here. the statistic is not shown if `this.active` is false, so this can be simplified to:

```suggestion
        <BoringSpellValueText spell={TALENTS.ENERGY_COMPRESSION_TALENT}>
          <ItemDamageDone amount={this.damageEnergyCompression} />
        </BoringSpellValueText>
```

src/analysis/retail/priest/shadow/modules/talents/Archon/ResonantEnergy.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  checkAmpDamage(event: DamageEvent) {
+    //Damage from Resonant Energy
+    const target = this.enemies.getEntity(event);
+    if (!target) {
+      return;
+    }
+    const ampRE =
+      target.getBuffStacks(SPELLS.SHADOW_PRIEST_ARCHON_RESONANT_ENERGY_DEBUFF.id) *
+      ARCHON_RESONANT_ENERGY_MULTIPLIER;
```
emallson, line 44: this is probably accidentally quadratic, but it might be okay if Resonant Energy doesn't trigger very often on each target. if you encounter performance issues, this would be something to potentially look at.

src/analysis/retail/priest/shadow/modules/talents/Archon/SustainedPotency.tsx
```diff
@@ -0,0 +1,80 @@
...
+    enemies: Enemies,
+  };
+
+  protected enemies!: Enemies;
+
+  durationSustainedPotency = 0;
+
+  //When DA/VF is active, casting Halo extends its duration by 1 second.
+  //When DA/VF is not active, casting Halo gives a buff, that causes the next DA/VF to be longer.
+  //Every 1 Halo cast causes 2 additional Halos that aren't casts.
```
emallson, line 25: amazing

src/analysis/retail/priest/shadow/modules/talents/Archon/SustainedPotency.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onCD(event: CastEvent) {
+    const buffer = 50;
+    //For some reason, the buffer does not work properly?
+    //So we look 10ms backwards to see how many stacks we had at that time.
+    //Usually, the buff falls off 1-3ms after the voidform cast finishes, but 1-3ms before the voidform buff starts.
+    this.durationSustainedPotency += this.selectedCombatant.getBuffStacks(
+      SPELLS.SHADOW_PRIEST_ARCHON_SUSTAINED_POTENCY_BUFF.id,
+      event.timestamp - 10,
+      buffer,
+    );
```
emallson, line 57: i don't have an explanation of this behavior. would need to dig into the details of it.

emallson (approved):
this looks good overall. a couple of minor things that you can fix or leave in.

src/analysis/retail/priest/shadow/modules/spells/Voidform.tsx
DoxAshe: I had just been testing some other values while looking at logs, and didn't change them back properly.

src/analysis/retail/priest/shadow/modules/talents/Archon/EnergyCompression.tsx
DoxAshe: Good catch, I've changed it to so that the event handler always runs, because halo is always talented even with the other hero choice node.

src/analysis/retail/priest/shadow/modules/talents/Archon/ResonantEnergy.tsx
DoxAshe: I haven't noticed any performance issues even in some keys, but I've noted this potential issue in the code in case it does become a problem.  If there is another way to calculate this damage amp that doesn't have that potential issue I can change it before it becomes a problem.

src/analysis/retail/priest/shadow/modules/talents/Archon/SustainedPotency.tsx
DoxAshe: I could not determine the problem either, but at least this current work around works.

src/analysis/retail/priest/shadow/modules/talents/Archon/Concentrated Infusion.tsx
DoxAshe: I hadn't even considered a PI refresh being a problem, thank goodness that's fixed (for now).

PR #7171, opened 2024-11-02
Improve Active Time estimation for Arms Warrior in Cataclysm

emallson:
### Description

In Cata, Arms is a *Downtime Spec.* There are not enough abilities or resources to fill every GCD, so some are unavoidably left empty.

Currently, the Guide explicitly calls this out and reduces the Active Time targets to compensate (based on, honestly, rough guesswork). This PR adjusts the Active Time calculation to count the unavoidable downtime as "uptime", the same way that other sources of unavoidable downtime (like GCDs, or casting) are counted.

To do that, it uses an APL to track what the expected cast is at each event. If there is no expected cast, then we treat it as downtime. (The method to do this might be too magical / have bad defaults and need adjusting).

In the process, I started refactoring some of the APL code to allow using APLs as checkers within Analyzers.

Additionally, I did some archaeology on Haste & the GCD in Classic, and I've become pretty confident that **haste does not impact the melee GCD.** As a result, I changed the Arms GCDs to be static at 1.5 seconds.

Specifically, classic has 3 types of haste: Melee, Ranged, and Spell. In patch 2.4, Spell Haste was changed to impact the GCD. Melee/Ranged haste were not. The wiki indicates that this was the status quo until WoD prepatch, which unified all 3 varieties of haste into a single "Haste" stat. Prior to this, melee abilities would not benefit from Spell Haste. 

### Testing

*Sample Report*

`/report/zHX3VMN6YAPRFTKg/32-Heroic+Chimaeron+-+Kill+(3:12)/Dolbob/standard`

Before
image

After
image

*Ahlaundoh (Current R1)*

`/report/8ZndcNyGkjTmMxYB/29-Heroic+Chimaeron+-+Kill+(2:56)/25-Ahlaundoh/standard`

Before
image

After
image

*Random Recent Log*

`/report/BDzNRdqAba9M6cCP/10-Heroic+Chimaeron+-+Kill+(3:45)/돌진후저세상/standard`

Before
image

After
image

worth noting that on each of these logs, all of the GCDs still pass:
image

### DIFF
--- a/src/common/SPELLS/classic/warrior.ts
+++ b/src/common/SPELLS/classic/warrior.ts
@@ -118,7 +118,7 @@ const spells = {
   BATTLE_TRANCE: {
     id: 12964,
     name: 'Battle Trance',
-    icon: 'inv_helmet_06',
+    icon: 'inv_helmet_06.jpg',
   },
   PUMMEL: {
     id: 6552,
--- a/src/common/SPELLS/classic/warrior.ts
+++ b/src/common/SPELLS/classic/warrior.ts
@@ -115,6 +115,11 @@ const spells = {
     name: 'Taste for Blood',
     icon: 'ability_rogue_hungerforblood.jpg.jpg',
   },
+  BATTLE_TRANCE: {
+    id: 12964,
+    name: 'Battle Trance',
+    icon: 'inv_helmet_06',
+  },
   PUMMEL: {
     id: 6552,
     name: 'Pummel',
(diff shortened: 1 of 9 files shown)


src/common/SPELLS/classic/warrior.ts
```diff
@@ -115,6 +115,11 @@ const spells = {
     name: 'Taste for Blood',
     icon: 'ability_rogue_hungerforblood.jpg.jpg',
   },
+  BATTLE_TRANCE: {
+    id: 12964,
+    name: 'Battle Trance',
+    icon: 'inv_helmet_06',
```
jazminite:
```suggestion
    icon: 'inv_helmet_06.jpg',
```

jazminite (approved):
1 small change, otherwise lgtm

PR #7174, opened 2024-11-09
Added Two Trinkets to TWW (Mad Queens Mandate & Ovinax Mercurial Egg)

Mcheung7272:
### Description

I'm working on getting the core of Hunter Marksmanship up and running again. As I was going through the spellDebug, I saw the spells for Mad Queen's Mandate and Ovinax's Egg not being recognized. Decided to break those additions off into this PR as it would benefit others as well.

Caveat on the Egg logic - There's currently no way to check what the split for the egg is as it isn't being logged.

### Testing

If you test any report that use either trinket, currently it should be red and unrecognized. With the additions of these files + adding them to your specs CombatLogParser, it should be green and recognized. 

- Test report URL: `/report/AqYvnpjk2MCHmgca/21-Mythic+The+Bloodbound+Horror+-+Kill+(5:10)/Lùck/standard/debug`
- Screenshot(s):

image

### DIFF
--- a/src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
+++ b/src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
@@ -31,7 +31,7 @@ export default class MadQueensMandate extends Analyzer.withDependencies({
 
   private onDamage(event: DamageEvent) {
     if (event.overkill && event.overkill > 0) {
-      this.spellUsable.reduceCooldown(SPELLS.ABYSSAL_GLUTTONY.id, 60);
+      this.spellUsable.reduceCooldown(SPELLS.ABYSSAL_GLUTTONY.id, 60000);
     }
   }
 }
--- a/src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
+++ b/src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
@@ -1,0 +1,24 @@
+import ITEMS from 'common/ITEMS/thewarwithin/trinkets';
+import SPELLS from 'common/SPELLS/thewarwithin/trinkets';
+import Analyzer, { Options } from 'parser/core/Analyzer';
+import SPELL_CATEGORY from 'parser/core/SPELL_CATEGORY';
+import Abilities from 'parser/core/modules/Abilities';
+
+export default class MadQueensMandate extends Analyzer.withDependencies({
+  abilities: Abilities,
+}) {
+  constructor(options: Options) {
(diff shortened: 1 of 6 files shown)


Mcheung7272:
Just an update. I tried looking into getting the stacks for the egg buff and I'm not seeing any fields where I could figure out what the exact split is between two stats.

emallson:
Egg doesn't log stacks, unfortunately.

src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (!this.active) {
+      return;
+    }
+
+    this.deps.abilities.add({
+      spell: SPELLS.ABYSSAL_GLUTTONY.id,
+      category: SPELL_CATEGORY.COOLDOWNS,
+      cooldown: 120,
+    });
+  }
```
emallson, line 36: if you want to add the death logic for MQM, the way i'd do it is:

1. add `SpellUsable` as a dependency.
2. add a `Events.damage` handler to the module
3. if the damage event has overkill, reduce the cooldown by 1 minute. this will sometimes be wrong (some enemies don't actually die), but it will be right way, way more than it is wrong.

the handler would look something like:

```ts
private onDamage(event: DamageEvent) {
  if (event.overkill && event.overkill > 0) {
    this.spellUsable.reduceCooldown(SPELLS.ABYSSAL_GLUTTONY.id, 60);
  }
}
```

if you did something like this in the constructor:

```ts
this.addEventListener(Events.damage.spell(SPELLS.ABYSSAL_GLUTTONY), this.onDamage);
```

emallson (commented):
this overall looks fine. if you want to add some handling for the cd reduction on MQM we can get this merged. otherwise i'll probably add some tomorrow and merge

src/parser/retail/modules/items/thewarwithin/trinkets/MadQueensMandate.ts
Mcheung7272: Thanks for the help. Added the cd logic. Lmk if you need anything else to get this along

emallson: (i actually got the scale wrong, it should be 60000, not 60 since it is in milliseconds. but i will patch that in the morning and merge)

Mcheung7272: good catch. updated.

emallson (approved):

PR #7177, opened 2024-11-12
[Shadow] Add Basic Support for New Spells Added by Voidweaver Hero Talents.

DoxAshe:
### Description

Void weaver adds 3 new spells to shadow.  Entropic Rift, which is created at the end of void torrent; Void Blast, which replaces mind blast while Entropic Rift exists; and Voidwraith, which replaces mindbender/shadowfiend.  
This adds these spells to shadow's abilities, and tracks them in the guide view, and enables them when the proper talents are selected.

There is a lot more to do for these spells.  For Void Blast, their are some cases for cast efficiency where some time is counted as off cooldown that shouldn't be, and these Void Blasts should also count as Mind Blasts for that cast efficiency.  There are also several other voidweaver talents that are not yet implemented, but those should only be simple statistics.  

### Testing

I am going away for a couple weeks, so I wanted to get some basic functionality up for voidweaver before I left.  I was rushing a little bit, but don't think this causes any problems for other talent selections, and only adds improvements even if its not perfect.

- Test report URL: `/report/fVwz7GWd9y4jBc1g/15-Heroic+Sikran,+Captain+of+the+Sureki+-+Kill+(2:16)/Krisix/standard)`
- Test report URL: `report/gdBvxajDfACLJNPT/2-Normal+The+Bloodbound+Horror+-+Kill+(2:58)/Doxashe/standard`
- 
- Screenshots:
Void Blast
nshot(s):
Rift
voidwraith


src/analysis/retail/priest/shadow/modules/guide/CooldownGraphSubsection.tsx
```diff
@@ -117,13 +156,26 @@ const CoreCooldownsGraph = () => {
     </p>
   );
 
+  if (info!.combatant.hasTalent(TALENTS.VOID_BLAST_TALENT)) {
```
emallson, line 159: quick note since i'm seeing a lot of the `info!` assertion. you can put something like this at the top (after all of the hooks) and then typescript will know that `info` isn't null/undefined:

```ts
if (!info) {
  return null;
}
```

`info` is only ever null/undefined when stuff is still loading. in practice i don't think it will *ever* be null/undefined for a component like this.

(you don't need to fix that here---but future note)

emallson (approved):

PR #7178, opened 2024-11-14
Updated Hunter MM Tier Set + Ran Update Script

Mcheung7272:
### Description

1. Ran update script to fetch most current spells/talents for Live WOW Build 11.0.5.57388.
2. Updated tier set to match the current tier set.
3. Added some dps information regarding the tier set effects (2P/4P)

### Testing

Run analysis on any Hunter MM Logs

- Test report URL: `/report/AqYvnpjk2MCHmgca/21-Mythic+The+Bloodbound+Horror+-+Kill+(5:10)/Yellot/standard/statistics`
- Screenshot(s):

image

### Additional Comments

I rebased my main and it looks like this PR still has commits from the prior PR. Not sure why.

I noticed the previous naming for the tier set was T29MMTier2P. Where exactly is this naming convention from? The name I've given is more of a guess. Also feel free to double check my tier set dps logic.. makes sense to me but I could've missed something obvious.


Mcheung7272:
Looks like running the update script does break typing for other classes. Should I only include the hunter updates?

emallson:
> Should I only include the hunter updates?

for now, yes. ill get a full update in soon:tm: and go through updating other specs / pinging people that need to do extra work

Mcheung7272:
Ok, only kept hunter common changes .

emallson (approved):

PR #7180, opened 2024-11-15
Update/Add CDR logic

Mcheung7272:
### Description

1. Updated/Added some of the CDR logic to fix errors showing up in spellUsable debug menu.
2. Simplified cdr logic in one place

### Testing

Analyze any MM logs.

- Test report URL: `/report/ZCjp1DgdVfbxnRBK/26-Mythic+The+Bloodbound+Horror+-+Kill+(3:23)/Pbpbman/standard/debug`
- Screenshot(s):

BEFORE: 

image

AFTER: 

image

### DIFF
--- a/src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
+++ b/src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
@@ -5,6 +5,11 @@ import TALENTS from 'common/TALENTS/hunter';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 export default [
+  change(date(2024, 11, 15), <>Added reset logic for <SpellLink spell={SPELLS.RAPID_FIRE} /> and <SpellLink spell={TALENTS.AIMED_SHOT_TALENT} /> from <SpellLink spell={SPELLS.WAILING_ARROW_DAMAGE} /> and <SpellLink spell={TALENTS.READINESS_TALENT} /> </>, Yellot),
+  change(date(2024, 11, 15), <>Added <SpellLink spell={SPELLS.RAPID_FIRE}/> reset logic from <SpellLink spell={TALENTS.SURGING_SHOTS_TALENT} /></>, Yellot),
+  change(date(2024, 11, 15), <>Simplified reset logic inside of <SpellLink spell={TALENTS.LOCK_AND_LOAD_TALENT} /></>, Yellot),
+  change(date(2024, 11, 15), <>Inclusion of <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT} /> and <SpellLink spell={SPELLS.STEADY_FOCUS_BUFF} /> for CDR logic in Aimed Shot</>, Yellot),
+  change(date(2024, 11, 15), <>Add <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT}/> constant for CDR Reduction</>, Yellot),
   change(date(2024, 11, 13), 'Updated tier set', Yellot),
   change(date(2024, 11, 10), 'Enable Marksmanship again for usage', Yellot),
   change(date(2024, 11, 10), <>Added <SpellLink spell={TALENTS.BLACK_ARROW_TALENT} />, <SpellLink spell={TALENTS.WAILING_ARROW_TALENT} /> and <SpellLink spell={SPELLS.FORTITUDE_OF_THE_BEAR} /></>, Yellot),
--- a/src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
+++ b/src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
@@ -5,11 +5,15 @@ import TALENTS from 'common/TALENTS/hunter';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 export default [
-  change(date(2024, 11, 15), <>Added reset logic for <SpellLink spell={SPELLS.RAPID_FIRE} /> and <SpellLink spell={TALENTS.AIMED_SHOT_TALENT} /> from <SpellLink spell={SPELLS.WAILING_ARROW_DAMAGE} /> and <SpellLink spell={TALENTS.READINESS_TALENT} /> </>, Yellot),
-  change(date(2024, 11, 15), <>Added <SpellLink spell={SPELLS.RAPID_FIRE}/> reset logic from <SpellLink spell={TALENTS.SURGING_SHOTS_TALENT} /></>, Yellot),
-  change(date(2024, 11, 15), <>Simplified reset logic inside of <SpellLink spell={TALENTS.LOCK_AND_LOAD_TALENT} /></>, Yellot),
-  change(date(2024, 11, 15), <>Inclusion of <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT} /> and <SpellLink spell={SPELLS.STEADY_FOCUS_BUFF} /> for CDR logic in Aimed Shot</>, Yellot),
-  change(date(2024, 11, 15), <>Add <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT}/> constant for CDR Reduction</>, Yellot),
+  change(date(2024, 11, 15), <>Added reset logic for <SpellLink spell={SPELLS.RAPID_FIRE} /> and <SpellLink spell={TALENTS.AIMED_SHOT_TALENT} /> 
+  from <SpellLink spell={SPELLS.WAILING_ARROW_DAMAGE} /> and <SpellLink spell={TALENTS.READINESS_TALENT} /> 
+  <br/>
+  Simplified reset logic inside of <SpellLink spell={TALENTS.LOCK_AND_LOAD_TALENT} />
+  <br/>
+  Inclusion of <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT} /> and <SpellLink spell={SPELLS.STEADY_FOCUS_BUFF} /> for CDR logic in Aimed Shot
+  <br/>
+  Add <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT}/> constant for CDR Reduction
+  </>, Yellot),
   change(date(2024, 11, 13), 'Updated tier set', Yellot),
   change(date(2024, 11, 10), 'Enable Marksmanship again for usage', Yellot),
   change(date(2024, 11, 10), <>Added <SpellLink spell={TALENTS.BLACK_ARROW_TALENT} />, <SpellLink spell={TALENTS.WAILING_ARROW_TALENT} /> and <SpellLink spell={SPELLS.FORTITUDE_OF_THE_BEAR} /></>, Yellot),
--- a/src/analysis/retail/hunter/marksmanship/modules/Abilities.tsx
+++ b/src/analysis/retail/hunter/marksmanship/modules/Abilities.tsx
@@ -24,7 +24,7 @@ class Abilities extends CoreAbilities {
         category: SPELL_CATEGORY.ROTATIONAL,
         cooldown: (haste: number) => {
           if (this.selectedCombatant.hasTalent(TALENTS.TACTICAL_RELOAD_TALENT)) {
-            return 12 / (1 + haste + TACTICAL_RELOAD_CDR_REDUCTION);
+            return (12 * (1 - TACTICAL_RELOAD_CDR_REDUCTION)) / (1 + haste);
           }
           return 12 / (1 + haste);
         },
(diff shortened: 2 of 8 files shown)


Mcheung7272:
Still trying to figure out the remainder ones left.... pr in the future

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
```diff
@@ -77,9 +88,23 @@ class AimedShot extends Analyzer {
...
     let modRate = 1;
+
+    // Sum the total cdr reduction from Trueshot, Tactial Reload, Steady Focus
+    let reduction = 0;
     if (this.selectedCombatant.hasBuff(SPELLS.TRUESHOT.id)) {
-      modRate /= 1 + TRUESHOT_AIMED_SHOT_RECHARGE_INCREASE;
+      reduction += TRUESHOT_AIMED_SHOT_RECHARGE_INCREASE;
+    }
+
+    if (this.selectedCombatant.hasTalent(TALENTS_HUNTER.TACTICAL_RELOAD_TALENT)) {
```
Pewtro: Since this talent statically reduces the cooldown - you can just do handle it in the abilities file. You should be able to find many examples of spells being reduced with X talent in there.

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
```diff
@@ -77,9 +88,23 @@ class AimedShot extends Analyzer {
...
+      reduction += TACTICAL_RELOAD_CDR_REDUCTION;
+    }
+
+    if (this.selectedCombatant.hasBuff(SPELLS.STEADY_FOCUS_BUFF.id)) {
+      reduction += STEADY_FOCUS_HASTE_PERCENT[1];
```
Pewtro: You can just add the steady focus haste to the haste file and it will automatically be handled in the abilities file since it's already hasted cooldown in there.

src/analysis/retail/hunter/marksmanship/modules/talents/LockAndLoad.tsx
```diff
@@ -77,15 +77,7 @@ class LockAndLoad extends Analyzer {
...
-      const expectedCooldownDuration = this.abilities.getExpectedCooldownDuration(
-        TALENTS_HUNTER.AIMED_SHOT_TALENT.id,
-      );
-      if (expectedCooldownDuration) {
-        const newChargeCDR =
-          expectedCooldownDuration -
-          this.spellUsable.cooldownRemaining(TALENTS_HUNTER.AIMED_SHOT_TALENT.id);
-        this.spellUsable.reduceCooldown(TALENTS_HUNTER.AIMED_SHOT_TALENT.id, newChargeCDR);
-      }
+      this.spellUsable.endCooldown(TALENTS_HUNTER.AIMED_SHOT_TALENT.id, event.timestamp);
```
Pewtro, line 80: Why is this being removed? Did functionality of LnL change?

src/analysis/retail/hunter/shared/talents/WailingArrow.tsx
```diff
@@ -25,12 +32,37 @@ class WailingArrow extends Analyzer {
...
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS_HUNTER.AIMED_SHOT_TALENT),
+      this.onAimedShot,
+    );
+  }
+
+  //Readiness grants Wailing Arrow the ability to reset the cooldown of Rapid Fire and 2 charges of Aimed Shot
+  onWailingCast(event: CastEvent) {
+    if (this.selectedCombatant.hasTalent(TALENTS.READINESS_TALENT)) {
+      this.spellUsable.endCooldown(SPELLS.RAPID_FIRE.id, event.timestamp);
+      this.chargesLeft = 2 - this.spellUsable.chargesAvailable(TALENTS_HUNTER.AIMED_SHOT_TALENT.id);
```
Pewtro: .endCooldown has a boolean property for restoring all charges, wouldn't that be better in this case?

src/analysis/retail/hunter/marksmanship/modules/talents/LockAndLoad.tsx
Mcheung7272: Unless I'm misunderstanding, isnt the logic here just removing/resetting the cooldown? 

Current logic - 
1. Get expected cooldown for aimed shot for current timestamp
2. If valid value, get the remaining cd
3. Reduce cooldown of aimed shot by the remaining cd ie resetting it

The outcome we want is to reset the cooldown of aimed shot so instead of doing the math to reduce it, i figured we just reset it and skip the math.

lmk if i missed anything tho

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
Mcheung7272: sorry, could you clarify where exactly you're suggesting to put this? I tried looking for the haste file but can't seem to find it.

src/analysis/retail/hunter/shared/talents/WailingArrow.tsx
Mcheung7272: Didn't know this was available. Thanks for pointing it out. Removed any tracking logic for residual charges and utilized the properties in endCooldown

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
Mcheung7272: I think I addressed it? Not entirely sure, feel free to take a look and lmk if thats what you meant.

I moved the Tactical reload talent cdr logic to aimed shot in abilities file.

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
Pewtro: There is a file called Haste in src/parser/shared/modules that already has Steady Focus buff defined as a 7% buff - you can update it to 8%, but shouldn't need any handling inside the Aimed Shot file.

Mcheung7272: Good call out. Thanks for pointing this out. Changed the value and reverted my changes in Aimed Shot.

Follow up question, is it safe to assume any buffs/talents that contributes a static haste % should live in this haste file?

src/analysis/retail/hunter/marksmanship/modules/talents/SurgingShots.tsx
```diff
@@ -49,8 +49,14 @@ class SurgingShots extends Analyzer {
...
 
-  onAimedShotCast() {
+  onAimedShotCast(event: CastEvent) {
     this.aimedShotCasts += 1;
+    if (
+      this.spellUsable.isOnCooldown(SPELLS.RAPID_FIRE.id) &&
+      this.spellUsable.chargesAvailable(SPELLS.RAPID_FIRE.id) === 0
+    ) {
+      this.spellUsable.endCooldown(SPELLS.RAPID_FIRE.id, event.timestamp);
+    }
```
kfinch: My reading of the Surging Shots talent is that Aimed Shot only has a small chance to reset Rapid Fire, but this implementation always resets Rapid Fire on Aimed Shot usage. I'm sure this reduced the number of errors in the debug, but it is adding a new non-debug-error problem: we'll frequently show Rapid Fire as not on cooldown when it actually is.

My guess is there is no event in logs that indicates when a proc has occured, which unfortunately means there is no foolproof way for WoWA to see every proc. The best you can do is look for "too early" Rapid Fire usages and infer that a proc occured.

src/analysis/retail/hunter/marksmanship/modules/talents/LockAndLoad.tsx
kfinch: I think this simplification is correct, but make sure to run some relevant logs through your local to sanity check that nothing changed in the module.

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
```diff
@@ -62,6 +62,13 @@ class AimedShot extends Analyzer {
     );
   }
 
+  //Pin Cushion reduces the cooldown of Aimed Shot by 2 seconds
+  onSteadyShot(event: CastEvent) {
+    if (this.selectedCombatant.hasTalent(TALENTS_HUNTER.PIN_CUSHION_TALENT)) {
+      this.spellUsable.reduceCooldown(TALENTS_HUNTER.AIMED_SHOT_TALENT.id, 2000, event.timestamp);
```
kfinch: the final field `event.timestamp` is optional and doesn't need to be specified when using the event's timestamp (that's the default)

src/analysis/retail/hunter/marksmanship/modules/Abilities.tsx
```diff
@@ -21,7 +22,12 @@ class Abilities extends CoreAbilities {
         spell: TALENTS.AIMED_SHOT_TALENT.id,
         enabled: this.selectedCombatant.hasTalent(TALENTS.AIMED_SHOT_TALENT),
         category: SPELL_CATEGORY.ROTATIONAL,
-        cooldown: (haste: number) => 12 / (1 + haste),
+        cooldown: (haste: number) => {
+          if (this.selectedCombatant.hasTalent(TALENTS.TACTICAL_RELOAD_TALENT)) {
+            return 12 / (1 + haste + TACTICAL_RELOAD_CDR_REDUCTION);
+          }
+          return 12 / (1 + haste);
+        },
```
kfinch: I don't think this is correct, because talents like this usually stack multiplicatively with haste, not additively. Also "CD reduced by 10%" in blizz speak means "multiply by 0.9", not "divide by 1.1". The correct formula should be `12 * (1 - TACTICAL_RELOAD_CDR_REDUCTION) / (1 + haste)`

src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
```diff
@@ -5,6 +5,11 @@ import TALENTS from 'common/TALENTS/hunter';
 import ITEMS from 'common/ITEMS';
 import SPELLS from 'common/SPELLS';
 export default [
+  change(date(2024, 11, 15), <>Added reset logic for <SpellLink spell={SPELLS.RAPID_FIRE} /> and <SpellLink spell={TALENTS.AIMED_SHOT_TALENT} /> from <SpellLink spell={SPELLS.WAILING_ARROW_DAMAGE} /> and <SpellLink spell={TALENTS.READINESS_TALENT} /> </>, Yellot),
+  change(date(2024, 11, 15), <>Added <SpellLink spell={SPELLS.RAPID_FIRE}/> reset logic from <SpellLink spell={TALENTS.SURGING_SHOTS_TALENT} /></>, Yellot),
+  change(date(2024, 11, 15), <>Simplified reset logic inside of <SpellLink spell={TALENTS.LOCK_AND_LOAD_TALENT} /></>, Yellot),
+  change(date(2024, 11, 15), <>Inclusion of <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT} /> and <SpellLink spell={SPELLS.STEADY_FOCUS_BUFF} /> for CDR logic in Aimed Shot</>, Yellot),
+  change(date(2024, 11, 15), <>Add <SpellLink spell={TALENTS.TACTICAL_RELOAD_TALENT}/> constant for CDR Reduction</>, Yellot),
```
kfinch: When making a lot of changes to the same category of thing, you only need to add one changelog entry

kfinch (changes requested):

src/analysis/retail/hunter/marksmanship/modules/talents/SurgingShots.tsx
Mcheung7272: I mostly followed this example [here](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/analysis/retail/hunter/beastmastery/modules/core/SpellUsable.tsx#L77). Where it looks like it takes an optimistic approach to reset (same low percentage scenario for reset). 

You're correct, there rly is no sign that surging shots triggers. I'll remove this though, didnt think about the wider impact of showing rapid fire off cd when it could be.

src/analysis/retail/hunter/marksmanship/modules/Abilities.tsx
Mcheung7272: appreciate the correct formula. Was totally a guess in the dark on my end

src/analysis/retail/hunter/marksmanship/modules/talents/AimedShot.tsx
Mcheung7272: removed.

src/analysis/retail/hunter/marksmanship/modules/talents/LockAndLoad.tsx
Mcheung7272: tested same log with current code and my simplification. I didn't see any changes on the end result.

src/analysis/retail/hunter/marksmanship/CHANGELOG.tsx
Mcheung7272: gotcha, consolidated all of em.

kfinch (approved):

PR #7190, opened 2024-11-19
Feature Survival Guides

Kivlov:
### Description
Enabling the guide module for Survival and getting core rotation abilities and explanations in. I am having trouble wiith the APL section so I have just commented it out, and plan to make a PR for it separately as I believe there is something awry with how it's tracking focus or how the APL checker checks focus. (As well as an issue with how it's checking for serpent sting).

### Testing
http://localhost:3000/report/Jb8HakY274dhCyg6/52-Mythic+Nexus-Princess+Ky'veza+-+Kill+(6:05)/Percival/standard
Log with lots of pulls on princess to check.
https://www.warcraftlogs.com/reports/akmJh3DWp1Gy47zP#fight=37&type=damage-done&source=8
heroic log for Thyminde which will have few errors.
- Screenshot(s):
image

I believe many of these red dots are because Wildfire Bomb is cast pre-pull, and only on the damage instance does it set the cooldown back to 0 and so it thinks wildfire is 1-2s longer cooldown that it is in reality. Have not implemented (SHARED) logic for Kill Command reset from Frenzied Tear in Pack Leader.


Kivlov:
and here I thought I had it somewhat figured git out but it brought in all the commits that had already been merged in I thought.

The start of the feature stuff should be from using shared deathblow forward in time.

src/analysis/retail/hunter/survival/modules/guide/sections/rotation/RotationSection.tsx
```diff
@@ -2,8 +2,9 @@ import { t, Trans } from '@lingui/macro';
 import { GuideProps, Section, SubSection } from 'interface/guide';
 import CombatLogParser from 'analysis/retail/hunter/survival/CombatLogParser';
 import TALENTS from 'common/TALENTS/hunter';
-import * as AplCheck from 'analysis/retail/hunter/survival/modules/apl/AplCheck';
-import { AplSectionData } from 'interface/guide/components/Apl';
+//Commented out because they will be reactived relatively quick.
```
Kivlov, line 5: I know these should just be deleted but I have a bad memory and didn't want to have to remember how to re-implement them once I have the APL section working as I'd like. It's like 90% there for pack leader except for an issue with Serpent Sting debuff tracking and checking for focus levels. (Both of which I feel might be outside of my skill level to resolve or track down)

src/analysis/retail/hunter/shared/HowlOfThePack.tsx
```diff
@@ -0,0 +1,211 @@
...
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE(7)}
+        size="flexible"
+        dropdown={
+          <>
+            <table className="table table-condensed">
+              <thead>
+                <tr>
+                  <th>Haste-Bonus</th>
```
emallson:
```suggestion
                  <th>Crit Damage Bonus</th>
```

right?

src/analysis/retail/hunter/shared/talents/ExplosiveShot.tsx
```diff
@@ -63,6 +65,25 @@ class ExplosiveShot extends Analyzer {
       </Statistic>
     );
   }
+  get guideSubsectionSV() {
+    const explanation = (
+      <p>
+        <strong>
+          <SpellLink spell={TALENTS.EXPLOSIVE_SHOT_TALENT} />
+        </strong>{' '}
+        deals damage after 3 second delay <SpellLink spell={TALENTS.TIP_OF_THE_SPEAR_TALENT} />.
```
emallson:
```suggestion
        deals damage after 3 second delay.
```

this looks like an accidental paste?

src/analysis/retail/hunter/shared/talents/ExplosiveShot.tsx
```diff
@@ -63,6 +65,25 @@ class ExplosiveShot extends Analyzer {
...
+        Always use it on cooldown. You may spend a{' '}
+        <SpellLink spell={TALENTS.TIP_OF_THE_SPEAR_TALENT} /> on a cast but do not delay a cast
+        specifically to tip it. Cast prior to using{' '}
+        <SpellLink spell={TALENTS.COORDINATED_ASSAULT_TALENT} /> as Explosive Shot does not snapshot
+        it's damage on cast.
```
emallson:
```suggestion
        its damage on cast.
```
grammar

src/analysis/retail/hunter/survival/CombatLogParser.ts
```diff
@@ -91,6 +106,10 @@ class CombatLogParser extends CoreCombatLogParser {
     rejuvenatingWind: RejuvenatingWind,
     trailblazer: Trailblazer,
     tranquilizingShot: TranquilizingShot,
+    SurvivalOfTheFittest: SurvivalOfTheFittest,
+    //Items
+    madQueensMandate: MadQueensMandate,
+    skardynsGrace: SkardynsGrace,
```
emallson: these should probably be added in the core `CombatLogParser` (the one that this class extends; if they aren't already) instead, since any spec can equip these items.

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,196 @@
...
+      (tense) => {
+        //const isSerpentStingMissing = cnd.debuffMissing(SPELLS.SERPENT_STING_SURVIVAL);
+        const isHowlOfThePackFading = cnd.not(
+          cnd.buffRemaining(SPELLS.HOWL_OF_THE_PACK_BUFF, 8, { atLeast: 1400 }, false),
+        );
+
+        // if (isSerpentStingMissing) {
+        //   return <>Serpent Sting is missing</>;
+        // } else
+        if (isHowlOfThePackFading) {
```
emallson: this is not going to work like you want. you need two separate rules for this. the return value of `cnd.not(...)` is a *condition object* which is used by the checker to evaluate the condition. it is not the result of evaluating the condition.

as a result, this `if` branch will *always* be taken since a non-null object is truthy

in this case, it looks like you refactored the rule to only include the "howl of the pack is fading" case, in which case you can remove most of this and just use `<>Howl of the Pack {tenseAlt(tense, 'is', 'was')} about to fade</>`

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,196 @@
...
+export const apl = build([
+  {
+    spell: TALENTS.KILL_COMMAND_SURVIVAL_TALENT,
+    condition: cnd.describe(
+      cnd.and(
+        cnd.hasTalent(TALENTS.RELENTLESS_PRIMAL_FEROCITY_TALENT),
+        cnd.buffMissing(SPELLS.TIP_OF_THE_SPEAR_CAST),
+        cnd.buffPresent(SPELLS.COORDINATED_ASSAULT_BUFF),
+      ),
+      (tense) => <>you have 0 tip stacks & Relentless Primal Ferocity is active</>,
```
emallson:
```suggestion
      (tense) => <>you {tenseAlt(tense, 'have', 'had')} 0 tip stacks & Relentless Primal Ferocity is active</>,
```

use `tenseAlt` to adjust the tense of the description so that it is correct when shown in past-tense contexts (usually describing rotational errors, ex: "you should have cast this because you had 0 tip stacks" instead of "you should have cast this because you have 0 tip stacks")

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
```diff
@@ -0,0 +1,196 @@
...
+      (tense) => (
+        <>
+          you {tenseAlt(tense, <>have</>, <>had</>)} at most 77 focus and don't have
+          <SpellLink spell={TALENTS.RELENTLESS_PRIMAL_FEROCITY_TALENT} /> up OR you have
+          <SpellLink spell={TALENTS.RELENTLESS_PRIMAL_FEROCITY_TALENT} /> with either no Tip stacks
+          or less than 30 focus.
+        </>
+      ),
+    ),
+  },
```
emallson: this condition might be *too complex* to be useful in practice. it probably warrants splitting up into multiple rules:

- cast kill command to apply coordinated assault (if you won't overcap focus)
- cast kill command to generate tip of the spear stacks (if you won't overcap focus)
- cast kill command to generate focus when you're low

the sequencing of the rules functions like an or, and can be easier to understand than a single large condition. combining a couple can be okay (i think both the tip and CA rules are kinda related? so maybe they go together?), but at 3 clauses its pretty chonky

src/analysis/retail/hunter/survival/modules/guide/sections/rotation/Cooldown.tsx
```diff
@@ -0,0 +1,79 @@
...
+      />
+      {info.combatant.hasTalent(TALENTS.SPEARHEAD_TALENT) && (
+        <CastEfficiencyBar
+          spellId={TALENTS.SPEARHEAD_TALENT.id}
+          gapHighlightMode={GapHighlight.FullCooldown}
+        />
+      )}
+
+      <SubSection title="Trinket Cooldowns">
+        {info.combatant.hasTrinket(133282) && (
```
emallson: use the named entry in `ITEMS` (probably `ITEMS.SKARDYNS_GRACE`) rather than the raw id

src/analysis/retail/hunter/survival/modules/guide/sections/rotation/Cooldown.tsx
```diff
@@ -0,0 +1,79 @@
...
+              <SpellLink spell={TALENTS.COORDINATED_ASSAULT_TALENT} /> when
+              <strong> NOT</strong> using <SpellLink spell={TALENTS.SENTINEL_WATCH_TALENT} />. If
+              using
+              <SpellLink spell={TALENTS.SENTINEL_WATCH_TALENT} /> then use them together when both
+              are available, but do not delay one for the other if it will cost you casts later.
+            </p>
+            <CastEfficiencyBar spellId={92099} gapHighlightMode={GapHighlight.FullCooldown} />
+          </>
+        )}
+        {info.combatant.hasTrinket(212454) && (
```
emallson: use the named entry in `ITEMS`

src/analysis/retail/hunter/survival/modules/guide/sections/rotation/RotationSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+    </Section>
+  );
+}
+/*
+TODO: add back in after rotational cooldowns subsection. Current problem is it's not counting Focus properly for hasResource.
+Recommending to Raptor when you're <77 focus and you pushed KC instead.
+<SubSection title="APL Analysis">
+        <AplSectionData checker={AplCheck.checkApl} apl={AplCheck.apl} />
+      </SubSection>
+*/
```
emallson: focus might be reported x100, like Rage is. should check that, as it'd mean that being at 1 focus logs as 100 "focus"

src/analysis/retail/hunter/survival/modules/talents/WildfireBomb.tsx
```diff
@@ -194,6 +259,37 @@ class WildfireBomb extends Analyzer {
...
+        should be kept off maximum charges and always be cast with{' '}
+        <SpellLink spell={SPELLS.TIP_OF_THE_SPEAR_CAST.id} />. It can go untipped if any of:
+        <br></br>
+        1. You are capped on bomb charges. <br></br>
+        2. Lunar Storm is ready <br></br>
+        3. You are about to press Coordinated Assault and have{' '}
+        <SpellLink spell={TALENTS.BOMBARDIER_TALENT} /> talented.<br></br>
+        4. You are about to press Butchery and the cooldown reduction from Frenzied Strikes would
+        overcap bomb.
+      </p>
```
emallson: don't use `<br/>` for line formatting. it breaks on differing screen sizes. for this case, you can use `<ol>` and `<li>` to get a formatted ordered list with proper spacing and automatic line breaks.

emallson (changes requested):
this is a big one. most of the comments are pretty minor (wording/grammar, formatting, minor code bits).

there are some comments in the APL about conditional usage. it kinda looks like it got left in an in-progress state and then not totally cleaned up.

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
Kivlov: Was struggling to get it to show anything other than the Serpent Sting missing so that explains why.
 I was trying to avoid having a lot of separate APL lines and get a single priority to display what  they did incorrectly at the time since these 2 basically will never conflict in actual play and are your highest priority action if either is true outside of RPF CA. Would it be better to just go with multiple lines and just accept that there will be some bloat to the list? I had this same issue with wildfire bomb below where the APL line is a single line and so they're all of equal importance but end up being split into 3 different lines in the APLchecker.

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
Kivlov: By splitting into multiple rules you mean like I have Wildfire bomb with a 1.9 fractional, 1.5 fractional w/ Tip, and Bombardier is talented & CA is ready right?

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
emallson: it might be. im not sure how fundamentally complicated the SV apl is vs what are artifacts of simc. however, if you want to have different descriptions then the only option is to have different rules.

src/analysis/retail/hunter/survival/modules/apl/AplCheck.tsx
emallson: something like that, yes. the goal is clarity. too many rules can hurt that, so it is a bit of a judgment call. sometimes doing an combined rule is better. my 2c is to focus on *intent* for it. what you're describing with wildfire bomb sounds like:

- don't cap wildfire bomb charges
- spend tip on wildfire bombs before CA

is that right?

Kivlov: Don't cap bombs
Dump bombs before CA if you have bombardier talented because bombardier refreshes bomb to 2 charges
Bomb if you're ~1.5 charges and have a tip stack - The reason for this one is if you don't throw it at roughly this point you'll very rapidly approach the 1.9 and cap since we have so much bomb CDR.

Then there is a lower priority wildfire bomb that I haven't implemented yet which accounts for if explosive, killshot, butchery are all on cooldown and you're high focus, to just use the bomb charge since it's still better than raptoring.

The APL itself in simc isn't particularly large but it's largely because they combined a lot of the bomb conditions into single lines since they're all ~ equal.

So if that's the case I think multiple lines and then I'll have to put a section above it discussing that certain priorities are equal despite having an apparent hierarchy in the APL list?

emallson: like i said, its a judgment call. what do you think is clear and easy to understand? if you were to explain to someone in discord when to push bomb, what would you tell them?

there's room for the checker in wowa to be more forgiving / allow slightly sub-optimal play in pursuit of being easier to understand and still catching a ton of common problems.

you can also do things that combine multiple different spells into one. for example a "dump focus" rule at the bottom of the APL that combines Wildfire Bomb, Raptor Strike, or whatever else as a catch-all "push something, anything is better than nothing" rule

emallson: (also, it doesn't need to be perfect. i'm prone to trying to make it so, but everything here is iterative. its okay to merge something that has flaws and then adjust later)

emallson (approved):

emallson:
thanks!

PR #7205, opened 2024-12-18
[Warrior] [Fury] Revamp Fury modules for The War Within

GeoffreyBalshaw:
### Description

Warrior hasn't been updated with major talent and theorycrafting updates in a few years. In particular, the current iteration of WoWAnalyzer places a very high priority on effectively using Rampage and avoiding wasted Rage, which is no longer a very high priority for Fury in most cases. 

This PR updates some existing modules, mostly to account for Rampage's relatively lower priority, as well as adding a few new modules to account for common errors in the most common Fury raid spec

### Testing

- Test report URL: `/report/vM8zdCPFhZkxfW3y/45/13`, `/report/bzfPd1NBxRa9hG7n/17/15`
- Screenshot(s):
image

### DIFF
--- a/src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
+++ b/src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
@@ -1,0 +1,193 @@
+import { defineMessage } from '@lingui/macro';
+import SPELLS from 'common/SPELLS';
+import talents from 'common/TALENTS/warrior';
+import { SpellLink } from 'interface';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, {
+  ApplyBuffEvent,
+  ApplyDebuffEvent,
+  ApplyDebuffStackEvent,
+  CastEvent,
+  EventType,
+  RefreshBuffEvent,
+  RemoveBuffStackEvent,
+  RemoveDebuffEvent,
+  RemoveDebuffStackEvent,
+} from 'parser/core/Events';
+import { ThresholdStyle, When } from 'parser/core/ParseResults';
+import { currentStacks } from 'parser/shared/modules/helpers/Stacks';
+import SpellUsable from '../features/SpellUsable';
+import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
+
+/*  Example log:
+ *  https://www.warcraftlogs.com/reports/vM8zdCPFhZkxfW3y?fight=45&type=casts&source=13
+ *  https://www.warcraftlogs.com/reports/bzfPd1NBxRa9hG7n?fight=17&type=casts&source=15
+ */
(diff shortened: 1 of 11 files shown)


src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
```diff
@@ -0,0 +1,193 @@
...
+    let usedForAshenJuggernaut = false;
+    let usedForSuddenDeath = false;
+    let usedForMarkedForExecution = false;
+
+    if (this.ashenJuggernautExpirationTimestamp - event.timestamp < BUFF_REFRESH_BUFFER) {
+      usedForAshenJuggernaut = true;
+    }
+    // Assume new AJ duration on cast of Execute
+    // which is true unless the execute is parried
+    this.ashenJuggernautExpirationTimestamp = event.timestamp + ASHEN_JUGGERNAUT_DURATION;
```
emallson, line 136: making a note for future improvement that this is a case where an event link could be useful. by linking the execute to its damage, you can look ahead and check if it was parried so you don't need to assume (without having to delay processing the cast til a later point when the damage actually happens---sounds like execute has some weirdness with that)

src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  suggestions(when: When) {
+    when(this.suggestionThresholds).addSuggestion((suggest, actual, recommended) =>
+      suggest(
+        <>
+          You cast execute prematurely {actual} times. Execute has powerful buffs and debuffs
+          associated with it, but does relatively little damage on its own. Refer to Wowhead or
+          Maxroll guides for a full description of when to best use{' '}
+          <SpellLink spell={SPELLS.EXECUTE_FURY} />
+        </>,
```
emallson, line 179: is the maxroll guide for fury actually good? i know for a lot of specs their guides have been crap but maybe they got a good fury writer.

emallson (approved):
this looks like a great start. thanks!

src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
GeoffreyBalshaw: Agreed, will look into adding this in future updates (which will be coming for 11.1 since they're changing how RA works again)

src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
GeoffreyBalshaw: Yeah, it's written (_actually_ written, not GPTd as far as I know) by Revvez, and tends to have very solid info. It sometimes takes him a bit to update after RWF, hence recommending Wowhead as well

PR #7208, opened 2024-12-22
Initial Chronowarden support for Aug

KYZ124:
### Description

- Enabled the Chrono Flame and Thread of Fate modules for Aug. (These already existed for Pres and require no changes to be used for Aug.)
- Adjusted the Reverberations module to also work for Aug.
- Fixed small error causing Mass Eruption module to load based on whether Eruption, rather than Mass Eruption, was talented.

### Testing

ChronowardenAugInitial

[Log link](https://www.warcraftlogs.com/reports/dARpygajT62QLVPn?fight=15&source=52)


Krealle:
LGTM.

PR #7212, opened 2024-12-29
Update guide section

KYZ124:
### Description

Update guide section. Removed Dragonflight season 3/4 4pc.

### Testing

MandatoryScreenshot


src/analysis/retail/evoker/augmentation/modules/Abilities.tsx
```diff
@@ -129,6 +130,7 @@ class Abilities extends CoreAbilities {
         spell: SPELLS.TREMBLING_EARTH_BUFF.id,
         category: SPELL_CATEGORY.OTHERS,
         gcd: null,
+        enabled: combatant.has4PieceByTier(TIERS.DF3) || combatant.has4PieceByTier(TIERS.DF4),
```
emallson: these tier sets are disabled now, right? if so, this should be removed rather than gated by tier set.

KYZ124: These tier sets are disabled, yes. I'll remove the old tier in a separate pull request.

There's currently an error which I couldn't work out exactly how to fix (still fairly new to this repo) where a few spells, including this one, show up in the cooldowns section when they clearly shouldn't. I've asked Voll about it, but in the meantime, requiring that tier set avoids the issue on this spell for anyone who doesn't have that 4pc equipped.

src/analysis/retail/evoker/augmentation/modules/guide/CoreRotation.tsx
```diff
@@ -28,7 +28,10 @@ export function CoreRotationSection({ modules, events, info }: GuideProps<typeof
         <SpellLink spell={SPELLS.LIVING_FLAME_CAST} /> to fill in the gaps.
         <br />
         Your main cooldown, <SpellLink spell={TALENTS_EVOKER.BREATH_OF_EONS_TALENT} />, should be
-        used alongside your other DPS players' major cooldowns, since it amplifies their damage.
+        used alongside your other DPS players' major cooldowns, since it amplifies their damage. As
+        Scalecommander, it should instead be used as often as possible where mobs will survive the
+        full duration.
```
Krealle: The Scalecommander part should be conditional based on whether or not the player is using those Hero Talents.
```suggestion
        used alongside your other DPS players' major cooldowns, since it amplifies their damage. {info.combatant.hasTalent(TALENTS_EVOKER.MASS_ERUPTION_TALENT) && (As
        Scalecommander, it should instead be used as often as possible where mobs will survive the
        full duration.)}
```

PR #7213, opened 2025-01-02
Several Chronowarden and Augmentation talents; Cirral Concoctory

KYZ124:
### Description

Implements modules for:
- Time Spiral
- Time Convergence
- Master of Destiny
- Motes of Acceleration
- Golden Opportunity (Aug)
- Overlord
- Hoarded Power (Aug)
- Cirral Concoctory

### Testing

TSGOOlHPMoA
TCMoDCirral


src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MasterOfDestiny.tsx
```diff
@@ -0,0 +1,111 @@
...
+    );
+    this.addEventListener(
+      /**
+       * TODO: Check what happens if all eligible players have a Thread of Fate and it is reapplied by a different player.
+       * Currently assumes that it removes the previous buff, then applies a new one, so refreshbuff does not need to handle the case where a different player is refreshing the buff.
+       * Very unlikely case to happen in practice.
+       */
+      Events.refreshbuff.by(SELECTED_PLAYER).spell(SPELLS.THREAD_OF_FATE_BUFF),
+      this.onRefreshBuff,
+    );
```
Krealle: If it were to get applied by another player on top of one of your own, wouldn't they simply stack or have yours removed? Since you won't see an external application event from another player(except for on yourself), you wouldn't be able to account for that either.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MasterOfDestiny.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const averageThreadDuration = this.totalThreadDuration / this.totalThreadsApplied;
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.CORE(5)}
+        size="flexible"
+        category={STATISTIC_CATEGORY.HERO_TALENTS}
+      >
+        <TalentSpellText talent={TALENTS_EVOKER.MASTER_OF_DESTINY_TALENT}>
+          <div>
+            <DamageIcon /> {formatNumber(this.totalThreadExtension)} sec
```
Krealle: This shouldn't be a `DamageIcon`, since it's not showing DPS.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MotesOfAcceleration.tsx
```diff
@@ -0,0 +1,68 @@
...
+      >
+        <TalentSpellText talent={TALENTS_EVOKER.MOTES_OF_ACCELERATION_TALENT}>
+          <div>
+            <InformationIcon /> {this.motesCollected}
+            <small> motes collected</small>
+            <br />
+            <WarningIcon /> {motesWasted}
+            <small> motes unused</small>
+          </div>
+        </TalentSpellText>
```
Krealle: You should only show the unused amount if there are any.
```suggestion
        <TalentSpellText talent={TALENTS_EVOKER.MOTES_OF_ACCELERATION_TALENT}>
          <div>
            <InformationIcon /> {this.motesCollected}
            <small> motes collected</small>
            {motesWasted > 0 && (
              <>
                <br />
                <WarningIcon /> {motesWasted}
                <small> motes unused</small>
              </>
            )}
          </div>
        </TalentSpellText>
```

src/analysis/retail/evoker/shared/constants.ts
```diff
@@ -14,6 +14,8 @@ export const BASE_EVOKER_RANGE = 25;
 export const CLOBBERING_SWEEP_CDR = 45;
 export const HEAVY_WINGBEATS_CDR = 45;
 
+export const TIME_SPIRAL_BASE_DURATION = 10;
```
Krealle: It would make more sense for this const to be in milliseconds.
```suggestion
export const TIME_SPIRAL_BASE_DURATION_MS = 10_000;
```

src/analysis/retail/evoker/shared/modules/talents/TimeSpiral.tsx
```diff
@@ -0,0 +1,173 @@
...
+              personal <SpellLink spell={SPELLS.HOVER} /> casts wasted
+            </small>
+            <br />
+            <InformationIcon /> {this.externalBuffsUsed}
+            <small> external buffs used</small>
+            <br />
+            <WarningIcon /> {externalBuffsWasted}
+            <small> external buffs unused</small>
+          </div>
+        </TalentSpellText>
```
Krealle: The wasted buffs should only be shown if relevant.
```suggestion
        <TalentSpellText talent={TALENTS_EVOKER.TIME_SPIRAL_TALENT}>
          <div>
            <SoupIcon /> {this.personalBuffsUsed}
            <small>
              {' '}
              personal <SpellLink spell={SPELLS.HOVER} /> casts gained
            </small>

            {personalBuffsWasted > 0 && (
              <>
                <br />
                <WarningIcon /> {personalBuffsWasted}
                <small>
                  {' '}
                  personal <SpellLink spell={SPELLS.HOVER} /> casts wasted
                </small>
              </>
            )}

            <br />
            <InformationIcon /> {this.externalBuffsUsed}
            <small> external buffs used</small>
            
            {personalBuffsWasted > 0 && (
              <>
                <br />
                <WarningIcon /> {externalBuffsWasted}
                <small> external buffs unused</small>
              </>
            )}
          </div>
        </TalentSpellText>
```

src/analysis/retail/evoker/shared/constants.ts
```diff
@@ -72,6 +74,7 @@ export const FAN_THE_FLAMES_INCREASE = 1;
 export const RED_HOT_INCREASE = 0.2;
 
 export const PRIMACY_HASTE_PER_STACK = 3;
+export const THREAD_OF_FATE_BASE_DURATION = 10;
```
Krealle: This should also be in milliseconds.
```suggestion
export const THREAD_OF_FATE_BASE_DURATION_MS = 10_000;
```

src/analysis/retail/evoker/augmentation/constants.tsx
```diff
@@ -9,6 +9,8 @@ export const EBON_MIGHT_BASE_DURATION_MS = 10000;
 
 // Mastery aura extension
 export const TIMEWALKER_BASE_EXTENSION = 0.04;
+// Multiplier for Sands coefficient to get the extension coefficient
+export const TIMEWALKER_EXTENSION_MULTIPLIER = 0.5 / 0.34;
```
Krealle: So looking at the current implementations that interact with mastery, they are all related to Time Walker and not Shift Sands. So I think we should simply change the `masteryCoefficient` value in `SPECS.ts` to use Time Walker coefficient instead. Then there won't be any need for the `TIMEWALKER_EXTENSION_MULTIPLIER` constant.

I think for clarity sake we can make a const for each coefficient and just import the Time Walker one into `SPECS.ts`.
I doubt that Shifting Sands coefficient will actually be used in analysis.
```suggestion
// Mastery Versatility buff
export const SHIFTING_SANDS_MASTERY_COEFFICIENT = 0.34;
// Mastery aura extension
export const TIMEWALKER_BASE_EXTENSION = 0.04;
export const TIMEWALKER_MASTERY_COEFFICIENT = 0.5;
```

src/analysis/retail/evoker/augmentation/modules/talents/Overlord.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  onDamage(event: DamageEvent) {
+    if (
+      this.selectedCombatant.hasBuff(SPELLS.DEEP_BREATH) ||
+      this.selectedCombatant.hasBuff(TALENTS.BREATH_OF_EONS_TALENT) ||
+      this.selectedCombatant.hasBuff(SPELLS.BREATH_OF_EONS_SCALECOMMANDER)
+    ) {
+      this.overlordDamage += event.amount;
+    }
+  }
```
Krealle: Not sure if it's possible for the damage events to happen after the buff is removed? Since it is commonplace for SC to cancelaura.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/GoldenOpportunity.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * Pres [NYI]: Casting Echo has a 20% chance to cause your next Echo to copy 100% more healing.
+ */
+class GoldenOpportunity extends Analyzer {
+  static dependencies = {
+    stats: StatTracker,
+  };
+  protected stats!: StatTracker;
+  goldenPrescienceApplyTimestamps: { [key: number]: number } = {};
+  goldenPrescienceTimestampExists: { [key: number]: boolean } = {};
+  masteryAtPrescienceApplication: { [key: number]: number } = {};
```
Krealle, line 34: This isn't necessarily something you need to change, but I'm not a huge fan of having three records that all pertain to the same application. Ideally you would store the relevant information for each application as an object, in a singular record.
eg.
```ts
goldenPrescienceRecord: {
  [key: number]: {
    timestamp: number;
    mastery: number;
  };
} = {};

goldenPrescienceRecord[targetID] = {
  timestamp: event.timestamp,
  mastery: this.stats.currentMasteryPercentage,
};

const application = this.goldenPrescienceRecord[targetID];
if(!application) { 
  return; 
}
application.timestamp // access the values

this.goldenPrescienceApplications[targetID] = nil // remove the application
```

You could also look into learning a bit about `Maps`. They are a great way to store data in a structured way, and come with a lot of built in functions that make it easier to work with.

I'll show an example of how you could use one here:
```ts
// Define the type of the data we want to store
type GoldenPrescienceApplication = {
  timestamp: number;
  baseDuration: number;
};

goldenPrescienceMap: Map<number, GoldenPrescienceApplication> = new Map();

onFightEnd(event: FightEndEvent) {
  this.goldenPrescienceMap.forEach((application) => {
    this.onPrescienceRemove(application.timestamp, event.timestamp, true);
  });
}

onPrescienceApply(targetID: number, timestamp: number) {
  // Calculating base duration here instead of carrying around the mastery value makes more sense
  const basePrescienceDuration =
    (PRESCIENCE_BASE_DURATION_MS *
      (1 + TIMEWALKER_BASE_EXTENSION + this.stats.currentMasteryPercentage)) /
    1000;
  this.goldenPrescienceMap.set(targetID, {
    timestamp,
    baseDuration: basePrescienceDuration,
  });
}

onPrescienceRemove(targetID: number, timestamp: number, fightEnd: boolean) {
  const application = this.goldenPrescienceMap.get(targetID);
  if (!application) {
    return;
  }

  const prescienceDuration = (timestamp - application.timestamp) / 1000;
  if (!fightEnd && prescienceDuration >= application.baseDuration * 1.9) {
    // Use 1.9 as a threshold to account for variations in timestamps.
    this.totalPrescienceExtension += prescienceDuration / 2;
  } else {
    // Prescience was removed early due to fight end, cancelaura, or death.
    // Approximately calculate extension value based on Mastery at the time of application.
    const extensionValue = prescienceDuration - application.baseDuration;
    if (extensionValue > 0) {
      this.totalPrescienceExtension += extensionValue;
    }
  }
  this.goldenPrescienceMap.delete(targetID);
}
```

Krealle (changes requested):
So I'll just preface this with saying that I think the scope of this PR is too big - in the future it would be best to split it up into to smaller PRs.
You also need to provide links to some logs so it's easier to test out changes.

With that out of the way, for the most part I think it looks fine. Left you a couple comments to look at.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MasterOfDestiny.tsx
KYZ124: Any recommendations for which icon could be used here? `InformationIcon` is used directly below, so I was using `DamageIcon` to avoid using it consecutively.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MasterOfDestiny.tsx
KYZ124: As of beta, Thread of Fate doesn't stack, and instead overwrites. It only does this if all eligible players already have Thread of Fate, however. This isn't an impossible scenario (e.g. it could happen if you had two Chronowardens in M+, with one or both running Time Skip or Master of Destiny), but as noted it's very unlikely.
As you've pointed out though, it's not something that can be accounted for if it does actually happen, so I'll remove the note.

src/analysis/retail/evoker/augmentation/modules/talents/Overlord.tsx
KYZ124: It is possible (and was the subject of an entire bug allowing Overlord Eruptions to apply Bombardments), but requires extremely tight timing - Breath must be cancelled within the travel time of Eruption, which is obviously hitting a target right next to you - and is unlikely to be done unintentionally, especially in AOE where targets may not be stacked tightly enough to allow this sort of cancellation.
Not sure if there's a way to check if `this.selectedCombatant` has recently had the buff? If there is, it might be possible to check for that case as well. This would also require checking that there is no castlink with an actual Eruption cast, as Breath can be cancelled directly into an Eruption cast without any GCD or anything in-between.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/GoldenOpportunity.tsx
KYZ124: Noted - will use a single object in future for this kind of thing.
This sort of happened because initially I was just storing a timestamp per player, and then also a boolean to check if a timestamp had been saved, and then even later the Mastery at the time of application due to this being snapshot.

src/analysis/retail/evoker/shared/modules/talents/hero/chronowarden/MasterOfDestiny.tsx
Krealle: Could maybe use the `Uptime` icon for the average and `InformationIcon` for the extra

src/analysis/retail/evoker/augmentation/modules/talents/Overlord.tsx
Krealle: One of the args for `hasBuff` is `bufferTime` which will let you check backwards yea.
Prolly better to solve with a castlink though, it's very easy to distinguish between the sources due to Eruption not having a travel time, and breath having a windup period. Then you can also just get the events from Breath Cast instead of going over every damage event.

Just need to check if there is a `ERUPTION_CAST_DAM_LINK` link, which there *shouldn't* be for the overlord events.

Krealle:
LGTM, let's send it.

PR #7215, opened 2025-01-06
Fix crashes in Surging Totem & Ascendance analysis

emallson:
### Description

reported in discord.

- Ascendance (@Seriousnes): https://www.warcraftlogs.com/reports/AndLvgWprm73JNMK?fight=last&type=casts&source=75&ability=114050 note that the player basically never casts the spell
- Surging Totems (@danny-janse): couple reports, but easy to see here: https://www.warcraftlogs.com/reports/ZfNXvxwmznpPF63Y?type=summary&source=15&fight=12&view=events `Ctrl-F` Whirling. They lose the buff without casting Surging Totem because they cast it pre-pull. Ultimately, this should get pre-pull normalization to properly reflect cooldown. However, for the moment this stops the crash.

### DIFF
--- a/src/analysis/retail/shaman/elemental/modules/talents/Ascendance.tsx
+++ b/src/analysis/retail/shaman/elemental/modules/talents/Ascendance.tsx
@@ -132,6 +132,8 @@ class Ascendance extends MajorCooldown<AscendanceCooldownCast> {
     this.addEventListener(Events.any.by(SELECTED_PLAYER), this.onCast);
 
     if (this.selectedCombatant.hasTalent(TALENTS.DEEPLY_ROOTED_ELEMENTS_TALENT)) {
+      // initialize the ascendance proc counter in case the player cast a proc-eligible spell before casting ascendance
+      this.castsBeforeAscendanceProc.push({ count: 0 });
       this.addEventListener(
         Events.cast
           .by(SELECTED_PLAYER)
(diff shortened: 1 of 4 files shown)


src/analysis/retail/shaman/elemental/modules/talents/Ascendance.tsx
```diff
@@ -132,6 +132,8 @@ class Ascendance extends MajorCooldown<AscendanceCooldownCast> {
     this.addEventListener(Events.any.by(SELECTED_PLAYER), this.onCast);
 
     if (this.selectedCombatant.hasTalent(TALENTS.DEEPLY_ROOTED_ELEMENTS_TALENT)) {
+      // initialize the ascendance proc counter in case the player cast a proc-eligible spell before casting ascendance
+      this.castsBeforeAscendanceProc.push({ count: 0 });
```
Seriousnes, line 136: I think this is going to make the statistic wrong, it's expected that the array only included elements that lead to a proc.

emallson: what is the intended state if you cast a spell that could result in a proc before casting ascendance then? that's the current crash. i can pull that out of this pr and leave the fix to you if you prefer

Seriousnes: I was only going off memory, you are probably right. This PR is fine, I'll have a look a little later on

PR #7222, opened 2025-01-28
[Shadow]  Additional support for Voidweaver Hero Talents.

DoxAshe:
### Description

This adds statistics for Inner Quietus, Void heart, Devour Matter, Void Empowerment, Depth of Shadows, and Darkening Horizon. 
 These are all talents added by Shadow's Voidweaver Hero tree, and now all of the talents in that tree have statistics.  

In addition, this adds important information about entropic rift into the guide view when using the voidweaver talents.  With this, both of shadow's hero talents are now fully supported.  There is still more to do in the guide view with voidweaver, as Void Blast and Mind Blast need to be presented better, but all the information is correct and there.

### Testing

These talents are not taken very often, so I only had a few logs to check them against.  

- Test report URL: 
`report/G3YJa9XQkgRptTfb/40-Heroic+Queen+Ansurek+-+Kill+(5:56)/Selur/standard/statistics'
'report/hRbDaN4BgJCcrZyk/35-Normal+Queen+Ansurek+-+Kill+(4:34)/Emoladots/standard/overview'

- Screenshot(s):
- 
guide additions

statistics


src/analysis/retail/priest/shadow/modules/talents/Voidweaver/DevourMatter.tsx
```diff
@@ -0,0 +1,51 @@
...
+      this.onDamage,
+    );
+  }
+
+  onDamage(event: DamageEvent) {
+    if (event.absorbed !== undefined) {
+      //console.log(event.absorbed, "absorbed")
+      //Since Devour Matter deals a set amount of damage that is not seperated from SW:D in the logs,
+      //and that SW:D has several different buffs effecting its damage, I don't know how to seperate the damage contribution.
+      //this.damage += calculateEffectiveDamage(event, 0);
```
emallson, line 29: i went to read the talent, thinking i could help and...i mean it is probably possible to estimate it but wow that seems hard. leaving out the damage number seems like a better choice to me

src/analysis/retail/priest/shadow/modules/talents/Voidweaver/Voidheart.tsx
```diff
@@ -0,0 +1,48 @@
...
+    this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onDamage);
+  }
+
+  onDamage(event: DamageEvent) {
+    if (this.selectedCombatant.hasBuff(SPELLS.SHADOW_PRIEST_VOIDWEAVER_ENTROPIC_RIFT_BUFF)) {
+      //console.log(event.ability.name,"Spell of type", event.ability.type)
+      //During Entropic Rift, if the ability that caused the damage event is a shadow spell, we have a damage increase.
+      //shadow damage is type 32.
+      //the multi school spells with shadow should also be increased, but the only one this spec has access to is shadowfrost (48)
+      if (event.ability.type === 32 || event.ability.type === 48) {
```
emallson: you can use the `MAGIC_SCHOOL` helpers to handle this:

```suggestion
      //During Entropic Rift, if the ability that caused the damage event is a shadow spell, we have a damage increase. This also applies to multi-school spells (like Shadowfrost spells)
      if (isMatchingDamageType(event.ability.type, MAGIC_SCHOOLS.ids.SHADOW)) {
```

important note: if there are spells in Shadow's kit that *change types*, those changes are *not* reflected on WCL for technical reasons. for the purposes of Voidheart that is probably fine, but Ret has had some issues before.

src/analysis/retail/priest/shadow/modules/talents/Voidweaver/EntropicRift.tsx
```diff
@@ -41,22 +76,163 @@ class EntropicRift extends Analyzer {
...
+      this.currentDP = 0;
+
+      //Darkening Horizon
+      if (this.currentVB > 3) {
+        this.currentVB = 3;
+      } //VB can only give 3 seconds of extension
+      this.totalVB += this.currentVB;
+      this.finalizeRiftTime();
+      this.currentVB = 0;
+    }
```
emallson: this probably works fine. if you run into timestamp issues, you might try an `EventLink` to connect the damage events to the cast and process the damage events linked from the cast event directly. don't need to go out and rewrite working code for this PR though

emallson (approved):
minor comment on the ability type check in Voidheart. if you don't have time to revisit right now i'm good to go ahead and merge

src/analysis/retail/priest/shadow/modules/talents/Voidweaver/EntropicRift.tsx
DoxAshe: Right now the timestamps are all at the same time, and there can never be more than one rift out a a time.  But in the future I'll change it to use EventLinks so it works better in case that ever changes.

src/analysis/retail/priest/shadow/modules/talents/Voidweaver/DevourMatter.tsx
DoxAshe: For now I'm just having it show how many shields are hit by it.  But I'm planning on revisiting it down the line since its powerful enough for priests to switch hero trees on fights with shields.

src/analysis/retail/priest/shadow/modules/talents/Voidweaver/Voidheart.tsx
DoxAshe: That is a good change, and thankfully shadow doesn't have anything like that.  I'll leave a comment in there as a reminder just in case that changes in the future.

PR #7226, opened 2025-02-09
Update Resto Shaman for 11.0.7 

foobartles:
### Description

Fix various broken modules and add some missing modules for restoration shaman. 
List of changes:

- Fix downpour as being an incorrect cast for unleash life
- Fix Nature's swiftness module for farseer
- Add coalescing water module
- Fix downpour wasted procs module
- Fix mana stats on water shield/resurgence
- Add multiple different modules for totemic
- general cleanup, etc.

### Testing

- Test report URL: https://www.warcraftlogs.com/reports/N8tgmCPvHRk4V3Zc?fight=18&type=summary
- 
image
New totemic hero talent section

- 
image
New coalescing water module

I made a lot of changes for this so I am probably missing some stuff here.


src/analysis/retail/shaman/restoration/modules/Abilities.ts
```diff
@@ -28,7 +28,10 @@ class Abilities extends CoreAbilities {
         spell: TALENTS.RIPTIDE_TALENT.id,
         category: SPELL_CATEGORY.ROTATIONAL,
         enabled: combatant.hasTalent(TALENTS.RIPTIDE_TALENT),
-        charges: combatant.hasTalent(TALENTS.ECHO_OF_THE_ELEMENTS_TALENT) ? 2 : 1,
+        charges:
+          1 +
+          (combatant.hasTalent(TALENTS.ECHO_OF_THE_ELEMENTS_TALENT) ? 1 : 0) +
+          (combatant.hasTalent(TALENTS.ELEMENTAL_REVERB_TALENT) ? 1 : 0),
```
emallson: this can be simplified to:
```suggestion
        charges:
          1 + combatant.getTalentRank(TALENTS.ECHO_OF_THE_ELEMENTS_TALENT) +
          combatant.getTalentRank(TALENTS.ELEMENTAL_REVERB_TALENT),
```

src/analysis/retail/shaman/restoration/normalizers/CastLinkNormalizer.ts
```diff
@@ -325,6 +328,42 @@ const EVENT_LINKS: EventLink[] = [
       return c.hasTalent(talents.WHIRLING_ELEMENTS_TALENT);
     },
   },
+  // Lively Totems : When you summon a Healing Tide Totem, Healing Stream Totem, Cloudburst Totem, Mana Tide Totem, or Spirit Link Totem you cast a free instant Chain Heal at 100% effectiveness.
+  {
```
emallson, line 332: would be good to add a `GlobalCooldown` override that prevents these casts from triggering the GCD. i don't think the uptime stats are currently being used by rsham, though, so it isn't high priority

emallson (commented):
minor comments. once you fix the type errors, this looks fine to merge :+1:

src/analysis/retail/shaman/restoration/normalizers/CastLinkNormalizer.ts
foobartles: Do you have any examples of this? I wasn't aware of this functionality

emallson: uh we don't have a lot of these. we've used the `freecast` type in normalizers before to do this, though it is a small lie. it might be best to do that since it is pretty standard. and that prevents issues with other analyzers thinking these were "real" casts.

depends on your goals for it.

- `freecast` normalizer: look at `UnhingedMortalStrikeNormalizer` as a reference for how to do this based on event links. this will make everything looking for player casts ignore it (but the healing etc will still show)
- GCD: add an overriden `GlobalCooldown` module to your spec, then override the `isOnGlobalCooldown` method to check for the event links that indicate that it was cast for free.

after having thought about it i'm pretty sure the `freecast` approach will "just work" while the GCD one will leave some lingering analysis thinking the free casts are actual player casts, but either should work

PR #7227, opened 2025-02-10
New Downtime Foundation section for Melee specs

emallson:
### Description

Better to show than tell. This is going up on `next.wowanalyzer.com` as well.
image
image

This replaces the existing downtime section, which is primarily the active time line chart. The new version is interactive, and includes a miniature version of the player's ability timeline when zoomed.

Additionally, this introduces some suggestions into the guide view. The specific one here covers small GCD gaps, though we could add more. I tried a bunch of different approaches to this, and ultimately just plugged in the old `Suggestion` component and moved on. I needed to stop cooking and present my work.

### Additional Maintenance

If we were to adopt this, there would be a bit of extra maintenance every tier in setting up the boss abilities / debuffs. I help a lot with Wipefest setup each tier, and could pull from that to help set these up. We're coming up on both Dragon Soul and Undermine releases so we'd be presented a real test of that idea.

### Technical Notes

The main *big* technical thing here is the use of `svg` rather than HTML. The thought process was that the existing timeline makes extensive use of `position:absolute` to bypass normal layout, so why not use `svg` to draw the diagram instead?

~~This works and notably makes the scrolling performant enough to not really need tricks. This might also be implementable as regular horizontal scrolling (TBD---i'll probably investigate that early this week).~~ This uses native scroll now.

I think that overall the SVG code is easier to follow than the existing timeline code, though it admittedly is also doing less complicated individual displays and more of small-repeated-elements.

There are some downsides. Primarily: SVG is not HTML, which means that you need to embed things in `<foreignObject>`. This is *reasonably* seemless, but not totally. The `viewBox` property, which I'd originally thought to use for zoom/pan, is basically unusable in combination with `<foreignObject>` for spell icons. SVG being separate also means that there is an extra learning curve to working with it.

### Testing

You can pick any Foundation-level melee spec to see this diagram. Additionally, this is enabled on Blood DK and Brewmaster in retail.

I tested a bunch on my own logs, like [this](https://next.wowanalyzer.com/report/BtLVN31xwmnr4Kcf/26-Mythic+Queen+Ansurek+-+Kill+(7:17)/Eisenpelz/standard), as well as other brewmaster logs.


src/interface/guide/foundation/FoundationDowntimeSection.tsx
```diff
@@ -20,60 +22,70 @@ export function FoundationDowntimeSection(): JSX.Element | null {
...
+              {formatPercentage(alwaysBeCasting.activeTimePercentage, 1)}%
+            </PerformanceStrong>{' '}
             <ByRole>
-              <Role.Melee>
-                There should be no gaps between the end of one <GCD /> and the start of the next.
-              </Role.Melee>
               <Role.Caster>
-                There should be no gaps between the end of one spell cast and the start of the next.
+                Cancelled Casts:{' '}
+                {cancelledCasts /* need to check this because the parameters are evaluated even for melee */ && (
```
GeoffreyBalshaw, line 64: Is this a TODO for a future PR, something to be done in this PR, or can this comment be removed?

emallson: that branch is still evaluated for melee players. it is an explanatory note about why the check is needed

src/interface/guide/foundation/FoundationDowntimeSectionV2.tsx
```diff
@@ -0,0 +1,513 @@
...
+  meleeGaps,
+  globalMeleeGaps,
+  debuffSegments: rawDebuffSegments,
+}: Props): JSX.Element | null {
+  const info = useInfo();
+  const { fight } = useFight();
+
+  const boss = findByBossId(fight.boss);
+  const debuffSegments = useMemo(
+    () => rawDebuffSegments?.filter((segment) => segment.end - segment.start >= 4000) ?? [],
```
trevorm4: What's the significance of 4000 here? Maybe put it into a constant with a detailed name

src/interface/guide/foundation/FoundationDowntimeSectionV2.tsx
```diff
@@ -0,0 +1,513 @@
...
+        element: null,
+        hidden(x) {
+          return x(info.fightStart + 1000) - x(info.fightStart) < 16;
+        },
+      },
+      {
+        height: 16,
+        element: <PlayerAbilityTimeline info={info} />,
+        hidden(x) {
+          return x(info.fightStart + 1000) - x(info.fightStart) < 16;
```
trevorm4: Similar maybe put magic numbers into constants for readability

src/interface/guide/foundation/analyzers/MeleeUptimeAnalyzer.tsx
```diff
@@ -0,0 +1,159 @@
...
+  public get meleeUptimeGaps(): typeof this.gaps {
+    return this.gaps;
+  }
+
+  public get meleeUptimePercentage(): number {
+    const downtime = this.gaps.reduce((total, gap) => total + gap.end - gap.start, 0);
+    return (this.owner.fightDuration - downtime) / this.owner.fightDuration;
+  }
+
+  public get meleeUptimePerformance(): QualitativePerformance {
```
trevorm4, line 73: Consider using `evaluateQualitativePerformanceByThreshold` instead of manually doing calcs

src/interface/guide/foundation/FoundationDowntimeSectionV2.tsx
emallson: just an arbitrary threshold to cut out very short debuffs. it is a bit of an artifact of when i was trying to pick debuffs automatically instead of just using a list and can probably be removed entirely.

src/interface/guide/foundation/FoundationDowntimeSectionV2.tsx
emallson: this is just 1 second. calculating the pixel width of it to control turning on the extra timeline bar. but it could be clearer for sure

PR #7234, opened 2025-02-20
Update Aug for 11.1

KYZ124:
### Description

- Load Symbiotic Bloom module if the talent isn't taken but Motes of Possibility is, as motes can proc the buff.
- Disable the Shifting Sands module if Motes of Possibility is taken, as most 'failed' uses are more likely to be lucky mote streaks.
- Update Ebon Might buff helper to use the timestamps the player applied/removed Ebon Might on, rather than fixed 30/27 second intervals.
- Added a module for Upheaval Essence Burst uses.

### Testing

- Test report URLs: `https://www.warcraftlogs.com/reports/KwDT7GfLbFvVWdRn?fight=last&source=1` (Symbiotic Bloom, Upheaval); `https://www.warcraftlogs.com/reports/PdfBQVJRmDZpA6hX?fight=5&type=damage-done&source=13` (Ebon Might buff helper)
- Screenshot(s): 
SymbioticMotes
Upheaval
EbonHelper

### DIFF
--- a/src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
+++ b/src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
@@ -218,7 +218,9 @@ class BuffTargetHelper extends Analyzer {
 
     const fetchPromises: Promise<DamageTables>[] = [];
     while (currentTime < this.fightEnd) {
-      fetchPromises.push(this.getDamage(currentTime));
+      if (this.ebonRemoveTimestamps[index]) {
+        fetchPromises.push(this.getDamage(currentTime, this.ebonRemoveTimestamps[index]));
+      }
       index += 1;
       currentTime = this.ebonApplyTimestamps[index];
     }
@@ -272,12 +274,12 @@ class BuffTargetHelper extends Analyzer {
     });
   }
 
-  async getDamage(currentTime: number): Promise<DamageTables> {
+  async getDamage(currentTime: number, endTime: number): Promise<DamageTables> {
     const normalDamage = await fetchWcl<WCLDamageDoneTableResponse>(
       `report/tables/damage-done/${this.owner.report.code}`,
       {
         start: currentTime,
-        end: currentTime + this.interval,
+        end: endTime,
         filter: this.getFilter(false),
       },
     );
@@ -285,7 +287,7 @@ class BuffTargetHelper extends Analyzer {
       `report/tables/damage-done/${this.owner.report.code}`,
       {
         start: currentTime,
-        end: currentTime + this.interval,
+        end: endTime,
         filter: this.getFilter(true),
       },
     );
--- a/src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
+++ b/src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
@@ -7,7 +7,12 @@ import classColor from 'game/classColor';
 import ROLES from 'game/ROLES';
 import SPECS from 'game/SPECS';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { ApplyBuffEvent, RefreshBuffEvent, RemoveBuffEvent } from 'parser/core/Events';
+import Events, {
+  ApplyBuffEvent,
+  FightEndEvent,
+  RefreshBuffEvent,
+  RemoveBuffEvent,
+} from 'parser/core/Events';
 import Combatants from 'parser/shared/modules/Combatants';
 import { isMythicPlus } from 'common/isMythicPlus';
 import '../../Styling.scss';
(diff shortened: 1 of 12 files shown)


src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
```diff
@@ -138,6 +160,24 @@ class BuffTargetHelper extends Analyzer {
...
+  }
+
+  onEbonRemove(event: RemoveBuffEvent) {
+    this.ebonRemoveTimestamps.push(event.timestamp);
+  }
+
+  onEbonRemoveFightEnd(event: FightEndEvent) {
+    this.ebonRemoveTimestamps.push(event.timestamp);
+  }
+
```
Krealle: You should check if you currently have EM active before pushing in a removal event.

src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
```diff
@@ -171,12 +211,14 @@ class BuffTargetHelper extends Analyzer {
 
     // Start 4 seconds in since you start the fight with 2x Prescience -> Ebon Might
     // This will also show MUCH better value targets
-    let currentTime = this.fightStart + this.fightStartDelay;
+    let index = 0;
+    let currentTime = this.ebonApplyTimestamps[index];
 
     const fetchPromises: Promise<DamageTables>[] = [];
     while (currentTime < this.fightEnd) {
```
Krealle: Since you are creating intervals based on an array, instead of intervals. You should prolly just iterate the array instead of using while.

Also looking a bit into `this.getDamage`, this is simply expecting a `startTime` arg, and will use `startTime + this.interval` for the `endTime` in the query.
I'm assuming you are trying to get damage based on the windows people used? In which case you would need to refactor `this.getDamage` to accept both `startTime` & `endTime` args.

You could then potentially also push to the `fetchPromises` array on `removebuff`/`fightend` events instead of doing it here.

src/analysis/retail/evoker/augmentation/modules/talents/SymbioticBloom.tsx
```diff
@@ -22,14 +22,17 @@ import { SYMBIOTIC_HEALING_INCREASE } from '../../constants';
...
   // This is an approximation. See the reasoning below.
   totalHealingFromSymbioticBloomBuff = 0;
 
-  symbioticBloomHealIncrease =
-    SYMBIOTIC_HEALING_INCREASE *
-    this.selectedCombatant.getTalentRank(TALENTS.SYMBIOTIC_BLOOM_TALENT);
+  symbioticBloomHealIncrease = this.selectedCombatant.hasTalent(TALENTS.SYMBIOTIC_BLOOM_TALENT)
+    ? SYMBIOTIC_HEALING_INCREASE *
+      this.selectedCombatant.getTalentRank(TALENTS.SYMBIOTIC_BLOOM_TALENT)
+    : SYMBIOTIC_HEALING_INCREASE * 2;
```
Krealle, line 35: Is this correct?
If Symbiotic is talented with 1 point, the buff from motes will be equal to 1 point. But if only motes is talented, it will be equal to 2 points?

src/analysis/retail/evoker/augmentation/modules/thewarwithin/T33Augmentation4P.tsx
```diff
@@ -0,0 +1,141 @@
...
+import SpellLink from 'interface/SpellLink';
+import { combineQualitativePerformances } from 'common/combineQualitativePerformances';
+
+type UpheavalCast = {
+  event: EmpowerEndEvent;
+  essenceBurstStacks: number;
+};
+/**
+ * (4) Set Augmentation: Upheaval deals 30% increased damage and increases the damage of your next 2 Eruption casts by 30%.
+ */
```
Krealle: Isn't this T32?

Krealle (changes requested):
Looks mostly fine.
The changes in the `BuffTargetHelper` module are not going to work as intended and will need to be reworked a bit, left a comment regarding what needs changing.

src/analysis/retail/evoker/augmentation/modules/talents/SymbioticBloom.tsx
KYZ124: Yes, this is correct.

src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
KYZ124: `this.getDamage` argument issue has been fixed.
Will refactor `BuffTargetHelper` to iterate over the array and push to `fetchPromises` on `removebuff`/`fightend` events in a subsequent pull request.

src/analysis/retail/evoker/augmentation/modules/features/BuffTargetHelper/BuffTargetHelper.tsx
```diff
@@ -216,7 +218,9 @@ class BuffTargetHelper extends Analyzer {
...
```
Krealle: After looking at this with some rested eyes, I realize that this while loop should never actually be able to broken out of unless you somehow manage to apply an EM on the same tick that the fight ends.
It should be made into some sort of iteration across the timestamp arrays.
eg.
```ts
    for (let i = 0; i < this.ebonApplyTimestamps.length; i += 1) {
       fetchPromises.push(this.getDamage(this.ebonApplyTimestamps[i], this.ebonRemoveTimestamps[i]))
    }
```
The while loop will also end up trying to index undefined fields, so you could do an undefined check and break the loop based on that, but a for loop makes more sense.

PR #7239, opened 2025-02-25
Update talents for patch 11.1

emallson:
### Description

This one's a doozy. There are a TON of removed spells in the patch notes, with some (like Rapture for Disc) being pretty fundamental to the spec.

This PR handles those removals, plus some renames. As much as possible, I tried to keep things that were just renamed/combined. For example: the `PurgeTheWicked` module is still present because the talent now applies to SWP (though a disc main would need to say whether it is worth enabling without Encroaching Shadows enabled). However, in the case of reworks or removals I just deleted things.

### Testing

I have not tested every spec extensively. I'm relying a bit on the e2e tests to find issues here, because the changes are SO big.


GeoffreyBalshaw (approved):
Warrior changes look good, I can't really comment on other classes though

PR #7244, opened 2025-03-01
[Shadow] Update Shadow Priest for 11.1 and add Season 2 tier set.

DoxAshe:
### Description

This updates shadow for its 11.1 changes, and adds support for the new tier set.

### Testing

The new tier set for shadow is causing some problems.  When a Jackpot! occurs, a void bolt without a cast event appears and damages a target, and sometimes causes a cast event of Power Infusion.  I am unable to find a way to properly track these Jackpot! Power Infusion Casts, so its creating issues with Power Infusion spell usable and graph.

- Test report URL: `/reports/tKzRwp2QrP6AvFCq?type=summary&fight=11`
- Screenshot(s):
- 
PI issue


emallson (approved):
👍 if you want to fix the PI thing, the way to do that is *probably* to mutate the event type to `freecast` in a normalizer. but that can be done later.

PR #7253, opened 2025-03-03
[Shadow] Fix Power Infusion with TWW season 2 tier set

DoxAshe:
### Description

I found a work around for shadow's TWW season 2 tier set casting free Power Infusions.  Since a normal cast of PI always targets an additional target in range, while the tier set casts never do, we can use this to create an event link that can separate them, then use that event link to properly change the Jackpot! casts as free casts.


emallson (approved):
so the other potential way to do this is to try linking it to a void bolt that triggers it. however, since PI is off-gcd you could hit that condition naturally.

i think this seems like a good solution to try, and maybe ends up more reliable in practice because twins range is _pretty big_

PR #7258, opened 2025-03-05
[Elemental & Enhancement] TWW S2 Update

Seriousnes:
### Description

Pretty big update to most of the Enhancement and Elemental modules

Summary of core changes:

- Added `spellToAbility` as the inverse to the existing `abilityToSpell`
- Updated title in `ExplanationRow` from `string` to `string | ReactNode`
- To `Events` and `EventMetaLib`, added a non-enhanced/non-inefficient `additionalCastInfo` property that's not so noisy on the timeline
- Added `fractionalChargesAvailable` to `SpellUsable` (used the BM Barbed Shot implementation), plus changed visibility of a few fields to `protected`


emallson (commented):
looked over the core changes and they look okay. i didn't read the shaman parts

PR #7259, opened 2025-03-05
Update VoidTorrent.tsx

jordantoine:
fix typo where "whith" should be "with"

### Description

Update description for void torrent usage. There exists a typo, this commit fixes it.

### Testing

Load a shadow priest report with Void Torrent talented

- Screenshot(s):


emallson (approved):
thanks for the PR!

normally we require changelog entries and contributors to add themselves to the `CONTRIBUTORS` file, but since this is just a typo fix i can bypass that CI check if you want

PR #7260, opened 2025-03-07
[Shadow] Fix Power Infusion without TWWS2 Four Piece

DoxAshe:
### Description

My last change fixed Power Infusion with the Four Piece, but I missed that it caused Power Infusion to be wrong without the four piece.  This makes my last change only happen if the priest has the new Four Piece.


emallson (commented):
does mean that the `isFreePI` check is incorrect? since it should only be able to trigger when you have the 4pc, right?

DoxAshe:
> does mean that the `isFreePI` check is incorrect? since it should only be able to trigger when you have the 4pc, right?

Sort of.  The linking relationship makes a Link on PI casts that do have an additional target.  Since the Linking Relation is only enabled when the four piece is equipped, and isFreePI checks if the linking relationship is not there, it is always returning True.

I've renamed isFreePI to isUnpairedPI to make it a little clearer, that it isn't checking to see if its free, just if there was another target.  And I've added a check to make sure the player has the 4 piece along side it in the normalizer, so it now works with the tier piece and without the tier piece.

emallson:
ooooh that makes so much sense

emallson (approved):

PR #7264, opened 2025-03-10
windwalker: update APL for tww2

smvoss:
### Description

Add TWW2 APL

### Testing

- Test report URL: `/report/AD1yaYTtCfN6PpZR/3-Heroic+Sprocketmonger+Lockenstock+-+Kill+(5:25)/Ausgemisted/standard`
- Screenshot(s): 
image

### DIFF
--- a/src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
+++ b/src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
@@ -4,136 +4,330 @@ import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 import SpellLink from 'interface/SpellLink';
 import { suggestion } from 'parser/core/Analyzer';
 import { AnyEvent } from 'parser/core/Events';
-import aplCheck, { build, CheckResult, PlayerInfo, tenseAlt } from 'parser/shared/metrics/apl';
+import aplCheck, { Apl, build, CheckResult, PlayerInfo, tenseAlt } from 'parser/shared/metrics/apl';
 import annotateTimeline from 'parser/shared/metrics/apl/annotate';
 import { AplRuleProps } from 'parser/shared/metrics/apl/ChecklistRule';
 import {
   and,
+  buffMissing,
   buffPresent,
+  buffRemaining,
   buffStacks,
   describe,
   hasResource,
   hasTalent,
   lastSpellCast,
   not,
+  or,
+  spellCooldownRemaining,
 } from 'parser/shared/metrics/apl/conditions';
 import {
   ASCENSION_ENERGY_MAX_ADDITION,
   BASE_ENERGY_MAX,
   INNER_PEACE_ENERGY_MAX_ADDITION,
 } from '../resources/EnergyCapTracker';
+import Combatant from 'parser/core/Combatant';
 
-const hasChi = (min: number) => hasResource(RESOURCE_TYPES.CHI, { atLeast: min });
-
-export const apl = build([
-  {
-    spell: SPELLS.TIGER_PALM,
-    condition: describe(
+const hasChi = (min: number) =>
+  describe(
+    or(
       and(
-        hasResource(RESOURCE_TYPES.CHI, { atMost: hasTalent(TALENTS.ASCENSION_TALENT) ? 4 : 3 }),
-        hasResource(RESOURCE_TYPES.ENERGY, {
-          // over 85% of max energy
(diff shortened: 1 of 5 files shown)


src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
```diff
@@ -118,22 +151,188 @@ export const apl = build([
...
+      ),
+    ),
+  },
+  {
+    spell: SPELLS.TIGER_PALM,
+    condition: describe(
+      and(
+        hasResource(RESOURCE_TYPES.ENERGY, {
+          atLeast: hasTalent(TALENTS.CELERITY_TALENT) ? 55 : 60,
+        }),
```
emallson: this won't work correctly. you have a couple of options here.

1. have your apl built by a function that takes the `combatant` as a parameter, then use `combatant.hasTalent(TALENTS.CELERITY_TALENT)` to check this. like `apl = (combatant: Combatant) => build([...]);`
2. use a 2nd rule, one with the 55 and `hasTalent`, and the other with 60 and `not(hasTalent)`

the reason for this is that `condition.hasTalent(...)` returns a *condition object*, not true/false. the condition object is used to check the condition at every event.

src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
```diff
@@ -118,22 +151,188 @@ export const apl = build([
...
+      and(
+        hasResource(RESOURCE_TYPES.CHI, { atMost: hasTalent(TALENTS.ASCENSION_TALENT) ? 4 : 3 }),
+        hasResource(RESOURCE_TYPES.ENERGY, {
+          // over 85% of max energy
+          atLeast:
+            (BASE_ENERGY_MAX +
+              (hasTalent(TALENTS.ASCENSION_TALENT) ? ASCENSION_ENERGY_MAX_ADDITION : 0) +
+              (hasTalent(TALENTS.INNER_PEACE_TALENT) ? INNER_PEACE_ENERGY_MAX_ADDITION : 0)) *
+            0.85,
+        }),
```
emallson: same note here about `hasTalent` usage. these cases both complex enough (in terms of how many options there are) and simple enough (only controlling a threshold for a `hasResource` check) that using the function approach seems better to me

src/analysis/retail/monk/windwalker/modules/apl/AplCheck.tsx
```diff
@@ -4,28 +4,53 @@ import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
...
 } from '../resources/EnergyCapTracker';
 
-const hasChi = (min: number) => hasResource(RESOURCE_TYPES.CHI, { atLeast: min });
+const hasChi = (min: number): Condition<number> => {
+  const resource = hasResource(RESOURCE_TYPES.CHI, {
+    atLeast:
+      hasTalent(TALENTS.ORDERED_ELEMENTS_TALENT) && buffPresent(TALENTS.STORM_EARTH_AND_FIRE_TALENT)
+        ? min - 1
+        : min,
+  });
```
emallson: very similar note here, except that you can't use `combatant.hasBuff` in this location because if you do it will check the buffs present at the start of the fight rather than at each event.

you could do it without relying on a `combatant` parameter like this, i think:

```tsx
const resource = or(
  and(hasTalent(TALENTS.ORDERED_ELEMENTS_TALENT), buffPresent(TALENTS.STORM_EARTH_AND_FIRE_TALENT), hasResource(RESOURCE_TYPES.CHI, { atLeast: min - 1 })),
  hasResource(RESOURCE_TYPES.CHI, { atLeast: min })
);
```

emallson (changes requested):

src/analysis/retail/monk/windwalker/modules/spells/HeartOfTheJadeSerpent.tsx
```diff
@@ -101,14 +101,16 @@ class HeartOfTheJadeSerpent extends HotJS {
         )}
         {this.currentSotwl !== 0 && (
           <>
+            <br />
```
emallson, line 104: note: generally should avoid `<br/>` for layout as it can produce awkward results on differing screen sizes. however, tooltips can be finicky. in the interest of timeliness i'm going to merge the APL changes, since this is super minor.

emallson (approved):

PR #7269, opened 2025-03-10
Adding subtlety Rogue support for 11.1.0

dboisvert:
### Description

Updated the subtlety rogue section based on the talent changes for 11.1.0

### Testing

I tested with my own local instance and everything appears to be running appropriately.

- Test report URL: `https://www.warcraftlogs.com/reports/WYaZm8M62Drqg39H/?fight=29&source=4`
- Screenshot(s):
image


src/CHANGELOG.tsx
```diff
@@ -35,12 +35,14 @@ import {
...
   Yellot,
   KYZ,
+  Dboisvert,
 } from 'CONTRIBUTORS';
 import { ItemLink } from 'interface';
 import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
+  change(date(2025, 3, 10), 'Added support for sublety rogue 11.1.0', Dboisvert),
```
emallson: this should be in the subtlety rogue change log in `src/analysis/retail/rogue/subtlety/CHANGELOG.tsx`

emallson:
im going to take a look at this PR this afternoon before it gets too much bigger

src/analysis/retail/rogue/shared/talents/ColdBlood.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    this.addEventListener(Events.fightend, this.adjustMaxCasts);
+
+    (options.abilities as Abilities).add({
+      spell: TALENTS.COLD_BLOOD_TALENT.id,
+      category: SPELL_CATEGORY.COOLDOWNS,
+      cooldown: this.cooldown,
+      gcd: {
+        base: 1000,
+      },
```
emallson: this should probably be a static gcd

```suggestion
      gcd: {
        static: 1000,
      },
```

src/analysis/retail/rogue/shared/talents/ColdBlood.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    this.addEventListener(
+      Events.damage.by(SELECTED_PLAYER).spell(TALENTS.COLD_BLOOD_TALENT),
+      this.onDamage,
+    );
+
+    this.addEventListener(Events.fightend, this.adjustMaxCasts);
+
+    (options.abilities as Abilities).add({
```
emallson, line 39: generally we'd add this to the abilities list directly, and just set `enabled: combatant.hasTalent(...)` on it. i think it is ok to leave it here if this is just being used to auto-enable it for each spec.

src/analysis/retail/rogue/shared/talents/ColdBlood.tsx
```diff
@@ -0,0 +1,109 @@
...
+      spell: TALENTS.COLD_BLOOD_TALENT.id,
+      category: SPELL_CATEGORY.COOLDOWNS,
+      cooldown: this.cooldown,
+      gcd: {
+        base: 1000,
+      },
+      castEfficiency: {
+        maxCasts: () => this.maxCasts,
```
emallson: i don't see what this is accomplishing? `maxCasts` overrides are only needed if the max casts is not derived from the cooldown divided by fight duration, such as spells that are tied to the presence of a buff or to target hp.

EDIT: oh i see it might be copied from `Flagellation`, but that code appears to be vestigial from how the ability worked previously. as it is, this is just reproducing the default behavior manually i think.

src/analysis/retail/rogue/subtlety/CombatLogParser.tsx
```diff
@@ -76,23 +95,21 @@ class CombatLogParser extends CoreCombatLogParser {
...
-    sepsis: Sepsis,
-    stealthAbilityFollowingSepsis: StealthAbilityFollowingSepsis,
     echoingReprimand: EchoingReprimand,
-    flagellation: Flagellation,
 
     // Covenants
 
     // Racials
-    arcaneTorrent: [ArcaneTorrent, { gcd: 1000 }] as const,
+    // arcaneTorrent: [ArcaneTorrent, { gcd: 1000 }] as const,
```
emallson: what is the reasoning for removing arcane torrent?

src/analysis/retail/rogue/subtlety/modules/spells/SymbolsOfDeath.tsx
```diff
@@ -0,0 +1,191 @@
...
+    let details: JSX.Element;
+
+    if (energyAtCast === 100 && !hasSecretTechnique) {
+      performance = QualitativePerformance.Fail;
+      summary = <div>Major Energy Waste</div>;
+      details = (
+        <div>
+          You used <SpellLink spell={SPELLS.SYMBOLS_OF_DEATH} /> at **{energyAtCast}** energy
+          without **Secret Technique**. This means you **wasted {wastedEnergy} energy**. Try using
+          abilities before activating Symbols of Death!
```
emallson: note that markdown formatting is not supported here. you probably want to use `<strong>` instead to bold these properly:

```suggestion
          You used <SpellLink spell={SPELLS.SYMBOLS_OF_DEATH} /> at <strong>{energyAtCast}</strong> energy
          without <strong>Secret Technique</strong>. This means you <strong>wasted {wastedEnergy} energy</strong>. Try using
          abilities before activating Symbols of Death!
```

src/analysis/retail/rogue/subtlety/modules/spells/SymbolsOfDeath.tsx
```diff
@@ -0,0 +1,191 @@
...
+        </div>
+      );
+    } else if (isEnergyWasted) {
+      performance = QualitativePerformance.Good;
+      summary = <div>Energy Usage Could Be Improved</div>;
+      details = (
+        <div>
+          You used <SpellLink spell={SPELLS.SYMBOLS_OF_DEATH} /> with **{energyAtCast} energy**,
+          resulting in some wasted energy ({wastedEnergy}). Try to use abilities before activating
+          Symbols of Death to **optimize energy gain**.
```
emallson: same note about markdown

src/analysis/retail/rogue/subtlety/modules/spells/SymbolsOfDeath.tsx
```diff
@@ -0,0 +1,191 @@
...
+        </div>
+      );
+    } else {
+      performance = QualitativePerformance.Perfect;
+      summary = <div>Perfect Energy Management</div>;
+      details = (
+        <div>
+          You activated <SpellLink spell={SPELLS.SYMBOLS_OF_DEATH} /> with **optimal energy
+          management** ({energyAtCast}). Well played!
+        </div>
```
emallson: and again (and further down but i'm not going to flood the review with this)

src/analysis/retail/rogue/subtlety/modules/spells/Eviscrate.tsx
```diff
@@ -0,0 +1,62 @@
...
+import Events, { CastEvent } from 'parser/core/Events';
+import SPELLS from 'common/SPELLS/rogue';
+import { QualitativePerformance } from 'parser/ui/QualitativePerformance';
+import SpellLink from 'interface/SpellLink';
+import { HideGoodCastsSpellUsageSubSection } from 'parser/core/SpellUsage/HideGoodCastsSpellUsageSubSection';
+import { logSpellUseEvent } from 'parser/core/SpellUsage/SpellUsageSubSection';
+import CastPerformanceSummary from 'analysis/retail/demonhunter/shared/guide/CastPerformanceSummary';
+import { createSpellUse } from 'parser/core/MajorCooldowns/MajorCooldown';
+
+export default class Eviscerate extends Analyzer {
```
emallson: typo in the *file* name (`Eviscrate` should be `Eviscerate`).

src/analysis/retail/rogue/subtlety/modules/spells/Flagellation.tsx
```diff
@@ -0,0 +1,209 @@
...
+    const hasBladesAvailable = this.spellUsable.isAvailable(TALENTS.SHADOW_BLADES_TALENT.id);
+    const hasSymbolsAvailable =
+      (this.spellUsable.chargesAvailable(SPELLS.SYMBOLS_OF_DEATH.id) === 1 &&
+        this.spellUsable.cooldownRemaining(SPELLS.SYMBOLS_OF_DEATH.id, event.timestamp) < 10) ||
+      this.spellUsable.chargesAvailable(SPELLS.SYMBOLS_OF_DEATH.id) >= 2;
+    const hasShadowDanceAvailable =
+      (this.spellUsable.chargesAvailable(SPELLS.SHADOW_DANCE.id) === 1 &&
+        this.spellUsable.cooldownRemaining(SPELLS.SHADOW_DANCE.id, event.timestamp) < 40) ||
+      this.spellUsable.chargesAvailable(SPELLS.SHADOW_DANCE.id) === 2;
+    // const hasSecretTechniqueAvailable = true;
```
emallson: dead comment

src/analysis/retail/rogue/subtlety/modules/spells/Rupture.tsx
```diff
@@ -0,0 +1,233 @@
...
+          <div>
+            You used <SpellLink spell={SPELLS.RUPTURE} /> at {comboPointsAtCast} combo points. Try
+            to use it at 6 or more CP for a finisher**.
+          </div>
+        ),
+      },
+    );
+  }
+
+  private rupturePerformance(
```
emallson, line 114: this method is *really* complicated. the extended (50 line!) chained ternary is probably not maintainable long-term. this method should be restructured a bit to correct that.

for example:

- first check if you are dealing with the initial rupture cast. if so, provide output for that via `return`, exiting the method.
- if that isn't true, handle the gap case. 
- finally, handle the early refresh case

in each case, you can likely simplify your return into a single JSX output that uses variables like `refreshTIme` to fill in the text instead of manually writing out all 9 cases.

src/analysis/retail/rogue/subtlety/modules/spells/ShadowBlades.tsx
```diff
@@ -0,0 +1,135 @@
...
+            {hasShadowDanceBuff ? (
+              <>
+                ✔ <SpellLink spell={SPELLS.SHADOW_DANCE} /> buff was present.
+              </>
+            ) : (
+              <>
+                <SpellLink spell={SPELLS.SHADOW_DANCE} /> buff was not present.
+              </>
+            )}
+            <br />
```
emallson: avoid using `<br/>` for formatting, as it behaves badly across different screen sizes.

src/analysis/retail/rogue/subtlety/modules/spells/ShadowBlades.tsx
```diff
@@ -0,0 +1,135 @@
...
+      'flagellation_alignment',
+      { event },
+      {
+        performance,
+        summary: <div>Buff Alignment</div>,
+        details: (
+          <div>
+            {hasShadowDanceBuff ? (
+              <>
+                ✔ <SpellLink spell={SPELLS.SHADOW_DANCE} /> buff was present.
```
emallson: these should *probably* be separate checks on the item, rather than using the UTF checkmark like this.

emallson (changes requested):
its good to see sub getting attention for the first time in a long time, but there are some issues to address.

generally, the code looks fine. the issues are primarily stylistic / organizational, not bugs (at least, that i can see).

there is a broad issue with using markdown syntax where it isn't supported, like this:
image

i would *encourage* doing a pass on the formatting of the cast details. for example, this is what Flagellation looks like:
image

note how the content on the right "stretches" the gap between the performance markers on the left, making it unclear which is which. the use of inline ✔️ further confuses which performance mark goes with which text.

however, i don't want to get caught up pursuing perfect presentation before getting anything online. the things that i think are *necessary* to fix before merging are:

- the sub-specific changelog entry in the global changelog (just move it to the sub `CHANGELOG` file)
- the use of markdown syntax in guide text (convert to using HTML tags instead)
- the `rupturePerformance` method's structure

past that, this can be refined over time. getting this online to support sub in 11.1 will be good

src/analysis/retail/rogue/subtlety/modules/spells/Rupture.tsx
samuelmaverick: Got it! We’ll simplify it.

src/analysis/retail/rogue/shared/talents/ColdBlood.tsx
dboisvert: That was the idea on this one. Let the spec enable it.

src/analysis/retail/rogue/subtlety/modules/spells/Flagellation.tsx
```diff
@@ -135,72 +135,109 @@ export default class FlagellationAnalysis extends Analyzer {
...
+  private bladesPerformance(
+    event: CastEvent,
+    hasBladesAvailable: boolean,
+  ): ChecklistUsageInfo | undefined {
+    return createChecklistItem(
+      'flagellation_blades',
+      { event },
+      {
+        performance: hasBladesAvailable ? QualitativePerformance.Good : QualitativePerformance.Fail,
+        summary: <div>Shadow Blades Cooldown</div>,
```
emallson, line 152: for future ref, you can use react fragments instead of divs here (and in similar locations). don't need to change it at this point though

```suggestion
        summary: <>Shadow Blades Cooldown</>,
```

emallson (approved):
approved with some small changes (a lint fix + preventing some broken text generation)

PR #7281, opened 2025-03-19
[Pres Evoker] DB no longer counts non-player targets

Don-Osipov:
### Description

DB was previously counting pets as valid target hits, which caused issues counting the number of targets hit in m+. Now only counts player targets

### Testing

- Test report URL: /reports/nBJXYDWVryRNQmZb?fight=77&type=summary
- Screenshot(s):
Logs show targets hit as 3, then 4, then 5. Analyzer now reflects that and doesn't count shadow fiend.
image
image


emallson (approved):

emallson:
seems like a simple enough fix. sneaky error, too

PR #7283, opened 2025-03-21
fix: correct cooldown on Xuen and SotWL based on chosen talents

smvoss:
### Description

Analyzer was not correctly considering the cooldown reduction on Xuen / Strike of the Windlord based on chosen talents.

### Testing

http://localhost:3000/report/9bxVkfYFXyT7jzdC/60-Mythic+Cauldron+of+Carnage+-+Kill+(6:37)/Protmonkey/standard/overview

- Screenshot(s):

before: 
image

after: 
image

### DIFF
--- a/src/analysis/retail/monk/windwalker/CHANGELOG.tsx
+++ b/src/analysis/retail/monk/windwalker/CHANGELOG.tsx
@@ -5,6 +5,7 @@ import { Durpn, emallson } from 'CONTRIBUTORS';
 import SpellLink from 'interface/SpellLink';
 
 export default [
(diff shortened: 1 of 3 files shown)


vohrr:
There is a HeartofTheJadeSerpent module in monk/shared. Is it possible to merge them, or is there a reason this was implemented separately?

smvoss:
Good call out, I actually meant to inherit from that one when I originally implemented the ww-specific deviations. Forgetting to do that is why I didn't automatically get the heart reductions and ultimately why I ended up adding them here.

That's the more appropriate way to handle it, and I'll look to converge. I had reached out to @trevorm4 about whether it made sense to put the ww specific logic (execution of window) into the shared one and we agreed it didn't.

trevorm4:
Ya I think that inheriting should be fine from an organizational perspective as imo only should put shared logic in the shared file.

smvoss:
Taking a look back at the inheritance because of these comments made me realize I actually already did inherit, and the crux of the impactful changes here is actually correctly setting the cooldown of Xuen and SotWL

PR #7306, opened 2025-03-28
[Enhancement] fixed some interactions with Whirling Fire

Spruudel:
### Description

Because the Hot Hand application event happens after the inital (already buffed) Lava Lash that triggered it, that Lava Lash was not tracked as part of that window.

Also, a Hot Hand triggered by Whirling Fire should not refund the CD of the initial Lava Lash the way a natural Hot Hand would.

To fix this, I added an `EventLink` between each Hot Hand `ApplyBuff` and the potential related Whirling Fire `RemoveBuff` and Lava Lash cast that triggered it. The `forwardBufferMs` and `backwardBufferMs` values I chose seemed to work fine when I tested a couple logs.
The Lava Lash CD is then only reset if it wasn't a Whirling Fire proc and the missing Lava Lash cast (+ GCD) is manually added.

I think I accounted for everything, but I don't think I can rule out any sideeffects given how new I am to the codebase.

### Testing

- Test report URL: `/report/Y4TDJ3FhpXrgGj9W/24-Mythic+Sprocketmonger+Lockenstock+-+Kill+(6:28)/Coneform/standard/overview`
- Log with event filters: [Log](https://www.warcraftlogs.com/reports/Y4TDJ3FhpXrgGj9W?fight=24&type=auras&source=5&view=events&pins=2%24Off%24%23244F4B%24auras-gained%24-1%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24215785%2435%5E0%24Separate%24%23909049%24auras-gained%24-1%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24453405%2436%5E0%24Off%24%23a04D8A%24expression%24%28IN+RANGE+FROM+type+%3D+%27applybuff%27+AND+ability.name+%3D+%27Whirling+Fire%27+TO+type+%3D+%27removebuff%27+AND+ability.name+%3D+%27Whirling+Fire%27+GROUP+BY+target+ON+source+END%29+AND+%28ability.id+IN+%2860103%29%29&translate=true)

#### CD reset
grafik

grafik

#### Hot Hands guide
grafik

grafik

### DIFF
--- a/src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
+++ b/src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
@@ -2,7 +2,7 @@ import { Options } from 'parser/core/Analyzer';
 import BaseEventLinkNormalizer, { EventLink } from 'parser/core/EventLinkNormalizer';
 import SPELLS from 'common/SPELLS';
 import TALENTS from 'common/TALENTS/shaman';
-import { EventType } from 'parser/core/Events';
+import { ApplyBuffEvent, EventType, GetRelatedEvent } from 'parser/core/Events';
 import { NormalizerOrder } from './constants';
 import {
   EnhancementEventLinks,
@@ -124,6 +124,7 @@ const whirlingFireHotHandLink: EventLink = {
   linkingEventType: EventType.ApplyBuff,
   referencedEventId: SPELLS.WHIRLING_FIRE.id,
   referencedEventType: EventType.RemoveBuff,
+  reverseLinkRelation: EnhancementEventLinks.WHIRLING_FIRE_LINK,
   forwardBufferMs: 5,
 };
 const whirlingFireLavaLashLink: EventLink = {
@@ -134,6 +135,18 @@ const whirlingFireLavaLashLink: EventLink = {
   referencedEventType: EventType.Cast,
   backwardBufferMs: EventLinkBuffers.CAST_DAMAGE_BUFFER,
   anyTarget: true,
+  additionalCondition: (le, _) => {
+    if (le.type === EventType.RemoveBuff && le.ability.guid === SPELLS.WHIRLING_FIRE.id) {
+      return (
+        GetRelatedEvent<ApplyBuffEvent>(
+          le,
+          EnhancementEventLinks.WHIRLING_FIRE_LINK,
+          (e) => e.type === EventType.ApplyBuff && e.ability.guid === SPELLS.HOT_HAND_BUFF.id,
+        ) !== undefined
+      );
+    }
+    return false;
+  },
 };
 
 class EventLinkNormalizer extends BaseEventLinkNormalizer {
--- a/src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
+++ b/src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
@@ -122,9 +122,11 @@ const whirlingFireLink: EventLink = {
   linkRelation: EnhancementEventLinks.WHIRLING_FIRE_LINK,
   linkingEventId: SPELLS.HOT_HAND_BUFF.id,
   linkingEventType: EventType.ApplyBuff,
-  referencedEventId: SPELLS.WHIRLING_FIRE.id,
-  referencedEventType: EventType.RemoveBuff,
-  forwardBufferMs: EventLinkBuffers.WHIRLING_FIRE_BUFFER,
+  referencedEventId: [SPELLS.WHIRLING_FIRE.id, TALENTS.LAVA_LASH_TALENT.id],
+  referencedEventType: [EventType.RemoveBuff, EventType.Cast],
+  forwardBufferMs: 5,
+  backwardBufferMs: EventLinkBuffers.CAST_DAMAGE_BUFFER,
+  anyTarget: true,
 };
 
 class EventLinkNormalizer extends BaseEventLinkNormalizer {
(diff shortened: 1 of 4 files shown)


Spruudel:
I think I misunderstood how Whirling Fire works. While the first Lava Lash that triggers Hot Hand does trigger Reactivity, it is missing the damage buff from Hot Hand and is technically not part of the window.

Depending on if it's desired not to include it, only the first commit which fixes the CD issue can be merged

(edit: probably not a big deal, just a tiny technicality where I don't want to make the decision. I'd prefer to include it)

emallson (commented):
code looks fine to me, but i don't understand enh very well. @Seriousnes if you have a chance to give a 2nd opinion?

src/analysis/retail/shaman/enhancement/modules/talents/HotHand.tsx
```diff
@@ -207,19 +208,41 @@ class HotHand extends MajorCooldown<HotHandProc> {
...
 
+      // make sure to include first Lava Lash of the window, when triggered by Whirling Fire
+      let lavaLashCastEvent: CastEvent | undefined;
+      if (whirlingFireRemovedEvent) {
+        lavaLashCastEvent = GetRelatedEvent<CastEvent>(
+          event,
+          EnhancementEventLinks.WHIRLING_FIRE_LINK,
+          (e) => e.type === EventType.Cast,
+        );
+      }
```
Seriousnes: We can add additional information to the timeline to indicate this Hot Hand was triggered by Whirling Fire. E.g.
```suggestion
      let lavaLashCastEvent: CastEvent | undefined;
      if (whirlingFireRemovedEvent) {
        lavaLashCastEvent = GetRelatedEvent<CastEvent>(
          event,
          EnhancementEventLinks.WHIRLING_FIRE_LINK,
          (e) => e.type === EventType.Cast,
        );
        lavaLashCastEvent && addAdditionalCastInformation(
          lavaLashCastEvent,
          <>
            <SpellLink spell={TALENTS.HOT_HAND_TALENT} /> was applied by{' '}
            <SpellLink spell={SPELLS.WHIRLING_FIRE} />
          </>,
        );
      }
```
Or `addEnhancedCastReason` if you think it's more appropriate.

src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
```diff
@@ -118,6 +118,16 @@ const sunderingDamageLink: EventLink = {
...
+const whirlingFireLink: EventLink = {
+  linkRelation: EnhancementEventLinks.WHIRLING_FIRE_LINK,
+  linkingEventId: SPELLS.HOT_HAND_BUFF.id,
+  linkingEventType: EventType.ApplyBuff,
+  referencedEventId: [SPELLS.WHIRLING_FIRE.id, TALENTS.LAVA_LASH_TALENT.id],
+  referencedEventType: [EventType.RemoveBuff, EventType.Cast],
+  forwardBufferMs: 5,
+  backwardBufferMs: EventLinkBuffers.CAST_DAMAGE_BUFFER,
+  anyTarget: true,
+};
```
Seriousnes, line 151: The ordering seems pretty consistent: Lava Lash Cast > Hot Hand Applied > Whirling Fire (WF) Removed.

I think this should be split into two different EventLinks; HH applied linking to WF removed, and another for WF removed to LL cast. They can use the same link relation, but for actual linking they should probably be separated.

HH seems to be applied within 5 ms of WF being removed, and LL cast is typically no more than 25ms before WF is removed. 100ms for either feels too lenient to start.

Seriousnes (approved):
Looks good. Two suggestions which you can do what you like with.

This highlights some other issues with Hot Hand that have changed since DF, which I can work on in a separate PR.

src/analysis/retail/shaman/enhancement/modules/normalizers/EventLinkNormalizer.tsx
Spruudel: Having two EventLinks sounds good; regarding the LL timing this is the biggest time difference I was able to find by just randomly trying a few logs. The damage event seemed to be always very close to the WF removal, so I assumed `CAST_DAMAGE_BUFFER` would be the usually used value in this situation

[Log](https://www.warcraftlogs.com/reports/8vbwaNRMDn3qcy9F?fight=61&type=auras&source=33&view=events&pins=2%24Off%24%23244F4B%24auras-gained%24-1%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24215785%2435%5E0%24Separate%24%23909049%24auras-gained%24-1%240.0.0.Any%240.0.0.Any%24true%240.0.0.Any%24true%24453405%2436%5E0%24Off%24%23a04D8A%24expression%24%28IN+RANGE+FROM+type+%3D+%27applybuff%27+AND+ability.name+%3D+%27Whirling+Fire%27+TO+type+%3D+%27removebuff%27+AND+ability.name+%3D+%27Whirling+Fire%27+GROUP+BY+target+ON+source+END%29+AND+%28ability.id+IN+%2860103%29%29&translate=true)

grafik

Seriousnes: I've used `CAST_DAMAGE_BUFFER` for spell casts such as lightning bolt, chain lightning, tempest, etc to link the cast to it's damage events.

If you've found one with a 45ms difference, we can try using `CAST_DAMAGE_BUFFER`'s 100ms, however I think we'll need to put an additional restriction on the WF -> Lava Lash.

`EventLinks` are evaluated in the order they're passed to the constructor, so we should **a)** make the link from WF removal to HH applied _first_ and the WF removal to LL cast second, and **b)** add an additional condition to the WF/LL link to check that a WF/HH link exists.

E.g. Something like this
```
const whirlingFireHotHandLink: EventLink = {
  ...,
  reverseLinkRelation: EnhancementEventLinks.WHIRLING_FIRE_LINK, // make the link available in both directions, needed for whirlingFireLavaLashLink
};
const whirlingFireLavaLashLink: EventLink = {
  ...,
  additionalCondition: (le, _) => {
    if (le.type === EventType.RemoveBuff && le.ability.guid === SPELLS.WHIRLING_FIRE.id) {
      return GetRelatedEvent<RemoveBuffEvent>(le, EnhancementEventLinks.WHIRLING_FIRE_LINK, (e) => e.type === EventType.RemoveBuff) !== undefined
    }
    return false;
  }
}
```

PR #7307, opened 2025-03-29
[Discipline] More updates for 11.1

Saeldur:
### Description
Fix some things Hana missed. Still not entirely up to date but hopefully this makes their life a bit easier.

* Add Twinsight Spellids
* Fix Pain and Suffering Multiplier
* Fix Penance bolts count
* Add Twinsight spells to Scov
* Haste Penance & PWS CDs (and Mindblast)
* Remove unused Constants.
* Update some hotfixed Changes.
* Fix Dot Spread
* Fix Death Throes Mana

### Testing

- Test report URL: `/reports/tpgM1azdPw7JCNRK?fight=18&type=healing&source=38` `/reports/C2TYBkbNR8Qqdgw3?fight=2&type=summary&source=90`
- Screenshot(s):


src/analysis/retail/priest/discipline/modules/features/AtonementHealingBreakdown.tsx
```diff
@@ -53,7 +67,7 @@ const AtonementHealingBreakdown = ({
                 <tr key={ability.guid}>
                   <td style={{ width: '30%' }}>
                     <SpellLink spell={abilityToSpell(ability)} icon={false}>
-                      <Icon icon={ability.abilityIcon} /> {ability.name}
+                      <Icon icon={ability.abilityIcon} /> {getAbilityNameReporting(ability)}
```
emallson: whats the story here? and does this become simpler by using `<SpellLink spell={ability.guid} />` instead to have the spell data from our internal list looked up?

src/analysis/retail/priest/discipline/modules/spells/BlazeOfLight/BlazeOfLightSourceDonut.tsx
```diff
@@ -33,7 +36,7 @@ function generateHealingItems(
     return {
       color: `${COLORS[ability.guid]}`,
       spellId: ability.guid,
-      label: ability.name,
+      label: getAbilityNameReporting(ability),
```
emallson: this could be `getSpell(ability.guid).name`

Saeldur: Probably a better solution. Wasn't aware of this being valid.
There are duplicated spells with different ids but same name on logs that I wanted to differentiate on the report to make it less confusing to read, in this case I differentiated it inside of the spell list so this would actually work fine.

src/analysis/retail/priest/discipline/modules/features/AtonementHealingBreakdown.tsx
Saeldur: Unsure, I just adapted what was there. Changed it to this.

src/analysis/retail/priest/discipline/modules/spells/Penance.jsx
oisinod, line 1: I'm not sure I'm a fan of counting twinsight bolts as penance bolts for the purpose of this module. 

But if we do - we would need to increase the expected bolts by the twinsight amount. I'd rather just keep it at the actual casted bolts. 

Which do you prefer?

oisinod (changes requested):
All good other than one comment on the Penance module

src/analysis/retail/priest/discipline/modules/spells/Penance.jsx
Saeldur: The expected bolts was already getting increased by the original code, which was causing it to say we were missing like 173 penance bolts... I just added the missing elements (tracking the twinsight bolts themselves). Although I assume that was down to necessity due to being written before blizz split the Spell IDs.
I'm unsure on whether its worth tracking them - As long as Penance ticks three times the twinsight bolts will still occur but if you do cancel it before the third tick and lose the Twinsight bolts as well as your regular bolts it should probably tell you, as thats extra bad.
Either way works, given its mainly trying to tell you that you are cutting penance too short and it does not do any logic on the damage/healing itself in this module.

PR #7324, opened 2025-04-17
Update Outlaw Rogue for 11.1

Krealle:
### Description

Updates Outlaw rogue to support 11.1.

Outlaw currently has 2 Meta builds, `Hidden Opportunity` and `Keep it Rolling` - which this PR mainly focuses on.

- Updated `Abilities`
- Disabled `Check List`
- Added `buffChannelSpec` for `Killing Spree`
- Added `Downtime Foundation` section
- Added `Cooldown Graph` section
- Updated `APLCheck` module
- Updated `Finisher` analysis
- Updated `Builder` analysis
- Added `Adrenaline Rush` module
- Cleaned up some unwanted `RefreshBuff` events
- Handle `Between the Eyes` CD reset provided from `Crackshot` talent
- Update CDR handling for `Restless Blades` & `Float Like a Butterfly`

Also refactors the `hasResources` apl condition slightly to better handle resources for builders.
This is necessary since the builders for Rogue have the `ResourceChange` trigger before the `Cast` event, and such would always be wrongly evaluated. 
I've added a `getResourceBeforeCast` arg that can be used to instead return the previous value, ie. before the resource was generated. `getResourceBeforeCast` gets ignored if the `Cast` event has the proper resource props in it, since those can be used to properly evaluate resources. These are generally only carried in spenders.
Ideally we would like to access the `comboPointTracker` analyzer to get the resource values, but that would require the APLCheck module to be reworked into a proper analyzer.

### Testing

Hidden Opportunity: 
`/report/rjcWD3NgPtMaf6Qb/2-Mythic+One-Armed+Bandit+-+Kill+(6:47)/Sjakal/standard/overview`

image

Keep it Rolling: 
`/report/gb1jYWaRxCMdtcAJ/26-Mythic+Sprocketmonger+Lockenstock+-+Kill+(5:58)/Vaiolation/standard/overview`

image


src/analysis/retail/rogue/outlaw/CombatLogParser.ts
```diff
@@ -21,7 +21,7 @@ import OutlawComboPointTracker from './modules/core/OutlawComboPointTracker';
 import OutlawEnergyCapTracker from './modules/core/OutlawEnergyCapTracker';
 import RestlessBlades from './modules/core/RestlessBlades';
 import AlwaysBeCasting from './modules/features/AlwaysBeCasting';
-import Checklist from './modules/features/Checklist/Module';
+//import Checklist from './modules/features/Checklist/Module';
```
emallson: should delete instead of commenting out

src/analysis/retail/rogue/outlaw/Guide.tsx
```diff
@@ -68,22 +79,29 @@ function ResourceUsageSection({ modules, info }: GuideProps<typeof CombatLogPars
...
           <RoundedPanel>{modules.finisherUse.chart}</RoundedPanel>
         </SideBySidePanels>
         <p></p>
-        <p>
+        {/* <p>
           <Trans id="guide.rogue.outlaw.sections.resources.comboPoints.buildersBreakdown">
             -- WIP section -- Higlight which builders the user is commonly overcapping with.
           </Trans>
-        </p>
+        </p> */}
```
emallson: is this still planned? if not, delete

src/analysis/retail/rogue/outlaw/Guide.tsx
```diff
@@ -149,3 +164,63 @@ function ActionPriorityList({ modules, info }: GuideProps<typeof CombatLogParser
...
+          </li>
+          <li>
+            <strong>Yellow segments</strong> indicate cooldown time.
+          </li>
+          <li>
+            <strong>Red segments</strong> highlight areas where an extra cooldown could have fit.
+          </li>
+        </ul>
+        {/* For Subtlety, <strong>Symbols of Death</strong> and <strong>Shadow Dance</strong> usage is
+        crucial, as they define your burst windows. */}
```
emallson: delete

src/analysis/retail/rogue/outlaw/modules/core/BuilderUse.tsx
```diff
@@ -96,6 +78,25 @@ export default class BuilderUse extends Analyzer {
...
 
+  private pistolShotUsage(cpAtCast: number) {
+    if (this.hasKeepItRolling) {
+      return (this.selectedCombatant.hasBuff(SPELLS.BROADSIDE) && cpAtCast <= 1) || cpAtCast <= 3;
+    }
+
+    return cpAtCast <= 4;
+  }
+
+  private ambushShotUsage(cpAtCast: number) {
```
emallson: should this be `ambushUsage`?

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -86,32 +69,12 @@ export default class FinisherUse extends Analyzer {
...
-      cpsSummary = <div>Spent &gt;= {targetCps} CPs</div>;
-      cpsDetails = <div>You spent {cpsSpent} CPs.</div>;
-    }
-
-    //- Finisher choice performance
-    const checklistFinisherChoice: ChecklistUsageInfo = {
+    // TODO: Finisher choice performance
+    // Determine if the proper finisher was used according to a priority list
+    // Can mostly just rely on the APLCheck for that for now
+    /* const checklistFinisherChoice: ChecklistUsageInfo = {
```
emallson: should *probably* just delete, since it'll need to change anyway for updated analysis right?

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -121,288 +84,266 @@ export default class FinisherUse extends Analyzer {
...
+        this.spellUses.push(createSpellUse({ event }, [comboPointPerformance]));
         break;
     }
+  }
+
+  private stealthPerformance(event: CastEvent, shouldBeInStealth?: boolean) {
+    const isInStealth = this.finishers.isInStealth();
+
+    if (shouldBeInStealth) {
+      createChecklistItem(
```
emallson: missing `return` here? i think this branch isn't reached currently so the bug isn't observable

src/analysis/retail/rogue/outlaw/modules/core/FinisherUse.tsx
```diff
@@ -121,288 +84,266 @@ export default class FinisherUse extends Analyzer {
...
-      event as TargettedEvent<any>,
+  get guide(): JSX.Element {
+    const explanation = (
+      <>
+        <p>
+          <strong>Finishers</strong> should typically be used at <strong>6 or more</strong> combo
+          points.{' '}
+          {this.selectedCombatant.hasTalent(TALENTS.SUBTERFUGE_TALENT) && (
+            <>
+              When inside of <SpellLink spell={SPELLS.SUBTERFUGE_BUFF} /> <strong>Finishers</strong>{' '}
```
emallson:
```suggestion
              When inside of <SpellLink spell={SPELLS.SUBTERFUGE_BUFF} />, <strong>Finishers</strong>{' '}
```

small grammar tweak

src/analysis/retail/rogue/outlaw/modules/features/Finishers.tsx
```diff
@@ -18,10 +23,42 @@ class Finishers extends FinisherTracker {
     return '';
   }
 
+  isInStealth() {
+    return (
+      this.selectedCombatant.hasBuff(
+        SPELLS.SUBTERFUGE_BUFF.id,
+        null,
+        0,
+        BUFF_REACTION_TIME_LATENCY,
```
emallson, line 32: it *probably* won't matter but using this parameter forces the `hasBuff` call to take the `O(events)` slow path instead of the `O(1)` fast path. if there are performance issues (especially in m+), this would be something to look at.

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
```diff
@@ -101,56 +79,59 @@ const COMMON_COOLDOWN: Rule[] = [
...
+    ),
+  },
+  {
+    spell: SPELLS.VANISH,
+    condition: and(
+      buffPresent(TALENTS.ADRENALINE_RUSH_TALENT),
+      hasFinisherCondition(),
+      notInStealthCondition(),
+    ),
+  },
```
emallson: *generally* shouldn't include cooldowns in the apl in case people are holding them for reasons. however i'm not really sure if vanish counts as a cooldown for this? since you're only casting it during AR already, so if you always use Vanish during AR it seems fine?

emallson: although AR itself is barely a cooldown for outlaw i guess

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
```diff
@@ -30,55 +32,31 @@ import { ROLL_THE_BONES_BUFFS } from '../../constants';
...
-
-//  /!\ Not sure this is working
-//  minEnergyThreshold will be the hard enforced energy rule while maxEnergyThreshold is the leeway accorded to the user
-const energyCondition = (minEnergyThreshold: number, maxEnergyThreshold: number) => {
-  return or(
-    hasResource(RESOURCE_TYPES.ENERGY, { atMost: minEnergyThreshold }),
-    optionalRule(hasResource(RESOURCE_TYPES.ENERGY, { atMost: maxEnergyThreshold })),
+  return describe(
+    hasResource(RESOURCE_TYPES.COMBO_POINTS, { atLeast: 6 }, undefined, usePrevAmount),
+    (tense) => <>the finisher condition {tenseAlt(tense, 'is', 'was')} met</>,
```
emallson: should explain what "the finisher condition" is (i know the text is unchanged from before, but it is simple enough to change to something like `<>you {tenseAlt(tense, 'have', 'had')} 6+ CP</>`

emallson (changes requested):
the code itself looks good. a good number of "delete instead of comment out" comments, as well as a couple of other minor things.

however, checking the sample log i still see quite a few cooldown tracking issues in the debug view ([link](http://localhost:3000/report/rjcWD3NgPtMaf6Qb/2-Mythic+One-Armed+Bandit+-+Kill+(6:47)/Sjakal/standard/debug)). are these fixable? or due to difficult/impossible to track resets?
image

(there are some gcd issues too but it looks primarily like Feint needs its gcd fixed)

src/analysis/retail/rogue/outlaw/Guide.tsx
Krealle: I did actually put a todo in the `BuilderUse` module, so will remove this.

Krealle:
> the code itself looks good. a good number of "delete instead of comment out" comments, as well as a couple of other minor things.
> 
> however, checking the sample log i still see quite a few cooldown tracking issues in the debug view ([link](http://localhost:3000/report/rjcWD3NgPtMaf6Qb/2-Mythic+One-Armed+Bandit+-+Kill+(6:47)/Sjakal/standard/debug)). are these fixable? or due to difficult/impossible to track resets? image
> 
> (there are some gcd issues too but it looks primarily like Feint needs its gcd fixed)

image

Completely missed that page existed actually, very handy! 
The CDR issues were because of how the extra (hidden) combo points spent from Super Charged/Coup De Grace, weren't being accounted for. Added some logic for that and it seems to have cleaned it up.
On the example log there is one instance of Adrenaline Rush and Sprint being used with a few seconds left on CD, but can't really figure out why that is, and looks fine for other fights.

The remaining GCD issue is with Killing Spree, but gonna assume that's just because of the channeling normalizer.

emallson:
that looks much better!

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
Krealle: Yea looking at it a bit more, I think it might be a good idea to exclude them for now. Having them in does makes the checker punish top logs too harshly.

src/analysis/retail/rogue/outlaw/modules/apl/AplCheck.tsx
Krealle: Cleaned up the APL check, should be more clear now.

emallson (approved):

PR #7326, opened 2025-04-17
[Core] Add patch 11.1.5

Krealle:
### Description

Adds patch 11.1.5.

Bumped all specs that are currently supported for 11.1.0, since the relevant changes are fairly minor.

### DIFF
--- a/e2e/poms/fight-selection-page.ts
+++ b/e2e/poms/fight-selection-page.ts
@@ -11,7 +11,6 @@ export class FightSelectionPage {
 
   async goto(reportCode: string) {
     await this.page.goto(`/report/${reportCode}`);
-    await this.expectFightSelectionHeaderToBeVisible();
   }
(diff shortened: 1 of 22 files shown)


ToppleTheNun:
Can you bump this as well: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/game/VERSIONS.ts#L7

emallson:
i believe the e2e failure is due to this info needing to be updated to an 11.1.5 log: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/8161b383d0982fce92ba3dda5fcece89f40a7d03/e2e/report-selection.spec.ts#L3-L12

we should potentially just change that test suite to handle the "previous patch" dialog

emallson:
doing the close + reopen thing to kick CI

PR #7327, opened 2025-04-18
Preparation Gem Subsection

CericME:
### Description

Been looking at guides and character tools since beginning of War Within and none of them told me I could have had more gem slots.  Saw the Preparation Section on WowAnalyzer and thought to myself this would be a good place for something like that.

I based this off the Enchant section.  The Gem Checker goes through the gear that can have gems, check if they have gem slots, if those slots are filled, can a slot be added, and are the gems filled in those slots of the highest crafting quality.

A piece of gear that has:

-   all gem slots filled with Max Crafted Gems is marked perfect.
-   all gem slots filled is marked good.
-   has a gem slot filled but not all ok.
-   No gem slots filled fail.

Pieces that can have slots added are considered to always have that number of slots even if in their current form they don't for the sake of above.  Gear indication is the background block.

A Gem is:

-   Perfect if it's 3 Star.
-   Good if it's 2 Star.
-   Ok if it's 1 Star or not able to be found in the known gem list. (Cyrce's Circlet Gems end up here.)

Gem quality is the border around the gem icon.

Currently Recommended Gems are not supported but, there is scaffolding for in the future once I figure out how Enchants get Recommended Enchants passed through.

localization/en/message.json does have values for the few messages that are produced.  I like to do that for the main description blob couldn't figure it out and I don't want perfect to get in the way of the good.

I tried to conform to the standards that I saw within code.  Currently only the War Within is supported.

### Testing

Bring up a log like you normally would in WowAnalyzer.
Go to a character.
Scroll to the Preparation section.
Third Subsection should be Gem and the slots with Gems or can have a gem slot added will be showing.

- Test report URL: https://www.warcraftlogs.com/reports/3Zb2BHt9cWNCfawP/?fight=18&source=19
- Screenshot(s): 
image
image
image
image

This is actually the log that prompted me to do this work.

- Test report URL: https://www.warcraftlogs.com/reports/DZQty6qwXnLjWmrG/?fight=17&source=2
- Screenshot(s):  
image

Later Raid with all the gems.  Except the Great Vault ones.  It's a bit early for that.


src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+class GemChecker extends Analyzer {
+  //Which slots can have a gem slots added
+  static twoAddableGemSlots = [GEAR_SLOTS.NECK, GEAR_SLOTS.FINGER1, GEAR_SLOTS.FINGER2];
+  static oneAddableGemSlot = [GEAR_SLOTS.HEAD, GEAR_SLOTS.WAIST, GEAR_SLOTS.WRISTS];
+
+  get GemableSlots(): Record<number, JSX.Element> {
+    return {}; //Keeping form with EnchantChecker and having the user override this
+  }
+
+  get GemableGear(): any {
```
Krealle: You shouldn't use the `any` type.
```suggestion
  get GemableGear(): Record<number, Item> {
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </Trans>,
+        );
+      } else {
+        tooltipContent.push(
+          <Trans id="shared.GemChecker.guide.FullyGemmed">
+            {slotName} slots are fully gemmed!
+          </Trans>,
+        );
+      }
+    }
```
Krealle: These conditions don't appear to work for non-jewelry items, causing the tooltips to be empty. You might need to update the bonus IDs for those, or simply check if a gem is present (prolly the ideal method).

src/interface/guide/components/Preparation/GemSubSection/GemBoxRow.tsx
```diff
@@ -0,0 +1,82 @@
+import styles from './GemBoxRow.module.scss';
+import { QualitativePerformance } from 'parser/ui/QualitativePerformance';
+import { CSSProperties, ReactNode } from 'react';
+import { Tooltip } from 'interface/index';
+import Item from 'common/ITEMS/Item';
+import Icon from 'interface/Icon';
+import { Gem as EventGem } from 'parser/core/Events';
```
Krealle: You are using two different `Item` types causing type errors. It looks like you want to use the one from `parser/core/Events` but you are importing the one from `common/ITEMS/Item` here.
```suggestion
import Icon from 'interface/Icon';
import { Gem as EventGem, Item } from 'parser/core/Events';
```

src/parser/core/itemHelper.ts
```diff
@@ -0,0 +1,26 @@
+import { Item } from 'parser/core/Events';
+
+/**
+ * Helper class for item-related operations that do not fit into the other classes.
+ */
+export class ItemHelper {
```
emallson: this doesn't need to be a class, it can just be a free-standing function with some local `const`s. like:

```ts
export const SINGLE_GEM_BONUS_ID = 10878;
export const DOUBLE_GEM_BONUS_ID = 10879;
export const TRIPLE_GEM_BONUS_ID = 10880;

export function hasBonusId(item: Item, bonusId: number): boolean {
  // existing function body
}
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+    slotName: JSX.Element,
+    slotNumber: number,
+    recommendedGems: GemItem[] | undefined,
+  ) {
+    const tooltipContent: JSX.Element[] = [];
+
+    //Note for Future me: <Trans id="..."> the text comes from locales/<Localization Code>/messages.json
+
+    //#region Special Cases
+    if (item.icon === 'inv_siren_isle_ring.jpg') {
```
emallson: use `item.id` to check for a specific item, not `item.icon`

src/parser/shared/modules/items/GemChecker.tsx
emallson: this can be simplified quite a bit:

```tsx
      // computing `maxGems` could be replaced with a method
      let maxGems = 0;

      if (hasBonusId(item, TRIPLE_GEM_BONUS_ID)) {
        maxGems = 3;
      } else if (hasBonusId(item, DOUBLE_GEM_BONUS_ID)) {
        maxGems = 2;
      } else if (hasBonusId(item, SINGLE_GEM_BONUS_ID)) {
        maxGems = 1;
      }

      const missingGems = maxGems - (item.gems?.length ?? 0);
      if (missingGems > 0) {
        tooltipContent.push(
          <Trans id="shared.GemChecker.MissingGemsMultiple">
            You are missing {missingGems} gems on your {slotName}.
          </Trans>,
        );
      } else {
        tooltipContent.push(
          <Trans id="shared.GemChecker.guide.FullyGemmed">
            {slotName} slots are fully gemmed!
          </Trans>,
        );
      }
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    //X Missing Setting
+    if (GemChecker.twoAddableGemSlots.includes(slotNumber)) {
+      const missingGems = 2 - (item.gems?.length ?? 0);
+      if (missingGems > 0) {
+        tooltipContent.push(
+          <Trans id="shared.GemChecker.MissingSlotsCraftable">
+            You are missing {missingGems} possible slot on your {slotName}.<br />
+            Craft/Buy Magnificent Jeweler's Setting to add a slot per.
+          </Trans>,
```
emallson: don't use `<br/>` for formatting. it doesn't play nicely with variable screen widths.

```suggestion
          <Trans id="shared.GemChecker.MissingSlotsCraftable">
            <p>You are missing {missingGems} possible slot on your {slotName}.</p>
            <p>Craft/Buy <ItemLink id={ITEMS.MAGNIFICENT_JEWELERS_SETTING.id} /> to add slots.</p>
          </Trans>,
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+      GemChecker.twoAddableGemSlots.includes(slot) ||
+      ItemHelper.hasBonusId(item, ItemHelper.DOUBLE_GEM_BONUS_ID)
+    ) {
+      return (item.gems?.length ?? 0) >= 2;
+    } else if (
+      GemChecker.oneAddableGemSlot.includes(slot) ||
+      ItemHelper.hasBonusId(item, ItemHelper.SINGLE_GEM_BONUS_ID)
+    ) {
+      return (item.gems?.length ?? 0) >= 1;
+    }
```
emallson: We should not mark missing sockets on Helm/Bracers/Belt as failures (IMO) since they aren't something that are particularly controllable until late in a season (when taking tokens from vault for a socket becomes likely).

Prior to that, players should be taking gear most of the time and only get sockets if they're unlucky (or if they're lucky and get one with a socket baked in)

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </Trans>,
+        ); //ItemLink in the future when I figure out how they work
+      }
+    } else if (GemChecker.oneAddableGemSlot.includes(slotNumber)) {
+      if ((item.gems?.length ?? 0) !== 1) {
+        tooltipContent.push(
+          <Trans id="shared.GemChecker.MissingSlotsVault">
+            You are missing a possible slot on your {slotName}. <br />6 Algari Token of Merit for
+            S.A.D. to add a slot.
+          </Trans>,
```
emallson: keeping a tooltip explaining how to get a guaranteed socket on your item does seem good to me, though. i would probably tooltip it as:

```ts
<Trans id="shared.GemChecker.MissingSlotsVault">
You do not have a gem socket in your {slotName}. If you don't have good items in your Vault, you can get <ItemLink id={ITEMS.ALGARI_TOKEN_OF_MERIT.id} /> instead and trade them for <ItemLink id={ITEMS.SOCKET_ADDITION_DEVICE.id} /> at the nearby vendor.
</Trans>
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+    ) {
+      return (item.gems?.length ?? 0) >= 1;
+    }
+
+    return true;
+  }
+
+  //#region UI
+  //Add a row for the actual Gem in the future to evaluate each
+  boxRowPerformance(item: Item, recommendedGems: number[] | undefined, slot: number) {
```
emallson: this being a separate method from the tooltip means that you both duplicate some logic and can't tie performance to tooltips. For example: both the "Perfect" color and the "Good" color have the same tooltip, but it isn't clear how to get to Perfect from Good. (I believe the answer is "use higher quality gems"?)

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    if ((item.gems ?? []).length > 0) {
+      item.gems!.forEach((iGem) => {
+        const lookupGem = gemById[iGem.id];
+
+        let tempQP = QualitativePerformance.Fail;
+        const gemRec = recommendedGems?.includes(iGem.id);
+
+        if (gemRec) {
+          tempQP = QualitativePerformance.Perfect;
```
emallson, line 141: this is one of those things that has grown organically but: we generally use the `Perfect` color for things that are exceptional enough that we can't reasonably expect players to accomplish them. Something like getting 100% cast efficiency.

For consistency, I'd recommend adjusting this to:

- Max Slots filled with Max Rank gems: Good
- Low-Rank gems: Ok
- Missing Slots on Jewelry or Empty Slots: Fail

emallson (changes requested):
this is a good PR that adds something which, in hindsight, we should have added a while ago. it isn't quite merge-able yet, but it isn't far off.

I noted this in a comment but I want to reiterate: we shouldn't mark helm/belt/bracers as fails for not having a socket, since there is often not an action the player can take to fix those in the near term---long term you'll eventually have 3 crappy vaults and get guaranteed sockets, but short term your options are "get lucky", "get a crappy vault", or "take a socket even if it isn't the best choice".

regarding the structure of `GemChecker`: you have a lot of duplication around handling 1, 2, and 3 socket bonus ids. additionally, the separation of performance from (tooltip) explanation has led to some awkwardness and further duplication.

i'd recommend doing some light refactoring there like adding a `getMaxSockets(item: Item): number` method that handles the existing bonus id / item slot checks and returns the max number of sockets that exist (probably excluding the helm/belt/bracers from having the "guaranteed" 1 socket), then you can use that to avoid repeating these checks for each different bonus id. as i noted in one of my comments

Krealle (commented):
Just gonna add a couple things I noticed as well.

src/parser/shared/modules/items/GemChecker.tsx
CericME: The itemId wasn't working for this item initially I can check in a version swapped.

src/parser/shared/modules/items/GemChecker.tsx
CericME: What would you recommend for a newline?

CericME: I see it.  You would suggest a paragraph wrap ok.  That use to be frowned on for this case.

src/parser/shared/modules/items/GemChecker.tsx
CericME: This breaks one of the core functions I wanted for this.  Which is to notify people that there are slots that can be added to the gear and slots that can be gemmed.

Right now really only have 4 states.  Perfect, good, ok, fail.  I really need a fifth stated for essentially this is the last thing you should do to your gear.

Maybe it be best to swap that particular on to ok.

src/parser/shared/modules/items/GemChecker.tsx
CericME: I like that to.  I actually started with the Items but there not in the item list that pulls from at the moment and for this initial one I didn't want to fully figure out how to get all the right data for the items.

src/parser/shared/modules/items/GemChecker.tsx
CericME: I thought about this and went back and forth.  What I originally did was add a tooltip to go with the fully hemmed that explained why it was not perfect.  I'll flesh that out again.

src/parser/shared/modules/items/GemChecker.tsx
CericME: That makes sense from a action perspective but from a gearing point of view I'm not sure.

Going to the early not failing when we don't have a slot from the vault.  This would essentially move everything down one.

If we went with this abbreviated scale I be inclined to move missing gem to fail and missing slot to ok.  That way the vault one missing would be ok.

src/parser/shared/modules/items/GemChecker.tsx
CericME: Cool, I was mimicking what I saw from enchanted trying to keep the coding style.  I love strong typing so no problem there.

src/parser/shared/modules/items/GemChecker.tsx
CericME: Do you have an example log I can poke at?

Earlier comments I went off of the presence of gems.  I moved away from that as I haven't been able to confirm that a empty slot would show as a gem array entry.

src/parser/shared/modules/items/GemChecker.tsx
emallson: showing the item slots in the gem section will still clue people into the fact that they can be gemmed. we could also potentially show them in a "disabled" state when these slots don't have sockets

src/parser/shared/modules/items/GemChecker.tsx
Krealle: `/report/rjcWD3NgPtMaf6Qb/2-Mythic+One-Armed+Bandit+-+Kill+(6:47)/Vollmer/standard` has gems in everything

src/interface/guide/components/Preparation/GemSubSection/GemBoxRow.tsx
CericME: That's on purpose.  More to do with the naming used in the codename itself.

There is the one from the event.  Then there is a common one and different parts pass something different.  I'm not at my computer but it was a bugger to track down and figure out.

Krealle: Are you certain?

Let's say `ItemA` is the one from `'parser/core/Events'` and `ItemB` is the one from `'common/ITEMS/Item'`

You are using the `ItemB` type for the `item` prop in the `GemBoxRowEntry` interface.
You then use `GemBoxRowEntry[]` as the return type for `getGemBoxRowEntries`.
You map the `GemBoxRowEntry[]` array from the record you get from `GemableGear`, which is typed `Record<number, ItemA>`.

```ts
get GemableGear(): Record<number, ItemA> {
  const gemSlots = this.GemableSlots;
  return Object.keys(gemSlots).reduce<Record<number, ItemA>>((obj, slot) => {
    const innerSlot = Number(slot);

    obj[innerSlot] = this.selectedCombatant._getGearItemBySlotId(innerSlot);

    return obj;
  }, {});
}

getGemBoxRowEntries(recommendedGems: Record<number, GemItem[]> = {}): GemBoxRowEntry[] {
  const gear: Record<number, ItemA> = this.GemableGear;
  // rest of code

  return Object.keys(gear)
  .filter((slot) => {
  })
  .map<GemBoxRowEntry>((slot) => {
    const item: ItemA = gear[slotNumber];
    // rest of code

    return {
      item, // ItemA - where ItemB is expected
      // rest of code
    };
  });
}
```

CericME:
Things got a little hectic on Good Friday.

Here's my Plan of Attack:
1. Low hanging fruit to get those resolved.
2. Change the coloring to the scheme discussed with @emallson
3. Investigate and fix the bug around the BonusIds from @Krealle 
4. Refactor of some of the more prolific code.
5. PR Cleanup.

I've converted this to a draft as I still want the review and comments but until 3 in particular is resolved this can't be merged.

src/parser/shared/modules/items/GemChecker.tsx
CericME: Thanks

src/parser/core/itemHelper.ts
CericME: I'm not used to being able to do that.  That's pretty cool. Thanks.

src/parser/shared/modules/items/GemChecker.tsx
CericME: Disabled isn't a bad option.  I'll have to see how that is handled.
In fact I like disabled quite a bit for the single socket add ones.

Do the same on the 2 socket add ones when there is no socket.

src/parser/shared/modules/items/GemChecker.tsx
CericME: Was looking at the Raidbot BonusId dump: https://www.[raidbots.com/static/data/live/bonuses.json](https://www.raidbots.com/static/data/live/bonuses.json)

Looks like the best way would be to look up the bonus Id and see if it has a socket on it.  If no socket is found from the available bonus ids assume that if there are gems the maximum ones are added unless the gear has an addable socket.

On a side note:  There are a lot of Bonus Id's that are just socket: <number>  Feels super wasteful be curious what the actual underlying reason for that is.

src/parser/shared/modules/items/GemChecker.tsx
CericME: This became a five hour adventure which involved the Blizzard API and some build rules on an unrelated file but, I have this updated now.

src/parser/shared/modules/items/GemChecker.tsx
CericME: @emallson
Was thinking about this yesterday and today.  What I've done now is introduce a level of potential for the equipment.
After looking through some UI, Psych, Art concepts and talking with AI I picked the color #f472b6 (A pink.)

The new color needed to be distinct from our existing performance colors (Yellow, Green, Blue, etc.) and crucially, avoid conflicts with the World of Warcraft item quality colors (Purple=Epic, Blue=Rare, etc.), which users commonly recognize. 

Purple seems to be the goto for this type of meaning but has that Epic connotation with the users.  Teal is another but was to close to green in this case.  Pink is consider to mean vibrancy and more importantly isn't used in the current or WoW color schemes so should be just different.

image

src/parser/shared/modules/items/GemChecker.tsx
CericME: I think I have the specific problem you were seeing fixed with tweaks for other items earlier today.
image

Here's a later parse I saw of yours that indicate the head has a potential to have a socket.
image

src/parser/shared/modules/items/GemChecker.tsx
CericME: I consolidated down the Performance and the Tooltip functions.

CericME:
I achieved everything that I was trying to get done today.  I now need to find special cases with weird log testing.  Which means finding/making some weird gemming on logs.

Also I don't know what's up with the About link timeout from the Playwright test.  I'm not even sure where that is.

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -29,216 +32,170 @@ class GemChecker extends Analyzer {
...
-            You are missing a possible slot on your {slotName}. <br />6 Algari Token of Merit for
-            S.A.D. to add a slot.
-          </Trans>,
-        ); //ItemLink in the future when I figure out how they work
+    //#region Special Cases
+    if (item.id === ITEMS.CYRCES_CIRCLET.id) {
+      if (missingGems === 0) {
+        equipmentPerformance = EquipmentPerformance.Good;
+      } else {
+        equipmentPerformance = EquipmentPerformance.Good;
```
emallson: should probably still flag cyrce's with missing gems as a problem. 
```suggestion
        equipmentPerformance = EquipmentPerformance.Fail;
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -29,216 +32,170 @@ class GemChecker extends Analyzer {
...
-                maxGem = false;
-              } else {
-                tempQP = QualitativePerformance.Perfect;
-              }
+              tempQP = EquipmentPerformance.Good;
               break;
             case 2:
-              tempQP = QualitativePerformance.Good;
-              maxGem = false;
+              tempQP = EquipmentPerformance.Good;
```
emallson: r2 gems should just be Ok. there's easy (if sometimes expensive) room for improvement)
```suggestion
              tempQP = EquipmentPerformance.Ok;
```

src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -29,216 +32,170 @@ class GemChecker extends Analyzer {
...
             case 1:
-              tempQP = QualitativePerformance.Ok;
-              maxGem = false;
+              tempQP = EquipmentPerformance.Ok;
               break;
-            case undefined: //Consider Item Level when there isn't a lookup.
-              tempQP = QualitativePerformance.Ok;
-              maxGem = false;
+            case undefined: //In Futuer Consider Item Level when there isn't a lookup return.
+              tempQP = EquipmentPerformance.Ok;
```
emallson: for the time being, we should mark gems without quality as Fail. Cyrce's is already handled separately, so we don't need to worry about those missing quality.

this can help catch cases where someone has accidentally bought an old expansion gem and not realized it.
```suggestion
            case undefined: 
              tempQP = EquipmentPerformance.Fail;
```

emallson (commented):
better :+1: 

going to think about whether this `Potential` option makes sense for things besides gems. most things (like enchants) you can just go buy. you don't need a time-gated resource for it.

but maybe this opens up checking that you've used all your sparks? :thinking:

src/parser/shared/modules/items/GemChecker.tsx
CericME: I didn't really change it from where there are 3 levels on that one.  Easy enough.

CericME:
> going to think about whether this `Potential` option makes sense for things besides gems. most things (like enchants) you can just go buy. you don't need a time-gated resource for it.
> 
> but maybe this opens up checking that you've used all your sparks? 🤔

Gem Sockets in particular are an the interesting one for potential.  I need to read up on how you obtain the new head enchant.  I'm fairly sure in classic there were repped based ones as well you couldn't buy.

CericME:
Another thing that comes to mind for potential is upgrading gear and catalyzing.
I've been running into the Crest cap all season.

I would lean toward something like "These are the pieces of gear that look to be upgradeable.  Please get a SimC and go to Questionable Epic Live for simulation or consult your favorite to see which be best piece to start to focus on. "

I saw in the translation file a reference to Questionably Epic.  I'm assuming there is one for I think Raidbot is the DPS one.

CericME:
Current look with the last push.
image

Plus it has the Cyrce's bug I need to investigate.
Want to find what the icon is for an empty socket because I think I can make that work now but I don't know what color or type of socket from the information I can find so it would need to be generic.

CericME:
image

Having a hard time making/finding logs for the edge cases.  I had to do some under the hood data manipulation for this to show.  I made an empty gem icon as that doesn't seem to actually exist as thing you can get.

I need to prove that my showing logic is right for weapons and the other gear slots.
I'm also going to need redo some logic for showing empty gems.

emallson:
it is okay for this to not cover every possible edge case. when they happen, it will be a lot easier to fix them

emallson:
i'm gonna sit down this evening and make a few tweaks to this to get it merged. the changes i'm planning to make are mostly to do with organization and presentation

### Organization

I'm going to remove `EquipmentPerformance` for now in favor of a local `type GemPerformance = QualitativePerformance | 'vault-upgrade'`. Just to avoid over-abstracting at this stage. The conversation on this PR has given me a lot of ideas for what this section could look like for retail (and classic, for that matter), and I don't feel that the `EquipmentPerformance` enum is going to cover that.

### Presentation

I'm going to remove the individual gem performance colors from display due to UX. The information can still be displayed as part of the tooltip. Highlighting which *items* need higher quality gems is sufficient, IMO, but also this approach solves some problems with the current implementation. For example:

image

The two rings in this image have identical tooltips. It isn't clear why the 2nd ring has this yellow box in it unless you know the details of the gem checker.

Additionally, I'm going to adjust the color currently used for "potential" to be more muted (potentially just grey) so that issues stand out in comparison. For example:

image

both time-gated socket addition and simple gold-gated socket addition are shown as the same color, but one is fixable *now* while the other is likely not.

emallson:
okay, did some work on it. i'd be good merging it now (assuming CI is green). here are some test cases from my very lazy alts:

http://localhost:3000/report/2FJ7rKkVbDz3hdfg/39-3+Vexie+and+the+Geargrinders/Auhken/standard
image

http://localhost:3000/report/HPhcYD4FVQ8CwXpL/1-4+Vexie+and+the+Geargrinders/Auhken/standard
image

http://localhost:3000/report/jRmX1zAhv9dLFfp2/26-4+Rik+Reverb/Auhken/standard
image

src/interface/guide/components/Preparation/GemSubSection/GemBoxRow.tsx
```diff
@@ -63,16 +45,11 @@ const GemBoxRow = ({ values }: GemBoxRowProps) => {
...
-                  >
-                    <Icon icon={gem.gem.icon} />
-                  </div>
-                ),
-              )}
+              {value.value.gems.map((gem: { gem: EventGem }, gemIndex: number) => (
+                <div key={gemIndex} className={styles['gem-icon']}>
+                  <Icon icon={gem.gem.icon} />
+                </div>
+              ))}
```
CericME, line 52: Alright so the Gems themselves no longer have a quality of their own.  Just doing the gear as a whole.

src/parser/retail/modules/items/GemChecker.tsx
```diff
@@ -1,24 +1,111 @@
...
-    return GEMABLE_SLOTS;
+  private static retailGemSlots = {
+    [GEAR_SLOTS.NECK]: retailJewelrySlots,
+    [GEAR_SLOTS.FINGER1]: retailJewelrySlots,
+    [GEAR_SLOTS.FINGER2]: retailJewelrySlots,
+    [GEAR_SLOTS.HEAD]: retailBodySlots,
+    [GEAR_SLOTS.WRISTS]: retailBodySlots,
+    [GEAR_SLOTS.WAIST]: retailBodySlots,
+  };
+
```
CericME, line 32: Can you no longer have gems on other slots like chest and weapon?

CericME:
@emallson 

Log: https://www.warcraftlogs.com/reports/3Zb2BHt9cWNCfawP/?fight=18&source=19

image

The neck has 1 gem already in it but doesn't have a second socket so it can have one added.  The bit that adds the empty sockets isn't taking into account the scenario that it can have 2 socket added but it already has 1 socket and it's filled. 

A bit to tired to dig into this one right this sec as one of you examples show having 1 empty slot and missing a socket.
http://localhost:3000/report/2FJ7rKkVbDz3hdfg/39-3+Vexie+and+the+Geargrinders/Auhken/standard
image

src/interface/guide/components/Preparation/GemSubSection/GemBoxRow.tsx
CericME: I think I found it.  You moved that portion to the Tooltip.

CericME:
I forgot to mention that the refactoring as a whole is more streamlined and shows the touch of a veteran in this codebase and tech stack.

src/parser/retail/modules/items/GemChecker.tsx
emallson: correct. there are tinkers for some slots but are engineering specific and require specific items.

emallson:
> The neck has 1 gem already in it but doesn't have a second socket so it can have one added. The bit that adds the empty sockets isn't taking into account the scenario that it can have 2 socket added but it already has 1 socket and it's filled.

good catch, i'll take a look and fix that

EDIT: oh as i posted this github refreshed and i see that you already have!

src/interface/guide/components/Preparation/GemSubSection/GemBoxRow.tsx
CericME: I must have been pretty tired last night as I did read your full explanation of your thinking and you specifically talk about this.

emallson (approved):

PR #7330, opened 2025-04-20
[Mage-All Specs] Major Defensives Guide section 

earosselot:
### Description

Added Major Defensives section to all mage spec Guides. The deffensives covered on this section are Mirror Images, Greater Invisibility and Ice Block / Ice  Cold. 
As a starter for this section. I left out Alter Time since I think it might be more difficult to analyze and barriers due to their short cd.

#### With Ice Cold:
- Test report URL: `/report/cbpBCZ2ngDKXdvxJ/16-Heroic+Chrome+King+Gallywix+-+Kill+(7:03)/29-Raistlín/standard`
- Screenshot(s):
image

#### With Ice Block:
- Test report URL: `/report/Z4ktN6rV9GpvcJhf/17/220`
- Screenshot(s):
image


src/analysis/retail/mage/shared/defensives/IceBlock.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  onDamageTaken(event: DamageEvent) {
+    if (!this.defensiveActive) {
+      return;
+    }
+    this.recordMitigation({
+      event,
+      mitigatedAmount: absoluteMitigation(event, this.mitigation),
+    });
+  }
```
emallson: this will not work correctly with immunities since there is no damage value for immuned damage. with ice cold being as strong as it is, people typically only take ice block to cheese things so it is probably not worth trying to evaluate what was immuned (or doing very basic checks like seeing if you immuned anything)

emallson (changes requested):
this mostly looks good. i think just remove the `recordMitigation` for ice block right now to avoid the chart doing weird things. it can stay in the list for evaluating the cooldown, maybe?

src/analysis/retail/mage/shared/defensives/IceBlock.tsx
emallson: for ref, this is the implementation of `absoluteMitigation`: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/f9ca50914c2d122548a14b28eaf9c69478d95c2a/src/interface/guide/components/MajorDefensives/MajorDefensiveAnalyzer.tsx#L108-L112

if you pass 1.0 in as the `mitPct`, it will calculate:

```ts
const actualAmount = undefined; // probably undefined for an immuned damage event
const priorAmount = actualAmount * (1/ (1 - 1)); // = NaN? * 1 / 0 = NaN? * Infinity
return Infinity - NaN?;
```

earosselot: Thanks for taking a look and for the detailed explanation :)
I know that almost nobody uses IceBlock, I just included it for the sake of completion. I've corrected the issue, now it only shows the cd, but does not process the incoming damage.

emallson (approved):

PR #7332, opened 2025-04-22
Streamline how Gem Recommendations work

CericME:
### Description

Realized that with Gems we really don't care what slot the gem itself is in for retail.  (More complicated for versions of Wow that have socket bonuses and other things.) Simplified the Gem Recommendations to only need the IDs of the Gems that are recommended.

I also checked to see if the Recommended Enchants were working as come to find out no one is using them and I based the original GemChecker on the EnchantChecker.

### Testing

Commit 8ce467 has all the setup needed to test this with the below log.  I did not want to have the Mistweaver recommended Gems and Enchants to be in this particular pull request.

- Test report URL: http://localhost:3000/report/tqYCnVyDm9BN3bwz/16-Heroic+Mug'Zee,+Heads+of+Security+-+Wipe+16+(6:21)/2-Locsul/standard
- Screenshot(s):
 image


src/parser/shared/modules/items/GemChecker.tsx
```diff
@@ -255,8 +252,9 @@ class GemChecker extends Analyzer {
     );
   }
 
-  getGemBoxRowEntries(recommendedGems: Record<number, CraftedItem[]> = {}): GemBoxRowEntry[] {
+  getGemBoxRowEntries(recommendedGems: number[] = []): GemBoxRowEntry[] {
     const gear = this.GemableGear;
+    this.recommendedGems = recommendedGems;
```
emallson: `get`ers should generally not mutate state except for caching.

instead of setting `recommendedGems` here, specs can sub-class `GemChecker` and override `recommendedGems` if that property is changed from `private` to `protected`.

emallson (changes requested):
minor change, and then good to go :+1:

src/parser/shared/modules/items/GemChecker.tsx
CericME: I ended going back to how it was.  The minor saving of a class scope variable just couldn't make it work in the 2 hours I alotted with the index.tsx that uses.

emallson (approved):

PR #7338, opened 2025-04-25
[Shadow] Fix issues in Shadowy Insight and Deathspeaker Talents

DoxAshe:
### Description

This fixes an error with Shadowy Insight when using the Void Weaver talent tree.  Shadowy Insight procs were not considered good if they were used on Void Blast instead of Mind Blast.  This fixes that mistake and improves on checking to see if the buff was used properly.

Also, the Deathspeaker talent was not counting the overwritten buffs as missed.  If the buff is refreshed it now counts as a missed proc.


DoxAshe:
It seems like I'm having issues with the linter and prettier, and I'm not sure why.  I updated to NodeJS v22, but don't know what else to do.

ToppleTheNun:
ESLint is file path normalization issues.

Prettier is likely the comments in ShadowyInsight not starting with a space.

ToppleTheNun:
ESLint is fixed in the main repo, just need to merge it into your branch and fix the baseline.json

DoxAshe:
Thank you, I think I've got it all corrected now.  Sorry about the difficulties.

emallson (approved):

PR #7340, opened 2025-04-26
browser support for recording some error metrics

emallson:
this records:

- GCD errors per minute
- Cooldown errors per minute
- Active Time %
- Unknown abilities per minute

the goal is to start using this to see how good the fundamentals of spec analysis are across specs, especially for foundation specs that don't have a dedicated person looking at them all the time

the backend changes for this are here: https://github.com/WoWAnalyzer/server/pull/66


emallson:
the e2e errors are due to an unexpected `console.error`. i'll change how that logs to prevent e2e failures.

(the lint fail is just me 🙃)

PR #7341, opened 2025-04-27
Recommended Gems now shown to User.

CericME:
### Description

Was documenting how to add Recommended Gems and thought back to a point that emallson made in the original PR for the feature about the individual gem qualities showing up on the UI.  If you know how the underlying code works that makes sense but if not it really doesn't.  As we are trying to help the users perform better or know more I added in the Recommended Gems showing if there are some with item links.

### Testing
[Setup some recommend gems](https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Setting-up-Recommended-Gems)
Go to any report for that class and spec.  Now you will see the recommended gems listed out under the gear in the gem area.

Screenshot(s):
Before:
image

After:
Recommended Setup:
image

Recommended not Setup:
image

(Sidenote for anyone curious I'm trying to setup signed commits.)

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -43,7 +43,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
-  change(date(2025, 4, 27), 'Recommend Gems now show to the User if setup', Ceric),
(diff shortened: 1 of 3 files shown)


emallson (approved):
seems fine. i suspect the display of this info will get refined when someone starts using that feature

PR #7352, opened 2025-05-03
Tooltip error on get gear

CericME:
### Description
image
I'm fairly sure more then just Rogues can use the speed stat.

The error comes from the tertiary stats using Spell Icons that link to the spells.  As the primary stats do not have tooltips the solution was to simply not have tooltips on the tertiary stats.

### Testing
Go to any report and hover over the tertiary stats.


Pewtro:
Is it really worth turning off links because the tooltip is wrong? With that logic we should link to almost no special effect gear because their tooltips are almost always off with the values.

emallson:
i am kinda ambivalent on this. the link being for a "rogue spell" is kinda off and i don't really think that speed needs much explanation, but the description *is* right

CericME:
Look at the other ones.  Is there any reason to keep the spell links in that particular area?
Even if the tertiary stats give a more or less correct description.
image
image

The Primary Stats give no Tooltip what-so-ever.

Not consistent.

danny-janse:
Sorry to butt in. I'm taking a look at this using report http://wowanalyzer.com/report/cgbvxAzZFXPDfBTR/3-Heroic+One-Armed+Bandit+-+Kill+(5:04)/Revamp/standard/character

Exactly what is the error we're trying to resolve, I dont experience any issues?

PR #7354, opened 2025-05-06
[Warrior] [Arms] [Fury] Remove Storm Bolt suggestion and fix Brutal Finish buff application event

GeoffreyBalshaw:
### Description

- Remove Storm Bolt suggestion; this is _technically_ part of the APL, but it's almost 0 damage and people have been confused frequently why it was being suggested to use a stun on raid bosses
- Use a normalizer to fix the timestamp of the Brutal Finish buff being applied relative to the end of the Bladestorm which applied it. This fixes spells being flagged as inefficient which were queued between the Bladestorm end event and the Brutal Finish apply event.

### Testing

- Test report URL: `/report/FwpATytC4fjR1YMn/22-Heroic+Sprocketmonger+Lockenstock+-+Kill+(4:25)/Nesquik/standard/timeline`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/warrior/fury/modules/normalizers/BrutalFinishBuffNormalizer.tsx
+++ b/src/analysis/retail/warrior/fury/modules/normalizers/BrutalFinishBuffNormalizer.tsx
@@ -1,0 +1,59 @@
+import SPELLS from 'common/SPELLS';
+import EventLinkNormalizer, { EventLink } from 'parser/core/EventLinkNormalizer';
+import { AnyEvent, EventType, GetRelatedEvent, HasRelatedEvent } from 'parser/core/Events';
+import { Options } from 'parser/core/Module';
+
+/**
+ * Brutal Finish is a buff that is applied when Bladestorm is finished channeling
+ * But that buff doesn't actually get applied for 100-200ms after the bladestorm finishes
+ * so it's possible for the player to queue up a spell to cast in that 100-200ms window
+ * and any checks on that cast event related to having the Brutal Finish buff or not are incorrect
+ *
+ * This normalizer links the Brutal Finish buff apply events to the Bladestorm end event which
+ * applied it, and replaces the buff apply event so that it's effectively applied instantly
+ *
+ * Sample report: /report/FwpATytC4fjR1YMn/22-Heroic+Sprocketmonger+Lockenstock+-+Kill+(4:25)/Nesquik/standard/timeline - the Raging Blow at 2:31 was previously flagged as incorrect because the cast event happens between the Bladestorm
+ * end event and the Brutal Finish apply event
+ */
+
+const BRUTAL_FINISH_ON_BLADESTORM_END = 'Brutal-Finish-On-Bladestorm-End';
+const BLADESTORM_BUFF_END_BUFFER_MS = 200; // Brutal Finish buff is applied 100-200ms after Bladestorm ends
+
+const EVENT_LINKS: EventLink[] = [
+  {
+    linkRelation: BRUTAL_FINISH_ON_BLADESTORM_END,
+    linkingEventId: SPELLS.BRUTAL_FINISH_BUFF.id,
+    linkingEventType: EventType.ApplyBuff,
+    referencedEventId: SPELLS.BLADESTORM.id,
+    referencedEventType: EventType.RemoveBuff,
+    forwardBufferMs: BLADESTORM_BUFF_END_BUFFER_MS,
+    backwardBufferMs: BLADESTORM_BUFF_END_BUFFER_MS,
+    anyTarget: true,
+  },
+];
+
+class BrutalFinishBuffNormalizer extends EventLinkNormalizer {
+  constructor(options: Options) {
+    super(options, EVENT_LINKS);
+  }
+
+  normalize(rawEvents: AnyEvent[]): AnyEvent[] {
+    const events = super.normalize(rawEvents);
+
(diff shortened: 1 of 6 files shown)


Krealle:
You should consider using `EventOrderNormalizer`  for `Brutal Finish`, since it's made to handle your specific use case.

src/analysis/retail/warrior/fury/modules/normalizers/BrutalFinishBuffNormalizer.tsx
```diff
@@ -0,0 +1,59 @@
...
+    events.forEach((event, index) => {
+      if (
+        event.type === EventType.ApplyBuff &&
+        HasRelatedEvent(event, BRUTAL_FINISH_ON_BLADESTORM_END)
+      ) {
+        const bladestormEndEvent = GetRelatedEvent(event, BRUTAL_FINISH_ON_BLADESTORM_END);
+        events.splice(index, 1); // remove original Brutal Finish buff apply event
+        event.timestamp = bladestormEndEvent?.timestamp || event.timestamp; // replace BF buff apply timestamp with BS end
+        event.__modified = true;
+        events.push(event); // add modified event back in
```
emallson, line 52: you don't need to do the `splice`/`push` stuff. modifying the event modifies it in the list. one of the blessings and curses of mutable objects.

before approving: can you confirm that the effects Brutal Finish apply to spells queued like this?

```suggestion
        const bladestormEndEvent = GetRelatedEvent(event, BRUTAL_FINISH_ON_BLADESTORM_END);
        event.timestamp = bladestormEndEvent?.timestamp || event.timestamp; // replace BF buff apply timestamp with BS end
        event.__modified = true;
```

emallson:
also, re @Krealle's comment, sref said this which i kinda agree with:

> I am now of the opinion that reordering events is evil and should only be done as an absolute last resort

`EventOrderNormalizer` ended up removed from [the core tools wiki page](https://github.com/WoWAnalyzer/WoWAnalyzer/wiki/Core-Tools-Everyone-Should-Know), but it *is* intended for this use case.

if it is easy/practical, adding helper to check buff range + `GetRelatedEvent` might do what you want. however, if it needs to work in the APL i think we're probably stuck with modifying the timestamp unless you *only* want to check the related events because of APL / functional analyzer limitations.

GeoffreyBalshaw:
@Krealle @emallson 

With respect to using `EventOrderNormalizer` - that seems to set the timestamp of both events to be the timestamp of the later event, which in this case doesn't work - we need the timestamps to both be the earlier of the two. In the PR description's example, the bladestorm removebuff event ends up getting placed after the brutal finish applybuff event
image
and now both are after the cast event which comes 68ms before the brutal finish buff is applied
image

src/analysis/retail/warrior/fury/modules/normalizers/BrutalFinishBuffNormalizer.tsx
GeoffreyBalshaw: It seems like it does, yes. Brutal Finish is applied when Bladestorm ends, and is consumed when Rampage is cast. In this log https://www.warcraftlogs.com/reports/K9M6kac7VWNnyCHx?fight=10&type=summary&source=247&view=events&start=2190217&end=2193218 the queued cast event for Rampage is 1ms before the Brutal Finish buff is applied, and then it's removed just over a second later without any other Rampage casts (the cast in that time was Execute, which doesn't interact with Brutal Finish in any way).

This change in general is because it is technically optimal to overcap on rage to stack additional buffs on the Rampage that will consume the Brutal Finish - [this check is already in place](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/analysis/retail/warrior/fury/modules/spells/MissedRampage.tsx#L93-L95) but that Raging Blow can get incorrectly flagged as inefficient if its cast event happens in the delay between the Bladestorm removebuff event and the Brutal Finish applybuff event

emallson:
`EventOrderNormalizer` always moves events forward, yes. if this is being used in APL rules, I think it is fine to leave as is. if this is not being used in APL rules, then maybe take a look at using `GetRelatedEvent` directly instead of relying on event order / `hasBuff` checks.

GeoffreyBalshaw:
> `EventOrderNormalizer` always moves events forward, yes. if this is being used in APL rules, I think it is fine to leave as is. if this is not being used in APL rules, then maybe take a look at using `GetRelatedEvent` directly instead of relying on event order / `hasBuff` checks.

By "APL rules" do you mean an AplCheck like [this](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/the-war-within/src/analysis/retail/warrior/arms/modules/core/AplCheck.tsx) ? If so, currently no but only kind of - only Arms is set up to use an AplCheck, whereas Fury is sort of a collection of different discrete checks. The [Arms APL](https://github.com/simulationcraft/simc/blob/thewarwithin/ActionPriorityLists/warrior_arms.simc) doesn't currently care about Brutal Finish, so this particular issue isn't immediately relevant for that spec and its AplCheck, but it might become relevant in the future (either when Arms does start caring about Brutal Finish, or when Fury is changed to use an AplCheck)

emallson (approved):
okay, i'm going to approve this so that it doesn't cause a nasty footgun if/when Fury starts using `AplCheck`, since `EventOrderNormalizer` is soft deprecated and doesn't do quite the right thing anyway.

PR #7360, opened 2025-05-16
Fix Critical Hits for Healing and Damage

lolswirl:
### Description

A change made in October removed critical hits being calculated under the Ability Tracker for both healing and damage, so I've readded it.

### Testing

- Test report URL: `/report/kfnwY1ydMJ6pBbTm/12-Mythic+Chrome+King+Gallywix+-+Kill+(9:32)/Qtkushi/standard/statistics`
- [Log crit percentages](https://www.warcraftlogs.com/reports/kfnwY1ydMJ6pBbTm/?fight=12&source=21&type=healing&ability=25914&by=source)
- Screenshot(s):

Live:
image

Fix:
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -37,12 +37,14 @@ import {
   KYZ,
   oneunreadmail,
   Ceric,
+  swirl,
 } from 'CONTRIBUTORS';
 import { ItemLink } from 'interface';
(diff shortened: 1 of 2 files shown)


emallson (approved):
this looks like it was accidentally removed as part of the `DamageValue` refactoring. i double checked and `DamageValue` does not track number of hits, so this seems like the correct fix.

PR #7369, opened 2025-05-30
[Core] Add WCL user sign-in & private logs support

Krealle:
### Description

Related to https://github.com/WoWAnalyzer/server/pull/67

- Added a new `Warcraft Logs` user sign-in.
- Updated `Report Not Found` page to direct users to sign in with WCL.
- Updated `rawFetchWcl` to send necessary credentials for fetching private data.
- Updated error handling.

### Testing

To test follow the steps found in the server side PR.

- Screenshot(s):

image

image

image

image


src/interface/PremiumLoginPanel.tsx
```diff
@@ -136,29 +166,37 @@ const PremiumLoginPanel = () => {
             </h1>
             <div className="description">
               <Trans id="interface.premiumLoginPanel.panel.sign-in.description">
-                Sign in with your Patreon or GitHub account using the buttons below.
+                Sign in with your Patreon, GitHub or Warcraft Logs account using the buttons below.
               </Trans>
             </div>
           </div>
         </div>
         <div className="row" style={{ position: 'relative' }}>
```
emallson, line 174: nit: don't think we need this `row` class anymore

src/interface/PremiumLoginPanel.tsx
```diff
@@ -136,29 +166,37 @@ const PremiumLoginPanel = () => {
...
           <div className="loginBubble">Already unlocked Premium? Login here!</div>
-          <div className="col-lg-6" style={{ padding: 0 }}>
+          <div className="col-lg-12" style={{ padding: 0 }}>
```
emallson: and without the row class, we don't need these column classes. that probably removes the need for the padding override too

src/interface/report/handleApiError.tsx
```diff
@@ -27,15 +31,21 @@ export default function handleApiError(error: Error, onBack: () => void) {
...
           </Trans>
         </div>
         <div>
           <button type="button" className="btn btn-primary" onClick={onBack}>
             &lt; <Trans id="interface.report.handleApiError.back">Back</Trans>
           </button>
+          <a
+            className="btn btn-primary"
+            style={{ marginLeft: 20 }}
+            href={`${import.meta.env.VITE_SERVER_BASE}login/wcl`}
```
emallson: it would be nice if this also sent along the link back to this page so that after login you were redirected to the report. however, not necessary and if that ends up being a headache with the oauth plugin it can be skipped (since the plugin is already using the session, you can probably set the redirect url on the session cookie there and then use it on successful login with appropriate state value)

src/interface/PremiumLoginPanel.tsx
```diff
@@ -70,6 +76,30 @@ const LoggedIn = React.forwardRef<HTMLDivElement, unknown>((props, ref) => {
...
+                  <a
+                    className="btn btn-primary"
+                    style={{ display: 'block' }}
+                    href={`${import.meta.env.VITE_SERVER_BASE}login/wcl`}
+                  >
+                    <Trans id="interface.premiumLoginPanel.logged-in.continue">Continue</Trans>
+                  </a>
+                </div>
+              </div>
+            ))}
```
emallson, line 102: if you log in with WCL *first*, you have to log out and then in with a different method to get premium. keeping some small "Link Patreon" / "Link GitHub" buttons around in the "haven't unlocked premium yet" case would be nice to keep the flow from being *too* confusing

could potentially ignore the comments on the buttons below and change the layout of the buttons based on whether you're signed in with WCL only or if you have premium as well, like:

- not signed in: all 3 buttons
- signed in with no wcl, premium: wcl button only
- signed in with wcl, no premium: premium buttons only
- signed in with wcl, premium: no buttons

but ultimately this problem is present because we're tacking this onto the premium page :see_no_evil: 

don't spend too much time thinking about this one. the version of this page as it exists in this PR is good enough, the feature is worth a minor UX regression here. we can always come back with further revisions

emallson (commented):
some minor comments. i think this is mergeable pending the `cache-control` change on the backend, so consider this a :+1: (but don't merge before the server is in)

src/interface/PremiumLoginPanel.tsx
Krealle: Removing `row` class makes the styling a bit awkward, so think it makes sense to keep it in. At least for now.

image

src/interface/PremiumLoginPanel.tsx
Krealle: These classes/overrides do look like they can be removed without causing issues. Think they where needed so the buttons could be side-by-side on larger screens. But now they are always stacked.

emallson (approved):

PR #7376, opened 2025-06-07
[Unholy] Update Unholy Death Knight Abilities and Talents for Patch 11.1.5

brandrewsss:
### Summary

This PR is the first step toward fully updating the **_Unholy Death Knight_** specialization for patch 11.1.5.

### What’s Included
- ✅ Updated `src/common/SPELLS/deathknight.ts` with all Unholy-relevant abilities
- ✅ Updated `src/common/TALENTS/deathknight.ts` with all Unholy-relevant talents
- ✅ Updated `src/analysis/retail/deathknight/unholy/modules/Abilities.tsx` with full ability and talent tree data based on patch 11.1.5
- ✅ Corrected `runicPowerCost` values that were previously negative
- ✅ Added ranges to all abilities and talents in `Abilities.tsx`

### What’s Not Included (coming in future PRs)
- ⬜ Update Buffs
- ⬜ Update Cooldown Logic
- ⬜ Rotation tracking and Spell Priorities
- ⬜ Suggestions for Optimizations and Improvements

### Related Issue

This PR contributes to: [#7375]


src/common/TALENTS/deathknight.ts
```diff
@@ -74,7 +74,7 @@ const talents = {
     entryIds: [96333],
     definitionIds: [{ id: 101335, specId: 252 }],
     runesCost: 1,
-    runicPowerCost: -10,
+    runicPowerCost: 10,
```
Pewtro: You shouldn't change anything in the talents file, these are generated files.
These (and all the others you changed from negative to positive) are correctly marked as negative values because they generate runic power - that's how generation is displayed for Death Knights. 
image

src/parser/core/modules/Abilities.ts
```diff
@@ -12,6 +12,7 @@ import Ability, { SpellbookAbility } from './Ability';
 export const AbilityRange = {
   Melee: 5,
   FullRanged: 40,
+  Self: 0,
```
Pewtro: The range values are used in the APL checker only, I don't think there's a need for a self value because it would just always be true in the spellRange function, so there is no point in checking for if something was in range.

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -73,10 +79,30 @@ class Abilities extends CoreAbilities {
...
+        cooldown: 20,
+        charges: 1,
+        gcd: {
+          base: 1500,
+        },
+        range: 30,
+      },
+      {
+        spell: TALENTS.SOUL_REAPER_TALENT.id,
+        enabled: !combatant.hasTalent(TALENTS.SOUL_REAPER_TALENT),
```
Pewtro: This should probably be `combatant.hasTalent(TALENTS.SOUL_REAPER_TALENT),` without the negator

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -177,7 +219,7 @@ class Abilities extends CoreAbilities {
         gcd: {
           base: 1500,
         },
-        isDefensive: true,
```
Pewtro, line 180: Why should this no longer count as a defensive?

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -210,29 +243,38 @@ class Abilities extends CoreAbilities {
...
 
-      // utility
+      // region Utility
       {
         spell: SPELLS.DEATH_CHARGE.id,
         enabled: combatant.hasTalent(TALENTS.DEATH_CHARGE_TALENT),
         category: SPELL_CATEGORY.UTILITY,
         charges: 1,
         cooldown: 45,
-        gcd: null,
```
Pewtro: Is this on the gcd now? If not these null values should stay. Same goes to all the other places it was changed

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -250,15 +293,14 @@ class Abilities extends CoreAbilities {
...
           base: 1500,
         },
+        range: 20,
       },
       {
         spell: TALENTS.RAISE_DEAD_UNHOLY_TALENT.id,
         enabled: combatant.hasTalent(TALENTS.RAISE_DEAD_UNHOLY_TALENT),
         category: SPELL_CATEGORY.UTILITY,
-        cooldown: 30,
-        gcd: {
```
Pewtro, line 259: Did the GCD change for this?

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -73,10 +79,30 @@ class Abilities extends CoreAbilities {
...
+          base: 1500,
+        },
+        range: 30,
+      },
+      {
+        spell: TALENTS.SOUL_REAPER_TALENT.id,
+        enabled: !combatant.hasTalent(TALENTS.SOUL_REAPER_TALENT),
+        category: SPELL_CATEGORY.ROTATIONAL,
+        cooldown: 6,
+        gcd: null,
```
Pewtro: I'm pretty sure Soul Reaper has a GCD no?

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -166,9 +210,7 @@ class Abilities extends CoreAbilities {
           60 -
           Number(combatant.hasTalent(TALENTS.ANTI_MAGIC_BARRIER_TALENT)) * 20 +
           Number(combatant.hasTalent(TALENTS.UNYIELDING_WILL_TALENT)) * 20,
-        gcd: {
```
Pewtro, line 169: Assuming this is still off the gcd this should stay or be marked as gcd: null

Pewtro (changes requested):
Firstly, thanks for the PR :) 

There are some things that need to be fixed before this can be merged, at least the gcd and runicpowercost changes should be reverted - if you parsed a log with these changes you should begin seeing a lot of gcd errors with all the nulls removed while also messing up any resource cost/generation logic in statistics.

src/CONTRIBUTORS.ts
```diff
@@ -2724,7 +2724,25 @@ export const Leftyxiv: Contributor = {
     {
       name: 'Magicmushies',
       spec: SPECS.DEMONOLOGY_WARLOCK,
-      link: 'https://www.warcraftlogs.com/character/us/area-52/magicmushies?zone=43',
+      link: 'https://worldofwarcraft.blizzard.com/en-us/character/us/area-52/magicmushies',
```
Pewtro: Why did you change someone elses character link?

src/common/TALENTS/deathknight.ts
brandrewsss: Reviewing this and trying I understand the difference now. So Death Strike Cost RP and that's why it's positive and AOTD generates that's why it's negative. This is helpful to understand even if the talents file is generated. I was under the impression after seeing no changes on the front end of WoWAnalyzer that different contributors just had different ways of displaying so I was trying to normalize it.

src/parser/core/modules/Abilities.ts
brandrewsss: I was just trying to make sure every ability had a range in it for standardization, if it's not necessary then I can remove it off and only leave ranges on values that are positive.

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: I think this was just an autocomplete error! Will correct.

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: I don't understand why this is being used. It is the only instance of `isDefensive` being assinged a value and I didn't see a difference anywhere when testing. I assumed that the category was determined by category: `SPELL_CATEGORY.DEFENSIVE`.

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: This again was an attempt at trying to normalize. Icebound Fortitude is `null`, Summon Gargoyle doesn't have a gcd at all, and Anti-Magic shell is a `static: 0`. However if i'm looking at the metadata on Wowhead these all just display `GCD 0`. Where is the best place to determine when something is a `static: 0` vs a `null`. Also is the Gargoyle not having a gcd value at all is a mistake?

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: `
{
        spell: TALENTS.ANTI_MAGIC_ZONE_TALENT.id,
        category: SPELL_CATEGORY.DEFENSIVE,
        cooldown: combatant.hasTalent(TALENTS.ASSIMILATION_TALENT) ? 90 : 120,
        gcd: {
          base: 1500,
        },
        isDefensive: true,
      }`
      
      
`{
    spell: TALENTS.SACRIFICIAL_PACT_TALENT.id,
    category: SPELL_CATEGORY.DEFENSIVE,
    cooldown: 120,
    gcd: {
      base: 1500,
    },
  }`

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: Yes, the values that were there before were for the shared talent. Not for the Unholy DK talent which is required on the tree.

[https://www.wowhead.com/spell=46585/raise-dead](url)

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
brandrewsss: You are correct, `1500`. Along with the talent issue above I believe what happened is I reused a block and didn't double check the changes.

src/CONTRIBUTORS.ts
brandrewsss: This was purely a mistake. I copied his link to use in my profile and the Copilot addon updated the link for his profile. Will revert this also.

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
emallson: `gcd: { static: 0 }` is functionally equivalent to `null` in virtually all cases. `gcd: null` should be preferred. on a functional level, omitting the `gcd` field is also equivalent. however, stylistically off-gcd abilities should have explicit `gcd: null` assignments.

The `Abilities` types are ports of now-ancient JS so there are quirks like that left around from pre-TypeScript days.

emallson: (putput commented but GitHub isn't showing. Not sure if deleted or bug)

not 100% sure of the behavior when omitted. It goes through a couple layers. I checked the gcd analyzer but didn't read the core abilities so I could be wrong

brandrewsss: I saw that as well, no problem though. As long as `static: 0` and `null` are equivalent I will update all non-gcd abilities and talents to `null`!

emallson (approved):
looks like the comments have been addressed. thanks!

Issue #6969, opened 2024-08-20
Demo Warlock support for TWW


ZyerTCoder:
To fix:
-  [ ] Rework Doom Brand into Doom
-  [ ] Add Gloom/Charhounds
-  [ ] Demonic Healthstones support
-  [ ] Statistic for Wicked Maw/Shadowtouched dmg increase
-  [ ] Statistic for The Houndmaster’s Gambit uptime
-  [ ] Fix Soul Strike tracking for when overcapped

Leftyxiv:
Do we still need to add support for these?

```
Statistic for Wicked Maw/Shadowtouched dmg increase

Statistic for The Houndmaster’s Gambit uptime
```

I think these are dated.

Leftyxiv:
I think we can close this issue! :D

emallson:
👍

PR #7378, opened 2025-06-08
Fix Soul Strike overcapping tracking for Demonology Warlock

Leftyxiv:
## Summary
- Fixed Soul Strike talent to properly display wasted soul shards when overcapped
- Updated Demonology Warlock patch compatibility to 11.1.5

### Soul Strike Module (`SoulStrike.tsx`)
- Enhanced the statistic display to show wasted shards in the main statistics box

### Resource Tracking (`SoulShardTracker.tsx`)
- Added `onEnergize` method to handle resource gain events with proper decimal conversion
- Added `getAdjustedGain` method to convert gain/waste values from WoW's internal format (0-50) to player format (0-5)
- Ensures consistent handling of the `/10` conversion for soul shard values

### Patch Compatibility
- Updated from 11.0.2 to 11.1.5 to match current patch

## Test plan
- [x] Verify Soul Strike statistic displays generated shards correctly
- [x] Verify wasted shards appear when overcapping occurs
- [x] Verify wasted shards don't appear when no overcapping occurs
- [x] Check that values display correctly (e.g., 2 shards wasted)

## Screenshot
The Soul Strike statistic will now show:
```
Soul Strike
[damage amount]
[X] Shards generated
[Y] Shards wasted (only if Y > 0)
```

With tooltip showing additional overcapping details when hovering.

Screenshot 2025-06-08 at 6 21 17 AM
Screenshot 2025-06-08 at 6 21 10 AM

If I'm not mistaken this should also close out issue#6969

### DIFF
--- a/src/analysis/retail/warlock/demonology/CONFIG.tsx
+++ b/src/analysis/retail/warlock/demonology/CONFIG.tsx
@@ -1,4 +1,4 @@
-import { Zyer, Gazh } from 'CONTRIBUTORS';
+import { Zyer, Gazh, Leftyxiv } from 'CONTRIBUTORS';
 import GameBranch from 'game/GameBranch';
 import SPECS from 'game/SPECS';
 import Config, { SupportLevel } from 'parser/Config';
@@ -7,10 +7,10 @@ import CHANGELOG from './CHANGELOG';
 
 const config: Config = {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [Zyer, Gazh],
+  contributors: [Zyer, Gazh, Leftyxiv],
   branch: GameBranch.Retail,
   // The WoW client patch this spec was last updated.
-  patchCompatibility: '11.0.2',
+  patchCompatibility: '11.1.5',
   supportLevel: SupportLevel.MaintainedPartial,
   // Explain the status of this spec's analysis here. Try to mention how complete it is, and perhaps show links to places users can learn more.
   // If this spec's analysis does not show a complete picture please mention this in the `<Warning>` component.
--- a/src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
+++ b/src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
@@ -2,11 +2,12 @@ import { formatThousands } from 'common/format';
 import SPELLS from 'common/SPELLS';
 import TALENTS from 'common/TALENTS/warlock';
 import Analyzer, { Options, SELECTED_PLAYER_PET } from 'parser/core/Analyzer';
-import Events, { DamageEvent } from 'parser/core/Events';
+import Events, { DamageEvent, ResourceChangeEvent } from 'parser/core/Events';
 import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
 import ItemDamageDone from 'parser/ui/ItemDamageDone';
 import Statistic from 'parser/ui/Statistic';
 import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import RESOURCE_TYPES from 'game/RESOURCE_TYPES';
 
 import SoulShardTracker from '../resources/SoulShardTracker';
 
@@ -17,6 +18,8 @@ class SoulStrike extends Analyzer {
 
   soulShardTracker!: SoulShardTracker;
   damage = 0;
+  totalCasts = 0;
+  wastedShards = 0;
 
   constructor(options: Options) {
     super(options);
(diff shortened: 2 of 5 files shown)


src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
```diff
@@ -25,24 +28,55 @@ class SoulStrike extends Analyzer {
...
+  }
+
   statistic() {
-    const shardsGained = this.soulShardTracker.getGeneratedBySpell(SPELLS.SOUL_STRIKE_SHARD_GEN.id);
     return (
       <Statistic
         category={STATISTIC_CATEGORY.TALENTS}
         size="flexible"
-        tooltip={`${formatThousands(this.damage)} damage`}
+        tooltip={
```
Pewtro, line 62: I don't think there's a point in having wasted shards shown in both places, that seems redundant. Similiar story with the damage really, maybe this statistic doesn't need a tooltip at all?

src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
```diff
@@ -25,24 +28,55 @@ class SoulStrike extends Analyzer {
...
+      // Check if player was at shard cap (5 shards = 50 fragments)
+      // The event shows the state AFTER the resource change, so we need to check if waste > 0
+      if (event.waste > 0) {
+        this.wastedShards += 1;
+      }
+    }
+  }
+
+  get shardsGenerated() {
+    return this.totalCasts;
```
Pewtro, line 54: Do you want this to perhaps show "good" shards generated rather than all shards when you're showing the wasted shards separately?

src/analysis/retail/warlock/demonology/CONFIG.tsx
```diff
@@ -7,10 +7,10 @@ import CHANGELOG from './CHANGELOG';
 
 const config: Config = {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [Zyer, Gazh],
+  contributors: [Zyer, Gazh, Leftyxiv],
   branch: GameBranch.Retail,
   // The WoW client patch this spec was last updated.
-  patchCompatibility: '11.0.2',
+  patchCompatibility: '11.1.5',
```
Pewtro, line 13: If something as "basic" resource handling was off - is it really compatible with the current patch with no other changes? It might be, I don't know, just seems odd if something like that was the only thing 'off'

Pewtro:
Your PR mentions "Added tracking for wasted soul shards using getWastedBySpell()", however I don't see this?
SoulShardTracker is imported in a few places in the demo modules - do your changes keep those accurate in implementation and wording still?

src/analysis/retail/warlock/demonology/CONFIG.tsx
Leftyxiv: This is the last update from the tww demo issue, I can remove this until were sure. Am still missing jackpot bonuses I think, just lmk!

This should close out issue #6969

src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
Leftyxiv: good catch, thanks

Leftyxiv:
> Your PR mentions "Added tracking for wasted soul shards using getWastedBySpell()", however I don't see this? SoulShardTracker is imported in a few places in the demo modules - do your changes keep those accurate in implementation and wording still?

No sir, I refactored that out, good catch.

emallson (approved):

src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
Pewtro: This still does the same thing as when I commented - are you sure you want this to show TOTAL casts and not only "GOOD" generation? The statistic reads confusing to me showing you generated X and you wasted Y. That leads me to believe total was X + Y, but it's actually only X.

And if waste should be divided by 10 should totalCasts not ALSO be some sort of function and shards generated is not equal to casts?

src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
Pewtro: I still think this point stands, it's saying the exact same thing but in a tooltip

src/analysis/retail/warlock/demonology/modules/talents/SoulStrike.tsx
```diff
@@ -25,24 +28,56 @@ class SoulStrike extends Analyzer {
...
+    );
   }
 
   handleSoulStrikeDamage(event: DamageEvent) {
     this.damage += event.amount + (event.absorbed || 0);
   }
 
+  handleSoulStrikeShardGen(event: ResourceChangeEvent) {
+    if (event.resourceChangeType === RESOURCE_TYPES.SOUL_SHARDS.id) {
+      this.totalCasts += 1;
```
Pewtro, line 43: Shouldn't this entire function listen to the getAdjustedGain function which returns mathed out gain and waste, which you can then use in here?

PR #7384, opened 2025-06-09
Setup retribution paladin for TWW

danny-janse:
### Description

* Setup retribution paladin to have work done for TWW.
  * Updated the spellbook for Retribution paladin
  * Removed unused spells from the paladin spells list.
  * Replaced retribution spells with the coresponding talent that had the same id for all paladin specs.
* Added Lightforged draenei Racial ability
* Removed an unused image.

I did not go into detail as to whether or not the current statistics are functional enough. This is just a stepping stone for myself or others.

### Testing

A great report with three ret paladin and a prot paladin.
* /report/3DryBTAgdatw1CKj/1-Heroic+Vexie+and+the+Geargrinders+-+Kill+(2:10)


src/analysis/retail/paladin/retribution/modules/talents/ArtOfWar.tsx
```diff
@@ -26,6 +28,12 @@ class AoWProcTracker extends Analyzer {
 
   constructor(options: Options) {
     super(options);
+
+    this.active = this.selectedCombatant.hasTalent(TALENTS_PALADIN.ART_OF_WAR_TALENT);
+    if (!this.active) {
+      return;
+    }
```
emallson, line 35: these early returns are fine, but note that event listeners already don't run if the module is inactive.

emallson (changes requested):
looking through the debug info for this, i see a couple of issues:

image

- it looks like there's a pretty significant haste issue, as GCDs are underestimated a lot. i suspect the main culprit is Crusade haste, which doesn't seem to be present? this can be added in the core `Haste` module without doing an override for the spec.
- Divine Hammer does something funky where it casts multiple times repeatedly. maybe fake casts when it hits a target? i don't think this necessarily needs to be fixed, since from looking through the events it seems...awful

(source for img: http://localhost:3000/report/3DryBTAgdatw1CKj/1-Heroic+Vexie+and+the+Geargrinders+-+Kill+(2:10)/Annatarpal/standard/debug)

i think if crusade haste is added this is probably good to merge as iss

src/analysis/retail/paladin/holy/modules/features/Abilities.tsx
```diff
@@ -296,7 +296,8 @@ class Abilities extends CoreAbilities {
         category: SPELL_CATEGORY.ROTATIONAL,
         charges: combatant.hasTalent(TALENTS.LIGHTS_CONVICTION_TALENT) ? 2 : 1,
         cooldown: (haste) => {
-          const swCdr = hasSanctifiedWrath && combatant.hasBuff(SPELLS.AVENGING_WRATH.id) ? 0.2 : 0;
+          const swCdr =
+            hasSanctifiedWrath && combatant.hasBuff(TALENTS.AVENGING_WRATH_TALENT.id) ? 0.2 : 0;
```
emallson, line 300: something i was reminded by another PR: i don't think this is going to work reliably. some code assumes it can check the cooldown using an average haste value. since your PR isn't really holy-related its fine to just cut it. for similar things in Ret, you should use `SpellUsable` override. it's a bit verbose, unfortunately, but such is life

Texleretour:
@emallson is it okay if I try to complete this PR w/ the Crusade haste ? It looks like the op is inactive and I need his work to start working

emallson:
yup, go for it

emallson:
closing in favor of https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7417

PR #7399, opened 2025-06-30
Fix to innervate when ally casts it on player

karlpralow:
The innervate module does not correctly detect when an external player innervates the druid making it think there's more spells casted during innervate: 
Example:
image
In the above pictures it is joining 3 innervates casted at different times into 1:
image

This behavior happens because it is creating a new tracked innerate on player cast instead of on ApplyBuffEvent.

Here's a screenshot of the result after the changes:
image

I also changed the recommenced saved amount to 400000 instead of 7000.

### DIFF
--- a/src/CONTRIBUTORS.ts
+++ b/src/CONTRIBUTORS.ts
@@ -340,6 +340,13 @@ export const Gao: Contributor = {
 export const Oratio: Contributor = {
   nickname: 'Oratio',
   github: 'karlpralow',
+  mains: [
+    {
+      name: 'Oratio',
+      spec: SPECS.RESTORATION_DRUID,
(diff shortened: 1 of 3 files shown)


kfinch:
Was just sitting down to work on this, seems straightforward. The `applybuff.to` fix should work, but I think that means the "cast on" section in the Guide breakdown needs to be changed to make sense.

As for the 'mana spent' thing I've been unhappy with it for a while because in truth it scales with haste. I've been considering removing the mana check and replacing it with an "active time check" that maybe also checks the player is using it for rejuvs/WGs and not Regrowths that are nearly free anyway

kfinch:
Going to merge and then do my mana update on top

PR #7400, opened 2025-07-02
[Unholy] Unholy Death Knight Core Support for Patch 11.1.5

brandrewsss:
## Summary

This PR brings the Unholy Death Knight analyzer up to date with patch 11.1.5. It includes core analyzers that establish the baseline functionality, an updated Guide experience, updated statistics, minor bugfixes and a modernized internal structure.

### 2. Overview Tab Updates
Replaced the legacy Checklist structure with the modern Guide structure.

#### 2. Cooldown Tracking
- Added efficiency tracking bars for:
  - **Dark Transformation**
  - **Apocalypse**
  - **Unholy Assault**
  - **Abomination Limb**
  - **Death and Decay**
  
  
image

#### 3. Proc Efficiency Tracking
- **Sudden Doom Efficiency**:
  - Tracks proc uses vs. total procs
- **Soul Reaper Efficiency**:
  - Tracks number of casts vs. total possible casts only during the execute phase
  
  
image

#### 4. Resource Use
- **Festering Wound Tracker**:
  - Tracks applied vs. wasted wounds.
  - Flags suboptimal Apocalypse usage.
- **Rune Graph**:
  - Timeline view of Rune usage over time.
- **Runic Power Graph**:
  - Timeline view of Runic Power usage over time.

image

#### Festering Wound Efficiency
- Tracks optimal wound application %

#### Sudden Doom Improvements
- Now displays proc usage % instead of raw number of procs used

### 4. Fixed Rune Cooldown Calculation Bug
- Resolved issue where rune cooldown was being incorrectly calculated due to multiplier
#### Before: 
image

#### After: 
image

### 5. Modernized Folder Structure

Refactored Unholy DK's directory to align with current other up-to-date specs analyzer architecture. This improves will make it easier for me to maintain overtime

Inspired by Evoker, Blood, and Shaman structures.

```
src/
├── analysis/
│   └── retail/
│       └── deathknight/
│           └── unholy/
│               └── modules/
│                   ├── core/       # Core resources
│                   ├── features/   # Shared gameplay analyzers
│                   ├── guide/      # Guide UI components
│                   ├── spells/     # Spell-specific modules
│                   ├── talents/    # Talent-specific modules

```

### Adding full support for San'layn

Due to the haste buff from Essence of the Blood Queen many cooldown issues are being flagged in the debugger. I will be opening a new PR to address this issue as to not mix too much into an already large PR.

#### Linking @emallson's PR #7191 where he addressed this issue for Blood for reference.

image

## Related Issue

This PR concludes issue: [#7375]


brandrewsss:
In commit [d2f1300](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7400/commits/d2f13004fd4fd6f20f94d8cab9f4430476bbd0f0) I removed the Runic Corruption multiplier because it applied the multiplier at spell registration, rather than dynamically during uptime. This caused runes to regenerate twice as fast at all times, regardless of whether Runic Corruption was actually active. As a result, the analyzer overestimated rune availability, producing negative overcap percentages and false spell usability. This change reverts to baseline haste-based regen.

I'm adding this comment because as I work on the PR for San'layn I noticed that this is needed, but in a more accurate way of tracking and it is a big part of fixing the spellUsable issues.

src/analysis/retail/deathknight/unholy/CONFIG.tsx
```diff
@@ -2,44 +2,53 @@ import GameBranch from 'game/GameBranch';
...
 const config: Config = {
   // The people that have contributed to this spec recently. People don't have to sign up to be long-time maintainers to be included in this list. If someone built a large part of the spec or contributed something recently to that spec, they can be added to the contributors list. If someone goes MIA, they may be removed after major changes or during a new expansion.
-  contributors: [],
+  contributors: [Brandrewsss],
   branch: GameBranch.Retail,
   // The WoW client patch this spec was last updated.
-  patchCompatibility: '11.0.2',
-  supportLevel: SupportLevel.Foundation,
+  patchCompatibility: '11.1.5',
+  supportLevel: SupportLevel.MaintainedFull,
```
emallson: don't use `MaintainedFull` until the sanlayn issues are fixed, at a minimum, since the haste issues mean that cast efficiency etc will be off. use `MaintainedPartial` for now to indicate that it is being worked on but the core is not totally done

src/analysis/retail/deathknight/unholy/modules/Guide.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Section title="Resource Use">
+        <SubSection title="Festering Wounds">
+          <span>{props.modules.woundTracker.guideSubsection}</span>
+        </SubSection>
+
+        <SubSection title="Runes">
+          <div>
+            Runes are Unholy’s core resource used to cast key abilities like{' '}
+            <SpellLink spell={TALENTS.FESTERING_STRIKE_TALENT} />. Unlike cooldowns, Runes
+            regenerate passively over time — up to three at once — and managing them efficiently is
```
emallson: > Unlike cooldowns, Runes regenerate passively over time --- up to three at once [...]

- this is confusing, since cooldowns also regenerate passively over time
- the comparison with cooldowns isn't useful here, just simplify this. something like:

> Up to three Runes can regenerate at a time [...] <probably some of the followup about rune waste when 4+ are recharging

src/analysis/retail/deathknight/unholy/modules/Guide.tsx
```diff
@@ -0,0 +1,78 @@
...
+            crucial for maintaining a smooth rotation.
+          </div>
+
+          <div style={{ marginTop: '12px' }}>
```
emallson: use `<p>` instead of the inline `marginTop`. our `<p>` styles have top margins baked in.

src/analysis/retail/deathknight/unholy/modules/Guide.tsx
```diff
@@ -0,0 +1,78 @@
...
+            When you sit on 4 or more Runes, you risk overcapping and wasting regeneration time. On
```
emallson: > When you sit on 4 or more Runes, you risk overcapping and wasting regeneration time.

when you are at 4 or more runes, you don't *risk* overcapping. you *are* overcapping.

src/analysis/retail/deathknight/unholy/modules/guide/IntroSection.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+
+      <div>
+        The analysis here is based on general guidelines and doesn’t always account for specific
+        fight mechanics or edge cases. For the most accurate benchmarking, compare your performance
+        to other top Unholy Death Knights in the same encounter using Warcraft Logs. (
+        <a href="https://www.warcraftlogs.com/zone/rankings/42?boss=3014&class=DeathKnight&spec=Unholy">
+          Mythic One-Armed Bandit Top 100
+        </a>
+        )
```
emallson: don't link directly to a boss here. it'll linkrot easily (e.g. new tier in a month!)

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -36,6 +42,11 @@ class SoulReaper extends ExecuteHelper {
     }
 
     this.addEventListener(Events.fightend, this.adjustMaxCasts);
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS.SOUL_REAPER_TALENT),
+      () => this.totalCastsInExecute++,
+    );
```
emallson, line 48: this is wrong since Soul Reaper can be cast outside of execute, right?

you can combine an `EventLink` between the cast and the explosion with a check of target resources at the explosion (with the caveat that full absorbs don't log resources---so there might be some need for tracking hp, or just assume the player is right if no resources are present and be correct a lot of the time like on the ansurek worshippers) to check if this was cast during the execute.

`ExecuteHelper` may also have a dedicated method you can use to track proper execute casts instead of total casts

src/analysis/retail/deathknight/unholy/modules/features/WoundTracker.tsx
```diff
@@ -0,0 +1,318 @@
...
+      <div style={{ lineHeight: '1.6' }}>
+        <b>
+          <SpellLink spell={SPELLS.FESTERING_WOUND} />
+        </b>{' '}
+        is a key mechanic for Unholy DKs. You apply wounds using{' '}
+        <SpellLink spell={TALENTS.FESTERING_STRIKE_TALENT} /> and burst them with abilities like{' '}
+        <SpellLink spell={TALENTS.SCOURGE_STRIKE_TALENT} /> and{' '}
+        <SpellLink spell={TALENTS.APOCALYPSE_TALENT} />.
+        <br />
+        <br />
```
emallson: don't use `<br/>` for layout, use `<p>` instead.

src/analysis/retail/deathknight/unholy/modules/features/WoundTracker.tsx
```diff
@@ -0,0 +1,318 @@
...
+          </>
+        </BoringSpellValueText>
+      </Statistic>
+    );
+  }
+
+  get guideSubsection() {
+    const barColor = this.getBarColor();
+    const explanation = (
+      <div style={{ lineHeight: '1.6' }}>
```
emallson: don't override line height for text content like this. if you think the line height for guide content should get changed, we can talk about that as a separate change.

src/analysis/retail/deathknight/unholy/modules/spells/VirulentPlagueEfficiency.tsx
```diff
@@ -0,0 +1,268 @@
...
+          actual: this.Uptime,
+          isLessThan: {
+            minor: 0.85,
+          },
+          style: ThresholdStyle.PERCENTAGE,
+        };
+  }
+
+  get VirulentDuration() {
+    return this.selectedCombatant.hasTalent(TALENTS.EBON_FEVER_TALENT) ? 13.65 : 27.3;
```
emallson, line 71: where did these numbers come from? the spelldata has Virulent Plague at a 27 second duration.

though it looks like this may predate your changes. so it doesn't need to be fixed as part of handling review comments, but maybe something to check in the future.

(this might have been someone building in latency tolerance)

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -77,6 +88,137 @@ class SoulReaper extends ExecuteHelper {
...
+            <div
+              style={{
+                position: 'relative',
+                height: '24px',
+                backgroundColor: '#1a1a1a',
+                borderRadius: '12px',
+                overflow: 'hidden',
+              }}
+            >
+              <div style={{ position: 'absolute', width: '100%', height: '100%', display: 'flex' }}>
```
emallson, line 134: i like these usage bars and think they're a good candidate for making available as a common guide component, but there's some work to do:

1. these bars should use the standard colors for good/bad. the colors used are *very* bright, and contrast with the background too much. it is okay to use some bespoke colors---especially in the case of adapting our common bar colors for use in text (where contrast is often not good by default)---but these should just be the standard colors

2. these bars should not be build with (primarily) inline css. this is a broader issue with your PR---some inline CSS is fine to patch specific points, but structural CSS like this is preferably not inline.

3. for consistency with other ui elements, these should not be fully rounded. rounded edges can look nice but break down when you're lining up another element against them (like the spell icon). rounding one side can be okay (like we do on the timeline). (the spell icons are also not properly aligned (as you can see in your screenshots) 

so my 2c is that the bar code here should be moved to a component like `PerformanceProgressBar` and then used like:

```tsx
<PerformanceProgressBar height={'24px'} progress={efficiency} performance={QualitativePerformance.Fail} ticks={[70, 85, 95]} />
```

that will then allow deduplicating the code between the other places that this component is used

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -77,6 +88,137 @@ class SoulReaper extends ExecuteHelper {
...
+        HP and survives the debuff.
+      </p>
+    );
+
+    const usageBar = (
+      <RoundedPanel>
+        <div style={{ display: 'flex', alignItems: 'center', padding: '10px' }}>
+          <div style={{ marginRight: '20px' }}>
+            <SpellIcon spell={TALENTS.SOUL_REAPER_TALENT} />
+          </div>
```
emallson, line 120: i am not as convinced that this container should be made a commonly available element. since you're re-using it, you could refactor it out into one within the unholy spec (and if other people start to use it, well, then we can just move it to `core`)

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -77,6 +88,137 @@ class SoulReaper extends ExecuteHelper {
...
+                <div style={{ width: '5%', backgroundColor: '#444' }} />
+              </div>
+              <div
+                style={{
+                  position: 'absolute',
+                  width: `${efficiency}%`,
+                  height: '100%',
+                  background: `linear-gradient(90deg, ${barColor} 0%, ${barColor}dd 100%)`,
+                  boxShadow: `0 0 10px ${barColor}66`,
+                  transition: 'width 0.3s ease',
```
emallson, line 161: this transition doesn't do anything since the React element goes away between time range changes, causing a full rerender and no animation.

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -77,6 +88,137 @@ class SoulReaper extends ExecuteHelper {
...
+              <div>
+                <PerformanceMark
+                  perf={
+                    efficiency >= 95
+                      ? QualitativePerformance.Perfect
+                      : efficiency >= 85
+                        ? QualitativePerformance.Good
+                        : efficiency >= 70
+                          ? QualitativePerformance.Ok
+                          : QualitativePerformance.Fail
```
emallson, line 205: you can use `evaluateQualitativePerformanceByThreshold` for this

src/analysis/retail/deathknight/shared/talents/SoulReaper.tsx
```diff
@@ -77,6 +88,137 @@ class SoulReaper extends ExecuteHelper {
...
+              <div>
+                <TooltipElement
+                  content={`${actual} Soul Reaper casts in execute phase out of ${possible} total`}
+                >
+                  <span style={{ color: '#aaa' }}>
+                    {actual} / {possible} procced ({efficiency.toFixed(1)}%)
+                  </span>
+                </TooltipElement>
+              </div>
+              <div>
```
emallson, line 196: this results in the performance mark being placed too far away on larger screens (screenshot on my work pc below), causing it to be visually disconnected from the stat. should just move the mark to precede the text as is conventional elsewhere

image

emallson (changes requested):
this is a big pr. i left a bunch of comments. i think my main high level concerns are:

1. there is a lot of duplicated JSX (hundreds of lines of it!). a lot of that will be fixed by refactoring out the bar component, so i think it'll be okay once that is done

2. you used a *lot* of bespoke inline CSS for margins, line heights, etc. some of this is necessary CSS for complex things like the progress bar + the container. those should ideally be migrated to not be inline in the component to help reduce noise. some of it is not needed, and should be removed. generally, i'd take a look at the CSS outside the container/progress bar layout code and see how much is just unneeded by switching your divs-with-margins to `<p/>`

3. there are a couple of logic errors, e.g. the soul reaper issue i commented on.

overall, though, this PR is looking good and with a bit of work will be mergeable

brandrewsss:
This codebase moved ahead too far to salvage the old branch, so I’m starting fresh.

PR #7403, opened 2025-07-07
Add buff filter to timeline

Seriousnes:
### Description

Updates the timeline component with a filter to show or hide buffs

### Testing

Any report -> timeline tab

- Screenshot(s):

All buffs
image

Filtered buffs
image


src/interface/icons/Cog.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+type Props = Omit<React.ComponentPropsWithoutRef<'svg'>, 'xmlns' | 'viewBox' | 'className'>;
+
+const Icon = (props: Props) => (
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" className="icon" {...props}>
```
emallson, line 6: quick check, where is this icon from? (just want to make sure if there are any copyright notice reqs we're good. there are a few permissive icon repos that you might have pulled from)

emallson (approved):
approval conditional on the cog icon being clear to use. rest of the code looks fine and the feature is good

src/interface/icons/Cog.tsx
Seriousnes: Icon is AI generated. I can find an appropriately licenced icon to replace it as I'm not too sure what the AI used to create it

emallson: that'd be great if you can. If not, I mean I don't think anyone will be coming after us for this

Seriousnes: According to Anthropic's terms of service, users own the content generated by Claude, provided there has been meaningful human input.

I had a look at the noun project, there are literally thousands of nearly identical cogs. I think its probably pretty safe to use the AI one

PR #7409, opened 2025-07-21
[Classic] MoP T14 Raids

jazminite:
### Description

- Add T14 raids (MSV, HoF, ToES)
- Add boss headshot icons
- Add placeholder image for bosses
- Remove Cata Raids

### Testing

- Test report URLs: `/report/yztfxH3CLZ7bQjmJ` and `/report/XgDC4Nxf7zYGktJw`
- Screenshot(s):
image
image
image


src/game/classic/ZONES.ts
```diff
@@ -1,65 +1,31 @@
 import { Zone } from 'game/ZONES';
-import { bwd, bot, totfw } from 'game/raids/cata_bwd_bot_totfw';
-import firelands from 'game/raids/cata_firelands';
-import dragonsoul from 'game/raids/cata_dragon_soul';
+import { msv, hof, toes } from 'game/raids/mop_msv_hof_toes';
 
 const ZONES: Zone[] = [
   {
-    id: 1023,
+    id: 1038,
```
emallson, line 6: this got changed on the WCL side from the structure you have. originally it was everything in 1038, but now it is MSV in 1038 and then HoF / ToES in 1040 (iirc). i think it should be easy to split them up (personally i don't think it'd be necessary to move them from one folder to another---a lot of boilerplate for what is still effectively t14).

emallson (changes requested):
thanks for doing this! one small change needed to the top-level raid objects. (i don't think it will have big ramifications for other things? though we might need to make revisions to the character rankings page to accommodate there being multiple current WCL zones)

otherwise, lgtm. one of the wcl folks will get a PR in for the images as soon as they're ready

emallson (approved):

PR #7413, opened 2025-07-27
Use DBC data to generate ability lists for Classic specs

emallson:
### Description

This PR uses game data to generate spell lists for Blood DK and Survival Hunter that are sufficient to power the Foundation guide.

My thought process here is that by continuing to refine gamedata extraction (see [here](https://git.emallson.net/?p=wow-dbc.git)), we can rapidly get to all mop specs having foundation support---and then the focus can be on building improvements to foundation, custom analysis for classes, etc instead of doing the much more tedious spell list updates.

#### Tooling / HOWTO

To set up spell list gen for a class, create a file that imports something named `spell-lists_CLASSNAME_SPECNAME.BRANCH` and then run `scripts/spell-lists/generate-spell-lists.ts` on it.

Then, you can use that import to fill out a `genAbilities` call to generate an `Abilities` analyzer with appropriate spell data.

My thinking is that long-term this allows us to have some shorthands in `package.json` that *don't* require painful manual repetition or painful babel/vite plugin stuff (see, for example, `scripts/e2e/generate-configs.ts`). Basically convention over configuration.

e.g. to regenerate all spell lists for a branch, we can use

```
node --experimental-strip-types scripts/spell-lists/generate-spell-list.ts src/analysis/classic/**/*.ts
```

Not certain this is the right approach, though, and open to other ideas on how to do this.

#### Blood DK

http://localhost:3000/report/L1rtQ7YHmZKcMy9a/4-Heroic+The+Stone+Guard+-+Kill+(4:19)/Bdk/standard/overview
image
image

the errors are:

- cd: bone shield (pre-pull cast timestamp is set at 0:00 but should be earlier, the bone shield buff is 5m long)
- gcd: looks like ptr lag? there's a cast gap and then a too-fast gcd (like really fast). this is a ptr log of CN players, so latency is high

#### Survival Hunter

http://localhost:3000/report/kZhQ1RLMXWYDvJAa/12-Normal+Elegon+-+Kill+(6:43)/Chìmalé/standard/overview
image
image

the unknown casts are our old friend Synapse Springs

#### Remaining Work

i would call the state of this as-is a "second pass". (the first pass was posted in `#maintainers` but never PRd because it was a mess)

it isn't ready for merge yet. there are a few more things that I need to do:

- [ ] make it easier to add analysis props to the spell book (`damageSpellId` / `buffSpellId` type things that don't require a full override)
- [x] add a way to exclude spells that are going to be added by analyzers (Kill Shot has this problem, as it is added by an `ExecuteHelper`)
- [x] figure out what the appropriate way to interop generated spell data with internal spell data is (the notable missing piece is the `icon` field, so generated spells cannot be passed to methods typed as `spell: Spell`. I know how to get this from the DBC, but it requires downloading a very large datafile containing the name of every asset in the game files)
- [ ] shared cooldowns: probably not going to be a part of the first pass because it is *so niche*, but Survival has shared trap cooldowns (all Fire traps share one, and all Ice traps, etc) but this is not currently captured by the generated spell data.
- [ ] improve the abilities list on the stats page. this isn't *bad* but it suffers a bit from auto-generation.
- [ ] probably don't do a git dependency to my personal server?

you will also notice that despite this being for MoP, the type is named `RetailSpell`. Cata is a transition point for a lot of internal game stuff (afaik) and i was able to repurpose the gamedata extraction i wrote for retail with almost no changes to work with MoP. I'll get around to renaming the type (i swear).

#### What Remains Manual

Cooldown resets and Executes are the big remaining manual bits. Thankfully, MoP doesn't have much in the way of hidden resets (they seem to be mostly buff-based). Survival benefited from the LnL module that was written in Cata, which means that the cooldown resets for that "just worked" (but really required manual work that was done last year). Many of these effects are heavily script-driven, so automatic generation is probably never going to be possible.

Haste effects are also still manual, but I have some [existing work](https://git.emallson.net/?p=badlands.git;a=blob;f=analysis/scripts/generate-haste.ts;h=f1427439f4bd63794e39a8e272da3625936bd425;hb=refs/heads/main) that I'm going to work on porting as well so that we can use gamedata to generate haste data. The list of all haste spells is way too big to put in a client-side json file, but having haste data be a part of the server response is possible.


emallson:
i added brewmaster, windwalker, and mistweaver and updated arms war. i think i'm going to push this to `next` so people can poke around? couple other notes:

- resources are hosed on MoP logs right now, which impacts all healers (not just MW). need a fix on either WCL or blizzard's side
- as i'm working through these, i'm definitely feeling how lackluster the foundation guide has remained. it was always intended to be expanded upon but i never really got around to it. i have some other prototyping i've done that i think will be worth incorporating. (@ me in discord if you want the full spiel) 
- i went back and forth on whether to merge these specs into this pr or make a follow-up since it is already a big one. i ended up pulling it into this one because i was making improvements to the spellgen along the way that were difficult to split out into a separate PR.
- i think it is worthwhile to confirm whether automation is actually producing speedups. i've written my fair share of "automations" that take more time in maintenance than they save in usage. my experience with this thus far is that *even with* some added debugging of `wow-dbc`, i'm getting through these foundational spec supports much faster than when i was doing it for cata. ymmv

Texleretour:
pog

PR #7425, opened 2025-08-09
add rotation section for MoP Brewmaster

emallson:
(i am not actually playing MoP so this is purely based on reading)

the rotation is based on the wowhead guide, but with some light editing to better match top logs. I've reached out to the MoP brew guidewriter to ask about the diff, because there's a real chance that the top logs are doing bad things but cheesing vengeance harder than people playing the rotation better.

### Testing

- Test report URL: `/report/KtATD3zqmdPC8nZc/30-Heroic+Elegon+-+Kill+(6:09)/Zenmonk/standard`
- Screenshot(s):
image

- Test report URL: `/report/NKJLHjgAy4bWMCht/2/5`
image

### DIFF
--- a/src/analysis/classic/monk/brewmaster/CHANGELOG.tsx
+++ b/src/analysis/classic/monk/brewmaster/CHANGELOG.tsx
@@ -1,4 +1,7 @@
 import { change, date } from 'common/changelog';
 import { emallson } from 'CONTRIBUTORS';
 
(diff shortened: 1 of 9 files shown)


emallson:
okay i caught up with the guidewriter and tl;dr there's a parsing rotation that dumps energy on RJW even on ST at the expense of Chi generation for defensives. they don't recommend it in the guide because it can be a noob-trap. I've updated the brew guide to show the more standard rotation by default and then the parsing one in a tab.

image

PR #7427, opened 2025-08-10
[Warrior] [Arms] [Fury] Add support for 11.2 changes

GeoffreyBalshaw:
### Description

Update Arms and Fury with new recommendations for 11.2. Note that Arms didn't actually have any APL changes, but the config was updated to reflect that it's still accurate for the new patch.

### Testing

Manaforge Omega hasn't been released yet, so no logs are available from live servers. However, some PTR logs are available and were checked

Reports: `/report/TxwHYQmKgLN4G3cM/35/4` and `/report/JGzcMqhfnR7yLdCv/17/20`

### DIFF
--- a/src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
+++ b/src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
@@ -26,9 +26,10 @@ import { addInefficientCastReason } from 'parser/core/EventMetaLib';
  */
 const ASHEN_JUGGERNAUT_DURATION = 15000;
 const SUDDEN_DEATH_DURATION = 12000;
-const BUFF_REFRESH_BUFFER = 6000;
+const BUFF_REFRESH_BUFFER = 3000;
 const RAMPAGE_RAGE_COST = 80;
-const MAX_MARKED_FOR_EXECUTION_STACKS = 3;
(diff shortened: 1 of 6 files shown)


src/analysis/retail/warrior/fury/modules/spells/SlayerExecute.tsx
```diff
@@ -168,7 +169,7 @@ class SlayerExecute extends Analyzer {
       this.overusedExecutes += 1;
       addInefficientCastReason(
         event,
-        'Execute was used without 3 stacks of Marked for Execution, or when neither Ashen Juggernaut or Sudden Death were near expiring',
+        'Execute was used without high stacks of Marked for Execution, or when neither Ashen Juggernaut nor Sudden Death were near expiring',
```
Krealle, line 172: You've changed the stack threshold for `Marked for Execution` to `1`, but mention "high stacks". Should maybe change it to "Execute was without Marked for Execution, or..." if you can use it with any amount of stacks?

GeoffreyBalshaw: Good call, this was actually a mid-PTR-cycle APL change, where it was 2 stacks (out of a max of 3), then 1 stack, then changed back to 2 stacks but I forgot to update the const locally. Updated now :)

PR #7431, opened 2025-08-11
Trim Breath of Eons filtered IDs

KYZ124:
Updates the Breath of Eons filter to avoid query failures.
Resolved an issue where Breath of Eons module was including itself and other Augmentation-caused damage in its evaluation of each player.


Krealle:
We prolly want to find a better solution to this ever growing filter long-term, but for now this works to un-break the module.

PR #7434, opened 2025-08-16
[Shadow] Update Shadow to 11.2

DoxAshe:
### Description

Shadow has once again received a talent rework.  This removes all the old talents that are no longer being used, updates existing talents to their new functionality, and adds talents that add new abilities.  In the future I will add the new talents in.

The talents affected are:
Idol of Nzoth
Idol of Y'shaarj
Deathspeaker
Void Volley

### Testing

There are a few changes I made that I am uncertain of.  The first is that Idol of Y'shaarj gives a reduction in haste.  It looks to me like the HASTE model can handle that, but I want to be sure.

Deathspeaker changes the execute range for shadow word death.  I didn't have a good way of doing this, but what I did works.  If there is a better solution I am open to changing to it.

- Test report URL: `/report/nGr7Pv3hWjJbCAZX/1-LFR+Plexus+Sentinel+-+Kill+(4:26)/Doxashe/standard/`
- Screenshot(s):


src/analysis/retail/priest/shadow/modules/spells/ShadowWordDeathSpeaker.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events from 'parser/core/Events';
+import Abilities from 'parser/core/modules/Abilities';
+import SPELL_CATEGORY from 'parser/core/SPELL_CATEGORY';
+import ExecuteHelper from 'parser/shared/modules/helpers/ExecuteHelper';
+import { SHADOW_WORD_DEATH_EXECUTE_RANGE_DEATHSPEAKER } from '../../constants';
+import DeathAndMadness from '../talents/DeathAndMadness';
+
+//Making a whole copy of shadow word death for when deathspeaker is active seems silly
+//But I don't konw of another way to update the lowerthreshold.
```
emallson, line 11: hmm, that kinda sucks but i also can't think of a great solution since `lowerThreshold` is static.

src/parser/shared/modules/Haste.ts
```diff
@@ -87,6 +87,8 @@ const DEFAULT_HASTE_BUFFS: HasteBuffMap = {
   //region Priest
   [TALENTS_PRIEST.POWER_INFUSION_TALENT.id]: 0.2,
   [SPELLS.BORROWED_TIME_BUFF.id]: 0.08,
+  [SPELLS.IDOL_OF_YSHAARJ_CALL_OF_THE_VOID.id]: 0.15,
+  [SPELLS.IDOL_OF_YSHAARJ_OVERBURDENED_MIND.id]: -0.1,
```
emallson, line 91: this will work in the sense that it won't cause haste values to break. it may not be precisely accurate, since the calculation of `removeHaste(event, 0.1)` produces slightly different values from `addHaste(event, -0.1)`

for example:
```
removeHaste(1.2, 0.1)
(1.2 - 0.1)/(1 + 0.1) = 1.1 / 1.1 = 1

addHaste(1.2, -0.1)
1.2 (1 - 0.1) - 0.1 = 1.2 - 0.12 - 0.1 = 0.98
```

so the better solution is probably to patch this to switch the `addHaste`/`removeHaste` calls if the haste value is negative. *but* this is likely close enough to not cause problems. there are only a couple calls of each, and you don't need to touch the ones that involve haste rating, so it shouldn't be *too* big a patch but if its going to delay this pr much we can just merge and you can come back to it

emallson:
also the e2e failure is brewmaster related, not related to your code

src/parser/shared/modules/Haste.ts
DoxAshe: I think I can fix that pretty quickly by changing `removeHaste` and `addHaste` to:



```
static addHaste(baseHaste: number, hasteGain: number) {
    if(hasteGain < 0) //removeHaste Instead
        {return (baseHaste - Math.abs(hasteGain)) / (1 + Math.abs(hasteGain))}
    return baseHaste * (1 + hasteGain) + hasteGain;
}
```


and 

```
static removeHaste(baseHaste: number, hasteLoss: number) {
    if(hasteLoss < 0) //addHaste Instead
        {return baseHaste * (1 + Math.abs(hasteLoss)) + Math.abs(hasteLoss)}
    return (baseHaste - hasteLoss) / (1 + hasteLoss);
}
```

emallson: 👍 you might consider having them call each other, but i think its also fine to make them symmetric like that instead.

DoxAshe: I tried that first, but it didn't allow for them to both call each other.

emallson (approved):

PR #7452, opened 2025-08-26
[Ret] More guide updates and Statistics fixes

Texleretour:
### Description

This is a two part PR.
**Statistics page updates and spec fixes** :
([small fixes to the statistics page](https://github.com/WoWAnalyzer/WoWAnalyzer/commit/91aa724181d87bbd450b93d8789e90681deba61b)
[add crusadin strikes to abilities as Hidden to avoid spellusable warning](https://github.com/WoWAnalyzer/WoWAnalyzer/commit/ffef020f2a26ef485dc1c05167f046c276813bd8))

- Correctly marked Hammer of Light as 5HP instead of 3.
- Remove duplicate Execution Sentence in the abilities sections (now only shows in cooldowns)
- Remove Consecration from Abilities so that is isnt showed in the stats page anymore (it's become 100% passive)
- Change Crusading Strikes icons to stay coherent with the in game icon (in game, the one Crusading Strike that generates HP does not have a different icon like it has on wowhead)
- Remove Crusading Strikes from Abilities so that is isnt showed in the stats page anymore (it's 100% passive)
- Remove Blessing of Dawn stat because uptime does not mean anything anymore since its rework. percentage of spenders affected could be interesting to some people though it has no gameplay interactions.

**Guide update**
Add a Cooldown Usage section similar to Havoc. This tracks every cast of Wake of Ashes (now our most important cooldown thanks to Radiant Glory).
It tracks the following:
- Whether or not Execution Sentence was already on target (should be)
- How many Hammer of Light casts were fit into the window (expecting 1+ and 2+ with 4 piece)
- If Divine Hammer was cast during cooldowns (should be cast before. Only displayed if it happened)

### Testing

`/report/pXdNqrLYG4cwM8tC/16/6` with a lot of mistakes and no 4 piece
`/report/Cxq9LjpZQ2kBdftK/49-Heroic+Loom'ithar+-+Kill+(4:00)/Amenadiell/standard` with less mistakes and 4 piece
`/report/pXdNqrLYG4cwM8tC/16-Heroic+Loom'ithar+-+Kill+(6:25)/Poulpitus/standard` between the two, not full fail or full good
- Screenshot(s):
image
image
image
image
image

### DIFF
--- a/src/analysis/retail/paladin/retribution/CHANGELOG.tsx
+++ b/src/analysis/retail/paladin/retribution/CHANGELOG.tsx
@@ -1,7 +1,10 @@
 import { change, date } from 'common/changelog';
+import { TALENTS_PALADIN } from 'common/TALENTS';
 import { Texleretour, Vetyst } from 'CONTRIBUTORS';
+import SpellLink from 'interface/SpellLink';
 
 export default [
+  change(date(2025, 8, 22), <>Add {<SpellLink spell={TALENTS_PALADIN.WAKE_OF_ASHES_TALENT} />} cast analysis.</>, Texleretour),
   change(date(2025, 8, 22), 'Add Expurgation uptime bar and add Divine Hammer to the Cooldown Graph', Texleretour),
   change(date(2025, 8, 4), 'Few bugfixes and old modules cleaning', Texleretour),
   change(date(2025, 6, 9), 'Basic 11.1.5 support', Vetyst),
--- a/src/analysis/retail/paladin/retribution/CONFIG.tsx
+++ b/src/analysis/retail/paladin/retribution/CONFIG.tsx
@@ -8,8 +8,8 @@ import { Texleretour } from 'CONTRIBUTORS';
 const config: Config = {
   contributors: [Texleretour],
   branch: GameBranch.Retail,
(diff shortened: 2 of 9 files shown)


emallson (approved):
as a note, we generally want to gate analysis like this by checking for the talent. however, if something is *so fundamental* to the way the spec plays right now, it is okay to skip the check and just assume everyone has it. afaik radiant glory is like that (from my brief looking, it appears that that talent has 100% play rate)

Texleretour:
> as a note, we generally want to gate analysis like this by checking for the talent. however, if something is _so fundamental_ to the way the spec plays right now, it is okay to skip the check and just assume everyone has it. afaik radiant glory is like that (from my brief looking, it appears that that talent has 100% play rate)

Thanks for the comment. Indeed Radiant Glory as well as anything hero talent related is hard locked rn, other options seem largely behind in numbers...
As I have a few more things coming, I can add some talent checking at the same time if that's okay (I realize now that I've forgotten to add support for Final Reckoning -choice node with Execution Sentence-. It's nearly not played but I would not be surprised to see it in some logs like Soulbinder for instance.)

emallson:
that sounds good 👍

PR #7455, opened 2025-09-02
Survival updates

Kivlov:
### Description

Update survival for 11.2
Add hero talent spell IDs from the 11.1 Pack Leader revamp. There is 4 buffs, 3 of which spawn a beast, and the 4th is just the tracker for the next beast.

Explosive shot and Kill shot have largely fallen out of use in ST.

--
Analyzer for Howl of the Pack Leader and the Hunter T4 for Pack Leader.
This is a big one, both modules I am willing to remove and put them in their own pulls if they make this pull too large.

This may be a mis-use of the Qualatative performance, but I used the Perfect/Good/Ok to colour each spawn .There isn't any really "bad" uses of the beasts, you're supposed to hold the bear for a little bit to make sure Rend Flesh hits multiple targets, and survival cares about boar strider (which is what most of this module is dealing with.) so I thought it might be handy to just make each spawn easier to find. I can also split it into 3 sub sections, one for each easy enough.

--

### Testing

Test log has some AoE to it to see what the boar strider looks like.
- Test report URL: `https://www.warcraftlogs.com/reports/qmTx6JhgLAk1HRGV?fight=42&type=damage-done&source=34`
- Screenshot(s):
image
I pre-pull wildfire bomb, so many of the red are it thinking wildfire is on cooldown. It is flagging flanking strike as having no available charges, but I suspect this is because the pet is casting flanking strike and it's considering that as an illegal cast.


src/analysis/retail/hunter/shared/talents/KillShot.tsx
```diff
@@ -42,6 +48,11 @@ class KillShot extends ExecuteHelper {
     const ctor = this.constructor as typeof ExecuteHelper;
     ctor.executeSpells.push(this.activeKillShotSpell);
 
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS.KILL_SHOT_SURVIVAL_TALENT),
+      this.onSVCast,
+    );
```
emallson: this is added to the *shared* Kill Shot implementation. i checked a couple BM/MM logs and don't see Kill Shot casts. has it been removed from those specs?

if not: you should subclass this analyzer in the Survival folder and add the Survival-specific logic there.

if so: you can just move this to the survival folder as-is

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
```diff
@@ -0,0 +1,680 @@
...
+        </p>
+      </>
+    );
+
+    return (
+      <SpellUsageSubSection
+        title="Howl of the Pack Leader"
+        explanation={explanation}
+        uses={this.uses}
+        castBreakdownSmallText="- Colored bars: Wyvern (Perfect), Bear (Good), Boar (Ok). Click for per-spawn details."
```
emallson: i pulled this up locally to see what it looked like and, yea, don't do this. the colors have semantic meaning in this context.

image

since you're not actually using the performance value here, you might consider using the cooldown breakdown instead. here's an example from Resto Druid:

image
 
i am a little concerned that there will be too many pack leader instances for this to work well, so you might be best off doing as you mentioned in the PR and just separate the subsections for each type of beast.

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
```diff
@@ -0,0 +1,680 @@
...
+
+    if (this.isSurvival) {
+      this.addEventListener(
+        Events.cast.by(SELECTED_PLAYER).spell(TALENTS.KILL_COMMAND_SURVIVAL_TALENT),
+        (e: CastEvent) => {
+          const enemy = this.enemies.getEntity(e);
+          this.kcHistory.push({ ts: e.timestamp, targetId: e.targetID, targetName: enemy?.name });
+          if (this.kcHistory.length > KC_HISTORY_MAX) this.kcHistory.shift();
+        },
+      );
```
emallson: prefer writing out methods longer handlers like this. having a one-liner in the constructor is okay for short handlers (like the `startSpawn` handlers below) but having like 10 inline, anonymous event handlers of significant length is a pain to read.

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
```diff
@@ -0,0 +1,680 @@
...
+    return items;
+  }
+
+  private formatDamage(num: number) {
+    if (num === null || num === undefined) return '0';
+    if (Math.abs(num) >= 1.0e9) return (num / 1.0e9).toFixed(1) + 'B';
+    if (Math.abs(num) >= 1.0e6) return (num / 1.0e6).toFixed(1) + 'M';
+    if (Math.abs(num) >= 1.0e3) return (num / 1.0e3).toFixed(1) + 'K';
+    return num.toString();
+  }
```
emallson: we have a helper for this. `formatNumber` in `common/format`. if we need billions (because TWW numbers are silly), we can add it there. don't need a separate helper.

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
```diff
@@ -0,0 +1,680 @@
...
+//region Constants
+
+const CHARGE_CLUSTER_MS = 1500; // capture window per charge
+const BEAR_WINDOW_MS = 12_000; // bear’s full effect within 12s of spawn
+const OPEN_UNTIL_FAR_FUTURE = 9e15; // effectively “infinity”
+const KC_HISTORY_MAX = 50; // guard against unbounded growth
+
+//endregion
+//region Analyzer
+export default class HowlOfThePackleaderGuide extends Analyzer.withDependencies({
```
emallson: i'm not going to pick through all of the details of this. the analysis is hard to follow, but seems *mostly* correct. Wyvern seems to often have trouble tracking the buff, which may indicate that your spawn linking logic has a subtle error. for example [this log](https://www.warcraftlogs.com/reports/txFckrWYJ7m2fBAj?fight=8&type=damage-done&source=21) has "no uptime recorded" for 4 of 10 wyvern spawns.

but anyway, the point i actually wanted to make is that a lot of this analysis seems like it would be significantly simplified with some event links. I see you picked up use of them for Stampede, but for example:

- link Howl buff apply to the next KC (or remove to the previous KC), then you don't need to have this KC deque within the module
- link Wyvern's Cry to the most recent Wyvern Howl removebuff so you don't need to track it independently
- link Hogstrider to the Boar Howl removebuff
- etc

basically a lot of the machinery for "I want to find <events> related to <other event>" can be handled neatly with `EventLinkNormalizer` without needing to implement all of the tracking yourself.

(there are some tricks to this, like being careful limit to 1 linking event and then applying reverse links)

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
```diff
@@ -0,0 +1,680 @@
...
+          <b>Wyvern</b>: <SpellLink spell={SPELLS.WYVERNS_CRY} /> applies a % damage buff.
+          <br />
+          <b>Bear</b>: Summons a bear that applies <SpellLink spell={SPELLS.BEAR_REND_FLESH} />. Aim
+          to maximize the number of targets hit when spawning a bear.
+          <br />
+          <b>Boar</b>: A boar charges your target with a ~3s delay before charging again for a total
+          of 3 charges. Each charge deals damage and applies{' '}
+          <SpellLink spell={SPELLS.HOGSTRIDER_BUFF} />. In AoE/Cleave for Survival, consume
+          Hogstrider before the next charge.
+        </p>
```
emallson: don't use `<br/>` for formatting like this. (i let it slide in tooltips because those can be finicky)

src/analysis/retail/hunter/shared/herotalents/Stampede.tsx
```diff
@@ -0,0 +1,270 @@
...
+    if (unique_targets === 1) {
+      if (actual_ticks >= expected) return QualitativePerformance.Perfect;
+      if (actual_ticks >= expected - stampede_count) return QualitativePerformance.Ok;
+      return QualitativePerformance.Fail;
+    }
+    const ratio = actual_ticks / expected;
+    if (ratio == 1) return QualitativePerformance.Perfect;
+    if (ratio >= 0.75) return QualitativePerformance.Good;
+    if (ratio > 0.5) return QualitativePerformance.Ok;
+    return QualitativePerformance.Fail;
```
emallson: couple things:

1. don't use braceless `if`s. (i should check and see if we can add that to eslint/prettier---though prettier might have actually produced these)
2. you can use `evaluateQualitativePerformanceByThreshold` to implement both of these checks

src/analysis/retail/hunter/shared/herotalents/Stampede.tsx
```diff
@@ -0,0 +1,270 @@
...
+      return [s_ts, e_ts] as [number, number];
+    });
+    const intervals = merge_intervals(raw_intervals);
+
+    const dmg_in_intervals: DamageEvent[] =
+      intervals.length === 0
+        ? []
+        : damages.filter((d) => intervals.some(([a, b]) => d.timestamp >= a && d.timestamp <= b));
+
+    type DamageEventWithInstance = DamageEvent & { targetInstance?: number };
```
emallson, line 166: what's the story here? linter complaining about `??` on a non-optional field? (`DamageEvent` has `targetInstance` on it but it isn't marked as optional---probably should be though?)

src/analysis/retail/hunter/survival/modules/guide/sections/rotation/ActiveTime.tsx
```diff
@@ -57,6 +58,7 @@ export default function ActiveTime({ modules, events, info }: GuideProps<typeof
         fightStart={info.fightStart}
         fightEnd={info.fightEnd}
       />
+      <FoundationDowntimeSection />
```
emallson, line 47: probably don't need both graphs here

emallson (commented):
looks like quite a bit of work here :sweat_smile: 

i'm not gonna lie, it was hard to review the Howl and Stampede analyzers.

the Howl analyzer's issues are more structural: related code is spread between the constructor + scattered methods, it is hard to verify that the spawn lookups are going to work correctly, Howl is kinda sorta 3 analyzers (one per beast) jammed into one module, etc.

the Stampede analyzer i think mostly needs a brief comment describing what is happening (especially since the Stampede analyzer is secretly a Lead from the Front analyzer, so really its "when LftF happens, collect each stampede during it using link data + <whatever heuristic> to group them and check each stampede")

i made a comment about using event links more in Howl, since afaik you can't rapid-fire Howl triggers so fast that you start losing track. Stampede is different, based on my reading, so i'm not sure that more event links help there.

a cleanup/doc pass would be good

----

also, while i was looking at this i checked on Flanking Strike. it looks like the talent id for flanking strike (which is in `Abilities` as one of the cast ids) isn't actually a real cast. the spell data has 2 dummy effects, while the one you have listed as `FLANKING_STRIKE_PLAYER` has the actual damage effect etc. if you want to fix those casts, you should:

1. remove `FLANKING_STRIKE_TALENT` from `Abilities`
2. add `FLANKING_STRIKE_TALENT` to `CASTS_THAT_ARENT_CASTS`

src/analysis/retail/hunter/shared/talents/KillShot.tsx
Kivlov: No, it's still there (kind of). MM has Black Arrow which I believe has it's own analyzer and replaces killshot, and BM just literally never presses it anymore. It's a really bad button for all 3 classes, except when it's Black Arrow for Dark Ranger. SV also has their own spell ID for it, so was trying to keep it within the shared but not interfere with the existing logic that was there which pretty much just tracked onCast I think. 

It's a common theme I see when doing log reviews is people are still hitting it for survival but you only wanna cast it when <30 focus or out of melee so I could just split it into it's own SV subclass then since BM doesn't use it rotationally currently and MM has BA.

emallson: yea, you can just do a `KillShot extends SharedKillShot` in the survival folder then

src/analysis/retail/hunter/shared/herotalents/HowlOfThePackLeader.tsx
Kivlov: Alright. I think I'll just drop Wyvern entirely because.. kind of pointless to track it and was torn between leaving it out for that reason and for including it since it's part of the set of spawns. Bear also isn't super useful to track, almost nobody as SV holds the bear because of other rotational pressures which just leaves the boar.

emallson: IMHO it is okay to omit Wyvern/Bear if they aren't really things that people track & make decisions around.

for comparison, Brewmaster has Wisdom of the Wall that can give 4 different buffs, but there is only 1 that people really play around (Shadow) so that is the only one the brew analyzer covers.

Kivlov:
Removed howl analyzer. It is important for SV aoe, unecessary for PL AoE and so I'm going to just work on it when I go to work on AoE.

Removed the active time graph. I had initially liked having both since one was quite good at-a-glance but the more I thought about it the more I agree that don't need both.

Made adjustments to Stampede. Removed the merge on the intervals as we can't parse out which damage belongs to which, and so collecting the damage from the whole interval works fine. It belongs to the 4p, and lead from the front has no further considerations to gameplay once Stampede goes away, so despite really being a LFTF analyser, it's named Stampede as it will go away when Stampede goes away.

src/analysis/retail/hunter/shared/herotalents/Stampede.tsx
Kivlov: I believe it was a linting issue, but I can't recall now

emallson (approved):
the changes look good! thanks for taking the time :)

PR #7473, opened 2025-09-22
Cooldown Reductions for Markmanship

brothernog:
Added basic cooldown reductions for Aimed Shot from Ammo Conservation, Precise Shots sent, updated Lock and Load constants (up from a 8% chance last patch to a 10% chance) and also applied the same logic for explosive shots cooldown reductions from Explosive Gunpowder Talent.

Although they seem to be working, a very few Aimed Shots/Explosive Shots cooldown reductions are still not being correctly trackeable by the parser, it could be because:

- Precise Shots get removed in the same GCD, so the log doesnt indentify that there is a Precise Shot buff. In events, there is a rare case where you don't see the "Buff Gained" event for Precise Shots after an Aimed Shot cast.
- On Aimed Shot, it could be because of randomly haste benefits throughout the fight that doesnt get correctly calculated.

Either the way, it needs some debugging, although I'm still not expert enough to test the reasons above.

The before and after:

image

image


src/analysis/retail/hunter/marksmanship/modules/talents/FocusedAim.tsx
```diff
@@ -0,0 +1,35 @@
...
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS_HUNTER.FOCUSED_AIM_TALENT);
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.PRECISE_SHOTS),
+      this.onPSRemoved,
+    );
+  }
+
+  onPSRemoved = () => {
```
emallson: minor style nit: this should be something like: `private function onPSRemoved() { ... };`

src/analysis/retail/hunter/marksmanship/modules/talents/MagneticGunPowder.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onAimedCast() {
+    if (
+      this.selectedCombatant.hasBuff(SPELLS.LOCK_AND_LOAD_BUFF.id) &&
+      this.spellUsable.isOnCooldown(TALENTS_HUNTER.EXPLOSIVE_SHOT_TALENT.id)
+    ) {
+      this.spellUsable.reduceCooldown(TALENTS_HUNTER.EXPLOSIVE_SHOT_TALENT.id, 8000);
+    }
+  }
+  onPSRemoved() {
+    const base = this.selectedCombatant.hasTalent(TALENTS_HUNTER.FOCUSED_AIM_TALENT) ? 4000 : 2000;
```
emallson: what's the story here? i don't see this interaction listed on either the Focused Aim or Magnetic Gunpowder talent.

emallson (commented):
apologies for the super delayed review. changes look good overall. even if it isn't perfect, this is still an improvement.

the only thing i want before merging is some info on that bit in the Magnetic Gunpowder analyzer related to Focused Aim

src/analysis/retail/hunter/marksmanship/modules/talents/MagneticGunPowder.tsx
brothernog: Sorry, thanks for calling my attention on this. The line is actually for the Windrunner Quiver talent, not for the Focused Aim talent.

Windrunner Quiver makes PS 2 stacks so it reduces the cooldown of explosive shot by 4 seconds instead of 2 seconds.

PR #7474, opened 2025-09-24
use average instead of median for support stats.

emallson:
tl;dr - the calculation of median on the server is wrong due to the interaction of `GROUP BY` and `OVER (window)` in MariaDB/MySQL. There is no median aggregate function, and switching fully to window functions makes performance *even worse,* so this simply removes the use of the median.

This is why some specs would flip between 0% active time and some high % while having high % average active time (seen in the tooltip) with effectively no CD errors, etc.

There will be a separate backend PR to remove the median calculation.

### Before
image

### After
image

### DIFF
--- a/src/common/server-metrics.ts
+++ b/src/common/server-metrics.ts
@@ -25,7 +25,6 @@ export interface Aggregate {
   min: number;
   max: number;
   avg: number;
-  median: number;
 }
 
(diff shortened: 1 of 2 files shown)


trevorm4 (approved):
lgtm

PR #7475, opened 2025-09-27
Hunter - BM - Ability Updates

Kivlov:
### Description

Moved Exhil/Damage taken to Shared as I plan to implement it into BM. Updated my reference to it in the Survival CombatAnalyzer. Would've just added them in this PR but Smoke Screen, a Dark Ranger talent casts a free use at half power of Exhil or SotF when you use one or the other and so it is misleading until those are handled.

Dire Beast is no longer a cast so analysis of it's use was pruned down. 

Before ability changes
image

After Ability Changes
image

The remaining yellow are Forgeweaver and Netherprism trinkets (I couldn't figure out how to get them to register when I added their IDs to the common/spells/thewarwithin/trinkets.ts so I removed that. A couple of the yellow are the Phase Blink from the cloak on Plexus Sentinel.

The one red cast is Aspect of the Cheetah during the free use from Evokers (I think, I can't explain otherwise why I'd have been able to use it, AotC is a 3 min cd still.)
### Testing

- Test report URL: report/vkgjGRDtLPcyTn6C/7-Mythic+Plexus+Sentinel+-+Kill+(5:15)/Percival/standard/debug

I do have follow up Dark Ranger analysis to add in (Not the other PR with the graph) but sending this one in now to help keep the PR smaller.


src/common/TALENTS/hunter.ts
```diff
@@ -421,9 +421,9 @@ const talents = {
     definitionIds: [{ id: 131305, specId: 255 }],
   },
   DIRE_BEAST_TALENT: {
-    id: 120679,
+    id: 304051,
```
emallson: this is a generated file and changes to it will be un-done the next time we update talents. looking at the talent tree on wowhead, this is the correct spell id / icon for the talent itself, so the summon cast (in `SPELLS` now) being updated and different is correct. the change to this file should be un-done.

emallson (changes requested):
one minor change but otherwise looks good

emallson (approved):

PR #7488, opened 2025-10-15
replace babel with oxc and swc

ToppleTheNun:
### Description

use oxc to replace most of the cursed generate-configs-bootstrap.cjs
script.

use swc to bundle the app via vite, including using the swc emotion
plugin and swc lingui plugin.

### Testing

- Test report URL: `/report/fCx3FXwYRQ1pdLya/8-Mythic+One-Armed+Bandit+-+Kill+(6:43)/Meldmedaddy/standard`
- Screenshot(s):
image
image


emallson:
confirming my understanding: oxc isn't a drop-in replacement for swc, just happens to be named similarly.

so swc is used for performance within lingui/vite, and oxc is used to eliminate some spaghetti from our e2e tests. is that right?

assuming this all works right, i'll probably retarget it at `the-war-within`. or at least do some kind of merge-into-both thing to get the library updates in that branch sooner rather than later

ToppleTheNun:
yep! At some point, Vite will be using oxc once it switches to using rolldown instead of rollup, but that'll be later.

swc is faster than babel for the React side of things; oxc is faster and lets us reduce complexity for the E2E script.

emallson:
`next` has a version of this that is based on `the-war-within`

emallson:
i merged this via the TWW pr and then merging `the-war-within` into `midnight`, so i'm closing this PR

PR #7490, opened 2025-10-16
Implement Aug tier sets

KYZ124:
Implement T35 module. Remove remaining TWW tier module.

Currently tiers are permanently inactive as Midnight set bonuses aren't on Wowa yet, will be corrected when possible.


src/analysis/retail/evoker/augmentation/modules/abilities/EbonMight.tsx
```diff
@@ -264,12 +265,16 @@ class EbonMight extends Analyzer {
     const critChance = this.stats.currentCritPercentage;
     const critMod = 1 + SANDS_OF_TIME_CRIT_MOD * critChance;
 
+    // Update once Midnight tiers are implemented
+    //const eruptionExtension = this.selectedCombatant.has2PieceByTier(TIERS.MIDNIGHT1) ? ERUPTION_EXTENSION_MS + T35_AUGMENTATION_2PC_EXTENSION_MODIFIER : ERUPTION_EXTENSION_MS;
```
Krealle, line 269: You should make a `this.eruptionExtension` instead of trying to calculate it every time, since it's a static value.

Krealle (changes requested):
Overall looks good, just got a couple of things that needs to be changed.
You should rename `T35` to `MID1` since that is what Midnight tiers will be called. I've also added the tier set ids for `MID1` so you add them in here.

PR #7493, opened 2025-10-22
[Core] Guide Rework Part 1

Sharrq:
I am working to expand the available components for people to use and to rework some of the existing components so they are a little easier to work with and more consistent. I plan on doing this in chunks, with this being the first chunk.

I updated CastEfficiencyBar to accept a spell object instead of a Spell ID. I then changed every instance of that component to send the spell object instead. This brings it more in line with other components that have been shifting towards spell objects and accounts for 90% of the files changed.

For the remainder of these, i tried to give them all a uniform style throughout so they all look like they belong together. Some of these duplicate existing things but that is largely because i didnt want to force existing components over to them and break people's stuff (though i could do that in a later PR if people like them).

For BuffUptimeBar, i condensed the chart and moved the stats into their own boxes. I am still not sure how i like the condensed chart and might explore different ways of displaying that information.
Screenshot from 2025-10-22 19 54 12
Screenshot from 2025-10-22 19 40 50

CastDetail is a new component that can show per cast information in a more detailed way while still highlighting the performance for the cast overall and some individual performance on portions of that cast. This was in my mage rework and i redesigned it to bring it inline with the style i built in this PR. The boxes in the top right that show the number of perfect, good, etc casts are clickable to filter the results.
Screenshot from 2025-10-22 19 44 18

CastSummary is the component that everyone is used to using, but bundled together into a single component instead of building it manually around CastSummaryAndBreakdown or GradiatedPerformance. Additionally this new component can do both of those with the toggling of the showBreakdown parameter for when you want to show the individual cast boxes when you click on the summary bar. The contents of CastSummaryAndBreakdown were duplicated into here to open up the ability to deprecate CastSummaryAndBreakdown in favor of this at some point.
Screenshot from 2025-10-22 19 47 55
Screenshot from 2025-10-22 19 48 09

CastSequence is a new component that allows you to specify a cast window (like x seconds before and/or after an event) and get all the cast events in that window to display what spells the player cast in that window. Useful for things like Combustion to show their cast sequence during the combustion window, or showing the pre setup and buff window for something like Touch of the Magi.
Screenshot from 2025-10-22 19 50 18

StatisticCard is similar to CastDetail, but allows you to show a single component with multiple data points or statistics. Might rename this to something like CastOverview, but it lets you add however many statistics you want and it will just keep appending them. Additionally you can tie a performance color to some or all of them to color the box they are in, or omit them and it will just color them gray 
Screenshot from 2025-10-22 19 52 39


Sharrq:
Might be worth consolidating some of the stylized elements/css stuff because there is a lot of duplication of those between the different components, but i kept them separate for now.

Sharrq:
Added TipBox and added it to CastDetail to show a summary detail to explain the performance. TipBox is meant to be very generic to be used wherever you want to display a quick tip or suggestion or whatever.
Screenshot from 2025-10-22 22 59 26

Sharrq:
Reskinned the BuffUptimeBar graph to look a little bit better
Screenshot from 2025-10-23 18 07 07

Also renamed StatisticCard to CastOverview

src/parser/ui/CastEfficiencyBar.tsx
```diff
@@ -87,21 +101,51 @@ export default function CastEfficiencyBar({
...
+      )}
+      {hideEfficiency ? (
+        <CooldownBar
+          activeWindows={activeWindows}
+          spellId={rawSpellId}
+          gapHighlightMode={gapHighlightMode}
+          minimizeIcons={minimizeIcons}
+          slimLines={slimLines}
+        />
+      ) : (
```
emallson: this param is unnecessary, right? you can just use the `CooldownBar` component directly? or is there a gotcha with that that i'm not seeing in my reading of this

src/parser/ui/CastEfficiencyBar.tsx
```diff
@@ -24,32 +24,46 @@ interface Props {
...
   activeWindows,
+  hideEfficiency,
+  showExplanation,
 }: Props & {
   useThresholds?: boolean;
 }): JSX.Element {
-  const rawSpellId = spell ? spell.id : spellId!;
-  const spellObj = spell ? spell : spellId!;
+  const rawSpellId = typeof spell === 'number' ? spell : spell.id;
+  const spellObj = typeof spell === 'number' ? spell : spell;
```
emallson: this ternary looks wrong. both branches are the same value?

src/interface/guide/components/GuideDivs.tsx
```diff
@@ -67,3 +67,252 @@ export const PanelHeader = styled.div`
...
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+`;
+
+// ==================
+// Stats Grid Components (for detail views)
+// ==================
+
+/** Grid layout for displaying multiple stat items */
+export const StatsGrid = styled.div`
```
emallson: these elements that are internal to other components shouldn't be exported in a common location like this. some of these are probably okay, but others will break their respective components if changed. i'd suggest reverting the part of a801a66 that moves these for now so that they don't end up in use (yet).

if there are specific ones you think are generally useful, they can remain here (or can be exported from the source component)

src/interface/guide/components/CastSequence.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  const handlePrevious = () => {
+    setCurrentIndex((prev) => (prev > 0 ? prev - 1 : sequences.length - 1));
+  };
+
+  const handleNext = () => {
+    setCurrentIndex((prev) => (prev < sequences.length - 1 ? prev + 1 : 0));
+  };
```
emallson:
```suggestion
  const handlePrevious = useCallback(() => {
    setCurrentIndex((prev) => (prev > 0 ? prev - 1 : sequences.length - 1));
  }, [sequences.length]);

  const handleNext = useCallback(() => {
    setCurrentIndex((prev) => (prev < sequences.length - 1 ? prev + 1 : 0));
  }, [sequences.length]);
```

to prevent spurious rerenders

src/interface/guide/components/GuideDivs.tsx
```diff
@@ -67,3 +67,252 @@ export const PanelHeader = styled.div`
...
+  font-weight: 700;
+  color: #fff;
+  line-height: 1;
+  margin-bottom: 2px;
+  text-shadow: 0 1px 3px rgba(0, 0, 0, 0.5);
+`;
+
+/** Stat label (small descriptive text) */
+export const StatLabel = styled.div`
+  font-size: 10px;
```
emallson: 10px is generally too small. i know we have some existing uses of fonts this small, but we probably shouldn't in most cases.

i know a bunch of these were a part of a previous PR and I'm not going to demand changing all of them right now. however, please adjust the 10px and 11px fonts up to at least 12px / 1.2rem (and preferably a little larger, like 14px, if it fits in the UI) 

as an additional note: prefer scalable units (like `rem`) instead of fixed-size units (like `px`) because they will be handled better by browsers on more combinations of screen sizes / dpis. (i actually get some minor pixelation / artifacting at 10px on my hidpi work monitor)

as an aside: our `html` element (which determines `rem` scaling) has `font-size: 10px` set, which both makes conversion to `rem` easy and also is not good as a default font size (though we immediately change in `body` to `font-size: 14px`, which is much more reasonable). in a lot of modern codebases, `1rem` is a good body font size (set to something like 14 or 16, or unset to let the browser handle it), but for us it is definitely too small.

src/interface/guide/components/CastDetail.tsx
```diff
@@ -0,0 +1,316 @@
...
+      } else {
+        newSet.add(performance);
+      }
+      // Reset index if current cast is filtered out
+      if (currentIndex >= filteredCasts.length) {
+        setCurrentIndex(0);
+      }
+      return newSet;
+    });
+  };
```
emallson: same note about `useCallback`

emallson (changes requested):
some small-ish changes, but things mostly look fine.

src/interface/guide/components/GuideDivs.tsx
Sharrq: I'll look through all of these, but did want to comment on this one in particular. I might have moved too many of them to guide divs but the reason I did was because multiple of the components I added were using the same divs and they were statically featured l created in each file, in some cases with minor differences that caused them to have subtle unnecessary differences. So I consolidated just the ones that were duplicated (or at least I was trying to only target the duplicated ones, I might have accidentally moved too many. 

I can undo all of that if you want, but that was why I did it

emallson: maybe i misunderstood. it is hard to track them. if they're re-used, this is probably fine. especially if you'd want to keep changes in-sync between them.

Sharrq: Yea I'll double check but I'm pretty sure all the ones I moved are used in at least 2 of the components (largely because they all use similar layouts).

src/parser/ui/CastEfficiencyBar.tsx
Sharrq: Yea this was a misunderstanding of something i told copilot and i didnt catch it when i was looking stuff over. I removed it and the associated hideEfficiency parameter. Ill probably do more with this component in the next PR to bring it in line with the style i have been using for the components in this PR.

src/parser/ui/CastEfficiencyBar.tsx
Sharrq: I removed both of these, set the component to only accept a Spell instead of Spell | number and then set the references to spell.id instead of these weird ternaries

src/interface/guide/components/GuideDivs.tsx
Sharrq: There was only one that wasnt being used in multiple components (InsetContainer). I moved that out and kept the rest there

src/interface/guide/components/GuideDivs.tsx
Sharrq: I converted all these to rem and made some of the smaller ones bigger.

emallson (approved):

PR #7494, opened 2025-10-25
remove Checklist from the codebase

emallson:
This removes the `Checklist` and related code from the codebase. This feature has had waning use since the introduction of the `Guide` but still has a sizable code footprint (as evidenced by the `-7,009` lines on this PR!)

### Why?

I'm starting some work on the analysis Results page (aka the place where guide & tabs are shown) and eliminating this feature is going to make it easier. Eliminating it in a standalone PR should make it a bit easier to review. 

**I don't plan to remove Checklist-type things entirely.** I just don't want to have `Checklist` (old, largely unused) and some kind of `ChecklistV2` living in the same codebase for no reason.

### Testing

E2E currently fails because of a [backend 500 error](https://github.com/WoWAnalyzer/server/pull/71) that I'm going to fix in a few minutes (an unexpected undefined from an API). Otherwise, e2e runs against every spec that is marked as supported and confirms that they load. There may be some 4.x (Cata) classic specs that only had Checklists and now show nothing, but since Classic is now on 5.x (MoP) that isn't a big deal IMO.


jazminite (approved):
lgtm for the Classic side

PR #7496, opened 2025-10-25
remove Suggestions API from the codebase

emallson:
### Description

Depends on: https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7494

Since the branch for that isn't merged, the diff isn't directly viewable. You can view the new commits here: https://github.com/emallson/WoWAnalyzer/compare/push-ktvkksxvvzpo...emallson:WoWAnalyzer:remove-suggestions

This is a follow-up to the Checklist removal PR that removes the existing Suggestions API. For specs that use the Guide format, suggestions are entirely unusable in their present form: there is no place in which they appear.

I added some light support in a [recent Foundation PR](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7227), but as I started planning out what I'm gonna be working on for Midnight I decided that I'd rather start with a simpler Suggestions API without the baggage of an API that has been mostly unused for several years.

As a result, this PR cleans up (almost all of) the legacy `suggestions(when: When)` implementations in Analyzers and removes most Suggestions-related components. I haven't fully cleaned up `ParseResults` yet because it reaches into many core places and I figured adding a change that might require some detailed review into another 7000 PR might not be so helpful :upside_down_face: 

### Testing

Much like the Checklist PR, e2e is doing the heavy lifting. This content is viewable almost no-where. If you have cases you're using it that I broke, lmk


ToppleTheNun (approved):

jazminite (approved):
lgtm from the Classic side of things

PR #7501, opened 2025-11-02
[Core] Guide Rework Part 2

Sharrq:
Additional new guide components as part of the larger rework that i have been working on. Alongside this PR, I also split some of the previous components to create some more generalized base components that people can reuse for other purposes if they decide to use them for other purposes or use cases.

Cast Efficiency - I differentiated the layout depending on whether the spell has charges or not. If it does not have charges it will show colored segments for when the ability is on cooldown with white pins for the casts. Gaps are shown as red highlights in the background. For the charged spells, the bullet graph main bar shows the total casts with the target line for the number that was possible. The bar beneath it shows the amount of time that they were capped on charges. I separated SegmentedTimeline and BulletGraph from this to be reused elsewhere. This also shows the difference between the standard layout and the compact layout that consolidates things into a single row.
Screenshot from 2025-11-01 19 59 46
Screenshot from 2025-11-01 20 15 14

Intensity Graph - This separates the fight into blocks of equal time depending on the length of the fight and the number of blocks. In this example, we are taking the ignite damage ticks over the fight and creating a heatmap of where the largest spikes of ignite was. I separated Heatmap from this to be used for other use cases.
Screenshot from 2025-11-01 20 06 55

Intensity Bar - Similar to Intensity Graph, this shows the amount of time ignite was above particular thresholds. StackedBarChart was separated from this to be used elsewhere. 
Screenshot from 2025-11-01 20 07 06

Damage Contribution - This shows the amount of damage that was contributed to some time period. In this case any damage done while Combustion was active. DonutChart was separated from this to be used elsewhere.
Screenshot from 2025-11-01 20 12 05

Also in this PR I created GuideDataWrapper to consolidate the overall style elements that i have been including with every guide component i have been creating, this just helps to keep it consistent across them. I also moved the relevant Divs from GuideDivs to this since the wrapper can be used wherever in place of using all those style divs repeatedly.

I purposely didnt overwrite CastEfficiencyBar because i didnt want to replace all uses of it in this PR, so this does duplicate that component with some changes. Ill do that replacement in a separate PR assuming people like it. Along with this rework i removed some options like the gap highlight mode and just had it show the gaps regardless.


emallson:
i haven't forgotten about this PR, its just going to take a bit to get through. rather than forcing you to go back and split it up, i'm going to basically split it up in review, focusing on one or two components at a time to make it manageable. so if you see a review come in and it only covers a couple things, don't be surprised.

### Code Review Status

- [x] Bullet Graph
- [x] Cast Efficiency Ribbon/Panel
- [x] `DamageContribution`
- [x] Donut Chart
- [ ] Heatmap
- [ ] Intensity Bar / Chart
- [x] Segmented Timeline
- [ ] Stacked Bar

## Non-Code Notes

Besides code stuff, I have a few notes about the design & composition.

### Colors

Up front: colors are fucking hard. I don't have a silver bullet for this, but what I have found works well is using one highly-saturated color to highlight the thing you want, and then low-saturation colors for non-focus, informative elements.

An examples of what I'd say are not great choices:

image

Both colors here are low-saturation and low-brightness, which means the whole thing is kind of visually indistinct. One of the colors should be high-saturation / high-brightness so that it stands out. The current state makes the *pins* the element of focus (because they are bright, though they lack saturation), which is probably not correct.

For an element like this, my preference would be to make the *downtime* segment the high-vis color so that the eye is drawn to the problem points (this is what is done in the downtime timeline, seen below)
image

The other way to do it is done for the *uptime* segment to be bright, as done here:
image

I have my preference, but I'm open to persuasion that we should highlight uptime rather than downtime here.

Another element that I think does this quite well is this chart:
image

The high-DPS points are highly saturated, while the low dps points have low saturation. The whole thing has somewhat inconsistent brightness (low dps values have high brightness, high dps values have low brightness), but that is really annoyingly time-consuming to fix because simply adjusting the S value in HSL is not actually going to produce good-looking results. Overall, it looks quite good so having an inverse relation between saturation and brightness might be good tech for later stuff.

All of that said: *I don't think we need to let people choose colors for this.* Or if we do, it should be from a limited palette (magic schools don't work *great* as illustrated by the mage specs: everything is fire, or everything is frost, or everything is arcane). Individually, choosing colors can look appealing. However, across many repeated elements it becomes visually chaotic. Compare these:

current:
image


Arcane (rescaled using `hsl(from ${ARCANE} h 15% 20%)`)
image


Frost (rescaled using `hsl(from ${FROST} h 15% 20%)`)
image

These colors are not ideal (the Frost and Arcane colors are kinda defined by being bright & saturated) but IMO the gaps are much clearer. I think letting a spec page have a "theme color" to use (so spec pages look like spec pages) is okay, but varying it per-element is probably not what we want.

As an aside: I would like to make our color usage more consistent and readable, and this nudge to not introduce even more new, different colors is a part of that :)

### Role of Charts & Diagrams

I think it is helpful to look at a chart / diagram and ask "why shouldn't this be a table?". Nott literally an HTML `<table>`, but a structured list of numbers like this: 
image

For every one of these diagrams: why should we use it instead of a table? There are two common reasons:

1. There is too much data to put in a table. Timelines are a *great* example of this. They summarize thousands of data points in a compact form. Reading a table with thousands of rows isn't very useful.
2. They visualize the relation between values in the table (including in ways that are helpful for people that aren't numerically inclined)

Both of these do the latter:
image
image

Fundamentally, they're actually the same thing in a different form: they help visualize the ratio between *each pair* of numbers in the table.

Now, these: 
image

Written as a table:

| Spell | Efficiency | Casts | Max Casts | Capped Time |
| --- | --- | --- | --- | --- |
| Phoenix Flames | 99% | 51 | 52 | 3s |
| Fire Blast | 99% | 138 | 140 | 6s |

The table is *smaller* than the charts, so the reason for the diagram can't be "too much data", it has to be comparison.

What do you *want* to compare? Usually either:

- Casts to Max Casts (aka efficiency / downtime) within the same row (aka the same spell), or
- Efficiency across rows (aka different spells)

I'd argue that the structure of these diagrams actually encourages comparing *casts / max casts* with *capped time*, which is wrong! This is due to these bars being placed next to each other, while simultaneously being visually separated from other cast efficiency diagrams.

IMO the tabular form with embedded bars is both more space efficient and clearer.
image

It doesn't have the same eye-catching "what is this", but it is more *useful*. Making things look good is worthwhile, but it shouldn't be at the cost of (much) functionality. 

There's a balance here---once of the big bits of feedback we got from a very loud minority after subcreation migrated to Archon was that Archon was too space-inefficient. People couldn't fit nearly as much on one screen. But that spacing had a purpose: it made Archon look good! We made some compromises (eliminating some padding that cost more than it was worth), but ultimately it is a judgment call. my 2c is that the `BulletGraph` component doesn't pass the bar yet.

### Other Notes

*Intensity Chart*

I think this turned out really well. I have some concerns about comparability across different reports, but within the same report it is awesome. My one nit is that when viewing per target (or any multiple) that these to be laid out in tabular form instead of alternating labels and diagrams:

Here is what the per-target currently looks like
image

Here is some quick css hackery to make them more heatmap-style:

image

(yes, they're no longer square. Not sure they *need* to be square, but also with less-hacky changes (e.g. reducing the bin count slightly) they could be made square)

*Intensity Bar*

It feels weird that the data is basically repeated twice. Once in the "table" data-strip at the top, and once in the diagram itself:
image

*Segmented Timeline*

I like how the pins panned out. They may interfere with zero-padding stacking of these, but having a "start of bar" marker helps solve the problem of "perfectly back-to-back cooldowns looks like one long cooldown"

Sharrq:
Looking through the feedback. I am probably going to make some changes here
1. I agree that having the labels and boxes side by side instead of alternating on the intensity graph is better, ill change it
2. Ill eliminate the redundant information on intensitybar
3.  I do like having the bullet graph for the charged spells, but i agree about the minor bar conflicting, so ill probably move the capped time to a stat box alongside the cast efficiency percent and then just have the one bar/target line.
4. I see the point on the color saturation. I tried a few options originally including one that was a patterned background instead of a solid color (like a gray bar with red diagonals across it), but didnt really like it so i went back to the solid background. Ill mess with the colors and make them uniform throughout.
5. As for tables, I would argue that tables are often boring and require filtering through a bunch of text to find the info you are looking for (especially when they have a lot of data). At a glance, for something like cast efficiency for charged spells, I can easily see "bar has a lot of empty space and thats bad". Drawing attention to the good/bad first and the numbers secondarily.

src/interface/guide/components/DonutChart.tsx
```diff
@@ -0,0 +1,193 @@
...
+ * Base donut chart component that renders a simple donut visualization.
+ * For more complex use cases with spell tracking and legends, use DamageContribution.
+ *
+ * @param segments - Array of segments to display in the chart
+ * @param size - Size of the donut chart in pixels (default: 200)
+ * @param innerRadiusRatio - Inner radius ratio 0-1, where 0.5 = half radius (default: 0.6)
+ * @param showCenterText - Whether to show center text with total (default: true)
+ * @param centerText - Custom center text override
+ */
+export default function DonutChart({
```
emallson: what is the purpose of having a new `DonutChart` (instead of using `parser/ui/DonutChart`)? is it primarily the new aesthetic parts (e.g. hover animation?)

(it *does* look nicer than the old one)

src/interface/guide/components/DamageContribution.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * @param title - Title for the chart (defaults to "Damage Contribution")
+ * @param spells - List of spells to track with their display colors
+ * @param calculateContribution - Function that takes spellId and returns damage/healing amount
+ * @param otherColor - Color for the "Other" category (default: #666666)
+ * @param size - Size of the donut chart in pixels (default: 200)
+ * @param innerRadiusRatio - Inner radius ratio 0-1 (default: 0.6)
+ * @param helperText - Optional helper text to display below the header
+ * @param stats - Optional stat cards to display in the header
+ */
+export default function DamageContribution({
```
emallson, line 40: i don't believe that this component is at the correct level of abstraction. that is making it very narrow & specific.

this component merges bits of analysis (injected via `calculateContribution`) with display (`Legend` layout with `DonutChart`).

what i'd suggest is:

- move the Legend stuff to `DonutChart` (can be optional with a boolean prop)
- maybe add a `total` prop to `DonutChart` to make it easy to add the "other" segment (not convinced we need this, since ultimately it is just `total - sum(value)`)
- remove this component

based on your use of this component for Combustion, I don't think that it makes sense to have a separate component for `DamageContribution`. this component has almost no actual logic, and primarily just calls the user-supplied `calculateContribution` for every element. but: the hard part of using this is not *calling* `calculateContribution`, but *writing* it---and this doesn't help with that.

src/interface/guide/components/DamageContribution.tsx
```diff
@@ -0,0 +1,215 @@
...
+  const specifiedTotal = contributions.reduce((sum, contrib) => sum + contrib.amount, 0);
+
+  // Calculate total from all damage (including unspecified)
+  const overallTotal = calculateContribution(-1); // -1 signals to get total of all
+
+  // Calculate "Other" category
+  const otherAmount = Math.max(0, overallTotal - specifiedTotal);
+
+  // Add "Other" to contributions if it exists
+  const allContributions: Array<SpellContribution & { isOther?: boolean }> = [...contributions];
```
emallson, line 70: this copy is not needed, `contributions` is never used again.

src/interface/guide/components/DamageContribution.tsx
```diff
@@ -0,0 +1,215 @@
...
+      color,
+      amount: calculateContribution(spell.id),
+    }))
+    .filter((contrib) => contrib.amount > 0);
+
+  // Calculate total from all specified spells
+  const specifiedTotal = contributions.reduce((sum, contrib) => sum + contrib.amount, 0);
+
+  // Calculate total from all damage (including unspecified)
+  const overallTotal = calculateContribution(-1); // -1 signals to get total of all
```
emallson, line 64: i mentioned eliminating `calculateContribution` above but if it stays, it shouldn't have this -1 behavior. the user can just pass in the `overallTotal` prop. they would already to have to write a separate branch to handle the -1 case and calculate it, so it isn't saving any work to just pass the value in.

src/analysis/retail/mage/fire/guide/Combustion.tsx
```diff
@@ -253,6 +257,24 @@ class CombustionGuide extends Analyzer {
...
+                { spell: TALENTS.PYROBLAST_TALENT, color: '#ff6600' },
+                { spell: SPELLS.FIRE_BLAST, color: '#ff9933' },
+                { spell: SPELLS.PHOENIX_FLAMES_DAMAGE, color: '#ffcc00' },
+                { spell: SPELLS.IGNITE, color: '#910808ff' },
+                { spell: SPELLS.FLAMESTRIKE, color: '#cc3300' },
+                { spell: SPELLS.ARCANE_PHOENIX_DAMAGE, color: '#7b1d92ff' },
+              ]}
+              calculateContribution={(spellId: number) =>
+                this.combustionDamageTracker.getDamageForSpell(spellId)
+              }
```
emallson: so as an example of the issue with `calculateContribution`: this could be replaced by:

1. calling `[ ... ].map(entry => ({ ...entry, value: getDamageForSpell(entry.spell) }))`
2. passing `totalValue={combustionDamageTracker.totalDamage}`

with virtually no additional work on the consumer side, but without the added logic & complexity in the `DamageContribution` component.

emallson (changes requested):
i am not the biggest fan of donut charts---they are very big, and aside from looking nice they don't *do* much. comparing ratios is easier with a bar chart (or a stacked bar chart, like you have in this PR). I'd pitch deleting the Donut Chart entirely, personally, but not going to force that issue and am fine including it.

However, the `DamageContribution` component doesn't really have a place (IMO). it doesn't do enough as a helper to justify the extra space. The useful bits of it (mostly the legend) should be moved to the base component.

ToppleTheNun:
I haven't gone through this full thing yet but I wanted to echo the donut chart feedback.

Sharrq:
I didn't notice that there was a donut chart already, my mistake. I'll take a look at it and see if it's worth replacing it with mine.

emallson:
I looked a bit as I was reviewing and I think it probably is. I'm not sure
it's API compatible, but it probably could be

Sharrq:
Ok. I havent messed with the Donut Chart yet. But made some other adjustments. 

I reskinned the colors for things and picked out some muted spec colors and went that direction
Screenshot_2025-11-06_18-50-21

I also carried those colors into IntensityChart and IntensityBar
Screenshot_2025-11-06_18-49-20

I cleaned up IntensityBar a little bit and added a legend to it instead of the stat boxes because i didnt like that the damage thresholds were buried in the tooltip
Screenshot_2025-11-06_18-49-35

And I reworked the CastEfficiencyRibbon to not show a secondary bar on the bullet graph and to move the capped time into a stat box
Screenshot_2025-11-06_18-48-55

Also, I did not change the per target section of IntensityChart largely because It tends to make the boxes on the heatmap incredibly small, especially if the targets have long names or if it is constrained to a smaller container like if it is used alongside one of the explanations. I could have reduced the number of blocks to compensate I suppose, but it started to get really annoying to work with ... like the text of the names needed to be large enough, but the bigger you make the text the less room the heatmap boxes have .... so it seemed better to keep it the way it is.

src/parser/ui/DonutChart.tsx
```diff
@@ -20,105 +16,158 @@ export interface Item {
...
-            { color, label, tooltip, value, valuePercent = true, spellId, valueTooltip, itemLevel },
-            index,
-          ) => {
-            label = tooltip ? <TooltipElement content={tooltip}>{label}</TooltipElement> : label;
-            label = spellId ? (
+/**
+ * Legacy DonutChart component for backward compatibility.
+ * This renders a simple donut chart with a legend.
+ */
+export default function DonutChart({ items }: Props) {
```
emallson: image
this is overflowing stat boxes.

the merge of `midnight` is disabling all specs, which makes this annoying to check. but Outlaw has some donut charts that you can check against.

this also shouldn't be duplicated. from what I can see, this just copies *most of* the contents of the Donut Chart component you added into the `parser/ui/DonutChart` component. there should be one that actually does the rendering, and then the other (probably this one) renders it with whatever minimal code is needed to change one `Props` type to the other.

Sharrq: I would actually ignore donut chart for now. I didn't mean to commit that. I was playing with it a little bit and ran into the same things you are pointing out and decided to circle back to it. I thought I reverted it before I committed but apparently I didn't

Sharrq: Sorry about that, wasn't paying attention

src/interface/guide/components/BulletGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+  width: 100%;
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+`;
+
+const GraphBackground = styled.div`
+  width: 100%;
+  height: 32px;
```
emallson, line 108: avoid using fixed heights for things like this. it means that the resizing the parent element doesn't actually re-size the contents and results in visual breakage.

src/interface/guide/components/BulletGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+const PerformanceZone = styled.div<{ width: number; color: string }>`
+  width: ${(props) => props.width}%;
+  height: 100%;
+  position: relative;
+`;
+
+const MainBar = styled.div<{ width: number; color: string }>`
+  position: absolute;
+  left: 0;
+  top: 0;
```
emallson, line 127: conventionally, the background would be `absolute` and the content would be laid out. that isn't *always* the case because sometimes you need to do it the other way, but in this case I think it makes the most sense.

of note: using `position: absolute` like this prevents (or at least makes difficult) doing the center-alignment for a bullet graph like shown on wikipedia: 
image

(you can center in absolute with some `calc` magic but it is harder than `display: flex; align-items: center;`)

src/interface/guide/components/BulletGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+  position: relative;
+  box-shadow: inset 0 1px 3px rgba(0, 0, 0, 0.4);
+  border: 1px solid rgba(255, 255, 255, 0.05);
+`;
+
+const PerformanceZone = styled.div<{ width: number; color: string }>`
+  width: ${(props) => props.width}%;
+  height: 100%;
+  position: relative;
+`;
```
emallson, line 122: this doesn't set the `background-color` attribute using `color`, so the `PerformanceZone`s don't actually work.

```suggestion
const PerformanceZone = styled.div<{ width: number; color: string }>`
  width: ${(props) => props.width}%;
  background-color: ${(props) => props.color};
  height: 100%;
  position: relative;
`;
```

src/parser/ui/DonutChart.tsx
emallson: no worries!

emallson (changes requested):
## `BulletGraph`
image

The cast count bars here are `CastEfficiencyRibbon`s, which in turn render `BulletGraph`s. I think [bullet graphs](https://en.wikipedia.org/wiki/Bullet_graph) are an interesting potential visualization. We do a *lot* of "show point value X, vs (often hidden) target values A, B, C...", which bullet graphs cover.

However, I think this implementation needs work---both on the code side and the design side.

There are two distinguishing features for Bullet Graphs:

- The use of the background to communicate a (not necessarily linear) scale, like our Bad / Ok / Good values. (implemented as `PerformanceZone`)
- The presence of a target "tick" showing the goal, which the progress bar may exceed. (implemented as `TargetMarker`)

The implementation as it is in this PR does not have functional `PerformanceZone`s, but this is easy to fix so it isn't a big deal. However, after fixing them they aren't really...visible?

The colors used in `CastEfficiencyRibbon` are *very* muted (opacity of ~15% against a black-adjacent background), and the bar itself completely covers the backgrounds. This leads to a situation where *if* you can see the backgrounds, it looks more like a `StackedBar`.

A little bit more CSS allows insetting the bar, but tbh I wasn't able to find a combination of colors that I think worked well. That isn't to say we *can't*, but, well, colors are hard and using the bright primary red/yellow/green doesn't make it easier:
image

image

image

My 2c would be to cut `BulletGraph` from this PR. It wasn't really being used in the elements you presented (those were really just bars with a max-value line---which is still useful, but also much simpler and could be less code). I think they might be interesting to investigate further, but need some focused work to make good (and maybe some changes to our core color palette 🙃 )

if you are interested, here are the css changes I made for the first "fixed" screenshot:

```diff
diff --git a/src/interface/guide/components/BulletGraph.tsx b/src/interface/guide/components/BulletGraph.tsx
index 962ac69e28..04c0872ec2 100644
--- a/src/interface/guide/components/BulletGraph.tsx
+++ b/src/interface/guide/components/BulletGraph.tsx
@@ -104,35 +104,40 @@
 `;
 
 const GraphBackground = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+
   width: 100%;
-  height: 32px;
   background: rgba(0, 0, 0, 0.4);
   border-radius: 4px;
   overflow: hidden;
   display: flex;
-  position: relative;
   box-shadow: inset 0 1px 3px rgba(0, 0, 0, 0.4);
-  border: 1px solid rgba(255, 255, 255, 0.05);
+  z-index: 0;
+  box-sizing: content-box;
 `;
 
 const PerformanceZone = styled.div<{ width: number; color: string }>`
+  background-color: ${(props) => props.color};
   width: ${(props) => props.width}%;
   height: 100%;
   position: relative;
 `;
 
 const MainBar = styled.div<{ width: number; color: string }>`
-  position: absolute;
-  left: 0;
-  top: 0;
   width: ${(props) => props.width}%;
-  height: 32px;
   background: ${(props) => props.color};
   display: flex;
   align-items: center;
   padding-left: 8px;
   transition: width 0.3s ease;
-  border-radius: 4px 0 0 4px;
+  margin-top: 1rem;
+  margin-bottom: 1rem;
+  z-index: 1;
+  box-shadow: 0 0 3px black;
 `;
 
 const BarLabel = styled.span`
@@ -149,7 +154,7 @@
   position: absolute;
   left: 100%;
   top: 0;
-  height: 32px;
+  bottom: 0;
   display: flex;
   align-items: center;
   transform: translateX(-50%);
```

src/interface/guide/components/CastEfficiencyRibbon.tsx
```diff
@@ -0,0 +1,451 @@
...
+  width: 100%;
+  height: 40px;
+  display: flex;
+  align-items: center;
+`;
+
+// Helper functions to convert combat log events to timeline segments/markers
+function createCooldownSegments(
+  spellId: number,
+  events: ReturnType<typeof useEvents>,
```
emallson, line 35: nit: this is clearer as `AnyEvent[]`
```suggestion
  events: AnyEvent[],
```

src/interface/guide/components/SegmentedTimeline.tsx
```diff
@@ -0,0 +1,192 @@
...
+ * - Phase indicators
+ *
+ * @param windows - Optional time windows to display (defaults to full fight)
+ * @param segments - Colored segments to render on the timeline
+ * @param markers - Optional markers (pins) to show on the timeline
+ */
+export default function SegmentedTimeline({
+  windows,
+  segments,
+  markers = EMPTY_MARKERS,
```
emallson, line 60: nit: this can just be `[]`
```suggestion
  markers = [],
```

src/interface/guide/components/SegmentedTimeline.tsx
```diff
@@ -0,0 +1,192 @@
...
+  timestamp: number;
+  /** Tooltip label for the marker */
+  label: string;
+  /** Color of the marker (default: white) */
+  color?: string;
+}
+
+export interface SegmentedTimelineProps {
+  /** Optional windows to show on the timeline (defaults to full fight if not specified) */
+  windows?: TimeWindow[];
```
emallson, line 34: multiple windows makes this component a lot more complex to get right. however, 90% of that complexity comes from the possibility of overlapping windows.

consider adding a dev-mode error check like:

```tsx
if (import.meta.env.DEV) {
  windows.sort((a, b) => a.startTime - b.startTime);
  windows.forEach((window, index) => {
    const nextWindow = windows[index + 1];
    if (nextWindow && nextWindow.startTime < window.endTime) {
      throw new Error("must not provide overlapping windows to SegmentedTimeline");
    }
  });
}
```

alternately, we could just document that in the docblock and if people do it it will look broken :)

src/interface/guide/components/CastEfficiencyRibbon.tsx
```diff
@@ -0,0 +1,451 @@
...
+
+      if (wasCapped && windowEnd > segmentStart) {
+        wastedTime += windowEnd - segmentStart;
+      }
+    });
+
+    return wastedTime;
+  };
+
+  const wastedTime = calculateWastedTime();
```
emallson, line 376: this needs to be memoized with `useMemo` or every re-render is going to reprocess all events for every `CastEfficiencyRibbon` rendered.

```suggestion
  const wastedTime = useMemo(calculateWastedTime(), [eventHistory, hasCharges, maxCharges]);
```

the function is pretty complicated, so you might want to move it out of the component and call it with proper parameters instead like:

```suggestion
  const wastedTime = useMemo(() => calculateWastedTime(eventHistory, ability), [ability, eventHistory]);
```

emallson (commented):
### `CastEfficiencyRibbon` / `CastEfficiencyPanel` / `SegmentedTimeline`

aside from the issues i raised with the `BulletGraph` last time (which would simplify the charges version of the `CastEfficiencyRibbon` component, these look good

src/interface/guide/components/IntensityBar.tsx
```diff
@@ -0,0 +1,345 @@
...
+      s: Math.round(s * 100),
+      l: Math.round(l * 100),
+    };
+  }
+
+  // Default fallback to fire orange
+  return { h: 35, s: 90, l: 55 };
+}
+
+function generateGradient(baseColor: string): string[] {
```
emallson, line 67: not getting through a full heatmap review right now, but you can replace the color parsing and gradient construction code with css relative colors like:

```ts
function generateGradient(baseColor: string): string[] {
  return [
    // Lightest - tier 1
    `hsl(from ${baseColor} h calc(s - 10) calc(l + 20))`,
    // Lighter - tier 2
    `hsl(from ${baseColor} h calc(s - 5) calc(l + 10))`,
    // Base - tier 3
    baseColor,
    // Darker - tier 4
    `hsl(from ${baseColor} calc(h - 10) s calc(l - 10))`,
    // Darkest - tier 5
    `hsl(from ${baseColor} calc(h - 20) s calc(l - 20))`,
  ];
}
```
this will allow using colors that aren't written in hsl format (the browser will convert automatically)

src/interface/guide/components/IntensityBar.tsx
```diff
@@ -0,0 +1,345 @@
...
+        }
+      });
+    });
+
+    const dps = (totalAmount / sampleInterval) * 1000;
+    if (dps > 0) {
+      samples.push(dps);
+    }
+  }
+
```
emallson, line 170: this is very very quadratic complexity. there are some other places with quadratic complexity in this pr but they mostly involve small numbers so it is fine. this is `<target count> * <event count> * <fight duration in seconds>`.

unfortunately we don't have good primitives for making slicing like this easy, and it needs to be manually unrolled into a loop. ideally it'd be single-pass, but the caller is providing data broken out by target already so it can be a single pass per target like:

```ts
for (const target of targets) {
  const buckets = new Map();
  for (const event of target.events) {
    const index = Math.floor((event.timestamp - fightStart) / sampleInterval);
    const currentValue = buckets.get(index) ?? 0;
    buckets.set(index, currentValue + event.amount);
  }

  samples.push(...buckets.values());
}
```

src/interface/guide/components/IntensityBar.tsx
```diff
@@ -0,0 +1,345 @@
...
+  if (samples.length === 0) {
+    return null;
+  }
+
+  // Calculate median for threshold centering
+  const sortedSamples = [...samples].sort((a, b) => a - b);
```
emallson, line 176: don't copy samples here
```suggestion
  const sortedSamples = samples.sort((a, b) => a - b);
```

src/interface/guide/components/Heatmap.tsx
```diff
@@ -0,0 +1,205 @@
...
+const CellsContainer = styled.div<{ $gap: number }>`
+  display: flex;
+  gap: ${({ $gap }) => $gap}px;
+`;
+
+const Cell = styled.div<{ $color: string; $size: number }>`
+  width: ${({ $size }) => $size}px;
+  height: ${({ $size }) => $size}px;
+  flex-shrink: 0;
+  background-color: ${({ $color }) => $color};
```
emallson, line 195: using `$prop` like this doesn't seem to be [officially supported](https://github.com/emotion-js/emotion/issues/2193) but I don't see the junk props on elements in the inspector so...i guess its fine?

src/interface/guide/components/IntensityBar.tsx
```diff
@@ -0,0 +1,345 @@
...
+    // Darkest - tier 5
+    `hsl(${Math.max(0, h - 20)}, ${s}%, ${Math.max(0, l - 20)}%)`,
+  ];
+}
+
+function roundThreshold(value: number): number {
+  if (value > 100000) return Math.round(value / 10000) * 10000;
+  if (value > 50000) return Math.round(value / 5000) * 5000;
+  return Math.round(value / 1000) * 1000;
+}
```
emallson, line 88: i guess its intended that each of these is bucketed into values 1/10 of the threshold? (so 100k gets 10k, 50k gets 5k)?

src/interface/guide/components/IntensityBar.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const step = roundThreshold(median * 0.4);
+
+  // Create 5 intensity tiers centered around median
+  const thresholds = [
+    { min: 0, max: Math.max(0, medianRounded - step), label: 'Very Low' },
+    { min: Math.max(0, medianRounded - step), max: medianRounded, label: 'Low' },
+    { min: medianRounded, max: medianRounded + step, label: 'Medium' },
+    { min: medianRounded + step, max: medianRounded + step * 2, label: 'High' },
+    { min: medianRounded + step * 2, max: Infinity, label: 'Very High' },
+  ];
```
emallson, line 189: the way these are auto-scaled is reasonable as an informative element. however, these aren't comparable across fights. they don't necessarily *need* to be, but conventionally our stuff *is*.

there isn't really a way to make it comparable across reports without picking a somewhat-arbitrary center point for the scale, unfortunately. so this is probably okay, but we'll see what people's feedback is.

src/interface/guide/components/IntensityChart.tsx
```diff
@@ -0,0 +1,353 @@
...
+    // Darkest - tier 5
+    `hsl(${Math.max(0, h - 20)}, ${s}%, ${Math.max(0, l - 20)}%)`,
+  ];
+}
+
+function roundThreshold(value: number): number {
+  if (value > 100000) return Math.round(value / 10000) * 10000;
+  if (value > 50000) return Math.round(value / 5000) * 5000;
+  return Math.round(value / 1000) * 1000;
+}
```
emallson, line 91: this is like, entirely duplicated from `Heatmap` with minor (incompatible!) changes. it should not be.

src/interface/guide/components/IntensityChart.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+        // Build data for each unique target name
+        return Object.entries(grouped).map(([name, targets]) => buildTargetData(targets, name));
+      })()
+    : [buildTargetData(data, 'Overall')];
+
+  // Calculate stats
+  const allBlocks = heatmapData.flatMap((t) => t.blocks);
+  const nonZero = allBlocks.filter((v) => v > 0);
+  const sorted = [...nonZero].sort((a, b) => a - b);
```
emallson, line 204: another needless copy.

src/interface/guide/components/IntensityChart.tsx
```diff
@@ -0,0 +1,353 @@
...
+            }
+            acc[target.targetName].push(target);
+            return acc;
+          },
+          {} as Record<string, TargetData[]>,
+        );
+
+        // Build data for each unique target name
+        return Object.entries(grouped).map(([name, targets]) => buildTargetData(targets, name));
+      })()
```
emallson, line 198: use an if-else instead of a ternary + IIFE

src/interface/guide/components/IntensityChart.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const median = sorted[Math.floor(sorted.length / 2)] || maxValue / 2;
+  const medianRounded = roundThreshold(median);
+  const step = roundThreshold(median * 0.4);
+  const thresholds = [
+    0,
+    medianRounded - step,
+    medianRounded,
+    medianRounded + step,
+    medianRounded + step * 2,
+  ];
```
emallson, line 223: this logic is also duplicated from `IntensityBar`

src/interface/guide/components/StackedBar.tsx
```diff
@@ -0,0 +1,178 @@
...
+  );
+
+  const defaultTooltipFormatter = (segment: StackedBarSegment, percent: number) => (
+    <>
+      <strong>{segment.label}</strong>
+      <br />
+      Value: {segment.value.toFixed(0)}
+      <br />
+      Percentage: {percent.toFixed(1)}%
+    </>
```
emallson, line 83: use `<div>` not `<br/>`

src/interface/guide/components/StackedBar.tsx
```diff
@@ -0,0 +1,178 @@
...
+  /** Whether to show percentage labels on segments. Default: true */
+  showLabels?: boolean;
+  /** Minimum percentage required to show a segment. Default: 0.5 */
+  minSegmentPercent?: number;
+  /** Minimum percentage required to show labels on a segment. Default: 5 */
+  minLabelPercent?: number;
+  /** Custom label formatter. Receives segment and percentage, returns label content */
+  labelFormat?: (segment: StackedBarSegment, percent: number) => React.ReactNode;
+  /** Custom tooltip formatter. Receives segment and percentage, returns tooltip content */
+  tooltipFormat?: (segment: StackedBarSegment, percent: number) => React.ReactNode;
```
emallson, line 36: pick one method to do labels / tooltips. either they should be supplied on `Segment`s (like you're doing currently), or they should be `format` props. Not both.

my 2c is to remove these and keep the segment props.

emallson (changes requested):
sorry for taking so long, final round of comments for the initial PR contents. Heatmap/Intensity bar stuff looks good visually, but has some duplication issues internally. these things are pretty tied together, so it might make sense for them to be in the same file and sharing the same internals.

Alternately, make the stuff that is duplicated in `IntensityChart` exported from `Heatmap` / `IntensityBar` so avoid duplication.

Sharrq:
Moved this back to draft for now. Going to work through the feedback, but wont be able to test those changes until i get the mage specs re-enabled in Midnight which i dont want to do in this PR.

Sharrq:
I am going to close this for now. I have been periodically making improvements to it, but im going to wait until i have arcane and fire updated and then do some of this in smaller pieces.

PR #7502, opened 2025-11-03
New results page header

emallson:
### Description

This PR replaces the existing results page header with a newer, more compact option. In the process, I simplified logic around phase selection / dungeon pull selection that predates using the WCL properties to drive these features.

The motivation for this was primarily the *vertical space* used up by the existing header, and secondarily by some additional features that I want to build that (I think) will hook into the header. for comparison, i posted [this screenshot](https://media.discordapp.net/attachments/376695647891488768/1432124691719520326/image.png?ex=690a758e&is=6909240e&hm=d22ffde2da02519c16b922762a953fa8bb579a77f39c77668d128765d25b65c0&=&format=webp&quality=lossless&width=1216&height=721) in discord of the view in production on my laptop.

I took elements of how WCL's design system is structured for Archon work, which makes it a little simpler to build new elements that look like they belong as part of the site. These will be a little out of place at the moment but as I work things should become more...standard (famous last words, etc etc).

This is deploying to https://next.wowanalyzer.com

### Screenshots!

*Large Desktop*
image

(at this screen size, the ad doesn't reduce content width whatsoever, so it is a strict upgrade over the current state of affairs)

*Small Desktop / Laptop*

(showing with and without premium)
image
image
image

I don't love the columns wrapping, but no matter what I did there were always cases where wrapping occurred. Some specs have more or fewer tabs, and affect where the "tab list wraps" breakpoint is hit. I think it looks *fine* when wrapped, even if not ideal.

I also didn't spend too much time on this particular piece because I am not sure where tabs end up as we centralize things onto the guide (with the checklist being removed). Tabs may end up rendered inline, with a Table of Contents (sidebar or menu, depending on screen size). But that is a future problem, and this is still a big improvement in the amount of info within the initial view *even with the side ad.*

I am a little nervous about the reception to the side ad because it goes from something you can simply scroll past to something that is now always present and always consuming space. But :shrug: guess we'll see

*Tablet*
image

(note: the ad is just disabled at these screen sizes)

*Mobile*
image
(fixing the guide layout on mobile was *not* a priority lmao)

*Filter Menu*
image
image
image


emallson:
(of course within minutes of posting i find more issues, but at least they're small ones)

image

need to remove left/right padding from timeline here

image

and fix some stuff with the new ad container

EDIT:

image

another one. this is filtered to 02:00 to 09:01. a few places referencing the wrong start time :skull:

src/interface/report/Results/Header/FilterButton.tsx
```diff
@@ -0,0 +1,318 @@
...
+import { createPortal } from 'react-dom';
+import TimeFilter from '../TimeFilter';
+import Fight from 'parser/core/Fight';
+import { SELECTION_ALL_PHASES, SELECTION_CUSTOM_PHASE } from 'interface/report/hooks/usePhases';
+import { Filter } from 'interface/report/hooks/useTimeEventFilter';
+import { useFight } from 'interface/report/context/FightContext';
+import Select from 'interface/controls/Select';
+import useClickOutsideHandler from 'interface/hooks/useClickOutsideHandler';
+
+const Btn = styled.button`
```
emallson, line 16: while i put the `Select` element in a re-usable place (`interface/controls/Select`), i haven't done it with the Button element yet. on my TODOs

src/interface/timeline-diagram/TimelineDiagram.tsx
```diff
@@ -346,31 +347,45 @@ PhaseHeader.HEIGHT = 30;
...
+  while (currentTimestamp < info.fightEnd) {
+    timestamps.push({
+      timestamp: currentTimestamp,
+      label: formatDurationMinSec(
+        (currentTimestamp - info.originalFightStart) / 1000,
+        timestampInterval === 60_000,
+      ),
+    });
+    currentTimestamp += timestampInterval;
+  }
```
emallson, line 369: i should probably have moved this change to a separate PR because it isn't *really* related to the header, except that I found it while testing the behavior of the dungeon pull selector

emallson:
I merged midnight in, which disables all retail specs. you can still test with classic specs. i'm probably going to merge tomorrow unless there are comments saying to wait. (design level / "this looks bad" comments are also fine! don't feel like you need to code review this (large) pr to give any feedback)

PR #7513, opened 2025-11-11
Feat/update sigil properties + CoB draft tww

Quaarkz:
### Description

Updated SoS and SoC to TWW numbers.

Started the updating of CoB as now only SoF and SoD reduces cooldowns, for that i tried to add Sigil of Doom from zero imitating previous implementations.

They have to be filtered, so i pass the guid in the  reduceSigilCooldowns function call, then fixate their ids to compare with the ones from the function parameter so the cdr only occurs when any of them match, what its happening right now is that Sigil of Doom never matches, SoF works as before.

Sigil of Doom is excluded on purpose from benefiting from CDR as it shares CD with SoF.

### Testing
Should be possible with any log with Cycle of Binding and better if its FS to see SoD behaviour

- Test report URL: `/report/Anf1XDcjJ2payFBr/1-Mythic++Eco-Dome+Al'dani+-+Kill+(17:59)/Quaarkz/standard/statistics`
- Screenshot(s):
imagen_2025-11-11_134247608

imagen_2025-11-11_134154621


Quaarkz:
You can test in any CoB log, it misses a few casts tho, the CDR in wowanalyzer represents 58 casts while in the log there are 65, i dont think that is new and i have no idea about what to do about it
/report/Anf1XDcjJ2payFBr/1-Mythic++Eco-Dome+Al'dani+-+Kill+(17:59)/Quaarkz/standard/statistics

ToppleTheNun (approved):

PR #7515, opened 2025-11-15
Enable Brewmaster in Midnight

emallson:
### Description

This enables Brewmaster in Midnight, as well as enabling spell list generation for Retail specs. In addition, it adds support for processing spell modifiers for cooldowns, gcds, and charges.

The change to `wow-dbc` to support this can be seen [here](https://github.com/RPGLogs/wow-dbc/commit/37d21d9bf878abf2fbe27d21d5ce650039c5c06b#diff-22b30e7434e26076ab6c5b5ff28c28b3f1c8d5859c9f17bfc245c7411c592f12).

Because the `TALENTS` file is shared with other monk specs, I have left it in place (which also means that I haven't removed all references to it from Brewmaster modules)

I did a couple of updates (removing Weapons of Order & related talents, removing Diffuse/Dampen) but mostly just got it compiling.

### Testing

There aren't real brewmaster logs yet (that I could find), so I did very minimal testing on somone that pulled the Normal Belo'ren boss: https://www.warcraftlogs.com/reports/Vw2r3zNJqPvpRAcj/#fight=6&source=1
image


emallson:
the spell file update for destro does actually break it. i don't think it is any of the new changes

emallson:
image

ok, fixed that. the issue was (1) we didn't have [this change](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7505/commits/2e4de42094c2a4245a29c8a7a829ecd56c146d9c#diff-e9c2e8419c822157ccbd34ee69402840b891b2569f5b660c813fa404208a1f74) from the TWW branch, and (2) the change didn't distinguish between spells whose effects are passive (aka "always on") and those that weren't. [Fire and Brimstone](https://www.wowhead.com/mop-classic/spell=108683/fire-and-brimstone) overrides a TON of buttons while active

(i did also check and confirm that the abilities table looks like same for Demo lock, which needed the fix in the first place)

PR #7523, opened 2025-11-18
update midnight to React 19

ToppleTheNun:
### Testing

- Screenshot(s):


emallson:
i see this is a draft. is it good to put on `next`? (i guess i can wait for tests to pass :sweat_smile:)

ToppleTheNun:
should be good to do so - there were a lot of lint issues that I'm going to fix, I just wanted to get the bones up off my PC for looking at

emallson:
note before i forget: i believe `next` has a few commits that aren't on this branch. i can sort out merging later. i am gonna try to do more testing (maybe getting e2e tests back on?) so we can move this forward and not get it stuck like the react 18 pr was

PR #7543, opened 2025-11-25
Created FILO talent

Quaarkz:
### Description

I had to change First In, Last Out talent ID because it was the ability id, not the shield id so as i copied Soul Barrier implementation it wasn't detecting the onApplyBuff and the other ones.

### Testing

Check with any specialization and FILO talented, it will appear in the statistics tab
imagen_2025-11-25_012900242

- Screenshot(s):

### DIFF
--- a/src/common/TALENTS/demonhunter.ts
+++ b/src/common/TALENTS/demonhunter.ts
@@ -715,7 +715,7 @@ const talents = {
     definitionIds: [{ id: 117839, specId: 577 }],
   },
   FIRST_IN_LAST_OUT_TALENT: {
-    id: 1266497,
+    id: 1266619,
     name: 'First In, Last Out',
     icon: 'ability_demonhunter_demonictrample',
     maxRanks: 1,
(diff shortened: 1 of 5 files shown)


src/common/TALENTS/demonhunter.ts
```diff
@@ -715,7 +715,7 @@ const talents = {
     definitionIds: [{ id: 117839, specId: 577 }],
   },
   FIRST_IN_LAST_OUT_TALENT: {
-    id: 1266497,
+    id: 1266619,
```
emallson: do not change talent ids here. if the talent id is not the id of the ability that is actually cast, add it to `SPELLS` instead. this is a generated file and will be overwritten, silently breaking your change

Quaarkz: Fantastic, i will do it that way, thank you!

Quaarkz: I implemented it the way you told me and as the Talents class would be automatically overwritten i changed it back to test the changes

PR #7548, opened 2025-11-30
refactor: convert DonutChart and Requirement to functional components with proper memoization

Soto-J:
- DonutChart: Add useMemo hooks for total and spec calculations to prevent unnecessary re-renders
- DonutChart: Replace static defaultProps with ES6 default parameters
- Requirement: Convert from class component to functional component
- Requirement: Replace Threshold<any> with RequirementThresholds type for better type safety
- Requirement: Simplify performance calculation by removing getter method


src/parser/ui/DonutChart.tsx
```diff
@@ -69,56 +95,24 @@ class DonutChart extends PureComponent<Props> {
         )}
       </div>
     );
-  }
-  renderChart(items: Item[], chartSize: number, innerRadiusFactor: number) {
-    const innerRadius = chartSize * innerRadiusFactor;
+  };
+
+  const renderChart = () => {
```
emallson, line 100: now that this is a functional component, these calls are actually no longer needed. you can remove the inner functions and inline all of the JSX into a single `return` statement.

src/parser/ui/DonutChart.tsx
```diff
@@ -69,56 +95,24 @@ class DonutChart extends PureComponent<Props> {
...
+    const data = { items };
```
emallson, line 101: this (the `data = { items }`) should also be memoized. i know it wasn't before either but it should've been and we should fix it while we're touching this code.

emallson (changes requested):
this is much better :+1: couple of comments, but nothing big

as a meta thing: future PRs of this type should target the `midnight` branch. changes to `the-war-within` at this point will mostly be classic support & retail bugfixes. depending on how things progress, i may backport core changes beforehand but there are a lot of breaking changes on `midnight` atm.

that is particularly relevant for this PR because the `Requirement` component was actually removed in the `midnight` branch as part of the removal of the Checklist feature.

i squashed/rebased the changes onto the `midnight` branch and the patch to `DonutChart` applies cleanly, at least: https://github.com/WoWAnalyzer/WoWAnalyzer/compare/midnight...emallson:WoWAnalyzer:class-refactor-midnight-donut-chart. if you want to start from this commit on midnight, you're welcome to. otherwise i'll probably do the rebase manually once you get an approval and apply the actual patch on `midnight` only.

Soto-J:
Thanks for the clarification and guidance! 🙏
I went ahead and created a class-refactor-midnight branch and migrated the relevant commits onto midnight (excluding the removed Requirement component, per your note).

Apologies for the confusion earlier, and thank you for your patience and the detailed explanation. It really helped me understand the project workflow better. If anything looks off or can be improved, I’m happy to make adjustments.

emallson:
closed in favor of https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7549

PR #7555, opened 2025-12-12
[Mistweaver] Update reworked Tear of Morning

lolswirl:
### Description

modified invigorating mists to swap to sheilun's gift, if talented
added `during celestial` attribution to easier identify envs within that window - logic applied onto those envs to calc on the extra duration of direct and amped healing

note: cleave healing is agnostic of the actual cleaved % since the event is direct instead of ticking from the hot so values will be the same on live vs midnight

### Testing

- Test report URL: 
  - `/report/rGqTxmtvPjW1pknQ/13-Mythic++Magisters'+Terrace+-+Wipe+1+(22:21)/1-Swirlanir/standard`
  - `/report/RJ6wpn8CAMNhqBjY/25-Normal+Lightblinded+Vanguard+-+Kill+(8:51)/Jeypii/standard/statistics`

Live:
image
image

Post change:
SG/Chi-Ji
image
image

Vivify/Yu'lon
image

### DIFF
--- a/src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
+++ b/src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
@@ -1,7 +1,13 @@
 import talents, { TALENTS_MONK } from 'common/TALENTS/monk';
 import spells from 'common/SPELLS/monk';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { CastEvent, GetRelatedEvents, HealEvent } from 'parser/core/Events';
+import Events, {
+  CastEvent,
+  GetRelatedEvents,
+  HealEvent,
+  ApplyBuffEvent,
+  RemoveBuffEvent,
+} from 'parser/core/Events';
 import { calculateEffectiveHealing } from 'parser/core/EventCalculateLib';
 import Statistic from 'parser/ui/Statistic';
 import ItemHealingDone from 'parser/ui/ItemHealingDone';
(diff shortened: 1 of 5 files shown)


src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
```diff
@@ -55,10 +111,84 @@ class TearOfMorning extends Analyzer {
...
+      (this.envelopingMistHealing +
+        this.tftCleaveHealing +
+        this.envExtraDurationHealing +
+        this.envExtraDurationBonusHealing) /
+        this.envelopCasts
     );
   }
 
+  onEnvApply(event: ApplyBuffEvent) {
+    if (this.celestialHooks.celestialActive) {
```
vohrr: This map is essentially duplicating what the hotattributor/tracker already manage. Imo would be better to implement the the env applications into the HotAttributor.ts. From here you can just grab the hot from the tracker and filter via `.fromTom` etc. and do your duration calcs based off of the tracker properties. 
I would recommend that since we will want to update the attributor either way to accommodate this talent rework and will simplify this file a lot

src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
```diff
@@ -43,10 +85,24 @@ class TearOfMorning extends Analyzer {
       Events.cast.by(SELECTED_PLAYER).spell(talents.ENVELOPING_MIST_TALENT),
       this.onCast,
     );
+    this.addEventListener(
```
vohrr: Need a listener for refresh events as well

vohrr (changes requested):

src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
lolswirl: added to attributor and replaced relevent fns

src/analysis/retail/monk/mistweaver/modules/spells/TearOfMorning.tsx
lolswirl: no longer needed with attributor

vohrr (approved):
Looks great!

PR #7556, opened 2025-12-15
Update package manager from pnpm to bun

akemmanuel:
This PR updates the package manager from pnpm to bun, which offers several benefits:

🚀 **Performance Improvements:**
- Up to 2x faster dependency installation
- Significantly faster package resolution
- Reduced disk space usage

🛠️ **Developer Experience:**
- Single binary installation (no separate CLI)
- Built-in test runner and bundler
- Better TypeScript support out of the box

📦 **Migration Benefits:**
- Compatible with existing package.json files
- Same command structure (bun install, bun run build, etc.)
- Zero configuration needed

The changes replace all pnpm references with bun equivalents in the README.md file while maintaining the same functionality. This migration will improve development speed and reduce setup complexity for new contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,19 +19,19 @@ First make sure you have the following:
 - [git](https://git-scm.com/)
   - Optional: Get a UI such [GitHub Desktop](https://github.com/apps/desktop) [TortoiseGit](https://tortoisegit.org/)
 - [Node.js](https://nodejs.org/). We recommend the _LTS_ version.
-- [pnpm](https://pnpm.io/)
+- [bun](https://bun.sh/)
 
 Now you need to pull a copy of the codebase onto your computer. Make a fork of the repo by clicking the **Fork** button at the top of this page. Next, click the green button **Clone or download** and copy your _Clone with HTTPS_ URL, and then run the command `git clone <paste link>`. This will take a minute.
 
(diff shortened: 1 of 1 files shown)


ToppleTheNun:
Just changing the references in the README.md isn't a sufficient migration. There's CI to consider and scripts in the package.json.

Furthermore, this is a change that affects every contributor to WoWAnalyzer and would need extensive discussion.

Pewtro:
This PR only updates the readme - it does not actually swap to bun. 

> Reduced disk space usage

I don't think this is true. pnpm saves each package once globally and symlinks to it. 

> Single binary installation (no separate CLI)

What is different versus pnpm here? 

> Better TypeScript support out of the box

Where does it give better typescript support?

> 📦 Migration Benefits:
Compatible with existing package.json files
Same command structure (bun install, bun run build, etc.)
Zero configuration needed

This is definitely NOT true - it is not compatible.

emallson:
We discussed using `bun` back when we made the switch off of `yarn`. At the time, `bun` lacked good windows support which was a major blocker. These days, we don't have any issues with `pnpm` that justify the churn that switching to `bun` would cause so this is a "no" right now.

you are free to use `bun` yourself locally, but the project isn't switching atm

PR #7563, opened 2025-12-22
Adding copilot instructions

Seriousnes:
### Description

Adding comprehensive copilot instructions for use with coding agents


.github/instructions/analyzer-patterns.md
```diff
@@ -0,0 +1,307 @@
...
+    abilities: Abilities,
+    enemies: Enemies,
+    spellUsable: SpellUsable,
+    maelstromTracker: MaelstromTracker,
+  };
+
+  protected abilities!: Abilities;
+  protected enemies!: Enemies;
+  protected spellUsable!: SpellUsable;
+  protected maelstromTracker!: MaelstromTracker;
```
emallson: may want to switch this to the `.withDependencies` helper (which fixes some gotchas with `dependencies` like them not being present during the constructor (see the notes in `Module.applyDependencies`) or forgetting to spread parent dependencies). not sure if the LLM will be able to understand the usage of it, but it probably has enough examples.

(if you change this, there are 2 places above with `static dependencies` usage. `.withDependencies` usage is technically compatible with `this.<moduleName>` usage but won't typecheck unless you use `this.deps.<moduleName>` iirc)

emallson (approved):
this seems okay to me. there is a lot of text, which has me a little wary of it. but cross the maintenance bridge when we come to it

.github/instructions/analyzer-patterns.md
Seriousnes: That's a good catch. LLMs are pretty good at understanding usage from examples, but just to be certain I've explicitly documented that `.withDependencies` is the preferred option and `static dependencies` should only be used when `withDependencies` can't be (i.e. `MajorCooldown`). I've also documented that `static dependencies` aren't available in the constructor, and parent dependencies must be spread

emallson:
looks like one more prettier fix, then we can merge

PR #7569, opened 2026-01-07
#7565 refactor loading of players within a fight

### LINKED ISSUE #7565: Spec is not found when player changes it during a report
The spec of a player seems to depend on the global information of the report, which makes it so that things such as `player.icon` do not convey the player spec for a particular fight.

Example here: https://wowanalyzer.com/report/nXgDxrhPVq2NKA3F/6-Heroic+Jin'rokh+the+Breaker+-+Kill+(3:58). Image below for reference.

Image

By cloning the repo locally and manually forcing config to return `ClassicWarlockDestruction` from `getConfig` if `!config && player.type === "Warlock"` I am able to get a correct analysis. See images below.

Image

Image

Furthermore, the error message that is displayed does not convey the real issue, as it shows that "the spec is not currently supported", when the issue is an internal error in correctly inferring the spec.

emallson:
closes #7565

this replaces the use of combatantinfo with the use of the playerDetails v2 api (see backend PR https://github.com/WoWAnalyzer/server/pull/73).

this is a relatively big patch because of the way that so many pieces relied on the initial `combatantinfo` data directly, but most of those places only used the id, name, or guid---all of which are set on `playerDetails` as well

## Before
https://wowanalyzer.com/report/nXgDxrhPVq2NKA3F/6-Heroic+Jin'rokh+the+Breaker+-+Kill+(3:58)

image

## After
image

(i don't have a bnet api key set up locally on this computer to load avatars)


emallson:
e2e tests are going to fail until the backend part is merged. i haven't run locally yet bc i am not set up properly to run playwright atm, but will confirm e2e passes before merging

src/interface/report/PlayerLoader.tsx
```diff
@@ -1,291 +1,79 @@
...
-  const renderError = (error: Error) => {
-    return handleApiError(error, () => {
-      dispatchFC({ type: 'reset' });
-      // We need to set the combatants in the global state so the NavigationBar, which is not a child of this component, can also use it
-      setCombatants(null);
-      navigate(makeAnalyzerUrl());
-    });
-  };
+    for (const player of data.players) {
+      result.ilvl += player.ilvl;
```
Krealle: Looks like the `ilvl` prop can be undefined. Example: `report/GNyjgnAz6rDLxJ4b/13-Mythic+Mug'Zee,+Heads+of+Security+-+Kill+(5:30)`

Image

Live behavior treats it as a `0`
Image

src/interface/report/PlayerTile.tsx
```diff
@@ -82,7 +81,7 @@ const PlayerTileContents = ({ avatar, player, spec }: PlayerTileContentsProps) =
           </small>
           <div className="flex text-muted text-small">
             <div className="flex-main">
-              <Icon icon="inv_helmet_03" /> {Math.round(getAverageItemLevel(player.combatant.gear))}
+              <Icon icon="inv_helmet_03" /> {player.ilvl}
```
Krealle: Same here, `ilvl` can be undefined which is treated as a `0` on live.

Krealle (approved):
Tested with the backend checked out and besides the undefined item levels, everything seems to be working well.
Should be gtg once that and tests are fixed.

PR #7577, opened 2026-01-12
Enable Shadow Priest for Midnight

DoxAshe:
### Description

This gets shadow priest started for the midnight expansion. It removes the old unused abilities and is able to analyze a Midnight Beta Log without crashing. More updates will be needed, but this is a good start.


src/analysis/retail/priest/shadow/modules/core/Pet.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
emallson: rather than this `eslint-disable`, use `pnpm run lint-baseline update`. that will allow current uses through the linter while excluding new ones. (however, it is unfortunately file-position-based so adjusting comments at the top of the file is going to impact the baseline)

src/analysis/retail/priest/shadow/modules/features/DotUptimes.tsx
```diff
@@ -45,11 +45,11 @@ class DotUptimes extends Analyzer {
     return explanationAndDataSubsection(explanation, data);
   }
 
-  get guideSubsectionDP() {
+  /*get guideSubsectionDP() {
```
emallson, line 48: its generally better to just delete rather than mass commenting-out like this. we all always intend to come back and fix the things we comment out, but, well, it doesn't always happen :) and deleting leaves the codebase in a better state.

however, for practical reasons i think the sanest solution is to merge with the commented out stuff because the overall code changes look fine and i need to get gamedata updates some (and i can do better fixes with the PR'd specs merged)

emallson (approved):
im going to push a baseline update for the `any` thing rather than the file-wide disable, but otherwise good to merge.

PR #7583, opened 2026-01-17
Feature survival boomstick

Kivlov:
### Description

I have another PR after this one that does further cleanup and adds in the missing new ability spellIDs in SPELLS but because Boomstick is integral to the rotation now, and is a complete new addition I'm adding it on it's own.

I used claude opus to plan and base the analyzer off of the windwalker Fist of Fury analyzer using the instructions in the .github. The damage is dealt in 4 ticks over a 3s (hasted) channel so I have it group the damage into the ticks for display & for performance.

In order to capture if it was clipped, I set the normalizer to 100ms buffer but I found there was the odd cast the player would have a very minimal delay after the channel ended and so if it picks up a cast but it makes 4 buckets/ticks of damage then it considers it not-clipped. The ability does a significant amount of damage and so missing a tick of it is really bad.

3f2051a8fe3d8e4f19136b6b962b34f25c78450c
45126d9ee204fbcf294b4af8fb81e23e47a8c0c2
These 2 commits have simple versions of the analyzer. If you look at the haste calculation commit and the analysis is too hard to follow (I leaned on claude to refactor for the hasted ticks rather than just make a plan), I will just get rid of it and fall back to the simpler one for now.

- Test report URL: /report/AZ8jBw1CHx76r9pm/1-Mythic++Pit+of+Saron+-+Kill+(27:27)/5-Giftia/standard)
- Test report with lots of clipped and missed casts: report/Bx96hAMnzVZKyGRq/1-Mythic++Skyreach+-+Kill+(57:02)/8-Giftia/standard
- I have no raid log to test with unfortunately at the moment. I'll take a look and on the PR that will follow this one that does more cleanup, I'll see if I can get a log to replace the example log. I'll have my own pre-patch logs in a week to use if I cant find a good one.

### DIFF
--- a/src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
+++ b/src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
@@ -5,6 +5,7 @@ import TALENTS from 'common/TALENTS/hunter';
 import { SpellLink } from 'interface';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Events, {
+  BeginCastEvent,
   CastEvent,
   DamageEvent,
   EndChannelEvent,
@@ -77,7 +78,7 @@ class Boomstick extends Analyzer.withDependencies({ haste: Haste }) {
 
     const tipped = this.selectedCombatant.hasBuff(SPELLS.TIP_OF_THE_SPEAR_CAST.id, cast.timestamp);
     const tickResults = this.analyzeTicksForCast(cast);
-    const nextAbility = GetRelatedEvent(event, BOOMSTICK_NEXT_CAST);
+    const nextAbility = GetRelatedEvent<CastEvent | BeginCastEvent>(event, BOOMSTICK_NEXT_CAST);
 
     // Aggregate stats
     const ticksHit = tickResults.filter((t) => t.hit).length;
@@ -169,7 +170,7 @@ class Boomstick extends Analyzer.withDependencies({ haste: Haste }) {
     ticksHit: number,
     wasClipped: boolean,
     missedTicks: number[],
-    nextAbility: ReturnType<typeof GetRelatedEvent>,
+    nextAbility: CastEvent | BeginCastEvent | undefined,
   ): { value: QualitativePerformance; header: JSX.Element; clippingInfo: JSX.Element | null } {
     let value: QualitativePerformance;
     let header: JSX.Element;
@@ -286,6 +287,7 @@ class Boomstick extends Analyzer.withDependencies({ haste: Haste }) {
         <BoringSpellValueText spell={TALENTS.BOOMSTICK_TALENT}>
           <>
             <ItemDamageDone amount={this.totalDamage} />
+            <br />
             {this.totalHits} <small>targets hit</small>
             <br />
             {(this.totalTicks > 0 ? this.totalHits / this.totalTicks : 0).toFixed(1)}{' '}
--- a/src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
+++ b/src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
@@ -41,8 +41,6 @@ class Boomstick extends Analyzer {
   private useEntries: BoxRowEntry[] = [];
   private clippedCasts = 0;
 
-  private castTippedStatus = new Map<number, boolean>();
-
   private static readonly BUCKET_WINDOW_MS = 100;
   private static readonly EXPECTED_TICKS = 4;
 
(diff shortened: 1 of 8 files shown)


src/common/SPELLS/hunter.ts
```diff
@@ -386,6 +386,11 @@ const spells = {
     name: 'Merciless Blow',
     icon: 'ability_hunter_swiftstrike',
   },
+  BOOMSTICK_DAMAGE: {
+    id: 1261215,
+    name: 'Merciless Blow',
+    icon: 'ability_hunter_swiftstrike',
```
emallson: are these right? the beta spell with that id: https://www.wowhead.com/beta/spell=1261215/boomstick

src/analysis/retail/hunter/survival/normalizers/BoomstickNormalizer.ts
```diff
@@ -0,0 +1,61 @@
...
+export const BOOMSTICK_CAST_HIT = 'boomstick-cast-hit';
+export const BOOMSTICK_CAST_END = 'boomstick-cast-end';
+export const BOOMSTICK_NEXT_CAST = 'boomstick-next-cast';
+const MAX_CHANNEL_BUFFER_MS = 4000;
+
+const EVENT_LINKS: EventLink[] = [
+  // Link damage to cast
+  {
+    linkRelation: BOOMSTICK_CAST_HIT,
+    reverseLinkRelation: BOOMSTICK_CAST_HIT,
```
emallson, line 17: i am not sure that the link normalizer will behave as expected if you supply the same relation name in both directions. maybe it will?

src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
```diff
@@ -0,0 +1,306 @@
...
+    }, []);
+
+    // Calculate expected tick times based on haste
+    const hasteAtCast = this.deps.haste.current;
+    const tickInterval = Boomstick.BASE_TICK_INTERVAL_MS / (1 + hasteAtCast);
+
+    // Match each expected tick slot to actual damage buckets
+    return [0, 1, 2, 3].map((tickIndex) => {
+      const tickNumber = tickIndex + 1;
+      const expectedTime = cast.timestamp + tickInterval * tickIndex;
```
emallson, line 149: while this method likely works, i think it is more complicated and (probably) more likely to run into latency-related issues. you're already grouping hits into buckets (lines 131-139). do you need to do this step, or can you treat each bucket as a tick?

src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
```diff
@@ -0,0 +1,306 @@
...
+    });
+  }
+
+  // Determine cast performance and generate header/clipping info
+  private classifyCast(
+    tipped: boolean,
+    ticksHit: number,
+    wasClipped: boolean,
+    missedTicks: number[],
+    nextAbility: ReturnType<typeof GetRelatedEvent>,
```
emallson: should be either `AnyEvent` or `CastEvent|BeginCastEvent`, not a return type referencing `GetRelatedEvent`.

src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
emallson: oh, after re-reading and checking the code out i see the reason: what happens if a tick hits 0 targets? that makes sense. i like showing the ticks that entirely missed 👍

emallson (changes requested):
image

UI issues on the statistic but otherwise this looks pretty good.


not sure exactly how `EventLinkNormalizer` behaves with forward/reverse links having the same key, but if it works it works

src/analysis/retail/hunter/survival/normalizers/BoomstickNormalizer.ts
Krealle: It works fine, Evoker has a bunch of castlinks like that.

src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
Kivlov: Updated to use castevent|begincastevent|undefined

src/common/SPELLS/hunter.ts
Kivlov: Copy pasted and then changed the ID but forgot to update the name and icon.

src/analysis/retail/hunter/survival/modules/talents/Boomstick.tsx
Kivlov: image
and also fixed the statistic appearance

Kivlov:
Did I still need to fix something or just waiting on you having time to get to it (Dont wanna pester since I suspect you're quite busy but just wanna make sure I didnt miss fixing something)

emallson:
> Did I still need to fix something or just waiting on you having time to get to it (Dont wanna pester since I suspect you're quite busy but just wanna make sure I didnt miss fixing something)

afaik no, i just need to catch up on reviews

emallson (approved):

PR #7584, opened 2026-01-18
game data updates for 12.0.1

emallson:
### Description

this updates talent data to version 12.0.1.65337 (aka current beta version). i did some best effort updates to the affected specs. the major changes are:

- Havoc DH: Netherwalk removed
- Mistweaver / Windwalker (big changes!): Conduit has distinct spell ids for MW now (confirmed that the new id is actually cast [here](https://www.warcraftlogs.com/reports/27BkY8zMhbLd4P6D?fight=4&type=casts&source=6))
- Shadow Priest: Depth of Shadows removed / baked into Shadowfiend

I tested with the sample reports, but not all of them are updated.


DoxAshe:
This looks fine from the shadow priest side of things.  I am going to have to redo all of the pets talents soon.

PR #7592, opened 2026-01-22
[#7589] fix missing Combatant values for other players in a log

### LINKED ISSUE #7589: `this.combatants.players` only includes the selected player
This has been an issue on live since 8th January. Had previously assumed it was a live-only issue as I was unable to replicate it on my Midnight branch, but it's also possible that I hadn't merged in the commit that caused the issue.

Haven't found any reports of it occurring before 8th January, so may be due to #7569 which was merged in on that date.

`this.combatants.players` only includes the selected player, not any other players. This in turn breaks several modules relying on it, including `PrescienceNormalizer.tsx` and all related Prescience modules for Aug.

emallson:
closes #7589 

### Description

in #7569, we refactored the loading of characters for a fight. we no longer load combatantinfo for the fight independently. there was a bug in the implementation, where we passed in an array of combatantinfo but this only included the selected player's info---resulting in `Combatants` containing only the player.

this fixes it by decoupling `Combatants` from `combatantinfo`. non-player combatants are constructed without `combatantinfo` and do not have gear etc. the player combatant is required to have `combatantinfo`. this was technically a requirement prior to #7569 to even select a player. however, it is no longer a prerequisite to selecting the player---which leads to a potential page crash if combatantinfo is actually missing. i'm making an issue for providing a better error, but don't want to delay fixing the breakage for Aug

while I was here, I removed the final vestiges of `combatantinfo` being a separate set of events in core normalizers.

### Testing

This is deploying to `next` for testing. I tested on: `/report/VgJ3xvBqmXAR2Wck/3-Mythic+Plexus+Sentinel+-+Kill+(1:00)/11-Trinvoker/standard/overview`

The most relevant things to check are:

- are pre-pull buffs still set correctly? (yes)
image
- are other players present in `combatants`? (i don't know how to verify this through the Prescience analysis that triggered this PR, but I checked via debugger and a cheeky console.log)
image

the main risks of this PR are that the changes could break:
- pre-pull buff normalization that previously relied on other players' combatantinfo (but this would already be broken)
- normalizers that assumed there were no combatantinfo events
- code that assumed all `Combatant`s had combatantinfo-related data, using non-null assertions or old untyped JS code


Krealle:
Didn't look at the code, but tested the Augmentation modules and they all look to behave as expected with these changes.
There is some pre-pull buff normalization that Aug does which relies on other players' combatantinfo, but it's nbd if that stays broken. It's **very** situationally relevant.

emallson:
> There is some pre-pull buff normalization that Aug does which relies on other players' combatantinfo, but it's nbd if that stays broken. It's very situationally relevant.

we could also reintroduce this load if it is relevant, either at the top level or with an extra load for aug later. (maybe not in this PR though)

Krealle:
> we could also reintroduce this load if it is relevant, either at the top level or with an extra load for aug later. (maybe not in this PR though)

It's definitely fine to skip for now. Will have to talk with @KYZ124 about the future of Aug, might not need it at all going forwards.

PR #7595, opened 2026-01-23
[Survival] Further Cleanup and Midnight Abilities addition

Kivlov:
### Description

Cleaning out old implementations of talents. Tip of the spear is now different (has been all of TWW but the analysis for it didnt really matter so it was never touched).

The CastLinkNormalizer and Wildfire Bomb normalizer changes are included but I have not yet included the change to the WildfireBomb.tsx as the analysis for it is going to change drastically and would make this a much longer review. I can just remove the normalizer entirely if you'd like to avoid having it added bu tnot hooked up to anything (the cast link normalizer for SV is currently not used anywhere anyways).

Added in CDR talents for Wildfire Shells and Lethal Calibration. Lunge was changed to be a flat 1s reduction instead of 1s per auto. They made Kill Command 5s with 2 charges base and Flanker's Advantage does something entirely different now so deleted that analyser also.

This should be the last bulk cleanup. There is likely still some unused stuff that will be purged.

### Testing

Shouldn't need much testing as it's largely pruning old abilities. 

- Test report URL: `report/FmKC8ZJxaN34fMVY/33/2`
Grabbed a random report from Fractillus to make sure it loaded raid kills properly.

`report/CFZXHhr17vpTfdRz/18/8` Soul hunters report to see the new CDR analyzers.


emallson (approved):
> I can just remove the normalizer entirely if you'd like to avoid having it added bu tnot hooked up to anything (the cast link normalizer for SV is currently not used anywhere anyways).

just mentioning it like this is fine 👍

Issue #7593, opened 2026-01-22
Show a better error page when selecting a player that has no combatantinfo


emallson:
Sometimes combatantinfo isn't dumped for a fight due to game bugs. Prior to #7569 we would not allow you to select the player. After it, we rely on other information to generate the player list and it is now possible to select players without combatantinfo.

While this used to be common in classic, it is now far less common. (more comparable with retail)

We should have a dedicated error screen for this case, regardless.

PR #7606, opened 2026-01-27
add a warning for unsupported fights

ToppleTheNun:
### Description

adds a warning for dummy fights

### Testing

- Screenshot(s):
image
image


emallson:
this fixes dummies but doesn't close #7593. it is possible to have boss fights without `combatantinfo` events as well as trash fights (which normally do not have `combatantinfo`)

ToppleTheNun:
I'll get that added in this evening

emallson:
my thought for this was to move the code in `useEventParser` that finds the player's combatantinfo up a level into the report loader, which can then use it to show an error dialog if the user has no combatantinfo. we could also reintroduce the bulk combatantinfo load to prevent clicking one of these in the first place (partially reverting the previous PRs refactoring this)

PR #7613, opened 2026-01-29
Fix/i18n ru core pages

SaltyRain:
### Description

This PR improves internationalization support and UI semantics for core pages.

Changes are split into logical commits for easier review, but represent one cohesive i18n improvement:

- Improved `<Panel>` component to better support translatable headings:
  - Added support for translation IDs (`titleTransId`)
  - Added configurable heading levels (`h1`–`h6`)
  - Added optional anchor generation with safer anchor IDs
- Refactored existing pages and subcomponents to use the enhanced `<Panel>` instead of plain `div` headers
- Added missing EN message keys as the source locale
- Added full RU localization for core pages:
  - `/news`
  - `/specs`
  - `/about`
  - `/premium`
  - `/help-wanted`
- Synced message catalogs across all locales to include new IDs
- Added a single consolidated changelog entry for this improvement

### Testing

- Switch UI language to **RU** and verify translations on the pages listed above
- Verify that headings render with correct semantic levels (`h1` → `h2` where appropriate)
- Verify that heading anchors work when enabled
- Confirm that all locales build without missing message IDs

- Test report URL: N/A
- Screenshot(s): N/A


SaltyRain:
**Questions / follow-ups for the future**

1. I noticed some inconsistency in localization key naming, for example:
   `interface.premiumLoginPanel.logged-in.premiumNotUnlocked`

   Is there a preferred convention for message IDs in this project?
   - camelCase vs kebab-case
   - are segments like `logged-in` / `sign-in` intentional exceptions, or historical leftovers?

   Happy to follow (or help enforce) a consistent convention if there is one.

2. For this PR I intentionally combined localization changes and related functional updates into a single PR, so that they result in a single consolidated changelog entry.

   For future contributions, would you prefer:
   - one PR with multiple logical commits (as done here), or
   - separate PRs for functional changes vs localization updates, even if that leads to multiple changelog entries?

3. When adding new localization keys, I manually went through all locale files to add empty placeholders for the new IDs.

   Is there an existing script or recommended workflow in the project to automatically sync localization keys across languages (e.g. add missing keys with empty values), or is the manual approach expected?

Thanks!

SaltyRain:
Okay I would require some help here. I can't figure out why tests are failing. Locally everything works.

Sharrq:
Try pushing a trivial change to get it to run the tests again. It looks like they just timed out

ToppleTheNun:
For one of the tests:
image
It's looking for `About WoWAnalyzer` but the i18n is failing to put the title in.

Same for the premium and help wanted tabs.

src/interface/routes/about.tsx
```diff
@@ -17,7 +17,7 @@ export function Component() {
     <>
       <DocumentTitle title="About WoWAnalyzer" />
 
-      <Panel title={<Trans id="interface.aboutPage.title">About WoWAnalyzer</Trans>}>
+      <Panel title="About WoWAnalyzer" titleTransId="interface.aboutPage.title" addAnchor={false}>
```
ToppleTheNun: you don't need to change how the title is passed here to get the same effect; keep the `Trans` component usage and add the `addAnchor={false}` prop.

src/interface/routes/about.tsx
```diff
@@ -67,7 +67,11 @@ export function Component() {
       </Panel>
       <br />
 
-      <Panel title="The WoWAnalyzer Discord bot">
+      <Panel
+        title="The WoWAnalyzer Discord bot"
+        titleTransId="interface.aboutPage.introducingWOWABotHeading"
```
ToppleTheNun: same as the above regarding the use of the `Trans` component.

src/interface/routes/about.tsx
```diff
@@ -67,7 +67,11 @@ export function Component() {
...
+        titleTag="h2"
```
ToppleTheNun: this was `h1` before due to it being another panel section. it's fine to leave it as an `h1`.

src/interface/routes/help-wanted.tsx
```diff
@@ -1,4 +1,5 @@
-import { Trans, t } from '@lingui/macro';
+import { Trans } from '@lingui/react/macro';
+import { t } from '@lingui/core/macro';
```
ToppleTheNun: where possible, we should be using the `useLingui()` hook instead of `t` from `@lingui/core/macro`.

src/interface/routes/help-wanted.tsx
```diff
@@ -11,7 +12,9 @@ export function Component() {
       <DocumentTitle title="Help wanted" />
 
       <Panel
-        title={<Trans id="interface.helpWantedPage.helpWantedTitle">Help wanted</Trans>}
+        title="Help wanted"
```
ToppleTheNun: same as the above regarding the use of the `Trans` component.

src/interface/routes/premium.tsx
```diff
@@ -27,200 +29,188 @@ export function Component() {
...
         </div>
         <div className="col-md-8 col-sm-7">
-          <div className="panel">
-            <div className="panel-heading">
-              <h1>
-                <Trans id="interface.premiumPage.premium">WoWAnalyzer premium</Trans>
-              </h1>
+          <Panel
+            title="WoWAnalyzer premium"
+            titleTransId="interface.premiumPage.premium"
```
ToppleTheNun: same as the above regarding the use of the `Trans` component.

src/interface/routes/premium.tsx
```diff
@@ -1,4 +1,5 @@
-import { t, Trans } from '@lingui/macro';
+import { t } from '@lingui/core/macro';
```
ToppleTheNun: same thing regarding the usage of `useLingui()`

src/interface/ChangelogPanel.tsx
```diff
@@ -14,51 +15,44 @@ const ChangelogPanel = () => {
...
-      </div>
-      <div className="panel-body pad">
-        <select
-          className="form-control"
-          value={changelogType}
-          onChange={(e) => setChangelogType(Number(e.target.value))}
-        >
-          <option value={0}>
-            {t({ id: 'interface.changelogPanel.option.core', message: 'Core' })}
+    <Panel title="Changelog" titleTransId="interface.changelogPanel.heading" titleTag="h2">
```
ToppleTheNun: same as the above regarding the use of the `Trans` component.

src/interface/Panel.tsx
```diff
@@ -1,46 +1,80 @@
...
+  actions,
+  backButton,
+  as = 'h1',
+}: HeadingProps) => {
+  if (title == null) return null;
+
+  const isStringTitle = typeof title === 'string';
+  const anchorId = isStringTitle ? makeAnchorId(title) : undefined;
+
+  const translatedTitle =
```
ToppleTheNun: while I appreciate the effort here, adding this here just obfuscates the translation being done. this should be handled in-place by users of the `Heading` / `Panel` components by providing a `Trans` component or a string.

src/interface/Panel.tsx
```diff
@@ -1,46 +1,80 @@
...
+  addAnchor = true,
+  explanation,
+  actions,
+  backButton,
+  as = 'h1',
+}: HeadingProps) => {
+  if (title == null) return null;
+
+  const isStringTitle = typeof title === 'string';
+  const anchorId = isStringTitle ? makeAnchorId(title) : undefined;
```
ToppleTheNun: I do very much like the anchor ID stuff. good work.

src/interface/Panel.tsx
```diff
@@ -1,46 +1,80 @@
...
+    .replace(/^-+|-+$/g, '');
+
+const Heading = ({
+  title,
+  titleTransId,
+  addAnchor = true,
+  explanation,
+  actions,
+  backButton,
+  as = 'h1',
```
ToppleTheNun: we generally try to avoid this pattern and instead use an indicator like `subheading: boolean` where it changes how we render the component.

ToppleTheNun (changes requested):
Always nice to see translation work being done. Core stuff needs it a lot.

I commented on a few things that need to change to keep with our patterns.

SaltyRain:
@ToppleTheNun Thanks a lot for the detailed feedback — and especially for the screenshot, that helped a lot 👍

I traced the issue down to the fact that Trans from @lingui/react/macro does not support dynamic ids properly. As an experiment, I tried importing Trans from @lingui/react, which does handle dynamic ids at runtime, and that fixed the rendering issue.

That said, I see your point about this being an undesirable pattern in this codebase. I’ll revert this approach and update the usage to match the existing conventions as you suggested.

src/interface/Panel.tsx
SaltyRain: Makes sense. Should I replace as with subheading?: boolean (h1 vs h2), or do you prefer Panel to always render h1?

src/interface/Panel.tsx
SaltyRain: Thanks! Glad the makeAnchorId helper makes sense 👍

One thing I want to call out though: in the previous implementation anchors were effectively only added when title happened to be a plain string. If a panel title was passed as a <Trans /> node (or any other ReactNode), the anchor was silently not applied.

With the current logic (anchor only when typeof title === 'string'), keeping the <Trans /> usage means this behavior remains — i.e. addAnchor has no effect for ReactNode titles, and anchors are still missing in those cases.

I just wanted to highlight this, since it looks like an existing inconsistency rather than an intentional choice.

src/interface/Panel.tsx
ToppleTheNun: I prefer `subheading?: boolean` where it defaults to `h1` if `false` or `undefined`.

ToppleTheNun:
> @ToppleTheNun Thanks a lot for the detailed feedback — and especially for the screenshot, that helped a lot 👍
> 
> I traced the issue down to the fact that Trans from @lingui/react/macro does not support dynamic ids properly. As an experiment, I tried importing Trans from @lingui/react, which does handle dynamic ids at runtime, and that fixed the rendering issue.
> 
> That said, I see your point about this being an undesirable pattern in this codebase. I’ll revert this approach and update the usage to match the existing conventions as you suggested.

yep, it's because the macros are done at build time while the non-macro variants are not.

src/interface/Panel.tsx
ToppleTheNun: As I'm thinking about this, it may be worth adding `anchor` as a prop instead of `addAnchor`. We could make it a warning in ESLint. Panel isn't used so frequently that `// eslint-disable-next-line`-ing the warning for the few that don't need an anchor would be problematic.

@emallson thoughts?

emallson: i think that making the prop `anchor` seems good. we could do a pass on current callers and set anchors for them. not sure how many there are off-hand but you're right that it is relatively uncommon. doing a pass on that now is maybe not in scope for this PR? uncertain

src/interface/routes/help-wanted.tsx
SaltyRain: Here we use 

`import { useLingui } from '@lingui/react';`

or

`import { useLingui } from '@lingui/react/macro';`

I can see that in the project we use `import { useLingui } from '@lingui/react'`;

ToppleTheNun: the former; you can then do:
```ts
const { i18n } = useLingui();

i18n.t({
	// whatever
});
```

SaltyRain: ah I see, 
`import { useLingui } from '@lingui/react';`  does not have a `t`

SaltyRain:
I’ve addressed the review comments and pushed the updates.

One e2e test is currently failing due to a “Partial support” overlay intercepting clicks in the report view — I’m not sure if this is related to the changes here

emallson (approved):
looks like you hit all the review comments 👍 i'll worry about the e2e tests, any breakage here is either unrelated or incidental to label changes

PR #7614, opened 2026-01-30
[Mistweaver] Update Sheilun's Gift modules

lolswirl:
### Description
Updated all Sheilun's Gift modules for Mistweaver.
- For all SG modules, normalize Invigorating Mists healing on SG's main target to effectively remove Invig healing from their calcs entirely and just utilize base value for unskewed data
- Created Invigorating Mists module to calc the 500% increase on Sheilun's Gift or just simply get Invig Mists healing from Vivify
- Created Emperor's Favor module to calc 20% amp on the heal and gets your cast time saved from the instant cast down from 2s (hasted)
- Created Tranquil Tea module to use the cloud stack gain via Mana Tea drinking you made from last cast to latest cast of SG to get the ratio from Tranquil Tea's contribution

### Testing
 `report/FMKmTZk9NhrwGqtp/7-Mythic++Maisara+Caverns+-+Kill+(27:41)/255-Ðextermonk/standard/statistics`
Sheilun's + Legacy + Veil
image
image
image
image

`report/Ac7K16v3abnMPCrF/2-Mythic++Algeth'ar+Academy+-+Kill+(28:12)/1-Vek/standard/statistics`
Sheilun's + Emperor's Favor + Veil
image
image
image
image

`/report/FWpdMHLzXD3f6Yam/14-Mythic++Magisters'+Terrace+-+Kill+(32:06)/3-Toixic/standard/statistics`
Sheilun's + Legacy + Tranquil Tea
image
image
image
image
image
image

`/report/CRX1Z7y84HdbvzPk/36-Mythic+Imperator+Averzian+-+Wipe+17+(3:38)/19-Vohrr/standard/statistics`
Vivify + Invig
image


src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -167,6 +169,32 @@ function getClosestEvent(timestamp: number, events: AnyEvent[]): AnyEvent {
...
+  const index = healEvents.indexOf(mainTarget);
+
+  const multiplier = 1 + INVIGORATING_MISTS_INCREASE;
+
+  healEvents[index] = {
+    ...mainTarget,
+    amount: mainTarget.amount / multiplier,
+    absorbed: (mainTarget.absorbed || 0) / multiplier,
+    overheal: (mainTarget.overheal || 0) / multiplier,
+  };
```
emallson: rather than rewriting the `healEvents` array, you should return a new array to prevent spooky action at a distance

emallson (changes requested):
one change, otherwise looks good. hard for me to tell if this `normalizeSheilunsGiftMainTarget` call actually impacts other analyzers. i don't think it does because `GetEvents` returns a new list (via the initial `.filter` call)

src/analysis/retail/monk/mistweaver/modules/spells/InvigoratingMists.tsx
vohrr, line 1: Recommend only loading this and doing analysis for when SG is talented. The healing contribution from invig with vivify is very clear just from looking directly at a log, and the events are already being evaluated in the Vivify module so doing it here is a duplication.

src/analysis/retail/monk/mistweaver/modules/spells/EmperorsFavor.tsx
```diff
@@ -0,0 +1,82 @@
...
+import Uptime from 'interface/icons/Uptime';
+import { formatPercentage } from 'common/format';
+import { SpellLink } from 'interface';
+import StatisticListBoxItem from 'parser/ui/StatisticListBoxItem';
+import { EMPERORS_FAVOR_INCREASE } from '../../constants';
+import { calculateEffectiveHealing } from 'parser/core/EventCalculateLib';
+import Haste from 'parser/shared/modules/Haste';
+import GlobalCooldown from '../core/GlobalCooldown';
+
+const SG_BASE_CAST_TIME = 2000;
```
vohrr, line 19: Would prefer to pull this from spell-list[...].castTime or some such similar. Or atthe very least toss it in the constants.ts file

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -167,6 +169,32 @@ function getClosestEvent(timestamp: number, events: AnyEvent[]): AnyEvent {
...
+    const highestTotal = effectiveHealing(highest) + (highest.overheal || 0);
+    return currentTotal > highestTotal ? heal : highest;
+  });
+}
+
+// normalize sheilun's gift heal events to remove invigorating mist increase
+export function normalizeSheilunsGiftMainTarget(healEvents: HealEvent[]): void {
+  if (!healEvents || healEvents.length === 0) return;
+
+  const mainTarget = getHighestHeal(healEvents);
```
vohrr: Small thing but we can just normalize and early return when the length is 1 instead of calling `getHighestHeal()` (use case for Emperor's Favor, but would be true for any array length of 1 in this case)

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
```diff
@@ -306,6 +334,13 @@ export function getSheilunsGiftHits(event: CastEvent): HealEvent[] {
   return GetRelatedEvents<HealEvent>(event, SHEILUNS_GIFT);
 }
 
+export function getSheilunsGiftMainTargetHit(event: CastEvent): HealEvent | undefined {
```
vohrr: We could do this nicely with an event link matching on event.targetIDs between the Heal and CastEvent. Matching based on the highest number isn't factoring in crits, which could subject this to breaking in the future if the 500% gets brought down low enough.

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
vohrr: Actually on second thought we probably shouldn't be doing this for the emperor's favor healing analysis because it will undervalue the talent by a significant amount

src/analysis/retail/monk/mistweaver/modules/spells/EmperorsFavor.tsx
```diff
@@ -0,0 +1,82 @@
...
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS_MONK.SHEILUNS_GIFT_TALENT),
+      this.onCast,
+    );
+  }
+
+  onCast(event: CastEvent) {
+    const sgHealEvents = getSheilunsGiftHits(event);
+    if (!sgHealEvents || sgHealEvents.length === 0) return;
+
+    normalizeSheilunsGiftMainTarget(sgHealEvents);
```
vohrr: We probably shouldn't be doing this for the emperor's favor healing analysis because it will undervalue the talent by a significant amount
`base heal = 100` 
`invigMod = baseHeal * (1 + 5) => 600`
`empFavMod = invigMod * (1 + 0.2) => 720` 
`effective increase => 120`
compared to after normalization:
`720 / (1 + 5) = 120`
`effective increase => 20`

src/analysis/retail/monk/mistweaver/modules/spells/TranquilTea.tsx
```diff
@@ -0,0 +1,116 @@
...
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS_MONK.SHEILUNS_GIFT_TALENT),
+      this.onSheilunsGiftCast,
+    );
+  }
+
+  onManaTeaStackRemove(event: RemoveBuffEvent | RemoveBuffStackEvent) {
+    if (this.selectedCombatant.hasBuff(SPELLS.MANA_TEA_CAST.id, event.timestamp, 50)) {
+      this.extraCloudsAtLastCast += 1;
+      this.totalExtraClouds += 1;
+    }
```
vohrr, line 61: Would be nice to add a metric here for wasted cloud generation from overdrinking

src/analysis/retail/monk/mistweaver/modules/spells/SheilunsGift.tsx
```diff
@@ -48,21 +60,39 @@ class SheilunsGift extends Analyzer {
     );
   }
 
+  calcEmperorsFavor(healEvents: HealEvent[]) {
```
vohrr, line 85: I'm torn on this. I get what you're going for with continuing the deconstruction of all the modifiers from other affecting talents, but its contrary to how we analyze talent healing contributions in most other cases. 
I won't hold the PR up for this change since its just accommodating the new talent, but maybe something to chew on in the future -- combine all these into a TalentAggregateStatistic like Rising Mist so that the whole picture/breakdown is more clear and presented in a single graphic

vohrr (changes requested):

src/analysis/retail/monk/mistweaver/modules/spells/InvigoratingMists.tsx
lolswirl: I totally disregarded the Vivify module - will just analyze for SG only for this

src/analysis/retail/monk/mistweaver/modules/spells/EmperorsFavor.tsx
lolswirl: Does MW have this spell-list file? I only saw a classic module when I initially looked, unless I'm completely blind

src/analysis/retail/monk/mistweaver/modules/spells/SheilunsGift.tsx
lolswirl: I agree with this and honestly don't mind holding off - I wanted to keep the same logic used for Legacy/SG that effectively removed each other's contributions (differing from the other modules that we currently have), but if you feel differently I can rework them all

lolswirl:
> one change, otherwise looks good. hard for me to tell if this `normalizeSheilunsGiftMainTarget` call actually impacts other analyzers. i don't think it does because `GetEvents` returns a new list (via the initial `.filter` call)

`normalizeSheilunsGiftMainTarget` should not impact other analyzers, yeah - its solely for use within the modules of this PR to remove the 500% amp on invigorating mists. `CastLinkNormalizers` is probably not a great place for it to live, since it doesn't change all instances of the events. I'll take Vohrr's suggestion for creating an event link directly [here](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7614#discussion_r2750459561)

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
lolswirl: Resolving as EventLink instead of this implementation anyways - going to redo EF analysis

src/analysis/retail/monk/mistweaver/modules/spells/TranquilTea.tsx
lolswirl: image

src/analysis/retail/monk/mistweaver/modules/spells/EmperorsFavor.tsx
lolswirl: Removed normalizing entirely for EF

lolswirl: Bit more realistic here 
image

src/analysis/retail/monk/mistweaver/normalizers/CastLinkNormalizer.ts
lolswirl: Not entirely sure the best way to get this across in Typescript - I believe maps create new arrays directly (and are a bit more readable) but is another method better? I implemented in this way https://github.com/lolswirl/WoWAnalyzer/blob/22e695895d47955d198f913795d2a6e66317efb0/src/analysis/retail/monk/mistweaver/modules/spells/SheilunsGift.tsx#L36

emallson: `.map` like this is good 👍

src/analysis/retail/monk/mistweaver/modules/spells/EmperorsFavor.tsx
vohrr: It has been enabled for retail but the MW file hasn't been generated yet -- doesn't have to happen in this pr

vohrr (approved):

PR #7616, opened 2026-01-31
Add warning for missing combatantinfo

### LINKED ISSUE #7593: Show a better error page when selecting a player that has no combatantinfo

emallson:
### Description

closes #7593

i don't have a handy test log, but if you change the `useMemo` to just be `undefined`, the error screen will show

image


src/parser/core/EventFilter.ts
```diff
@@ -1,8 +1,6 @@
-import { EventType } from 'parser/core/Events';
+import type { EventType } from 'parser/core/Events';
 import { SELECTED_PLAYER, SELECTED_PLAYER_PET } from './Analyzer';
 
-const VALID_BY_FLAGS = SELECTED_PLAYER | SELECTED_PLAYER_PET;
-
```
emallson, line 5: this is the bit that i moved around to eliminate the circular reference error. i have *no idea* why this triggered. slapping `import type` on everything clearly didn't help, so 🤷

Krealle (approved):
Looks good! Was able to test with `/report/zQtqGWjMkPh48b6V/1-Heroic+Jin'rokh+the+Breaker+-+Kill+(3:26)/33-Melinoé/standard`

Image

PR #7618, opened 2026-02-01
[Survival] Sentinel Talent: Moonlight Chakrams

Kivlov:
### Description

Moonlight Chakram for Sentinel Hero talents.
It gives Survival 10s of wildfire bomb, and must be used within 15s of Takedown (Trueshot for MM) cast or it fades.
The goal is to minimise waste, not miss a cast, and try not to cast it on something low hp since it deals it's damage over time (up to 10s with the bug).

This contains analysis for Marksman. They get Lock and Load for free, so if they have the buff when they press Chakram then it's a bad use. Survivals is a sliding  amount of valid waste that will be tuned as sims come out.

Once we're level 90, I can remove all the checks for stalk and strike because right now they are not taken in single target, but they are taken in AoE.

-  `/report/B23zGhvLk1AxKFw4/16-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:44)/114-Percival/standard` <- Stalkless
- `/report/yMqfrbH9nv6txVRd/33/184` <- Stalk and Strike report
-

### DIFF
--- a/src/analysis/retail/hunter/shared/herotalents/MoonlightChakram.tsx
+++ b/src/analysis/retail/hunter/shared/herotalents/MoonlightChakram.tsx
@@ -31,9 +31,10 @@ export default class MoonlightChakram extends Analyzer {
   protected spellUsable!: SpellUsable;
 
   private totalDamage = 0;
-  private totalHits = 0;
   private chakramCasts = 0;
   private missedCasts = 0;
+  private totalWildfireBombCdr = 0;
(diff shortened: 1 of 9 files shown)


emallson (approved):
> Once we're level 90, I can remove all the checks for stalk and strike because right now they are not taken in single target, but they are taken in AoE.

unless these are hero talents (which everyone is guaranteed to have), imo leaving it in place since the work is already done may be better? in case the meta build changes, the analysis will still work

PR #7622, opened 2026-02-01
[BrM] rewrite Stagger tracking for Midnight

emallson:
### Description

Stagger tracking is the oldest Brewmaster code, dating back to when event fabrication was a normal means of operation.

These days, we not only have new, dedicated events for stagger (with some caveats, see the docstring for `StaggerPool`), but we also have newer tools in WoWA for handling complicated cases like this.

In addition, the correctness of the PB/Stagger implementation has bit-rotted a bit due to the number of secondary sources of stagger clearing. PB's purification is overestimated in the current version, because secondary sources (Staggering Strikes, Quick Sip, Tranquil Spirit, etc) are underestimated.

This removes the existing Stagger and Purifying Brew implementations, replacing the Stagger implementation with a new one that closely tracks ticks to prevent the analyzer's pool value drifting from the actual amount ticking out in the log.

Purifying Brew doesn't have new implementation yet, so this is a draft. I may not put the new UI for the guide in this PR because its already getting pretty large.

### Testing

image

- Quick Sip only: `report/NpA8mKBc76hJL3nf/44-Mythic+Fractillus+-+Kill+(1:07)/3-Eisenpelz/standard/overview`
- Tranquil Spirit: `/report/Hn9MRdgbpBfDmzqV/15-Mythic+Dimensius,+the+All-Devouring+-+Kill+(3:45)/2-Arim/standard/stagger`
- Tranquil + NR: https://www.warcraftlogs.com/reports/3XxdLQpYf1KGvFWZ?fight=14&type=summary&source=13&view=events&pins=0%24Off%24%23244F4B%24expression%24type+in+("staggerclear"%2C+"staggerprevented")^2%24Off%24%23909049%24expression%24ability.name+in+("Niuzao's+Resolve"%2C+"Expel+Harm"%2C+"Stagger")
- M+ log that led to the discovery of the `staggerclear` race condition: https://www.warcraftlogs.com/reports/V4ptm1Y9dnJckjhB?fight=15&type=damage-done&source=223

### TODOs

- [x] check for Awakened Spirit logging quirks
- [x] Purifying Brew analyzer


emallson:
image
new stagger section, using the new table code

PR #7624, opened 2026-02-03
[Survival] Sentinel Mark - Hero Talent

Kivlov:
### Description

Sentinel's Mark is an upgraded version of Spotter's Mark for Marksman that replaces it for Marksman, but gives the mechanic as a new mechanic to Survival.
On consuming Precise shots (MM) or Tip of the Spear (SV), you have a chance to apply sentinel's mark to a random target you're in combat with. It can be a low hp mob, a mob behind you. Generally pretty irritating.

It's consumed by hitting it with a bomb or an aimed shot and it makes it do more damage to the target with the mark. With hero talents, it also refunds aimed shot or wildfire cooldown.

The tricky part is because these can go up on a target, and then up on another target and you consume the 2nd one before the first one, I opted to collect the marks, and then resolve them as the analyser came across remove debuff events and then sort them after we've resolved everything. This seems to work reasonably well.

My dimmy log provides almost all cases, except the fight ending without it being consumed which is added in as an edge case for fights like Ras in Dawnbreaker where they just leave at 60%.

For analysis, since it's quite ambiguous what is a good or bad mark in some situations, I've opted to provide the HP the target was on when it applied, how long the debuff was up, and how many GCDs the player had between the ability that triggered the mark, and the one that removed it. For most people, I think the colour coding will be sufficient, and the tooltip values are just some extra value to be able to tell if you had a fair chance at consuming Mark before the target died or if it expired during some downtime mechanic.

- Test report URL: `/report/B23zGhvLk1AxKFw4/16-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:44)/114-Percival/standard`
- Screenshot(s):

### DIFF
--- a/src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
+++ b/src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
@@ -1,0 +1,460 @@
+import type { JSX } from 'react';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import Events, {
+  ApplyDebuffEvent,
+  RefreshDebuffEvent,
+  RemoveDebuffEvent,
+  DamageEvent,
+  HasRelatedEvent,
+  GetRelatedEvents,
+  GlobalCooldownEvent,
+} from 'parser/core/Events';
+import TALENTS from 'common/TALENTS/hunter';
+import SPELLS from 'common/SPELLS';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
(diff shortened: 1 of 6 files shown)


src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
```diff
@@ -0,0 +1,460 @@
...
+    if (outcome.type === 'consumed') {
+      return {
+        applyTimestamp: mark.applyTimestamp,
+        value: QualitativePerformance.Good,
+        tooltip: (
+          <>
+            <h5 style={{ color: GoodColor }}>GOOD: Consumed Mark</h5>
+            {baseTooltip}
+            <br />
+            Consumed @ <strong>{this.owner.formatTimestamp(outcome.timestamp)}</strong>
```
emallson, line 206: nit: use `<div>`/`<p>` instead of `<br/>` if possible

emallson (approved):

PR #7627, opened 2026-02-05
[Survival] Hero Talent Pack Leader - Stampede and Howl

Kivlov:
### Description

Adjusted Pack Leader APL now that sims are functional: Hatchet Toss is bad! So any cast of it is bad even with hogstrider. I've placed it below Raptor Strike, and it shouldn't show up if they dont cast it but it should show up as a 'never cast this ability' reminder if they do cast it.

Stampede is back, no longer a tier set unfortunately. It luckily can no longer double stampede, and is triggered from a new Stampede! buff that is applied when you use Takedown or Bestial Wrath and then Kill Command after. Simplified the analyser as well.

Howl of the Pack Leader - Boar Charge. This is currently the only useful or interesting Howl spawn to track. Wyvern is a damage buff for some time, and bear just spawns, deals a bleed, and then autos whatever you're attacking, there isn't any gameplay usage to it. Boar though, they recently buffed it by 200% and it hits like a truck and so making sure it actually hit something and didnt hit a wall, and making sure it actually hits a good amount of targets in aoe since it charges at whatever you target with the Kill Command that spawns it.

Boar also grants Hogstrider buff, which makes Hatchet Toss do more damage, and aoe and same for cobra shot for BM. However, they nerfed it from 700% down to 200% and it sucks now so we don't press it. If they ever buff it, this analyser will be extended to track the usage of hog strider since for survival you can sit on it for 20s to use it at the optimal moment. BM can also hold it for a little while but not as long.

and per last PR, cleaned up the sentinel mark <br /> tags.

- Test report URL: `/report/cN7DPVYJyHv6n1LB/26-Mythic+Fractillus+-+Kill+(1:24)/29-Samfar/standard`

### DIFF
--- a/src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
+++ b/src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
@@ -164,20 +164,13 @@ export default class SentinelsMark extends Analyzer.withDependencies({
     const gcdsBetweenApplyAndRemove = Math.max(mark.gcdCount - 1, 0);
 
     const baseTooltip = (
-      <>
+      <div>
         <div>
           Target: <strong>{mark.targetName}</strong>
-          {mark.hpPercent !== null && (
-            <>
-              <br />
-              HP: {mark.hpPercent.toFixed(1)}%
-            </>
-          )}
-          <br />
-          Duration: {(duration / 1000).toFixed(1)}s
-          <br />
-          GCDs Between Apply and Remove: {gcdsBetweenApplyAndRemove}
         </div>
+        {mark.hpPercent !== null && <div>HP: {mark.hpPercent.toFixed(1)}%</div>}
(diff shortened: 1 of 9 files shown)


src/analysis/retail/hunter/shared/herotalents/SentinelsMark.tsx
```diff
@@ -202,8 +197,9 @@ export default class SentinelsMark extends Analyzer.withDependencies({
           <>
             <h5 style={{ color: GoodColor }}>GOOD: Consumed Mark</h5>
             {baseTooltip}
-            <br />
-            Consumed @ <strong>{this.owner.formatTimestamp(outcome.timestamp)}</strong>
+            <div>
+              Consumed @ <strong>{this.owner.formatTimestamp(outcome.timestamp)}</strong>
+            </div>
```
emallson, line 202: 🫡

emallson (approved):

PR #7632, opened 2026-02-10
Prevent duplicate GCDs from getting triggered by normalized channels

Krealle:
### Description

This applies a global fix to `GlobalCooldown` that's already been implemented in a couple different spec analyzers.

Basically what's causing this is that when a channel gets normalized with the `Channeling` normalizer, the event sequence becomes `CastEvent` => `BeginChannelEvent`. Where as the "natural" sequence is `BeginCastEvent/BeginChannelEvent` => `CastEvent`.
This results in us trying to trigger duplicate GCD's for the same spell. This is already handled in `triggerGlobalCooldown`, but we still get debug annotations for trying.

image

I've gone through and removed the spec specific implementations for this since it's now being handled by the base `GlobalCooldown` module. Tested them all and they appear to still be working as intended. (Along with fixing the spells that weren't included in them)
I didn't touch the one for Mistweaver since they have a custom normalizer for handling Celestial Conduit channel fabrications, which I wasn't sure if it would break.

### Testing

Before
image

After
image

- Evoker: 
`/report/AmfBDLCGgRrF21Ph/45-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:34)/123-Vossopher/standard/debug`
- Shadow: 
`/report/AmfBDLCGgRrF21Ph/45-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:34)/103-Blessing/standard/debug`
- Devourer: 
`/report/AmfBDLCGgRrF21Ph/45-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:34)/3-Zahadius/standard/debug`
- Havoc: 
`/report/ZJyaVLcRTAWf1g87/22-Mythic+Nexus-King+Salhadaar+-+Kill+(1:51)/328-Draezeik/standard/debug`
- Survival: 
`/report/AmfBDLCGgRrF21Ph/45-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:34)/5-Fapandtrap/standard/debug`
- Windwalker: 
`/report/drHKpygjBZ8X21b6/32-Mythic+Plexus+Sentinel+-+Kill+(2:56)/28-Krexonaut/standard/debug`
- Classic Shadow:
`/report/qK4WNjgtQVYZcRdk/3-Heroic+Jin'rokh+the+Breaker+-+Kill+(1:49)/31-Barcode/standard/debug`


src/parser/shared/modules/GlobalCooldown.tsx
```diff
@@ -247,6 +256,20 @@ class GlobalCooldown extends Analyzer {
...
+        priority: -Infinity,
+        details: (
+          <div>
+            <dl>
+              <dt>Expected GCD Duration</dt>
+              <dd>{(event.duration / 1000).toFixed(2)}s</dd>
+            </dl>
+          </div>
+        ),
+      });
```
Krealle, line 272: This adds an annotation for the initial GCD triggered

src/parser/shared/modules/GlobalCooldown.tsx
```diff
@@ -76,7 +76,10 @@ class GlobalCooldown extends Analyzer {
    * If the channel of the cast was cancelled before it was finished (in the case of cast-time abilities, not channels), the GCD event will *not* be fired since it will reset upon cancel. We have no way of knowing *when* the cancel is (regardless if it's 100ms into the channel or 1400ms), but in most cases not triggering the entire GCD is enough.
    */
   onBeginChannel(event: BeginChannelEvent) {
-    this._currentChannel = event;
+    if (!event.trigger || event.trigger.type === EventType.BeginCast) {
+      // if the trigger is a cast event, then the beginchannel event came after the cast event, in which case we don't want to set the current channel to avoid getting false matches in the next `onCast` trigger
```
emallson: would it be clearer to explicitly check if the trigger type is `cast` rather than implicitly by elimination (assuming `event.trigger.type === EventType.BeginCast` is equivalent to `event.trigger.type !== EventType.Cast`?)

did a double take when i read the comment and then the condition

emallson (approved):

src/parser/shared/modules/GlobalCooldown.tsx
Krealle: Looking at the current implementations of `trigger` event being set it will always be either `CastEvent` or `BeginCastEvent` so they should be equivalent yea.

PR #7633, opened 2026-02-11
Update Embedded Timeline for easier usage

emallson:
### Description

This PR greatly simplifies the usage of the embedded timeline. Existing code is quite brittle and generally assumes a fixed layout with narrow ranges of possible durations shown.

As part of the ongoing work on cooldown/buff windows, I've added a `timeline` option to `CooldownExpandable` using this tech.

image

http://localhost:3000/report/NpA8mKBc76hJL3nf/11-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:10)/19-Setthh/standard

image

http://localhost:3000/report/NpA8mKBc76hJL3nf/11-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:10)/3-Eisenpelz/standard

I will be putting this up on https://next.wowanalyzer.com soon:tm: (probably tomorrow morning) so people can poke around.

This involved a pretty significant refactor of the Timeline internals in order to make things display in an auto-sizing (as opposed to fixed-size) container. Along the way, I fixed a few pieces of bit-rot in the main timeline itself, like the cooldown legend disappearing when you scroll past the first (horizontal) page. **This PR should have no functional changes to the main timeline** (aside from some spacing changes between timeline sections) so if you see any while poking around on `next`, please let me know.

The end result is that there is now an `<EmbeddedTimeline>` component that is usable without having to manage events / cooldowns / buffs yourself. You can render it manually like:

```tsx
<EmbeddedTimeline
  range={{ start, end }}
  cooldowns={[ SPELLS.BLACKOUT_KICK, SPELLS.BREATH_OF_FIRE_TALENT, ... ]}
  auras={[ SPELLS.SECRET_INFUSION_HASTE_BUFF, ... ]} 
  />
```

You can directly render one in a `CooldownExpandable` using the `timeline` prop (which just takes the `EmbeddedTimeline` props and passes them down).

## `StateHistory`

As part of this, I added a `StateHistory` tool to the repo that makes it easy(er) to access events based on timestamp after the analysis phase of processing, using binary search to limit the runtime complexity. This is a strategy that I used in [another project](https://git.emallson.net/?p=badlands.git;a=summary) and was quite happy with. This is used to provide access to slices of the event history quickly, without either (a) each component needing to re-scan the event list, or (b) requiring a top-level "scan once & split into parts" manager.

This is a class and not a function to leave the "sequential optimization" open in the future (we process events moving forward in time, so most of the time we'll be calling `getBefore(X)` then `getBefore(X + 100)`, etc and can use that to skip work by tracking the last timestamp used). However, getting that optimization right is not needed for this PR.

## Where Consumers?

I left the Invoke code out of this PR. it is in [this branch](https://github.com/emallson/WoWAnalyzer/compare/embedded-timeline-cooldown-expandable...emallson:WoWAnalyzer:embedded-timeline-cooldown-expandable-invoke). I'll be PRing the Invoke Niuzao parts for brewmaster, and removing the Invoke Yu'lon parts (unless the MW maintainers ask me to leave it in).


src/interface/report/Results/Timeline/Auras.tsx
```diff
@@ -0,0 +1,258 @@
...
+    this._applied[spellId] = event;
+    this._levels[level] = spellId;
+    this._maxLevel = Math.max(this._maxLevel, level);
+
+    return this.renderIcon(event, level, {
+      className: 'hoist',
+      children: <div className="time-indicator" />,
+    });
+  }
+  renderRemoveAura(event: AbilityEvent<EventType>, endsAfterFight = false) {
```
Krealle, line 131: similarly to `renderApplyAura` do we want to prevent rendering removals that occur before the start of the window?

src/interface/report/Results/Timeline/Auras.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  render() {
+    const { style } = this.props;
+
+    let combinedHistory = new StateHistory<AnyEvent>([]);
+
+    for (const auraId of this.aurasToRender()) {
+      const history = this.props.auras.history(auraId);
+      const slice = history.slice(this.props.start, this.props.end, true);
+      console.log(auraId, slice);
```
Krealle: leftover console.log

src/interface/report/Results/Timeline/Casts.tsx
```diff
@@ -138,7 +148,7 @@ const Casts = ({ start, windowStart, secondWidth, events, movement, ...others }:
     return (
       <Fragment
         // It's possible this complains about "encountered two children with the same key". This is probably caused by fabricating a channel event at a cast time. If you can fix it by removing one of the events that would be great, otherwise you may just have to ignore this as while it's showing a warning, deduplicting the icons is correct behavior.
-        key={`cast-${left}-${event.ability.guid}`}
+        key={`cast-${event.timestamp}-${event.ability.guid}-${className}`}
```
Krealle: not sure how relevant this comment is anymore, but we could append `event.type` to this key?

src/interface/report/Results/Timeline/EmbeddedTimeline.tsx
```diff
@@ -49,4 +61,156 @@ export const SpellTimeline = ({
...
+    end: number;
+  };
+  /**
+   * IDs of auras (buffs) to show on the timeline.
+   * If not provided or set to `true`, get the ids from the `Buffs` core analyzer.
+   * If set to `false`, do not show auras.
+   *
+   * At this time, ALL AURAS must be provided in your specs `Auras` / `Buffs` core analyzer. This is used to find the event data without re-scanning the entire fight.
+   */
+  auras?: (number | Spell)[] | boolean;
```
Krealle, line 167: would it make sense to add a debug annotation if you try to pass an aura that isn't provided in the core analyzer?

Krealle (commented):
Got a couple comments, but overall this is looking good!

Played around with your example implementation and It looks like a scroll bar won't show up if you increase the size range of the `CooldownExpandable` component. I think this is something that would make sense to be able to enable, potentially just being able to pass down a custom `secondWidth` arg like you can with `EmbeddedTimelineContainer`.

e.g. this is roughly how big the range would be for Devastation's Dragonrage.

Image

src/interface/report/Results/Timeline/Auras.tsx
emallson: ah, probably yea.

src/interface/report/Results/Timeline/EmbeddedTimeline.tsx
emallson: i think a `throw` in dev mode actually might make the most sense? we can tell which auras are in `Auras` even if they don't appear in the log, and throwing in dev seems the clearest way to signal to somebody "hey, you need to do X or this won't work"

emallson:
> Played around with your example implementation and It looks like a scroll bar won't show up if you increase the size range of the `CooldownExpandable` component. I think this is something that would make sense to be able to enable, potentially just being able to pass down a custom `secondWidth` arg like you can with `EmbeddedTimelineContainer`.

hmm, maybe a `minSecondWidth` that starts scrolling once things get too narrow? the issue with making it generally scrollable is that the autosizer can then start growing it infinitely and things break. but if things are too narrow the autosizer can be turned off in favor of a minimum width + scrollbar

src/interface/report/Results/Timeline/EmbeddedTimeline.tsx
Krealle: yea that does sounds better

PR #7636, opened 2026-02-16
[Enhancement] Quick guide fixes

Texleretour:
# Edits
- Fixed a few missing spaces in the guide
- Added some talent checks so that Lightning strikes doesn't fail checks when you aren't talented in it.
- Fixed PStorm GCD

There's still a problem with Thorim's Invocation as it expects all GCDs during the window to be Crash Lightning/Windstrike/Stormstrike, even when playing Totemic.
Thorim's invocation statistic doesn't work either on pre-patch logs, but works on earlier ones. I don't know why.
@Seriousnes 

# Testing
/report/Rzc6xtHmZL1JDGpr/29-Mythic+Fractillus+-+Kill+(1:25)/4-Aman/standard

### DIFF
--- a/src/analysis/retail/shaman/enhancement/modules/talents/PrimordialStorm.tsx
+++ b/src/analysis/retail/shaman/enhancement/modules/talents/PrimordialStorm.tsx
@@ -181,13 +181,13 @@ class PrimordialStorm extends MajorCooldown<PrimordialStormCast> {
       performance: lotfwActive ? QualitativePerformance.Perfect : QualitativePerformance.Fail,
       summary: (
         <>
-          <SpellLink spell={SPELLS.LIGHTNING_STRIKES_BUFF} /> {lotfwActive ? '' : 'not'}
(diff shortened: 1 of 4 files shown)


src/analysis/retail/shaman/enhancement/modules/talents/PrimordialStorm.tsx
Seriousnes, line 1: _Legacy of the Frost Witch_ has been removed, so the guide checks referencing it are fully out of date. _Lightning Strikes_ has interaction with _Primordial Storm_, so the correct fix here is to just remove anything referencing LotFW. There's no need to replace or add checks for _Lightning Strikes_

Seriousnes (changes requested):
Hi @Texleretour thanks for the PR. I've left one comment on the `PrimordialStorm.tsx` changes. You're right to call out the incorrect checks, I've not fully removed all _Legacy of the Frost Witch_ related checks, thanks for picking this up.

src/analysis/retail/shaman/enhancement/modules/talents/PrimordialStorm.tsx
Texleretour: Thanks for the review!

PR #7637, opened 2026-02-16
Add Throughput Table component

emallson:
### Description

This PR adds a `ThroughputTable` component to `interface`. This component allows creation of (optionally-whitelisted) Damage/Healing tables, without having to manually build up the tables yourself.

It also exposes the internals as a `Table` component, which allows others to build similarly-styled tables that show custom values. (I will be using this when I revisit my Stagger PR to show a table of Staggered damage, so if it isn't good enough for that yet, it will be soon. I didn't want to bloat this PR with that since it isn't critical-path work)

The table component can be seen on https://next.wowanalyzer.com right now on Brewmaster as part of the `CooldownGrid`:

image

https://next.wowanalyzer.com/report/KvHmTR2QCBY1c6yG/1-Mythic++Algeth'ar+Academy+-+Kill+(25:43)/4-Ace/standard/overview

image

https://next.wowanalyzer.com/report/bg42WrMFkR1Dvn9B/11-Mythic+Dimensius,+the+All-Devouring+-+Kill+(4:47)/20-Setthh/standard

The objective of the `ThroughputTable` itself is to enable maintainers to easily add damage/healing tables that cover interesting spans of time (like cooldowns) or interesting effects (like talents).

Of particular note: I left out the DPS/HPS column. My thinking there is that in one of the intended use-cases (short spans of time), the DPS/HPS values will not be particularly useful since they'll be highly inflated by the denominator (time) being small. However, it should be easy to add this in a follow-up if people disagree with me on that.

Another noteworthy change is that I replaced the CSS for spell schools and unit types in `Game.scss` with the values from WCL's CSS. The existing values (and the values in `MAGIC_SCHOOLS`) are taken from the game itself, and are very highly saturated, making them too intense for comfortable display on a table like this. I did not update `MAGIC_SCHOOLS` (yet?) but can.

### TODOs

- [x] Absorb healing (you'd think I'd have remembered this as a monk player)
- [ ] Exclude healing done to pets (maybe? not sure what healers think here)
- [x] ~~Handle support event reattribution~~
   - This turned out to be a much bigger task because we don't load the necessary events right now. It requires either a second query, or a WCL backend change. As a result, I punted it. We don't handle support events elsewhere anyway.

### DIFF
--- a/src/interface/guide/components/CooldownExpandable.tsx
+++ b/src/interface/guide/components/CooldownExpandable.tsx
@@ -14,15 +14,26 @@ export interface CooldownExpandableItem {
   details?: ReactNode;
 }
 
-interface Props {
+type RangeProps =
+  | { range?: undefined; timeline?: undefined; table?: undefined }
+  | {
+      range: { start: number; end: number };
+      /**
+       * Enable the embedded timeline. See `EmbeddedTimeline`. Accepts the props of the `EmbeddedTimeline` component. Use `timeline={true}` if you just want default props.
+       */
+      timeline?: Omit<EmbeddedTimelineProps, 'range'> | true;
+      /**
+       * Enable the embedded throughput table. See `ThroughputTable`. Accepts the props of the `ThroughputTable` component.
+       */
+      table?: Omit<ThroughputTableProps, 'range'>;
+    };
+
+type Props = {
   header: ReactNode;
   checklistItems?: CooldownExpandableItem[];
   detailItems?: CooldownExpandableItem[];
   perf?: QualitativePerformance;
-  range?: { start: number; end: number };
-  timeline?: Omit<EmbeddedTimelineProps, 'range'>;
-  table?: Omit<ThroughputTableProps, 'range'>;
-}
+} & RangeProps;
 
 /**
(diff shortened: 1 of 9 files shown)


emallson:
image

absorbs are in now (not on next bc branching shenanigans)

src/interface/guide/components/CooldownExpandable.tsx
```diff
@@ -50,8 +54,30 @@ export const CooldownExpandableDataList = ({
...
+  checklistItems,
+  detailItems,
+  perf,
+  timeline,
+  range,
+  table,
+}: Props) => {
   const [isExpanded, setIsExpanded] = useState(false);
+
+  // i don't like the `throw`, but the typing for this is very clunky in practice.
```
Krealle: could be good to add comments to the props/method so these conditional requirements are made clear prior to getting thrown an error

Krealle (approved):
Looks good!

> Exclude healing done to pets (maybe? not sure what healers think here)

Not a healer, but I think we should? Keeps it consistent with WCL, and pet healing is (usually) not important. Could also just make it an optional arg, not sure how much pet filtering is done in the various modules.

emallson:
image

excluded pets from healing done by default

PR #7643, opened 2026-02-22
Initial Pass to enable Midnight Affliction Warlock

katogaming88:
### Description

Enabling Affliction Warlock for midnight

Mostly deleting removed talents and adding Dark Harvest tracking.

Added Dark Harvest Talent to Cooldown tracker, but don’t know how to do the CDR on it with the talent Cull the Weak (https://www.wowhead.com/spell=1259886/cull-the-weak), as can be seen on the debug screen with the red dots under spellUsable. I do plan to come back to this and ask how to properly implement it, but would like to get Demonology also running before I start diving into fleshing everything out.

### Testing

- Test report URL: /report/1QDFd84qnzKARLPr/30-Mythic+Nexus-King+Salhadaar+-+Kill+(3:16)/1-Katorrí/standard/
- Screenshot(s): 
image

image


emallson:
> don’t know how to do the CDR on it with the talent Cull the Weak

the tl;dr is that you make an analyzer with `SpellUsable` as a dependency, then call `this.deps.spellUsable.reduceCooldown(DARK_HARVEST, 1500)` (i don't remember the exact params, so you'd need to double-check that).

that method returns the actual amount of CDR applied, which you can use to track CDR waste (if it is relevant)

src/analysis/retail/warlock/affliction/CHANGELOG.tsx
```diff
@@ -1,24 +1,25 @@
...
 import TALENTS from 'common/TALENTS/warlock';
-import { Arlie, Jonfanz, Lithix, Meldris, ToppleTheNun, dodse, Gazh } from 'CONTRIBUTORS';
+import { Arlie, Jonfanz, Lithix, Meldris, Topple, dodse, Gazh, Katorri} from 'CONTRIBUTORS';
 import { SpellLink } from 'interface';
 
 export default [
+  change(date(2026, 2, 21), "Initial Midnight update to activate Affliction", Katorri),
   change(date(2024, 10, 1), <>Add support for <SpellLink spell={SPELLS.DEMONIC_HEALTHSTONE} /> </>, Gazh),
-  change(date(2024, 9, 28), <>Fix <SpellLink spell={TALENTS.SHADOW_EMBRACE_TALENT} /> error </>, Gazh),
+  change(date(2024, 9, 28), <>Fix SHADOW_EMBRACE_TALENT error </>, Gazh),
```
emallson:
```suggestion
  change(date(2024, 9, 28), <>Fix Shadow Embrace error </>, Gazh),
```

if you want to keep the log entry, it should have the proper name for removed spells instead of the const name. generally we'd just delete all the old log entries. changes from 2023-2024 aren't super relevant in 2026

emallson (approved):
minor thing with the changelog, but this looks like a good start 👍

src/analysis/retail/warlock/affliction/CHANGELOG.tsx
katogaming88: ah ok, i'll just clear out the changelog then, since the spell is removed anyways, so it doesn't matter anymore.

PR #7644, opened 2026-02-22
Reskin CastSummary, CastDetail, & CastOverview

Sharrq:
Aesthetic rework of CastDetail, CastSummary, and CastOverview.

Current Live version
Screenshot_2026-02-22_01-15-22
Screenshot_2026-02-22_01-15-38

Reskinned Versions
Screenshot_2026-02-22_01-17-28
Screenshot_2026-02-22_01-17-43


src/interface/guide/components/GuideDivs.tsx
```diff
@@ -70,136 +70,131 @@ export const PanelHeader = styled.div`
...
+  font-size: 1.1rem;
   font-weight: 500;
-  color: rgba(255, 255, 255, 0.7);
+  color: rgba(255, 255, 255, 0.65);
   text-transform: uppercase;
   letter-spacing: 0.5px;
   text-align: left;
-  margin-bottom: 4px;
+  margin-bottom: 2px;
 `;
```
emallson: these are unused now. are they going to be part of future stuff? if not, maybe we cut them. unused stuff tends to bitrot.

emallson (commented):
i know this is still marked as WIP so i didn't go in too much detail looking at this, but I figured i could pull it up and take a look.

overall, this looks good 👍 this is cleaner and more refined than the previous PR (and I appreciate that you also made it a much smaller PR!)

i still have my issues with the big `GuideDivs` file. some parts of it I feel are clear if split into their own files, but I'm not going to nit that one too hard 

#### Text Contrast

Image

the "13" text in this needs more contrast with its background. could be either by reducing the saturation/lightness of the background, or by brightening the color used by the text in the foreground. i think this text color matches the one used elsewhere, so maybe the background is the proper adjustment.

the boxes with 0 of their type also should be a *little* brighter so that they're easier to read, or hidden entirely.

#### `CastCard` animation

I would prefer if at least the `translateY` part of the animation didn't replay every time. I find the little "bounce" distracting. Having only the opacity seems fine to me for a quick animation. (Not sure there's an easy way to have the translate happen initially then never again---but i'm not sure the animation is even visible initially)

#### `CastSummary` animation

this animation jumps on click because the helper text at the bottom appears/disappears instantly. i poked around and if you use a CSS `transition` then the "jump" is prevented and you get a smooth transition:

```diff
diff --git a/src/interface/guide/components/CastSummary.tsx b/src/interface/guide/components/CastSummary.tsx
index 5fd5981d7e..4cc4082721 100644
--- a/src/interface/guide/components/CastSummary.tsx
+++ b/src/interface/guide/components/CastSummary.tsx
@@ -160,11 +160,9 @@
               </BoxRowScaler>
             </BreakdownContainer>
           </ControlledExpandable>
-          {!isExpanded && (
-            <HelperText style={{ marginTop: '4px' }}>
-              Click the bar above for per-cast breakdown
-            </HelperText>
-          )}
+          <DisappearingHelperText style={{ height: isExpanded ? 0 : '1lh' }}>
+            Click the bar above for per-cast breakdown
+          </DisappearingHelperText>
         </>
       ) : (
         <BarContainer>
@@ -180,6 +178,12 @@
   );
 }
 
+const DisappearingHelperText = styled(HelperText)`
+  margin-top: 4px;
+  overflow-y: clip;
+  transition: height 0.5s; /* 0.5s matches react-animate-height */
+`;
+
 const BarContainer = styled.div`
   border-radius: 4px;
   overflow: hidden;
```

Sharrq:
Thanks. I havent started going through the changes to see what got duplicated yet (which is the biggest reason it is still WIP). 

For the grayed out Perfect/OK. I initially had them hidden but didnt really like it so i opted to gray out the ones where there are no results. I can make it a little more visible though. I think my issue with them being hidden is that it looks kinda weird when all the results are the same (like they are all Perfect), which kinda makes it seem less like a filter and more just like an additional info box, so i thought graying it out would read better.

emallson:
yea that makes sense. a little more contrast is prob the play then

Sharrq:
> the "13" text in this needs more contrast with its background. could be either by reducing the saturation/lightness of the background, or by brightening the color used by the text in the foreground. i think this text color matches the one used elsewhere, so maybe the background is the proper adjustment.

I made the disabled badges a little lighter so they are a little more visible and increased the contrast on the background. In both of these examples, the top is before and the bottom is after
Screenshot_2026-02-28_20-47-20
Screenshot_2026-02-28_20-47-43

Sharrq:
Reworked the CastDetail Navigation and Timeline a little bit
Screenshot_2026-02-28_23-33-15

Sharrq:
Forgot to include an example of CastSequence in this as well
Screenshot_2026-03-01_00-24-06

Sharrq:
This should be ready to be reviewed now.

emallson:
looking through on http://localhost:3000/report/KvHmTR2QCBY1c6yG/1-Mythic++Algeth'ar+Academy+-+Kill+(25:43)/2-Khaelt/standard

image

when there are more casts in the sequence than fit in the display, the overflowing ones aren't viewable (can't scroll). fixable by adding `overflow-x: auto` to the `Sequence` component, but maybe cast sequences shouldn't be that long and it should be left alone?

that text color still needs more contrast, but after toying around with it more i'm pretty sure that we just need to use a different color. i wouldn't worry about it for this PR.

overall the changes look good. let me know what you think about the cast sequence thing. if you think we should just say "that is too long for a cast sequence" then IMO we add a doc comment stating the intention there, and merge

Sharrq:
> looking through on http://localhost:3000/report/KvHmTR2QCBY1c6yG/1-Mythic++Algeth'ar+Academy+-+Kill+(25:43)/2-Khaelt/standard
> 
> image
> when there are more casts in the sequence than fit in the display, the overflowing ones aren't viewable (can't scroll). fixable by adding `overflow-x: auto` to the `Sequence` component, but maybe cast sequences shouldn't be that long and it should be left alone?
> 
> that text color still needs more contrast, but after toying around with it more i'm pretty sure that we just need to use a different color. i wouldn't worry about it for this PR.
> 
> overall the changes look good. let me know what you think about the cast sequence thing. if you think we should just say "that is too long for a cast sequence" then IMO we add a doc comment stating the intention there, and merge

It was definitely supposed to scroll, just all my examples werent long enough to show it so i didnt notice. I had a ScrollableContainer for it, but when the Sequence is being added as additionalContent for CastDetail, it doesnt have that scrollable container. So i moved the scrolling functionality into Sequence and replaced ScrollableContainer with the SectionContainer from GuideDataWrapper

emallson (approved):

PR #7648, opened 2026-03-02
Initial update for Midnight Demonology Warlock

katogaming88:
### Description

First pass at Midnight Demonology to enable spec. There are multiple parts of the spec that need way more "complex" changes due to the rework to Summon Demonic Tyrant. Namely, SummonDemonicTyrant.tsx needs to be switched from the information saying that Tyrant buffs the other demons, and now it's the other demons buff the Tyrant, and Tyrant no longer extends active pets, unless talented into Reign of Tyranny, but that talent only extends the duration of Dreadstalkers. I will take the time in the coming weeks to look into making this change.

I need to find a way to track the new Grimoires for Demonology as well, Grimoire: Imp Lord (spell ID 1276452) and Grimoire: Fel Ravager (spell ID 1276467). Currently, in Warcraftlogs, when a player uses one of the Grimoires, it shows as what the pet casts, not the actual pet that is summoned (maybe it is a different ID?), but in the damage breakdown of the player, it does show the Grimoire, so I'm not sure how to track that as a cooldown just yet. 

One thing I am going to look into is a tracker that looks at Implode efficiency with how many imps were imploded, but that is not crucial at the moment.

Aside from the minor issues that I need to fix above, it is in a working state with as close to the most up to date information as possible for now.

### Testing

- Test report URL: /report/WyvVJzPYBjMHQFC2/21-Mythic+Mug'Zee,+Heads+of+Security+-+Kill+(1:58)/1-Katorrí/standard
- Screenshot(s):


src/analysis/retail/warlock/demonology/modules/features/CooldownThroughputTracker.ts
```diff
@@ -1,20 +1,21 @@
...
-      spell: TALENTS.GRIMOIRE_FELGUARD_TALENT.id,
-      duration: 15,
-      summary: [BUILT_IN_SUMMARY_TYPES.DAMAGE],
-    },
+    // To do: change to Imp Lord and Fel Ravager
+    // {
+    //  spell: TALENTS.GRIMOIRE_FELGUARD_TALENT.id,
+    //  duration: 15,
+    //  summary: [BUILT_IN_SUMMARY_TYPES.DAMAGE],
+    // },
```
emallson, line 15: prefer deleting rather than commenting out, unless you're coming back to this immediately

src/analysis/retail/warlock/demonology/modules/features/Abilities.ts
```diff
@@ -173,34 +152,22 @@ class Abilities extends SharedAbilities {
...
+      //  gcd: {
+      //    base: 1500,
+      //  },
+      //  castEfficiency: {
+      //    suggestion: true,
+      //    recommendedEfficiency: 0.9,
+      //    averageIssueEfficiency: 0.8,
+      //    majorIssueEfficiency: 0.7,
+      //   },
+      // },
```
emallson, line 170: prefer deletion over commenting out

emallson (changes requested):
apologies for the delayed review. i'd request that you delete the code in these block comments instead of leaving them as commented-out code. commented-out code has a nasty tendency to stick around forever unless immediately dealt with.

if your next PR is going to deal with the commented-out code, then we can just merge. but otherwise the code should get deleted.

katogaming88:
so I commented out because I'm not as familiar with the way things are set up for wowanalyzer yet, so I didn't want to completely remove then not know how to add it back.

But I do intend to resolve most of the commented out code in my next PR, and if I can't figure out how to fix the things, then I will just delete.

emallson (approved):

PR #7658, opened 2026-03-08
Basic accessibility patch for Div like Buttons and Toggle Buttons

RussellFeinstein:
### Description

Added required semantic information to use a div as a button. Previously, Divs across the entire site were inaccessible to keyboard users. Screenreader users also had no semantic information.

Note that this is a band-aid fix. A more accessible design is a having the Arrow be the button and using native html.

### Testing

Go to any section of a report without this patch and click inside it to apply focus to the section.
Use Shift+Tab to tab backwards.
User will not be able to put focus on the Div title.

- Screenshot(s):
Before patch:
image
image

After patch:
image
Can press enter to interact with div:
image


emallson (approved):
good improvement 👍 

not sure what the normal way of doing a clickable header like this is. `<button>` + `<label for="...">`?

ToppleTheNun:
> good improvement 👍
> 
> not sure what the normal way of doing a clickable header like this is. `<button>` + `<label for="...">`?

it's the `details` tag but you style it!

emallson:
oh, duh. that makes sense

PR #7661, opened 2026-03-09
Created Demonic Tyrant Analyzer

katogaming88:
### Description

Adds a <b>Demonic Tyrant guide analyzer</b> for Demo Warlock.

This analyzer evaluates each <b>Summon Demonic Tyrant</b> window and provides feedback on how effectively the setup was executed.

The guide currently evaluates:
<ul>
<li> Number of <b>Hand of Gul'dan</b> casts </li>
<li> Number of <b>Wild Imps</b> summoned</li>
<li>Whether <b>Call Dreadstalkers</b> was active when Tyrant was cast</li>
<li>Whether <b>Call Dreadstalkers was cast too early</b> before Tyrant</li>
<li>Whether <b>Implosion</b> or <b>Power Siphon</b> were cast before Tyrant (which reduces demon count)</li>
</ul>

Each Tyrant cast is rated based on the number of <b>Hand of Gul'dan</b> casts during the Tyrant window, and a
<b>cast sequence view</b> shows the spells used around the Tyrant setup.

Also added logic to detect if <b>Call Dreadstalkers</b> was cast too far in advance before Tyrant and therefore expired too soon

### Testing

- Test report URL: 
`report/gnxHv1YBbTdzP3aC/9-Mythic+Gemellus+-+Kill+(2:04)/4-Katorrí/standard`
`report/gnxHv1YBbTdzP3aC/14-Mythic+Degentrius+-+Kill+(1:44)/4-Katorrí/standard`
- Screenshot(s):
image
image

### DIFF
--- a/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
+++ b/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
@@ -40,9 +40,9 @@ export default class DemonicTyrant extends Analyzer {
   onTyrantCast = (event: CastEvent) => {
     const timeSinceDreadstalkers = event.timestamp - this.lastDreadstalkersCast;
 
-    const dreadstalkersActive = timeSinceDreadstalkers <= 12000;
+    const dreadstalkersActive = timeSinceDreadstalkers <= DREADSTALKERS_DURATION;
 
-    const dreadstalkersTooEarly = dreadstalkersActive && timeSinceDreadstalkers > 7000;
+    const dreadstalkersTooEarly = dreadstalkersActive && timeSinceDreadstalkers > DREADSTALKERS_EARLY_CUTOFF_MS;
 
     const tyrant: TyrantCastData = {
       cast: event.timestamp,
--- a/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
+++ b/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
@@ -2,7 +2,6 @@ import SPELLS from 'common/SPELLS';
 import { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
 import Analyzer from 'parser/core/Analyzer';
 import Events, { CastEvent, SummonEvent } from 'parser/core/Events';
-import TALENTS from 'common/TALENTS/warlock';
 
 const DREADSTALKERS_DURATION = 12000;
 const TYRANT_WINDOW_MS = 20000;
@@ -19,7 +18,7 @@ export default class DemonicTyrant extends Analyzer {
     // Tyrant cast
     this.addEventListener(
       Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SUMMON_DEMONIC_TYRANT),
-      (event) => this.onTyrantCast(event),
+      this.onTyrantCast,
     );
 
     // Hand of Gul'dan cast
@@ -31,24 +30,13 @@ export default class DemonicTyrant extends Analyzer {
     // Wild imp summons
     this.addEventListener(
       Events.summon.by(SELECTED_PLAYER).spell(SPELLS.WILD_IMP_HOG_SUMMON),
-      (event) => this.onImpSummon(event),
+      this.onImpSummon,
     );
 
     // Dreadstalkers cast
     this.addEventListener(
       Events.cast.by(SELECTED_PLAYER).spell(SPELLS.CALL_DREADSTALKERS),
-      (event) => this.onDreadstalkersCast(event),
-    );
-
-    // Implosion cast
-    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(SPELLS.IMPLOSION_CAST), (event) =>
-      this.onImplosionCast(event),
-    );
-
-    // Power Siphon cast
-    this.addEventListener(
-      Events.cast.by(SELECTED_PLAYER).spell(TALENTS.POWER_SIPHON_TALENT),
-      (event) => this.onPowerSiphonCast(event),
+      this.onDreadstalkersCast,
     );
   }
 
@@ -65,8 +53,6 @@ export default class DemonicTyrant extends Analyzer {
       impsSummoned: 0,
       dreadstalkersActive,
       dreadstalkersTooEarly,
-      castedImplosion: false,
-      castedPowerSiphon: false,
     };
 
     this.tyrantData.push(tyrant);
@@ -98,16 +84,6 @@ export default class DemonicTyrant extends Analyzer {
   onDreadstalkersCast(event: CastEvent) {
     this.lastDreadstalkersCast = event.timestamp;
   }
-
-  onImplosionCast(event: CastEvent) {
-    if (!this.currentTyrant) return;
-    this.currentTyrant.castedImplosion = true;
-  }
-
-  onPowerSiphonCast(event: CastEvent) {
-    if (!this.currentTyrant) return;
-    this.currentTyrant.castedPowerSiphon = true;
-  }
 }
 
 export interface TyrantCastData {
@@ -116,6 +92,4 @@ export interface TyrantCastData {
   impsSummoned: number;
   dreadstalkersActive: boolean;
   dreadstalkersTooEarly: boolean;
-  castedImplosion?: boolean;
-  castedPowerSiphon?: boolean;
 }
--- a/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
+++ b/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
@@ -7,6 +7,8 @@ export default class DemonicTyrant extends Analyzer {
   tyrantData: TyrantCastData[] = [];
   currentTyrant: TyrantCastData | null = null;
 
+  lastDreadstalkersCast = 0;
+
   constructor(options: Options) {
     super(options);
 
@@ -28,13 +30,26 @@ export default class DemonicTyrant extends Analyzer {
       Events.summon.by(SELECTED_PLAYER).spell(SPELLS.WILD_IMP_HOG_SUMMON),
       (event) => this.onImpSummon(event),
     );
+    // Dreadstalkers cast
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.CALL_DREADSTALKERS),
+      (event) => this.onDreadstalkersCast(event),
+    );
   }
 
   onTyrantCast = (event: CastEvent) => {
+    const timeSinceDreadstalkers = event.timestamp - this.lastDreadstalkersCast;
+
+    const dreadstalkersActive = timeSinceDreadstalkers <= 12000;
+
+    const dreadstalkersTooEarly = dreadstalkersActive && timeSinceDreadstalkers > 7000;
+
     const tyrant: TyrantCastData = {
       cast: event.timestamp,
       handOfGuldanCasts: 0,
       impsSummoned: 0,
+      dreadstalkersActive,
+      dreadstalkersTooEarly,
     };
 
     this.tyrantData.push(tyrant);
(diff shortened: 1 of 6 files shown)


src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -0,0 +1,95 @@
...
+      (event) => this.onImpSummon(event),
+    );
+    // Dreadstalkers cast
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.CALL_DREADSTALKERS),
+      (event) => this.onDreadstalkersCast(event),
+    );
+  }
+
+  onTyrantCast = (event: CastEvent) => {
```
emallson:
```suggestion
  onTyrantCast(event: CastEvent) {
```

should use method syntax instead of property-assignment syntax. the difference only really comes into play in edge-cases, but the property-assignment syntax is inconsistent with the rest of the codebase and comes with some surprising footguns

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -0,0 +1,95 @@
...
+    const timeSinceDreadstalkers = event.timestamp - this.lastDreadstalkersCast;
+
+    const dreadstalkersActive = timeSinceDreadstalkers <= 12000;
+
+    const dreadstalkersTooEarly = dreadstalkersActive && timeSinceDreadstalkers > 7000;
```
emallson: rather than `12000` or `7000`, prefer named constants. maybe this is:

```suggestion
    const dreadstalkersActive = timeSinceDreadstalkers <= DREADSTALKERS_DURATION;

    const dreadstalkersTooEarly = dreadstalkersActive && timeSinceDreadstalkers > DREADSTALKERS_EARLY_CUTOFF_MS;
```

(you might consider simply using `DREADSTALKERS_DURATION / 2` for the latter, calling it waste if at least half of Dreadstalkers was over before Tyrant. however, I am not an expert on Demo and there may be reasons for the the 7s number)

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+      }
+      if (handOfGuldanCasts === 5) {
+        return QualitativePerformance.Good;
+      }
+      if (handOfGuldanCasts >= 3) {
+        return QualitativePerformance.Ok;
+      }
+
+      return QualitativePerformance.Fail;
+    }
```
emallson: functions that only use their parameters and file-level declarations should be declared outside the component. this prevents them from being re-defined each time the component renders

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    function getTyrantFeedback(
```
emallson: same note here (i think)

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+        );
+      }
+
+      const castedImplosion = preTyrantCasts.some(
+        (cast) => cast.spellId === SPELLS.IMPLOSION_CAST.id,
+      );
+
+      const castedPowerSiphon = preTyrantCasts.some(
+        (cast) => cast.spellId === TALENTS.POWER_SIPHON_TALENT.id,
+      );
```
emallson: ideally, these analysis pieces would be moved into the `Analyzer` and the guide file would remain focused on presenting the information. in some cases the lines are blurred, but each of these values for example could be tracked on `DemonicTyrant` using `lastCastPowerSiphon` / `lastCastImplosion` fields.

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+          'Entering Tyrant with more imps already active will significantly increase its damage.',
+        );
+      }
+      return (
+        <>
+          {feedback.map((line, i) => (
+            <div key={i}>{line}</div>
+          ))}
+        </>
+      );
```
emallson: prefer using `<ul>` / <`li>` for lists, rather than `<div />`s

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+          <li>
+            Imp Gang Bosses summoned by <SpellLink spell={SPELLS.IMPLOSION_CAST} /> or{' '}
+            <SpellLink spell={TALENTS.POWER_SIPHON_TALENT} /> with the talent{' '}
+            <SpellLink spell={TALENTS.TO_HELL_AND_BACK_TALENT} />
+          </li>
+        </ul>
+        <br />
+        During the Tyrant window, aim to cast as many{' '}
+        <SpellLink spell={SPELLS.HAND_OF_GULDAN_CAST} /> as possible to summon additional imps and
+        increase Tyrant's damage.
```
emallson: use `<p>` rather than a manual `<br/>`

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+  get guideSubsection(): JSX.Element {
+    const tyrant = <SpellLink spell={SPELLS.SUMMON_DEMONIC_TYRANT} />;
+
+    const explanation = (
+      <>
+        <b>{tyrant}</b> deals increased damage based on the number of active demons during its
+        duration. To maximize its effectiveness, summon as many pets as possible before and during
+        the Tyrant window.
+        <br />
+        <br />
```
emallson: same deal, `<p>` instead of manual breaks. the padding of the next line is probably fine if it is also wrapped in `<p>`

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+const TYRANT_POST_WINDOW = 20000;
+
+class DemonicTyrantGuide extends Analyzer {
+  static dependencies = {
+    demonicTyrant: DemonicTyrant,
+    eventHistory: EventHistory,
+  };
+
+  protected demonicTyrant!: DemonicTyrant;
+  protected eventHistory!: EventHistory;
```
emallson: this doesn't need to be an analyzer. it can simply be a react component. as a rule of thumb: if you aren't setting up event handlers, it shouldn't be an analyzer. you can rewrite this into a react component like:

```tsx
function DemonicTyrantGuide(): JSX.Element | null {
   const demonicTyrant = useAnalyzer(DemonicTyrant);
   const eventHistory = useAnalyzer(EventHistory);

  if (!demonicTyrant || !eventHistory) {
    return null;
  }

  // the contents of `get guideSection()` go here
}
```

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -0,0 +1,216 @@
...
+          icon: event.ability.abilityIcon.replace('.jpg', ''),
+        }));
+
+        return {
+          data: cast,
+          start: windowStart,
+          end: windowEnd,
+          casts,
+        };
+      });
```
emallson: the body of this `.map` uses `EventHistory.getEvents`, which is expensive. in order to avoid re-computing it on every render, this should be `useMemo`'d (which requires that this be converted from an `Analyzer` to a react component) or computed via event handler during processing.

the easier path for this is definitely converting to a react component.

```suggestion
    const tyrantSequenceEvents: CastSequenceEntry<TyrantCastData>[] = useMemo(() => {
     return demonicTyrant.tyrantData.map((cast) => {
        const windowStart = cast.cast - TYRANT_PRE_WINDOW;
        const windowEnd = cast.cast + TYRANT_POST_WINDOW;

        const castEvents = eventHistory.getEvents([EventType.Cast], {
          searchBackwards: false,
          startTimestamp: windowStart,
          duration: windowEnd - windowStart,
        });

        const casts: CastInSequence[] = castEvents.map((event) => ({
          timestamp: event.timestamp,
          spellId: event.ability.guid,
          spellName: event.ability.name,
          icon: event.ability.abilityIcon.replace('.jpg', ''),
        }));

        return {
          data: cast,
          start: windowStart,
          end: windowEnd,
          casts,
        };
      });
    }, [demonicTyrant, eventHistory]);
```

emallson (changes requested):
this is a good start 👍 i've requested a few changes. the bulk of the requested changes are organization / cleanup, though the organizational changes in `DemonicTyrantGuide` double as performance changes due to the assumptions that React makes about component code.

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
katogaming88: Ah ok, thanks. Yeah, I wasn't sure the best way to go about it. Basically the goal is "were the dreadstalkers cast too early" and my thought was slightly more than half the actual duration of dreadstalkers, but half the duration works too, not that big of a deal.

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -0,0 +1,121 @@
...
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SUMMON_DEMONIC_TYRANT),
+      (event) => this.onTyrantCast(event),
+    );
+
+    // Hand of Gul'dan cast
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.HAND_OF_GULDAN_CAST),
+      (event) => this.onHandOfGuldanCast(event),
+    );
```
emallson, line 28:
```suggestion
    this.addEventListener(
      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.HAND_OF_GULDAN_CAST),
      this.onHandOfGuldanCast,
    );
```

don't need to wrap these function refs in anonymous functions. this is mostly a style thing (no sneaky footguns), so not going to block a merge. just fyi for future

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -24,66 +28,86 @@ export default class DemonicTyrant extends Analyzer {
...
+
+  onImplosionCast(event: CastEvent) {
+    if (!this.currentTyrant) return;
+    this.currentTyrant.castedImplosion = true;
+  }
+
+  onPowerSiphonCast(event: CastEvent) {
+    if (!this.currentTyrant) return;
+    this.currentTyrant.castedPowerSiphon = true;
+  }
```
emallson: these values appear to be unused (i may be reviewing an in-progress part, apologies if so). i see the guide file is still using the `preCastEvents` data.

if you want to reproduce that logic here, you want to track the last observed power siphon / implosion *independently* of the current tyrant cast (rather than attached to it).

i know you've got a lot under review right now. since the guide now `useMemo`'s the values computed from event history, it is not *as* important to compute them in the analyzer. its only an organizational thing, rather than also impacting performance.

i believe you've addressed all the other comments. if you want, I can just cut these two properties/methods and merge the rest so you can focus on getting the other PRs in

katogaming88: > if you want, I can just cut these two properties/methods and merge the rest so you can focus on getting the other PRs in

Yeah, lets do that.

PR #7664, opened 2026-03-09
Add <main> Landmarks and Fix Heading Hierarchy

RussellFeinstein:
## Description
Two accessibility improvements across the interface layer.

### `<main>` Landmark
Top level page containers were plain `<div>`s, no Landmark region to navigate to.
- Converted to `<main>` on: CharacterParses, GuildReports, FightSelection, PlayerLoader, Results, UnsupportedSpecBouncer, MissingCombatantInfoBouncer, PatchChecker
 
### Heading Hierarchy
Duplicate `<h1>`s break document outline expectations.
- Demoted misused `<h1>`s to `<h2 className="h1">`s in Footer, ReportRaidBuffList, CharacterParses, GuildReports, and PremiumLoginPanel
- Replaced non-heading `<h1>`s with `<span>`s in ReportSelectionHeader (brand name) and PlayerTile (player names)
  - CSS updated to match
- Promoted `BossMiniBox` in Results to `<h1>` (no clear h1, either boss or player)
- SectionHeader and SubSection in the guide view now render `<h2>` and `<h3>` respectively
- Fixed heading order on Premium page
  - Passed `<h1>` to DOM first, then visually reordered with CSS

### Supporting Infrastructure
- Added `.h1` utility class to App.scss to decouple visuals from semantics
- Added a `subheading` prop to Panel that applies `.h1` to all subheadings

### Minor
- Added `title` prop to DocumentTitle on News page

## Testing
- Changes are primarily semantic
- Use HeadingsMap Extension to verify Heading Hierarchy
- Visual smoke test pages for regressions

### Screenshots
Player Selection Page:
image
image

Player Report:
image
image


emallson:
i want to spend a little time poking around before merging this so i prob won't get it in tonight. appreciate the work on a11y tho!

RussellFeinstein:
Oh no rush at all. Definitely worth making sure I didn't break anything else.

And yea, happy to do what I can!

emallson (changes requested):
i haven't done a full code review yet. but i pulled this up locally and there are a number of style changes as a result of this, including some that I'd consider breaking.

My general recommendation is that:

1. we can merge the `<main>` landmarks separately if they're split out. whether this is worth doing depends on how much work the next point is.
2. to fix the heading issue, we need to remove the baseline `<hN>` CSS from the site, replacing it with class-based styles where necessary to preserve layout.

I don't believe that trying to patch away the base `<hN>` styles at every location that uses a heading is sustainable long-term if the intention is to use the heading elements more for semantic/a11y reasons. The element-based styles are a holdover from the days of bootstrap CSS and are fine to be relocated to more narrowly-scoped rules.

To make that swap easier, we can add a top-level 

## Boss Header

### Before

Image

### After

Image

Note the change in text color, the sizable increase in line-height, the different font, and the all-caps style.

(font differences may not be visible depending on your system's available fonts)

## Section Headers

### Before

Image

### After

Image

Different font, and the sub-headings have a bottom border that is not properly padded away from the content.

## Player Selection

### Before

Image

### After

Image

i'm noting this as a difference on the principle that changes should be *intentional.* However, i think the *After* version of this is much more readable and would be a good change.

## Character Parses

### Before

Image

### After

Image

note the font difference in both the character name and "We couldn't find any logs" text. also note that the *After* version is missing padding on the sides. The missing padding is not a big deal---regaining that 30px of space is fine (especially if it was out of line with other pages, or other pages also get updated to regain that space).

RussellFeinstein:
Hey! Thanks for taking a look. I knew I would miss some spots, it was inevitable with what I was doing.

Agreed on all fronts. Honestly I was just trying to see if I could make the heading fix quickly as a band-aid without making an even broader change.

Happy to pull out the '<main>' change so that can get merged in on its own. I'd like to work on the actual modules for a bit now that I'm getting back to this project. After some time on Guardian and Balance (the two I'm working on), I'll circle back and work on pulling out the Headings entirely.

I'll close this and reopen with just the '<main>' changes.

emallson:
sounds reasonable to me 👍 thanks!

PR #7665, opened 2026-03-10
Update Affliction module with new analyzers and guides

katogaming88:
### Description

This PR rebuilds and improves several <b>Affliction Warlock </b> analyzers and guides, focusing on <b>Unstable Affliction (UA), Corruption/Wither, and Agony.</b> Key changes include

- Created Guide
  - Was previously just using FoundationGuide to get it functioning, now has a dedicated Guide to put new modules into.
- Unstable Affliction
  - Created new UA analyzer to track uptime, and give Qualitative Performance metric as well.
    - Currently, the QP metrics are very lenient, and will update it further once raids come out and actual uptime information can be gained
  - Updated UA spellID 
  - Added UA Uptime Guide
- Corruption/Wither
  - Recreated the Corruption analyzer and created Corruption guide to track uptime
  - Includes Qualitative Performance metrics
- Agony
  - Rebuilt Agony analyzer for clarity
  - includes Qualitative Performance metrics
- Cleanup
  - Restructured the Affliction folders to better align with more up to date specs
  - Removed outdated or redundant files to support the new structure
  - Rearranged files to better fit the new structure
- Resources
  - Copied Resources files from Demo into Aff
  - Added Aff back to Defensives guide
  - Added Defensives Guide to Guide

Plans in the future

- More robust UA analyzer to track more information (TBD)
- Update Darkglare information about extending dots (doesn't work that way)
- Add Dark Harvest analyzer
- Add reasoning to Dot Uptime trackers Qualitative Performance (currently just shows number)
- 

### Testing

- Test report URL: `/report/1QDFd84qnzKARLPr/4-Mythic+Plexus+Sentinel+-+Kill+(2:24)/1-Katorrí/standard`

- Screenshot(s):
<b>Dot Uptimes</b>

image

<b>Unstable Affliction section</b>

image

<b> Defensives Guide </b>

image

<b> Resource Graph </b>

image


src/analysis/retail/warlock/affliction/modules/analyzers/Corruption.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Determine which DoT was actually used
+  get activeDot() {
+    const witherUptime = this.enemies.getBuffUptime(SPELLS.WITHER_DEBUFF.id);
+
+    if (witherUptime > 0) {
+      return SPELLS.WITHER_DEBUFF;
+    }
+    return SPELLS.CORRUPTION_DEBUFF;
+  }
```
emallson: is this checkable based on a talent? or is this one of those annoying glyphs that changes the spell id?

if there is a talent for this, you can select the spell to use in the constructor:

```tsx
  super(options);

  this.activeDot = this.selectedCombatant.hasTalent(...) ? SPELLS.WITHER_DEBUFF : SPELLS.CORRUPTION_DEBUFF;
}
```

src/analysis/retail/warlock/affliction/modules/analyzers/Haunt.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  statistic() {
+    return (
+      <Statistic
+        category={STATISTIC_CATEGORY.TALENTS}
+        size="flexible"
+        tooltip={
+          <>
+            {formatThousands(this.bonusDmg)} bonus damage
+            <br />
```
emallson, line 103:
```suggestion
```

un-necessary break (i think?)

src/analysis/retail/warlock/affliction/modules/analyzers/Haunt.tsx
```diff
@@ -0,0 +1,187 @@
...
+          </>
+        }
+      >
+        <BoringSpellValueText spell={TALENTS.HAUNT_TALENT}>
+          {formatPercentage(this.uptime)} % <small>uptime</small>
+          <br />
+          {formatNumber(this.dps)} DPS{' '}
+          <small>
```
emallson: you can try formatting this with `<div>`s instead of `<br/>` but it may violate nesting rules. if it does, the `<br/>` is fine

src/analysis/retail/warlock/affliction/modules/analyzers/Haunt.tsx
```diff
@@ -0,0 +1,187 @@
...
+            damage amplification by <b>2%</b>.
+          </p>
+        )}
+
+        {this.selectedCombatant.hasTalent(TALENTS.SHADOW_OF_NATHREZA_3_AFFLICTION_TALENT) && (
+          <p>
+            <SpellLink spell={TALENTS.SHADOW_OF_NATHREZA_3_AFFLICTION_TALENT} /> increases Haunt's
+            damage amplification by <b>2%</b>.
+          </p>
+        )}
```
emallson: this is more suitable for a tooltip. you can wrap the actual damage bonus in on line 143 in `<TooltipElement>` to get a "this is hoverable" underline on it, and set this as the content of the tooltip, maybe with some adjustments like:

```tsx
<>
Haunt's damage bonus:
<ul>
  <li>+12% baseline</li>
  <li>+X% from Shadow of Nathreza</li>
</ul>
</>
```

we often use lists for this, but if you feel that the textual format is better that's fine!

src/analysis/retail/warlock/affliction/modules/analyzers/Haunt.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  get shadowOfNathrezaBonus() {
+    let bonus = 0;
+
+    if (this.selectedCombatant.hasTalent(TALENTS.SHADOW_OF_NATHREZA_2_AFFLICTION_TALENT)) {
+      bonus += 0.02;
+    }
+    if (this.selectedCombatant.hasTalent(TALENTS.SHADOW_OF_NATHREZA_3_AFFLICTION_TALENT)) {
+      bonus += 0.02;
+    }
```
emallson: this is wrong (i think) in an annoying way from how talents are generated and apex talents are laid out

Image

`SHADOW_OF_NATHREZA_2_AFFLICTION_TALENT` should (i believe) correspond to both tiers of the `Rank 2` value shown in this screenshot. so this should actually be:

```suggestion
      bonus += 0.02 * this.selectedCombatant.getTalentRank(TALENTS.SHADOW_OF_NATHREZA_2_AFFLICTION_TALENT);
    }
```

(having the `_3` talent means you must have both ranks of `_2`, so this is only wrong if you have exactly 2 points of `_2` but didn't take `_3` at all)

src/analysis/retail/warlock/affliction/modules/guide/UnstableAfflictionGuide.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  // Statistic bar at the top
+  statistic() {
+    return (
+      <StatisticBar wide position={STATISTIC_ORDER.CORE(1)}>
+        {this.unstableaffliction.subStatistic()}
+      </StatisticBar>
+    );
+  }
```
emallson: this should just be on the `UnstableAffliction` analyzer since the guide analyzer isn't adding anything to it

src/analysis/retail/warlock/affliction/modules/resources/SoulShardGraph.tsx
```diff
@@ -0,0 +1,65 @@
+import ResourceGraph from 'parser/shared/modules/ResourceGraph';
+import { formatTime } from 'parser/ui/BaseChart';
+import SoulShardTracker from './SoulShardTracker';
+
+class SoulShardGraph extends ResourceGraph {
```
emallson, line 5: is this duplicated with the one in the demo PR? if so, delete this one and move the demo one from `warlock/demonology` to `warlock/shared` and use it in both specs

src/analysis/retail/warlock/affliction/modules/guide/UnstableAfflictionGuide.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { STATISTIC_ORDER } from 'parser/ui/StatisticsListBox';
+import type { JSX } from 'react';
+import { SpellLink } from 'interface';
+import SPELLS from 'common/SPELLS';
+import UnstableAffliction from '../analyzers/UnstableAffliction';
+import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
+import { RoundedPanel } from 'interface/guide/components/GuideDivs';
+import { TALENTS_WARLOCK } from 'common/TALENTS';
+
+class UnstableAfflictionGuide extends Analyzer {
```
emallson: similar note as in the demo pr: this doesn't need to be an `Analyzer`. However, this one also doesn't need the memoization that the Demo one needed so its less important to address

src/analysis/retail/warlock/affliction/modules/analyzers/UnstableAfflictionEfficiency.ts
```diff
@@ -0,0 +1,75 @@
...
+    const targetID = event.targetID;
+
+    // Find the last cast on this target that wasn't wasted
+    const lastCast = [...this.casts]
+      .reverse()
+      .find((c) => c.cast.targetID === targetID && !c.wasted);
+
+    if (lastCast) {
+      lastCast.wasted = true;
+    }
```
emallson: so, two separate notes about this:

First: `removedebuff` happens when things die *as well as* when things naturally expire. marking every `removedebuff` as wasted means that a kill log should mark every cast as wasted since (typically) every enemy target in a kill log dies.

to address this, you could add an event link that ties `removedebuff` events to any `death` events that occur immediately after to the same target. then you can mark `removedebuff`s that are not tied to deaths as wasted, which will be *mostly correct* (boss immunity / RP phases often remove all debuffs, but that is harder to deal with in a generic way) 

Second: doing a `.reverse()` + `.find` through all events is very inefficient. We've had issues with similar code causing very long processing times on M+ logs in the past. What I would suggest is:

- use a map to obtain relevant casts like `castsByTarget: Map<string, UACastData[]>;`, using `encodeTargetString` / `encodeEventTargetString` as the map key
- when marking a cast as wasted, mark all casts in the map for that target as wasted *and then* reset that map entry to an empty array. something like:

```tsx
if (wasted) {
  const target = encodeEventTargetString(event)
  const casts = this.castsByTarget.get(target);
  if (!casts) {
    // potentially broken fight. just ignore
    return;
  }
  for (const cast of casts) {
    cast.wasted = true;
  }
  this.castsByTarget.set(target, []);
}
```

you can either simultaneously insert the same object into multiple places in `onUACast` (like `const cast = { ... }; this.casts.push(cast); this.castsByTarget.get(target).push(cast);`) or move them from `castsByTarget` to `this.casts` when they're removed. if you do the latter, then you also need to handle the `fightend` event to track non-wasted casts.

emallson (changes requested):
a few comments, but nothing too big. there are two potential correctness issues (the Shadow of Nathreza talent thing, and the UA `removedebuff` thing). the UA one is the most important thing to address

src/analysis/retail/warlock/affliction/modules/analyzers/Corruption.ts
katogaming88: This is checkable based on a talent. The Hellcaller Hero tree for Affliction and Destruction change Corruption/Immolate respectively to Wither

src/analysis/retail/warlock/affliction/modules/resources/SoulShardGraph.tsx
katogaming88: Ah ok, I understand. Yeah, i will look to do that, since Demo was originally the one with the Soul Shard Graph, but I've added it to Aff and destro, so i can look at just putting it in shared, and calling it in each guide file for each spec

katogaming88:
here are a couple other reports, which will show the Haunt bonuses working

`/report/gQbcDRCvm2dxfKBh/13-Mythic+Restless+Heart+-+Kill+(1:25)/3-Katorrí/standard`
`/report/78qVrHhYNw4WDjKJ/17-Mythic+Arcanotron+Custos+-+Kill+(1:12)/4-Katorrí/standard`

emallson (approved):

emallson:
thanks :)

PR #7680, opened 2026-03-15
Disc Priest - Protector of the Frail

danny-janse:
Added damage reduction statistics for disc priest talent `Protector of the Frail`

/report/9fHxMGzXcr18aTKh/6-Mythic+Imperator+Averzian+-+Kill+(6:40)/95-Robpri/standard/statistics
image

### DIFF
--- a/src/analysis/retail/priest/discipline/modules/spells/ProtectorOfTheFrail.tsx
+++ b/src/analysis/retail/priest/discipline/modules/spells/ProtectorOfTheFrail.tsx
@@ -1,0 +1,70 @@
+import { TALENTS_PRIEST } from 'common/TALENTS';
+import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
+import BoringSpellValueText from 'parser/ui/BoringSpellValueText';
+import Statistic from 'parser/ui/Statistic';
+import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
+import STATISTIC_ORDER from 'parser/ui/STATISTIC_ORDER';
+import SPELLS from 'common/SPELLS';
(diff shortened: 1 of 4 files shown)


src/analysis/retail/priest/discipline/modules/spells/ProtectorOfTheFrail.tsx
```diff
@@ -0,0 +1,70 @@
...
+        category={STATISTIC_CATEGORY.TALENTS}
+      >
+        <>
+          <BoringSpellValueText spell={TALENTS_PRIEST.PROTECTOR_OF_THE_FRAIL_TALENT}>
+            <SpellIcon
+              spell={TALENTS_PRIEST.PAIN_SUPPRESSION_TALENT}
+              style={{
+                height: '1.3em',
+                marginTop: '-1.em',
+              }}
```
emallson: couple minor issues here:

1. the `marginTop` doesn't actually work because `-1.em` is not valid css. needs to be `-1.0em` or `-1em`.
2. i'm not sure that this is needed? i believe this is just making the icon larger relative to the text. i'm ok including it but this is what it looks like without the `style` at all for ref:

Image

danny-janse: I copied this from blackout kick from the monk. But ur absolutelyright. not really required at all. in fact the -1em  misplaces the icon overall. Thanks for pointing it out.

emallson: ah, not surprised to see some ancient stuff in old monk statistics 🙈

emallson (approved):

PR #7684, opened 2026-03-16
Remove `@emotion/styled`, replace with css modules

Seriousnes:
### Description

* `@emotion/styled` package removed
* added CSS Modules for every use of styled components

### Testing

- Screenshot(s):

### DIFF
--- a/.serena/memories/project_overview.md
+++ b/.serena/memories/project_overview.md
@@ -1,0 +1,3 @@
+WoWAnalyzer is a client-side React/TypeScript app for analyzing World of Warcraft raid performance and surfacing gameplay metrics/suggestions.
+Repo root includes src/, scripts/, docs/, e2e/, public/, and packages/.
+Main app code lives under src/, with analysis code split by game version/spec under src/analysis/.
--- a/src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
+++ b/src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
@@ -26,6 +26,11 @@
   display: inline-flex;
   flex-direction: row;
   gap: 1px;
+
+  & > .coloredBlock {
+    height: 100%;
+  }
+
   max-height: 1.5em;
   height: 100%;
 }
--- a/src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
+++ b/src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
@@ -1,0 +1,37 @@
+.layout {
+  display: grid;
+  grid-template-columns: 45% 1fr;
+  gap: 2em;
+}
+
+.runicPowerSummary {
+  margin-top: 1em;
+}
+
+.suggestionsColumn {
+  display: flex;
+  flex-direction: column;
+  gap: 1em;
+}
+
+.rpTableContainer {
+  display: grid;
+  grid-template-columns: max-content max-content 1fr;
+  gap: 2px 0.5em;
+  align-items: center;
+  align-content: start;
+}
+
+.blockRow {
+  display: inline-flex;
+  flex-direction: row;
+  gap: 1px;
+  max-height: 1.5em;
+  height: 100%;
+}
+
+.coloredBlock {
+  display: inline-block;
+  box-sizing: content-box;
+  height: 100%;
+}
--- a/src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
+++ b/src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
@@ -1,5 +1,17 @@
-import styled from '@emotion/styled';
+import { forwardRef, type ComponentPropsWithoutRef } from 'react';
 
-export const ExplanationSection = styled.section`
-  margin-bottom: 20px;
-`;
+import styles from './CommonComponents.module.scss';
+
+export const ExplanationSection = forwardRef<HTMLElement, ComponentPropsWithoutRef<'section'>>(
+  ({ className, ...props }, ref) => (
+    <section
+      {...props}
+      ref={ref}
+      className={
+        className ? `${styles.explanationSection} ${className}` : styles.explanationSection
+      }
+    />
+  ),
+);
+
+ExplanationSection.displayName = 'ExplanationSection';
--- a/src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
+++ b/src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
@@ -1,17 +1,10 @@
+import { clsx } from 'clsx';
 import { forwardRef, type ComponentPropsWithoutRef } from 'react';
 
 import styles from './CommonComponents.module.scss';
 
 export const ExplanationSection = forwardRef<HTMLElement, ComponentPropsWithoutRef<'section'>>(
   ({ className, ...props }, ref) => (
-    <section
-      {...props}
-      ref={ref}
-      className={
-        className ? `${styles.explanationSection} ${className}` : styles.explanationSection
-      }
-    />
+    <section {...props} ref={ref} className={clsx(styles.explanationSection, className)} />
   ),
 );
-
-ExplanationSection.displayName = 'ExplanationSection';
--- a/src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
+++ b/src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
@@ -1,0 +1,34 @@
+.hitTimelineContainer {
+  display: grid;
+  grid-template-columns: calc(150px - 1rem) 1fr;
+  gap: 1rem;
+  height: 20px;
+  padding: 0 10px;
+  margin: 5px 0;
+
+  > :first-child {
+    justify-self: start;
+    align-self: start;
+    padding-left: 1rem;
+  }
+}
+
+.hitTimelineBar {
+  position: relative;
+  width: 100%;
+  height: 20px;
+}
+
+.hitTimelineSlice {
+  height: 100%;
+  position: absolute;
+  top: 0;
+  border: 1px solid black;
+  box-sizing: content-box;
+}
+
+.highlight {
+  background-color: var(--highlight-background-color);
+  color: var(--highlight-text-color);
+  padding: 0 3px;
+}
--- a/src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
+++ b/src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
@@ -6,7 +6,7 @@
   padding: 0 10px;
   margin: 5px 0;
 
-  > :first-child {
+  & > :first-child {
     justify-self: start;
     align-self: start;
     padding-left: 1rem;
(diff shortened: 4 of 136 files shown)


src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.rpTableContainer {
+  display: grid;
+  grid-template-columns: max-content max-content 1fr;
+  gap: 2px 0.5em;
+  align-items: center;
+  align-content: start;
+}
+
+.blockRow {
```
ToppleTheNun, line 25: this is missing the `& > .coloredBlock`

src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
```diff
@@ -1,5 +1,17 @@
...
-  margin-bottom: 20px;
-`;
+import styles from './CommonComponents.module.scss';
+
+export const ExplanationSection = forwardRef<HTMLElement, ComponentPropsWithoutRef<'section'>>(
+  ({ className, ...props }, ref) => (
+    <section
+      {...props}
+      ref={ref}
+      className={
```
ToppleTheNun: we can clean this up with clsx:
```tsx
className={clsx(styles.explanationSection, className)}
```

src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
```diff
@@ -0,0 +1,34 @@
+.hitTimelineContainer {
+  display: grid;
+  grid-template-columns: calc(150px - 1rem) 1fr;
+  gap: 1rem;
+  height: 20px;
+  padding: 0 10px;
+  margin: 5px 0;
+
+  > :first-child {
```
ToppleTheNun: missing the `&`

src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.tsx
```diff
@@ -123,27 +95,53 @@ function HitTimeline({ hits, showSourceName, unmitigatedContent }: HitTimelinePr
...
+      ...style,
+      '--highlight-background-color': color,
+      ...(textColor ? { '--highlight-text-color': textColor } : {}),
+    };
+
+    return (
+      <span
+        {...props}
+        ref={ref}
+        className={className ? `${styles.highlight} ${className}` : styles.highlight}
```
ToppleTheNun: clsx here too

src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.tsx
```diff
@@ -123,27 +95,53 @@ function HitTimeline({ hits, showSourceName, unmitigatedContent }: HitTimelinePr
...
-  background-color: ${(props) => props.color};
-  padding: 0 3px;
-  ${(props) => (props.textColor ? `color: ${props.textColor};` : '')}
-`;
+type HighlightProps = ComponentPropsWithoutRef<'span'> & {
+  color: string;
+  textColor?: string;
+};
+
+type HighlightStyle = CSSProperties & {
```
ToppleTheNun, line 121: unsure how I feel about using variables this way, but it beats having to have a single class per color (unless we start doing utility classes 😉)

src/analysis/retail/evoker/preservation/Guide.tsx
```diff
@@ -1,19 +1,23 @@
...
-  grid-column-gap: 0;
-  grid-template-columns: 154px 1fr;
-  align-items: center;
-`;
+export const GuideContainer = forwardRef<HTMLDivElement, ComponentPropsWithoutRef<'div'>>(
+  ({ className, ...props }, ref) => (
+    <div
+      {...props}
+      ref={ref}
+      className={className ? `${styles.guideContainer} ${className}` : styles.guideContainer}
```
ToppleTheNun: clsx here too

src/analysis/retail/monk/brewmaster/modules/core/StaggerPool/StaggerPoolSection.module.scss
```diff
@@ -0,0 +1,27 @@
+.sideBySide {
+  margin-top: 1.5rem;
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: 1.5rem;
+  align-items: start;
+
+  header {
```
ToppleTheNun: missing the `&`

src/analysis/retail/monk/brewmaster/modules/core/StaggerPool/StaggerPoolSection.module.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.summaryDl {
+  display: grid;
+  grid-template-columns: 1fr auto;
+  gap: 0.5rem 1rem;
+  width: max-content;
+  font-size: 95%;
+
+  dt {
```
ToppleTheNun: missing the `&`

src/analysis/retail/monk/brewmaster/modules/core/StaggerPool/StaggerPoolSection.module.scss
```diff
@@ -0,0 +1,27 @@
...
+    font-weight: normal;
+  }
+
+  dd {
```
ToppleTheNun: missing the `&`

ToppleTheNun (changes requested):
didn't get through all of it yet because I have to go to work but there are some common patterns/issues we can resolve

Seriousnes (commented):
n/a

src/analysis/retail/deathknight/blood/modules/spells/DeathStrike/DeathStrikeSection.module.scss
Seriousnes: Fixed

src/analysis/retail/demonhunter/shared/guide/CommonComponents.tsx
Seriousnes: Fixed

src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.module.scss
Seriousnes: Fixed

emallson: this does not appear to be fixed, at least in the version that has been pushed

emallson:
as a note: Github's review UI does *not* handle force-pushing after review comments are added, especially with large PRs. if there are any more changes to this before merge, please add commits rather than force-pushing. its going to get squashed into a single commit on merge anyway

.serena/memories/project_overview.md
emallson, line 1: i notice this folder got committed in https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7678.

i'm wary of committing the setup/config of *every* tool people use to the repo, because its hard to tell if it can ever be safely removed. there's a balance between utility and maintenance. if committing this folder was intentional, i'm not going to mandate its removal now. but in the future please ask first.

src/interface/report/Results/TimeFilter.module.scss
```diff
@@ -0,0 +1,29 @@
...
+    cursor: not-allowed;
+    color: #ccc;
+  }
+}
+
+.resetButton {
+  background: hsl(44, 7%, 10%);
+  border-color: #303030;
+  box-shadow: 0 1px 4px black;
+}
```
emallson, line 24: the variables from `interface/design-system` shouldn't have their values inlined like this. we have sass/scss in the repo, so we can move the definitions from `interface/design-system` into variables in a .scss file and `@use` it

```suggestion
.resetButton {
  background: design.$level1-background;
  border-color: design.$level2-border;
  box-shadow: design.$level1-shadow;
}
```

the intent is to improve readability (`0 1px 4px black` vs `design.$level1-shadow`) and to slightly discourage making tiny adjustments that result in an inconsistent UI

src/interface/CooldownGrid/CooldownGrid.tsx
```diff
@@ -205,18 +143,23 @@ function CooldownGridElementRaw({
   }
 
   return (
-    <CooldownGridElementContainer
+    <div
+      className={styles.elementContainer}
+      style={{
+        border: `1px solid ${design.level2.border}`,
+        background: design.level2.background,
+      }}
```
emallson, line 151: these shouldn't be inline styles

src/interface/CooldownGrid/CooldownGrid.tsx
```diff
@@ -225,17 +168,23 @@ function CooldownGridElementRaw({
         </tbody>
       </table>
       {renderContents && timeline && (
-        <CooldownGridTimelineContainer>
+        <div
+          className={styles.timelineContainer}
+          style={{
+            boxShadow: `inset ${design.level1.shadow}`,
+            background: design.level1.background,
+          }}
```
emallson, line 176: same here

src/analysis/retail/demonhunter/vengeance/guide/HitTimeline.tsx
emallson: these don't need to be CSS vars unless other consumers are overriding the styles instead of passing them as params. could write:

```tsx
style={{
  backgroundColor: color,
  color: textColor,
}}
```

and get the same effect *except* if there are overrides elsewhere

emallson (commented):
the overarching changes look reasonable to me. i'll put this up on `next` sometime tomorrow.

i have a few thoughts about structure, but that might be best left to a follow-up. to avoid trying to re-review this monster repeatedly, i'm going to limit my change requests. a lot of things can be addressed after the fact with judicious use of `grep` and `sed`. i'll defer to you on what is easy to have the llm adjust vs things i should just slam together a script for, so each of these is *non-blocking* for a merge.

- references to the `interface/design-system` values should use SCSS variables. `interface/design-system` should become an SCSS file.
   - as long as none of these are changed, this will be relatively easy to fix with grep/sed
- we should *probably* add a helper for defining these `forwardRef` elements. slightly-sassy name would be `styled`, but that is probably confusing to people coming in with experience elsewhere. `element(<elementName>, <className>) { return forwardRef(...); }` is probably reasonable. typescript should be able to fill in the type params for `forwardRef`
   - this will be relatively easy to fix with some find-references / codemod-style work
- *i had another one but its slipped my mind, i'll edit this later if it returns to me*

however, there are a number of display issues currently that *are* blocking a merge:

## Guide Section Backgrounds

### Production

Image

### This PR

Image

notice how it is both narrower and has the wrong background color. this happens on every guide section across each spec that i've looked at.

## Cooldown Grid Columns

### Production

Image

### This PR

Image

this appears to actually be an issue with the responsive timeline container. specifically:

the *old* implementation set a `width: ${width}px` css rule via emotion to a fixed value, and emotion then overwrote that entirely with the `width: 100%` rule from the responsive component. the *new* implementation sets a `width: 100%` css rule, and then a `width: ${width}px` `style` attribute which *cannot* be overridden by a css rule. using a CSS var (like `width: var(--timeline-width);` and then `style={{ '--timeline-width': `${width}px` }}`) would likely restore the old behavior.

## Fight View - Time Filter

### Production

Image

### This PR

Image

probably just needs some judicious use of `display: inline-block`

PR #7698, opened 2026-03-20
update food buffs for midnight

tyler-lutz:
### Description

Added missing food buffs for midnight.


emallson (approved):
haven't checked every id but i spot checked a few and this looks good 👍

PR #7712, opened 2026-03-22
windwalker: massive buffs overhaul

smvoss:
### Description

Add in-depth buff tracking for consuming our proc-based buffs: docj, bok!/combo breaker, and rushing wind kick.

Also fix talent for Crashing Fists to instead tick 6 times instead of 5 in the fists of fury clip analysis.

Also add zenith and rwk to mastery tracking.

### Testing

- Test report URL: `/report/a3MmGKFdyJ9ftbYX/10-Normal+Vorasius+-+Kill+(4:18)/40-Kholer/standard/overview`
- Screenshot(s):
image
image
image
image
image


src/analysis/retail/monk/windwalker/modules/core/AplProcWindowDetail.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    style={{
+                      display: 'inline-block',
+                      width: '18px',
+                      height: '10px',
+                      borderRadius: '3px',
+                      background: 'rgba(220,220,220,0.75)',
+                      marginRight: '8px',
+                      verticalAlign: 'middle',
+                    }}
+                  />{' '}
```
emallson: should probably pull this out as a (local) component with a `backgroundColor` prop rather than duplicating 3x.

src/analysis/retail/monk/windwalker/normalizers/ComboBreakerCastLinkNormalizer.ts
```diff
@@ -0,0 +1,42 @@
...
+    anyTarget: true,
+    maximumLinks: 1,
+  },
+];
+
+class ComboBreakerCastLinkNormalizer extends EventLinkNormalizer {
+  constructor(options: Options) {
+    super(options, EVENT_LINKS);
+  }
+}
```
emallson, line 36: because you're using the results of this normalizer in other normalizers, you should set a higher priority to guarantee it runs before the other normalizers that depend on it.


```suggestion
class ComboBreakerCastLinkNormalizer extends EventLinkNormalizer {
  priority = -100;
  constructor(options: Options) {
    super(options, EVENT_LINKS);
  }
}
```

src/analysis/retail/monk/windwalker/modules/core/aplProcWindows.ts
emallson, line 1: this file could use some top-level explanatory comments. what kind of proc windows? what is being tracked? does it work with other proc windows than the ones you designed it for?

it looks like this is showing APL state within proc windows (based on reading + the screenshots in the PR, which look nice). that's all fine, but you'll appreciate having the comments when you look at this in 6 months and do a gandalf.jpg

src/analysis/retail/monk/windwalker/modules/spells/FistsofFury.tsx
```diff
@@ -24,7 +23,10 @@ import { STATISTIC_ORDER } from 'parser/ui/StatisticBox';
 
 // Inspired by the penance bolt counter module from Discipline Priest
 
-const FISTS_OF_FURY_MINIMUM_TICK_TIME = 100; // This is to check that additional ticks aren't just hitting secondary targets
+// Fists of Fury only has 5 real tick moments per channel. On multi-target pulls the log can
+// interleave same-tick damage events across targets with >100ms gaps, so use a larger threshold
+// to collapse those into one tick while still separating the actual channel ticks.
+const FISTS_OF_FURY_MINIMUM_TICK_TIME = 500;
```
emallson: what % haste do you need to reach for this to break? remember that while static 100% haste is basically unreachable, short-term 100% haste is very attainable if ww is picking up haste on gear.

emallson (changes requested):
minor comments. the only one that is really blocking is actually the `priority` one because that one can sneakily break by other seeemingly-unrelated changes.

also: i like seeing more use of APL stuff outside the main apl section. I did some similar things with classic Arms War and with BDK for Bone Shield and was pretty happy with how both turned out. This looks nice

src/analysis/retail/monk/windwalker/modules/spells/FistsofFury.tsx
smvoss: i tightened this a bit and changed how im tracking, but i'll likely revisit this in the future and see if i can optimize this even further

PR #7715, opened 2026-03-22
RDruid Mana Efficiency and Cooldowns tab update

foobartles:
### Description
Update the mana efficiency tables and cooldown tab for resto druid. Removed all hardcoded mana values from rdruid for now since I did not see any value in hardcoding these. 

Updated base mana value for all healers.
Also updated the wowhead tooltips config to no longer link to Beta cooldowns.
### Testing

Verify that the new mana efficiency table is up to date (old table showed mana costs in the millions)
Verify that new cooldowns tab shows casts leading up to tranq and convoke casts for proper ramp analysis

- Test report URL:  https://www.warcraftlogs.com/reports/1qwDQb6vCHNMAjXm?fight=52&type=summary&source=1
- Screenshot(s):
image
image
image


src/common/TALENTS/druid.ts
Krealle, line 1: `TALENTS` files are auto-generated, so you shouldn't make changes to them as they'll be overridden the next time they are re-generated.

Krealle (changes requested):

src/common/TALENTS/druid.ts
foobartles: These mana costs are wrong though 🤔 Can we fix the generation to not include incorrect mana costs?

Krealle: think the resource values used in the talent regeneration needs to be updated as well, will take a look

foobartles: ah yeah https://github.com/WoWAnalyzer/WoWAnalyzer/blob/midnight/scripts/talents/generate-talents.ts#L43

need to update this too. I can do that in this PR

foobartles: Ok I updated these values but when running the script it was creating a massive diff in these talents files 🤔

Krealle: Yea when you generate the talents it will have wrong formatting. You can run `oxfmt src/common/TALENTS` to format them, then the only diff should be to the `manaCost` props.

emallson: it looks like the invocation of `oxfmt` in the `package.json` shortcuts is wrong. i'm fixing it and going to push a commit that regenerates talents after running it

emallson:
i'm not sure what the correct way to resolve the merge conflict in `Lifebloom.tsx` is. @foobartles will need to do that. the changes look reasonable tho

foobartles:
@emallson I fixed the merge conflicts but am seeing some unrelated pipeline failures? Not really sure how to address these

PR #7722, opened 2026-03-24
Add Ancestral Call (Mag'har Orc) racial support

MarchingCube:
### Description
When checking my own horrible log I've noticed that Mag'har orc racial is not detected correctly. Implemented something similar to regular Orc setup and it seems to work now.

### Testing

- Test report URL: /report/GAzkTHPbF2KQnpZm/33-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(9:33)/228-Scharfer
- Screenshot(s):
image

and after
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -27,12 +27,14 @@ import {
   Rex,
   squided,
   Gambyt,
+  MarchingCube,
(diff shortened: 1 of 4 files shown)


emallson (approved):
thanks!

PR #7725, opened 2026-03-24
[Core] Add Darkmoon Sigil: Hunt embellishment support

MarchingCube:
### Description
Tracks https://www.wowhead.com/item=245876/darkmoon-sigil-hunt in a similar way as  TWW ascension. Embellishment can proc several stat buffs based on creature type.

I've added stat tracking since it seems this embellishment was one source of wrong rune regen tracking for unholy DK (which I am trying to fix a bit).

In general I've tried to mostly copy TWW setup so if anything should be done differently now - let me know.
I'm linking to the craft spell because there is no single spell that maps cleanly here like in ascension case.

### Testing
- Test report URL: /report/GAzkTHPbF2KQnpZm/33-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(9:33)/228-Scharfer/standard/statistics
- Screenshot(s):
image
image

### DIFF
--- a/src/CHANGELOG.tsx
+++ b/src/CHANGELOG.tsx
@@ -34,6 +34,7 @@ import SpellLink from 'interface/SpellLink';
 
 // prettier-ignore
 export default [
+  change(date(2026, 3, 24), <>Add <SpellLink spell={ITEMS.DARKMOON_SIGIL_HUNT} /> embellishment analyzer with stat buff tracking.</>, MarchingCube),
   change(date(2026, 3, 24), "Add Ancestral Call (Mag'har Orc) racial support", MarchingCube),
   change(date(2026, 3, 23), "Update close kill times to match new WCL API", Putro),
   change(date(2026, 3, 20), 'Update FoodChecker for Midnight', Gambyt),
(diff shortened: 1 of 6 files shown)


emallson (approved):

Dambroda:
@MarchingCube I was making this change in a local  branch myself to fix similar cooldown issues I was seeing, and then when I went to rebase I saw you had already done this. Just to confirm something, I saw you used 56 (per the wowhead data) as the secondary stat buff, but I was seeing an 85 in-game tooltip. Did you also see that discrepancy? Wondering if I should update this to fix it.

MarchingCube:
> @MarchingCube I was making this change in a local branch myself to fix similar cooldown issues I was seeing, and then when I went to rebase I saw you had already done this. Just to confirm something, I saw you used 56 (per the wowhead data) as the secondary stat buff, but I was seeing an 85 in-game tooltip. Did you also see that discrepancy? Wondering if I should update this to fix it.

@Dambroda  I didn't notice until you mentioned this. Buffs on this embellishment are a big mystery but we should update to 85 to match. When you proc the buff, does it increase stats for 85 too?

Mostly because sometimes what Blizzard has in tooltips is not exactly what happens :)

PR #7726, opened 2026-03-24
[Unholy] Fix cooldown tracking and add Forbidden Knowledge spells

MarchingCube:
### Description
- Fix Anti-Magic Shell cooldown to account for Anti-Magic Barrier talent (40s vs 60s)
- Fix Death Charge to track 2 charges with Death's Echo talent
- Add Necrotic Coil and Graveyard spell definitions (Forbidden Knowledge replacements for Death Coil and Epidemic
  after Army of the Dead)
- Add Unholy Aura haste buff spell definition (from Magus of the Dead)

### Testing
- Test report URL: /report/GAzkTHPbF2KQnpZm/33-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(9:33)/228-Scharfer/standard
- Screenshot(s): 
image

image

Debug annotations are getting better:
image


danny-janse (approved):
lgtm

src/analysis/retail/deathknight/unholy/modules/Abilities.tsx
```diff
@@ -59,6 +59,24 @@ class Abilities extends CoreAbilities {
...
+      },
+      {
+        spell: SPELLS.GRAVEYARD.id,
+        enabled: combatant.hasTalent(TALENTS.FORBIDDEN_KNOWLEDGE_1_UNHOLY_TALENT),
+        category: SPELL_CATEGORY.ROTATIONAL_AOE,
+        gcd: {
+          base: 1500,
+        },
+        range: 20,
+      },
```
emallson: these should be added via `ExecuteHelper` implementations. you can see a really simple example of that here: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/b4580d6a6b2df149b9995243d7b0c1ebe9fe05e7/src/analysis/retail/monk/brewmaster/modules/spells/EmptyTheCellar.tsx#L14

using `ExecuteHelper` with an activation buff sets up cast efficiency / max casts data correctly.

a slightly more complicated one that tracks `maxCasts` for a spell that is usable multiple times per buff is Void Volley: https://github.com/WoWAnalyzer/WoWAnalyzer/blob/b4580d6a6b2df149b9995243d7b0c1ebe9fe05e7/src/analysis/retail/priest/shadow/modules/spells/VoidVolley.tsx#L19

MarchingCube: This is great, had no idea that this existed. Modified my code to use it and track casts within forbidden knowledge windows.

src/common/SPELLS/deathknight.ts
```diff
@@ -658,12 +705,6 @@ const spells = {
     icon: 'ability_maldraxxus_deathknight',
   },
 
-  ABOMINATION_LIMB_TICK: {
```
MarchingCube, line 661: Graveyard uses the same spell id. From what I see damage tick of limb was removed and it only ticks actual grips.

emallson (approved):

PR #7728, opened 2026-03-24
[Survival] Raptor Swipe Analyzer

Kivlov:
### Description

Adding Analyser for Raptor Swipe Apex. It is a massive misplay to fail to tip swipe as it both hits hard & when tipped causes a 300% strike as one hit so making a cast summary that makes it obvious if it wasn't tipped.

Still no changelog as I have a PR for Takedown (where I'll add the changelog) and then an AplCheck.tsx PR that will come on it's own after that.

- Test report URL: `/report/y6mgBPCHWG72bAx4/10-Heroic+Fallen-King+Salhadaar+-+Kill+(5:49)/203-Percival/standard`


Pewtro:
This looks to have chakram stuff in it too? Is that intentional

Kivlov:
> This looks to have chakram stuff in it too? Is that intentional

This was because I made this branch off the last branch for chakram fix so it should be the same commit

Pewtro (approved):

PR #7733, opened 2026-03-26
Add `WeightedPerformance` as alternative to `QualitativePerformance` + Elemental Shaman implementation

Seriousnes:
#### Elemental Shaman
Implements the below `WeightedPerformance` for `MaelstromSpenders` analysis.

---

#### Summary

Introduces `WeightedPerformance.ts` as an opt-in alternative to the existing
`getLowestPerf` / `getAveragePerf` helpers. Instead of collapsing multiple
checks into a single qualitative bucket, each check produces a continuous
**0.0–1.0 score** that is combined via a weighted average before being mapped
back to the existing `Perfect / Good / Ok / Fail` enum for display.

#### Motivation
The current helpers have two opposing failure modes:

- `getLowestPerf` - a minor mistake in a low-priority check drags the entire
  cast rating down to Fail, even if the important checks were fine.
- `getAveragePerf` - a critical mistake can be "averaged away" by several
  perfect checks, understating the severity.

Neither approach allows a module author to express that "wasting 30 Maelstrom
is 3× more important than missing a MoTE buff".

#### Design

**API** - `src/parser/ui/WeightedPerformance.ts`

**`scoreFromBreakpoints(actual, breakpoints)`**
Maps a raw value to a score using **linear interpolation** between breakpoints.
Clamps to the nearest boundary outside the defined range.
Breakpoints must be sorted ascending by `value`; direction (lower-is-better vs
higher-is-better) is encoded through the `score` direction.

```ts
// "lower is better" - resource waste
scoreFromBreakpoints(2, [
  { value: 0,  score: 1.0 },   // 0 waste  → perfect
  { value: 5,  score: 0.75 },  // 5 waste  → good edge
  { value: 25, score: 0.0 },   // 25 waste → fail
])
// actual=2 → interpolated: 1.0 + (2/5) * (0.75-1.0) = 0.90
```

```ts
// "higher is better" - uptime
scoreFromBreakpoints(0.82, [
  { value: 0.0,  score: 0.0 },
  { value: 0.75, score: 0.75 },
  { value: 1.0,  score: 1.0 },
])
// actual=0.82 → interpolated: 0.84
```

**`computeScore(actual, strategy)`**
Top-level dispatcher: accepts either a `ScoreBreakpoint[]` (delegates to
`scoreFromBreakpoints`) or a `ScoreFunction` — a plain
`(actual: number) => number` for custom mappings (exponential decay,
logarithmic scaling, etc.). Output is always clamped to [0, 1].

```ts
// Custom ScoreFunction — exponential decay for resource waste
computeScore(5, (waste) => Math.exp(-waste / 10))
// → ~0.607
```

**`getAggregatedScore(checks: ScoredCheck[])`**
Penalty-based weighted average: `1 - Σ(effectivePenalty × weight) / Σ(weight)`
where `effectivePenalty = 1 - score`. Without `maxPenalty` this is equivalent
to a standard weighted average of the scores. Only the ratio between weights
matters, not the absolute values.

`ScoredCheck.maxPenalty` (optional, 0.0–1.0) caps how much a single failing
check can drag the overall rating down — useful for secondary checks that
should influence but not dominate the result.

```ts
getAggregatedScore([
  { score: 0.90, weight: 3 },  // waste  (high importance)
  { score: 0.40, weight: 1 },  // MoTE   (lower importance)
])
// → 1 - ((0.10×3 + 0.60×1) / 4) = 0.775  →  Good

// With maxPenalty — cap a secondary check's influence
getAggregatedScore([
  { score: 0.90, weight: 3 },
  { score: 0.00, weight: 1, maxPenalty: 0.3 },  // even total failure = 0.3 penalty max
])
// → 1 - ((0.10×3 + 0.30×1) / 4) = 0.850  →  Good
```

**`scoreToQualitativePerformance(score, thresholds?)`**
Maps a 0–1 score to `QualitativePerformance`.
Default thresholds (overridable per call):

```
≥ 0.95  →  Perfect
≥ 0.75  →  Good
≥ 0.50  →  Ok
 < 0.50  →  Fail
```

### Testing

- Screenshot(s):

### DIFF
--- a/src/analysis/retail/shaman/shared/Abilities.tsx
+++ b/src/analysis/retail/shaman/shared/Abilities.tsx
@@ -177,9 +177,7 @@ class Abilities extends CoreAbilities {
         enabled: combatant.hasTalent(TALENTS.SPIRITWALKERS_GRACE_TALENT),
         category: SPELL_CATEGORY.UTILITY,
         cooldown: combatant.hasTalent(TALENTS.GRACEFUL_SPIRIT_TALENT) ? 90 : 120,
-        gcd: {
-          static: 0,
-        },
+        gcd: null,
       },
       {
         spell: TALENTS.PURGE_TALENT.id,
--- a/src/analysis/retail/shaman/shared/Abilities.tsx
+++ b/src/analysis/retail/shaman/shared/Abilities.tsx
@@ -177,7 +177,9 @@ class Abilities extends CoreAbilities {
         enabled: combatant.hasTalent(TALENTS.SPIRITWALKERS_GRACE_TALENT),
         category: SPELL_CATEGORY.UTILITY,
         cooldown: combatant.hasTalent(TALENTS.GRACEFUL_SPIRIT_TALENT) ? 90 : 120,
-        gcd: null,
+        gcd: {
+          static: 0,
+        },
       },
       {
         spell: TALENTS.PURGE_TALENT.id,
--- a/src/parser/ui/WeightedPerformance.ts
+++ b/src/parser/ui/WeightedPerformance.ts
@@ -19,19 +19,19 @@ export interface ScoreBreakpoint {
 }
 
 /**
- * A single scored check with a relative importance weighting and optional penalty cap.
+ * A single scored check with a relative importance weighting and optional score floor.
  *
  * `score` should be in [0.0, 1.0].
  * `weight` is a relative importance factor — only the ratio between weights matters.
- * `maxPenalty` (optional, 0.0–1.0) caps how much penalty this check can contribute.
- *   Without it, a score of 0.0 contributes a full penalty of 1.0.
- *   With maxPenalty=0.3, even a score of 0.0 contributes at most 0.3 penalty.
+ * `minScore` (optional, 0.0–1.0) floors the effective score this check contributes.
+ *   Without it, a score of 0.0 contributes 0.0 to the weighted average.
+ *   With minScore=0.7, even a score of 0.0 contributes at least 0.7.
  *   This allows secondary checks to influence the result without dominating it.
  */
 export interface ScoredCheck {
   score: number;
   weight: number;
-  maxPenalty?: number;
+  minScore?: number;
 }
 
 /**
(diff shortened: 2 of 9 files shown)


src/analysis/retail/shaman/shared/Abilities.tsx
```diff
@@ -177,7 +177,9 @@ class Abilities extends CoreAbilities {
         enabled: combatant.hasTalent(TALENTS.SPIRITWALKERS_GRACE_TALENT),
         category: SPELL_CATEGORY.UTILITY,
         cooldown: combatant.hasTalent(TALENTS.GRACEFUL_SPIRIT_TALENT) ? 90 : 120,
-        gcd: null,
+        gcd: {
+          static: 0,
+        },
```
emallson: is this not equivalent to `gcd: null`?

src/parser/ui/WeightedPerformance.ts
```diff
@@ -0,0 +1,181 @@
...
+ * `maxPenalty` (optional, 0.0–1.0) caps how much penalty this check can contribute.
+ *   Without it, a score of 0.0 contributes a full penalty of 1.0.
+ *   With maxPenalty=0.3, even a score of 0.0 contributes at most 0.3 penalty.
+ *   This allows secondary checks to influence the result without dominating it.
+ */
+export interface ScoredCheck {
+  score: number;
+  weight: number;
+  maxPenalty?: number;
+}
```
emallson: overall, i think this API looks good. its flexible, but doesn't look too complicated to use.

the one major gripe i have with the API itself is the use of both `score` and `penalty`. negation trips people up often, and double negation doubly so. the actual aggregation function is equivalent to:

```
Sum_i [ max(1 - maxPenalty, score_i) * weight_i ] / Sum_i [ weight_i ]
```

so introducing the negated value doesn't seem good to me. simpler to keep everything on the same scale and rename `maxPenalty` to `minScore`

src/parser/ui/WeightedPerformance.ts
```diff
@@ -0,0 +1,181 @@
...
+ * Breakpoints must be sorted ascending by `value`. Values outside the breakpoint
+ * range are clamped to the nearest boundary's score.
+ */
+export function scoreFromBreakpoints(actual: number, breakpoints: ScoreBreakpoint[]): number {
+  if (breakpoints.length === 0) {
+    return 0;
+  }
+  if (breakpoints.length === 1) {
+    return Math.max(0, Math.min(1, breakpoints[0].score));
+  }
```
emallson, line 101: these should throw rather than producing junk. obvious errors are preferable to subtle ones

emallson (changes requested):
some api design comments but otherwise this is mergeable

src/parser/ui/WeightedPerformance.ts
Seriousnes: Done

src/parser/ui/WeightedPerformance.ts
Seriousnes: Done

src/analysis/retail/shaman/shared/Abilities.tsx
Seriousnes: It is, I'll revert this

emallson (approved):

PR #7736, opened 2026-03-26
[Holy Priest] Prayer of Healing & Lightweaver section update

Amiphite:
- **Update Prayer of Healing** 
  - Remove dependency on `SpellUsable` and checks for Holy Word: Sanctify cooldown
  - Remove unused `prayerCircle`, add `SurgeOfLight` instead
  - Format `TALENTS_PRIEST` to `TALENTS` 
  - Add conditional `if/else` logic to the guide to adapt suggestions if not talented into Surge of Light
  - Minor visual guide update

- **Update Lightweaver section**
  - Remove complex calculateHealing method and separated healing source tracking
  - Simplify guide subsection to focus on Flash Heal usage and stack management
  - `Heal` was removed and replaced by `PrayerOfHealing` where it was used
  - Minor visual guide update
  
 ###  **Testing**
report/LTCMtPfKFHBz7Gh1/23-Heroic+Imperator+Averzian+-+Wipe+15+(4:48)/15-Amiphite/standard
report/VKRBjcw1dG6AanqJ/37-Mythic+Imperator+Averzian+-+Kill+(6:53)/48-Smörflugan/standard
report/8cNrzxvFAHh3GkQW/36-Mythic+Imperator+Averzian+-+Kill+(7:04)/8-芝士熊师傅/standard

**New with Lightweaver talent:**
image

**New without Lightweaver talent:**
image

**New without Spiritwell talent:**
image

**Old:**
image


Amiphite:
Sorry for too many commits after PR. Found out about augmentation styling and wanted to have similar visuals =)

src/analysis/retail/priest/holy/modules/Styling.scss
```diff
@@ -0,0 +1,101 @@
+@use 'interface/Theme.scss';
+@use 'interface/guide/Guide';
```
emallson: these `@use` calls import variables into the file. since you're not using variables currently, they aren't needed

src/analysis/retail/priest/holy/modules/Styling.scss
emallson, line 1: the way this CSS is set up, it will be applied *globally* as soon as you load a holy priest log (which will trigger loading the CSS)

you should:

- change this from `Styling.scss` -> `Styling.module.scss`
- import the styles from it in the places you want to use them (`import styles from '../Styling.module.scss'`)

i only see a few of these classes (`.goodCast` / `.badCast` / `.perfectCast`) used. any that are un-used can be deleted. the others can be referenced via SCSS module import.

if you *do* need to override global css, use a class on the container, and nest a `:global` selector like this:

```scss
.gridContainer {
  :global(.grid) {
    // ...
  }
}
```

then import and use the `.gridContainer` class on a container. this will scope the change only to where you intend it to occur

emallson (changes requested):

emallson (approved):

PR #7740, opened 2026-03-28
Destruction: Update Backdraft analyzer

katogaming88:
## 1. Backdraft Analyzer & Guide

- Updated the original **Backdraft** analyzer (`modules/analyzers/Backdraft.tsx`) that tracks:
  - Backdraft stacks consumed per spell (**Incinerate**, **Chaos Bolt**, **Soul Fire**)
  - Wasted stacks from **overcapping** and **expiration**
  - Per-cast **SpellUse** performance entries

- Created a **BackdraftGuide component** (`modules/guide/BackdraftGuide.tsx`) with a spell usage breakdown.

- Removed the old **Flashover module** entirely (talent was removed for dragonflight)

---

### 2. Guide Restructure

- Added a new top-level **Rotation** section to the Destruction guide.
- Moved **HavocGuide** into the Rotation section (previously a standalone top-level section).
- Added **BackdraftGuide** as a subsection under Rotation when the talent is active.

---

### 3. Folder Restructure

- Split `modules/features/` into:
  - `modules/core/`
  - `modules/analyzers/`

- Moved Destruction core modules (Abilities, ABC, CooldownThroughputTracker) into **core/**
- Moved all talent/feature analyzers into **analyzers/**
- Updated imports in `CombatLogParser.ts` and guide files accordingly

### Testing

- Test report URL: 
Backdraft only: `/report/CQgrp23KT4kBFDAq/16-Heroic+Lightblinded+Vanguard+-+Kill+(5:17)/2-Katorrí/standard`
Rotation section with havoc: `/report/CQgrp23KT4kBFDAq/11-Heroic+Vaelgor+&+Ezzorak+-+Kill+(6:40)/73-Speakntongue/standard`

- Screenshot(s):
**Backdraft**

image
image
image

**Overall views**
image
image


src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+      this.wastedOvercapStacks += overcapped;
+      this._currentStacks = this._maxStacks;
+
+      this.uses.push({
+        event: event as unknown as CastEvent,
+        performance: QualitativePerformance.Fail,
+        checklistItems: [],
+        performanceExplanation: `Overcapped Backdraft by ${overcapped} stack${overcapped > 1 ? 's' : ''}`,
+      });
+    } else {
```
emallson: this condition will never trigger. you need to check for `refreshbuff` events that occur while at 2 stacks. [example](https://www.warcraftlogs.com/reports/b61z94vk7w2FrTna?fight=24&source=17&type=auras&ability=117828&view=events&start=8731535&end=8770877)

src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+    if (spellId === SPELLS.CHAOS_BOLT.id || spellId === TALENTS.SOUL_FIRE_TALENT.id) {
+      performance = QualitativePerformance.Good;
+    }
+
+    let spellName = 'Unknown Spell';
+
+    if (spellId === SPELLS.CHAOS_BOLT.id) spellName = 'Chaos Bolt';
+    if (spellId === SPELLS.INCINERATE.id) spellName = 'Incinerate';
+    if (spellId === TALENTS.SOUL_FIRE_TALENT.id) spellName = 'Soul Fire';
```
emallson:
```suggestion
    const spellName = SPELLS[spellId]?.name ?? 'Unknown Spell';
```

src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+      event,
+      performance,
+      checklistItems: [],
+      performanceExplanation: `Consumed Backdraft with ${spellName}`,
+    });
+  }
+
+  getSpellUsesWithPotentialMisses(fightStart: number, fightEnd: number): SpellUse[] {
+    return this.uses
+      .filter((use) => use.event.timestamp >= fightStart && use.event.timestamp <= fightEnd)
```
emallson: what problem does this `.filter` solve? the events being processed are already within the fight boundaries. if the goal is simply to be defensive against this case, the filter should be removed. many things will break in much worse ways if that occurs.

the sort i can understand (it is possible to add `uses` that are out of order), though i'm unsure if the sort is needed in practice here.

src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+  _maxStacks = 2;
+  _currentStacks = 0;
+
+  wastedOvercapStacks = 0;
+  wastedExpiredStacks = 0;
+
+  _buffedChaosBoltCasts = 0;
+  _buffedIncinerateCasts = 0;
+  _buffedSoulFireCasts = 0;
```
emallson: use `private` instead of *or* in addition to the underscores


```suggestion
  private _maxStacks = 2;
  private _currentStacks = 0;

  wastedOvercapStacks = 0;
  wastedExpiredStacks = 0;

  private _buffedChaosBoltCasts = 0;
  private _buffedIncinerateCasts = 0;
  private _buffedSoulFireCasts = 0;
```

src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(SPELLS.CHAOS_BOLT),
+      this.onChaosBoltCast,
+    );
+    this.addEventListener(
+      Events.cast.by(SELECTED_PLAYER).spell(TALENTS.SOUL_FIRE_TALENT),
+      this.onSoulFireCast,
+    );
+
+    this.addEventListener(Events.cast.by(SELECTED_PLAYER), this.onCast);
```
emallson: you can pass multiple spells to a filter to run the listener on specific spells. this allows removing the spell id check from the handler itself.


```suggestion
    this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell([
      SPELLS.INCINERATE,
      SPELLS.CHAOS_BOLT,
      TALENTS.SOUL_FIRE_TALENT
    ]), this.onCast);
```

src/analysis/retail/warlock/destruction/modules/analyzers/Backdraft.tsx
```diff
@@ -0,0 +1,286 @@
...
+    if (this._currentStacks > 0) {
+      this.wastedExpiredStacks += this._currentStacks;
+
+      this.uses.push({
+        event: event as unknown as CastEvent,
+        performance: QualitativePerformance.Fail,
+        checklistItems: [],
+        performanceExplanation: `Backdraft expired with ${this._currentStacks} stack${this._currentStacks > 1 ? 's' : ''} remaining`,
+      });
+    }
```
emallson: this check is incorrect and causes double-counting of backdraft consumption specifically when you go to 2 stacks of backdraft and then spend back down to 0. i hacked in timestamps in the tooltips to verify locally:

Image

Image

[sample log](https://www.warcraftlogs.com/reports/b61z94vk7w2FrTna?fight=30&source=17&type=auras&ability=117828&view=events&start=11985763&end=11991004)

tracking expiration in general is hard. the most robust way we have is to use an `EventLinkNormalizer` to connect the cast event of spells that consume the buff to a `removebuff` or `removebuffstack` that happens in a short (typically <100ms) window after.

if you don't want to dive into that for this PR, a slightly less robust but also far less complicated way is to track the timestamp of the most recent cast that could consume a buff, and then check if `event.timestamp - this.previousCast.timestamp > BUFF_REMOVAL_THRESHOLD` in this handler. if so, assume it was wasted. if not, assume it wasn't. this will break in a few niche cases (like if you die right after completing a cast), but will *mostly* work.

emallson (changes requested):
a couple of style comments, but those aren't as important. the main thing to fix before we can merge this is the correctness problems around tracking Backdraft state: currently it will never report an overcap, and it will always report a wasted buff due to expiration if you go to 2 stacks and then spend down to 0.

katogaming88:
sorry, i rebased before i saw your message, but hopefully this resolved the issues

emallson:
im going to re-rebase this so that the diff shows the correct changes, for my sanity. its probably fine, but with such a large diff it is easy to miss accidentally-reverted changes.

emallson (approved):

emallson:
@katogaming88 you'll need to re-merge the midnight branch. sorry about that. i was going to push a merge to it after but it won't let me because the PR is closed

PR #7742, opened 2026-03-29
[Mistweaver] Improve Celestial analysis + other CD updates

lolswirl:
### Description

- Implemented a Soothing Mist normalizer since the channel doesn't have an actual end channel event. This was fine previously, but with target swapping being added baseline, the Soothing Mist removal and subsequent apply events that occur when swapping targets cause a "fake" end on the first target swap. 
- Updated Spiritfont tracking for addition in the "other cds" section in the existing SF module for usage perfs and moved all spiritfont event links into an apex only file
- Added base creation for base celestial analyzer
- Updated Celestial analysis for Spiritfont active during the cd, heart of the jade serpent active with good perf before or ok perf during, and yu'lon specific with soothing mist being used for all enveloping mist casts
- Added Celestial Conduit to healing cooldowns usage alongside Invoke & Revival
- Updated APL for TotM to 4 stacks, removed Shaohao's entry, added Spiritfont at 2 stacks

### Testing

Reports:
`report/yJBtr1THY2mjV3Kx/1-Heroic+Crown+of+the+Cosmos+-+Kill+(9:34)/5-Velomunk/standard/overview`
`report/wHX7QZrFnBTh1xak/26-Mythic+Fallen-King+Salhadaar+-+Kill+(6:01)/10-Slapthatmw/standard/overview`
`report/B8MRAjLyN4vZt9z2/10-Heroic+Lightblinded+Vanguard+-+Kill+(5:45)/15-Magnimonk/standard`

image
image
image
image
image
image


src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
```diff
@@ -80,14 +90,28 @@ class BaseCelestialAnalyzer extends Analyzer {
...
     );
-    this.addEventListener(
-      Events.cast.by(SELECTED_PLAYER).spell(TALENTS_MONK.ENVELOPING_MIST_TALENT),
-      this.onEnvmCast,
-    );
     this.addEventListener(Events.cast.by(SELECTED_PLAYER), this.onAction);
     this.addEventListener(Events.damage.by(SELECTED_PLAYER), this.onAction);
     this.addEventListener(Events.heal.by(SELECTED_PLAYER), this.onAction);
     this.addEventListener(Events.cast.by(SELECTED_PLAYER).spell(this.currentRskTalent), this.onRsk);
+    this.addEventListener(
```
vohrr: Nit for consistency; use a prop `isFlowingWisdomActive` to conditionally run that perf analysis

src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
```diff
@@ -152,11 +219,22 @@ class BaseCelestialAnalyzer extends Analyzer {
...
+  onHotjsApply(event: ApplyBuffEvent) {
+    this.hotjsApplyTimestamp = event.timestamp;
     if (!this.celestialActive) {
       return;
     }
-    this.castTrackers.at(-1)!.totalEnvM += 1;
+    this.castTrackers.at(-1)!.hotjsDuringWindow = true;
+  }
+
+  onHotjsRemove(event: RemoveBuffEvent) {
```
vohrr, line 251: I think we should be resetting the apply timestamp here before the early return like is done in `onHotjsAppy`

src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
```diff
@@ -218,6 +296,68 @@ class BaseCelestialAnalyzer extends Analyzer {
         details: <>{this.getSiBuffType(cast)}</>,
       });
     }
+
+    // heart of the jade serpent (flowing wisdom) buff
+    if (this.selectedCombatant.hasTalent(TALENTS_MONK.FLOWING_WISDOM_TALENT)) {
+      let hotjsPerf = QualitativePerformance.Fail;
+      // tolerance set for tft rem global, should be active on celestial press
+      const expectedDuration = HEART_OF_THE_JADE_SERPENT_DURATION - 1.5 / (1 + cast.averageHaste);
```
vohrr: This is going to be slightly inaccurate since the average haste over the entire celestial will be lower than the haste value at the time of gcd press, especially for jade bond uses.
Recommend tracking the haste value at point in time in `onHotjsApply` or just using the buffed haste value i.e. `baseHaste * flowing_wisdom_haste`

src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
```diff
@@ -218,6 +296,68 @@ class BaseCelestialAnalyzer extends Analyzer {
...
+              : 'None'}
+          </>
+        ),
+      });
+    }
+
+    // spiritfont active - r2/3 of spiritfont increases envm/rsk further during it
+    if (this.selectedCombatant.hasTalent(TALENTS_MONK.SPIRITFONT_2_MISTWEAVER_TALENT)) {
+      const spiritfontPerf = cast.spiritfontActiveDuring
+        ? QualitativePerformance.Good
```
vohrr, line 373: Would be cool if we could also add some analysis to see the avg # of envs active during the buffs duration (doesn't have to be in this PR)

src/analysis/retail/monk/mistweaver/normalizers/SoothingMistLinkNormalizer.tsx
```diff
@@ -0,0 +1,102 @@
...
+  HasAbility,
+  HasSource,
+  HasTarget,
+} from 'parser/core/Events';
+import {
+  SOOTHING_MIST_CHANNEL_START,
+  SOOTHING_MIST_CHANNEL_END,
+} from './EventLinks/EventLinkConstants';
+
+const REAPPLY_BUFFER = 10; // ms
```
vohrr: Gut feeling is this seems like its too low for higher latency logs - standard buffer is 250ms

vohrr (changes requested):

src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
lolswirl: Went with current haste on apply to be more accurate on the global

src/analysis/retail/monk/mistweaver/modules/spells/BaseCelestialAnalyzer.tsx
lolswirl: Great idea, I had a few more additions I wanted to add to the Spiritfont module in general so that might be a good time for it

src/analysis/retail/monk/mistweaver/normalizers/SoothingMistLinkNormalizer.tsx
lolswirl: The remove/apply events of soom are on the same tick, do you think a buffer is even necessary? I didn't go with CAST_BUFFER_MS since I couldn't find any cases where it _wasn't_ on the same timestamp
image

vohrr: Had convo offline, should be fine

vohrr (approved):

PR #7743, opened 2026-03-29
Midnight Frost mage update pt2

Dambroda:
Continuing on @earosselot 's midnight update, another large set of changes for better analysis.

### Description

Changes:
- Fixed a number of incorrect ability/spell/buff definitions and added a few missing ones
- Added CDR stats for Glaciate and Spellfrost Teachings
- Simplified fingers of frost feedback analyzer
- Added Ice Lance feedback section, as its our main damage source (via Freezing shatter) which checks for freezing utilization
- Updated Flurry analyzer to remove obsolete references to icicles and hard casts, instead properly only caring about Brain Freeze (and not munching Thermal Void, if talented)
- Remove obsolete references of the pre-midnight version of shatter

There is still a lot of work to do here and I hope to follow up with more PRs, especially:
- Support comet storm a bit better, and frostfire in general
- Specific feedback for common failure cases, especially something like casting ice lance when you have brain freeze active (a lot of us pre-midnight mages have muscle memory to use ice lance first... which is now incorrect)
- Confirm glacial spike usage numbers are correct, and add feedback section for (it's pretty important)
- I saw the new APL analyzer support (well, new since I last contributed 5 years ago...), and want to implement that for frost -- which may obsolete some of these other changes.

### Testing

- Test report URL: `/report/RC9wBdgvN7QbA4pq/50-Heroic+Vaelgor+&+Ezzorak+-+Kill+(6:32)/38-Berling/standard/overview`

### Screenshot(s):

image

image
image
image


src/analysis/retail/mage/frost/talents/Glaciate.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const MS_PER_FREEZING_STACK = 100;
+
+class Glaciate extends Analyzer {
+  static dependencies = {
+    spellUsable: SpellUsable,
+    iceLance: IceLance,
+  };
+  protected spellUsable!: SpellUsable;
+  protected iceLance!: IceLance;
```
Dambroda, line 23: I wasn't sure if this was the right way to do this. I didn't want to duplicate code that was calculating how many Freezing stacks each cast of ice lance shattered, so I made the IceLance module a dependency of this one and had it expose its `Record<Timestamp, IceLanceCast>` analysis with that metadata in it. Is that the job of a normalizer? If I remember correctly normalizers dont have access to stuff like Enemies.getBuff so I didn't think so. Should it have been a standalone module? Or perhaps, rather than exposing a Record it should be setting properties on the events themselves? Not sure, looking for confirmation or correction on this

src/interface/report/Results/Timeline/Casts.tsx
```diff
@@ -68,7 +69,7 @@ export const isApplicableEvent =
  */
 export const highlightInefficientCast = (
   event: CastEvent | BeginChannelEvent | CastEvent[] | BeginChannelEvent[],
-  tooltip: string,
+  tooltip: ReactNode | MessageDescriptor,
```
Dambroda: I was tempted to use `Parameters<typeof addInefficientCastReason>[1]`, but I've been in some codebases that don't like that style. Any opinions?

Pewtro: Honestly, that type is re-used 4 times in the file (EventMetaLib), just make it a type in there and export it

earosselot (approved):
lgtm

src/analysis/retail/mage/frost/talents/Glaciate.tsx
emallson: we generally try to avoid modifying events these days. event modification doesn't work well with devtools reloading code, and any added properties on events or fabricated events are functionally shared global state.

using a dependency and exposing the `Record` so that another module can consume it is good. the other main way of accomplishing this is inverting it: having `IceLance` call `Glaciate` when something happens. that is more often used when multiple analyzers need to interact with a single common one. I think exposing the `Record` via dependency is better here.

src/analysis/retail/mage/frost/talents/CometStorm.tsx
```diff
@@ -47,9 +47,9 @@ class CometStorm extends Analyzer {
       if (enemy && !enemies.includes(enemy.guid)) {
         enemies.push(enemy.guid);
       }
-      if (enemy && SHATTER_DEBUFFS.some((effect) => enemy.hasBuff(effect.id, d.timestamp))) {
-        shattered += 1;
-      }
+      //if (enemy && SHATTER_DEBUFFS.some((effect) => enemy.hasBuff(effect.id, d.timestamp))) {
+      //  shattered += 1;
+      //}
```
emallson:
```suggestion
```

src/analysis/retail/mage/frost/talents/Flurry.tsx
```diff
@@ -37,85 +41,81 @@ class Flurry extends Analyzer {
...
-    ).length;
-    const withoutBrainFreezeBad = this.flurryEvents.filter((flurry) => flurry.noBfBadCast()).length;
+    if (this.hasThermalVoidTalent && hasThermalVoid && hasBrainFreeze) {
+      highlightInefficientCast(event, munchedThermalVoidTooltip);
+    }
+  }
 
-    const flurryCasts: { label: string; color: string; value: number }[] = [];
+  analyzeFlurry(): { label: any; color: string; value: number }[] {
+    const flurryCasts: { label: any; color: string; value: number }[] = [];
```
emallson: this shouldn't be passing the linter 👀 

it looks like `frost` is still listed in the ignore list in `.oxlintrc.json`. for all of our sanity, i'm not going to demand it be turned on in the middle of this PR, but:

1. fix these `any`s (should probably be `React.ReactNode`)
2. the next PR for frost mage needs to enable linting of the spec.

src/analysis/retail/mage/shared/Abilities.tsx
```diff
@@ -146,7 +128,7 @@ class Abilities extends CoreAbilities {
       //       ? 25 / 1.3
       //       : 25,
       //   gcd: {
-      //     base: 1500,
+      //     base: 500,
       //   },
       // },
```
emallson, line 133: should delete this instead of trying to keep commented-out code in sync

emallson (changes requested):
fix the `any`s and this is mergeable. future PRs need to enable linting for the spec. looking through it, i think a few other specs are currently not being linted. might be an oversight from the transition from eslint to oxlint (e.g. we snapshotted an old version of the eslint ignore list in the oxlint pr)

src/analysis/retail/mage/frost/talents/Flurry.tsx
Dambroda: Oooh, I was surprised I didn't hit more linter errors. I was using any's as placeholders a fair bit and missed this one.

It wasnt too bad to enable the linter for frost, so I've gone ahead and done it here

src/interface/report/Results/Timeline/Casts.tsx
Dambroda: I've exported "Reason", as thats what EventMetaLib calls all of its parameters that use the type. I don't love the name, so open to suggestions if we have a problem with it.

src/analysis/retail/mage/shared/Abilities.tsx
Dambroda: This is for a spec maintained by someone else so I didn't want to just delete it in case they wanted the placeholder there for some reason

PR #7747, opened 2026-03-31
[Survival] Tip of the Spear Analyser Adjustment

Kivlov:
### Description

Couple small fixes with this one. Had Don't Look Back and Lunar Storm spell name's backwards so they were showing incorrectly in the takedown analyser. Raptor Swipe's target's hit for SaO was counting damage events, not unique targets and our pet strikes twice on a swipe.
Not sure how I caused the merge branch commit on the 27th and that march 25th commit I think is the same commit from my last PR.. not sure how I did that either. Thought I had opened a clean branch after my last PR was merged and pulled to my origin/midnight /shrug

Anyways, onto the Tip analysis.
Couple things changing: A while back, they changed how much focus our spells cost, added a cooldown to Kill Command, and removed all of our reset mechanics for it which has left us in an awkward spot where focus went from not mattering at all, to mattering quite a bit. This change is focused arounded cleaning up some of the wording in the Focus and Tip management under the Resource section, and then adding in a performance box row that only captures untipped abilities so that you can see at a glance how many times it occurred. 

The instances you send an untipped ability that isn't raptor strike is quite small, so it's a pretty good indicator of poor play. I also modified some of the rules, adding in specifics between what is acceptable and unacceptable to cause waste with between the two hero trees.

I also encountered an issue where when you get the focus at the cast event, it grabs the post-generation focus (This is a problem within the AplCheck also that my next PR will address) so I added a helper to check against the pre-cast focus for determining if casting kill command at low focus occurred as this can be acceptable wastage.

- Test report URL: `/report/qPHQbNdcG8gRk9L1/21-Mythic+Imperator+Averzian+-+Kill+(6:56)/18-Percival/standard`


src/analysis/retail/hunter/survival/modules/talents/TipOfTheSpear.tsx
```diff
@@ -38,25 +45,104 @@ class TipOfTheSpear extends BuffStackTracker {
...
+      tippableSpells.push(TALENTS.FLAMEFANG_PITCH_TALENT);
+    }
+    if (this.selectedCombatant.hasTalent(TALENTS.RAPTOR_SWIPE_1_SURVIVAL_TALENT)) {
+      tippableSpells.push(SPELLS.RAPTOR_SWIPE_DAMAGE);
+    }
+    if (!this.hasTwinFangs && this.selectedCombatant.hasTalent(TALENTS.TAKEDOWN_TALENT)) {
+      tippableSpells.push(TALENTS.TAKEDOWN_TALENT);
+    }
+
+    if (tippableSpells.length > 0) {
```
Pewtro: This is always > 0 with WFB in there by default

src/analysis/retail/hunter/survival/modules/talents/TipOfTheSpear.tsx
```diff
@@ -38,25 +45,104 @@ class TipOfTheSpear extends BuffStackTracker {
...
+        Events.cast.by(SELECTED_PLAYER).spell(tippableSpells),
+        this.onTippableCast,
+      );
+    }
   }
 
-  get totalWastedStacks(): number {
-    return this.wastedStacks;
+  private getFocus(event: CastEvent): number {
+    return event.classResources?.find((r) => r.type === FOCUS_RESOURCE_TYPE)?.amount ?? Infinity;
```
Pewtro: Consider if the fallback should be 0 instead so events without class resources dont get flagged

src/analysis/retail/hunter/survival/modules/talents/TipOfTheSpear.tsx
```diff
@@ -38,25 +45,104 @@ class TipOfTheSpear extends BuffStackTracker {
...
     const currentStacks = this.current;
-    const hasPrimalSurge = this.selectedCombatant.hasTalent(TALENTS.PRIMAL_SURGE_TALENT);
-    const stacksGained = hasPrimalSurge ? 2 : 1;
-    const potentialStacks = currentStacks + stacksGained;
+    const stacksGained = this.hasPrimalSurge ? 2 : 1;
 
-    const currentFocus =
-      event.classResources?.find((r) => r.type === FOCUS_RESOURCE_TYPE)?.amount ?? Infinity;
+    const potentialStacks = currentStacks + stacksGained;
+    const currentFocus = this.getFocus(event) - this.focusGained;
```
Pewtro: You might need to look at the generated resource of the event because if they're at a value above MAX-20 (or 25) then current focus will be incorrect

src/analysis/retail/hunter/survival/modules/talents/TipOfTheSpear.tsx
```diff
@@ -10,13 +10,13 @@ import STATISTIC_CATEGORY from 'parser/ui/STATISTIC_CATEGORY';
...
 import CastSummaryAndBreakdown from 'interface/guide/components/CastSummaryAndBreakdown';
+import { PerformanceBoxRow } from 'interface/guide/components/PerformanceBoxRow';
 import { explanationAndDataSubsection } from 'interface/guide/components/ExplanationRow';
 import { BoxRowEntry } from 'interface/guide/components/PerformanceBoxRow';
 import { BadColor, GoodColor, OkColor } from 'interface/guide';
 
 const MAX_STACKS = 3;
 const LOW_FOCUS_THRESHOLD = 30;
-// Focus is resource type 2 (see RESOURCE_TYPES.FOCUS)
 const FOCUS_RESOURCE_TYPE = 2;
```
Pewtro: Should probably just use RESOURCE_TYPES.FOCUS.id instead of magic number

src/analysis/retail/hunter/survival/modules/talents/TipOfTheSpear.tsx
Kivlov: Will this matter since the only check on this is to see if they were low focus when they cast it?
```  if (isLowFocus) {
      // Low focus is always an acceptable reason to Kill Command regardless of stacks.
      value = QualitativePerformance.Good;
      header = `Good: low focus (${currentFocus})`;
      color = GoodColor;
    } else if (this.isPackLeader) {
```
    
This is what I'm doing with the currentFocus, is jsut checking if it's < 30. ie could they have sent a raptor instead

Pewtro: It matters to the degree that wowanalyzer has a lot of copying internally so if someone were to copy it without knowing that it's inaccurate at various points they'd have something incorrect

Kivlov: Good point. I have added a normalizer to catch the KC resource change event from the cast. Also added the focus level + change + waste to the tooltip for the summary bars.

PR #7750, opened 2026-04-01
[Holy Priest] Prayer of Healing update

Amiphite:
### Changes:
- Divinity talent support.
- Divinity added to `SPELLS/priest.ts`
- Additional checks for combatant talents.
- Guide section update.
- Better code readability.


src/analysis/retail/priest/holy/modules/spells/PrayerOfHealing.tsx
```diff
@@ -282,35 +335,55 @@ class PrayerOfHealing extends Analyzer {
...
         castBreakdownSmallText = (
           <>
             {' '}
-            - <span className={styles.perfectCast}>Blue</span> is a perfect cast with both{' '}
-            <SpellLink spell={TALENTS.LIGHTWEAVER_TALENT} /> and{' '}
-            <SpellLink spell={TALENTS.SURGE_OF_LIGHT_TALENT} /> active.{' '}
+            - <span className={styles.perfectCast}>Blue</span> is a perfect cast with{' '}
+            <SpellLink spell={TALENTS.LIGHTWEAVER_TALENT} />,{' '}
+            <SpellLink spell={TALENTS.SURGE_OF_LIGHT_TALENT} /> and{' '}
+            <SpellLink spell={TALENTS.DIVINITY_TALENT} /> active.{' '}
```
emallson: noting that this is not actually what is checked for `Perfect`. Divinity appears to be ignored for this condition. if that's intended, should update this text. otherwise, should adjust the conditions for `Perfect`.

emallson (changes requested):
question about inconsistency between guide text and the actual check, but the code looks reasonable

src/analysis/retail/priest/holy/modules/spells/PrayerOfHealing.tsx
Amiphite: I don't know how good this code is from a programming pov, I just made a lot of scenarios for different sets of talents, it still confused me a little, I will fix it very soon

Amiphite:
I'm not sure about `case 'surge-divinity'`, I'll fix it later when it turns out it's useless, make sure everything is fine now, I think I failed to make friends with GitHub again by clicking the wrong buttons

emallson (approved):

PR #7755, opened 2026-04-02
[Unholy] Add Unholy Aura haste tracking

MarchingCube:
### Description
Unholy Aura grants 1%/2% haste per active Magus of the Dead (based on talent rank). Register a new analyzer for unholy aura that integrates haste buff and tracks it.

I've also added statistics tracking for this talent.

### Testing

- Test report URL: report/RgH3wjBdvynZDL7C/7-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(4:52)/37-Scharfer/standard/overview
- Screenshot(s):
image


src/analysis/retail/deathknight/unholy/modules/talents/UnholyAura.tsx
emallson, line 1: unless I've missed something, the `averageStacks` calculation in this file can be replaced by:

```tsx
get averageStacks() {
  return this.selectedCombatant.getStackWeightedBuffUptime(SPELLS.UNHOLY_AURA_BUFF.id) / this.owner.fightDuration;
}
```

(might be worth double checking that it produces the same value, or at least is close)

if you check within that method, you can also see the method call to get the time at each stack count, which you can use for the table.

emallson (changes requested):

src/analysis/retail/deathknight/unholy/modules/talents/UnholyAura.tsx
MarchingCube: This is exactly what I needed. It produces the same value. Simplified the analyzer a lot.

emallson (approved):

PR #7756, opened 2026-04-02
[Fury] Remove Old Talents and Spells

tyler-lutz:
### Description

- Removed Old Talents and Spells
- Updated example report


emallson:
code looks fine, would like a quick look from @GeoffreyBalshaw for correctness

PR #7757, opened 2026-04-02
Replaced Outdated Browser

danny-janse:
### Description

Noticed the link to outdatedbrowser.com has become quite shady redirecting to a different domain. Found an alternative that gives the visitors suggestions to install a new browser/update the current one.

Old site:
image

New Site for old browser:
Screenshot_7


emallson (approved):
damn, good catch

PR #7759, opened 2026-04-03
[Unholy] Sudden doom rework

MarchingCube:
### Description
 - Rewrites the Unholy Death Knight Sudden Doom analyzer to properly handle    
  multi-stack procs using ApplyBuffStack/RemoveBuffStack events
  - Introduces a dedicated SuddenDoomLinkNormalizer (EventLinkNormalizer) that  
  links buff removal events to their consuming cast, eliminating false-positive 
  overwrite/expiration counts
  - Fixes double-counting issues where WCL fires RemoveBuffStack + RefreshBuff  
  at the same timestamp during simultaneous consumption and new proc            
  - Adds a donut chart breaking down consumed vs expired vs overwritten procs

### Testing
I've added a set of unit tests since manual testing is quite complicated.

- Test report URL: /report/RgH3wjBdvynZDL7C/7-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(4:52)/37-Scharfer/standard/statistics
- Screenshot(s):
image

cc: @brandrewsss


src/analysis/retail/deathknight/unholy/normalizers/SuddenDoomLink.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const { normalizer: SuddenDoomLinkNormalizer, linkHelper: SuddenDoomConsumption } =
+  EventLinkNormalizer.build({
+    linkRelation: 'sudden-doom-consumption',
+    linkingEventType: [EventType.RemoveBuff, EventType.RemoveBuffStack],
+    linkingEventId: SPELLS.SUDDEN_DOOM_BUFF.id,
+    referencedEventId: SD_CONSUMERS,
+    referencedEventType: EventType.Cast,
+    backwardBufferMs: 500,
+    forwardBufferMs: 500,
```
emallson, line 27: 500ms on both sides seems quite long for this (1s total). often 100ms (for buffs that virtually always log removal immediately) or 250ms (for those with a bit more delay) is more appropriate.

are there logs where the actual delay between cast and removal is this high?

emallson (approved):
the changes look good overall. need to run `oxfmt` on the test files before merging. i had a question about the size of the window used for the link, but it may simply be that the game logs this with large delay

src/analysis/retail/deathknight/unholy/normalizers/SuddenDoomLink.ts
MarchingCube: So I based this on the report mentioned in the MR description. Key information:
- Deltas are always negative (buff always gets removed before the cast)
- Maximum delta is 284ms
- Values generally cluster around 250ms.

This is why I have picked 500ms to be on the safe side for detecting linked casts. 300ms might be too tight. And 250ms would already miss some events.

MarchingCube: Also pushed formatting changes. Wanted to do it earlier but Easter got in my way :)

emallson: okay. that seems like a good reason to have a wider range. note that the range doesn't need to be symmetric (you can have 500ms before and 50ms after, for example). though i don't think that will matter here, so i'll merge once CI passes.

MarchingCube: I will try to do more exhaustive check after we get spec to a better state. I have bunch of logs from my own runs that I can use.

PR #7760, opened 2026-04-03
Add HP plot to Brew stagger graph

NotStirred:
### Description
Adds player HP to the stagger graph, which is useful to understand why Purifying Brew was cast at a specific moment.

I would like to have the HP graph as a % of max which requires having independent scale, but I don't know how to do that without also forcing the purify casts to have their own independent scale.

### Testing
- Test report URL: `/report/cVTwyWK2fn1FvB8L/1-Mythic++Algeth'ar+Academy+-+Kill+(26:24)/5-Bluefists/standard/stagger`
image

### DIFF
--- a/src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
+++ b/src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
@@ -3,7 +3,7 @@ import talents from 'common/TALENTS/monk';
 import { SpellLink } from 'interface';
 import { Panel } from 'interface';
 import Analyzer, { Options, SELECTED_PLAYER } from 'parser/core/Analyzer';
-import Events, { DeathEvent } from 'parser/core/Events';
+import Events, { DamageEvent, DeathEvent, HealEvent } from 'parser/core/Events';
 import BaseChart, { formatTime } from 'parser/ui/BaseChart';
 import { VisualizationSpec } from 'react-vega';
 import AutoSizer from 'react-virtualized-auto-sizer';
@@ -29,6 +29,7 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
   stagger: StaggerPool,
   pb: PurifyingBrew,
 }) {
+  _hpEvents: (HealEvent | DamageEvent)[] = [];
   _deathEvents: DeathEvent[] = [];
   _lastHp: number | null = null;
   _lastMaxHp: number | null = null;
@@ -36,6 +37,8 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
   constructor(options: Options) {
     super(options);
 
+    this.addEventListener(Events.damage.to(SELECTED_PLAYER), this._damage);
+    this.addEventListener(Events.heal.to(SELECTED_PLAYER), this._heal);
     this.addEventListener(Events.death.to(SELECTED_PLAYER), this._death);
   }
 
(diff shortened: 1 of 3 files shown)


src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -29,13 +29,16 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
   stagger: StaggerPool,
   pb: PurifyingBrew,
 }) {
+  _hpEvents: (HealEvent | DamageEvent)[] = [];
```
emallson: type this as `HitpointsEvent<EventType>`

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -164,6 +211,15 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
...
       );
 
+      let hpEvents = this._hpEvents
+        .filter((event) => event.hitPoints !== undefined)
+        .map(({ timestamp, hitPoints }) => {
+          return {
+            x: timestamp - startTime,
+            hitPoints: hitPoints!!, // !! filtered above
+          };
+        });
```
emallson: once the array type is change, the filter and non-null assertion can be removed


```suggestion
      const hpEvents = this._hpEvents
        .map(({ timestamp, hitPoints }) => {
          return {
            x: timestamp - startTime,
            hitPoints
          };
        });
```

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -193,6 +250,14 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
     }
   }
 
+  _damage(event: DamageEvent) {
+    this._hpEvents.push(event);
+  }
+
+  _heal(event: HealEvent) {
+    this._hpEvents.push(event);
+  }
```
emallson, line 258: use the `HasHitpoints` helper to gate these .push calls


```suggestion
  _damage(event: DamageEvent) {
    if (!HasHitpoints(event)) { return; }
    this._hpEvents.push(event);
  }

  _heal(event: HealEvent) {
    if (!HasHitpoints(event)) { return; }
    this._hpEvents.push(event);
  }
```

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -81,6 +84,50 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
...
+              axis: {
+                labelExpr: formatTime('datum.value'),
+                tickCount: 25,
+                grid: false,
+              },
+              scale: {
+                nice: false,
+              },
+              title: null,
+            },
```
emallson: use the `xAxis` const defined above for the axis to make sure it stays in sync. if the x axis definitions are different the charting library starts behaving weirdly

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -81,6 +84,50 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
...
+          },
+          transform: [
+            {
+              calculate: `datum.timestamp - ${startTime}`,
+              as: 'timestamp_shifted',
+            },
+          ],
+          encoding: {
+            x: {
+              field: 'x',
```
emallson: this transform is unused because the field is defined as `x`, not `timestamp`. `timestamp_shifted` is unused.

you can either eliminate the `transform` on this layer or eliminate the `.map` call when building the array. since the filter call should also be eliminated, that would simplify the array being passed to the chart to simply `this._hpEvents`

emallson (changes requested):
a few small changes. looks reasonable overall, just some simplifications. you should add yourself to `CONTRIBUTORS.tsx` and add an entry to the `retail/monk/brewmaster/CHANGELOG.tsx` file.

also: might be worth loading this on an M+ log and making sure it doesn't break with 30 minutes of point data 🙃

NotStirred:
Let me know if you want a squash 

Tested fine on a long M+ key so should be fine :+1:.

emallson (approved):

emallson:
> Let me know if you want a squash

iirc we have squash-on-merge enabled so no need to worry about that during a PR

PR #7766, opened 2026-04-05
[Arcane Mage] Midnight Update 3 

Sharrq:
Overhaul of the Arcane modules to support the latest rotations and APLs for Arcane Mage.
- Adds support for Spellslinger Missile and Orb builds
- Overhauled Touch of the Magi, Arcane Surge, Arcane Missiles, Arcane Barrage, Arcane Orb, and Presence of Mind to support the latest rotations
- Removed Clearcasting guidance around buff expiration and overcapping since it isnt relevant anymore
- Added Arcane Pulse to the Arcane Spellbook
- Adds CDR for Barrier Diffusion for all Mage Specs
- Added CDR for Spellfrost Teachings

Current - Live
WowaLive

Spellslinger Orb Build
SpellslingerOrb

Spellslinger Missile Build
SpellslingerMissile

Sunfury
Sunfury

Also in this PR is an update to the core TipBox implementation to add additional TipBox types and colors as well as a variation of TipBox called PerformanceTipBox which colors and styles based on a provided QualitativePerformance
image
image


Sharrq:
@emallson When you have a moment, can you review the TipBox and PerformanceTipBox changes in core here? You can ignore all the Arcane specific stuff as I am still working on it and can review those myself. I only need your feedback on TipBox and PerformanceTipBox.

Also as I am still working on Arcane stuff, please dont merge once approved. I can merge it once i am done with my arcane changes.

Sharrq:
I am going to do one last review on my arcane changes to confirm i didnt miss anything, but once TipBox and PerformanceTipBox are reviewed/approved, then this should be good to go. 

I can handle reviewing the Arcane stuff myself, just need a review for the core changes.

Sharrq:
Also fyi @Dambroda @earosselot, I added barrier diffusion's cdr here and added it to frost as well

src/analysis/retail/mage/shared/analyzers/BarrierDiffusion.tsx
```diff
@@ -0,0 +1,39 @@
...
+  static dependencies = {
+    spellUsable: SpellUsable,
+  };
+  protected spellUsable!: SpellUsable;
+
+  constructor(options: Options) {
+    super(options);
+    this.active = this.selectedCombatant.hasTalent(TALENTS.BARRIER_DIFFUSION_TALENT);
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(TALENTS.PRISMATIC_BARRIER_TALENT),
```
Dambroda, line 19: Does this also reduce the cooldown if the barrier just expires? That may be fine as an edge case as this definitely improves the vast majority of cases, but could be worth a TODO comment or note

OR, maybe it's the case that expiring also reduces the cooldown in-game... I havent tested that; seems counter intuitive but I'm not 100% sure

Sharrq: If it expires then spell usable will just try to reduce a cool down that isn't on cool down and will just do nothing. Pretty sure that's just how spell usable works in general

Dambroda: Oh, I forgot only frost has two charges, haha. I was picturing a case where someone spams two barriers in a row and then the first expires while the second charge is still on CD

src/interface/guide/components/TipBox.tsx
```diff
@@ -1,58 +1,187 @@
...
+            <Header>
+              <TitleWrapper $color={color}>
+                {icon && <IconWrapper>{icon}</IconWrapper>}
+                {title && <strong>{title}</strong>}
+              </TitleWrapper>
+            </Header>
+          )}
+          <Content>{children}</Content>
+        </>
+      ) : (
```
emallson: it looks like the `block` layout is unused now. everything is `inline`.

src/analysis/retail/mage/shared/analyzers/BarrierDiffusion.tsx
Sharrq: I actually didn't realize frost has two charges. I can probably add a check to see if it's currently on cool down or not before it reduces it.... But now I'm wondering what would happen in game if they used ice barrier while it's already active and the first naturally expired ... If it would still reduce the CD on the second charge or not...

src/interface/guide/components/TipBox.tsx
```diff
@@ -1,58 +1,187 @@
...
 
 const IconWrapper = styled.span`
-  display: flex;
-  align-items: center;
-  line-height: 1.5;
+  display: inline-block;
+  vertical-align: -0.2em;
+  line-height: 1;
+  margin-right: 0.4em;
+  font-size: 1.4em;
```
emallson:
```suggestion
  vertical-align: -0.1em;
  line-height: 1;
  margin-right: 0.4em;
  font-size: 1.4em;
```

to me, this adjusts the icons to look a little less off-center. ymmv

src/interface/guide/components/TipBox.tsx
```diff
@@ -1,58 +1,187 @@
...
+
+const Header = styled.div`
+  margin-bottom: 6px;
 `;
 
 const Content = styled.div`
-  flex: 1;
   margin-bottom: 0;
   line-height: 1.5;
 `;
```
emallson: it doesn't look like this is doing much anymore. the only diff from containing css now is the`line-height` setting (which may be relevant)

emallson (commented):
the changes to `TipBox` look fine. IMO the variants should be *shortcuts* for common usages, instead of being the *only* usages (effectively exposing the internal `TipBoxLayout` component). but there are enough options that, tbh, i think the vast majority of usages will be covered as-is.

also noting that the `WithTimestamps` variants are now unused as well and could be cut.

Sharrq:
> also noting that the `WithTimestamps` variants are now unused as well and could be cut.

I think in one of my iterations I had it so as part of cast detail or whatever you could call out something that was commonly messed up and it would tell you all the timestamps where that mistake happened. Sort of as a pseudo replacement for suggestions. I'll probably just remove it and maybe add that functionality later if it ends up being useful. 

I'll work on the rest of the feedback

src/analysis/retail/mage/shared/analyzers/BarrierDiffusion.tsx
Dambroda: If you want to just leave a TODO for frost charges specifically here that's fine with me. Not sure what the best fix is currently. Probably trying to link the remove to a damage event or something.

Sharrq:
@emallson I believe i resolved everything you brought up. If you dont have any additional feedback, then i will merge it so i can start getting feedback on the analysis

emallson (approved):

PR #7769, opened 2026-04-05
Implement script for generating enchants

Krealle:
### Description

This is partly vibecoded as I wanted to test out some agentic gameplay.

Raidbots has item enchantment data readily available for us to use in their `enchantments.json` file.
Raidbots also has temporary enchants available in their `temp-enchants.json` file, however, this file lacks an effect id, so we have to find that ourselves. Technically you can find it in the `enchantments.json` file, but some of them are missing, so we try to find it from the DBC instead.
Update: I asked seriallos about adding the effect id to the `temp-enchants.json` file, and it should now be present going forward. So the manual matching is now gated behind effect ids not being present.

I imagine we can also use this for other item generation (maybe even all of it?). The `enchantments.json` file also contain gem data, and there are other files for potions, food, etc. But that will be a future project.

Also abstracted some duplicate logic into a shared helper, tested the other scripts and they still seem to run fine.

The clanker wanted to add some caching to the fetching which works pretty nicely for testing since the DBC data can take a decent while to re-fetch each time, especially when we need 3 different tables for the item->effect matching.


scripts/enchants/enchants-helpers.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+function mapItemEnchantmentToEnchant(entry: ItemEnchantmentStaticDataEntry): Enchant {
+  return {
+    id: entry?.itemId ?? -1,
+    name: entry?.baseDisplayName ?? entry.itemName ?? entry.displayName,
+    icon: entry?.itemIcon ?? entry.spellIcon,
+    effectId: entry.id,
+    craftQuality: entry.craftingQuality,
+  };
```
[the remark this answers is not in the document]
Krealle: Item enchants in the "other" category appear like this, so you can't really use `displayName` in these cases. Not really sure why only Mastery seems to be truncated :thinking: 
```json
{
    "id": 7934,
    "displayName": "32 Int & 70 Sta",
    "spellId": 1229437,
    "spellIcon": "inv_tailoring_spellthread_orange_spellthread",
    "itemId": 240094,
    "itemName": "Sunfire Silk Spellthread",
    "itemIcon": "inv_tailoring_spellthread_orange_spellthread",
    "quality": 4,
    "expansion": 11,
    "craftingQuality": 1,
    "tokenizedName": "sunfire_silk_spellthread_1",
    "stats": [
      {
        "type": "int",
        "amount": 32
      },
      {
        "type": "sta",
        "amount": 70
      }
    ],
    "equipRequirements": {
      "itemClass": 4,
      "itemSubClassMask": 30,
      "invTypeMask": 128
    }
  }
```
Whilst the entry for the example looks like this:
```json
{
    "id": 7964,
    "baseDisplayName": "Enchant Ring - Amani Mast",
    "displayName": "Enchant Ring - Amani Mast 1",
    "spellId": 1236058,
    "spellIcon": "ui_profession_enchanting",
    "itemId": 243954,
    "itemName": "Amani Mastery",
    "itemIcon": "inv_12_profession_enchanting_enchantedvellum_green",
    "quality": 3,
    "expansion": 11,
    "craftingQuality": 1,
    "tokenizedName": "amani_mastery_1",
    "stats": [
      {
        "type": "mastery",
        "amount": 22
      }
    ],
    "equipRequirements": {
      "itemClass": 4,
      "itemSubClassMask": 31,
      "invTypeMask": 2048
    },
    "categoryId": 2273,
    "categoryName": "Rings Enchants"
  }
```

src/common/ITEMS/midnight/enchants.ts
```diff
@@ -554,14 +539,14 @@ const enchants = {
...
     name: 'Enchant Tool - Amani Resourcfulness',
     icon: 'inv_12_profession_enchanting_enchantedvellum_blue',
     effectId: 7976,
     craftQuality: 1,
   },
-  TOOL_AMANI_RESOURCFULNESS_R2: {
+  TOOL_AMANI_RESOURCEFULNESS_R2: {
     id: 243967,
     name: 'Enchant Tool - Amani Resourcfulness',
     icon: 'inv_12_profession_enchanting_enchantedvellum_blue',
```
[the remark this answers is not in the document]
Krealle: Does the clanker not know what is removed and what is added when it does a review? :confused:

scripts/enchants/README.md
```diff
@@ -0,0 +1,28 @@
...
+**For PTR Data:**
+
+```bash
+pnpm run generate-enchants:ptr
+```
+
+#### New Expansions
+
+When generating enchants for a new expansion, you will need to update the `EXPANSION` constant in `scripts/enchants/enchants-helpers.ts` to match the new expansion.
+
```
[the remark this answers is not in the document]
Krealle: @copilot apply changes based on [this feedback](https://github.com/WoWAnalyzer/WoWAnalyzer/pull/7769#discussion_r3036917228)

scripts/utils/dbc-types.ts
```diff
@@ -0,0 +1,52 @@
...
+  ID: number;
+  SpellID: number;
+}
+
+// https://wago.tools/db2/SpellEffect
+export interface SpellEffectEntry {
+  SpellID: number;
+  EffectMiscValue_0: number;
+  Effect: number;
+}
```
emallson: my 2c is that there are basically 2 ways to handle these types:

1. generate the full type and put it in a common location (presumably `dbc-types.ts`)
2. write out a partial type (like these) and make it a local type to whatever is using it

since there are also different column sets in different game versions (e.g. `EffectDieSides` in [old `SpellEffect`](https://wago.tools/db2/SpellEffect?build=1.15.8.66564) that is missing in the [new one](https://wago.tools/db2/SpellEffect)), i lean towards local types since that just seems less likely to trip us up later.

scripts/enchants/generate-enchants.ts
```diff
@@ -0,0 +1,66 @@
...
+} from 'scripts/enchants/enchants-types';
+import {
+  mapItemEnchantmentStaticDataToInternalEntries,
+  getKeyedEnchantmentInternalEntries,
+  mapTempEnchantsStaticDataToInternalEntries,
+  printEnchants,
+} from 'scripts/enchants/enchants-helpers';
+import { RaidbotsStaticDataFile } from 'scripts/utils/raidbots-types';
+
+const ENCHANTS_DIR = `./src/common/ITEMS/midnight/enchants.ts`;
```
emallson: nit: this is a filename, not a dirname

scripts/enchants/enchants-helpers.ts
```diff
@@ -0,0 +1,292 @@
...
+  ${enchants.enchants
+    .map(({ key, value }) => {
+      return `${key}: ${JSON.stringify(value)},`;
+    })
+    .join('\n')}
+  // endregion
+  `;
+}
+
+export function getKeyedEnchantmentInternalEntries(entries: EnchantmentInternalEntry[]) {
```
emallson: this is essentially:

```tsx
export function groupBy<T, K extends keyof T>(key: K, entries: T[]): { [key]: K; values: T[]}[]  {
```
right?

not sure the return type i wrote is valid. doesn't need to be rewritten, but i think a clearer name is in order because (a) the result isn't keyed in the conventional sense (you cannot access by key), and (b) the key used is not the key produced by `createEnchantKey`.

`groupEnchantsByType` simpler + clearer

scripts/enchants/generate-enchants.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const tempEnchantsForExpansion = tempEnchantsData.filter(filterToExpansion);
+  const temporaryEnchantmentsInternalEntries = await mapTempEnchantsStaticDataToInternalEntries(
+    tempEnchantsForExpansion,
+    isPTR,
+  );
+  const keyedTemporaryEnchantsByType = getKeyedEnchantmentInternalEntries(
+    temporaryEnchantmentsInternalEntries,
+  );
```
emallson: more of a structural comment but: most (or at least *some*) of these function calls should be inlined. for example, `mapItemEnchantmentStaticDataToInternalEntries` has the body:

```tsx
  return entries.map((entry) => {
    const categoryName = getItemEnchantmentCategoryName(entry);
    return {
      type: categoryName,
      key: createEnchantKey(entry.itemName, entry.craftingQuality, categoryName),
      value: mapItemEnchantmentToEnchant(entry),
    };
  });
```

that's barely longer than the method call, and much more legible to have inline instead of having to jump-to-definition.

*most* is doing a bit of heavy lifting because, for example, the temp enchant handling is much more involved (and very worthy of being its own function). and there are a few actually shared functions. given the exhaustively verbose function names, i'm gonna guess these were mostly llm generated

emallson (changes requested):
`generateEnchants` has a bit of a "sprawl" problem where it separates work out into enough intermediate method chains that end up harming legibility. would like some cleanup there, and a pass on the method names in `enchants-helpers` after.

this is ideally the type of code that we only have to touch ~once per expansion, so legibility & clarity is v important.

otherwise, this is 👍 love to see more automation of data entry

scripts/enchants/generate-enchants.ts
Krealle: > given the exhaustively verbose function names, i'm gonna guess these were mostly llm generated

yea it went a bit wild with the naming here :sweat_smile:, gave it a pass and should be more eatable

Krealle:
I think some of the sprawling lied in the thinking that this would be re-usable for essentially all item generation since I believe Raidbots has static data for everything that we use. But that is a future project!

emallson (approved):

PR #7771, opened 2026-04-05
Implement dungeon pull navigation buttons.

zprobinson:
It is unwieldy for users to leverage wow analyzer for pull by pull m+ analysis because each time they want to view another pull it requires 3 clicks. Filter -> Drop down -> Pull. By creating navigation buttons, we allow users to have a single click that will let them more easily analyze their performance in a dungeon in a linear way.

### Description
Updated the filter button to include pull navigation buttons when analyzing a dungeon.

Some weirdness I could use some guidance on:
- Not sure what to do when in dungeon mode and the user tries to analyze via the "time" `FilterMode`. Currently the 'prev' button is disabled and clicking 'next' will move it to "phase" `FilterMode` pull index 0.
- You can go from "entire dungeon" to "pull 1" by clicking the "next" button. Once you enter this flow, there is no single button click to go back to "entire dungeon" view. No button for resetting the filter as it were. (at least not for the pull

### Testing

- Screenshot(s):

image
image
image

Noticeably, no dungeon pull navigation buttons when showing raid boss attempt:
image


src/interface/report/Results/Header/FilterButton.tsx
```diff
@@ -14,18 +14,35 @@ import Select from 'interface/controls/Select';
 import useClickOutsideHandler from 'interface/hooks/useClickOutsideHandler';
 import Button from 'interface/controls/Button';
 
-const Btn = styled(Button)`
+const FilterContainer = styled.div`
   grid-area: filter;
-
-  align-self: start;
```
emallson, line 20: restore the `align-self: start;` and the filter will be top-aligned again.

src/interface/report/Results/Header/FilterButton.tsx
```diff
@@ -14,18 +14,35 @@ import Select from 'interface/controls/Select';
...
+  padding: 0 0.5rem;
+
+  &:disabled {
+    opacity: 0.4;
+    cursor: not-allowed;
+    filter: none;
+  }
+
+  & .glyphicon {
+    font-size: 75%;
```
emallson, line 42: add `top: 0;` here to properly center the arrows


```suggestion
    font-size: 75%;
    top: 0;
```

src/interface/report/Results/Header/FilterButton.tsx
```diff
@@ -291,3 +315,50 @@ function usePhases() {
...
+  const goToPrevPull = useCallback(() => {
+    if (canGoPrev) {
+      handlePhaseSelection(selectedPhaseIndex - 1);
+    }
+  }, [canGoPrev, handlePhaseSelection, selectedPhaseIndex]);
+
+  const goToNextPull = useCallback(() => {
+    if (canGoNext) {
+      handlePhaseSelection(
+        selectedPhaseIndex === SELECTION_ALL_PHASES ? 0 : selectedPhaseIndex + 1,
```
emallson: imho hitting `>` while having a time filter should behave like "All Phases". there's not a great way to be clear about it, but at least it can be consistent. to do that, use:

```suggestion
        (selectedPhaseIndex === SELECTION_ALL_PHASES || selectedPhaseIndex === SELECTION_CUSTOM_PHASE) ? 0 : selectedPhaseIndex + 1,
```

emallson (changes requested):
minor changes and we can get this merged

src/interface/report/Results/Header/FilterButton.tsx
zprobinson: included!

src/interface/report/Results/Header/FilterButton.tsx
zprobinson: introduced!

src/interface/report/Results/Header/FilterButton.tsx
zprobinson: reintroduced

emallson (approved):

PR #7772, opened 2026-04-05
(Outlaw) Enabling the spec for Midnight

tinkerer12345:
### Description

I've updated most of the outlaw kit and spellbook to correct the errors provoked by the spell pruning that happened from The war within into midnight.

The red dots that are present on the screenshot are due to not having implemented yet the new talent called **Deadly Pursuit** since I wanted to keep this PR as short as possible fixing only what's most accessible. I've also added the new Preparation spell since it's easy to implement.

#### To do implementations
- [x] Fix Roll the bones %
- Better in-depth on combo point spending
- Fixing the APL
- Blade Flurry statistics
- Deadly Pursuit CDR

I'm not sure if i should set myself as contributor/maintainer for Outlaw but Vollmer was there and he wasn't the one who changed things, so I didn't mean to stain his name in case my changes were undesirable! I also didn't turn it on because the spec overview is still broken.

### Testing

Uncomment this block on the Outlaw rogue config.tsx

```
parser: () =>
  import('./CombatLogParser' /* webpackChunkName: "OutlawRogue" */).then(
    (exports) => exports.default,
  ),
```
I've already updated the base log to use my own parse on Vorasius for test purposes too

image

#### Spellbook
image


src/analysis/retail/rogue/outlaw/modules/apl/KeepItRollingRotation.tsx
```diff
@@ -21,7 +20,11 @@ const hasLowCPFinisherCondition = () => {
 
 const rtbKirCondition = () => {
   return or(
-    buffsCount(ROLL_THE_BONES_BUFFS, 3, 'lessThan'),
+    buffPresent(ROLL_THE_BONES_BUFFS[2]) ||
+      buffPresent(ROLL_THE_BONES_BUFFS[3]) ||
+      buffPresent(ROLL_THE_BONES_BUFFS[4]),
```
Krealle: If you are going to reference specific buffs like this you should use the spell definitions directly instead, since this is both ambiguous and would break if you were to move spells around in the `ROLL_THE_BONES_BUFFS` array.

src/analysis/retail/rogue/outlaw/modules/apl/KeepItRollingRotation.tsx
```diff
@@ -21,7 +20,11 @@ const hasLowCPFinisherCondition = () => {
...
+
+    // buffsCount(ROLL_THE_BONES_BUFFS, 3, 'lessThan'),
```
Krealle: should remove this line instead of commenting it out

src/analysis/retail/rogue/outlaw/modules/apl/KeepItRollingRotation.tsx
```diff
@@ -30,7 +33,10 @@ const rtbKirCondition = () => {
 const COOLDOWNS: Rule[] = [
   {
     spell: TALENTS.KEEP_IT_ROLLING_TALENT,
-    condition: buffsCount(ROLL_THE_BONES_BUFFS, 4, 'atLeast'),
+    condition:
+      buffPresent(ROLL_THE_BONES_BUFFS[2]) ||
+      buffPresent(ROLL_THE_BONES_BUFFS[3]) ||
+      buffPresent(ROLL_THE_BONES_BUFFS[4]),
```
Krealle: same as above, use spell definitions instead

Krealle: also since this condition if repeated, consider making a generic condition that can be used both places instead

src/analysis/retail/rogue/outlaw/modules/spells/RollTheBonesBuffs.tsx
```diff
@@ -44,49 +44,60 @@ class RollTheBonesBuffs extends Analyzer {
...
+          1 and 2.
+          <br />
+          <br />
+          The current odds of receiving the buffs seem to be 55% chance to gain Stage 1, 30% chance
+          to gain Stage 2, 10% chance to gain Stage 3, and a 5% chance to gain Stage 4. (Source:{' '}
+          <a href="https://www.icy-veins.com/wow/outlaw-rogue-pve-dps-rotation-cooldowns-abilities">
+            Icy veins
+          </a>
+          )
         </p>
```
Krealle, line 72: Instead of using the `<br />` tags inside of the `<p>` elements, you should use multiple `<p>` elements to achieve the line breaks.

e.g.
```
<p>
  Lorem Ipsum
</p>
<p>
  Lorem Ipsum
</p>
```

src/analysis/retail/rogue/outlaw/modules/spells/RollTheBonesBuffs.tsx
```diff
@@ -44,49 +44,60 @@ class RollTheBonesBuffs extends Analyzer {
...
-          </li>
-          <li>
-            <b>
-              <SpellLink spell={SPELLS.TRUE_BEARING} />
-            </b>{' '}
-            - <SpellLink spell={SPELLS.RESTLESS_BLADES_TALENT} /> increased to 1.5 seconds per combo
-            point spent.
+            - Critical strike chance increased by 10%.
           </li>
+          <br />
```
Krealle: Remove these `<br />` tags as well.

Krealle (changes requested):
Thanks for the PR! Got a couple of things for you to take a look at but overall it's looking good.

> I'm not sure if i should set myself as contributor/maintainer for Outlaw but Vollmer was there and he wasn't the one who changed things, so I didn't mean to stain his name in case my changes were undesirable! I also didn't turn it on because the spec overview is still broken.

Feel free to set yourself as the maintainer, I'm not actively maintaining the spec.

src/analysis/retail/rogue/outlaw/modules/spells/RollTheBonesBuffs.tsx
Krealle, line 137: I tried it out locally and I think this part needs some changes since there is so much text in the description.
Wrapping the `<RoundedPanel>` in a div will prevent the rounded panel from growing too much. And increasing the `explanationPercent` in `explanationAndDataSubsection` will give some breathing room for the description.
```tsx
    const data = (
      <div>
        <RoundedPanel>
          <strong>
            <SpellLink spell={SPELLS.ROLL_THE_BONES} /> uptime
          </strong>
          {uptimeBarSubStatistic(this.owner.fight, rollTheBonesBarSpec, subBuffsBarSpecs)}
        </RoundedPanel>
      </div>
    );

    return explanationAndDataSubsection(explanation, data, 40);
  }
```

How it currently looks:
Image

With the proposed changes:
Image

Krealle (changes requested):
Thanks for making the changes! Got one more place for you to look at and then it's good to go.

Also took a look at a couple of logs and it seems like the APL checker is marking most top logs as having an accuracy of ~50%, so that will need to be looked at some more before the spec is enabled. Since the spec isn't getting enabled in this PR, feel free to wait with that till later a PR.

src/analysis/retail/rogue/outlaw/modules/spells/RollTheBonesBuffs.tsx
tinkerer12345: Thank you for the help @Krealle ! I'm still learning the project and javascript in general so any suggestions you got are super welcome! I also agree and I will change accordingly. I don't expect to enable the spec until most of the things mentioned above are solved and accurate so I wanted to start small and make smaller scale PRs to carve the path towards a good review tool. This one just happened to be many changes due to many broken things 😄 
If you got anything else you want to point out feel free and I'll change as soon as I can!

Krealle (approved):

PR #7774, opened 2026-04-06
feat(pulls): Improve selection for close calls

### LINKED ISSUE #6266: [UI] Fight Selection Panel/ Boss Percentage
Currently if you wipe on a Boss with 0,X% Life left the HP Progress-Bar is filled up completly.

image

Wouldn't it be better if the percentaged get floored by Math.floor?
So there still would be a gap visible - currently you can't tell why this bar is orange instead of green since it's completly full.

https://github.com/WoWAnalyzer/WoWAnalyzer/blob/940a0e1ed0970f5d5732c729b6b904d94cad73e1/src/interface/report/FightSelectionPanelList.tsx#L94C85-L94C85

math280h:
### Description

This PR Aims to improve the visual language around close calls as they can be hard to see with the current progress bar in pull selection.

Replace the "Wipe x" text with the remaining boss HP% on wipes as the wipe is implied by the boss not being killed, moved the pull count to a small text after the boss HP

It makes the boxes slightly taller to make space for the new text but only by a few px

Closes #6266 

### Testing

- Test report URL: `/report/xZmqapzPBVMfvTbk`
- Screenshot(s):
image
image

Prod:
image

### DIFF
--- a/src/interface/report/FightSelectionPanelList.tsx
+++ b/src/interface/report/FightSelectionPanelList.tsx
@@ -68,6 +68,10 @@ class FightSelectionPanelList extends PureComponent<Props> {
                     {pulls.map((pull) => {
                       const duration = Math.round(pull.end_time - pull.start_time);
                       const Icon = pull.kill ? SkullIcon : CancelIcon;
+                      // pull.fightPercentage is in 1/100ths of a percent (10000 = 100% HP remaining).
+                      const bossHpRemaining = pull.kill ? 0 : pull.fightPercentage! / 100;
+                      // Flag wipes where the boss had less than 2% HP left as a close call.
+                      const isCloseCall = !pull.kill && bossHpRemaining > 0 && bossHpRemaining < 2;
 
                       return (
                         <Link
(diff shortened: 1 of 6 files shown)


src/interface/report/FightSelectionPanelList.tsx
```diff
@@ -90,11 +94,38 @@ class FightSelectionPanelList extends PureComponent<Props> {
...
+                                      color: '#ffd100',
+                                      fontSize: 10,
+                                      lineHeight: 1,
+                                      pointerEvents: 'none',
+                                      whiteSpace: 'nowrap',
+                                    }}
+                                  >
+                                    Close Call: {bossHpRemaining.toFixed(2)}%
+                                  </div>
+                                )}
```
emallson: neat idea, but i don't think this is worthwhile in the current fight list layout. the tooltips are good though

emallson (changes requested):
the tooltips are a good addition, but i don't like the close call text as it stands. i could see us adding the fight % to every Wipe, but the conditional underlay looks off to me.

(also, apologies. that issue should've been closed as NOTPLANNED and I missed it. guess i need to do a pass on the issues)

emallson:
that is to say: if you adjust the wipe boxes to include the fight %, that would potentially be mergeable. but if you don't want to do that right now, the tooltips are mergeable on their own

math280h:
Updated PR based on discussion in discord, also updated images in PR description

ToppleTheNun (approved):

src/interface/report/FightSelectionPanelList.tsx
math280h: Removed

PR #7783, opened 2026-04-06
feat(combatant): Refactor _combatantInfo pattern

### LINKED ISSUE #2996: Replace external usages of _combatantInfo with proper solutions
The private property is used in a lot of places. This is bad practice and makes it harder to refactor.

Use existing props/methods were possible. Make new ones where not.

math280h:
### Description

Refactor `_combatantInfo` to be a true protected type. Also renamed to `combatantInfo`

Added a new enum:
- `Faction`

Added two new getters:
- `get faction(): Faction`
- `get pullStats(): Stats`

Updated all the callers to use the new pattern and no longer reach into the `_combatantInfo` directly

Closes #2996 

### Testing

- Test report URL (Vengance): `/report/hHmNaTyJW8nBwjkr/31-Heroic+Imperator+Averzian+-+Kill+(5:06)/2-Del/standard/character`
- Test report URL (Shaman): `/report/GC9dD2PTcL6FkAJY/27-Mythic+Imperator+Averzian+-+Kill+(6:53)/2-Gamerbutbrs/standard/timeline`


src/parser/core/Combatant.ts
```diff
@@ -78,7 +80,37 @@ class Combatant extends Entity {
...
+    }
+    return {
+      strength: info.strength,
+      agility: info.agility,
+      intellect: info.intellect,
+      stamina: info.stamina,
+      crit: info.critSpell,
+      // Haste is reported per attack type; pick whichever the spec uses.
+      // Falls back to 0 because some test fixtures omit all three fields.
+      haste: info.hasteSpell || info.hasteRanged || info.hasteMelee || 0,
```
emallson: in retail this is correct enough, but in classic this will break since you can have all 3 types of haste and crit at varying values. example:

- [varying haste values](https://fresh.warcraftlogs.com/reports/2MvnkFHt1qWX6Z4N?fight=4&type=summary&source=9&view=events)
- [varying crit values](https://fresh.warcraftlogs.com/reports/2MvnkFHt1qWX6Z4N?fight=4&type=summary&view=events&source=5)

i see that the old code was broken, which is a big sigh, but it should be fixed while we're here. the *most correct* way to fix it is to check the independent values for different uses. however, that would be a huge haste refactor for little benefit and is not practical. the *second most correct* way to fix it is to slap a `Math.max` on it like so:


```suggestion
      crit: Math.max(info.critSpell ?? 0, info.critMelee ?? 0, info.critRanged ?? 0),
      // Haste is reported per attack type; pick whichever the spec uses.
      // Falls back to 0 because some test fixtures omit all three fields.
      haste: Math.max(info.hasteSpell ?? 0, info.hasteMelee ?? 0, info.hasteRanged ?? 0),
```

emallson (changes requested):
good change, small correction we should add while we're here

emallson (approved):

PR #7785, opened 2026-04-07
Run item stat scaling script

Dambroda:
### Description

Ran the item stat scaling value generation script so that an upcoming trinket module I have will work correctly; seems to have been broken with the stat squish

### Testing

Unit tests updated. Can also test with any additional items you have in-game to test against their tooltips.


src/parser/retail/modules/items/midnight/embellishments/DarkmoonSigilHunt.tsx
```diff
@@ -10,7 +10,7 @@ import UptimeBar from 'parser/ui/UptimeBar';
 import SpellLink from 'interface/SpellLink';
 import { formatDuration, formatPercentage } from 'common/format';
 
-const STAT_RATING = 56;
+const STAT_RATING = 85;
```
Dambroda, line 13: @MarchingCube I did confirm 85 was correct in-game when I had the stat buff

emallson (approved):

PR #7789, opened 2026-04-09
Remove Dire Beast for BM Hunter

swhild3n:
### Description

Edited 04/12/2026

Changed PR to be to remove the Dire Beast statistic from BM hunter. It no longer provides a haste buff so tracking uptime isn't necessary


src/analysis/retail/hunter/beastmastery/modules/talents/DireBeast.tsx
```diff
@@ -1,81 +1,44 @@
...
-    this.direbeastDuration +=
-      DIRE_FRENZY_INCREASE_DB_TIME[
-        this.selectedCombatant.getTalentRank(TALENTS.DIRE_FRENZY_TALENT)
-      ];
     this.addEventListener(Events.damage.by(SELECTED_PLAYER_PET), this.onPetDamage);
     this.addEventListener(
       Events.summon
         .by(SELECTED_PLAYER)
-        .spell([SPELLS.DIRE_BEAST_SUMMON, SPELLS.DIRE_BEAST_GLYPHED]),
+        .spell([SPELLS.DIRE_BEAST_SUMMON, SPELLS.DIRE_BEAST_GLYPHED, SPELLS.DARK_HOUND_SUMMON]),
```
Pewtro: Since the dark hound has a separate spell id, you can always know it's a Dark Hound, so we can separate it out or check by spellId to mark things as dark hound procs.

src/analysis/retail/hunter/beastmastery/modules/talents/DireBeast.tsx
```diff
@@ -93,9 +56,14 @@ class DireBeast extends Analyzer {
       event.timestamp - this.lastKillCommandCast < MS_BUFFER_100 &&
       this.selectedCombatant.hasTalent(TALENTS.DIRE_COMMAND_TALENT)
     ) {
-      //Then the summon is caused by a Dire Command proc
-      this.isDireBeastSummon = false;
-      return;
+      this.direCommandProc += 1;
```
Pewtro: Maybe all the Dire Command logic is better put into a Dire Command specific module rather than mixing it with Dire Beast, similiarly with Corpsecaller? Especially since you can have Corpsecaller and/or Dire Command without Dire Beast talented

src/analysis/retail/hunter/beastmastery/modules/talents/DireBeast.tsx
Pewtro: Simply putting all Dire Beasts summoned into this is overvaluing the talent given that there are several other things that can spawn Dire Beasts, Pack Leader bear spawns 2, Bestial Wrath spawns 1 guaranteed with tier (+1 with Wailing Dead)

Pewtro (commented):
Thanks for the PR - left a few notes. 

Potentially consider if such a Dire Beast statistic is even interesting, it's not something you play around or can change your gameplay around - it just happens. Perhaps it might be better to delete it entirely?

swhild3n:
@Pewtro Agreed that it might not be worth keeping around. I did ask in the discord and it was suggested to keep it just because it may be useful to know that a log you are comparing yourself to just out-procced dire beast. I personally do not see a great deal of value, but I want this to be useful to all BM hunters.

I'm very open to just removing the statistic. I don't know who can make that final call, though

Pewtro:
What discord? I don't see any posts in Trueshot Lodge related to this to see the context of this discussion. 
I can make that call - as it is in the PR it is misleading and incorrect, and is better off removed. If there is a desire for it in the future it can be added. If you want it to remain, you will need to correct the module to account for all other sources of Dire Beast and appropriately split out the other summon triggering abilities

swhild3n:
In the WowAnalyzer discord. Should I be asking in Trueshot Lodge?
Also yes I agree the change is misleading and if it's desired to be kept I will make the necessary fixes.
Should we gather sentiment from the Trueshot Lodge discord or just go ahead and remove the statistic?

Pewtro:
Ah, I assumed the hunter discord. No you dont need to gather sentiment just remove it

Pewtro (approved):

PR #7796, opened 2026-04-11
12.0.5 Aug changes

KYZ124:
Implements the new Mighty Inferno and Chronal Dynamo talents, as well as Afterimage. For the latter two, only the Aug-relevant parts have been added, and anything relating only to Preservation is not yet implemented.
Updated Accretion module to account for Afterimage and Duplicate.
Removes Molten Embers and Energy Cycles.

/report/AWafk9bJxCZFRK2t/
A
B

/report/VRY4NJMjLWnGbmK8/10-Mythic+Imperator+Averzian+-+Kill+(4:16)/5-Evook/
C


src/analysis/retail/evoker/augmentation/modules/talents/MightyInferno.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  onInfernosApply(targetID: number, timestamp: number) {
+    this.infernoApps.push({
+      playerID: targetID,
+      timestamp,
+      masteryPercentage: this.stats.currentMasteryPercentage,
+    });
+  }
+
+  onInfernosRemove(targetID: number, timestamp: number, fightEndOrRefresh: boolean) {
```
Krealle: `fightEndOrRefresh` should be renamed to `fightEnd` since you are passing in `false` on the refresh event, unless that should be passing `true`?
Actually this param isn't even used so should just omit it.

src/analysis/retail/evoker/augmentation/modules/talents/MightyInferno.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  onRefreshBuff(event: RefreshBuffEvent) {
+    this.onInfernosRemove(event.targetID, event.timestamp, false);
+    this.onInfernosApply(event.targetID, event.timestamp);
+  }
+
+  onFightEnd(event: FightEndEvent) {
+    Object.keys(this.infernoApps).forEach((targetID) => {
+      this.onInfernosRemove(Number(targetID), event.timestamp, true);
+    });
```
Krealle: I don't think this is doing what you intend it to. Using `Object.keys(this.infernoApps)` here won't make keys out of `playerID`, but instead it will make keys out of the indices.
```ts
this.infernoApps = [
  { playerID: 10 },
  { playerID: 5 },
];

Object.keys(this.infernoApps).forEach((targetID) => console.log(targetID)) // "0", "1"
this.infernoApps.forEach((app) => console.log(app.playerID)) // 10, 5
```

You should loop over the array directly instead
```ts
this.infernoApps.forEach((app) => this.onInfernosRemove(app.targetID, event.timestamp, true))
``` 

Also, this appears to be a list of currently active applications that you later splice out as they lose the application. So I would properly change this to either a `Record` or `Map` that uses the `playerID` as the key.
Then you could get entries from an event's `targetID` directly instead of having to do a `findIndex` lookup.
Would then also rename it to `activeInfernoApps`.

src/analysis/retail/evoker/augmentation/modules/talents/MightyInferno.tsx
```diff
@@ -0,0 +1,137 @@
...
+    if (index === -1) {
+      return;
+    }
+    const infernosDuration = (timestamp - this.infernoApps[index].timestamp) / 1000;
+    // While refreshing Inferno's Blessing with Fire Breath will appear to set the duration to 10 or 11 sec,
+    // this is actually 8 sec and then immediately being extended by 2 or 3 sec.
+    const baseInfernosDuration =
+      (INFERNOS_BLESSING_BASE_DURATION_MS *
+        (1 + TIMEWALKER_BASE_EXTENSION + this.infernoApps[index].masteryPercentage)) /
+      1000;
```
Krealle: base duration should be calculated on application?

src/analysis/retail/evoker/augmentation/modules/talents/MightyInferno.tsx
```diff
@@ -0,0 +1,137 @@
...
+  statistic() {
+    return (
+      <Statistic
+        position={STATISTIC_ORDER.OPTIONAL(13)}
+        size="flexible"
+        category={STATISTIC_CATEGORY.TALENTS}
+      >
+        <TalentSpellText talent={TALENTS.MIGHTY_INFERNO_TALENT}>
+          <ItemDamageDone amount={this.damage} />
+          <div></div>
```
Krealle: `<div></div>` should not be used as an alternative to `<br />`! you should wrap the above and below elements in separate `div`s instead to achieve it.

src/analysis/retail/evoker/augmentation/CHANGELOG.tsx
```diff
@@ -28,10 +29,10 @@ export default [
   //change(date(2025, 1, 16),  <>Implement <SpellLink spell={TALENTS.TIME_SPIRAL_TALENT}/>, <SpellLink spell={TALENTS.TIME_CONVERGENCE_TALENT}/>, <SpellLink spell={TALENTS.MASTER_OF_DESTINY_TALENT}/>, <SpellLink spell={TALENTS.MOTES_OF_ACCELERATION_TALENT}/>, <SpellLink spell={TALENTS.GOLDEN_OPPORTUNITY_TALENT}/>, <SpellLink spell={TALENTS.OVERLORD_TALENT}/>, and <SpellLink spell={TALENTS.HOARDED_POWER_TALENT}/> modules</>, KYZ),
   change(date(2024, 12, 30), "Update guide section", KYZ),
   change(date(2024, 12, 27), <>Implement <SpellLink spell={TALENTS.MOMENTUM_SHIFT_TALENT}/>, <SpellLink spell={TALENTS.PRIMACY_TALENT}/> and Volcanic Upsurge modules</>, KYZ),
-  change(date(2024, 12, 25), <>Add <SpellLink spell={TALENTS.REVERBERATIONS_TALENT}/> damage to <SpellLink spell={TALENTS.MOLTEN_EMBERS_TALENT} /> statistics</>, Vollmer),
+  change(date(2024, 12, 25), <>Add <SpellLink spell={TALENTS.REVERBERATIONS_TALENT}/> damage to Molten Embers statistics</>, Vollmer),
```
Krealle: lets just remove all the pre-midnight entries instead

src/analysis/retail/evoker/shared/modules/normalizers/ChronowardenCastLinkNormalizer.ts
```diff
@@ -0,0 +1,110 @@
...
+      return (
+        !HasRelatedEvent(referencedEvent, AFTERIMAGE_DAMAGE_LINK) &&
+        !HasRelatedEvent(referencedEvent, LIVING_FLAME_CAST_HIT) &&
+        !HasRelatedEvent(referencedEvent, PUPIL_OF_ALEXSTRASZA_LINK) &&
+        HasRelatedEvent(referencedEvent, AFTERIMAGE_CAST_LINK) &&
+        HasRelatedEvent(linkingEvent, UPHEAVAL_CAST_DAM_LINK)
+        // If something like the Undermine tier set is added again, the last two conditions will have to change.
+      );
+    },
+  },
```
Krealle, line 83: This won't behave the way that you imagine it will, and needs to be reworked. Will DM specifics.

src/analysis/retail/evoker/shared/modules/normalizers/ChronowardenCastLinkNormalizer.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.priority += 100;
+  }
+}
+
+export function isFromAfterimageDamage(event: DamageEvent): boolean {
+  return HasRelatedEvent(event, AFTERIMAGE_DAMAGE_LINK);
+}
+
+export function getChronoFlameDamageLink(event: DamageEvent): DamageEvent {
+  return GetRelatedEvent(event, CHRONO_FLAME_DAMAGE_LINK) as DamageEvent;
```
Krealle: You should do `GetRelatedEvent<DamageEvent>` instead of casting.

src/common/TALENTS/evoker.ts
Krealle, line 1: Did you manually edit this? You can (and should) generate PTR talents using `npm run generate-talents:ptr`

Krealle (changes requested):

Krealle (approved):

PR #7807, opened 2026-04-11
[Interface] Introduce toaster to better display errors

Hezaerd:
### Description

- Introduce [Sonner](https://sonner.emilkowal.ski/getting-started) as a lightweight and unstyled library for toasts (~4kb).
- Split fatal and non-fatal uncaught errors in the app shell.
- Replace blocking browser alerts with toasts.
- Replace blocking non-fatal error window with toasts.

#### Notes
- Will probably need better styling
- Will definitely need better phrasing
- Resolve [this todo](https://github.com/WoWAnalyzer/WoWAnalyzer/blob/927d46c09d786157adc14d009beea6421c79077f/src/interface/RootErrorBoundary.tsx#L141) about displaying a toast to user instead of hiding the entire app and that all errors aren't fatal ones

### Testing

- Check report selector, name search, debug annotations, and player loading flows for Sonner toasts instead of browser alerts.
- Check blocked player tiles now explain the state with a tooltip.
- Trigger a recoverable global error and confirm the app stays interactive while showing a toast.
- Trigger a fatal render/analysis error and confirm the fullscreen error state still appears.

### Screenshots:
empty-report-toast
manual-error-toast1
manual-error-toast2
{DC30C03E-21CF-440C-9EF4-69B598695113}

### DIFF
--- a/src/interface/DebugAnnotationsTab.tsx
+++ b/src/interface/DebugAnnotationsTab.tsx
@@ -9,6 +9,7 @@ import { Ability, AnyEvent, HasAbility, HasSource, HasTarget } from 'parser/core
 import { useMemo, useState, useCallback } from 'react';
 import { useCombatLogParser } from './report/CombatLogParserContext';
 import { formatDuration } from 'common/format';
+import { toast } from 'sonner';
 import SpellLink from './SpellLink';
 
 export default function DebugAnnotationsTab({ parser }: { parser: CombatLogParser }) {
@@ -270,7 +271,7 @@ function CopySpellData({ ability }: { ability: Ability }) {
       const text = `${key}: ${data},`;
       await navigator.clipboard.writeText(text);
     } catch {
-      alert('Unable to copy data to clipboard');
+      toast.error('Unable to copy data to clipboard');
     }
   }, [ability]);
   return <CopyTextLink onClick={copy}>(copy definition)</CopyTextLink>;
--- a/src/interface/NameSearch.tsx
+++ b/src/interface/NameSearch.tsx
@@ -7,6 +7,7 @@ import { REALM_LIST, CLASSIC_REALM_LIST } from 'game/RealmList';
 import { FormEvent, useCallback, useEffect, useRef, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import SelectSearch from 'react-select-search';
+import { toast } from 'sonner';
 import { useLingui } from '@lingui/react';
 import { RETAIL_EXPANSION_NAME, CLASSIC_EXPANSION_NAME } from 'game/Expansion';
 
@@ -43,7 +44,7 @@ const NameSearch = ({ type }: Props) => {
       const makePageUrl = type === SearchType.CHARACTER ? makeCharacterPageUrl : makeGuildPageUrl;
 
       if (!game || !region || !realm || !name) {
-        alert(
+        toast.warning(
           i18n._(
             defineMessage({
               id: 'interface.nameSearch.pleaseSelect',
@@ -56,7 +57,7 @@ const NameSearch = ({ type }: Props) => {
 
       // Checking for guild-exists here makes it more user-friendly and saves WCL-requests when guild doesn't exist
       if (loading) {
-        alert(
+        toast.info(
           i18n._(
             defineMessage({
               id: 'interface.nameSearch.stillWorking',
@@ -76,7 +77,7 @@ const NameSearch = ({ type }: Props) => {
           response = await fetch(makeCharacterApiUrl(undefined, region, realm, name));
         }
         if (response.status === 500) {
-          alert(
+          toast.error(
             i18n._(
               defineMessage({
                 id: 'interface.nameSearch.noResponse',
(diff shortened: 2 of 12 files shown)


src/interface/report/PlayerLoader.tsx
```diff
@@ -161,7 +162,7 @@ const PlayerLoader = ({ children }: Props) => {
         ),
       );
     } else if (!config) {
-      alert(
```
emallson, line 164: these should be `alert`s

src/interface/report/PlayerTile.tsx
```diff
@@ -26,14 +27,11 @@ interface BlockLoadingProps {
...
-      alert(message);
-    }}
-  >
-    {children}
-  </span>
+  <Tooltip content={message}>
+    <span className="player" style={{ cursor: 'not-allowed', opacity: 0.6 }}>
+      {children}
+    </span>
+  </Tooltip>
```
emallson, line 34: i think the tooltip here is a good addition, but this should also handle the user's click with an `alert` as before.

src/interface/DebugAnnotationsTab.tsx
```diff
@@ -270,7 +271,7 @@ function CopySpellData({ ability }: { ability: Ability }) {
       const text = `${key}: ${data},`;
       await navigator.clipboard.writeText(text);
     } catch {
-      alert('Unable to copy data to clipboard');
+      toast.error('Unable to copy data to clipboard');
```
emallson, line 274: this should be an `alert`

src/interface/NameSearch.tsx
```diff
@@ -43,7 +44,7 @@ const NameSearch = ({ type }: Props) => {
       const makePageUrl = type === SearchType.CHARACTER ? makeCharacterPageUrl : makeGuildPageUrl;
 
       if (!game || !region || !realm || !name) {
-        alert(
+        toast.warning(
```
emallson, line 47: this should be an `alert`

src/interface/NameSearch.tsx
```diff
@@ -98,7 +99,7 @@ const NameSearch = ({ type }: Props) => {
           setLoading(false);
           return;
         } else if (!response.ok) {
-          alert(
+          toast.error(
```
emallson, line 102: these should be `alert`s.

the "name not found" case might warrant its own ui state, but for now an alert is fine

src/interface/ReportSelecter.tsx
```diff
@@ -157,7 +158,7 @@ const ReportSelecter = () => {
     e.preventDefault();
 
     if (!reportCode) {
-      alert('Enter a report first.');
+      toast.warning('Enter a report first.');
```
emallson, line 161: this should be an `alert`

emallson (changes requested):
up front: i don't think that every place that this PR converts to a toast is a good usage of toasts.

to explain, i think it is useful to look at what toasts (at least as they exist in `sonner`) are:

a toast is:

- **ephemeral.** it appears, then disappears on its own.
- **out-of-the-way.** it appears in a static location, neither being connected to the action the user took (like a tooltip) nor stealing focus (like a modal or alert)
- **small.**

i don't think that toasts make sense for most errors. this PR acknowledges that, and fatal errors are handled the same as they have been (with some improvements to the error processing logic). however, errors tied to *user input* are moved to toasts. i think that most of these are worse than `alert` usage.

fundamentally: if the user clicks something, the UI should obviously do something. if an error occurs because of something the user does *and* that error prevents the UI from doing what they expect it to do, it should not be a toast. that is more akin to a fatal error.

on the other hand, if it is either something that the user didn't trigger (like a background update) or something that doesn't prevent the ui from behaving as expected (like failing to load extra details from bnet), then a toast for the error is fine.

in my PR comments, i'm just saying "should be `alert`s", though adding a prettier alert/modal box wouldn't be a bad thing.

math280h:
> up front: i don't think that every place that this PR converts to a toast is a good usage of toasts.
> 
> to explain, i think it is useful to look at what toasts (at least as they exist in `sonner`) are:
> 
> a toast is:
> 
>     * **ephemeral.** it appears, then disappears on its own.
> 
>     * **out-of-the-way.** it appears in a static location, neither being connected to the action the user took (like a tooltip) nor stealing focus (like a modal or alert)
> 
>     * **small.**
> 
> 
> i don't think that toasts make sense for most errors. this PR acknowledges that, and fatal errors are handled the same as they have been (with some improvements to the error processing logic). however, errors tied to _user input_ are moved to toasts. i think that most of these are worse than `alert` usage.
> 
> fundamentally: if the user clicks something, the UI should obviously do something. if an error occurs because of something the user does _and_ that error prevents the UI from doing what they expect it to do, it should not be a toast. that is more akin to a fatal error.
> 
> on the other hand, if it is either something that the user didn't trigger (like a background update) or something that doesn't prevent the ui from behaving as expected (like failing to load extra details from bnet), then a toast for the error is fine.
> 
> in my PR comments, i'm just saying "should be `alert`s", though adding a prettier alert/modal box wouldn't be a bad thing.

I mentioned similar in a conversation in the discord where I think some things like user input should be actually "attached" to the field e.g. error text right below or in the input field and the button disabled when we know for a fact its wrong (e.g. no link)
image

However it may be a large ask for this PR and should possibly be a follow-up
https://discord.com/channels/316864121536512000/316864385052049420/1492656276738543676

emallson:
yea, i think focusing this PR on:

- improvements to the error boundary's error handling, and
- toasts (if any)

is right. other improvements to error presentation should be separate

PR #7813, opened 2026-04-12
fix(typing): Limit usage of ts-ignore

math280h:
### Description

This PR intends to limit the usage of ts-ignore by fixing typing of multiple files.

I've had to keep the ignore for CSSProperties as the solution marked as accepted on the package repo simply didnt work anymore and I couldn't find a better solution that wouldn't require big changes

### Testing

- Test report URL: `/report/2ygv3QaYBfnRA14c/22-Mythic+Vaelgor+&+Ezzorak+-+Kill+(6:57)/70-Tabbìe/standard`

### DIFF
--- a/src/analysis/retail/priest/holy/modules/core/EchoOfLightMastery.tsx
+++ b/src/analysis/retail/priest/holy/modules/core/EchoOfLightMastery.tsx
@@ -18,7 +18,8 @@ import { ABILITIES_THAT_TRIGGER_MASTERY } from '../../constants';
  * UNTIL THEN, JUST THE FRONT END IS DISABLED.
  */
 
-const DEBUG = false;
+// widened to boolean so TS doesn't narrow to literal `false` and flag DEBUG && ... as unreachable
+const DEBUG = false as boolean;
 const CUTOFF_PERCENT = 0.01;
 
 interface EoLHealEvent extends HealEvent {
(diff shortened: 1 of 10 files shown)


types/css-custom-properties.d.ts
```diff
@@ -0,0 +1,8 @@
+import 'react';
+
+declare module 'react' {
+  interface CSSProperties {
+    // oxlint-disable-next-line consistent-indexed-object-style -- template literal pattern keys cannot use Record syntax
+    [key: `--${string}`]: string | number | undefined;
+  }
+}
```
emallson, line 8: mixed feelings on patching the react type like this, but especially as we look toward migrating to css modules (away from `emotion`), this would become increasingly annoying.

and, anyway, this saves `as CSSProperties` casts all over the place

src/analysis/retail/priest/holy/modules/core/EchoOfLightMastery.tsx
```diff
@@ -18,7 +18,8 @@ import { ABILITIES_THAT_TRIGGER_MASTERY } from '../../constants';
  * UNTIL THEN, JUST THE FRONT END IS DISABLED.
  */
 
-const DEBUG = false;
+// widened to boolean so TS doesn't narrow to literal `false` and flag DEBUG && ... as unreachable
+const DEBUG = false as boolean;
```
emallson: i think this is a flag in your editor? removing the widening passes `pnpm run typecheck` and doesn't show any errors in the file (via LSP)


```suggestion
const DEBUG = false;
```

i only care about this because the `const DEBUG` pattern is pretty common and i want to avoid an unnecessary cast getting copied around

emallson (changes requested):
cssprops is fine, that comment is more thinking-out-loud. the type extension in `EchoOfLightMastery` is odd to me

src/analysis/retail/priest/holy/modules/core/EchoOfLightMastery.tsx
math280h: It was from the oxc extension but will double check

emallson (approved):

PR #7818, opened 2026-04-14
[RDruid] Guide Improvements + Bug Fixes

foobartles:
### Description

Improve the guide for resto druid by adding an abundance stack graph and lightening up the requirements on what a "good" convoke is. Also fixed a bug with Abundance/Strategic Infusion % crit increase. 

### Testing

- Test report URL: `/report/mA7d1YcV9aRJWjQC/65-Mythic+Chimaerus,+the+Undreamt+God+-+Kill+(6:30)/341-Solareana/standard`
- Screenshot(s):
image


src/parser/core/EventCalculateLib.tsx
foobartles, line 1: Requesting a review since this is in Core.

I am pretty sure this is a long standing unnoticed bug with `calculateEffectiveHealingFromCritIncrease` and having a `critHealMultiplier  > 2`

When calculating the baseCritHeal it wasn't account for critBonusFactor. It was assuming that a crit does 2x the base heal(the crit bonus is exactly equal to baseHeal) But when critHealMultiplier is not 2, the crit bonus portion is actually baseHeal * (critHealMultiplier - 1), not just baseHeal.

emallson: i think you're on the right track, and i am not sure that the new implementation is correct. the current implementation is *also* confusing, so apologies for the long comment.

stepping back to think about this: the actual thing being done is saying "if you crit 5% more on average due to this effect, then we're going to count 5% of the *additional healing* done by a crit toward this effect".

the additional healing done by a crit is:

```
additionalHealing = raw - raw / critHealMultiplier
                  = (1 - 1 / critHealMultiplier) raw
```

`raw / critHealMultiplier` is called `baseHeal` in the code and does appear repeatedly (sort of). but its hard to track, so i'm gonna work through the details. i'm going to side-step the "what does this mean" stuff right now to not get side-tracked, but our statement is essentially: 

```
additionalHealing' = (c + p) / c * additionalHealing
```

where `additionalHealing'` is the additional healing with increased crit rate. that statement is nonsense when viewing a single event, but makes sense when viewing the total (aka it is an *amortized* statement). if you do a little algebra, you get something that looks an awful lot like the `baseCritHeal` calculation in the current code:

```
additionalHealing' * c / (c + p) = additionalHealing
```

the output of the function should be (ignoring overkill, we'll come back to it in a sec):

```
additionalHealing' - additionalHealing
= additionalHealing' - additionalHealing' * c / (c + p)
```

or, put into code:

```tsx
const additionalHealingFromCrit = raw - raw / critHealMultiplier;
const amountFromCritIncrease = additionalHealingFromCrit - additionalHealingFromCrit * currentCrit / (currentCrit + percentCritIncrease);
```

now: overheal. if more crit just makes you overheal more, it isn't *doing* anything. so the current method is eliminating overhealing from the amount awarded to the effect being tracked, but it does so in a roundabout way via base heal calculations. once you have `amountFromCritIncrease`, you can simply do:

```tsx
return Math.max(0, amountFromCritIncrease - overheal);
```

- if `overheal > amountFromCritIncrease`, then all of the value of the crit increase is wasted.
- if `overheal = 0`, then `amountFromCritIncrease` is the full effect. (remember that we start from the larger value and *remove* from it, so we can't produce *extra* (theoretical) overhealing as part of our calculation)
- if `amountFromCritIncrease > overheal > 0`, then we remove all of the overhealing from the amount we calculated and return the rest.

so, to wrap it up, IMO this implementation is a good deal clearer:


```tsx
/**
 * Calculate what percent of a heal event can be attributed to a flat crit bonus (like +4% crit chance).
 *
 * ## Technical Details
 * This function uses an *amortized* model of events. It "steals" a part of each event proportional to the crit increase
 * and returns it (after removing overhealing). This doesn't make sense on any individual heal (crit % 
 * increases don't increase the amount healed), but does make sense when you add up the totals 
 * at the end, because you should get extra total healing on average.
 *
 * @param event a critical heal event
 * @param currentCrit the current critical strike percentage without the crit buff (example: 0.05 is 5%)
 * @param flatCritIncrease the critical strike increase being added by the buff (example: if your buff adds 4% crit, this is 0.04)
 * @param critHealMultiplier the multiplier for critical healing. this is usually 2 (the default value), but some effects increase it 
 */
export function calculateEffectiveHealingFromCritIncrease(
  event: LightWeightHealingEvent,
  currentCrit: number,
  flatCritIncrease: number,
  critHealMultiplier = 2,
): number {
  // 1 - the total amount of healing done (including overhealing)
  const raw = event.amount + (event.absorbed ?? 0) + (event.overheal ?? 0);
  // 2 - the amount of (1) that is done due to it being a critical heal
  const additionalHealingFromCrit = raw - raw / critHealMultiplier;
  // 3 - the amount of (2) that is attributed to increased crit chance.
  const amountFromCritIncrease = additionalHealingFromCrit - additionalHealingFromCrit * currentCrit / (currentCrit + flatCritIncrease);
  // 4 - remove overhealing from (3), returning 0 if it is all overhealing)
  return Math.max(0, amountFromCritIncrease - (event.overheal ?? 0));
}
```

i know that is a wall of text but i think it is worth being detailed here. would appreciate someone checking my reasoning 😅

foobartles: This definitely makes more sense to me, and I appreciate the in depth response here. Will update my changes to include this implementation. Thanks!

PR #7823, opened 2026-04-15
feat(combatant): Refactor the lazy import of glyphs

### LINKED ISSUE #7779: Refactor `hasGlyph` lazy imports in combatant
Currently we import `this._importGlyphs(this.combatantInfo);` on every single call to the function, we should import it in the constructor once instead.

math280h:
### Description

This PR intends to refactor `hasGlyph` because it was calling `this._importGlyphs` (It did nothing after the first call but there is no reason to call it again) on every call instead of just importing once.

Closes #7779 

### Testing

- Test report URL: `/report/zhLYXJWARjk3FtqP/4-Heroic+Horridon+-+Kill+(4:25)/63-Cvoidzone/standard/overview`


Krealle (approved):
looks like a good change, add a changelog entry and we can get it merged!

PR #7837, opened 2026-04-19
[Demonology] [Warlock] Tyrant analyzer and Guide Update

katogaming88:
## Summary

- **Demonic Tyrant analyzer overhaul**: tracks Demonic Power stacks, Soul Shards at cast and window end, Dreadstalker timing, Grimoire (Imp Lord / Fel Ravager), and Doomguard alignment
- **Weighted 0–100 scoring system**: score pool adjusts dynamically based on talented cooldowns — Grimoire, Doomguard, and Shards at end each take 5pts from the pool; base components (HoG/Ruination, Dreadstalkers, Shards at cast) redistribute proportionally. Score is capped at 100
- **Scoring thresholds**: Soul Shards at Cast thresholds differ by spec (Diabolist vs Soul Harvester). Perfect requires ≥95 and ≥8 (updated from ≥6) spender casts
- **Cast sequence filmstrip**: shows spells cast during the Tyrant window, with post-window casts shown desaturated for context
- **Score breakdown tooltip**: per-factor contribution shown on the Score stat card
- **Timestamps**: update the Cast Details timestamp to show the entire Tyrant/Apex window and not just the start time
- **Fight-ended window handling**: pro-rates the score based on how far into the window the fight ended; suppresses leftover-shard penalties and misleading feedback messages for those windows
- **Grimoire leeway**: no penalty if Grimoire came off cooldown during the Tyrant window; positive feedback message shown
- **Dreadstalkers timing**: accounts for pre-cast Dreadstalkers that become available again during the 25s window; feedback wording updated
- **Imp Wording**: remove incorrect wording regarding extending Imps
- **Soul Shards tracking fixes**: corrected `resourcechange` listener (`to` vs `by(SELECTED_PLAYER)`), fixed NaN/float display, rounded shard values for display, restored `soulShardTracker.current` for accurate cast-time shard counts. Initialized `SoulShardTracker.maxResource` to `5` to prevent NaN when the first log event is an energize with no classResources. Retroactively derives `shardsOnCast` from the first window spender's pre-spend amount minus gains since Tyrant cast, since combat logs report stale shard counts on free spells
- **Guide feedback improvements**: feedback items below the HoG/Ruination cast summary now render as a bulleted list; Grimoire "available but unused" feedback acknowledges intentional burn phase holds
- **Ruination (Diabolist) support**: casts folded into HoG cast count; labels adjust via `hasTalent`
- **Doomguard CDR analyzer**: tracks cooldown reduction granted by Doomguard
- **Guide restructure**: Demonic Tyrant guide moved into a nested subsection within Cooldowns, matching the Defensives section pattern
- **Support Level and Example Report**: Changed support level to Full and updated example report

- Comments were added to new code to briefly explain the purpose that code does  (mostly for me)

### Testing
- Test report URL: 
- `/report/wN3fvbZyQr7GgcJ2/9-Heroic+Imperator+Averzian+-+Kill+(3:49)/537-Katorrí/standard`
- `/report/FHfng1KdRc98Z6VP/27-Mythic+Imperator+Averzian+-+Kill+(6:53)/1-Katorrí/standard`
- `/report/FHfng1KdRc98Z6VP/9-Heroic+Belo'ren,+Child+of+Al'ar+-+Kill+(4:31)/1-Katorrí/standard`
- Screenshot(s):
image
image
image

image
image

### DIFF
--- a/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
+++ b/src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
@@ -9,6 +9,7 @@ import Events, {
   ApplyBuffStackEvent,
   CastEvent,
   FightEndEvent,
+  RemoveBuffEvent,
   ResourceChangeEvent,
 } from 'parser/core/Events';
 import SoulShardTracker from 'analysis/retail/warlock/shared/resources/SoulShardTracker';
@@ -16,7 +17,7 @@ import SoulShardTracker from 'analysis/retail/warlock/shared/resources/SoulShard
 const DREADSTALKERS_DURATION = 12000;
 const GRIMOIRE_DURATION = 20000;
 const DOOMGUARD_DURATION = 12000;
-export const TYRANT_WINDOW_MS = 25000; //adds 5 seconds to the end to account for Apex window
+export const TYRANT_WINDOW_MS = 25000;
 const DEMONIC_POWER_SPELL_ID = 1276788;
 
 export default class DemonicTyrant extends Analyzer {
@@ -33,8 +34,6 @@ export default class DemonicTyrant extends Analyzer {
   lastDreadstalkersCast = 0;
   lastGrimoireCast = 0;
   lastDoomguardCast = 0;
-  // Last known shard count from SoulShardTracker while inside a Tyrant window.
-  latestShardsInWindow = 0;
   // Tracks gains (in shards) after Tyrant cast to retroactively compute shardsOnCast from the first window spender.
   private gainsBeforeFirstWindowSpend = 0;
   private pendingRetroactiveShardsOnCast = false;
@@ -42,10 +41,23 @@ export default class DemonicTyrant extends Analyzer {
   constructor(options: Options) {
     super(options);
 
+    // Use the Apex talent buff (longer window) if the player has it, otherwise the base Tyrant buff.
+    const hasApex =
+      this.selectedCombatant.hasTalent(TALENTS.DOMINION_OF_ARGUS_1_DEMONOLOGY_TALENT) ||
+      this.selectedCombatant.hasTalent(TALENTS.DOMINION_OF_ARGUS_2_DEMONOLOGY_TALENT) ||
+      this.selectedCombatant.hasTalent(TALENTS.DOMINION_OF_ARGUS_3_DEMONOLOGY_TALENT);
+    const windowBuff = hasApex ? SPELLS.DOMINION_OF_ARGUS_BUFF : SPELLS.SUMMON_DEMONIC_TYRANT;
+
     // Tyrant cast
     this.addEventListener(
       Events.cast.by(SELECTED_PLAYER).spell(SPELLS.SUMMON_DEMONIC_TYRANT),
       this.onTyrantCast,
+    );
+
+    // Window end via buff removal
+    this.addEventListener(
+      Events.removebuff.to(SELECTED_PLAYER).spell(windowBuff),
+      this.onTyrantEnd,
     );
 
     // Hand of Gul'dan and Ruination (Diabolist) — both count as shard spender casts
@@ -78,9 +90,6 @@ export default class DemonicTyrant extends Analyzer {
 
     // Track shard gains — use to(SELECTED_PLAYER) since energize events target the player, not sourced by them.
     this.addEventListener(Events.resourcechange.to(SELECTED_PLAYER), this.onShardGain);
-
-    // Track shard costs for all player casts during the window
-    this.addEventListener(Events.cast.by(SELECTED_PLAYER), this.onAnyPlayerCast);
 
     // Finalize last window at fight end
     this.addEventListener(Events.fightend, this.onFightEnd);
@@ -126,9 +135,6 @@ export default class DemonicTyrant extends Analyzer {
           event.timestamp - this.lastDoomguardCast <= DOOMGUARD_DURATION
         : null;
 
-    // Close the previous window before opening a new one.
-    this.tryFinalizeWindow(event.timestamp);
-    this.latestShardsInWindow = this.soulShardTracker.current;
     this.gainsBeforeFirstWindowSpend = 0;
     // Only needed before the first ever spend — after that the tracker has accurate absolute state.
     this.pendingRetroactiveShardsOnCast = this.soulShardTracker.spent === 0;
@@ -156,12 +162,18 @@ export default class DemonicTyrant extends Analyzer {
     this.currentTyrant = tyrant;
   }
 
+  // Called when the tracked window buff (Dominion of Argus or Summon Demonic Tyrant) expires.
+  onTyrantEnd(event: RemoveBuffEvent) {
+    if (!this.currentTyrant) return;
+    if (this.currentTyrant.shardsAtWindowEnd !== null) return;
+
+    this.currentTyrant.shardsAtWindowEnd = Math.round(Math.max(0, this.soulShardTracker.current));
+    this.currentTyrant.actualWindowDurationMs = event.timestamp - this.currentTyrant.cast;
+    this.currentTyrant = null;
+  }
+
   onSpenderCast(event: CastEvent) {
     if (!this.currentTyrant) return;
-
-    this.tryFinalizeWindow(event.timestamp);
-
-    if (event.timestamp > this.currentTyrant.cast + TYRANT_WINDOW_MS) return;
 
     // Retroactively set shardsOnCast: first spender's pre-spend amount minus gains since Tyrant cast.
     if (this.pendingRetroactiveShardsOnCast) {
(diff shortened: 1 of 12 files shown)


katogaming88:
sorry for the large PR, been working on this a while

src/analysis/retail/warlock/demonology/modules/features/SummonDoomguard.ts
```diff
@@ -0,0 +1,76 @@
...
+      this.onCoreApply,
+    );
+    this.addEventListener(
+      Events.applybuffstack.by(SELECTED_PLAYER).spell(SPELLS.DEMONIC_CORE_BUFF),
+      this.onCoreStack,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.DEMONIC_CORE_BUFF),
+      this.onCoreRemove,
+    );
```
emallson: don't *need* to change this but: since Demonic Core is removed after the cast event, you can simply use `this.selectedCombatant.getBuff(SPELLS.DEMONIC_CORE_BUFF.id)` in your `onDemonboltCast` handler to check for the presence of the buff instead of tracking the stack count yourself

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -47,49 +118,213 @@ export default class DemonicTyrant extends Analyzer {
...
+
+    if (event.timestamp > this.currentTyrant.cast + TYRANT_WINDOW_MS) return;
+
+    if (event.ability?.guid !== DEMONIC_POWER_SPELL_ID) return;
+
+    let stacks = 1;
+
+    if ('stack' in event) {
+      stacks = (event as ApplyBuffStackEvent).stack ?? 1;
+    }
```
emallson:
```suggestion
    if (event.type === EventType.ApplyBuffStack) {
      stacks = event.stack ?? 1;
    }
```

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -21,23 +51,64 @@ export default class DemonicTyrant extends Analyzer {
...
+      this.onDemonicCoreApplyStack,
+    );
+    this.addEventListener(
+      Events.removebuff.by(SELECTED_PLAYER).spell(SPELLS.DEMONIC_CORE_BUFF),
+      this.onDemonicCoreRemove,
+    );
+    this.addEventListener(
+      Events.removebuffstack.by(SELECTED_PLAYER).spell(SPELLS.DEMONIC_CORE_BUFF),
+      this.onDemonicCoreRemoveStack,
+    );
```
emallson: same note about demonic core stack tracking. actually with it occurring twice, its probably worth touching up---at least here, since this module is so big

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -1,16 +1,46 @@
...
+  RemoveBuffStackEvent,
+  ResourceChangeEvent,
+} from 'parser/core/Events';
+import SoulShardTracker from 'analysis/retail/warlock/shared/resources/SoulShardTracker';
 
 const DREADSTALKERS_DURATION = 12000;
-const TYRANT_WINDOW_MS = 20000;
+const GRIMOIRE_DURATION = 20000;
+const DOOMGUARD_DURATION = 12000;
+export const TYRANT_WINDOW_MS = 25000; //adds 5 seconds to the end to account for Apex window
```
emallson: im assuming that the apex is probably a 100% play rate right now, but this doesn't handle a person not taking it.

generally, as the complexity of this module has gotten higher i think it is worth looking at changing how tyrant is tracked.

currently, you do:

- onTyrantCast -> end the previous tyrant (if present), start a new tyrant
- all other event handlers: check if the previous tyrant is still active (based on timestamp)

there are two possible buffs you could use to make this more automatic: https://www.warcraftlogs.com/reports/Faty8NcZ4DmTHgX7?fight=26&type=auras&pins=0%24Separate%24%23909049%24auras-gained%240%240.0.0.Any%24243109094.0.0.Warlock%24true%24243109094.0.0.Warlock%24false%241276166&source=2&ability=265187&view=events

- Dominion of Argus: the Apex talent buff, active for ~25s
- Summon Demonic Tyrant: the base buff for the tyrant being active, active for ~20s

in the `constructor`, you can check for the apex talent(s) used and then set up two listeners:

1. `onTyrantStart` that listens to the applybuff of whichever you're using (assuming apex buff if they have it, tyrant buff if not)
2. `onTyrantEnd` that listens to the removebuff of whichever you're using, handling the "end window" logic

(`onFightEnd` is still needed in case the fight ends before tyrant does)

this would allow you to:

- remove all timestamp checks for tyrant duration (the `if (event.timestamp > this.currentTyrant.cast + TYRANT_WINDOW_MS) return;`s)
- remove (some of) the extra soul shard state (`latestShardsInWindow`; since the window end handler is called when the window ends, you can just check `this.soulShardTracker.current` immediately when it ends).

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
```diff
@@ -47,49 +118,213 @@ export default class DemonicTyrant extends Analyzer {
...
+        ? this.lastDoomguardCast !== 0 &&
+          event.timestamp - this.lastDoomguardCast <= DOOMGUARD_DURATION
+        : null;
+
+    // Close the previous window before opening a new one.
+    this.tryFinalizeWindow(event.timestamp);
+    this.latestShardsInWindow = this.soulShardTracker.current;
+    this.gainsBeforeFirstWindowSpend = 0;
+    // Only needed before the first ever spend — after that the tracker has accurate absolute state.
+    this.pendingRetroactiveShardsOnCast = this.soulShardTracker.spent === 0;
```
emallson, line 141: this state should also be available once any resource change events are seen. you could assume starting the fight with 3 shards but i know people engage in prepull shenanigans from time to time

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -142,38 +361,140 @@ function DemonicTyrantGuide(): JSX.Element | null {
...
+                        >
+                          cast after the tyrant window ended
+                        </div>
+                        <SpellSequence casts={postWindowCasts} iconSize={40} />
+                      </>
+                    )}
+                  </>
+                ),
+              };
+            })()
```
emallson: do not use an IIFE (`(() => { ... })()`) for this. it looks like you used it for the ability to use variable declarations in the body. instead, use a `let`

```tsx
let additionalContent = null;
if (sequenceEntry) {
  // ...
  additionalContent = ...;
}
```

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -12,17 +12,152 @@ import {
...
+    label: string;
+    score: number;
+    max: number;
+    displayScore?: number;
+    displayMax?: number;
+  }[];
+}
+
+// Computes a weighted 0–100 score for a single Tyrant window across spender casts, cooldown usage, and resource management.
+function scoreTyrantWindow(cast: TyrantCastData, isDialobist: boolean): ScoreBreakdown {
```
emallson, line 36: this method is not merge-able as it stands. combining metrics to make a qualitative score is okay, the way it is done is messy.

the main issue is simply the nested ternary expressions. ternaries are fine. sometimes *shallow* nesting of them is fine. however, these go too far.

as an example:

```tsx
  const shardsOnCastScore = isDialobist
    ? Math.min(cast.shardsOnCast, 5)
    : cast.shardsOnCast === 2
      ? 5
      : cast.shardsOnCast <= 1
        ? 4
        : cast.shardsOnCast === 3
          ? 3
          : cast.shardsOnCast === 4
            ? 1
            : 0; // 5 shards
```
why does the `shardsOnCast` logic invert partway through this? you get:

- 1 point per shard as diabolist
- as a non-diabolist, you get:
   - 2 shards: 5 points
   - 1 or 0 shards: 4 points
   - 3 shards: 3 points
   - 4 shards: 1 point
   - 5 shards: 0 points

i would encourage you to simplify the scoring as well, though i won't *mandate* it. complicated scores produce surprising results---and not the good kind of surprise. keeping it simple will be more predictable and understandable. you might consider simplifying the qualitative logic to:

- Perfect: 8 HoGs + 2 pet cooldowns (Grimoire + Dogs)
- Good: 7+ HoGs + 1 pet cooldown
- Okay: 5-6 HoGs + 1 pet cooldown
- Bad: anything else

doing something like this would allow cutting most of the score-related code.

note that all of the info about shards is---in essence---double-counting the score for HoG. if you don't have the shards on cast, you can't hit the full 8 HoGs (I assume). if you end with unspent shards, you get fewer HoGs as well.

buuut the *main thing* to address with this method is organization and clarity. clarity can be addressed by simplifying the underlying scoring method, or by putting extra effort into making the complicated scoring method clearer.

emallson (changes requested):

src/analysis/retail/warlock/demonology/modules/features/DemonicTyrant.ts
katogaming88: So I had it set up for any resource change events. But on some fights before any shards were spent, i was getting NaNs for the Shards at Cast object. It was really only one fight or so (in raid) that was causing this issue, and yeah, i didn't want to always assume the player had 3 shards at the start of the fight (player could have forgotten to summon their demon and thus had to use a shard right before boss was pulled, or some other shenanigans lol)

katogaming88: I meant to add this link in my comment as well. At the start of the fight, WCL doesn't show the current shard count, and it shows as 0, so thats why i was using this method

https://www.warcraftlogs.com/reports/gcDwqAnKTbh64v8t?fight=2&source=1&type=resources&spell=107

emallson: okay, that's weird. imo keep your working solution then, since this just seems like the log being weird and not a wowa bug

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
katogaming88: Hopefully the changes I made resolve some/most of the issues you brought up on this suggestion.

The shard stuff as non diablolist was because Soul Harvester Demo gains 3 shards when casting tyrant. So when overcapping on shards, thats why the `shardsonCast` logic inverts. The path was more of the score you got, so it went from 5>0

src/analysis/retail/warlock/demonology/modules/guide/DemonicTyrantGuide.tsx
```diff
@@ -446,52 +436,52 @@ function DemonicTyrantGuide(): JSX.Element | null {
...
+                      </div>
+                      <SpellSequence casts={postWindowCasts} iconSize={40} />
+                    </>
+                  )}
+                </>
+              ),
+            };
+          }
+          return additionalContent;
+        })(),
```
emallson: the changes are good except for the IIFE here. what i meant was that you'd essentially lift this calculation up to like ~350, outside the `return` statement. like:

```tsx
// currently Line 351
const perf = scoreToPerf(
  score,
  scoreBreakdown.totalSpenderCasts,
  scoreBreakdown.maxExpectedCasts,
);

let additionalContent;

if (sequenceEntry) {
  // ...
  additionalContent = ...;
}

return {
  // ...
 additionalContent,
};

emallson (approved):

PR #7838, opened 2026-04-20
Midnight 12.0.5 talent regen

Krealle:
### Description

generated talents for 12.0.5 and patched the places that were breaking. seems to be overall pretty minor so shouldn't be any breaking changes


Pewtro (commented):
looks good from hunter pov

PR #7861, opened 2026-04-26
fix(components): bug where if mitigation was >=1

YepoBebop:
If mitPct was >=1 (like Aspect of the Turtle), it would calculate incorrectly and show there was 0 mitigation.


emallson (commented):
immunities do not provide damage values. they have `missType` set and `amount` of 0, with no `unmitigatedAmount`. so while this patch won't break anything, it also won't *do* anything

PR #7863, opened 2026-04-27
[brewmaster] Track Elevated Stagger on Overview Stagger Pool chart

renanthera:
### Description

Mark purify events in Overview Stagger Pool chart green if cast while Elevated Stagger is applied or High Tolerance isn't talented, otherwise red.
Add text dynamically to describe Overview Stagger Pool chart marking behaviour.
Allow spell linking of Elevated Stagger buff in Stagger Pool Section text.

### DIFF
--- a/src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
+++ b/src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
@@ -17,6 +17,7 @@ import AutoSizer from 'react-virtualized-auto-sizer';
 import StaggerPool from '../core/StaggerPool';
 import PurifyingBrew from '../talents/PurifyingBrew';
 import HighTolerance from '../spells/HighTolerance';
+import { OkColor } from 'interface/guide';
 
 interface StaggerEvent {
   timestamp: number;
@@ -153,6 +154,7 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
             filled: true,
             size: 60,
             opacity: 1,
+            strokeWidth: 1,
           },
           transform: [
             {
@@ -172,7 +174,14 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
                 test: 'datum["isElevated"] || datum["isElevated"] == null',
                 value: '#00ff96',
               },
-              value: '#c41f3b',
+              value: OkColor,
+            },
+            stroke: {
+              condition: {
+                test: 'datum["isElevated"] || datum["isElevated"] == null',
+                value: undefined,
+              },
+              value: 'black',
             },
             tooltip: [
               { field: 'amount', title: 'Amount Purified', format: '.3~s' },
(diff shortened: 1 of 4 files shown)


src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -213,6 +222,13 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
         };
       });
 
+      const purifyEvents = this.deps.pb.purifies.map((point) => ({
+        ...point,
+        isElevated: this.selectedCombatant.hasTalent(talents.HIGH_TOLERANCE_TALENT)
+          ? this.deps.ht.uptime.some((e) => e.start < point.timestamp && e.end > point.timestamp)
```
emallson:
```suggestion
          ? this.deps.ht.uptime.some((e) => e.start < point.timestamp && e.end >= point.timestamp)
```

to handle removals at the exact timestamp of the cast

Image

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
```diff
@@ -165,6 +167,13 @@ class StaggerPoolGraph extends Analyzer.withDependencies({
               type: 'quantitative' as const,
               title: null,
             },
+            color: {
+              condition: {
+                test: 'datum["isElevated"] || datum["isElevated"] == null',
+                value: '#00ff96',
+              },
+              value: '#c41f3b',
```
emallson: im going to adjust this color. partly for contrast reasons, and partly because purifies outside of elevated stagger are not definitionally *bad* when using HT

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
emallson: this is also quadratic. going to pull in `StateHistory` to make it `O(nlogn)` instead to avoid potential issues in M+

emallson (commented):
i have this pulled up locally right now and am gonna make a couple quick tweaks before merging

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
renanthera: I was originally using the red used in most other places, but I found it (subjectively) relatively poor contrast.

I then tweaked mark opacity, which also helped a lot, but didn't revisit that color.

src/analysis/retail/monk/brewmaster/modules/features/StaggerPoolGraph.tsx
renanthera: I was tempted to linearize it, but javascript doesn't really have the necessary tools to do so in a non-painful way. Thanks for the improvement.

PR #7864, opened 2026-04-27
[Ret] Spellbook, statistics and buffs

Texleretour:
### Description

**Timeline/Spellbook fixes**
- Add Wings, Divine Purpose and Art of War to the buffs timeline
- Remove Crusading strikes from the cast timeline
- Switch to Execute Helper to manage Hammer of Wrath, now appears on the cast timeline as an actual spell with a GCD

**12.0.5 Bump**
**New default log**
**Remove a few old things**
**Update some constants**

I am very unhappy with the way Hammer of Wrath is handled. I can't find another mechanic that is exactly the same: replaces Judgment with Hammer of Wrath for the whole Wings window so doesn't expect only 1 cast like Warriors' Thunder Blast. Has to actually stop Judgment cast expectations unlike most spells that replace a cooldown and not a rotational spell (Priests' Void Volley...)

I still think it is more informative that way.

I'm confident I messed something up with Crusade haste tracking and would love some feedback.

### Testing

- Test report URL: `/report/zKy7YmB8ahtL4cNG/51-Heroic+Fallen-King+Salhadaar+-+Kill+(4:44)/375-Ancalainur/standard/overview`


src/analysis/retail/paladin/retribution/modules/talents/Crusade.tsx
```diff
@@ -1,80 +1,60 @@
...
-  get badGlobalPercent() {
-    return this.badFirstGlobal / this.abilityTracker.getAbility(TALENTS.CRUSADE_TALENT.id).casts;
+  #updateCrusadeStacks(event: CastEvent): void {
+    // For some reason HoL acts as if it consumed 9 Holy Power - as of 12.0.5
+    const bonusHastePercent = event.ability.guid === SPELLS.HAMMER_OF_LIGHT.id ? 0.18 : 0.06;
+    if (this.#currentCrusadeHaste + bonusHastePercent > 0.2) {
+      this.#currentCrusadeHaste = 0.2;
+    } else {
+      this.#currentCrusadeHaste += bonusHastePercent;
+    }
```
emallson, line 45: this seems fine to me. it might be clearer to track the total HP spent during wings and then do `const hasteBonus = Math.min(CRUSADE_MAX_HASTE, CRUSADE_HASTE_PER_HP * this.#currentCrusadeHolyPowerSpent);` in the haste update.

but that is a nit, i think.

emallson (approved):
judgment cast efficiency is a tough one.

another way you could potentially handle it is to treat HoW and Judgment as two spells that share a CD (by setting `spell: [SPELLS.JUDGMENT.id, SPELLS.HAMMER_OF_WRATH.id]` in `Abilities`).

this would give you an accurate view of Judgment cast efficiency overall, though it would not give you a good view of HoW. HoW could then be handled using timeline windows, dedicated statistics, or cast sequences.

regardless, i think the current handling looks entirely reasonable, given the complexity of it

PR #7866, opened 2026-04-27
Improvements for Druid Balance

victor-brossard:
## Description

Improvements for Druid Balance
- Fix some abilities (GCD, CD, suggestions...)
- Fix the Haste modifiers
- Add missing buff `Cenarius might`
- Add spells list generated by the tool
- Override Spell Usable to handle CD reduction from `Control of the Dream`

### Test report URL
[Report at 100% performance](https://wowanalyzer.com/report/vFTDkdC9yYNBXghM/22-Mythic+Imperator+Averzian+-+Kill+(4:28)/24-M%C3%AEnd%C3%AF/standard/overview)
[Report at 16% performance](https://wowanalyzer.com/report/h9cxgWvBG2pbf8Ta/126/60)

### globalCooldown errors
- before
image

- after
image

Remaining errors are all due to `GCD from Force of Nature still active`, which is weird because in-game testing shows a correct GCD value 🤔

### Cooldowns efficiency

- before
image

- after
image

The number of possible casts is now correct, it's a 120s CD spell (2 charges) with the first cast benefiting from a 15s CDR because of `Control the Dream`. The following casts can be done during this 514s fight :
- t=0s (1st charge)
- t=20s (2nd charge)
- t=105s (1st cd reduced)
- t=225s (regular 120s CD)
- t=345s (regular 120s CD)
- t=465s (regular 120s CD)

⚠️ But the efficiency score needs to be discussed : I think it should be the % of casts out of possible casts. Currently the score is weird... [I have asked in Discord](https://discord.com/channels/316864121536512000/316864121536512000/1499758543946121238).

### Control of the Dream statistics

- before
image

- after
image
image

### Changelog

image


src/parser/shared/modules/CastEfficiency.tsx
```diff
@@ -319,8 +319,17 @@ class CastEfficiency extends Analyzer {
...
+    // So if you could restore 3 full CDs of 60s during a fight of 185s, it means that you can cast at
+    // 1) t=0s
+    // 2) t=60s
+    // 3) t=120s
+    // 4) t=180s
+    // which is 4 times. But in this case rawMaxCasts=3.08333.
+    // The solution is thus to count the CDs that can be fully restored (here 3) and
+    // add 1 because we are counting the number of casts and not the number of CDs.
+    // (same as counting the number of pickets based on a picket fence length and the total length of the fence).
+    const maxCasts = Math.floor(rawMaxCasts || 0) + 1;
```
Pewtro, line 332: Your change does the same as what it already did. 
Math.ceil(3.08) => 4

src/parser/shared/modules/SpellUsable.tsx
```diff
@@ -269,6 +269,12 @@ class SpellUsable extends Analyzer {
     } else {
       // trigger an end cooldown and then immediately a begin cooldown
       // we're treating this as a missed natural CD expiration, so pass true to 'reset cooldown'
+      console.warn(
```
Pewtro: If we do want this, haven't really thought it through, it should at least be hidden behind the DEBUG flag

src/parser/shared/modules/CastEfficiency.tsx
victor-brossard: Yep the result is the same, but based on the comment I have deleted, the previous logic was confusing and hard to understand.

src/parser/shared/modules/SpellUsable.tsx
victor-brossard: The idea was to mimick the WARN in SpellUsable `Ability is missing from spellbook (check out spellUsable in the debug annotation view for more info)` which I found super useful (and it does not require to enable DEBUG flag on this file.
But if you think it should be hidden behind a flag, tell me and I will modify !

emallson: the "cooldown error happened" info is shown in the debug view (go to about -> view debug info). [example here](https://wowanalyzer.com/report/TwgxtVCM19BK6AhW/29-Heroic+Chimaerus,+the+Undreamt+God+-+Kill+(2:14)/55-Setthh/standard/debug). does this add something that isn't covered by that view? or just something you didn't find?

victor-brossard: Yes it helped me when running unit tests, the message shows up in the output.
But kudos for the debug info view, super useful !

Again, if you think it should be hidden behind a flag, tell me and I will modify it 🙏

emallson: since this appears to be duplicative with the debug view, yes it should be behind a flag. we've had major pain with `console` spam in the past (still kind of do, but it isn't as bad as it was), to the point that the console was virtually unusable. so this `console.warn` should be off by default

victor-brossard: I totally understand the spam :)
I have pushed a change

> `DEBUG &&
>         console.warn(
>           'Cooldown began at %ims for Spell Id=%i but it was already on cooldown (until %ims).',
>           triggeringEvent.timestamp,
>           spellId,
>           cdInfo.expectedEnd,
>         );`

src/analysis/retail/druid/balance/spell-list_Druid_Balance.retail.ts
emallson, line 1: this file doesn't appear to be imported by the current implementation. delete the file if it is unused. (if i just missed the usage, my bad)

emallson (changes requested):
overall, the code here looks fine. will wait for putro to give his 👍 re: his comments. the tests are nice, especially for a wonky effect like this.

src/analysis/retail/druid/balance/spell-list_Druid_Balance.retail.ts
victor-brossard: Thanks 🙏 

> re: his comments

Sorry I'm not familiar with this, what does it mean ?

> this file doesn't appear to be imported by the current implementation. delete the file if it is unused. (if i just missed the usage, my bad)

You're right, I used it as a reference on the side but it is not referenced. I have deleted it !

victor-brossard: I will wait for @Pewtro 's feedback then !

Pewtro (approved):
I personally prefer the old way in CastEfficiency but won't let it block anything

victor-brossard:
> I personally prefer the old way in CastEfficiency but won't let it block anything

Thanks @Pewtro 🙏 
I feel like current code was misleading (based on my experience + the existing comment in the code) but I know it's a bit of a personal taste here !

